PR #9, opened 2017-01-03
add support to search through all repos of an organization

ryanbaxendale:


ryanohoro:
I spent a better part of today writing a patch for this, but ended up with a much meatier fork. Your solution is simple though, I like it.

https://github.com/ryanohoro/truffleHog/

dxa4481:
Thanks for the PR. I like how truffleHog is currently github agnostic (it's a Git tool) so what I did was create a scripts directory, made truffleHog callable like a library and put your code in there.

https://github.com/dxa4481/truffleHog/blob/master/scripts/searchOrg.py

PR #10, opened 2017-01-03
ignores duplicate diffs caused by branches

Stuart4:
Implements the second item in the wishlist:
> - Don't rescan diffs if already looked at in another branch


dxa4481:
Thank you!

PR #23, opened 2017-01-04
replaced bcolors class with bcolors dictionary

bandrel:
Replaced bcolors class with bcolors dictionary.


bandrel:
I know it's a trivial and a common occurrence, but because the class does not contain any methods and instances of the object are not going to be created it's cleaner to use a dictionary.

PR #28, opened 2017-01-10
Creating a PyPI package and command line tool out of truffleHog

kepoorhampond:
Pretty much a setup so that you can easily upload the package to PyPI. I would myself, but that would be really lame because then it would look like I had done it.

To upload it, you'll need to run these commands in the repos root directory:
```
python setup.py sdist
python setup.py bdist_wheel
python setup.py sdist bdist_wheel upload
```


setup.py
```diff
@@ -0,0 +1,18 @@
...
+    name='truffleHog',
+    version='1.0.0',
+    description='Searches through git repositories for high entropy strings, digging deep into commit history.',
+    url='https://github.com/dxa4481/truffleHog',
+    author='Dylan Ayrey',
+    author_email='dxa4481@rit.edu',
+    license='GNU',
+    packages =['truffleHog'],
+    install_requires=[
+        'GitPython == 2.1.1'
```
centralniak: Is exactly 2.1.1 required? 

A tool that would be installed globally should rather not enforce a particular version IMO.

kepoorhampond: It's what the `requirements.txt` requires, so that's what I put in. I haven't tested it with any other versions, but I will once I get home.

PR #29, opened 2017-01-11
Diff can contain UTF8

jeremydenoun:
Patch print printableDiff for display utf8


dxa4481:
Thank you, I fixed this with another PR.

Issue #27, opened 2017-01-09
Can't decode 0xa9, which is the copyright symbol


dullroar:
This is actually more of a problem with Python's UTF8 code, I think. But if you run the trufflehog script against [this repo](https://github.com/dullroar/TW5-atomfeed), you'll get the following error:

```
Traceback (most recent call last):
  File "/home/lehmer/Sandbox/trufflehog.py", line 130, in <module>
    project_path = find_strings(args.git_url)
  File "/home/lehmer/Sandbox/trufflehog.py", line 95, in find_strings
    printableDiff = blob.diff.decode()
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xa9 in position 96: invalid start byte
```

This is because an older version of the license file had a copyright symbol (0xa9, which is the rendered value of the HTML entity &copy;, for example). I figured it out by putting a `print(blob.diff)` right before line 95.

floatingatoll:
This also trips up on other non-utf8 values (0xA0). It would be most useful if the script proceeded rather than terminating unexpectedly.

dxa4481:
I think I fixed it. Reopen if it appears again.

PR #34, opened 2017-02-06
Gracefully Handle Failure to decode non 'utf-8' codec

### LINKED ISSUE #7: UnicodeEncodeError: 'charmap' codec can't encode characters ... character maps to <undefined>
Testing using a private Github repo (soon to be public) caused an error. It identified a few SHA1 hash strings in two different commits, then throws the following error:

```
Date: 2016-12-16 13:10:27
Branch: master
Commit: Nameserver overhaul (#17)

* go changes to support nameservers_from
... omitted ...

Traceback (most recent call last):
  File "truffleHog.py", line 108, in <module>
    find_strings(args.git_url)
  File "truffleHog.py", line 97, in find_strings
    print(printableDiff)
  File "c:\tools\python2\lib\encodings\cp437.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_map)
UnicodeEncodeError: 'charmap' codec can't encode characters in position 1292428-1292443: character maps to <undefined>
```

Environment is Python 2.7 running on Windows 10 x64 bit. There are 50 changed files in the commit and it doesn't say which file caused the error, but my guess is it doesn't like one of the static.go files that has base64+GZip encoded JS/CSS files. See attached git patch for commit details.

[0001-Nameserver-overhaul-17.zip](https://github.com/dxa4481/truffleHog/files/681656/0001-Nameserver-overhaul-17.zip)

cjbarker:
Fixes  #7 and #27 

Leverage replacement character when decode of unrecognized character encoding encountered and streamlined Python 2+ header encoding.

This is most likely occurring when git commit pulled down is NOT a UTF-8 encoded content-type response.  For example it might be legacy HTML encoded via  iso-8859-1.  Rather than try to determine string's encoding type (incredibly difficult) simply handle decode error gracefully.

Simple example
`'\xa9'.decode('utf-8', errors='replace')
u'\ufffd'`

### DIFF
--- a/truffleHog.py
+++ b/truffleHog.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 import shutil
 import sys
 import math
@@ -8,10 +10,6 @@ import tempfile
 import os
 import stat
 from git import Repo
-
-if sys.version_info[0] == 2:
-    reload(sys)  
-    sys.setdefaultencoding('utf8')
 
 BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 HEX_CHARS = "1234567890abcdefABCDEF"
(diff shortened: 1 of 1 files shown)


dxa4481:
🎆 thank you!

PR #35, opened 2017-02-17
Local path and output file.

comlaterra:
New features:

* Allow to use a local repository.
* Set an output file.

Also improved the README with Run and Output

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ pip install -r requirements.txt
 ```
 
 ## How it works
+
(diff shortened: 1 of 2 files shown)


dxa4481:
Thanks for the PR. The local file feature was described in #17 

With regards to the output file, #14 added json support, I think that's probably preferred for file output. I can add a feature that writes the json to a file, if you don't like redirecting STDIN to a file

PR #44, opened 2017-05-10
Add ability to output findings to a file

adamnfish:
While it's possible to redirect trufflehog's output to a file, this is much less convenient if trufflehog's find_strings is being called from other Python code.

This change adds an argument to find_strings that specifies a file to write the output to, defaulting to stdout. It also adds an argument to main's parser so this can be controlled when using trufflehog from the command line.


dustin-decker:

I recommend using the json output and redirect stdout to a file to solve this.

PR #45, opened 2017-05-10
Only output details of commits that *introduce* HE strings

adamnfish:
High-entropy strings that appear in the "context" of diffs, or those that are removed by a commit aren't particularly interesting when you're searching a repository's history for secrets. This change means trufflehog only searches diffs for lines that introduce potential secrets, dramatically reducing the amount of noise to search through in trufflehog's output.


amiller-isp:
Couldn't you do this will an allow regex ?
`"Only added lines": "regex:^+.*"`

dustin-decker:

PR #49, opened 2017-06-19
Update __init__.py

jingpengw:
this fixed the issue https://github.com/dxa4481/truffleHog/issues/39
not sure that whether this will effect python2 or not.


dxa4481:
Amazing, ty. That import was used in a previous build but I don't even need it anymore. Doesn't seem to break python2

Issue #42, opened 2017-04-20
Prints UnicodeEncodeError in some repos


moritz:
For some repos, trufflehog prints this message:
```
Traceback (most recent call last):
  File "/home/mlenz/src/truffleHog/venv-trufflehog/bin/trufflehog", line 9, in <module>
    load_entry_point('truffleHog==1.0.2', 'console_scripts', 'trufflehog')()
  File "build/bdist.linux-x86_64/egg/truffleHog/truffleHog.py", line 20, in main
  File "build/bdist.linux-x86_64/egg/truffleHog/truffleHog.py", line 135, in find_strings
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 72: ordinal not in range(128)
```

I can't make the repos available publicly, sorry. This is with python 2.7.9.

dxa4481:
Yeah.... this is a heisenbug that appears every now and again. If you have a repeatable public repo this occurs on, let me know so I can work on it.

doomnuggets:
@dxa4481 seems to be coming from the branch name.
You can test this against my repo located here:
https://github.com/doomnuggets/tst.git

B3nac:
@doomnuggets after applying your fix I get:

```
Traceback (most recent call last):
  File "truffleHog.py", line 143, in <module>
    main()
  File "truffleHog.py", line 20, in main
    output = find_strings(args.git_url, args.output_json)
  File "truffleHog.py", line 136, in find_strings
    print(printableDiff)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 42: ordinal not in range(128)
```
This happens only when I try to pipe the output > .txt. Trying to figure out what to do next.

dxa4481:
Yeah I've observed that a few times too. The temporary work around is to use the json output. I'm not sure why sometimes piping to a file wonks it up.

doomnuggets:
@dxa4481 @B3nac @moritz 
Could you guys check if my latest commit solves the issue?
https://github.com/dxa4481/truffleHog/pull/51/commits/0aade8ab264e371fc40cf94d400cfc511e8f88c0

AntonKuzminRussia:
No, fresh repo has this issue. May be it help: i fix it on line 136, replaced 
print(printableDiff)
by 
print(printableDiff.encode('utf8', 'ignore'))

dxa4481:
I think I fixed it. Reopen if it appears again.

PR #51, opened 2017-08-07
Update truffleHog.py

doomnuggets:
This might solve #42


dxa4481:
Fixed this with another PR, thank you though.

PR #65, opened 2017-11-28
--fail flag added to exit(1) when issues found

sebolabs:
--fail=False/True flag (default = False) to fail when issues found


dxa4481:
Status code's are now thrown. If you think this should be configurable, feel free to re-open

PR #70, opened 2017-12-08
Update regexChecks.py

carnal0wnage:
added a few regex additions.  The ones with *[\'|"]  weren't working for me but i didnt remove them i just created some additional ones that did work for me

i created a repo with fake secrets to test

https://github.com/carnal0wnage/fake_commited_secrets

### DIFF
--- a/truffleHog/regexChecks.py
+++ b/truffleHog/regexChecks.py
@@ -2,13 +2,24 @@ import re
 
 regexes = {
     "Internal subdomain": re.compile('([a-z0-9]+[.]*supersecretinternal[.]com)'),
+    "Generic AppSecret": re.compile('[a|A][p|P][p|P][s|S][e|E][c|C][r|R][e|E][t|T].*.[0-9a-zA-Z]{32,45}'),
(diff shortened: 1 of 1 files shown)


dxa4481:
I've merged the new ones, thanks for the additions. The ['|"] part ensures that the secret is rapped in quotes. This is to improve signal. A simple example with one of your newer ones, a github URL would likely flag on [g|G][i|I][t|T][h|H][u|U][b|B].*[0-9a-zA-Z]{35,40} such as https://github.com/google/googletest/commit/10ef1d9a1e7e6a0b3701d262cf1b889620fdbbc9

To prevent this URL from flagging, we ensure only secrets rapped in quotes will flag. This means we'll miss out on some real secrets, but we increase our signal to noise ratio in doing so.

carnal0wnage:
thanks for the explanation on the ['|"] part.  I dont see secrets wrapped in "" too frequently myself.  All the ones in that fake repo are ones i found on github or pastebin and they werent matching.  I appreciate the merge of the others!

dxa4481:
If you can find another way to catch those secrets without increasing noise too much, I'll merge it. Maybe make sure the string is wrapped by an equals or a colon on the left and a newline on the right?

carnal0wnage:
cool. if i find a better way i'll do a PR / issue / or an email.

thanks!

dxa4481:
@carnal0wnage 👋

PR #71, opened 2017-12-09
fix #69 - multiple regex matches overwrite diff

### LINKED ISSUE #69: The regex functionality is not working as expected
I git cloned the truffleHog repository. Changed my `regexChecks.py` file to look like below:

```
import re

regexes = {
    "Slack Token XOXP": re.compile('xoxp.*'),
    "Slack Token XOXB": re.compile('xoxb.*'),
    "Slack Token XOXO": re.compile('xoxo.*'),
    "Slack Token XOXA": re.compile('xoxa.*'),
    "AWS API Key": re.compile('AKIA.*'),
    "Private key": re.compile('-----BEGIN PRIVATE KEY-----.*')
}
```

I then installed the libraries required to run the tool by typing `pip install -r requirements.txt`. My requirements.txt file looked like below:

```
GitPython==2.1.5
gitdb2==2.0.2
smmap2==2.0.2
```

Finally, I ran the tool by typing - `python truffleHog.py --regex --entropy=False https://github.com/secretuser1/secretrepo.git`

It printed out the Private Key, Slack Token XOXP and Slack Token XOXB. It should have also printed out the AWS key here - https://github.com/secretuser1/secretrepo/blob/master/secretfile.txt#L2 but it did not, even though the regex is present. 

Any idea why?

milo-minderbinder:
Previously, the first matching regex pattern in a call to `regex_check()`
would overwrite the `printableDiff` and thereby cause any remaining
regexes that _should_ match to fail.

This PR also removes example regex patterns, and uses raw strings in all patterns to avoid issues with special and escaped characters.

### DIFF
--- a/truffleHog/regexChecks.py
+++ b/truffleHog/regexChecks.py
@@ -1,14 +1,17 @@
 import re
 
 regexes = {
-    "Internal subdomain": re.compile('([a-z0-9]+[.]*supersecretinternal[.]com)'),
-    "Slack Token": re.compile('(xox[p|b|o|a]-[0-9]{12}-[0-9]{12}-[0-9]{12}-[a-z0-9]{32})'),
-    "RSA private key": re.compile('-----BEGIN RSA PRIVATE KEY-----'),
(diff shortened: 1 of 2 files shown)


dxa4481:
I didn't merge this because there's another PR that adds regexes that would have heavily conflicted. Thank you for helping me diagnose the issue though, I've addressed this in #72

PR #81, opened 2018-01-18
use the IGNORECASE flag instead of [] groups

paralax:
for the regexes that use stuff like [Tt][Ww][Ii]... use re.I to specify case insensitive, makes it easier to see and manage


dxa4481:
Hi @paralax thanks for the PR. The reason I didn't do this initially is because I don't want the whole search to be case insensitive. Take facebook for example

```
"Facebook Oauth": re.compile('[f|F][a|A][c|C][e|E][b|B][o|O][o|O][k|K].*[\'|"][0-9a-f]{32}[\'|"]')
```

In this case, we actually don't want upper case letters. Our signal is higher staying case sensitive. That said, some of the regexes don't matter because we DO want upper case characters. I just left the brackets in for everything for consistency

PR #83, opened 2018-01-24
[ BUG FIX ] The first commit was missed

kings-way:
If some code exists in a repo from the first commit and stays unchanged,  or there is only one commit in a repo, trufflehog are not going to check the code in the first commit.

Here is a simple solution :)

### DIFF
--- a/truffleHog/truffleHog.py
+++ b/truffleHog/truffleHog.py
@@ -13,6 +13,7 @@ import json
 import stat
 from defaultRegexes.regexChecks import regexes
 from git import Repo
+from git import NULL_TREE
 
 def main():
     parser = argparse.ArgumentParser(description='Find secrets hidden in the depths of git.')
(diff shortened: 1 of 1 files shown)


dxa4481:
Hi @kings-way I've merged this into dev. There was a problem I was having, basically when you casted the entire commit tree to a list, it can consume lots of memory if the commit tree is huge. To solve this I've implemented a hacky work around you can see here https://github.com/dxa4481/truffleHog/blob/dev/truffleHog/truffleHog.py#L240

Basically, because we can't diff NULL_TREE against current diff, I flip them around for the first diff. I'm not super happy with this solution, because it pulls the wrong commit message and date for the first commit, and shows the diff in reverse, but it's at least showing results.

Do you have any ideas for how to fix this without casting the entire commit tree to a list?

kings-way:
Hi @dxa4481 
Actually, your code at this time does not work as well.....  For two scenarios：
1.  There is only one commit, your code works well.
2.  There are more than one commit, but the sensitive string exists from the first commit and never be changed. Which you still miss for now ......

For my own project, I got a very different solution. 
There is  no need to diff between every two commits.  In fact, we could diff every commit with NULL_TREE, and this works actually.

A sample code from my private project:

`
    for remote_branch in repo.remotes.origin.fetch():
        repo.git.checkout(remote_branch.name)
        for commit in repo.iter_commits(max_count=max_depth):
            diff_list = commit.diff(NULL_TREE, create_patch=True, unified=0)
            for item in diff_list:
                diff_text = item.diff.decode('utf-8', errors='replace')
                if diff_text.startswith("Binary files"):
                    continue
                # do something
            if str(commit) == since_commit:
                break
`

Issue #87, opened 2018-02-15
Feature Request: exit code if secrets found


calve:
Hi !

In order to batch truffleHog, it would be great to have different exit code when truffleHog find suspicious string and when it does not.

I would suggest to return 0 when the code is clean, and 1 otherwise.

calve:
fwiw here is my oneliner to exit whit a failure code when secrets are found:
 `trufflehog /path/to/git | ifne sh -c \"cat - && false\"`

 (`ifne` is in `moreutils` on debians)

calve:
just opened a PR to address this

dxa4481:
Done

PR #88, opened 2018-02-19
Allow to set a custom exit code when secrets are found

calve:
address #87


calve:
I'm open to suggestions for a better naming

calve:
I see that https://github.com/dxa4481/truffleHog/pull/65 also address this

dxa4481:
Error codes are now thrown on findings, feel free to re-open if you think this should be configurable

PR #97, opened 2018-04-10
Single character match doesn't need pipes

jeff-minard-ck:


dxa4481:
These rules were pulled out and made into a separate module. Feel free to open up the same PR there https://github.com/dxa4481/truffleHogRegexes/edit/master/truffleHogRegexes/regexes.json

PR #98, opened 2018-04-12
Add back repeated diff scanning prevention

qstearns:
This check was removed while resolving merge conflicts in f2f593a which
appears to have been accidental.

Thanks for you're work on truffleHog. It's a great tool

### DIFF
--- a/truffleHog/truffleHog.py
+++ b/truffleHog/truffleHog.py
@@ -267,6 +267,14 @@ def find_strings(git_url, since_commit=None, max_depth=1000000, printJson=False,
                 continue
             else:
                 diff = prev_commit.diff(curr_commit, create_patch=True)
+
+            # avoid searching the same diffs
+            hashes = str(prev_commit) + str(curr_commit)
+            if hashes in already_searched:
+                prev_commit = curr_commit
+                continue
+            already_searched.add(hashes)
+
             foundIssues = diff_worker(diff, curr_commit, prev_commit, branch_name, commitHash, custom_regexes, do_entropy, do_regex, printJson)
             output = handle_results(output, output_dir, foundIssues)
             prev_commit = curr_commit


dxa4481:
Hey @qstearns would you mind profiling this on some large repos? I've taken some steps to reduce out of memory issues recently, and I'm worried storing every commit hash in memory may grow out of control quickly. 

If you could scan a few repos that have many branches, and tens of thousands of commits, and report the memory usage, that would be great.

qstearns:
I'd be happy to. Can you point me towards your preferred profiler? I don't come from the Python community, so I'm not totally clear on what you'd normally use.

sainaen:
On a very small repo (~30MB) with lots of small commits (~8000) and _recent_ branches (~500 with very long common history), without a check like this truffleHog takes more than 10 hours on my machine (and it didn’t finish).

I don’t use Python often, so don’t know how to profile it either and my calculations could be wrong, but I think the memory required to store all those strings even for the repository with 100 times more commits would be under 100MB:
```
# number of commits * (49 bytes of overhead on every string + 80 bytes for characters)
800_000 * (49 + 80) = 103_200_000 ≈ 99.18MB
```

With a check similar to this, the time went down to about 6 minutes. (I must note that I’ve added other changes, but they are mostly irrelevant to the commit traversal time.)

Hope my anecdote helps with a decision on whether to merge this change. :slightly_smiling_face:

sainaen:
@dxa4481 
> The one thing I'm kinda iffy on is, if a secret shows up multiple times, folks might want to know everywhere it shows up. Maybe that can be added in the future.

I’m not sure I understand your concern here. This change is supposed to prevent truffleHog from scanning the _same_ pair of commits more than once. If a secret is present in _multiple_ commits it will still be shown multiple times even after its merged.

sainaen:
@dxa4481 
Oh, I see. You’re right, this could be something from the nice-to-have category.

PR #107, opened 2018-05-09
Suppress output when included as python module

### LINKED ISSUE #104: Feature Request: Suppress output when run as an included module
The app works great when run from command line. However, it would be nice if the findings output could be suppressed when included as a module in other python projects that only need access to portions of the built in functionality or the ability to customize the output format. One possible solution is to:

1. Remove the calling of `print_results()` from `diff_worker()`
2. Add a command line argument for enabling/disabling output (on by default)
3. Add logic to `main()` for running `print_results()` when required.

Let me know if you would like me to incorporate the change into a PR.

ray-yin:
Default bevior has not changed. Following changes made:

- Moved `print_results()` out of `diff_worker()` and into `main()`
- Removed `printJson` argument from `diff_worker` & `find_strings`
- Changed `handle_results()` output to appended list of foundIssues
- Updated `searchOrg.py` to output findings by default.
- Various code format changes due to pep8 linting in vscode :-)

Fixes #104

### DIFF
--- a/scripts/searchOrg.py
+++ b/scripts/searchOrg.py
@@ -5,14 +5,20 @@ via https://github.com/dxa4481/truffleHog/pull/9
 import requests
 from truffleHog import truffleHog
 
+
 def get_org_repos(orgname, page):
-    response = requests.get(url='https://api.github.com/users/' + orgname + '/repos?page={}'.format(page))
+    response = requests.get(
+        url='https://api.github.com/users/' + orgname + '/repos?page={}'.format(page))
     json = response.json()
     if not json:
         return None
     for item in json:
         if item['private'] == False:
             print('searching ' + item["html_url"])
-            truffleHog.find_strings(item["html_url"], do_regex=True, do_entropy=False, max_depth=100000)
+            output = truffleHog.find_strings(
+                item["html_url"], do_regex=True, do_entropy=False, max_depth=100000)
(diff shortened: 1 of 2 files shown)


dxa4481:
I like the idea of suppressing output for import, ideally I think it should be a config flag you can turn on and off. The way you implemented things all the issues are printed at the end, and I kinda liked seeing results as you go. Let me see if I can rework things a little

dxa4481:
Just merged in a change that should have the same affect, but preserves print-as-you-go when it runs. Feel free to open an issue if there's something I missed

ray-yin:
Thanks for getting this in so quickly! 

The primary reason I outputted everything at the end is because the default return object for the `find_strings` function only lists the path to a temp file with the actual data and I wanted to the full details as output. I can see value in both methods. 

I've tested an import and it fills my immediate need. It would be nice if:

1.  `diff_worker` outputted the email for the person who made the commit
2. I can set the base directory for where the output findings are stored. This would make cleanup easier for those times when I kill the process. 

Let me know if you would like me to add these items as a new feature request.

As an aside, I really appreciate the performance enhancements you made in the latest version. All my scans use custom regex patterns.  A scan of a repo with 900+ commits and 35 branches that used to take ~50 minutes now completes in ~3 minutes. I've also scanned a repo with 8k+ commits across 44 branches in ~20 minutes.

PR #110, opened 2018-06-11
Bugfix: Fix return of wrong commit hash

maennel:
Returned commit hash pointed the parent commit of the effectively wanted one. 

I.e. if a commit introduced a secret, truffleHog returned that commit's parent's hash value. The comment was from the correct commit, on the other hand. 

Also added a test working as a regression test.


tmsteen:
Appears to fix #113 ?

PR #128, opened 2018-10-01
Add report option

camaya:
Hi,

I added an option that generates a report like this: 

```
{
    "finding": "62cdb7020ff920e5aa642c3d4066950dd1f01f4d",
    "reason": "High Entropy",
    "commits": [
        {
            "date": "2017-04-07 10-47:27",
            "hash": "0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33",
            "authorName": "Foo Bar"
            "authorEmail": "foo@bar.com",
            "file": "src/file.go",
            "branch": "origin/master",
            "commitMsg": "Lorem ipsum dolor sit amet",
            "relevantLines": "+ secret = 62cdb7020ff920e5aa642c3d4066950dd1f01f4d"
        },
        {
            "date": "2017-04-07 10-47:27",
            "hash": "0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33",
            "author": "foo@bar.com",
            "file": "src/file.py",
            "branch": "origin/master",
            "commitMsg": "Lorem ipsum dolor sit amet",
            "relevantLines": "+ secret = 62cdb7020ff920e5aa642c3d4066950dd1f01f4d"
        }
    ]
}
```

It does a couple of things:

1. It only contains information that is relevant to the secrets instead of the full diffs to reduce noise and make it easy to spot them.
2. The findings are grouped so they appear only once and there's an array of commits where that finding was found.

Cheers

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ GitPython==2.1.1
 unittest2==1.1.0
 pytest-cov==2.5.1
 codecov==2.0.15
-truffleHogRegexes==0.0.4
+truffleHogRegexes==0.0.5
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='truffleHog',
-    version='2.0.93',
+    version='2.0.94',
     description='Searches through git repositories for high entropy strings, digging deep into commit history.',
     url='https://github.com/dxa4481/truffleHog',
     author='Dylan Ayrey',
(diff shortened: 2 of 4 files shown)


camaya:
The build is failing because Travis is configured with Python 2.7 and the script is importing a 3.X module which is easy to fix but I'm wondering what versions should be supported.

Correct me if I'm wrong, but having a look to the docker file (https://github.com/dxa4481/truffleHog/blob/dev/Dockerfile#L1), I understand it should be Python 3

dustin-decker:

PR #145, opened 2019-02-12
Only delete repositories that weren't provided with the repo_path argument.

bobthesecurityguy:
This checks if the repo_path argument was passed in before attempting to delete the repository directory after the run. If the user cloned their own repo, they should likely decide when to delete it.


dustin-decker:

PR #146, opened 2019-02-12
Updated command line arguments to make git_url optional if repo_path is used. 

bobthesecurityguy:
Uses argparse's mutually exclusive group to ensure that the user specified either git_url or repo_path.


dustin-decker:

PR #148, opened 2019-03-28
Initial take on trimming down output to the essential data

grzegorznowak:
Hey man,
we found our self getting lost in the amount of text we were getting back from the tool. Albeit useful in general, it was a blocker for our massive repo.
Attached version that adds `--contextify` flag  to considerably reduce the amount of output served.
Cheers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /truffleHog.egg-info/
 **/__pycache__/
 **/*.pyc
+/.idea
+/pyenv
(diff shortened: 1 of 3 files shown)


dustin-decker:

grzegorznowak:
oh ok, thanks @dustin-decker . You actually just reminded me I used to use the tool as a standalone sniffer back in the day. Now with all sort of pipelines in place it's indeed a good thing to have. Will take a look 👍🏾 
3-years-later-way-back-machine, open source in the nutshell ;)

cheers!

PR #149, opened 2019-04-01
Add option to not return the secrets value when one gets found

Mathdon:
Hello. 👋 
We were running this tool in our CI/CD pipeline and did not want the value of the secret printed to the screen when one was found.
I have added some functionality to allow this by passing in 'yes', or 'secret' to a --suppress_output parameter.
passing it yes will return nothing and passing secret will return everything apart from the secret.


dustin-decker:

PR #166, opened 2019-06-18
Avoid handling first commit if since_commit was reached

danielfn:
From running truffleHog in many of our repos we see that, although the `since_commit` hash is properly reached, secrets from the very first commit are still being printed out. I understand that this is not the intended behaviour with the `since_commit` option, therefore this patch.


dustin-decker:

Issue #137, opened 2018-12-17
Feature request: Ignore false-positives


iammarco:
How about defining regex that are false positives so those can be ignored?

L1lith:
Yes please. And maybe file paths regex?

electrofelix:
I think it would be useful to support both literal and regex patterns to be added to an 'allowed' list that can be ignored similar to `git-secrets` but allowing for possibility to display when these occur as well. Adding literals via the tool to have it automatically escape them is just a nice convenience, though being able to know which are literals to output what would be ignored might mean that storing them separately in rules would be useful.

Ignoring entire files does open the door to adding files blindly and then never spotting that a valid secret got added to the excluded file in the future or never running with the entropy scan enabled because whitelisting too many files reduces the usefulness

@L1lith is the false positibe file paths regex already covered by the exclude-patterns.txt?

gaurav-mmt:
As we perform scan and identify some false positives. Please let me know if i can at these false positives identified in first scan ignored as part of further scans on same repositories. Something like literal in git-secrets.

alishasinghania:
Is there any update on this how we can resolve the false positive issue?

davidasnider:
I’m interested in this feature as well and would contribute a PR for the functionality.  I’m thinking something as simple as a comment at the end of a line `# NotASecret` in the code and then a flag for truffleHog `—ignore NotASecret`.  Do the maintainers support that approach?

supermitch:
https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

Along with @davidasnider 's comment, you could even respect Flake8's `# noqa` comments instead of introducing a new one.

dxa4481:
Yeah, these are some good ideas. Let me see if I can do something like the suggested, I'll have to think about it becuase it has to be language agnostic

dustin-decker:
Hey there, we've just released the next major version of TruffleHog! 

This is a duplicate of #22 which we are tracking.

It is a complete rewrite that scans more data sources and now supports detecting and verifying over 600 credentials.
Please check it out when you can.

https://trufflesecurity.com/blog/introducing-trufflehog-v3

We're no longer maintaining v2 so I am closing this issue.

PR #175, opened 2019-07-28
With regex, exclude lines from entropy scaning

claer:
Add the option --entropy-exclude-regex ENTROPY_EXCLUDE_RE_FILE to exclude some lines from entropy scanning.
It can be an option to Feature request: Ignore false-positives #137


iamsudip:
@dxa4481 Hi, can we merge this? We see lot's of false positives, specially when we write unit tests.

ohthehugemanatee:
tbh I'm not sure how anyone can use this for CI _without_ the ability to ignore lines. High entropy strings pop up all over the place: sha sums, URLs, inline pubkeys...

PR #181, opened 2019-09-12
--since_date feature is added with some fixes(optimizations) below.

caglarsayin:
- Null tree comparision excluded for the search with --since_date and --since_commit attributes.
- epoc date from since_date attribute is used in git clone and git iter_commits and optimized the performace and disk space


dustin-decker:

PR #187, opened 2019-10-23
-ri, --rules-ignore-case: Ignore case in rules regex checks

knek-little-projects:
With this flag user can specify the rule:
```
    "Password": "password",
```
instead of 
```
    "Password": "[pP][aA][sS][wW][oO][rR][dD],
```


amiller-isp:
You can do this already
` "Password": "regex:[pP][aA][sS][wW][oO][rR][dD]",`
Can be written
` "Password": "regex:(?i:password)",`

dustin-decker:

PR #191, opened 2019-12-05
Update .gitignore

nsigarora:


dxa4481:
There shouldn't be any cpp files in the trufflehog working directory. Feel free to re-open if I've overlooked something here.

PR #217, opened 2020-05-08
Fix versions

changusmc:
GitPython 3.0.6 is not found in the requirements.

The current build is broken https://travis-ci.org/github/dxa4481/truffleHog


dxa4481:
I believe this is only an issue with python 2.7, which I don't plan to support anymore

Issue #218, opened 2020-05-20
since_commit is broken


dmrickert:
The since_commit feature is broken.  

It always scans the first commit of a repo due to the way the for loop "continue"s to loop through all commits until it reaches the oldest commit.  Further, the since_commit feature actually ignores the commit newer then the since_commit due to the way the prev_commit logic is written.

dmrickert:
To expand a bit, let's take a look at the tests and the Git log for the project the tests rely on.

```
commit 9ed54617547cfca783e0f81f8dc5c927e3d1e345
Author: flower <flower@flowers-MacBook-Pro.local>
Date:   Fri Dec 30 23:20:50 2016 -0600

    OH no a secret

commit 709f22821820a7815106f82a03e8d90f50b2b653
Author: flower <flower@flowers-MacBook-Pro.local>
Date:   Fri Dec 30 23:19:51 2016 -0600

    removing secret

commit d15627104d07846ac2914a976e8e347a663bbd9b
Author: flower <flower@flowers-MacBook-Pro.local>
Date:   Fri Dec 30 23:19:39 2016 -0600

    Oh no a secret file

commit 202564cf776b402800a4aab8bb14fa4624888475
Author: flower <flower@flowers-MacBook-Pro.local>
Date:   Fri Dec 30 23:16:57 2016 -0600

    updating the readme
```

So, a secret was added in d156... removed in 709f... and then added again in 9ed5.  The tests were set up to use d156 as the "since_commit".  This should've caught the secret being removed in 709f.  Instead, the since_commit actually caught the change in 9ed.  This was because of the bug I mentioned where the first commit since the designated "since_commit" actually gets ignored.

Also, the first commit was always getting scanned when the since_commit flag was used.  If you look at the logic in truffleHog.py specifically in the function "find_strings" the behavior once "since_commit" is matched is to keep looping through `repo.iter_commits(branch_name, max_count=max_depth)`.  While it keeps looping through, it continues to set `prev_commit = curr_commit` before it hits the continue.  Also, curr_commit keeps getting updated as it loops through the for loop and hits the continue as well.

That final curr_commit ends up always hitting the logic where it compares the commit to a "NULL_TREE" which is a blank git construct.  Thus, the first commit on a branch was always getting scanned when using since_commit.

Love the tool, impressed that it's so powerful with such few lines of code.

toshith:
[...]

PR #219, opened 2020-05-20
Fixed since_commit and tests

dmrickert:
Fixes broken since_commit referenced in #218 

The break is also obvious if you look at the git log and the hashes the since_commit tests were looking for.  There was a commit that should've been found sooner, and it should've been found in all 3 current branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /truffleHog.egg-info/
 **/__pycache__/
 **/*.pyc
+env/
--- a/test_all.py
+++ b/test_all.py
@@ -33,9 +33,9 @@ class TestStringMethods(unittest.TestCase):
         # Start at commit d15627104d07846ac2914a976e8e347a663bbd9b, which 
         # is immediately followed by a secret inserting commit:
         # https://github.com/dxa4481/truffleHog/commit/9ed54617547cfca783e0f81f8dc5c927e3d1e345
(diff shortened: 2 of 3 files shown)


cristianbriscaru:
Any progress on this? 
We are experiencing the same issue

dxa4481:
I'll try to take a look this weekend

salimane:
@dxa4481 a new release on pip would be welcomed

dxa4481:
Hey @salimane I cut a release on February 4th, 2.2.1, are you having issues with that release?

salimane:
`0f223225d6efc8c64504d9381eececb06b14c0e6` is the last commit on `https://github.com/dxa4481/truffleHog`

If I run the following command
```
truffleHog --regex --entropy=False --since_commit=0f223225d6efc8c64504d9381eececb06b14c0e6 https://github.com/dxa4481/truffleHog
```

I will expect no warnings but it still scans the full commit history.

dxa4481:
@salimane so trufflehog scans all branches, and looks for the since_commit in each branch it scans. Since the latest commit (in dev) isn't in the other branches, it still does a full scan of those https://github.com/dxa4481/truffleHog/blob/dev/truffleHog/truffleHog.py#L339

To get around this you can provide the branch flag to keep it only scanning a single branch.

I think that's it. Let me know if you find anything else wonky with it.

salimane:
@dxa4481 thanks. I have added the branch flag and it worked.

PR #238, opened 2020-12-09
Merge status code feature

dporr:
I found this feature useful for automated CI jobs. It would be great if It can be integrated into the main version.


dxa4481:
Hey, looks like this already exists in the main version https://github.com/dxa4481/truffleHog/blob/dev/truffleHog/truffleHog.py#L87

Feel free to re-open if I overlooked something

PR #262, opened 2021-09-13
Add staged file support

danzajork:
Hello,

I've added support for scanning staged files. This is extremely helpful for situations where a developer needs to know if files contain secrets before a commit is made. It also aids in the support of using truffleHog in a pre-commit githook to block unwanted commits.

Thank you,
Dan


dustin-decker:

It supports scanning unstaged files for local repos!

PR #276, opened 2022-01-31
Fix Error - can only concatenate str (not tuple) to str

Splint3r7:
## Issue details:

if the found_strings contains a value that is not strung then the concatenation of fails in `def regex_check` fails and throws the error:

Screenshot 2022-01-31 at 6 26 19 PM

## Fix

To fix the issue, I have converted `found_string` to str first and the issue resolved.


dustin-decker:

PR #293, opened 2022-04-05
Add JFrog Artifactory protobuf source and metadata

mcastorina:


proto/source_metadata.proto
```diff
@@ -188,6 +188,17 @@ message Teams {
...
 
+// https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-FileInfo
+message Artifactory {
+  string repo = 1;
+  string path = 2;
+  string link = 3;
+  string created_at = 4;
+  string created_by = 5;
+  string modified_at = 6;
+  string modified_by = 7;
```
dustin-decker, line 199: are we able to capture an email (preferred) or a username of the person who uploaded the package?

mcastorina: It looks like there's a [GetUserDetails](https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-GetUserDetails) API that contains an email.

I'm not certain, but I think `created_by` and `modified_by` is the username (see [FileInfo docs](https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-FileInfo)), so we can make another request to get the email.

mcastorina: Maybe we can simplify the metadata to choose whichever timestamp is the latest, and have a `username` and `timestamp` instead of capturing both creating and modification data.

`email` can be added too as best effort

dustin-decker: We might be able to build a lookup from one request to Get Users List. I'm not sure if it contains email though.

https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API#ArtifactoryRESTAPI-GetUsers

I think simplifying the metadata to `timestamp` and `username` is a good call.

dustin-decker (approved):

PR #302, opened 2022-04-06
Fix regex for Slack API Tokens

TarasZelyk:
Fixes regex for Slack API Token to match uppercase letters in token


dustin-decker:
Thanks tz

dustin-decker (approved):

dustin-decker:
Fixes #300

PR #317, opened 2022-04-08
Don't attempt checking for updates on dev builds

dustin-decker:

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,6 +116,10 @@ func main() {
 	if !*noUpdate {
 		updateCfg.Fetcher = updater.Fetcher(version.BuildVersion)
 	}
+	if version.BuildVersion == "dev" {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -116,6 +116,10 @@ func main() {
 	if !*noUpdate {
 		updateCfg.Fetcher = updater.Fetcher(version.BuildVersion)
 	}
+	if version.BuildVersion == "dev" {
+		updateCfg.Fetcher = nil
+	}
```
mcastorina, line 121: **Nit**: This could be more succintly represented with

```go
if version.BuildVersion != "dev" && !*noUpdate {
    updateCfg.Fetcher = updater.Fetcher(version.BuildVersion)
}
```

though maybe less readable that way.

mcastorina (approved):

PR #324, opened 2022-04-08
Update Dockerfiles

adriang-90:
# Update Dockerfiles 

I have updated Dockerfiles to  known best pracitces such as:
* use **WORKDIR** instead of **RUN mkdir**
* pre-build ``` go ``` dependencies for better caching
* add ``` --no-cache ``` option -> [Why?](https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache)
I did not achieve much of a space  (just **82MB** from **84.5**) but since we need ``` git ``` and ``` mv ``` command I do not see a reason to use ``` distroless ``` or ``` scratch ``` image.


dustin-decker:
Thank you, Adrian

dustin-decker (approved):

Issue #186, opened 2019-10-15
Install via Homebrew


cloudlena:
It would be great to get a [Homebrew](https://brew.sh/) formula for this project to make it easier to keep the installed version up to date on macOS.

drewstinnett:
I'd be happy to help with this if you all like. Goreleaser can build the homebrew package with a little extra config...we would just need somewhere to push it. Maybe a new repo for all the trufflesecurity formulas, like `homebrew-trufflehog`? Users could then do a `brew tap trufflesecurity/trufflehog` and `brew install trufflehog`.

dustin-decker:
This is now available! Thank you @drewstinnett!

PR #330, opened 2022-04-08
Build and Publish Homebrew Package

drewstinnett:
This will add a homebrew targets in, using the existing goreleaser process. This PR should address #186 

Note that I've got my personal repo in there now as a target (mainly so I could test that it actually works). These should be changed to whatever you folks think is appropriate 😁


dustin-decker:
Thanks a lot, Drew! Looking forward to getting this released

drewstinnett:
Happy to help however I can!

dustin-decker:
Do you know if there is any downside to having the brew formulas in this repository?

drewstinnett:
I don't think it can go directly in this repository...in order for the tap names to work, the repository name needs to be prefixed with `homebrew-`. So for example, it would need to go into a repository like:

`github.com/trufflesecurity/homebrew-trufflehog`

Then it could be tapped with: `brew tap trufflesecurity/trufflehog` (Note how the `homebrew-` is dropped from the tap cli)

What some folks do, is create a tap to hold all their formulas, which may be good here if there are other homebrew packages you wanna put up. In this case, you could have something like `trufflesecurity/homebrew-tap`, which would work with `brew tap trufflesecurity/tap`

Full docs on this is [here](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap), if it's helpful!

dustin-decker:
Thanks a lot, Drew. I've setup the new repository and a service account token in the secrets for this repo. I've also renamed the owner in your PR. I think it's ready to go 🚀

dustin-decker (approved):

drewstinnett:
Looks awesome Dustin! Feel free to ping me if anything is wonky and thanks!!

PR #344, opened 2022-04-08
Added sonarcloud scanner

roxanne-tampus:


pkg/detectors/sonarcloud/sonarcloud.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_SonarCloud,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://"+resMatch+"@sonarcloud.io/api/users/search", nil)
```
dustin-decker, line 50: It looks like there's a validation endpoint: https://sonarcloud.io/web_api/api/authentication?deprecated=false

dustin-decker: Is the credential that we are testing with as-is from sonarcloud? I'm asking because most have to be base64 encoded before they can be used with basic auth.

roxanne-tampus: Yes, the credential is generated in sonarcloud. It is located under My Account->Security. I just updated the endpoint and added organization as parameter since it seems validation endpoint still returns true even without token or credentials.
image

PR #345, opened 2022-04-08
Added image4 scanner

roxanne-tampus:


pkg/detectors/image4/image4.go
```diff
@@ -0,0 +1,73 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"image4"}) + `\b([0-9a-zA-Z]{22}\==)`)
```
dustin-decker, line 23: This looks like a base64 encoded token. The `==` at the end is padding, and might not always be like that.

This would be more correct:
`\b([0-9a-zA-Z]{22}[0-9a-zA-Z\=]{2})`

roxanne-tampus: I just commited the changes in regex

PR #348, opened 2022-04-09
Remove overseer warn, and don't os.Exit on git error.

bill-rich:


pkg/sources/git/git.go
```diff
@@ -262,8 +262,13 @@ func (s *Git) ScanCommits(repo *git.Repository, path string, scanOptions *ScanOp
...
 		return err
 	}
-	zerolog.SetGlobalLevel(zerolog.Disabled)
-	fileChan, err := glgo.GitLog(path, scanOptions.HeadHash)
+	if log.GetLevel() < log.DebugLevel {
+		zerolog.SetGlobalLevel(zerolog.Disabled)
+	}
+
+	// Errors returned on errChan aren't blocking, so just ignore them.
+	errChan := make(chan error)
```
dustin-decker, line 270: Should we do any logging with these, or is that happening already with zerolog when in debug mode?

dustin-decker (approved):

PR #360, opened 2022-04-09
Added holistic scanner

roxanne-tampus:


pkg/detectors/holistic/holistic.go
```diff
@@ -0,0 +1,86 @@
...
+			req.Header.Add("x-api-key", resMatch)
+			res, err := client.Do(req)
+			if err == nil {
+				bodyBytes, err := ioutil.ReadAll(res.Body)
+				if err != nil {
+					continue
+				}
+
+				bodyString := string(bodyBytes)
+				validResponse := strings.Contains(bodyString, `"data":[]`)
```
dustin-decker, line 64: This response is likely to change when there are projects to return. Maybe use ``"data":[` instead, which just contains the beginning of the JSON array?

dustin-decker (changes requested):

pkg/detectors/holistic/holistic.go
roxanne-tampus: In creating a Holistic account, an auto-generated sample project is created. So it will likely to return that project if the keys is correct, otherwise it will return null.

PR #367, opened 2022-04-11
added stormboard scanner

lonmarsDev:

### DIFF
--- a/pkg/detectors/stormboard/stormboard.go
+++ b/pkg/detectors/stormboard/stormboard.go
@@ -2,7 +2,6 @@ package stormboard
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
--- a/pkg/detectors/stormboard/stormboard.go
+++ b/pkg/detectors/stormboard/stormboard.go
@@ -1,0 +1,74 @@
+package stormboard
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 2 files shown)


pkg/detectors/stormboard/stormboard.go
```diff
@@ -0,0 +1,74 @@
+package stormboard
+
+import (
+	"context"
+	"fmt"
```
dustin-decker, line 5: It won't compile with this unused import. Also, the test seems to be failing right now.

dustin-decker (changes requested):

pkg/detectors/stormboard/stormboard.go
dustin-decker: Oh the test failing might be my very slow wifi timing out. Please remove the unused import though.

pkg/detectors/stormboard/stormboard.go
```diff
@@ -0,0 +1,73 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"stormboard"}) + `\b([a-z0-9]{40})\b`)
```
roxanne-tampus, line 23: suggestion: this regex can be made a bit stricter by `\b([0-9a-f]{40})\b` as the keys are in hex encoding so we can only work with `a-f`.

dustin-decker (approved):

PR #368, opened 2022-04-11
added apilayer scanner

lonmarsDev:

### DIFF
--- a/pkg/detectors/apilayer/apilayer.go
+++ b/pkg/detectors/apilayer/apilayer.go
@@ -1,0 +1,77 @@
+package apilayer
+
+import (
+	"context"
+	"net/http"
+	"regexp"
+	"io/ioutil"
+	"strings"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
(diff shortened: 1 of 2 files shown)


pkg/detectors/apilayer/apilayer.go
```diff
@@ -0,0 +1,77 @@
...
+			res, err := client.Do(req)
+			if err == nil {
+				defer res.Body.Close()
+				bodyBytes, err := ioutil.ReadAll(res.Body)
+				if err != nil {
+					continue
+				}
+				body := string(bodyBytes)
+
+				if !strings.Contains(body, "invalid_access_key") {
```
dustin-decker, line 63: Please invert this logic so that this will fail safely during a service outage. Check for contents of a successful response instead.

pkg/detectors/apilayer/apilayer.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Apilayer,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "http://apilayer.net/api/validate?access_key="+resMatch+"&number=09066200872", nil)
```
roxanne-tampus, line 50: Is the detector's site is https://apilayer.com? If it is, they might have updated their top-level domain and also the endpoint on verifying the number https://apilayer.com/marketplace/description/number_verification-api. They also updated the key pattern into base64 from hex encoding.

dustin-decker (approved):

PR #371, opened 2022-04-11
added paydirtapp scanner

lonmarsDev:


dustin-decker (changes requested):
Please rename all occurrences of paydirtyapp to paydirtapp, including the package name

PR #380, opened 2022-04-11
Use .Line to reduce diff strangeness

bill-rich:
Using .String() results in the op ("+") and extra newlines. Switching to .Line removes both of these. Expected chunks adjusted for raw chunk changes.

--trace option also added for easier debugging


pkg/sources/git/git.go
```diff
@@ -278,6 +278,7 @@ func (s *Git) ScanCommits(repo *git.Repository, path string, scanOptions *ScanOp
 
 	var depth int64
 	for file := range fileChan {
+		log.WithField("commit", file.PatchHeader.SHA).WithField("file", file.NewName).Trace("Scanning file from git")
```
mcastorina, line 281: **Nit:** There's also `log.WithFields` that I just learned about:

```go
log.WithFields(log.Fields{"commit": file.PatchHeader.SHA, "file": file.NewName}).Trace("scanning file from git")
```

mcastorina (approved):

PR #389, opened 2022-04-12
Update the Teams message proto to include access token auth.

ahrav:


dustin-decker:
Is this access token long lived? Or should this be refresh token?

dustin-decker (approved):

dustin-decker:
You may also have to regen following #387

ahrav:
> You may also have to regen following #387
Yep, i made sure to pull in his changes first.

PR #390, opened 2022-04-12
Address some linting issues

dustin-decker:


mcastorina (approved):
Thanks for doing this!

PR #395, opened 2022-04-14
added lunchmoney scanner

ladybug0125:

### DIFF
--- a/pkg/detectors/lunchmoney/lunchmoney.go
+++ b/pkg/detectors/lunchmoney/lunchmoney.go
@@ -21,7 +21,7 @@ var (
 	client = common.SaneHttpClient()
 
 	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"lunchmoney"}) + `\b([a-zA-Z0-9]{50})\b`)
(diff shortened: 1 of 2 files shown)


ahrav (approved):
lgtm

pkg/detectors/lunchmoney/lunchmoney.go
```diff
@@ -0,0 +1,75 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"lunchmoney"}) + `\b([a-zA-Z0-9]{50})\b`)
```
ahrav, line 24: suggestion: tighten this regex and use `\b([a-f0-9]{50})\b`

ahrav (changes requested):
Small tweak to the regex since lunchmoney only uses a-f

pkg/detectors/lunchmoney/lunchmoney.go
ladybug0125: I changed the regex and pushed it. Thanks!

dustin-decker: Some extra context here Mildred, this is a hex-encoded credential which only uses a-f rather than a-z

PR #396, opened 2022-04-14
added yelp scanner

ladybug0125:

### DIFF
--- a/pkg/detectors/yelp/yelp.go
+++ b/pkg/detectors/yelp/yelp.go
@@ -1,0 +1,74 @@
+package yelp
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct{}
+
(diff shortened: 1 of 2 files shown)


pkg/detectors/yelp/yelp.go
```diff
@@ -0,0 +1,74 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"yelp"}) + `\b([0-9A-Za-z]{15}-[0-9A-Za-z]{27}-[0-9A-Za-z]{84})\b`)
```
ahrav, line 24: suggestion: I would use ``\b([a-zA-Z0-9_\\=\.\-]{128})\b`` as your regex here because not all the API keys for yelp are broken down with `-` at those exact places. As an example [Yelp API Key](https://regex101.com/r/0NMud1/1) is an API key for yelp with a very different structure. 

I don't think you can you can make it more exact other than using a more generic `Bearer` match.

ahrav (changes requested):

pkg/detectors/yelp/yelp.go
ladybug0125: I changed the regex and pushed it. Thank you!

PR #397, opened 2022-04-14
Updated to new endpoint for IBM Cloud

jjasghar:
Taken from https://cloud.ibm.com/apidocs/codeengine#get-kubeconfig
The IAM is now pointing to long term one, and also updated the
additional information in the POST


dustin-decker:
Thank you for the contribution. Can you sign the CLA?

I tested this change and the test passes with the endpoint change but not with the payload change.

jjasghar:
Ah interesting, the update had `&refresh_token=${refreshToken}` as an option, i wasn't 100% sure how to add that. maybe just the updated URL is what we need?

dustin-decker (approved):

PR #398, opened 2022-04-14
added atera scanner

ladybug0125:

### DIFF
--- a/pkg/detectors/atera/atera.go
+++ b/pkg/detectors/atera/atera.go
@@ -2,7 +2,6 @@ package atera
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
@@ -53,7 +52,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				continue
 			}
 			req.Header.Add("Accept", "application/json")
-			req.Header.Add("X-API-KEY", fmt.Sprintf("%s", resMatch))
+			req.Header.Add("X-API-KEY", resMatch)
 			res, err := client.Do(req)
 			if err == nil {
 				defer res.Body.Close()
--- a/pkg/detectors/atera/atera.go
+++ b/pkg/detectors/atera/atera.go
@@ -1,0 +1,75 @@
+package atera
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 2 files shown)


pkg/detectors/atera/atera.go
```diff
@@ -0,0 +1,75 @@
...
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://app.atera.com/api/v3/alerts", nil)
+			if err != nil {
+				continue
+			}
+			req.Header.Add("Accept", "application/json")
+			req.Header.Add("X-API-KEY", fmt.Sprintf("%s", resMatch))
```
ahrav, line 56: nit: you don't need to use `fmt.Sprintf` here. You can just use `resMatch`.

pkg/detectors/atera/atera_test.go
```diff
@@ -0,0 +1,116 @@
...
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "found, unverified",
+			s:    Scanner{},
+			args: args{
+				ctx:    context.Background(),
+				data:   []byte(fmt.Sprintf("You can find a atera secret %s within but not valid", inactiveSecret)), // the secret would satisfy the regex but not pass validation
```
ahrav, line 59: nit: find `an` vs find `a`

pkg/detectors/atera/atera_test.go
```diff
@@ -0,0 +1,116 @@
...
+		args    args
+		want    []detectors.Result
+		wantErr bool
+	}{
+		{
+			name: "found, verified",
+			s:    Scanner{},
+			args: args{
+				ctx:    context.Background(),
+				data:   []byte(fmt.Sprintf("You can find a atera secret %s within", secret)),
```
ahrav, line 43: nit: same here `an` vs `a`

ahrav (approved):
Small nitpicks otherwise lgtm

pkg/detectors/atera/atera_test.go
ladybug0125: Thanks! I just pushed the changes.

pkg/detectors/atera/atera.go
ladybug0125: Thanks! I just pushed the changes.

pkg/detectors/atera/atera_test.go
ladybug0125: Thanks! I just pushed the changes.

PR #402, opened 2022-04-15
Added aha scanner

roxanne-tampus:


pkg/detectors/aha/aha.go
```diff
@@ -0,0 +1,75 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aha"}) + `\b([0-9a-zA-Z]{64})\b`)
```
ahrav, line 24: suggestion: You can tighten up this regex pattern by using `\b([0-9a-f]{64})\b` Aha's API keys are encoded so they only use a-f0-9.

ahrav (approved):
small suggestion otherwise lgtm.

pkg/detectors/aha/aha.go
roxanne-tampus: I just modified the regex as suggested. Thanks @ahrav!

dustin-decker: Some extra context: it's hex encoding that uses the a-f characters rather than a-z

PR #403, opened 2022-04-15
Added parsehub scanner

roxanne-tampus:


pkg/detectors/parsehub/parsehub.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Parsehub,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://www.parsehub.com/api/v2/projects?api_key="+resMatch+"&offset=0&limit=20&include_options=1", nil)
```
ahrav, line 50: nit: prefer `fmt.Sprintf("https://www.parsehub.com/api/v2/projects?api_key=%s&limit=1", resMatch)` and additionally you can also likely remove the additional query params to the request to make it simpler. I would potentially also just change the limit to 1 to also potentially reduce the request latency and size.

ahrav (approved):
small nit, otherwise lgtm

pkg/detectors/parsehub/parsehub.go
roxanne-tampus: Thanks @ahrav! Pushed the enhancement.

PR #404, opened 2022-04-15
Added packagecloud scanner

roxanne-tampus:


pkg/detectors/packagecloud/packagecloud.go
```diff
@@ -0,0 +1,73 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"packagecloud"}) + `\b([0-9a-z]{48})\b`)
```
ahrav, line 23: suggestion: this regex can be made a bit stricter `\b([0-9a-f]{48})\b` You can only look for `a-z` since this is an encoded.

ahrav (approved):
1 small suggestion otherwise lgtm. 👍

pkg/detectors/packagecloud/packagecloud.go
roxanne-tampus: Thanks! I just pushed the enhancement.

dustin-decker: Some extra context: it's hex encoding that uses the a-f characters rather than a-z

PR #406, opened 2022-04-15
Added flowdash scanner

roxanne-tampus:


pkg/detectors/flowdash/flowdash.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Flowdash,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://app.flowdash.com/api/v1/workflows", nil)
```
ahrav, line 51: nit: maybe use `https://app.flowdash.com/api/v1/workspace` as the endpoint vs workflows? The response object for workspace will generally always be a single item whereas for workflows i'm not certain on the resulting response objects size. Potentially a lot bigger if they are a large org and even moreso if their API doesn't support pagination. This way we lower the call's overhead and maybe reducing latency.

ahrav (approved):
tiny suggestion otherwise lgtm 👍

pkg/detectors/flowdash/flowdash.go
roxanne-tampus: Great! I just pushed the changes. Thanks @ahrav!

PR #407, opened 2022-04-15
Added flowdock scanner

roxanne-tampus:


pkg/detectors/flowdock/flowdock.go
```diff
@@ -0,0 +1,73 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"flowdock"}) + `\b([0-9a-z]{32})\b`)
```
ahrav, line 23: suggestion: this regex can be made a bit stricter `\b([0-9a-f]{32})\b`

pkg/detectors/flowdock/flowdock.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Flowdock,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.flowdock.com/flows", nil)
```
ahrav, line 50: nit: Maybe use `https://api.flowdock.com/flows?users=false` in their documentation they state:

`A boolean value (1/0) that controls whether a list of users should be included with each flow.` 
Based on that wording (1 before 0) that indicates to me the default behavior is to include users within the response if that is the case we should probably pass in `users=false` to return the response object size. If that is not the case feel free to ignore this 😄

ahrav (approved):
couple suggestions otherwise lgtm 👍

pkg/detectors/flowdock/flowdock.go
roxanne-tampus: Sure, I just pushed the changes. Thanks @ahrav!

pkg/detectors/flowdock/flowdock.go
dustin-decker: Some extra context: it's hex encoding that uses the a-f characters rather than a-z

PR #409, opened 2022-04-15
Added fibery scanner

roxanne-tampus:


pkg/detectors/fibery/fibery.go
```diff
@@ -0,0 +1,88 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"fibery"}) + `\b([0-9a-z.]{44})\b`)
```
ahrav, line 25: suggestion: I think we can make this a little bit more strict `\b([0-9a-f]{8}.[0-9a-f]{35})\b` all the keys are encoded so we can only work with `a-f` additionally they are all split into two section by a `.` separator. So we can use that to our advantage.

pkg/detectors/fibery/fibery.go
```diff
@@ -0,0 +1,88 @@
...
+	domainPat = regexp.MustCompile(detectors.PrefixRegex([]string{"fibery", "domain"}) + `\b([0-9a-z]{5,20})\b`)
```
ahrav, line 26: suggestion we can also make this a little more robust. After some testing it looks like they allow alphanumerics as part of the domain name and also allow a minimum of 2 characters and a max of around 40. So maybe lets go with `\b([0-9A-Za-z]{2,40})\b` ?

pkg/detectors/fibery/fibery.go
```diff
@@ -0,0 +1,88 @@
...
+
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_Fibery,
+				Raw:          []byte(resMatch),
+			}
+
+			if verify {
+				timeout := 10 * time.Second
+				client.Timeout = timeout
+				req, err := http.NewRequestWithContext(ctx, "POST", "https://"+resDomainMatch+".fibery.io/api/commands", nil)
```
ahrav, line 63: nit: Prefer `fmt.Sprintf` here: `fmt.Sprintf("https://%s.fibery.io/api/commands", resDomainMatch)`

ahrav (changes requested):
Few tweaks to the regex for additional coverage, otherwise lgtm

pkg/detectors/fibery/fibery.go
roxanne-tampus: Great! Pushed the enhancements. Thanks @ahrav!

dustin-decker (approved):

PR #410, opened 2022-04-15
Added typetalk scanner

roxanne-tampus:


pkg/detectors/typetalk/typetalk.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			resIdMatch := strings.TrimSpace(idMatch[1])
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_Typetalk,
+				Raw:          []byte(resMatch),
+			}
+
+			if verify {
+				req, err := http.NewRequestWithContext(ctx, "POST", "https://typetalk.com/oauth2/access_token?client_id="+resIdMatch+"&client_secret="+resMatch+"&grant_type=client_credentials", nil)
```
ahrav, line 58: nit: Prefer fmt.Sprintf when there are multiple variables within the string. 
`fmt.Sprintf("https://typetalk.com/oauth2/access_token?client_id=%s&client_secret=%s&grant_type=client_credentials", resIdMatch, resMatch)`

ahrav (approved):
nice LGTM 👍

pkg/detectors/typetalk/typetalk.go
roxanne-tampus: pushed the enhancement, thanks @ahrav!

PR #417, opened 2022-04-16
Handle nil file channel

dustin-decker:
Parser can return a nil channel and nil error under some conditions


pkg/sources/git/git.go
```diff
@@ -272,6 +272,10 @@ func (s *Git) ScanCommits(repo *git.Repository, path string, scanOptions *ScanOp
 	if err != nil {
 		return errors.WrapPrefix(err, "could not open repo path", 0)
 	}
+	// parser can return nil chan and nil error
```
ahrav, line 275: nit: Capitalize and period.

ahrav (approved):
lgtm

Issue #424, opened 2022-04-18
API keys not detected in Filesystem mode


SecTheBit:
**Describe the bug**
I created one file with securitytrails api key in it and tested it out with trufflehog in filesystem mode. But it is not able to detect the api key, even with `--no-verification` flag

**To Reproduce**
command used: 
```
sudo trufflehog filesystem --directory=work/trufflehog/ --debug --no-verification

```

**Output**
```
DEBU[0000] running version 3.3.2                        
2022/04/18 11:02:52 [updater parent] run
2022/04/18 11:02:52 [updater parent] checking for updates...
2022/04/18 11:02:57 [updater parent] failed to get latest version: already up to date
2022/04/18 11:02:57 [updater parent] starting /usr/local/bin/trufflehog
DEBU[0000] running version 3.3.2                        
2022/04/18 11:02:57 [updater child#1] run
2022/04/18 11:02:57 [updater child#1] start program
trufflehog 3.3.2
DEBU[0000] running with up to 8 workers                 
DEBU[0000] loaded 2 decoders                            
DEBU[0000] loaded 631 detectors total, 0 with verification enabled. 631 with verification disabled 
INFO[0000] starting pprof and fgprof server on :18066 /debug/pprof and /debug/fgprof 
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2022/04/18 11:02:58 [updater parent] checking for updates...
DEBU[0001] scanned 1 chunks                             
2022/04/18 11:02:58 [updater parent] prog exited with 0

```
**Key File Format**
```
SECURITYTRAILS:
apikeywdnwjfb3434343
```

dustin-decker:
The detector currently expects a 32 character token: https://github.com/trufflesecurity/trufflehog/blob/ea95e886ba366541ac08959d4e9d4e49d02c6863/pkg/detectors/securitytrails/securitytrails.go#L23

The example you posted is 20 characters. Have you found that your actual tokens are not 32 characters?

SecTheBit:
Hi @dustin-decker 

Yes, the actual tokens are of 32 characters

ahrav:
@SecTheBit the issue should be resolved in. #429

PR #429, opened 2022-04-18
[THOG-234] Update security trails detector's regex and keywords.

ahrav:
## What?
Update the security trails' detector's regex and keywords.
## Why?
The old regex was not capturing secrets that followed keywords after a new line.
By default, "." doesn't match newline. If you give the "s" flag, it does
## How?
Update the regex to include the "s" flag properly matches the secrets with keyword.
## Testing?
accompanying unit tests
## Screenshots (optional)
n/a
## Anything Else?
[Issue#424](https://github.com/trufflesecurity/trufflehog/issues/424)
regex validation: https://regex101.com/r/8T83x5/1

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -92,7 +92,7 @@ func CleanResults(results []Result) []Result {
 func PrefixRegex(keywords []string) string {
 	pre := `(?i)(?:`
 	middle := strings.Join(keywords, "|")
-	post := `).{0,40}`
+	post := `).|(?:[\n\r]){0,40}`
 	return pre + middle + post
 }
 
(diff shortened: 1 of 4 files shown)


pkg/detectors/detectors.go
```diff
@@ -92,7 +92,7 @@ func CleanResults(results []Result) []Result {
 func PrefixRegex(keywords []string) string {
 	pre := `(?i)(?:`
 	middle := strings.Join(keywords, "|")
-	post := `).{0,40}`
+	post := `).[\r\n]{0,40}`
```
dustin-decker, line 95: This needs to be `[.\r\n]` otherwise it captures one character and up to 40 CR/LF

ahrav: Had to separate out the class group from the `.` and put them all into a capture group. That seems to have worked properly.

dustin-decker:
Looks good

PR #437, opened 2022-04-20
[THOG-248] fix broken gitlab tests

ahrav:
## What?
Clean up gitlab scan code to remove duplicated scan logic for different authentication types.
Fix broken tests.
Update clone exec command.
## Why?
To fix the gitlab tests and allows for scanning of repos to work properly.
## How?
Modify the test code to include a defer to close the chunksChan.
Add the `.run()` to the exec command during repo clone.
Dry up logic around scanning repos for auth vs unauth
## Testing?
Updated unit test
## Screenshots (optional)
n/a
## Anything Else?


pkg/sources/gitlab/gitlab.go
```diff
@@ -262,119 +263,89 @@ func (s *Source) getRepos() ([]*url.URL, []error) {
...
+				path, repo, err = git.CloneRepoUsingToken(s.token, repoURL.String(), s.user)
 			}
-			wg.Add(1)
-			go func(ctx context.Context, repoURL *url.URL, i int) {
-				defer s.jobSem.Release(1)
-				defer wg.Done()
-				if len(repoURL.String()) == 0 {
-					return
-				}
-				s.SetProgressComplete(i, len(repos), fmt.Sprintf("Repo: %s", repoURL), "")
```
dustin-decker, line 285: These calls are needed to update progress

pkg/sources/gitlab/gitlab_test.go
```diff
@@ -92,18 +92,20 @@ func TestSource_Scan(t *testing.T) {
...
-			gotChunk.Data = nil
-			gotChunk.SourceMetadata = nil
-			if diff := pretty.Compare(gotChunk, tt.wantChunk); diff != "" {
-				t.Errorf("Source.Chunks() %s diff: (-got +want)\n%s", tt.name, diff)
+			for gotChunk := range chunksCh {
+				gotChunk.Data = nil
+				gotChunk.SourceMetadata = nil
+				if diff := pretty.Compare(gotChunk, tt.wantChunk); diff != "" {
+					t.Errorf("Source.Chunks() %s diff: (-got +want)\n%s", tt.name, diff)
+				}
```
mcastorina, line 108: **Nit:** I think this test will pass if we get 0 chunks. Could we add a counter or something to ensure we get at least 1?

pkg/sources/gitlab/gitlab.go
```diff
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/go-errors/errors"
+	git2 "github.com/go-git/go-git/v5"
```
mcastorina, line 13: **Nit:** `gogit` or `gg` would be a better name imo

mcastorina (approved):
Looks good, thanks for fixing the tests!

pkg/sources/gitlab/gitlab.go
ahrav: Oh yea i don't think i removed any of them. It's just moved up to line 292. I simply refactored both the conditionals into a single since the only difference i noticed was `git. CloneRepoUsingUnauthenticated` vs `git. CloneRepoUsingToken` the rest should be the same.

pkg/sources/gitlab/gitlab.go
ahrav: haha hey man Go's auto-importer would disagree. But yea i like it. gg it is .

pkg/sources/git/git.go
```diff
@@ -84,7 +84,7 @@ func (s *Source) JobID() int64 {
 }
 
 // Init returns an initialized GitHub source.
-func (s *Source) Init(aCtx context.Context, name string, jobId, sourceId int64, verify bool, connection *anypb.Any, concurrency int) error {
+func (s *Source) Init(aCtx context.Context, name string, jobId, sourceId int64, verify bool, connection *anypb.Any, _ int) error {
```
bill-rich, line 87: I think it would be better to pass `concurrency` along to `NewGit` instead of removing it here.

pkg/sources/gitlab/gitlab.go
```diff
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/go-errors/errors"
+	gg "github.com/go-git/go-git/v5"
```
bill-rich, line 13: **Nit**: I like `gogit` more than `gg` because it is easier to understand what it is without looking at the imports.

pkg/sources/gitlab/gitlab.go
```diff
@@ -262,119 +263,89 @@ func (s *Source) getRepos() ([]*url.URL, []error) {
...
-				// We are returning nil instead of the scanErrors slice here because
-				// we don't want to mark this scan as errored if we cancelled it.
-				return nil
+	errs = s.scan(ctx, chunksChan, repos, &wg, &errsMut, errs)
+	wg.Wait()
+
+	return errs
+}
+
+func (s *Source) scan(ctx context.Context, chunksChan chan *sources.Chunk, repos []*url.URL, wg *sync.WaitGroup, errsMut *sync.Mutex, errs []error) []error {
```
bill-rich, line 274: **Nit/Question:** Its very possible I'm overlooking something, but it seems like it would be easier to keep `scan` and `scanRepos` as one since `errs` and `errsMut` are being passed into `scan`

bill-rich (commented):
Looks good. Just one small change request and a couple nit/questions.

pkg/sources/gitlab/gitlab.go
ahrav: ah fair enough.

pkg/sources/git/git.go
ahrav: Yea so I had removed it because it wasn't being used in the `NewGit` constructor call, but what you are suggesting is i pass that in as opposed to the current implementation `runtime.NumCPU()`? Also follow up can concurrency be 0? In the even 0 is passed into `Init` can we pass 0 into `NewGit` or should we check that and if it's 0 set it to the `runtime.NumCPU()`

pkg/sources/gitlab/gitlab.go
ahrav: Glad you mentioned this, yea it actually now makes more sense within the original function. When i first broke it apart there was still a good amount of code within `scanRepos` that warranted a separate function for the actual scanning portion. Now i agree with you it can be condensed.
Combined back into a single fxn.

dustin-decker (approved):

dustin-decker:
Needs conflict resolved, but lgtm right now

PR #438, opened 2022-04-21
Ignore safedir when scanning git repo

bill-rich:
Fixes issue where safedir prevents running `git log` against a mounted docker directory where directory ownership doesn't match. The issue was preventing the trufflehog action from working properly.


ahrav (approved):
LGTM

Note: Might be worth adding a test for `ScanCommits`  since it's exported. Although given the dependency on glgo might be a little bit of work. Nonetheless could be beneficial maybe?

PR #440, opened 2022-04-21
Detectors that fail verification should still report the unverified secret

trufflesteeeve:
A number of detectors would fail to detect a secret if verification is turned on, but the verification request failed. This adjusts them to ensure the secret is still included if the verification request fails.

I do have a couple of questions that are slightly unrelated. I've included comments in the code for them, and will add PR comments on them as well.

### DIFF
--- a/pkg/detectors/github_old/github_old.go
+++ b/pkg/detectors/github_old/github_old.go
@@ -75,21 +75,24 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			req.Header.Add("Content-Type", "application/json; charset=utf-8")
 			req.Header.Add("Authorization", fmt.Sprintf("token %s", token))
 			res, err := client.Do(req)
-			if err != nil {
-				break
-			}
-			defer res.Body.Close()
-			if res.StatusCode >= 200 && res.StatusCode < 300 {
-				var userResponse userRes
-				err = json.NewDecoder(res.Body).Decode(&userResponse)
-				if err == nil {
-					s.Verified = true
+			if err == nil {
+				if res.StatusCode >= 200 && res.StatusCode < 300 {
+					var userResponse userRes
+					err = json.NewDecoder(res.Body).Decode(&userResponse)
+					res.Body.Close()
+					if err == nil {
+						s.Verified = true
+					}
 				}
 			}
+		}
+
+		if !s.Verified && detectors.IsKnownFalsePositive(token, detectors.DefaultFalsePositives, true) {
+			continue
 		}
 
 		results = append(results, s)
 	}
 
-	return
+	return // Should we add detectors.CleanResults(results), nil ?
 }
--- a/pkg/detectors/github_old/github_old.go
+++ b/pkg/detectors/github_old/github_old.go
@@ -94,5 +94,5 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		results = append(results, s)
 	}
 
-	return // Should we add detectors.CleanResults(results), nil ?
+	return detectors.CleanResults(results), nil
 }
--- a/pkg/detectors/jdbc/jdbc.go
+++ b/pkg/detectors/jdbc/jdbc.go
@@ -45,29 +45,26 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			Redacted:     redact,
 		}
 
-		if verify {
-			//TODO can this be verified? Possibly. Could triage verification to other DBMS strings
-			s.Verified = false
-			// client := common.SaneHttpClient()
-			// req, err := http.NewRequestWithContext(ctx, "GET", "https://jdbcci.com/api/v2/me", nil)
-			if err != nil {
-				continue
-			}
-			// req.Header.Add("Accept", "application/json;")
-			// req.Header.Add("Jdbc-Token", token)
-			// res, err := client.Do(req)
-			// if err != nil {
-			// 	break
-			// }
-			// if res.StatusCode >= 200 && res.StatusCode < 300 {
-			// 	s.Verified = true
-			// }
-		}
+		//if verify {
+		//	// TODO: can this be verified? Possibly. Could triage verification to other DBMS strings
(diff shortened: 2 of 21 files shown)


pkg/detectors/github_old/github_old.go
```diff
@@ -75,21 +75,24 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 
+		if !s.Verified && detectors.IsKnownFalsePositive(token, detectors.DefaultFalsePositives, true) {
+			continue
+		}
+
 		results = append(results, s)
 	}
 
-	return
+	return // Should we add detectors.CleanResults(results), nil ?
```
trufflesteeeve, line 97: This detector looks like it might have a lot of false positives. Should the results be cleaned like similar detectors?

pkg/detectors/microsoftteamswebhook/microsoftteamswebhook.go
```diff
@@ -27,7 +27,7 @@ var (
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"microsoft"}
+	return []string{"microsoft"} // should we instead look for webhook.office.com, or something required to be in the regex above?
```
trufflesteeeve, line 30: This is unrelated to the overall goal, so I'm happy to pull it into another PR. But I think if the key must have `webhook.office.com` in the key pattern, that's probably fine to use as the keyword as well.

pkg/detectors/onelogin/onelogin.go
```diff
@@ -63,26 +63,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+			if !s.Verified && detectors.IsKnownFalsePositive(string(s.Raw), detectors.DefaultFalsePositives, true) {
+				continue
 			}
 
 			results = append(results, s)
 		}
 	}
 
-	return
+	return detectors.CleanResults(results), nil
```
trufflesteeeve, line 84: I wasn't sure if this would have a number of false positives. So I could see this line being removed.

pkg/detectors/pivotaltracker/pivotaltracker.go
```diff
@@ -57,24 +56,21 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-			}
+		if !s.Verified && detectors.IsKnownFalsePositive(string(s.Raw), detectors.DefaultFalsePositives, true) {
+			continue
 		}
 
 		results = append(results, s)
 	}
 
-	return
+	return detectors.CleanResults(results), nil
```
trufflesteeeve, line 75: Again, trying to not get lots of false positives.

pkg/detectors/rechargepayments/rechargepayments.go
```diff
@@ -54,25 +54,22 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+			if !s.Verified && detectors.IsKnownFalsePositive(string(s.Raw), detectors.DefaultFalsePositives, true) {
+				continue
 			}
 
 			results = append(results, s)
 		}
 	}
 
-	return
+	return detectors.CleanResults(results), nil
```
trufflesteeeve, line 74: Another false positive check. Should it be here?

pkg/detectors/uri/uri.go
```diff
@@ -109,34 +108,25 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-			}
+		if !s.Verified && detectors.IsKnownFalsePositive(string(s.Raw), detectors.DefaultFalsePositives, false) {
+			continue
 		}
 
 		results = append(results, s)
 	}
 
-	return
+	return detectors.CleanResults(results), nil
```
trufflesteeeve, line 131: I should have started a review earlier, but yeah, another false positive check. Does it make sense?

pkg/detectors/github_old/github_old.go
dustin-decker: Yeah I think so. The old tokens were quite noisy (commit hashes are detected)

pkg/detectors/microsoftteamswebhook/microsoftteamswebhook.go
dustin-decker: Yes, that should replace the existing keyword. It is ideal to use something that must be in the pattern if possible.

pkg/detectors/jdbc/jdbc.go
```diff
@@ -45,29 +45,26 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		//		continue
+		//	}
+		//	req.Header.Add("Accept", "application/json;")
+		//	req.Header.Add("Jdbc-Token", token)
+		//	res, err := client.Do(req)
+		//	if err == nil {
+		//		if res.StatusCode >= 200 && res.StatusCode < 300 {
+		//			s.Verified = true
+		//		}
+		//	}
```
dustin-decker, line 63: Can you delete all of this? No way that we'll try to verify with this. We have ticket tracking this one. It'll attempt to connect directly from the detector.

pkg/detectors/squareapp/squareapp.go
```diff
@@ -75,21 +74,18 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-					return results, err
-				}
-				defer res.Body.Close()
+				if err == nil {
+					res.Body.Close() // The request body is unused.
 
-				// 404 = Correct crentials. The fake access token should not be found
-				if res.StatusCode == http.StatusNotFound {
-					s.Verified = true
+					// 404 = Correct crentials. The fake access token should not be found
```
dustin-decker, line 80: typo `crentials`

dustin-decker (commented):
LGTM except for the 3 requests that I commented on (keyword change, clean up commented code, and typo)

dustin-decker (approved):

PR #441, opened 2022-04-21
[THOG-204] Use oauth2 as username when cloning

ahrav:
## What?
Fix error when cloning private gitlab repos.
## Why?
In order to scan repos for secret we need to be able to clone them first.
## How?
Use oauth2 as the username when constructing the url for cloning a repo.
## Testing?
Updated unit tests and they are also now working.
## Anything Else?

### DIFF
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -324,7 +324,13 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk, 
 				}
 				s.SetProgressComplete(i, len(repos), fmt.Sprintf("Repo: %s", repoURL), "")
 
-				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), "placeholder")
+				// If a username is not provided we need to use a default one in order to clone a private repo.
+				// Not setting "placeholder" as s.user on purpose in case any downstream services rely on a "" value for s.user.
+				user := s.user
+				if user == "" {
+					user = "placeholder"
+				}
+				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), user)
 				defer os.RemoveAll(path)
 				if err != nil {
 					errsMut.Lock()
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -324,7 +324,7 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk, 
 				}
 				s.SetProgressComplete(i, len(repos), fmt.Sprintf("Repo: %s", repoURL), "")
 
-				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), "oauth2")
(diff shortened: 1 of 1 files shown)


pkg/sources/gitlab/gitlab.go
```diff
@@ -324,7 +324,7 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk,
 				}
 				s.SetProgressComplete(i, len(repos), fmt.Sprintf("Repo: %s", repoURL), "")
 
-				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), s.user)
+				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), "oauth2")
```
dustin-decker, line 327: Maybe "placeholder" instead? We don't support oauth2 for gitlab, this is kind of confusing.

ahrav: well the "oauth2" string is actually need in order to server as the user here: https://github.com/trufflesecurity/trufflehog/blob/6d82056a9b76247723cb31758301c6335b0a1e89/pkg/sources/git/git.go#L245

That way when we construct the url to clone it has the correct format of:
`git clone https://oauth2:${Personal Access Tokens}@gitlab.com/username/myrepo.git` Without oauth2 as part of the url string the clone command fails because it requires a username and password.

dustin-decker: Are you saying the specific string "oauth2" is required and a string that is "placeholder" (or similar) won't work? As far as I can tell, there is nothing related to oauth2 happening here so it's confusing.

ahrav: ahh got it, yea I was under the impression it needed to be oauth2 as the string before `:` in the url because that's what i was reading [Stackover answers](https://stackoverflow.com/questions/25409700/using-gitlab-token-to-clone-without-authentication)

but with some additional testing as long as it's a string it works, it just can't be an empty string. Interesting and weird 🤔  

Updated to "placeholder"

pkg/sources/gitlab/gitlab.go
```diff
@@ -324,7 +324,7 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk,
 				}
 				s.SetProgressComplete(i, len(repos), fmt.Sprintf("Repo: %s", repoURL), "")
 
-				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), s.user)
+				path, repo, err := git.CloneRepoUsingToken(s.token, repoURL.String(), "placeholder")
```
bill-rich, line 327: I think basic auth used to be supported on gitlab, so we should leave the option for a username. The on prem gitlab offering could still support it. Using "placeholder" if s.user is empty should work.

bill-rich (changes requested):

pkg/sources/gitlab/gitlab.go
ahrav: makes sense. Updated to check if s.user is set in the event it isn't we can use "placeholder".

dustin-decker (approved):

PR #449, opened 2022-04-22
added rentman detector

lonmarsDev:

### DIFF
--- a/pkg/detectors/rentman/rentman.go
+++ b/pkg/detectors/rentman/rentman.go
@@ -1,0 +1,75 @@
+package rentman
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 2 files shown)


pkg/detectors/rentman/rentman.go
```diff
@@ -0,0 +1,75 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"rentman"}) + `\b([a-zA-Z0-9._]{300,500})\b`)
```
roxanne-tampus, line 24: suggestion: this regex can be made a bit stricter by adding `ey` as it always starts with `\b(ey[a-zA-Z0-9]{34}.ey[a-zA-Z0-9._-]{250,300})` and they are all split into two section by `.` separator so we can use that to our advantage.

dustin-decker (approved):

PR #458, opened 2022-04-22
added docparser detector

lonmarsDev:

### DIFF
--- a/pkg/detectors/docparser/docparser.go
+++ b/pkg/detectors/docparser/docparser.go
@@ -21,7 +21,7 @@ var (
 	client = common.SaneHttpClient()
 
 	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"docparser"}) + `\b([a-zA0-9]{40})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"docparser"}) + `\b([a-fA0-9]{40})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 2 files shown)


pkg/detectors/docparser/docparser.go
```diff
@@ -0,0 +1,74 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"docparser"}) + `\b([a-zA0-9]{40})\b`)
```
roxanne-tampus, line 24: suggestion: this regex can be made a bit stricter by `\b([0-9a-f]{40})\b` as the keys are in hex encoding so we can only work with `a-f`.

pkg/detectors/docparser/docparser.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Docparser,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.docparser.com/v1/parsers?api_key="+resMatch, nil)
```
roxanne-tampus, line 51: nit: Prefer fmt.Sprintf here: fmt.Sprintf("https://api.docparser.com/v1/parsers?api_key=%s", resMatch)

pkg/detectors/docparser/docparser.go
roxanne-tampus: Kindly remove "A" in the regex.

dustin-decker (approved):

PR #475, opened 2022-04-24
added flightlabs detector

ladybug0125:

### DIFF
--- a/pkg/detectors/flightlabs/flightlabs.go
+++ b/pkg/detectors/flightlabs/flightlabs.go
@@ -2,6 +2,7 @@ package flightlabs
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"regexp"
@@ -21,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"flightlabs"}) + `\b([a-zA-Z-0-9.-_]{359})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"flightlabs"}) + `\b(ey[a-zA-Z0-9]{34}.ey[a-zA-Z0-9._-]{300,350})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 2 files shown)


pkg/detectors/flightlabs/flightlabs.go
```diff
@@ -0,0 +1,84 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"flightlabs"}) + `\b([a-zA-Z-0-9.-_]{359})\b`)
```
roxanne-tampus, line 24: suggestion: this regex can be made a bit stricter by adding `ey` as it always starts with it `\b(ey[a-zA-Z0-9]{34}.ey[a-zA-Z0-9._-]{300,350})` and they are all split into two section by `.ey` separator so we can use that to our advantage. Also added a range to it since it is in jwt encoding.

pkg/detectors/flightlabs/flightlabs.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_FlightLabs,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://app.goflightlabs.com/airports?access_key="+resMatch, nil)
```
roxanne-tampus, line 51: nit: Prefer fmt.Sprintf here: fmt.Sprintf("https://app.goflightlabs.com/airports?access_key=%s", resMatch)

ladybug0125: Hi Roxanne, modified to use fmt.Sprintf(...). Thanks!

pkg/detectors/flightlabs/flightlabs.go
ladybug0125: Hi Roxanne, fixed the regex. Thank you!

pkg/detectors/flightlabs/flightlabs.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			res, err := client.Do(req)
+			if err == nil {
+				bodyBytes, err := ioutil.ReadAll(res.Body)
+				if err != nil {
+					continue
+				}
+				bodyString := string(bodyBytes)
+				validResponse := !strings.Contains(bodyString, `"invalid access key`)
```
roxanne-tampus, line 64: Kindly check for a substring in a valid response. Otherwise this can false positive under certain conditions (deploy, downtime, etc).

ladybug0125: I changed the valid response filter. Thank you!

dustin-decker (approved):

PR #494, opened 2022-04-30
fix: regex for hostnames

dcRUSTy:
The current regex for some of the detectors had very minor issues which can lead verification requests to a different domain then intended.
ex- regex for hooks.slack.com will also match hooks1slack.com, hooks2slack.com and so on as "." matches any character (except for line terminators)


dustin-decker (approved):
Thank you!

PR #497, opened 2022-05-02
[THOG-270] Fix scanning crash for git source without TLS.

ahrav:
## What?
Fix crash when trying to scan a git source without TLS/SSL.
## Why?
Should be able to scan a source without TLS/SSL.
## How?
Update the switch case statement to include `http` as a case in addition to `https`
## Testing?
Updated the unit tests to cover the new scenario.
## Screenshots (optional)
n/a
## Anything Else?

### DIFF
--- a/pkg/sources/git/git_test.go
+++ b/pkg/sources/git/git_test.go
@@ -421,6 +421,12 @@ func TestPrepareRepo(t *testing.T) {
 			err:    nil,
 		},
 		{
+			uri:    "http://github.com/dustin-decker/secretsandstuff.git",
+			path:   true,
+			remote: true,
+			err:    nil,
+		},
+		{
 			uri:    "file:///path/to/file.json",
 			path:   true,
 			remote: false,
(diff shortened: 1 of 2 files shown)


pkg/sources/git/git_test.go
```diff
@@ -315,3 +316,44 @@ func TestSource_Chunks_Edge_Cases(t *testing.T) {
...
+func TestPrepareRepo(t *testing.T) {
+	tests := []struct {
+		uri    string
+		path   string
+		remote bool
+		err    error
+	}{
+		{
+			uri:    "https://github.com/dustin-decker/secretsandstuff.git",
+			path:   "/var/folders/1m/m4kl5q2102bfr4k4566nmf800000gn/T/trufflehog2276311773",
```
dustin-decker, line 329: See the other MR comments, we shouldn't expect a particular path here because it's OS dependent.

ahrav: great point.

ahrav: worth noting this can probably made a bit better with some more thought, but this should also work for the time being as the concept is tested I think well enough.

dustin-decker (approved):

dustin-decker:
Merge conflict now

PR #500, opened 2022-05-02
Syslog source

bill-rich:
Adds a syslog monitoring source. Listens on the specified protocol/port and looks secrets that would be logged. Useful for checking logs in realtime as they're sent to a central logging service, so that expensive db queries do not need to be used to check later.


dustin-decker (approved):
There is solid feedback from the linter, but other than those, lgtm

PR #502, opened 2022-05-03
added brightlocal detector

lonmarsDev:


galevalerie (commented):
- can you generate another api key and check if it will generate letters after "f". if not, change the regex to match only letters from a-f
- i think the batch id is user-specific. can you verify that and make it a user input instead and not hardcoded

galevalerie (approved):

PR #505, opened 2022-05-03
added createsend detector

lonmarsDev:


galevalerie (commented):
- kindly add this at the end of the regex `[ \r\n]{1}` instead of just removing `\b`. you can also replace the initial `\b' from the regex with the code above in case the API key starts with special characters. with this we can make sure that the whole API key accurately matches with the length provided in the regex

zricethezav:
Closing this PR as createsend is now campaign monitor. We can revisit this detector in the future.

PR #506, opened 2022-05-03
added getresponse detector

lonmarsDev:

### DIFF
--- a/pkg/detectors/getresponse/getresponse.go
+++ b/pkg/detectors/getresponse/getresponse.go
@@ -1,0 +1,74 @@
+package getresponse
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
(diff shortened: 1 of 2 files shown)


pkg/detectors/getresponse/getresponse.go
```diff
@@ -0,0 +1,74 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"getresponse"}) + `\b([a-z0-9]{32})`)
```
roxanne-tampus, line 24: suggestion: this regex can be made stricter by closing the word boundary at the end `\b([a-z0-9]{32})\b` so the keys will match the length specified in the regex.

roxanne-tampus: Also, the test is no longer passing.
```

--- FAIL: TestGetresponse_FromChunk (2.21s)
    --- FAIL: TestGetresponse_FromChunk/found,_verified (0.82s)
        getresponse_test.go:97: Getresponse.FromData() found, verified diff: (-got +want)
             [
              {
               DetectorType: 861,
            -  Verified: false,
            +  Verified: true,
               Raw: [
               ],
               Redacted: "",
               ExtraData: {
               },
               StructuredData: nil,
              },
             ]
FAIL
FAIL    github.com/trufflesecurity/trufflehog/v3/pkg/detectors/getresponse      2.621s
FAIL
```

lonmarsDev: I change the expired key to a new one.

dustin-decker (approved):

PR #508, opened 2022-05-03
added besnappy detector

lonmarsDev:

### DIFF
--- a/pkg/detectors/besnappy/besnappy.go
+++ b/pkg/detectors/besnappy/besnappy.go
@@ -22,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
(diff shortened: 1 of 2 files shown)


pkg/detectors/besnappy/besnappy.go
```diff
@@ -0,0 +1,76 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"besnappy"}) + `\b([a-z0-9]{64})\b`)
```
roxanne-tampus, line 25: suggestion: this regex can be made a bit stricter by `\b([a-f0-9]{64})\b` as the keys are in hex encoding so we can only work with `a-f`.

dustin-decker (approved):

PR #513, opened 2022-05-03
added kanbantool detector

lonmarsDev:

### DIFF
--- a/pkg/detectors/kanbantool/kanbantool.go
+++ b/pkg/detectors/kanbantool/kanbantool.go
@@ -22,7 +22,7 @@ var (
 
 	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"kanbantool"}) + `\b([0-9A-Z]{12})\b`)
-	domainPat = regexp.MustCompile(detectors.PrefixRegex([]string{"kanbantoolDomain"}) + `\b([a-z0-9A-Z._]{4,22})\b`)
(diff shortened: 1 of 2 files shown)


pkg/detectors/kanbantool/kanbantool.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"kanbantool"}) + `\b([0-9A-Z]{12})\b`)
+	domainPat = regexp.MustCompile(detectors.PrefixRegex([]string{"kanbantoolDomain"}) + `\b([a-z0-9A-Z._]{4,22})\b`)
```
roxanne-tampus, line 25: suggestion: `kanbantoolDomain` is unlikely to be found, recommended to just use `kanbantool`.

lonmarsDev: if we use kanbantool on domainPat as regexString and not kanbantoolDomain, regex will also get the value of keyPat

roxanne-tampus, line 25: suggestion: After some testing, they allow a minimum of 2 characters and maybe we can go with `\b([a-z0-9A-Z._]{2,22})\b`?

lonmarsDev: this one already resolve.

dustin-decker (approved):

PR #517, opened 2022-05-03
Reorganize GitHub source

mcastorina:
This breaks up the Chunks method into smaller sub-method calls to help
organize and better understand the logic flow. No logic has been
modified (except one obvious bug), just shuffling code around.

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/credentialspb"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/source_metadatapb"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/sourcespb"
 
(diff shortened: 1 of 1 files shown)


pkg/sources/github/github.go
```diff
@@ -132,173 +141,192 @@ func (s *Source) Init(aCtx context.Context, name string, jobId, sourceId int64,
...
-				errUser := s.addReposByUser(ctx, apiClient, org)
-				if errOrg != nil && errUser != nil {
-					log.WithError(errOrg).Error("error fetching repos for org or user: ", org)
-				}
+	if len(s.orgs) > 0 {
+		specificScope = true
+		for _, org := range s.orgs {
+			errOrg := s.addReposByOrg(ctx, apiClient, org)
+			errUser := s.addReposByUser(ctx, apiClient, org)
+			if errOrg != nil || errUser != nil {
```
mcastorina, line 196: As mentioned in the overview, this is the bug-fix (changed `&&` to `||` in this if-statement).

mcastorina: Please correct me if I'm mistaken in making this change.

dustin-decker: It's a bit unfortunate on how this was setup, but one of these will always error. Both are attempted  because users and organizations have the same url format and can be put in the same field.

pkg/sources/github/github.go
```diff
@@ -132,173 +141,192 @@ func (s *Source) Init(aCtx context.Context, name string, jobId, sourceId int64,
...
+		return nil, errors.New(err)
+	}
 
-		s.addGistsByUser(ctx, apiClient, user.GetLogin())
+	// If no scope was provided, enumerate them
+	if !specificScope {
+		s.addReposByUser(ctx, apiClient, user.GetLogin())
+		// Scan for orgs is default with a token. GitHub App enumerates the repositories
+		// that were assigned to it in GitHub App settings.
+		s.addOrgsByUser(ctx, apiClient, user.GetLogin())
```
dustin-decker, line 212: We should probably heed the linter here, check these errors and log appropriately.

PR #518, opened 2022-05-03
added copyscape detector

lonmarsDev:


pkg/detectors/copyscape/copyscape.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"copyscape"}) + `\b([0-9a-z]{16})\b`)
+	idPat = regexp.MustCompile(detectors.PrefixRegex([]string{"copyscapeId"}) + `\b([0-9a-zA-Z._-]{4,22})\b`)
```
roxanne-tampus, line 26: could you please  `copyscape` instead of `copyscapeId`. Also can you verify that `.` works? I tested that it only allows letters, numbers, or underscores.

pkg/detectors/copyscape/copyscape.go
```diff
@@ -0,0 +1,89 @@
...
+					// }
+					body := string(bodyBytes)
+					//Used error as base for the fail safe, due to late response on success token
+					if !strings.Contains(body, "error") {
+						s1.Verified = true
+					} else {
+						if detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
+							continue
+						}
+					}
```
dustin-decker, line 79: Could we instead check against the balance API with o=balance, and verify the authenticated response?
Assuming it's verified if error is not in the response is not reliable. If there is a server error, unavailability, etc this check will become a false positive.
cc @roxanne-tampus

dustin-decker: https://www.copyscape.com/api-guide.php#balance

zricethezav:
Closing due to lack of relevance

PR #521, opened 2022-05-04
Cleanup packages and include more detectors

dustin-decker:
Looks like some packages got duplicated somehow during the move to open source.
Also imports some new detectors that were merged.


ahrav (approved):
lgtm 👍

PR #523, opened 2022-05-04
Fix overlooked nil

bill-rich:


pkg/engine/engine.go
```diff
@@ -256,5 +256,9 @@ func fragmentFirstLine(chunk *sources.Chunk) (int64, *int64) {
 	case *source_metadatapb.MetaData_Gerrit:
 		fragmentStart = &metadata.Gerrit.Line
 	}
```
mcastorina, line 258: **Nit:** A `default` case is a bit more readable to me

```go
// ...
default:
    return 0, nil
}
return *fragmentStart, fragmentStart
```

mcastorina (approved):

pkg/engine/engine.go
bill-rich: That is much better! Switched it out.

bill-rich: That is much better! Switched it out.

PR #524, opened 2022-05-04
Disable tests that take too long

bill-rich:
Enum tests are taking too long and causing testing to fail. I've commented out the offending tests. We should set up creds with a more limited scope to get these re-enabled.


mcastorina:
Ideally we could move these tests to an integration test stage in the pipeline.

bill-rich:
Agreed. Most (if not all) of the existing GH tests should be moved to an integration test. These couple will still need some work. They're covering far more than needed to test the functionality. 

This is more of a quick fix to get to where we at least have the benefit of our working tests again.

mcastorina (approved):

dustin-decker:
Longer term, build tags with integration tests in one file is probably the way to go https://mickey.dev/posts/go-build-tags-testing/

PR #529, opened 2022-05-08
[THOG-314] Add new parameter to the Init method for the source interface.

ahrav:
In order to support the oauth token refresher service we need to be able to pass the source service client as an argument to to the scanner in order to request an oauth token from the gRPC API layer.

Add an interface as an additional param to the source interface.

NOTE: this should ideally be converted to a concrete interface type. This will likely need to have the same signature as the `SourceService` rpc service used for the oauth token refresher service. If we make this it's own service the interface would only need to contain the `GetToken` method to satisfy the interface.

### DIFF
--- a/pkg/engine/filesystem.go
+++ b/pkg/engine/filesystem.go
@@ -23,7 +23,7 @@ func (e *Engine) ScanFileSystem(ctx context.Context, directories []string) error
 	}
 
 	fileSystemSource := filesystem.Source{}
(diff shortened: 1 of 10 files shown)


dustin-decker:
Hmm our SourceService interface is in the enterprise tree. So maybe it would be easier to add a separate RPC service to sources.proto?

ahrav:
> Hmm our SourceService interface is in the enterprise tree. So maybe it would be easier to add a separate RPC service to sources.proto?

Yea that was my thinking as well. I don't think we want to mix the two. I can go ahead and do that instead. Thanks.

dustin-decker (approved):

Issue #493, opened 2022-04-30
Provide Docker image to be used in GitHub action runs


channelbeta:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Build and upload the Docker image to GitHub container registry and update the GitHub action to use the built image.

## ~Problem~ Request to be Addressed

Currently, when using this as a GitHub action, the runner will build the Docker image every time it is activated. This is because `action.yml` has `image: "Dockerfile"` in the run configuration. You can see it happening [here](https://github.com/trufflesecurity/trufflehog/runs/6236421083?check_suite_focus=true), in the Dogfood task.

## Description of the Preferred Solution

Set up an automation workflow that will build the Docker image and upload it to GitHub's own container registry, then use that image as the run directive on `action.yml`. This is what I did [here](https://github.com/channelbeta/secret-scan/commit/18bf4b94de1762419b2630ae1c76f12552401b66) (I already had Docker build automation). This can save some time on GitHub action runs for everyone.

## Additional Context

Since I already did this setup once, I'm able to work on a PR, if the feature request is accepted by maintainers.

One possible caveat that I didn't test yet is the possibility to use `latest` as the Docker image tag, instead of using a fixed version. If the image tag *must* be versioned, then it would create additional work for maintainers, since each new release should be followed by a change on `action.yml` to update the version number (can be automated, of course). If the request is approved, I will test what works and what doesn't and report on the PR.

dustin-decker:
Thanks for the report.

We already build and push an image to ghcr, check out .goreleaser.yml.

Would you like to make a PR for it to use that?

dustin-decker:
You can find the ghcr images here: https://github.com/trufflesecurity/trufflehog/pkgs/container/trufflehog

dustin-decker:
Oh, I assumed that it would be public because the repo is. I fixed that.

Here you are: https://github.com/trufflesecurity/trufflehog/pkgs/container/trufflehog

Using the latest tag would be great.

dustin-decker:
Fixed in #534

PR #534, opened 2022-05-09
Update GitHub action metadata to use ghcr.io image

channelbeta:
Hi, this PR implements what we discussed on #493. You can see my test results [here](https://github.com/channelbeta/trufflehog/actions/workflows/testpr.yaml).

Runs 1, 2 and 8 were done using the old mode (`Dockerfile`). They took around 3 minutes of runner time just to build the images.

Runs 5, 6 and 7 took around **3 seconds** to pull the prebuilt image from ghcr.io :exploding_head: 

I'm using the `latest` tag, as requested, and it seems to be working fine with regards to action metadata requirements. However, as you can see by the commit history on the `main` branch (and also comparing results from workflow runs 7 and 8), I couldn't get 
the new workflow to find any secrets on the repo.

I'm raising the PR as a draft while I investigate if this is an issue with the metadata change. If there is something I'm overlooking, please leave a comment here.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -17,7 +17,7 @@ branding:
   color: "green"
 runs:
   using: "docker"
-  image: "Dockerfile"
+  image: "docker://ghcr.io/trufflesecurity/trufflehog:latest"
   args:
     - git
     - file://${{ inputs.path }}


dustin-decker:
The current arg changes are scanning the whole repo, but the previous are intending to only scan the differences from the default branch, which is what we desire for this action.

dustin-decker:
It looks like it is working as intended if you revert the arg changes.

channelbeta:
@dustin-decker run 9 confirms that, since it is a branch that just changed the version of workflow to use my 9.9.9 test release.

I've rebased this PR and it is ready for review.

dustin-decker (approved):
Thanks for the contribution!

PR #535, opened 2022-05-10
[THOG-327] Resuse the strings builder when writing each fragment.

ahrav:
Following up #534 reusing the strings.builder does provide a pretty nice speedup during the gitscan.

Rough test: (as a quick rough estimate using a decent sized repo)
Screen Shot 2022-05-09 at 7 45 46 PM

NOTE: ideally we can get some even better benchmarks, but for now this looks to be quite promising.

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -331,8 +331,8 @@ func (s *Git) ScanCommits(repo *git.Repository, path string, scanOptions *ScanOp
 			when = file.PatchHeader.AuthorDate.String()
 		}
 
+		var sb strings.Builder
 		for _, frag := range file.TextFragments {
-			var sb strings.Builder
 			newLineNumber := frag.NewPosition
 			for _, line := range frag.Lines {
 				if line.Op == gitdiff.OpAdd {
@@ -349,6 +349,7 @@ func (s *Git) ScanCommits(repo *git.Repository, path string, scanOptions *ScanOp
 				Data:           []byte(sb.String()),
 				Verify:         s.verify,
 			}
+			sb.Reset()
 		}
 	}
 	return nil


mcastorina (approved):
Nice!

pkg/sources/git/git.go
```diff
@@ -349,6 +349,7 @@ func (s *Git) ScanCommits(repo *git.Repository, path string, scanOptions *ScanOp
 				Data:           []byte(sb.String()),
 				Verify:         s.verify,
 			}
+			sb.Reset()
```
dustin-decker, line 352: This sets the buffer to nil on strings.Builder so it doesn't actually prevent reallocation of the buffer like bytes.Buffer's Reset() does. 

The total run time from your test is misleading. Notice the user CPU time is actually greater on the second test. Hard to say what the performance difference is with this change, but I'm guessing it's probably not much.

dustin-decker: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/strings/builder.go;l=59

dustin-decker: and for comparison, the bytes.Buffer Reset(): https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/bytes/buffer.go;l=97

mcastorina: Huh, I was curious why it doesn't do the same `b.buf = b.buf[:0]` for the string builder and found this: https://github.com/golang/go/issues/24716

The returned `string` data *is* the `[]byte` and so reusing it would mean modifying a previous returned `string` (which should be immutable).

ahrav: I wonder why there isn't say a `Clear()` method on the Builder? I think that could be potentially useful in order to keep that underlying array in memory vs it getting re-allocated.

Alas i'll poke at this a bit later on....

dustin-decker: I guess it looks like a design tradoff in the strings builder. They opted for being able to output a string from bytes with zero allocations instead of being able to reuse the buffer: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/strings/builder.go;l=47

Issue #385, opened 2022-04-12
Add detector name to JSON output


chr-b:
**Is your feature request related to a problem? Please describe.**
When running trufflehog with `--json` output, a `DetectorType` field is available in the output. This field is an integer value, not a string identifier.

**Describe the solution you'd like**
Add an additional field to the JSON output. E.g. `DetectorName` that provides the name of the secrets detector (e.g. `AWS` for the AWS access key detector).

**Describe alternatives you've considered**
I am not sure how to perform a dynamic mapping from detector type (the integer ID) to the detector name.

**Additional context**
I have only tested this with the Github command.

mvastola:
I was literally about to post something similar. Alternatively/additionally, is there a way to obtain a list showing the ID to name mappings? (Especially for processing existing results.)

mvastola:
@kstilwell that helps. Not sure why it wasn't coming up in search. Thanks!

dustin-decker:
We'd like to change this to be the string value of the enum. Right now it's just directly marshaling the protobuf type which marshals the enums as an integer in the JSON.

richa-p:
+1 to this request, it will be a lot easier to relate to the finding.

mattnedrich:
+1 to this as well

dustin-decker:
Fixed in #538, released in v3.4.5

PR #538, opened 2022-05-13
[THOG-336] Use the string version of the Detector Type.

ahrav:
Instead of printing the DetectorType Enum use the string. This makes the output message a lot easier to analyze to determine which detector found a secret.
[#385 ](https://github.com/trufflesecurity/trufflehog/issues/385)

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


dustin-decker (changes requested):
We should probably preserve DetectorType and add DetectorName as the string to avoid making a breaking change

PR #542, opened 2022-05-14
Added twitch detector

roxanne-tampus:


galevalerie (approved):
moving forward, i think you can use `url.Values{}` to store the request data and `application/x-www-form-urlencoded` for content type.

galevalerie (commented):
can you remove the other detector that is unrelated to this PR

dustin-decker (approved):

PR #543, opened 2022-05-14
Added cloudconvert detector

roxanne-tampus:

### DIFF
--- a/pkg/detectors/cloudconvert/cloudconvert.go
+++ b/pkg/detectors/cloudconvert/cloudconvert.go
@@ -1,0 +1,75 @@
+package cloudconvert
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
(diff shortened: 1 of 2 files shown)


galevalerie (commented):
kindly validate and update the regex pattern. i tried creating API keys and some are not matching with the provided regex. create API keys using different scopes.

roxanne-tampus:
Updated the regex with ranges. Please check if this could suffice your created keys. Thanks!

dustin-decker:
You can try running JWT tokens (long tokens with ey prefix) through a JWT parser like https://jwt.io/ to see if it contains variables that can be different lengths.

galevalerie (approved):
changes in the regex seems okay to me. I've also tested it with my created tokens and with the jwt decoder.

dustin-decker (approved):

PR #549, opened 2022-05-14
Added twist detector

roxanne-tampus:

### DIFF
--- a/pkg/detectors/twist/twist.go
+++ b/pkg/detectors/twist/twist.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"net/url"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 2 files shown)


roxanne-tampus:
Updated this for auth enhancement. Kindly check, thanks!

roxanne-tampus:
Nice catch! I just pushed the changes.

pkg/detectors/twist/twist.go
```diff
@@ -0,0 +1,77 @@
...
+	dataStr := string(data)
+
+	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+
+	for _, match := range matches {
+		if len(match) != 2 {
+			continue
+		}
+		resMatch := strings.TrimSpace(match[1])
+		setAuth := resMatch
```
galevalerie, line 44: you can remove this part since we are expecting to detect tokens with this prefix already

pkg/detectors/twist/twist.go
```diff
@@ -0,0 +1,77 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	//Make sure that your group is surrounded in boundry characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"twist"}) + `\b([0-9a-f:]{40,47})\b`)
```
galevalerie, line 24: i think you should include `oauth2:` in the regex pattern. we expect that the token already contains `oauth2:`

roxanne-tampus: No we can't include the oauth2: in the regex pattern because token generated using Basic authentication doesn't have oauth2 in it. Please check it here https://developer.twist.com/v3/#login.

galevalerie (approved):

dustin-decker (approved):

PR #550, opened 2022-05-14
Added zipcodebase detector

roxanne-tampus:


galevalerie (commented):
- kindly change the regex so that it can match the exact pattern of the API Key.

galevalerie (approved):

dustin-decker (approved):

PR #552, opened 2022-05-14
Added duply detector

roxanne-tampus:


galevalerie (commented):
- kindly update both regex to match the exact pattern of each token.

galevalerie (approved):

dustin-decker (approved):

PR #555, opened 2022-05-14
Added ecostruxureit detector

roxanne-tampus:


galevalerie (commented):
-can you generate another token and check if the pattern is the same for all tokens? if yes, can you match the regex with the exact pattern of the token.

galevalerie (approved):

dustin-decker (approved):

PR #564, opened 2022-05-19
Shallow clone if --since-commit is provided

mcastorina:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -18,6 +18,7 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/google/go-github/v42/github"
 	"github.com/rs/zerolog"
 	log "github.com/sirupsen/logrus"
 	glgo "github.com/zricethezav/gitleaks/v8/detect/git"
@@ -201,7 +202,7 @@ func CleanOnError(err *error, path string) {
 	}
 }
 
-func CloneRepo(userInfo *url.Userinfo, gitUrl string) (clonePath string, repo *git.Repository, err error) {
+func CloneRepo(userInfo *url.Userinfo, gitUrl string, args ...string) (clonePath string, repo *git.Repository, err error) {
 	if err = GitCmdCheck(); err != nil {
 		return
 	}
@@ -217,8 +218,10 @@ func CloneRepo(userInfo *url.Userinfo, gitUrl string) (clonePath string, repo *g
 		return
 	}
 
+	gitArgs := []string{"clone", cloneURL.String(), clonePath}
+	gitArgs = append(gitArgs, args...)
 	cloneURL.User = userInfo
-	cloneCmd := exec.Command("git", "clone", cloneURL.String(), clonePath)
+	cloneCmd := exec.Command("git", gitArgs...)
 
 	output, err := cloneCmd.CombinedOutput()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/sources/git/git.go
```diff
@@ -472,7 +475,71 @@ func TryAdditionalBaseRefs(repo *git.Repository, base string) (*plumbing.Hash, e
...
+	uri, err := url.Parse(uriString)
+	if err != nil {
+		return "", false, fmt.Errorf("unable to parse Git URI: %s", err)
+	}
+
+	if uri.Scheme == "file" || uri.Host != "github.com" {
+		return PrepareRepo(uriString)
+	}
+
+	owner, repoName, found := strings.Cut(uri.Path, "/")
```
mcastorina, line 499: Oops.. looks like `strings.Cut` was added in go1.18

dustin-decker: #566 upgrades to 1.18 for CI and builds

dustin-decker:
Do you think the test failure might be related?

mcastorina:
It could be? The failure *is* related to cloning, but the tests pass locally for me. I'm rerunning the action now.

mcastorina:
Still failing :/

pkg/sources/git/git.go
```diff
@@ -472,7 +475,71 @@ func TryAdditionalBaseRefs(repo *git.Repository, base string) (*plumbing.Hash, e
...
+	var ts string
+	{
+		author := commit.GetAuthor()
+		if author == nil {
+			return PrepareRepo(uriString)
+		}
+		ts = author.GetDate().Format(time.RFC3339)
+	}
+
+	remotePath := fmt.Sprintf("%s://%s%s", uri.Scheme, uri.Host, uri.Path)
```
ahrav, line 518: are you missing a trailing `/` after the second `%s`

mcastorina: That's how the original code I copied this from does it, and I believe `uri.Path` begins with a leading `/`

pkg/sources/git/git.go
```diff
@@ -216,9 +217,11 @@ func CloneRepo(userInfo *url.Userinfo, gitUrl string) (clonePath string, repo *g
 		err = errors.WrapPrefix(err, "could not parse url", 0)
 		return
 	}
-
 	cloneURL.User = userInfo
-	cloneCmd := exec.Command("git", "clone", cloneURL.String(), clonePath)
+
+	gitArgs := []string{"clone", cloneURL.String(), clonePath}
```
mcastorina, line 222: Found the issue - I accidentally constructed `gitArgs` before setting `cloneURL.User` here (fixed in latest commit). Strange that it didn't happen locally, but I'm glad the CI caught it :)

pkg/sources/git/git.go
```diff
@@ -472,7 +475,71 @@ func TryAdditionalBaseRefs(repo *git.Repository, base string) (*plumbing.Hash, e
...
+	if !found {
+		return PrepareRepo(uriString)
+	}
+
+	client := github.NewClient(nil)
+	commit, _, err := client.Git.GetCommit(context.Background(), owner, repoName, commitHash)
+	if err != nil {
+		return PrepareRepo(uriString)
+	}
+	var ts string
```
bill-rich, line 509: nit: `ts` could be changed to something longer like `timestamp` so it's easier to understand what it is.

pkg/sources/git/git.go
bill-rich, line 518: Since you already have the url parsed, you could use `uri.String()` instead of reassembling it manually.

mcastorina: Ah yeah, good point!

PR #568, opened 2022-05-22
modified Alibaba detector to use standard library

roxanne-tampus:
REFERENCE: https://www.alibabacloud.com/help/en/elastic-compute-service/latest/api-reference
SIGNATURE: https://www.alibabacloud.com/help/en/elastic-compute-service/latest/request-signatures


pkg/detectors/alibaba/alibaba.go
```diff
@@ -17,68 +24,109 @@ type Scanner struct{}
...
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b([a-zA-Z0-9]{30})\b`)
+	idPat  = regexp.MustCompile(`\b(LTAI[a-zA-Z0-9]{17,21})[\"';\s]*`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"LTAI"}
+	return []string{"alibaba"}
```
dustin-decker, line 37: We should use `LTAI` as the keyword here because it is a unique prefix that is always part of the ID pattern. This substring will always be present but `alibaba` may not, so it is a better choice.

pkg/detectors/alibaba/alibaba.go
```diff
@@ -17,68 +24,109 @@ type Scanner struct{}
...
+func GetSignature(input, key string) string {
+	key_for_sign := []byte(key)
+	h := hmac.New(sha1.New, key_for_sign)
+	h.Write([]byte(input))
+	return base64.StdEncoding.EncodeToString(h.Sum(nil))
+}
+func buildStringToSign(input string) string {
+	filter := strings.Replace(input, "+", "%20", -1)
+	filter = strings.Replace(filter, "*", "%2A", -1)
+	filter = strings.Replace(filter, "%7E", "~", -1)
```
dustin-decker, line 60: Are the parameters swapped here? It looks like the other two are swapping the symbol for the url encoded version.

dustin-decker (approved):

dustin-decker:
Great job! Happy to remove some dependencies we don't need.

PR #579, opened 2022-05-24
Modified aws detector to use standard library 

roxanne-tampus:
References: https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
Calculate Signature: https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html
Code Sample in Python: https://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html#sig-v4-examples-get-query-string

### DIFF
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -2,11 +2,14 @@ package aws
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


dustin-decker (approved):
Great job, this one is a beast!

PR #584, opened 2022-05-25
README Total Detectors badge & small improvements

MNThomson:
# Description
[Total Detectors](/pkg/detectors)
Add a [shields.io](https://shields.io/) badge to display the total detectors & add YAML highlighting to GH Action example

## General Comments
This badge works by counting the directories in the `pkg/detectors` folder. Weirdly, this functionality is broken on shields production, so the suggestion from [badges/shields#5967](https://github.com/badges/shields/issues/5967) is to use the staging instance.
The LICENSE badge now links directly to the `LICENSE` file instead of an image of the badge.
Additionally, because a newline is missing in the GH Action example, the first line is cut off & the syntax highlighting is broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,14 @@
 
 ---
 
+<center>
 
 [![CI Status](https://github.com/trufflesecurity/trufflehog/actions/workflows/release.yml/badge.svg)](https://github.com/trufflesecurity/trufflehog/actions)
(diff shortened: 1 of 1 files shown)


dustin-decker:
This is really cool! Thanks for the contribution :)

PR #587, opened 2022-05-25
Increase default timeout for SaneHttpClient, but fail early.

dustin-decker:


dustin-decker:
Increases the default overall timeout on the SaneHttpClient to 5 seconds, but also allows it to fail early on connect and a few other timeouts which should be fast.

ahrav (approved):

PR #588, opened 2022-05-26
Enhancement in Gitlab detector

roxanne-tampus:


pkg/detectors/gitlab/gitlab_test.go
```diff
@@ -15,7 +15,7 @@ import (
 func TestGitlab_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors2")
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors4")
```
dustin-decker, line 18: Can we also test both tokens?

dustin-decker (approved):

PR #589, opened 2022-05-26
added common regex patterns for detectors

roxanne-tampus:
Below detectors has the imported function for common regex patterns.

1. ZulipChat
- common.BuildRegex(common.Base64Pattern, ..
- EmailPattern
- SubDomainPattern

2. CloudConvert
- JWT

3. Databox
- common.BuildRegex(common.RegexPattern, ..

4. Onesignal
- UUIDPattern

5. ClickSendSMS
- UUIDPatternUpperCase
- EmailPattern

6. CodeMagic
- BuildRegexWithChar

7: SpeechTextAI
- HexPattern

### DIFF
--- a/pkg/common/patterns.go
+++ b/pkg/common/patterns.go
@@ -1,0 +1,56 @@
+package common
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const EmailPattern = `\b([a-z0-9]{4,25}@[a-zA-Z0-9]{2,12}.[a-zA-Z0-9]{2,6})\b`
+const SubDomainPattern = `\b([0-9a-zA-Z]{2,40})\b`
+const UUIDPattern = `\b([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\b`
+const UUIDPatternUpperCase = `\b([0-9A-Z]{8}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{12})\b`
+
+const RegexNumPattern = "0-9"
+const RegexPattern = "0-9a-z"
+const RegexPatternUpperCase = "0-9A-Z"
+const Base64Pattern = "0-9a-zA-Z"
+const HexPattern = "0-9a-f"
+const HexPatternUpperCase = "0-9A-F"
+const HexPatternWithUpperCase = "0-9a-fA-F"
+
+//Custom Regex functions
+func BuildRegex(pattern string, length int) string {
+	return fmt.Sprintf(`\b([%s]{%s})\b`, pattern, strconv.Itoa(length))
+}
+
+func BuildRegexWithChar(pattern, specialChar string, length int) string {
+	return fmt.Sprintf(`\b([%s%s]{%s})\b`, pattern, specialChar, strconv.Itoa(length))
+}
+
+func BuildRegexJWT(firstRange, secondRange, thirdRange string) string {
+	first_range_split := strings.Split(firstRange, ",")
+	first_range_min := strings.TrimSpace(first_range_split[0])
+	first_range_max := strings.TrimSpace(first_range_split[1])
+	if first_range_min >= first_range_max {
+		log.Error("First min value should not be greater than or equal to max")
+	}
+
+	second_range_split := strings.Split(secondRange, ",")
+	second_range_min := strings.TrimSpace(second_range_split[0])
+	second_range_max := strings.TrimSpace(second_range_split[1])
+	if second_range_min >= second_range_max {
+		log.Error("Second min value should not be greater than or equal to max")
+	}
+
+	third_range_split := strings.Split(thirdRange, ",")
+	third_range_min := strings.TrimSpace(third_range_split[0])
+	third_range_max := strings.TrimSpace(third_range_split[1])
+	if third_range_min >= third_range_max {
+		log.Error("Third min value should not be greater than or equal to max")
+	}
+
+	return fmt.Sprintf(`\b(ey[%s]{%s}.ey[%s-\/_]{%s}.[%s-\/_]{%s})\b`, Base64Pattern, firstRange, Base64Pattern, secondRange, Base64Pattern, thirdRange)
+}
--- a/pkg/common/patterns.go
+++ b/pkg/common/patterns.go
@@ -8,8 +8,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const EmailPattern = `\b([a-z0-9]{4,25}@[a-zA-Z0-9]{2,12}.[a-zA-Z0-9]{2,6})\b`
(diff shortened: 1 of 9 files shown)


pkg/common/patterns.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const EmailPattern = `\b([a-z0-9]{4,25}@[a-zA-Z0-9]{2,12}.[a-zA-Z0-9]{2,6})\b`
+const SubDomainPattern = `\b([0-9a-zA-Z]{2,40})\b`
```
dustin-decker, line 12: This is just alphanumeric. Should it instead be a hostname pattern? (require a valid TLD, etc)

pkg/common/patterns.go
```diff
@@ -0,0 +1,56 @@
...
+const UUIDPattern = `\b([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\b`
+const UUIDPatternUpperCase = `\b([0-9A-Z]{8}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{12})\b`
+
+const RegexNumPattern = "0-9"
+const RegexPattern = "0-9a-z"
+const RegexPatternUpperCase = "0-9A-Z"
+const Base64Pattern = "0-9a-zA-Z"
```
dustin-decker, line 19: This is an Alphanumeric pattern

pkg/common/patterns.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+const EmailPattern = `\b([a-z0-9]{4,25}@[a-zA-Z0-9]{2,12}.[a-zA-Z0-9]{2,6})\b`
+const SubDomainPattern = `\b([0-9a-zA-Z]{2,40})\b`
+const UUIDPattern = `\b([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\b`
+const UUIDPatternUpperCase = `\b([0-9A-Z]{8}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{12})\b`
+
+const RegexNumPattern = "0-9"
+const RegexPattern = "0-9a-z"
+const RegexPatternUpperCase = "0-9A-Z"
```
dustin-decker, line 18: These are not very descriptive and probably simple enough that they should be removed.

pkg/common/patterns.go
```diff
@@ -0,0 +1,56 @@
...
+func BuildRegexWithChar(pattern, specialChar string, length int) string {
+	return fmt.Sprintf(`\b([%s%s]{%s})\b`, pattern, specialChar, strconv.Itoa(length))
+}
+
+func BuildRegexJWT(firstRange, secondRange, thirdRange string) string {
+	first_range_split := strings.Split(firstRange, ",")
+	first_range_min := strings.TrimSpace(first_range_split[0])
+	first_range_max := strings.TrimSpace(first_range_split[1])
+	if first_range_min >= first_range_max {
+		log.Error("First min value should not be greater than or equal to max")
```
dustin-decker, line 38: These should be a panic, which will be caught during the test. Also, does numerical comparison work on strings?

pkg/common/patterns.go
```diff
@@ -0,0 +1,56 @@
...
+	"strconv"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const EmailPattern = `\b([a-z0-9]{4,25}@[a-zA-Z0-9]{2,12}.[a-zA-Z0-9]{2,6})\b`
+const SubDomainPattern = `\b([0-9a-zA-Z]{2,40})\b`
+const UUIDPattern = `\b([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\b`
+const UUIDPatternUpperCase = `\b([0-9A-Z]{8}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{4}-[0-9A-Z]{12})\b`
```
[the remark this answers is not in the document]
dustin-decker: I was trying to comment just on line 12

pkg/common/patterns.go
dustin-decker: I was just trying to comment on line 19

pkg/common/patterns.go
roxanne-tampus: Based on what I've read here `https://stackoverflow.com/questions/31487694/how-can-i-convert-string-to-integer-in-golang`, as long as it doesn't need the fractional part, then it will work. I tested it in playground as well `https://go.dev/play/p/Oc-_X_cFzrd`

pkg/common/patterns.go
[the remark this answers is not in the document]
roxanne-tampus: Are we going to include the hostname along with the subdomain in GCP as well?

pkg/common/patterns.go
dustin-decker: The string comparison fails with "100,22" and "33,222"

pkg/common/patterns.go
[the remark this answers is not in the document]
dustin-decker: I'm not sure what you mean. Can you clarify?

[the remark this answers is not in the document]
roxanne-tampus: In GCP for subdomain we have like this `ZULIPCHAT_DOMAIN=secretscanner` so if we'll add it, the new value for this is `ZULIPCHAT_DOMAIN=https://secretscanner.zulipchat.com`. Then the pattern will be something like this `^(([a-zA-Z0-9-_]{1,61}[a-zA-Z0-9]))\.([a-zA-Z]{2,6}|[a-zA-Z0-9-]{2,30}\.[a-zA-Z]{2,3})$`

pkg/common/patterns.go
roxanne-tampus: Oh, thanks for pointing this out! I just pushed the changes and used parseInt()

pkg/common/patterns.go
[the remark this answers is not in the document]
dustin-decker: I see. Can you look at using the information here? https://stackoverflow.com/questions/7930751/regexp-for-subdomain

From the top-voted answer, 

>>>Each subdomain part must have a length no greater than 63.
Each subdomain part must begin and end with an alpha-numeric (i.e. letters [A-Za-z] or digits [0-9]).
Each subdomain part may contain hyphens (dashes), but may not begin or end with a hyphen.

pkg/common/patterns.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const EmailPattern = `\b([a-z0-9]{4,25}@[a-zA-Z0-9]{2,12}.[a-zA-Z0-9]{2,6})\b`
```
dustin-decker, line 11: Check out the answer here for the email regex from the latest RFC on it: https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

dustin-decker:
Merged in #654

PR #591, opened 2022-05-28
enhancements for doppler detector

galevalerie:
enhancements for doppler detector.
added test cases for the different token types.

### DIFF
--- a/pkg/detectors/doppler/doppler.go
+++ b/pkg/detectors/doppler/doppler.go
@@ -20,6 +20,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	// Note: For SCIM Token, you need to have an Enterprise subsciption and a verified domain where you can add a DNS Configuration given by the service
 	personalTokenPat = regexp.MustCompile(detectors.PrefixRegex([]string{"doppler"}) + `\b(dp\.pt\.[a-zA-Z0-9]{40,44})\b`)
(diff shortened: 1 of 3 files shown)


pkg/detectors/doppler/doppler.go
```diff
@@ -20,20 +20,35 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(dp\.pt\.[a-zA-Z0-9]{43})\b`)
+	// Note: For SCIM Token, you need to have an Enterprise subsciption and a verified domain where you can add a DNS Configuration given by the service
+	personalTokenPat = regexp.MustCompile(detectors.PrefixRegex([]string{"doppler"}) + `\b(dp\.pt\.[a-zA-Z0-9]{40,44})\b`)
+	serviceTokenPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"doppler"}) + `\b(dp\.st\.(?:[a-z0-9\-_]{2,35}\.)?[a-zA-Z0-9]{40,44})\b`)
+	auditTokenPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"doppler"}) + `\b(dp\.audit\.[a-zA-Z0-9]{40,44})\b`)
```
dustin-decker, line 26: These are specific enough that you can drop the prefix regex portion. That will allow us to catch more results by not requiring the `doppler` string within 40 chars of the match

zricethezav:
The doppler detector has since been updated to support all the different token types

PR #600, opened 2022-06-02
Update README.md

cyberqueen-meg:
Hello, 

Thanks for creating a great tool! I found another way to install trufflehog using the blackarch repository in Arch Linux, so I added that installation method to the README.md file to alert others to that installation option


dustin-decker:
Thank you for the contribution, but I do not want to add this repository because it is not an official arch repository, we do not control it and do not know if it is safe.

cyberqueen-meg:
No problem, thank you for your time @dustin-decker

PR #601, opened 2022-06-03
Allow github to resume from encoded resume info

trufflesteeeve:
This allows github to be resumed from whatever repos it was last working on.

I'm still working on fixing up a couple of the tests (I think I found a failing test case but am not sure why it is failing). But I wanted to get this out now to answer some questions. Like, this changes the randomization behavior of repo scanning. It would now be sorted. Is that an okay change to the defaults, or should we put it behind a flag?


dustin-decker:
Sorted is fine.

dustin-decker (approved):

mcastorina (approved):

PR #602, opened 2022-06-04
added mollie access token detector

galevalerie:
added mollie access token detector

### DIFF
--- a/pkg/detectors/mollieaccesstoken/mollieaccesstoken.go
+++ b/pkg/detectors/mollieaccesstoken/mollieaccesstoken.go
@@ -1,0 +1,80 @@
+package mollieaccesstoken
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


pkg/detectors/mollieaccesstoken/mollieaccesstoken.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+		log.Println(resMatch)
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_MollieAccessToken,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.mollie.com/v2/customers", nil)
```
roxanne-tampus, line 52: Could we just use the endpoint that list all the permissions granted to an access token? This endpoint can cause a false positive if a specific access is not granted to a token.

wywwzjj (approved):
It looks good.

PR #603, opened 2022-06-04
added mollie api key detector

galevalerie:
added mollie api key detector


pkg/detectors/mollieapikey/mollieapikey.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		resMatch := strings.TrimSpace(match[1])
+
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_MollieAPIKey,
+			Raw:          []byte(resMatch),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.mollie.com/v2/customers", nil)
```
roxanne-tampus, line 51: Could we just use the endpoint that list all the permissions granted to an access token? This endpoint can cause a false positive if a specific access is not granted to a token.

roxanne-tampus (approved):

PR #610, opened 2022-06-07
Initialize resumeInfoMutex in github

trufflesteeeve:
I forgot to initialize the mutex in github.


trufflesteeeve:
Already completed. Good!

PR #611, opened 2022-06-07
Allow gitlab to resume from encoded resume info

trufflesteeeve:
This also includes a commit to pull out the common functions between gitlab and github into potentially reusable ones for other sources. I put it in `resume.go` in the sources package, but could definitely see it making more sense elsewhere.


pkg/sources/resume.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+// RemoveRepoFromResumeInfo removes the repoURL from the resume info.
+func RemoveRepoFromResumeInfo(resumeRepos []string, repoURL string) []string {
+	index := -1
+	for i, repo := range resumeRepos {
+		if repoURL == repo {
+			index = i
```
mcastorina, line 14: **Nit (optional):** I know this is simply moving code around, but we can `break` out of this loop once we found the repo. Alternatively, since the `resumeRepos` are sorted, we could do a binary search using [sort.SearchStrings](https://pkg.go.dev/sort#SearchStrings)

pkg/sources/gitlab/gitlab.go
```diff
@@ -241,8 +245,8 @@ func (s *Source) getAllProjects(apiClient *gitlab.Client) ([]*gitlab.Project, er
 	return projectList, nil
 }
 
-func (s *Source) getRepos() ([]*url.URL, []error) {
-	var validRepos []*url.URL
+func (s *Source) getRepos() ([]string, []error) {
+	var validRepos []string
 	var errs []error
 	if len(s.repos) > 0 {
```
mcastorina, line 251: **Nit (optional):** I prefer structuring code to "fail early" which helps keep nested indentation low.

```go
if len(s.repos) == 0 {
    return nil, nil
}
// rest of logic that was previously inside the `if len(s.repos) > 0` block
```

pkg/sources/gitlab/gitlab.go
```diff
@@ -252,22 +256,24 @@ func (s *Source) getRepos() ([]*url.URL, []error) {
...
-func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk, repos []*url.URL) []error {
+func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk) []error {
 	wg := sync.WaitGroup{}
 	var errs []error
 	var errsMut sync.Mutex
 
-	for i, u := range repos {
+	// If there is resume information available, limit this scan to only the repos that still need scanning.
+	reposToScan, progressIndexOffset := sources.FilterReposToResume(s.repos, s.GetProgress().EncodedResumeInfo)
+	s.repos = reposToScan
```
mcastorina, line 274: **Question:** Is it okay to be possibly dropping some information here?

pkg/sources/gitlab/gitlab.go
```diff
@@ -252,22 +256,24 @@ func (s *Source) getRepos() ([]*url.URL, []error) {
 				continue
 			}
 
-			// The repo normalization has already successfully parsed the URL at this point, so we can ignore the error.
-			u, _ := url.ParseRequestURI(repo)
-			validRepos = append(validRepos, u)
+			validRepos = append(validRepos, repo)
```
mcastorina, line 259: **Question:** I'm not sure if we can trust the `repo` url directly or if we should do `url.ParseRequestURI` followed by `String()`. I would think parsing would help normalize the input, but perhaps it's redundant at this point.

pkg/sources/gitlab/gitlab.go
```diff
@@ -386,3 +405,18 @@ func (s *Source) basicAuthSuccessful(apiClient *gitlab.Client) bool {
...
 }
+
+// setProgressCompleteWithRepo calls the s.SetProgressComplete after safely setting up the encoded resume info string.
+func (s *Source) setProgressCompleteWithRepo(index int, repoURL string) {
+	s.resumeInfoMutex.Lock()
+	defer s.resumeInfoMutex.Unlock()
+
+	// Add the repoURL to the resume info slice.
+	s.resumeInfoSlice = append(s.resumeInfoSlice, repoURL)
+	sort.Strings(s.resumeInfoSlice)
```
mcastorina, line 416: **Nit (optional):** We could more efficiently insert `repoURL` into the slice since it should already be sorted, though I kind of like the stability of sorting every time.

pkg/sources/gitlab/gitlab.go
```diff
@@ -418,5 +418,6 @@ func (s *Source) setProgressCompleteWithRepo(index int, repoURL string) {
 	// Make the resume info string from the slice.
 	encodedResumeInfo := sources.EncodeResumeInfo(s.resumeInfoSlice)
 
-	s.SetProgressComplete(index, len(s.repos), fmt.Sprintf("Repo: %s", repoURL), encodedResumeInfo)
+	// Add the offset to both the index and the repos to give the proper place and proper repo count.
```
mcastorina, line 421: **Suggestion:** It would be nice to have a test for this to prevent regressions. If adding it would take too much time / effort, let's open an issue so it can be planned.

mcastorina (approved):
Nice work! Just had some optional nits / suggestions.

pkg/sources/resume.go
trufflesteeeve: Ah that's neat. I do like the `break`. But looking at `sort.SearchStrings`, it could return an index that we wouldn't want to use, because it doesn't actually contain the `repoURL`, and we'd have to check that that index did exactly equal the repo. But still very cool to know about.

pkg/sources/gitlab/gitlab.go
trufflesteeeve: Totally agree.

pkg/sources/gitlab/gitlab.go
trufflesteeeve: Are you referring to the dropping of the repositories that have already been scanned?

If so, I think it's fine. We don't actually store that information anywhere, as it's generated at the start of a scan. And the progressIndexOffset will allow us to get the proper count of the original number of repos.

However if you're referring to the potential to drop repos that may have been added between the time the scan started and when it was picked back up, I think that is okay. The goal here is to reduce the amount of time it takes to finish the scan, so the next scan will happen faster, and that scan will pick up the new repository.

pkg/sources/gitlab/gitlab.go
trufflesteeeve: Ah. The repo URL was already normalized when `giturl.NormalizeGitlabRepo(prj)` was run on it above. That eventually calls `url.Parse` here: https://github.com/trufflesecurity/trufflehog/blob/4218c39d99b5f30153f62471c1be1c1596f0a4d4/pkg/giturl/giturl.go#L41

Though now that you mention it, maybe `giturl` should be using `url.ParseRequestURI` instead of just `url.Parse`.

pkg/sources/gitlab/gitlab.go
trufflesteeeve: True. Maybe something to keep an eye on, as though I don't expect anyone to have like 10k+ repos to sort nor the concurrency to have that many listed in the resume info, you never know.

pkg/sources/gitlab/gitlab.go
mcastorina: This `if` condition must be inverted.

mcastorina (changes requested):
Thanks for adding the test! Just one thing needs to be fixed before merging.

mcastorina (approved):

PR #613, opened 2022-06-07
Mark detector tests with a build flag

mcastorina:


dustin-decker:
I'm concerned with how this affects the accessibility of testing. Would it be better to just filter the path like you've done?

dustin-decker (approved):

PR #614, opened 2022-06-07
Cleanup individual repositories after scanning

trufflesteeeve:
The deferred os.RemoveAlls mean that anytime there's a scan of a lot of repositories, they won't be deleted from disk until the entire scan is done. This fixes that.

But it also changes some behavior - currently if there's an error getting or scanning a repo, the scan ends. Should we keep that? I can easily change this to do so. Or should it log the issue and continue scanning, as I've currently implemented it? (Though for directories, the current behavior remains in place)


dustin-decker (approved):

dustin-decker:
We should probably preserve the current behavior. That can still be done with these anonymous functions.

PR #615, opened 2022-06-08
improve debug logging for GHE enum

dustin-decker:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -464,6 +464,8 @@ func handleRateLimit(errIn error, res *github.Response) bool {
 }
 
 func (s *Source) getReposByOrg(ctx context.Context, apiClient *github.Client, org string) ([]string, error) {
+	log := s.log.WithField("org", org)
+
 	repos := []string{}
 	opts := &github.RepositoryListByOrgOptions{
 		ListOptions: github.ListOptions{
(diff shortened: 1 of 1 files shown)


pkg/sources/github/github.go
```diff
@@ -464,6 +464,8 @@ func handleRateLimit(errIn error, res *github.Response) bool {
 }
 
 func (s *Source) getReposByOrg(ctx context.Context, apiClient *github.Client, org string) ([]string, error) {
+	log := log.WithField("org", org)
```
trufflesteeeve, line 467: __optional__: I'm wondering if it would be good to attach this field on top of the existing `s.log` which looks like this: ```s.log = log.WithField("source", s.Type()).WithField("name", name)```

So like, `log := s.log.WithField("org", org)`

Or, does this make for a cleaner log message, so not worth adding the source and source name.

pkg/sources/github/github.go
```diff
@@ -674,6 +677,7 @@ func (s *Source) addReposByApp(ctx context.Context, apiClient *github.Client) er
 }
 
 func (s *Source) addAllVisibleOrgs(ctx context.Context, apiClient *github.Client) {
+	log.Debug("enumerating all visibile organizations on GHE")
```
trufflesteeeve, line 680: __optional__: Unless it unnecessarily complicates the log message, might be best to use `s.log` for all of these `log.Debug`s that were added.

trufflesteeeve (approved):
Looks good. Just some optional comments for potentially using the sources default logger instead of the global logger.

pkg/sources/github/github.go
dustin-decker: Good suggestion

PR #618, opened 2022-06-09
Change GHE org enum to use since ID instead of pages

dustin-decker:
Enumeration on this endpoint does not use pages. it uses a since ID.
The endpoint will return organizations with an ID greater than the given since ID.
Empty org response is our cue to break the enumeration loop.

https://docs.github.com/en/rest/orgs/orgs#list-organizations

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -703,8 +703,8 @@ func (s *Source) addAllVisibleOrgs(ctx context.Context, apiClient *github.Client
 			break
 		}
 		lastOrgID := *orgs[len(orgs)-1].ID
-		orgOpts.Since = lastOrgID
 		s.log.Debugf("listed organization IDs %d through %d", orgOpts.Since, lastOrgID)
+		orgOpts.Since = lastOrgID
 
 		for _, org := range orgs {
 			var name string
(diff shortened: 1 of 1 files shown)


pkg/sources/github/github.go
```diff
@@ -695,7 +699,13 @@ func (s *Source) addAllVisibleOrgs(ctx context.Context, apiClient *github.Client
 			log.WithError(err).Errorf("Could not list all organizations")
 			return
 		}
-		s.log.Debugf("listed organization page %d/%d", orgOpts.Page, resp.LastPage)
+		if len(orgs) == 0 {
+			break
+		}
+		lastOrgID := *orgs[len(orgs)-1].ID
+		orgOpts.Since = lastOrgID
+		s.log.Debugf("listed organization IDs %d through %d", orgOpts.Since, lastOrgID)
```
bill-rich, line 707: `orgOpts.Since` was set to `lastOrgID` in the previous line so this will read "listed organization IDs X through X". Looks like moving it up a line should get the results you want.

bill-rich (approved):
1 issue on log content, but looks good.

pkg/sources/github/github.go
dustin-decker: Thanks Bill!

PR #619, opened 2022-06-10
Only include one result per AWS Key ID, preferably verified

trufflesteeeve:
Also ignore unverified results that match hashes, because they are
probably just hashes.

I'm currently working on tests for this, but wanted to get the logic out early to make sure we're comfortable with these changes. They will essentially increase the number of unverified results we return, while also helping us ignore the fact that any kind of hash technically matches the shape of an AWS secret, with an incredibly tiny probability of actually being one. But we'll still attempt to verify any hashes, so in case they are live, we'll still include them.


dustin-decker (approved):
Looks good!

PR #637, opened 2022-06-29
[THOG-531] - remove ending word boundary AWS detector

ahrav:
Remove the ending word boundary for the AWS dectector. This will prevent missing secrets that end with / due to it not being ASCII.

Note: I think this should be okay given we limit the regex to 40 chars and thus shouldn't over match? Let me know if there is a better approach.

https://github.com/trufflesecurity/trufflehog/issues/624


mcastorina:
Hm.. my regex skillz are lacking, so I would feel a lot more confident that this works if we had some tests for it. It will be good for preventing regressions as well.

dustin-decker:
I think the biggest danger to having no boundary is the potential for having many secret matches for 40+ char alphanumeric strings that might be in the file

dustin-decker:
How about we replace the trailing `\b` with `[ \r\n'"`]{1}` outside of the capturing group?

ahrav:
I don't think you can use ` in a string literal, but i think that pattern looks good. 
`(`\b([A-Za-z0-9+/]{40})[ \r\n'"\x60]`)`

PR #646, opened 2022-07-07
Disable sentiment verifier and update teams webhook to avoid being stateful

dustin-decker:


samdatkins:
I don't have much context surrounding this change or this section of the codebase, but it looks sane to me

samdatkins (approved):

PR #647, opened 2022-07-08
Add SlackTokens option for slack credentials

trufflesteeeve:
Is there a way to do this without the `tokens` name, and instead having just `bot_token` and `app_token` be the credential? I don't _think_ so but I am less familiar with protobufs.

### DIFF
--- a/proto/credentials.proto
+++ b/proto/credentials.proto
@@ -59,3 +59,8 @@ message GitHubApp {
   string installation_id = 2;
   string app_id = 3;
 }
+
+message SlackTokens {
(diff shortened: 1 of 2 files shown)


dustin-decker:
It's needed to ensure that we can differentiate on credential types. 
It is possible to include it as top level fields, but then you lose out on having different credential types and need to have application logic sift through sparse fields (ie. token is blank but app_token and bot_token are not, and the reverse. also the invalid combinations).

dustin-decker (approved):

PR #653, opened 2022-07-12
Add codeowners file

dustin-decker:
Adds an initial codeowners file with critical detectors defined [THOG-504]


dustin-decker:
> Looks like `@trufflesecurity/detection` doesn't exist

It does, but might not be visible to you. I'm the only one in it.

Issue #540, opened 2022-05-13
Make Github Action customizable


timmeinerzhagen:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

## Problem to be Addressed

The current options for usage of the GitHub Actions for Trufflehog are very limited. Only the options `path`, `base`, and `head` are allowed. This is not enough for many use cases.

## Description of the Preferred Solution

Additional options like `no-verify`, `include-` / `exclude-paths`, and output options would be needed to make it usable in many cases.
Ideally, an custom output for the GitHub Action for Trufflehog could leverage [Job Summary for GitHub Actions](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/), which would be a lot more usable instead of having to rely on the runner logs.

### Option 1: Include the current options that are available for the cli in the GitHub Action 

Example
```yml
- name: TruffleHog OSS
  uses: trufflesecurity/trufflehog@v3.4.3
  with:
    path: ./
    base: ${{ github.event.repository.default_branch }}
    no-verify: true
    output: json
```

### Option 2: Add a parameter like `options` to the GitHub Action that could be used to just specify cli options that will just be appended when trufflehog gets executed.

Example
```yml
- name: TruffleHog OSS
  uses: trufflesecurity/trufflehog@v3.4.3
  with:
    path: ./
    base: ${{ github.event.repository.default_branch }}
    options: --no-verify --json
```

### References

* [Current options for TruffleHog OSS Github Action](https://github.com/trufflesecurity/trufflehog#trufflehog-oss-github-action)
* [Job Summary for GitHub Actions](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/)

dustin-decker:
Either option would be a welcome contribution. Would you be able to take this on?

Cr0n1c:
It would be great to have the include-paths with an external file (e.g. --include-paths https://somesite.com/company_regex_file)

PR #670, opened 2022-07-25
Update github action to support generic cli command input

jwenz723:
This PR partially resolves https://github.com/trufflesecurity/trufflehog/issues/540. Specifically the portion of https://github.com/trufflesecurity/trufflehog/issues/540 regarding adding support for all trufflehog input arguments.

This is one possible way which can generically support passing in all of the available commands and arguments available for the trufflehog cli.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,6 @@ FROM alpine:3.15
 RUN apk add --no-cache git
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=builder /build/trufflehog /usr/bin/trufflehog
-ENTRYPOINT ["/usr/bin/trufflehog"]
+COPY entrypoint.sh /etc/entrypoint.sh
+RUN chmod +x /etc/entrypoint.sh
+ENTRYPOINT ["/etc/entrypoint.sh"]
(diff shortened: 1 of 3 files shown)


jwenz723:
Note that since declaring expressions in default values for github actions inputs isn't supported I have simplified the default command to `git file://./` which is slightly different than the default command as it currently exists in branch `main` where it is required that the caller of the action specify `inputs.path` and `inputs.base`.

dustin-decker:
This is great, thank you! Unfortunately it's a breaking change at the moment. In retrospect we probably should have versioned the GitHub action separately... 

What if we preserved the current inputs and arguments and expanded the `cli_cmd` args separately after those?

jwenz723:
@dustin-decker I've made the change as you suggested. I was able to validate this in my open github actions workflow using the step:

```yaml
      - name: TruffleHog OSS
        uses: jwenz723/trufflehog@f72ed383cee5994b1c430b5ded1bef6243f8c3eb
        with:
          path: "./"
          base: ${{ github.base_ref }}
          head: ${{ github.head_ref }}
          extra_args: --only-verified
```

I have `jwenz723/trufflehog@f72ed383cee5994b1c430b5ded1bef6243f8c3eb` because in that particular version of the action it makes use of [a docker image which I pushed into my own registry](https://hub.docker.com/layers/trufflehog/jwenz723/trufflehog/latest/images/sha256-3c58e615fdf922c282c265e42e89fc44b6c3989b58f43ad54663f2909d8a5612?context=explore) so that I could test with the entrypoint.sh script compiled into the image.

bill-rich (approved):
Looks good. Thanks @jwenz723 for the contribution!

PR #683, opened 2022-08-01
Archive decoder

bill-rich:
Adding file handlers along with an archive handler.


dustin-decker:
Please address the linter findings, looks like it caught some good stuff.

pkg/handlers/handlers.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+
+type Handler interface {
+	FromFile(io.Reader) chan ([]byte)
+	IsFiletype(io.Reader) bool
+	New()
+}
+
+func HandleFile(file io.Reader, chunkSkel *sources.Chunk, chunksChan chan (*sources.Chunk)) bool {
```
dustin-decker, line 21: Is there a way to handle the filename within an archive in the metadata?

pkg/sources/filesystem/filesystem.go
```diff
@@ -106,13 +108,37 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
...
+			if handlers.HandleFile(inputFile, chunkSkel, chunksChan) {
+				return nil
+			}
+
+			if err != nil && !errors.Is(err, archiver.ErrNoMatch) {
+				log.WithError(err).Debug("Error reading archive")
+				return nil
+			}
+
 			reader := bufio.NewReaderSize(bufio.NewReader(inputFile), BufferSize)
```
dustin-decker, line 140: At this point it's only safe to continue here if the reader is consumed (cursor is at EOF, processed by the handler) or if the original reader is untouched (no cursor advance)

pkg/handlers/archive.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		newReader := bytes.NewReader(fileBytes)
+		return d.openArchive(ctx, depth+1, newReader, archiveChan)
+	}
+	return fmt.Errorf("Unknown archive type: %s", format.Name())
+}
+
+// IsFiletype returns true if the provided reader is an archive.
+func (d *Archive) IsFiletype(reader io.Reader) bool {
+	format, readerB, err := archiver.Identify("", reader)
```
dustin-decker, line 97: We must return readerB here to be used for further processing. The Handler interface must be updated and it should include a comment that a new reader must be returned.

>The returned io.Reader will always be non-nil and will read from the same point as the reader which was passed in; it should be used in place of the input stream after calling Identify() because it preserves and re-reads the bytes that were already read during the identification process.

https://pkg.go.dev/github.com/mholt/archiver/v4#Identify

dustin-decker (changes requested):
I think there's a bug here re: reader consumption that needs fixing. It would also be nice to have some test coverage on this new feature.

pkg/handlers/handlers.go
bill-rich: The issue is that it would need to be added to the metadata so we'd need to include a way to handle each metadata source type. We could work around that by having the handler chunk handling be done by the source. This seemed like the best course since it makes the handlers easy to add to sources.

pkg/sources/filesystem/filesystem.go
bill-rich: Ended up with a mix of a couple of methods here. Got it all sorted out.

bill-rich:
I wasn't sure how long it would take to get the git binary stuff fixed, so I left the tests off this PR. The plan was to include them in a PR that uses an updated gitleaks fork so we can easily test against a repo with the artifacts.

bill-rich:
Nevermind on the tests in the git PR. After some more thought on it I rewrote it to not depend on git.

dustin-decker (approved):

pkg/handlers/archive.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// openArchive takes a reader and extracts the contents up to the maximum depth.
+func (d *Archive) openArchive(ctx context.Context, depth int, reader io.Reader, archiveChan chan ([]byte)) error {
+	if depth >= maxDepth {
+		return fmt.Errorf("max archive depth reached")
+	}
+	format, reader, err := archiver.Identify("", reader)
+	if err != nil {
+		if errors.Is(err, archiver.ErrNoMatch) && depth > 0 {
```
mcastorina, line 60: **Nit:** Negating this to make the if body smaller will make it easier to follow for me.

pkg/handlers/archive.go
```diff
@@ -0,0 +1,163 @@
...
+			for {
+				chunk := make([]byte, 10*1024)
+				n, _ := reader.Read(chunk)
+				archiveChan <- chunk
+				if n < 512 {
```
mcastorina, line 65: **Question:** Why 512 here?

pkg/handlers/archive.go
```diff
@@ -0,0 +1,163 @@
...
+					break
+				}
+			}
+			fileBytes, err := d.ReadToMax(reader)
```
mcastorina, line 69: **Question:** Wouldn't the reader already be empty at this point?

mcastorina (approved):
Nice job! I just had a few questions

pkg/handlers/archive.go
bill-rich: Missed this when switching chunk size to match filesystem chunk sizes

pkg/handlers/archive.go
bill-rich: It would. I forgot to remove this when merging several methods of handling archives.

PR #684, opened 2022-08-02
Support scanning binary files in git sources

bill-rich:


bill-rich:
Depends on #683

trufflesteeeve (approved):
Looks good!

pkg/sources/git/git.go
```diff
@@ -616,3 +634,49 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
 	}
 	return safeURL
 }
+
+func handleBinary(repo *git.Repository, chunksChan chan (*sources.Chunk), chunkSkel *sources.Chunk, commitHash plumbing.Hash, path string) error {
```
ahrav, line 638: suggestion: unwrap the second arg type.
`func handleBinary(repo *git.Repository, chunksChan chan *sources.chunk, ...)`

pkg/sources/git/git.go
```diff
@@ -616,3 +634,49 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
+
+	reader, err := file.Reader()
+	if err != nil {
+		return err
+	}
+	handled := handlers.HandleFile(reader, chunkSkel, chunksChan)
+
+	if !handled {
+		log.WithField("path", path).Trace("Binary file is not recognized by file handlers. Chunking raw.")
+		chunkSize := 10 * 1024
```
ahrav, line 658: nit: might be worth pulling this into a constant for a bit more clarity.
const bufferSize = 10 * 1024

aside: Might be worth down the road getting this logic into a more central place. I notice we have this calculation littered across the codebase. Some places we use a const others we don't.

pkg/sources/git/git.go
```diff
@@ -616,3 +634,49 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
+	file, err := commit.File(path)
+	if err != nil {
+		return err
+	}
+
+	reader, err := file.Reader()
+	if err != nil {
+		return err
+	}
+	handled := handlers.HandleFile(reader, chunkSkel, chunksChan)
```
ahrav, line 654: nit: You can remove the temp `handled` var by flipping the if check logic.
Ex:
```
if handlers.HandleFile(reader, chunkSkel, chunksChan) {
    return nil
}

log.WithField("path", path).Trace("Binary file is not recognized by file handlers. Chunking raw.")
chunkSize := 10 * 1024
reader, err = file.Reader()
if err != nil {
    return err
}
eof := false
for !eof {
    chunkData := make([]byte, chunkSize)
    n, err := reader.Read(chunkData)
    if err != nil {
        if !errors.Is(err, io.EOF) {
        return err
    }
    eof = true
}
if n < chunkSize {
    eof = true
}

chunk := *chunkSkel
chunk.Data = chunkData
chunksChan <- &chunk
}

return nil
```

ahrav (approved):
lgtm

pkg/sources/git/git.go
bill-rich: Funny you should mention that #703

PR #687, opened 2022-08-02
Add Honeycomb detector

MNThomson:
# Description
Add a [Honeycomb.io](https://www.honeycomb.io/) API key detector

## General Comments
Currently using the [auth](https://docs.honeycomb.io/api/auth/) endpoint (returns key's permissions) to verify the API key
All API keys require `X-Honeycomb-Team:`, but stuck instead with just `Honeycomb` as the keyword (for variables and such)


dustin-decker:
Thank you for the contribution! We'll add a credential and review this soon.

dustin-decker:
FYI I had to update the pattern in https://github.com/trufflesecurity/trufflehog/pull/687/commits/ad0d7f8b298a8ba250cec53fd20dee875546d834 for the test to pass. Please confirm it still works for you.

dustin-decker (approved):
Thanks again for your contribution! Sorry it took a while to get around to testing and merging.

MNThomson:
> FYI I had to update the pattern in [ad0d7f8](https://github.com/trufflesecurity/trufflehog/commit/ad0d7f8b298a8ba250cec53fd20dee875546d834) for the test to pass. Please confirm it still works for you.

Wow, my regex somehow got an additional `A` in there. However, my tests are now failing locally.
```bash
$ CGO_ENABLED=0 go test -tags=detectors -timeout=5m github.com/trufflesecurity/trufflehog/v3/pkg/detectors/honeycomb
--- FAIL: TestHoneycomb_FromChunk (0.00s)
    --- FAIL: TestHoneycomb_FromChunk/found,_verified (0.00s)
        honeycomb_test.go:100: Honeycomb.FromData() found, verified diff: (-got +want)
             [
            + {
            +  DetectorType: 877,
            +  Verified: true,
            +  Raw: [
            +  ],
            +  RawV2: [
            +  ],
            +  Redacted: "",
            +  ExtraData: {
            +  },
            +  StructuredData: nil,
            + },
             ]
    --- FAIL: TestHoneycomb_FromChunk/found,_unverified (0.00s)
        honeycomb_test.go:100: Honeycomb.FromData() found, unverified diff: (-got +want)
             [
            + {
            +  DetectorType: 877,
            +  Verified: false,
            +  Raw: [
            +  ],
            +  RawV2: [
            +  ],
            +  Redacted: "",
            +  ExtraData: {
            +  },
            +  StructuredData: nil,
            + },
             ]
FAIL
FAIL    github.com/trufflesecurity/trufflehog/v3/pkg/detectors/honeycomb        0.025s
FAIL
```
All HC API keys that I've ever created have been `[0-9a-f]{32}`, but it seems you updated to `[0-9Aa-z-A-Z]{22}`. Not sure if the length change was a mistake or you generated different looking keys than I.

For reference:
```bash
$ echo -n $HONEYCOMB | wc -c
32
```

Having the expanded char space still lets my tests pass, but the change in length now prevents the regex from matching the correct key

dustin-decker:
How are you generating these keys? The ones I've generated are 22 chars. Maybe it's an old token format?

dustin-decker:
Also, I made a few typos in the pattern, i've corrected it to be `[0-9a-zA-Z]{22}` but it has the same effect.

MNThomson:
Alright, did some digging. As of the [HC Classic Update](https://docs.honeycomb.io/honeycomb-classic/classic-changes/) on [June 15th](https://changelog.honeycomb.io/), the API keys switched from being `[0-9a-f]{32}` -> `[0-9a-zA-Z]{22}`.

For reference I have both a an older "classic" as well as a newer "environment" HC account. The API keys I used to test are from the old account, but my newer accounts keys are all of length `22`.

I feel the best option would be to include both formats, thoughts?

MNThomson:
@dustin-decker I can put together another PR. I'd assume a lot of older api keys are still being used, so I think it's best to include both into `\b([0-9a-f]{32}|[0-9a-zA-Z]{22})\b`. Thoughts?

dustin-decker:
Yeah, let’s support both. Thanks for digging into that! PR would be
appreciated.

PR #700, opened 2022-08-09
[Thog-371] Utilize config struct for engine scans

ahrav:
- Define a config struct within sources that encompasses all the configurations used for all the engine scanning integrations.
- Construct each config within main for each source using functional args.
- Pass the config as the second argument to each of the engine scan methods.
- Update all the tests.

### DIFF
--- a/main.go
+++ b/main.go
@@ -182,7 +182,7 @@ func run(state overseer.State) {
 
 	var repoPath string
 	var remote bool
-	var cfg sources.Config
+	cfg := sources.NewConfig()
 	switch cmd {
 	case gitScan.FullCommand():
 		repoPath, remote, err = git.PrepareRepoSinceCommit(*gitScanURI, *gitScanSinceCommit)
@@ -193,13 +193,16 @@ func run(state overseer.State) {
 			defer os.RemoveAll(repoPath)
 		}
 
-		cfg = sources.Config{
-			RepoPath: repoPath,
-			HeadRef:  *gitScanBranch,
-			BaseRef:  *gitScanSinceCommit,
-			MaxDepth: *gitScanMaxDepth,
-			Filter:   filter,
+		setup := func(c *sources.Config) {
+			c.RepoPath = repoPath
+			c.HeadRef = *gitScanBranch
+			c.BaseRef = *gitScanSinceCommit
+			c.MaxDepth = *gitScanMaxDepth
+			c.Filter = filter
 		}
+
+		cfg = sources.NewConfig(setup)
+
 		if err = e.ScanGit(ctx, cfg); err != nil {
 			logrus.WithError(err).Fatal("Failed to scan git.")
 		}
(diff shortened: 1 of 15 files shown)


pkg/sources/sources.go
```diff
@@ -42,6 +44,57 @@ type Source interface {
 	GetProgress() *Progress
 }
 
+// Config defines the optional configuration for a source.
```
mcastorina, line 47: This is an interesting approach that I hadn't thought of. It's nice that we have only one struct for everything, but my concern is it's not clear which options are used for which scan types.

Or do all of the scans support all of these options?

main.go
```diff
@@ -189,36 +192,73 @@ func run(state overseer.State) {
...
-		if err != nil {
+
+		cfg = sources.Config{
+			RepoPath: repoPath,
+			HeadRef:  *gitScanBranch,
+			BaseRef:  *gitScanSinceCommit,
+			MaxDepth: *gitScanMaxDepth,
+			Filter:   filter,
+		}
+		if err = e.ScanGit(ctx, &cfg); err != nil {
```
mcastorina, line 203: **Question/Suggestion:** What is the reasoning for using a pointer here for the `cfg`?

I would prefer passing the struct so there's no possiblity of null pointer dereference. Alternatively, if we need to pass a pointer, we should check for a nil config at the beginning of each of the scan functions.

pkg/sources/sources.go
ahrav: I wasn't sure on the req on the tix tbh. I just went with this approach because it made things a bit simpler. Also, I figured doing it this way we can leverage the zero value for the fields in the struct. Then we just configure the struct for each scan independently.

I contemplated a variadic fxn as well, but went with this easier approach. I can make separate structs for each scan, if that's what was needed.

main.go
ahrav: Will do.

pkg/sources/sources.go
ahrav: Update to use functional args. I think this makes it a bit better?

mcastorina: Ooh interesting. And sorry the ticket was underspecified :/

I also hadn't considered using functional args like that. I've seen that pattern used in other libraries except the library would have public functions that returned a private type so they had more control over the options.

I like that it minimizes the risk of null pointers, though technically it's still possible to pass one. We're the only ones using this code so I think it's an acceptable amount of risk though.

> I can make separate structs for each scan, if that's what was needed.

The intention of the ticket was to replace a long list of arguments with something to make the code easier to maintain, update, and read. Imo your solution ticks those boxes.

pkg/sources/sources.go
```diff
@@ -42,6 +44,66 @@ type Source interface {
...
+type Config struct {
```
mcastorina, line 48: **Praise:** Great comments in this struct!

mcastorina (approved):

main.go
```diff
@@ -189,36 +191,78 @@ func run(state overseer.State) {
...
+
+		github := func(c *sources.Config) {
+			c.Endpoint = *githubScanEndpoint
+			c.Repos = *githubScanRepos
+			c.Orgs = *githubScanOrgs
+			c.Token = *githubScanToken
+			c.IncludeForks = *githubIncludeForks
+			c.IncludeMembers = *githubIncludeMembers
+			c.Concurrency =
+				*concurrency
```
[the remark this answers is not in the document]
ahrav: weird. Not sure why this happened. Thanks

pkg/engine/git.go
```diff
@@ -49,15 +54,15 @@ func (e *Engine) ScanGit(ctx context.Context, repoPath, headRef, baseRef string,
...
+	if len(c.HeadRef) > 0 {
+		headHash := plumbing.NewHash(c.HeadRef)
+		if !plumbing.IsHash(c.HeadRef) {
+			head, err := git.TryAdditionalBaseRefs(repo, c.HeadRef)
 			if err != nil {
 				return errors.WrapPrefix(err, "unable to resolve head ref", 0)
 			} else {
-				headRef = head.String()
-				headCommit, _ = repo.CommitObject(plumbing.NewHash(baseRef))
+				c.HeadRef = head.String()
```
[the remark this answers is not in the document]
ahrav: yea i did realize that, but figured that it was okay since as you mentioned we only use the cfg in the switch.

PR #703, opened 2022-08-09
Use re-readable reader and common chunker

bill-rich:
This PR has two parts. Adding re-readable readers and adding a common chunking function.

Part 1:
File handlers require that readers be read more than once. Some sources readers are more difficult than others to use efficiently. For example, an HTTP response could be fully read and stored in memory, or two requests could be made, but both have their pitfalls. Using disk-buffer-reader makes it so we can use the same method everywhere. It allows readers to be reset to the beginning by utilizing a tmp file as a buffer.

Part 2:
Each source that deals with files re-implements chunking. A common chunking function gives us consistency and is easier to use.

### DIFF
--- a/pkg/common/chunker.go
+++ b/pkg/common/chunker.go
@@ -23,6 +23,7 @@ func ChunkReader(r io.Reader) chan []byte {
 			n, err := reader.Read(chunk)
 			if err != nil && !errors.Is(err, io.EOF) {
 				log.WithError(err).Error("Error chunking reader.")
+				break
 			}
 			peekData, _ := reader.Peek(PeekSize)
 			chunkChan <- append(chunk[:n], peekData...)
--- a/pkg/common/chunker.go
+++ b/pkg/common/chunker.go
@@ -1,0 +1,35 @@
+package common
+
+import (
+	"bufio"
+	"errors"
+	"io"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	ChunkSize = 10 * 1024
+	PeekSize  = 3 * 1024
+)
+
+func ChunkReader(r io.Reader) chan []byte {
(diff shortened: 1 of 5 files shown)


pkg/common/chunker.go
```diff
@@ -0,0 +1,35 @@
...
+	chunkChan := make(chan []byte)
+	go func() {
+		defer close(chunkChan)
+		reader := bufio.NewReaderSize(bufio.NewReader(r), ChunkSize)
+		for {
+			chunk := make([]byte, ChunkSize)
+			n, err := reader.Read(chunk)
+			if err != nil && !errors.Is(err, io.EOF) {
+				log.WithError(err).Error("Error chunking reader.")
+			}
```
mcastorina, line 26: **Question:** Should we `continue` here?

bill-rich: Good catch. I think we should `break` to avoid getting stuck in the loop.

mcastorina:
I'm curious about the implications this will have on the speed of chunking. I don't think it's a blocker as the current bottleneck is in detection, but just curious.

mcastorina (approved):

dustin-decker:
Shouldn't we use the TeeReader instead of copying to disk? https://stackoverflow.com/a/39792097/11976023

bill-rich:
@mcastorina In most cases nothing extra is being written to disk, so there is no performance hit. 

@dustin-decker I was originally going to use TeeReader, but wanted to use something that could provide an interface that would work like a standard reader. TeeReader writes to the writer as the first reader is read from. Since the first reader is not guaranteed to be fully read, the two readers need to be joined with something like io.MultiReader. That combined with concerns about memory usage if a large file is tee'd made me figure out another option.

PR #707, opened 2022-08-10
Use file handler and common chunker

bill-rich:
Add file handlers so archives in S3 can be unpacked, and use the common chunker for other objects.

### DIFF
--- a/pkg/sources/s3/s3.go
+++ b/pkg/sources/s3/s3.go
@@ -3,7 +3,6 @@ package s3
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/sources/s3/s3.go
```diff
@@ -302,14 +284,27 @@ func (s *Source) pageChunker(ctx context.Context, client *s3.S3, chunksChan chan
...
+				return
+			}
+
+			if err := reader.Reset(); err != nil {
+				log.WithError(err).Error("Error resetting reader to start.")
+			}
+			reader.Stop()
+
+			for chunkData := range common.ChunkReader(reader) {
+				chunk := *chunkSkel
```
mcastorina, line 297: I'm not certain here, but wouldn't this create a shallow copy? `SourceMetadata` is still a pointer so every `chunk` in this loop will point to the same `SourceMetadata` object.

Maybe a closure would be better:

```go
makeChunk := func(chunkData IdkTheType) *sources.Chunk {
    return &sources.Chunk{ /* ... */ }
}
```

bill-rich: It does, but all of chunks created will be the same except for `Data`.

mcastorina: I think that'll be okay as long as we're only ever reading the chunk SourceMetadata.

mcastorina (approved):

PR #709, opened 2022-08-12
[chore] - Use config struct instead of pointer for engine scans.

ahrav:
Use a config struct instead of pointer when scanning engine sources.
Remove the nil check from each scan.

### DIFF
--- a/main.go
+++ b/main.go
@@ -200,7 +200,7 @@ func run(state overseer.State) {
 			c.Filter = filter
 		}
 
-		if err = e.ScanGit(ctx, sources.NewConfig(g)); err != nil {
+		if err = e.ScanGit(ctx, *sources.NewConfig(g)); err != nil {
 			logrus.WithError(err).Fatal("Failed to scan Git.")
 		}
 	case githubScan.FullCommand():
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -200,7 +200,7 @@ func run(state overseer.State) {
 			c.Filter = filter
 		}
 
-		if err = e.ScanGit(ctx, sources.NewConfig(g)); err != nil {
+		if err = e.ScanGit(ctx, *sources.NewConfig(g)); err != nil {
```
mcastorina, line 203: **Suggestion:** make `sources.NewConfig` return the struct instead of a pointer so we don't have to deref here.

mcastorina (approved):
Nice!

main.go
ahrav: yea i had thought about going that route. I guess I left it this way since it was the standard way that everyone is used to seeing this functional options implemented. One of those things where when you take a quick glance you just kinda go, "oh its doing that" and almost automatically think it should return a pointer. Similar to how we are used to seeing `err != nil` and when we see `err == nil` it usually requires a double take.

That being said I don't have a strong personal pref on this.

mcastorina: Cool, thanks for sharing your reasoning! I'm not familiar with that pattern so it didn't occur to me.

PR #710, opened 2022-08-12
[Thog-628] update detector results hash v2

ahrav:
- include RawV2 to all the detectors that require the multi part creds.
- This will prevent collisions if different secrets have the same ID.


mcastorina (approved):
Looks like tedious work, thanks for doing it!

PR #715, opened 2022-08-16
Use pointer to config

bill-rich:
A pointer is needed for the config maker function otherwise the functions aren't applied to the returned config.


dustin-decker (approved):

ahrav:
Nice catch!

Issue #735, opened 2022-08-26
Allow secrets (tokens and access keys) to be passed as env vars


fenak:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Allow secrets to be passed to `trufflehog` as environment variables.

I got a PR opened for this, see #736 

## Problem to be Addressed

When scanning private repos on a Github organization, repos on Gitlab or S3 buckets, the secrets need to be passed via cli arguments, which might expose those secrets in logs and files like `.bash_history`.

## Description of the Preferred Solution

It should be possible to pass those secrets as env vars. Using AWS as an example, when running the `trufflehog` container in a cloud environment like AWS ECS, the secrets can be injected from AWS Secrets Manager as environment variables, like described here: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data-secrets.html#secrets-envvar

## Additional Context

edit: added reference to opened PR

### References

* #736

dustin-decker:
Added in #736

dustin-decker:
This was released in v3.10!

https://github.com/trufflesecurity/trufflehog/releases/tag/v3.10.0

PR #736, opened 2022-08-26
Allow secrets to be passed as env vars

fenak:
This change enables the following secrets to be passed as environment variables:
- Github token
- Gitlab token
- AWS access keys for S3 scanning

There are no automated tests for the cli parameters, and I haven't worked on that. I manually tested to see the change working for Github, Gitlab, and S3.

Open to suggestions for the env var names. I tried to make them as descriptive as possible, and used the familiar names for AWS access keys (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`) as they are the well known env vars for them, although they are still being prefixed to explicitly relate to `trufflehog` S3 scanning.

This PR is related to #735, please refer to it for more context.

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,7 +60,7 @@ var (
 	githubScanEndpoint   = githubScan.Flag("endpoint", "GitHub endpoint.").Default("https://api.github.com").String()
 	githubScanRepos      = githubScan.Flag("repo", `GitHub repository to scan. You can repeat this flag. Example: "https://github.com/dustin-decker/secretsandstuff"`).Strings()
 	githubScanOrgs       = githubScan.Flag("org", `GitHub organization to scan. You can repeat this flag. Example: "trufflesecurity"`).Strings()
-	githubScanToken      = githubScan.Flag("token", "GitHub token.").Envar("TRUFFLEHOG_GITHUB_TOKEN").String()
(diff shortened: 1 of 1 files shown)


dustin-decker:
Hey @fenak, thanks for the contribution!

Do you think we should drop the `TRUFFLEHOG_` prefix and stick with the well-known env vars? Or might that cause people to use the wrong ones, or unintentionally use them? I think it'll be a bit easier to use if we drop the prefix.

fenak:
@dustin-decker The well-known are the AWS ones. I thought about that, ended up erring on the explicit side. The `TRUFFLEHOG_` prefix clearly shows intention. 

For AWS, it's common for tooling that interacts with the AWS SDKs to use the well-known variables, like `AWS_PROFILE`, `AWS_ACCESS_KEY_ID`, etc, as the AWS SDK would look for them to handle authentication. So it could be a case for dropping the prefix.

Not sure about `GITHUB_TOKEN` and `GITLAB_TOKEN`, but they are obvious enough as is, and happy to drop the prefix.

I added some details for the CLI `--help` as well, so people can learn about the new capability.

dustin-decker (approved):

PR #737, opened 2022-08-26
Import 27 new detectors

dustin-decker:


bill-rich (approved):
Assuming the original detectors are good, the changes lgtm.

PR #738, opened 2022-08-27
Add common sentry recover library and add into goroutines

dustin-decker:
This will enable us to capture panics if sentry is configured.

### DIFF
--- a/pkg/common/recover.go
+++ b/pkg/common/recover.go
@@ -1,0 +1,29 @@
+package common
+
+import (
+	"fmt"
+	"os"
+	"runtime/debug"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+)
+
+// Recover recovers panics and reports to Sentry before exiting
+func Recover(ctx context.Context) {
+	if err := recover(); err != nil {
+		panicStack := string(debug.Stack())
+		eventID := sentry.CurrentHub().Recover(err)
+		if eventID != nil {
+			ctx.Logger().Info("panic captured", "event_id", *eventID)
+		}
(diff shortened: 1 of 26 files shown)


pkg/common/recover.go
```diff
@@ -0,0 +1,29 @@
...
+	"os"
+	"runtime/debug"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+)
+
+// Recover recovers panics and reports to Sentry before exiting
```
ahrav, line 14: nit: `Recover panics and report to Sentry before exiting.` Slight less stutter.

pkg/common/recover.go
```diff
@@ -0,0 +1,29 @@
...
+func Recover(ctx context.Context) {
+	if err := recover(); err != nil {
+		panicStack := string(debug.Stack())
+		eventID := sentry.CurrentHub().Recover(err)
+		if eventID != nil {
+			ctx.Logger().Info("panic captured", "event_id", *eventID)
```
ahrav, line 20: nit: inline
```
if eventID := sentry.CurrentHub().Recover(err); eventID != nil {
    ctx.Logger().Info("panic captured", "event_id", *eventID)
}
```

pkg/common/recover.go
```diff
@@ -0,0 +1,29 @@
...
+		}
+		fmt.Fprint(os.Stderr, panicStack)
+		flushed := sentry.Flush(time.Second * 5)
```
ahrav, line 23: ditto: Albeit more optional here.
```
if !sentry.Flush(time.Second * 5) {
    ctx.Logger().Info("sentry flush failed")
}
```

ahrav (approved):
Very nifty. I like it 👏

Issue #724, opened 2022-08-19
Codacy detector should be more specific and should not use /version endpoint for token verification


dinvlad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently, Codacy detector detects lines like
```
2022-08-19 19:22:38.155Z  info [ReportRules] Generated coverage report: /tmp/codacy-coverage-15620814908828598437.json (145.15 kB)  - (ReportRules.scala:265)
```
as verified secrets (`15620814908828598437` in this case).

Btw, this is from off-the-shelf [codacy/codacy-coverage-reporter-action](https://github.com/codacy/codacy-coverage-reporter-action).

## Problem to be Addressed

Codacy detector should be more sensitive and should not be [using ](https://github.com/trufflesecurity/trufflehog/blob/c4ca7d7c8bd79eaec415a7c13fbd94966224189f/pkg/detectors/codacy/codacy.go#L50) `https://app.codacy.com/api/v3/version` as token verification endpoint, as this endpoint is unauthenticated and always returns 200.

## Description of the Preferred Solution

Perhaps we should add `/tmp/codacy-coverage` to the list of known FPs for this detector? This seems like it will be a common occurrence, coming from the official [codacy/codacy-coverage-reporter-action](https://github.com/codacy/codacy-coverage-reporter-action).

Additionally, FPs marked in this pattern consist of 20 digit-only characters, which seems like an outlier because typical Codacy tokens have various alphanumeric characters, not just digits. Perhaps there's improvement to be made there in the regex.

Also, we should use an authenticated endpoint for verification (I'm not sure which one, however).

dustin-decker:
Thank you, this should be fixed in #758 and should be released soon.

dinvlad:
Thanks! Would it be possible to make this detector more specific also?

PR #758, opened 2022-08-31
fix issue in codacy 

roxanne-tampus:
[](https://github.com/trufflesecurity/trufflehog/issues/724)

### DIFF
--- a/pkg/detectors/codacy/codacy_test.go
+++ b/pkg/detectors/codacy/codacy_test.go
@@ -1,6 +1,3 @@
-//go:build detectors
-// +build detectors
-
 package codacy
 
 import (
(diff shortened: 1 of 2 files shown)


pkg/detectors/codacy/codacy_test.go
```diff
@@ -1,6 +1,3 @@
-//go:build detectors
```
dustin-decker, line 1: Hey Roxanne, could you put these tags back?

You can run these tests by adding `-tags=detectors` with your `go test` command.

dustin-decker: Could you also update these docs with that? https://github.com/trufflesecurity/trufflehog/tree/main/hack/docs

dustin-decker (approved):

PR #762, opened 2022-08-31
added nubela detector

lonmarsDev:

### DIFF
--- a/pkg/detectors/nubela/nubela_test.go
+++ b/pkg/detectors/nubela/nubela_test.go
@@ -1,0 +1,117 @@
+
+package nubela
+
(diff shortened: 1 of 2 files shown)


pkg/detectors/nubela/nubela_test.go
```diff
@@ -0,0 +1,117 @@
+
```
roxanne-tampus, line 1: Please add the tags back. You can run the test by `go test ./pkg/detectors/nubela -tags=detectors`
```
//go:build detectors
// +build detectors
```

roxanne-tampus (approved):

PR #763, opened 2022-08-31
added infobip detector

lonmarsDev:


roxanne-tampus (approved):

pkg/detectors/infobip/infobip.go
```diff
@@ -0,0 +1,82 @@
+package infobip
```
roxanne-tampus, line 1: Please add the tags back. You can run the test by `go test ./pkg/detectors/infobip -tags=detectors`
```
//go:build detectors
// +build detectors
```

roxanne-tampus (approved):

Issue #704, opened 2022-08-09
include members and include-forks not working with org


rahuljaisinghani:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version
3.7.3

### Expected Behavior

command should scan organisation user's public repo and forks for credentials after passing --include members and include-forks flags

### Actual Behavior

command does not scan organisation user's public repo for credentials after passing --include members and include-forks flags
It only scan organisation repos

### Steps to Reproduce
Run trufflehog github --org=<org> --include-members --include-forks

## Environment
 * OS: Mac intel
 * VersionMojave

### References

* #0000

nyanshak:
In both `getReposByOrg` and `getReposByUser`, there is a check for fork status + fork flag status before adding a repo to the list to be scanned. I think this issue can be closed out now.

kashifkhan0771:
It looks like this has been resolved. @rahuljaisinghani feel free to open a new issue if you encounter any further problems.

PR #773, opened 2022-09-02
Fix include-members not working on github

trufflesteeeve:
This is a partial fix for https://github.com/trufflesecurity/trufflehog/issues/704 - `--include-members` did not have an effect in most cases. It doesn't change the `--include-forks`, which _should_ work, but might not?


dustin-decker:
Pagination for `addMembersByApp` appears to be getting stuck as well

dustin-decker:
Related to OC-94

dustin-decker (approved):

PR #777, opened 2022-09-02
Fix Honeycomb detector with older key format

MNThomson:
## Description
Add the older HC api key format. See #687 for discussion.
Expanding regex to `\b([0-9a-f]{32}|[0-9a-zA-Z]{22})\b`

image
[Regex visualization](https://www.debuggex.com/r/spDIDRvm-uLZzAqS)


dustin-decker (approved):
Thank you!

PR #778, opened 2022-09-02
fix regex pattern for confluent detector

### LINKED ISSUE #770: Confluent detector can be improved
### Community Note

### TruffleHog Version

v 3.10.0

### Trace Output

N/A

### Expected Behavior

1. The two regexes [here](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/confluent/confluent.go#L25-L26) are buggy : 
    a. There is an extra `-` before `0-9` in each of the regex. It is not needed.
    b.  The regex for detecting confluent API key secret [here](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/confluent/confluent.go#L26) should also have detection for `+` and `/` characters. I confirmed using regex101.com that a secret with `+` and '/' character would not be detected with current pattern. 

The correct regex pattern would be : 

```
keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"confluent"}) + `\b([a-zA-Z0-9]{16})\b`)
secretPat = regexp.MustCompile(detectors.PrefixRegex([]string{"confluent"}) + `\b([a-zA-Z0-9\+\/]{64})\b`)
```

~~2. Confluent has [different types of API keys](https://docs.confluent.io/cloud/current/access-~~management/authenticate/api-keys/api-keys.html). There is cloud API key , which provides control plane API access and then there are resource specific API keys. I think we can improve this detector by adding validations for all types of API keys.~~

3. The [API currently used](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/confluent/confluent.go#L63) to detect validity of secret may result in **false negative**,  depending on the role assigned to a user and/or the type of key. ~~We can use various resource specific and identity APIs to accurately detect Cloud and resource specific API keys.  See [this](https://docs.confluent.io/cloud/current/api.html)~~ . We can use the IAM read API key endpoint instead. See [this](https://docs.confluent.io/cloud/current/api.html#tag/API-Keys-(iamv2)/operation/getIamV2ApiKey)

### Actual Behavior

A confluent secret with `+` and '/' character was not detected with [current](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/confluent/confluent.go#L26) regex.

### Steps to Reproduce

 4. Scroll down to '....'
 5. See error

## Environment
 * OS: macOS
 * Version 12.5.1

## Additional Context

I work for Confluent Security team and would be happy to send a PR to improve Confluent detector. 

### References

* #0000

ap00rv:
Closes #770 

Go version : 1.18.0 
OS : macOS 12.5.1

### changes

- Fixed regex for detecting secret and key ID 
- changed validation endpoint
- At L75, changed parameter to secret value instead of key ID 

### Tests (failing) 

- Created `/tmp/.env` file with following content 

```
CONFLUENT_TOKEN=<valid_secret>
CONFLUENT_KEY=<valid_key_id>
CONFLUENT_INACTIVE=<invalid_secret>
```
- Ran `export TEST_SECRET_FILE="/tmp/.env" && go test ./pkg/detectors/confluent -tags=detectors`

- Tests fail with following message : 

```
--- FAIL: TestConfluent_FromChunk (0.93s)
    --- FAIL: TestConfluent_FromChunk/found,_verified (0.77s)
        confluent_test.go:107: Confluent.FromData() found, verified diff: (-got +want)
             [
              {
               DetectorType: 60,
               Verified: true,
               Raw: [
               ],
               RawV2: [
            -   86,
            -   89,
            -   67,
           ...
               ],
               Redacted: "",
               ExtraData: {
               },
               StructuredData: nil,
              },
             ]
```

### DIFF
--- a/pkg/detectors/confluent/confluent.go
+++ b/pkg/detectors/confluent/confluent.go
@@ -22,8 +22,8 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"confluent"}) + `\b([a-zA-Z-0-9]{16})\b`)
-	secretPat = regexp.MustCompile(detectors.PrefixRegex([]string{"confluent"}) + `\b([a-zA-Z-0-9]{64})\b`)
+	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"confluent"}) + `\b([a-zA-Z0-9]{16})\b`)
+	secretPat = regexp.MustCompile(detectors.PrefixRegex([]string{"confluent"}) + `\b([a-zA-Z0-9\+\/]{64})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
@@ -60,7 +60,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			if verify {
 				data := fmt.Sprintf("%s:%s", resMatch, resSecret)
 				sEnc := b64.StdEncoding.EncodeToString([]byte(data))
-				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.telemetry.confluent.cloud/v2/metrics/cloud/descriptors/resources", nil)
+				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.confluent.cloud/iam/v2/api-keys/"+resMatch, nil)
 				if err != nil {
 					continue
 				}
@@ -72,7 +72,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 						s1.Verified = true
 					} else {
 						// This function will check false positives for common test words, but also it will make sure the key appears 'random' enough to be a real key.
-						if detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
+						if detectors.IsKnownFalsePositive(resSecret, detectors.DefaultFalsePositives, true) {
 							continue
 						}
 					}
--- a/pkg/detectors/confluent/confluent.go
+++ b/pkg/detectors/confluent/confluent.go
@@ -54,7 +54,6 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_Confluent,
 				Raw:          []byte(resMatch),
-				RawV2:        []byte(resMatch + resSecret),
 			}
 
 			if verify {
--- a/pkg/detectors/confluent/confluent.go
+++ b/pkg/detectors/confluent/confluent.go
@@ -54,6 +54,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_Confluent,
 				Raw:          []byte(resMatch),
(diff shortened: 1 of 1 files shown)


ahrav:
Hi @ap00rv thanks so much for the contribution. It actually looks like your code is completely fine. The issue looks to be stemming from the `RawV2` field that was added to the `Result` proto message. When the test is doing a diff against the result and the expected value the `RawV2` field isn't being set. Since that field isn't actually something we need to test against for the detector you can actually do the comparison and ignore a field of a struct. A great library to use would be `go-cmp` (https://pkg.go.dev/github.com/google/go-cmp@v0.5.8/cmp) namely the `Diff` function. The last argument of the function can then take in `cmpopts.IgnoreFields` with the `RawV2` field. 

It could look something like:
```
ignoreOpts := cmpopts.IgnoreFields(detectors.Result{}, "RawV2")
if diff := cmp.Diff(tt.want, got, ignoreOpts); diff != "" {
    t.Errorf("Confluent.FromData() %s diff: (-got +want)\n%s", tt.name, diff)
}
```

Apologies about that, we will try to get the template sorted for detectors that use the `RawV2` field as part of the Result proto.

Otherwise LGTM ✅  on the PR, thanks again :)

ap00rv:
@ahrav  thanks for the quick feedback ! I removed `RawV2` field from [here](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/confluent/confluent.go#L57) and the test is working. Can this be an alternative solution or should I go ahead with the solution you suggested ?

ahrav:
> @ahrav thanks for the quick feedback ! I removed `RawV2` field from [here](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/confluent/confluent.go#L57) and the test is working. Can this be an alternative solution or should I go ahead with the solution you suggested ?

Actually, I think you can remove it like in your example. We had scenarios in the past where the KeyID for detectors could be the same for multiple accounts, so to disambiguate them we combined the KeyID and SecretID and then hashed those values. In this case I do not believe PATs would be the same, and therefore just using the Raw field should be sufficient.

Feel free to submit with that change, appreciate it.

ahrav:
@ap00rv I spent a little more time looking into this and come to find out we will actually need `RawV2` as part of the result struct. So your initial solution is actually correct even though the tests would currently fail.

I had mistakenly interpreted the `keyPat` as a personal access token, when in fact it's still a keyID "pattern". For that reason since the same KeyID can be used throughout the codebase with a potentially different secret. If we do **NOT** include the secret in the HashV2 we can run into the situation where the same KeyID is found with a different secret that is potentially inactive and would have the same hash value (since we only included the KeyID in the hash), thus invalidating the initial active credential.

Sorry for my mistake on the initial explanation and taking a while to get back to you on this. I was double checking some of the logic currently in place to make sure I didn't miss anything else. Once the RawV2 field is added back we should be good. 😄

ap00rv:
happy to make that change @ahrav  :) . What about tests ? should I also write the logic to ignore RawV2 field in tests ? Thanks !

ahrav:
> happy to make that change @ahrav :) . What about tests ? should I also write the logic to ignore RawV2 field in tests ? Thanks !

For now you don't need to worry about the tests w/ the new field. So long as the tests pass without that field you should be good. I'm gonna spend some time this week to get a solution in place so the templates for the new detectors will account for certain credentials that do require the RawV2 field.

That's why for now your initial submission is good to be merged.

ahrav (approved):

dustin-decker (approved):
Thank you for your contribution!

PR #783, opened 2022-09-07
[THOG-681] - Handle errors sources

ahrav:
- Update all sources Progress completion if there are no errors during scanning.

### DIFF
--- a/pkg/sources/gitlab/gitlab_test.go
+++ b/pkg/sources/gitlab/gitlab_test.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/kylelemons/godebug/pretty"
 	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/protobuf/types/known/anypb"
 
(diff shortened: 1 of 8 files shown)


pkg/sources/gitlab/gitlab_test.go
```diff
@@ -247,3 +249,41 @@ func Test_setProgressCompleteWithRepo_Progress(t *testing.T) {
...
+		wantErr      bool
+	}{
+		{
+			name:         "no repos",
+			wantComplete: true,
+		},
+		{
+			name:    "no repos",
+			repos:   []string{"a"},
+			wantErr: true,
```
mcastorina, line 267: **Nit (optional):** Maybe `name` should be `"one invalid repo"`? Invalid because it doesn't end in `.git`

mcastorina (approved):

dustin-decker (approved):

pkg/sources/gitlab/gitlab_test.go
ahrav: thanks for catching this. I updated all the other tests as well 🙇

PR #784, opened 2022-09-08
Improve GitHub debug logging

dustin-decker:
- Increases debug logging for github
- Frees resources sooner

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -483,7 +483,9 @@ func (s *Source) getReposByOrg(ctx context.Context, apiClient *github.Client, or
 	var numRepos, numForks int
 	for {
 		someRepos, res, err := apiClient.Repositories.ListByOrg(ctx, org, opts)
-		res.Body.Close()
+		if err == nil {
+			res.Body.Close()
+		}
 		if handled := handleRateLimit(err, res); handled {
 			continue
 		}
@@ -534,7 +536,9 @@ func (s *Source) getReposByUser(ctx context.Context, apiClient *github.Client, u
 	}
 	for {
 		someRepos, res, err := apiClient.Repositories.List(ctx, user, opts)
-		res.Body.Close()
+		if err == nil {
+			res.Body.Close()
+		}
 		if handled := handleRateLimit(err, res); handled {
 			continue
 		}
@@ -561,7 +565,9 @@ func (s *Source) getGistsByUser(ctx context.Context, apiClient *github.Client, u
 	gistOpts := &github.GistListOptions{}
 	for {
 		gists, res, err := apiClient.Gists.List(ctx, user, gistOpts)
-		res.Body.Close()
+		if err == nil {
+			res.Body.Close()
+		}
 		if handled := handleRateLimit(err, res); handled {
 			continue
 		}
@@ -610,7 +616,9 @@ func (s *Source) addMembersByApp(ctx context.Context, installationClient *github
 	for _, org := range installs {
 		for {
 			members, res, err := apiClient.Organizations.ListMembers(ctx, *org.Account.Login, optsOrg)
-			res.Body.Close()
+			if err == nil {
+				res.Body.Close()
+			}
 			if handled := handleRateLimit(err, res); handled {
 				continue
 			}
@@ -644,7 +652,9 @@ func (s *Source) addReposByApp(ctx context.Context, apiClient *github.Client) er
 	}
 	for {
 		someRepos, res, err := apiClient.Apps.ListRepos(ctx, opts)
-		res.Body.Close()
+		if err == nil {
+			res.Body.Close()
+		}
 		if handled := handleRateLimit(err, res); handled {
 			continue
 		}
@@ -680,7 +690,9 @@ func (s *Source) addAllVisibleOrgs(ctx context.Context, apiClient *github.Client
 	}
 	for {
 		orgs, resp, err := apiClient.Organizations.ListAll(ctx, orgOpts)
-		resp.Body.Close()
+		if err == nil {
+			resp.Body.Close()
+		}
 		if handled := handleRateLimit(err, resp); handled {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


pkg/sources/github/github.go
```diff
@@ -483,16 +483,14 @@ func (s *Source) getReposByOrg(ctx context.Context, apiClient *github.Client, or
 	var numRepos, numForks int
 	for {
 		someRepos, res, err := apiClient.Repositories.ListByOrg(ctx, org, opts)
-		if err == nil {
-			defer res.Body.Close()
-		}
+		res.Body.Close()
```
ahrav, line 486: suggestion: We might want to check err != nil here first because if err != nil, res will be nil and will panic.

pkg/sources/github/github.go
```diff
@@ -536,15 +534,14 @@ func (s *Source) getReposByUser(ctx context.Context, apiClient *github.Client, u
 	}
 	for {
 		someRepos, res, err := apiClient.Repositories.List(ctx, user, opts)
-		if err == nil {
-			defer res.Body.Close()
-		}
+		res.Body.Close()
```
ahrav, line 537: ditto

pkg/sources/github/github.go
```diff
@@ -575,24 +572,23 @@ func (s *Source) getGistsByUser(ctx context.Context, apiClient *github.Client, u
 	gistURLs := []string{}
 	gistOpts := &github.GistListOptions{}
 	for {
-		gists, resp, err := apiClient.Gists.List(ctx, user, gistOpts)
-		if err == nil {
-			defer resp.Body.Close()
-		}
-		if handled := handleRateLimit(err, resp); handled {
+		gists, res, err := apiClient.Gists.List(ctx, user, gistOpts)
+		res.Body.Close()
```
ahrav, line 576: ditto

pkg/sources/github/github.go
```diff
@@ -575,24 +572,23 @@ func (s *Source) getGistsByUser(ctx context.Context, apiClient *github.Client, u
...
+		if handled := handleRateLimit(err, res); handled {
 			continue
 		}
 		if err != nil {
 			log.WithError(err).Warnf("could not list repos for user %s", user)
 			return nil, fmt.Errorf("could not list repos for user %s: %w", user, err)
 		}
+		s.log.Debugf("Listed gists for user %s page %d/%d", user, gistOpts.Page, res.LastPage)
```
ahrav, line 584: linter: Pointing out you might have a nil pointer dereference here, might need to check that res != nil first.

pkg/sources/github/github.go
```diff
@@ -575,24 +572,23 @@ func (s *Source) getGistsByUser(ctx context.Context, apiClient *github.Client, u
...
 		for _, gist := range gists {
 			gistURLs = append(gistURLs, gist.GetGitPullURL())
 		}
-		if resp == nil || resp.NextPage == 0 {
+		if res == nil || res.NextPage == 0 {
```
ahrav, line 588: linter: pointing out the pointer can be nil 👀

pkg/sources/github/github.go
```diff
@@ -626,9 +622,7 @@ func (s *Source) addMembersByApp(ctx context.Context, installationClient *github
 	for _, org := range installs {
 		for {
 			members, res, err := apiClient.Organizations.ListMembers(ctx, *org.Account.Login, optsOrg)
-			if err == nil {
-				defer res.Body.Close()
-			}
+			res.Body.Close()
```
ahrav, line 625: ditto, res can potentially be nil here as well?

pkg/sources/github/github.go
```diff
@@ -661,20 +656,20 @@ func (s *Source) addReposByApp(ctx context.Context, apiClient *github.Client) er
 	}
 	for {
 		someRepos, res, err := apiClient.Apps.ListRepos(ctx, opts)
-		if err == nil {
-			defer res.Body.Close()
-		}
+		res.Body.Close()
```
ahrav, line 659: ditto

pkg/sources/github/github.go
```diff
@@ -697,9 +692,7 @@ func (s *Source) addAllVisibleOrgs(ctx context.Context, apiClient *github.Client
 	}
 	for {
 		orgs, resp, err := apiClient.Organizations.ListAll(ctx, orgOpts)
-		if err == nil {
-			defer resp.Body.Close()
-		}
+		resp.Body.Close()
```
ahrav, line 695: ditto

pkg/sources/github/github.go
```diff
@@ -735,16 +728,15 @@ func (s *Source) addOrgsByUser(ctx context.Context, apiClient *github.Client, us
 	}
 	for {
 		orgs, resp, err := apiClient.Organizations.List(ctx, "", orgOpts)
-		if err == nil {
-			defer resp.Body.Close()
-		}
+		resp.Body.Close()
```
ahrav, line 731: last one 😺

pkg/sources/github/github.go
```diff
@@ -483,16 +483,14 @@ func (s *Source) getReposByOrg(ctx context.Context, apiClient *github.Client, or
...
-		}
+		res.Body.Close()
 		if handled := handleRateLimit(err, res); handled {
 			continue
 		}
 		if err != nil {
 			return nil, fmt.Errorf("could not list repos for org %s: %w", org, err)
 		}
-		log.Debugf("listed repos page %d/%d", opts.Page, res.LastPage)
+		s.log.Debugf("Listed repos for org %s page %d/%d", org, opts.Page, res.LastPage)
```
ahrav, line 493: question: Not sure if the check for `err != nil` means res != nil, because if err is not nil but `res` is still nil `res.LastPage` will panic.

NOTE: this applies to the debugs below as well.

ahrav (approved):

pkg/sources/github/github.go
dustin-decker: You're totally right, not sure what I was thinking. Reverted to only close if err is nil like before, but not deferring the close

pkg/sources/github/github.go
dustin-decker: Oh, I see there is a nil check, i'll move it down.

ahrav:
> Let's merge #783 first

This should be in.

PR #792, opened 2022-09-10
[THOG-162] Implement JDBC verification for select drivers

mcastorina:
Also includes integration tests for postgres and mysql via docker. To run, execute the following (untested what will happen if the docker images aren't installed):

go test -tags=detectors,integration ./pkg/detectors/jdbc


pkg/detectors/jdbc/jdbc.go
```diff
@@ -15,7 +18,7 @@ type Scanner struct{}
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`(?i)jdbc:[\w]{3,10}:\/\/\w[\s\S]{0,512}?password[=: \"']+(?P<pass>[^<{($]*?)[ \s'\"]+`)
+	keyPat = regexp.MustCompile(`(?i)jdbc:[^\s"']+`)
```
mcastorina, line 21: This might be a bit too permissive a pattern for unverified results.. I'll revisit this.

pkg/detectors/jdbc/jdbc.go
```diff
@@ -30,38 +33,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-		//		}
-		//	}
-		//}
+		if verify {
+			s.Verified = false
+			j, err := newJDBC(jdbcConn)
+			if err != nil {
+				continue
+			}
+			s.Verified = j.ping()
```
dustin-decker, line 50: Don't we need a timeout here?

dustin-decker (approved):
We may need some timeout considerations but otherwise lgtm

pkg/detectors/jdbc/jdbc.go
mcastorina: Yeah, the timeout is hidden in each implementation of the `ping()` method, usually by calling this helper function:

```go
func pingErr(driverName, conn string) error {
	db, err := sql.Open(driverName, conn)
	if err != nil {
		return err
	}
	defer db.Close()

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second) // <---------
	defer cancel()
	if err := db.PingContext(ctx); err != nil {
		return err
	}
	return nil
}
```

I'll surface it, good callout.

dustin-decker (approved):

mcastorina:
It's already [there](https://github.com/trufflesecurity/trufflehog/blob/b02cf7e032422d514d010d44afab8d4c02269d0c/pkg/engine/defaults.go#L754), this PR is just adding verification for it.

PR #793, opened 2022-09-11
Add support for MongoDB detector.

ahrav:
- Support for Standalone connection strings to MongoDB.

Regex used: https://regex101.com/r/eTsgOz/1

NOTE: TODO Add support for replica set, sharded db, and Atlas Deployment.


pkg/detectors/mongodb/mongodb.go
```diff
@@ -0,0 +1,71 @@
...
+				if err != nil {
+					return
+				}
+				defer func() {
+					if err := client.Disconnect(ctx); err != nil {
+						return
+					}
+				}()
+				if err := client.Ping(ctx, readpref.Primary()); err != nil {
+					return
```
dustin-decker, line 62: Does this mean that we would not have unverified results?

dustin-decker: Oh nevermind, I see it's in an anonymous func

dustin-decker (approved):

PR #801, opened 2022-09-16
Use recover to catch panic in dep for old rars

bill-rich:
Some old rars can cause a crash in rardecoder. Catching the panic since I can not find an example that reproduces the panic. An issue is open in rardecode to track https://github.com/nwaples/rardecode/issues/30


KidiIT (commented):
Approved for the same price and the other one was for a different brand and the price was the only difference that we could get from it so we could have it delivered on the day that you guys would like to get the same thing and then you can get the exact amount you need it for your house.✔️🫶

PR #804, opened 2022-09-16
Include apiClient in Github source

bill-rich:
Include `apiClient` in Github Source to make function signatures smaller and improve readability. Will also be needed to make getting repo public/private status easier.


mcastorina (approved):
Nice refactor!

PR #806, opened 2022-09-17
[THOG-712] - Monotonically increasing progress counter

ahrav:
Initial implementation to make progress increment monotically. The main goal i'm trying to accomplish here is separating out progress completion from source. With our current implementation the source updates the progress, but that tends to happen in goroutines and using a loop variable. This can lead to buggy situations and incorrect progress completion values while scanning. 

- Add Counter struct to Progress struct that monotonically increases. This will keep track of each scan for each source. This will also be safe to use in goroutines.
- Refactor SetProgressComplete into two separate methods: `Update` and `Complete`. Update works on updating progress during a scan and Complete gets called at the end of the scan. (idk if this is too much refactoring, it just seemed a little more obvious to me than explicitly passing in the len of repos or buckets etc. This abstracts that away and just uses the total count from the counter and sets the progress explicitly)
- Include multiple fields in the counter to keep track of successful scan counts and total scan counts. This is so we can correctly keep track of both the percent complete (using the successCnt) and also update the progress in the logs when concurrently scanning.
- TotalCnt and SuccessCnt should have the same end value if all scans succeed. They only differ during scans to prevent multiple scans from having the same counter value prior to successfully finishing. Once a scan successfully finishes scanning the SuccessCnt is also incremented. 
- Both TotalCnt and SuccessCnt and safe for concurrent use.


mcastorina:
Overall I think it looks good. A few thoughts:

* There's possible confusion with the function names `IncSuccess` and `Complete`
* We need to decide what percent complete means -- successful completions? Work completed? I can see arguments to both, though I think "work completed" is slightly better (mostly because we're using this to indicate how many items we've gone through)

pkg/sources/gitlab/gitlab.go
```diff
@@ -315,28 +322,30 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk)
...
+		})
 	}
-	wg.Wait()
-	if len(errs) == 0 {
-		s.SetProgressComplete(len(s.repos), len(s.repos), fmt.Sprintf("Completed scanning source %s", s.name), "")
+
+	_ = s.jobPool.Wait()
+	if len(scanErrs) == 0 {
+		s.Complete(fmt.Sprintf("Completed scanning source %s", s.name))
 	}
```
mcastorina, line 346: **Question:** Should a scan be considered complete even if we encountered errors? What would this look like to the customer?

pkg/sources/sources.go
```diff
@@ -104,39 +104,81 @@ func NewConfig(opts ...func(*Config)) Config {
...
+	mu                   sync.Mutex
+}
+
+// IncTotal increments the total field of counter and returns the new value.
+func (c *Counter) IncTotal() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.totalCnt++
+	return int(c.totalCnt)
+}
```
mcastorina, line 124: **Suggestion (optional):** I think the `atomic` package can be used here, particularly [AddUint32](https://pkg.go.dev/sync/atomic#AddUint32) which atomically adds to the counter and returns the new value.

I marked it at optional because I don't think it matters too much in this case to have a lock free solution.

pkg/sources/sources.go
```diff
@@ -104,39 +104,81 @@ func NewConfig(opts ...func(*Config)) Config {
 	return *c
 }
 
+// Counter is a simple monotonically-increasing counter for tracking progress.
+// It includes a successCnt that records the number of successful scans.
+// It also includes a totalCnt that records the total number of scans.
+// This is necessary in order to display the number of concurrent scans
+// regardless of whether they are successful or not.
+// successCnt is the only field used when calculating the percentage complete.
```
mcastorina, line 112: **Question:** Same question about calculating percentage.. I think we should increase the percent even if it failed to indicate we are progressing.

pkg/sources/gitlab/gitlab.go
ahrav: For this I would've thought we only consider it complete if len(scanErrs) is 0, which would mean no errors i think?

pkg/sources/sources.go
ahrav: I started with that as well, but thought this slighlty more verbose way was a lot easier to read.

pkg/sources/sources.go
ahrav: Would you mind clarifying this statement? I'm not sure I follow.

ahrav:
> Overall I think it looks good. A few thoughts:
> 
> * There's possible confusion with the function names `IncSuccess` and `Complete`
> * We need to decide what percent complete means -- successful completions? Work completed? I can see arguments to both, though I think "work completed" is slightly better (mostly because we're using this to indicate how many items we've gone through)

Yea so I just had it as Complete since it's a method on `Progress` so complete means we are done progressing. I guess then we need to clarify progress a bit more? It sounds and reads a bit weird to have progress.WorkComplete? It's almost like another noun suddenly gets injected into the meaning.

If everyone things that seems better i'm more than happy to go with that approach.

pkg/sources/sources.go
ahrav: As maybe a follow up question for you. in the `Update` method we need to get `successCnt` from the `Counter` would atomic also work in that case? I'm actually not super familiar with using atomic for anything other than incrementing a value

pkg/sources/sources.go
mcastorina: Sorry for poor wording.. I edited it with some commas. But I was referring to this:

> successCnt is the only field used when calculating the percentage complete

If we are scanning 10 repos and one of them fails, I think the percent complete should be 100%, not 90%.

ahrav: Oh interesting. I was under the impression percent complete can only be 100% if there are no failures.

ahrav: Because i guess that way in the UI it doesn't look like a success?

ahrav: Maybe i confused healthy w/ complete

mcastorina: Hm, yeah. That's the thread that spawned the idea of "scan reports" which will more accurately capture which items were successfully scanned.

In the mean time, I'm not sure which we should go with.. I still think from a user perspective I would want to see the progress increase even if a repo failed to scan, but I agree that the entire scan shouldn't be considered healthy in that case.

ahrav: Hmm so I think maybe we need to wait until we have scan reports because I'm not so sure this PR helps. I think we still end at the same place of 99/100 repos fail to scan we still would get 100% scan completed progress which is probably less helpful to the user.

pkg/sources/sources.go
```diff
@@ -104,39 +105,75 @@ func NewConfig(opts ...func(*Config)) Config {
...
+// IncSuccess increments the success field of counter and returns the new value.
+func (c *Counter) IncSuccess() int {
+	return int(atomic.AddUint32(&c.successCnt, 1))
+}
+
+// Get returns the current successCnt.
+func (c *Counter) Get() uint32 {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.successCnt
```
mcastorina, line 133: **Nit (optional):** There's also a [LoadUint32](https://pkg.go.dev/sync/atomic#LoadUint32) function if you want to get rid of the mutex.

mcastorina:
Maybe we should get another opinion on how to calculate percent complete before merging.

pkg/sources/git/git.go
```diff
@@ -189,8 +189,14 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
 		return errors.New("invalid connection type for git source")
 	}
 
-	for i, u := range s.conn.Directories {
-		s.SetProgressComplete(i, len(s.conn.Repositories), fmt.Sprintf("Repo: %s", u), "")
+	for _, u := range s.conn.Directories {
+		_ = s.Counter.IncTotal()
```
mcastorina, line 193: **Nit (optional):** Unless this causes a lint error, I think we can remove the `_ =` here. Maybe it's just me, but it looks like we're ignoring an error instead of ignoring the new value.

pkg/sources/git/git.go
```diff
@@ -189,8 +189,14 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
...
+		go func() {
+			defer func() {
+				s.Counter.IncSuccess()
+			}()
```
mcastorina, line 197: **Nit (optional):** `defer s.Counter.IncSuccess()` should be equivalent

pkg/sources/github/github.go
```diff
@@ -470,31 +470,36 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
 			if common.IsDone(ctx) {
 				return nil
 			}
+			cnt := s.Counter.IncTotal()
 
-			// TODO: set progress complete is being called concurrently with i
-			s.setProgressCompleteWithRepo(i, progressIndexOffset, repoURL)
+			defer func() {
+				s.Counter.IncSuccess()
+			}()
```
mcastorina, line 484: **Nit (optional):** `defer s.Counter.IncSuccess()`

pkg/sources/github/github.go
```diff
@@ -470,31 +470,36 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
 				s.resumeInfoSlice = sources.RemoveRepoFromResumeInfo(s.resumeInfoSlice, repoURL)
 			}(s, repoURL)
 
+			var err error
 			if !strings.HasSuffix(repoURL, ".git") {
-				scanErrs = append(scanErrs, fmt.Errorf("repo %s does not end in .git", repoURL))
+				err = fmt.Errorf("repo %s does not end in .git", repoURL)
+				scanErrs = append(scanErrs, err)
 				return nil
 			}
```
mcastorina, line 498: **Nit (optional):** The assignment to `err` here isn't needed since we `return nil`.

pkg/sources/gitlab/gitlab.go
```diff
@@ -265,34 +270,33 @@ func (s *Source) getRepos() ([]string, []error) {
...
-				return
+				return nil
 			}
 
-			s.setProgressCompleteWithRepo(i, progressIndexOffset, repoURL)
-			// Ensure the repo is removed from the resume info after being scanned.
+			cnt := s.Counter.IncTotal()
+			defer func() {
+				s.Counter.IncSuccess()
+			}()
```
mcastorina, line 297: **Nit (optional):** `defer s.Counter.IncSuccess()`

pkg/sources/gitlab/gitlab.go
```diff
@@ -315,28 +319,28 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk)
 			}
 			defer os.RemoveAll(path)
 			if err != nil {
-				errsMut.Lock()
-				errs = append(errs, err)
-				errsMut.Unlock()
-				return
+				scanErrs = append(scanErrs, err)
+				return nil
 			}
```
mcastorina, line 324: I think we might still need to lock here if we are concurrently appending to `scanErrs`.

mcastorina: Random thought - a named return parameter / deferred function might work well here:

```go

s.jobPool.Go(func() (err error) {
    defer func() {
        if err != nil {
            // lock and append to scanErrs
         }
         // always set err to nil to prevent marking as error
         err = nil
    }()
})

**Review by mcastorina — COMMENTED** [review]
Overall looks good, just had one concern about locks

File: pkg/sources/git/git.go [hunk 11]
```diff
@@ -189,8 +189,12 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
 		return errors.New("invalid connection type for git source")
 	}
 
-	for i, u := range s.conn.Directories {
-		s.SetProgressComplete(i, len(s.conn.Repositories), fmt.Sprintf("Repo: %s", u), "")
+	for _, u := range s.conn.Directories {
+		s.Counter.IncTotal()
+		go func() {
+			defer s.Counter.IncSuccess()
+		}()
```
mcastorina, line 196: I think the `defer s.Counter.IncSuccess()` should probably go at the end of the for loop?

pkg/sources/gitlab/gitlab.go
```diff
@@ -319,28 +322,30 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk)
...
 			defer os.RemoveAll(path)
 			if err != nil {
-				errsMut.Lock()
-				errs = append(errs, err)
-				errsMut.Unlock()
-				return
+				scanErrs = append(scanErrs, err)
+				return nil
```
mcastorina, line 326: This'll need the mutex too :/

pkg/sources/git/git.go
ahrav: You mean not use the defer and just do s.Counter.IncSuccess()?

mcastorina: Yeah, or wrap the entire loop body in a function to defer.

Since we have a defer in a go func here I think it'll just immediately execute currently.

PR #808, opened 2022-09-19
Browserstack regex

rahuljaisinghani:
- The old regex do not consider all the types of keys. The new regex covers all the old and new format of browserstack keys
- wrt documentation and sample scripts, added more keywords for improving detection


ahrav (approved):
LGTM 👍 
Thanks a lot for updating this. We really appreciate it.

PR #810, opened 2022-09-19
[THOG-709] - Recover from detector panics

ahrav:


pkg/engine/engine.go
```diff
@@ -163,7 +163,12 @@ func (e *Engine) detectorWorker(ctx context.Context) {
 	for chunk := range e.chunks {
 		fragStart, mdLine := fragmentFirstLine(chunk)
 		for _, decoder := range e.decoders {
-			decoded := decoder.FromChunk(chunk)
+			decoded := func(ctx context.Context) *sources.Chunk {
+				defer func(ctx context.Context) {
+					common.Recover(ctx)
```
dustin-decker, line 168: This function calls os.Exit(1), maybe we should update it to optionally exit

mcastorina: Maybe renaming `Recover` to `RecoverAndExit` or something would make it more clear, then have `Recover` not exit.

dustin-decker: Yes, let's do that

dustin-decker: Also the integration tests are fixed in main, please update your branch

mcastorina (approved):

dustin-decker (approved):

PR #817, opened 2022-09-22
[THOG-709] - Fix recover

ahrav:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -163,12 +163,7 @@ func (e *Engine) detectorWorker(ctx context.Context) {
 	for chunk := range e.chunks {
 		fragStart, mdLine := fragmentFirstLine(chunk)
 		for _, decoder := range e.decoders {
-			decoded := func(ctx context.Context) *sources.Chunk {
-				defer func(ctx context.Context) {
-					common.Recover(ctx)
-				}(ctx)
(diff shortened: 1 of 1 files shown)


pkg/engine/engine.go
```diff
@@ -186,16 +181,21 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
 						continue
 					}
-					ctx, cancel := context.WithTimeout(ctx, time.Second*10)
-					defer cancel()
-					results, err := detector.FromData(ctx, verify, decoded.Data)
+
+					results, err := func() ([]detectors.Result, error) {
+						ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+						defer cancel()
+						defer common.Recover(ctx)
```
mcastorina, line 188: Do you know if the order matters?

mcastorina (approved):

pkg/engine/engine.go
ahrav: Yea i think we want recover, then cancel. So with defer it's LIFO so I think this works.

ahrav: hm... or maybe we want that flipped.

ahrav: OKay yep cancel then recover looks correct.

PR #818, opened 2022-09-22
Move head and base normalization to source

bill-rich:
Prevents a bad head or base being used in git source.

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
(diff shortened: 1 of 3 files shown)


pkg/sources/git/git.go
```diff
@@ -453,6 +458,55 @@ func (s *Git) ScanRepo(ctx context.Context, repo *git.Repository, repoPath strin
 	return nil
 }
 
+func normalizeConfig(scanOptions *ScanOptions, repo *git.Repository) (err error) {
```
mcastorina, line 461: **Suggestion (optional):** We may be able to minimize our if statements by starting the function with something like:

```go
if scanOptions.BaseHash == "" || scanOptions.HeadHash == "" {
    // Nothing to do.
    return nil
}

// ...
```

mcastorina (approved):

PR #819, opened 2022-09-24
[THOG-729] - Add support for file and stdin scans.

ahrav:
Screen Shot 2022-09-24 at 5 16 44 PM


main.go
```diff
@@ -76,6 +76,9 @@ var (
 	// filesystemScanIncludePaths = filesystemScan.Flag("include-paths", "Path to file with newline separated regexes for files to include in scan.").Short('i').String()
 	// filesystemScanExcludePaths = filesystemScan.Flag("exclude-paths", "Path to file with newline separated regexes for files to exclude in scan.").Short('x').String()
 
+	fileScan     = cli.Command("file", "Scan a file (defaults to standard in)")
+	fileScanPath = fileScan.Flag("path", "Path of the file to scan").Default(os.Stdin.Name()).String()
```
mcastorina, line 80: **Thought:** It would make more sense to me to put file scanning under the `filesystem` subcommand. In general I think the command should look like `trufflehog filesystem FILE_OR_DIR [FILE_OR_DIR...]` but I know we have the `--directory` flag. Maybe we can start with a repeatable `--file` flag?

main.go
```diff
@@ -238,6 +241,14 @@ func run(state overseer.State) {
 		if err = e.ScanFileSystem(ctx, sources.NewConfig(fs)); err != nil {
 			logrus.WithError(err).Fatal("Failed to scan filesystem")
 		}
+	case fileScan.FullCommand():
+		fs := func(c *sources.Config) {
+			c.FilePath = *fileScanPath
+		}
```
mcastorina, line 247: **Question:** Does this mean we will only scan one file at a time?

pkg/common/chunker.go
```diff
@@ -13,6 +13,8 @@ const (
 	PeekSize  = 3 * 1024
 )
 
+// ChunkReader is a wrapper around an io.Reader that chunks the data into smaller
+// chunks of size ChunkSize. It returns a channel of []byte that can be read from.
```
mcastorina, line 17: Good comment, thanks for adding that!

pkg/sources/file/file.go
```diff
@@ -0,0 +1,151 @@
...
+	var conn sourcespb.File
+	if err := anypb.UnmarshalTo(connection, &conn, proto.UnmarshalOptions{}); err != nil {
+		return fmt.Errorf("failed to unmarshal connection: %w", err)
+	}
+
+	s.path = conn.Path
+	// Set the path to stdin if one is not provided.
+	if s.path == "" {
+		s.path = os.Stdin.Name()
+	}
```
mcastorina, line 69: **Thoughts:** This can be unexpected and the user may think the scan is hanging when it's just waiting for input. I personally prefer using `-` as an indicator for stdin, but either way I'd like the user to intentionally request stdin scanning somehow.

Another way to handle it is to check if stdin is a tty or not. Usage wise, that would look like:

```bash
# detects stdin is not a tty and scans stdin
echo 'hello' | trufflehog filesystem

# scans stdin regardless of tty
trufflehog filesystem -
```

pkg/sources/file/file.go
```diff
@@ -0,0 +1,151 @@
...
+	s.fileName = filepath.Base(s.path)
+
+	fi, err := os.Stat(s.path)
+	if err != nil {
+		return fmt.Errorf("failed to stat file: %w", err)
+	}
+	if !fi.Mode().IsRegular() {
+		return fmt.Errorf("file is not a regular file: %v", s.path)
+	}
```
mcastorina, line 78: **Suggestion:** Maybe check if the file is a directory here instead? Otherwise I think this check prevents scanning symlinks, which I could see a legitimate use for.

pkg/sources/file/file.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to open file: %w", err)
+	}
+	defer func(file *os.File) {
+		if err := file.Close(); err != nil {
+			s.log.WithError(err).Errorf("Failed to close file: %v.", s.path)
+		}
+	}(file)
+
+	reReader, err := diskbufferreader.New(file)
```
mcastorina, line 96: **Question:** Will this create a temp file? Also, why do we need to reread the data?

proto/source_metadata.proto
```diff
@@ -79,6 +79,11 @@ message Filesystem {
   string email = 3;
 }
 
+message File {
+  string name = 1;
```
mcastorina, line 83: **Question:** What is name in this context?

ahrav: file name

main.go
ahrav: Yep, I was adding the same functionality as enterprise. Ideally we use only one implementation and then that can handle multiple files, but for now I felt for MVP it was sufficient.

pkg/sources/file/file.go
ahrav: Yes, I believe this is how we scan files for our other source so I was using our existing implementation.

main.go
dustin-decker: I don't think it necessarily needs to match enterprise (the CLIs are already pretty divergent and have opportunity to improve), but what about stdin in this case? Maybe File + Directory are covered under Filesystem, and Stdin becomes its own subcommand?

Probably a separate discussion, but if we were to redo this, what would the preferred terminology be that is more technically accurate and human friendly?

Obak3:
Heyo! Do you guys potentially need any help in this? My golang isn't fantastic, but what else needs to be done?

PR #829, opened 2022-09-28
Added digitaloceanv2 detector

ladybug0125:

### DIFF
--- a/pkg/detectors/digitaloceanv2/digitaloceanv2.go
+++ b/pkg/detectors/digitaloceanv2/digitaloceanv2.go
@@ -1,0 +1,74 @@
+package digitaloceanv2
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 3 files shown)


pkg/detectors/digitaloceanv2/digitaloceanv2.go
```diff
@@ -0,0 +1,74 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(dop\_v1\_[a-f0-9]{64})\b`)
```
dustin-decker, line 24: We'll need the other two prefixes supported here as well. You can use a regex OR to do that. They'll also need to be added into the keywords

dustin-decker: Oh, nevermind, we’ll need to add separate verifiers for those types in later

dustin-decker (approved):

PR #832, opened 2022-09-30
Digitaloceanv2 detector

ladybug0125:


pkg/detectors/digitaloceanv2/digitaloceanv2.go
```diff
@@ -21,13 +21,13 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(dop\_v1\_[a-f0-9]{64})\b`)
+	keyPat = regexp.MustCompile(`\b(dop\_v1\_[a-f0-9]{64} | doo\_v1\_[a-f0-9]{64} | dor\_v1\_[a-f0-9]{64})\b`)
```
ahrav, line 24: suggestion: Since the only differences in these regex are the first 3 characters we can just or that portion. The rest will remain the same. Maybe something like: `\b(?:(dop|doo|dor)_v1_[a-f0-9]{64})\b`. Additionally `_` does not need to be escaped so we can omit that.

ahrav (approved):

pkg/detectors/digitaloceanv2/digitaloceanv2.go
ladybug0125: I modified the regex based on the suggestion.
Thank you.

pkg/detectors/digitaloceanv2/digitaloceanv2.go
```diff
@@ -54,15 +56,39 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 			if err == nil {
-				defer res.Body.Close()
-				if res.StatusCode >= 200 && res.StatusCode < 300 {
-					s1.Verified = true
-				} else {
-					// This function will check false positives for common test words, but also it will make sure the key appears 'random' enough to be a real key.
-					if detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
+				switch {
+				case strings.HasPrefix(resMatch, "dor_v1_"):
+					bodyBytes, err := io.ReadAll(res.Body)
```
dustin-decker, line 63: This requires a different request: https://docs.digitalocean.com/reference/api/oauth-api/#refresh-token-flow

dustin-decker (approved):

pkg/detectors/digitaloceanv2/digitaloceanv2.go
```diff
@@ -41,28 +42,60 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			continue
 		}
 		resMatch := strings.TrimSpace(match[1])
+		println("resmatch: " + resMatch)
```
dustin-decker, line 45: Please remove this and it should be ready to go

dustin-decker (approved):

PR #835, opened 2022-10-05
[THOG-746] - Add decoder type to results.

ahrav:
-Include the decoder type in the secret results.
Screen Shot 2022-10-05 at 4 35 20 PM


dustin-decker (approved):
Great! We'll also want to include the archive handler even though it's not a decoder package

ahrav:
> Great! We'll also want to include the archive handler even though it's not a decoder package

Ahh good point.

ahrav:
As a note, we will add archive decoding info on the file level. So I will leave it out of this PR.

PR #841, opened 2022-10-11
added npm detector

ankushgoel27:
Added NPM detector. Only using regex for the old format. NPM introduced a new format which is NPM_[0-9a-zA-z]{36}. Not sure how to add both regexes to the same npmtoken.go file


ahrav:
> 

@ankushgoel27 thanks for adding this detector, really appreciate it. As for the approach I would take in this case, I would probably include a second `keyPat` regex that covered the new format. Initially I was thinking maybe we could use a single regex leveraging maybe `|` if they were similar enough something similar to (https://github.com/trufflesecurity/trufflehog/blob/85467538f6e62595a5c6905e223740d520583538/pkg/detectors/digitaloceanv2/digitaloceanv2.go#L25) (not quite the same). That might not be the best way because the regexes look rather different to me.

I'm also sure there is a way to write a regex to cover both cases, but at that point reading the regex might be more cumbersome that writing two.

ankushgoel27:
They are very dissimilar. I tried creating one regex for both but its more complicated. how do i include a second keypat? sorry, not so very well versed in golang. maybe you can help :)

the new format regex is `npm_[0-9a-zA-Z]{36}`

ankushgoel27:
ok, figured it out. Will send a pull request soon.

ankushgoel27:
Hi,

during my testing, i had a file with multiple valid old format tokens and new format tokens. But trufflehog was detecting only 1 of each. it should detect all tokens present in the file.

ankushgoel27:
Fixed naming convention and detection of multiple tokens. Ready to be merged

ahrav (approved):
LGTM. Thanks a lot for adding these.

PR #845, opened 2022-10-12
Buffer commit log processing

mcastorina:
Some very large commits take a lot of time to process, which we can make progress on while we are scanning the contents of other commits.

Tested with `duckduckgo/tracker-radar`:
* Without buffered channel: 200 minutes
* With buffered channel: 110 minutes
* Speedup: 1.8x

Note: `git log` for the repo took around ~60 minutes which is the theoretical fastest we could scan the repo (via that dependency).


trufflesteeeve (approved):
Really cool find! LGTM!

PR #852, opened 2022-10-18
[THOG-767] - Ignore Bitbucket and Gitlab repos

ahrav:
- update protos to allow for ignoring repos during a scan.


proto/sources.proto
```diff
@@ -69,6 +69,7 @@ message Bitbucket {
 
   }
   repeated string repositories = 5;
+  repeated string ignoreRepos = 6;
```
mcastorina, line 72: I always forget, should protos be `camelCase` or `snake_case`?

We kind of have both but I feel like I'm seeing more snake case.

mcastorina (approved):

PR #855, opened 2022-10-20
Check rate limit when getting github user

trufflesteeeve:
Also, don't fetch a github user or their token when both are known. This currently only affects the Github Token auth type. Github App installations will continually fetch tokens every time we clone a repo. In the future we should check the `ExpiresAt` field of the Github App token and determine if we need to fetch a new one at that point.

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -43,11 +43,12 @@ const (
 )
 
 type Source struct {
-	name     string
-	token    string
-	sourceID int64
-	jobID    int64
-	verify   bool
+	name        string
+	githubUser  string
+	githubToken string
+	sourceID    int64
+	jobID       int64
+	verify      bool
 	repos,
 	orgs,
 	members,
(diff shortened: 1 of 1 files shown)


pkg/sources/github/github.go
```diff
@@ -544,13 +567,27 @@ func (s *Source) cloneRepo(ctx context.Context, repoURL string, installationClie
 		if err != nil {
 			return "", nil, fmt.Errorf("error cloning repo %s: %w", repoURL, err)
 		}
-	default:
-		var token string
-		user, token, err := s.UserAndToken(ctx, installationClient)
+
+	case *sourcespb.GitHub_GithubApp:
```
trufflesteeeve, line 571: This should be functionally equivalent to what was here before. The only change is in the Github_Token section below, which will skip fetching a user and token if the user and token are both known.

bill-rich: Can we do the same thing for `GithubApp` too or does it only make sense for a `Token`?

bill-rich (approved):

pkg/sources/github/github.go
trufflesteeeve: We can eventually. And actually, the Github App token has an `ExpiresAt` field, which will make that decision relatively easy. But for now I'm just doing this for Github Token, because we can't currently refresh that token anyway, so once we have it, we never need to fetch it again.

Issue #844, opened 2022-10-12
Trufflehog only lists one token per detection if multiple tokens are present


ankushgoel27:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version
Trufflehog dev

### Expected Behavior
For example, if multiple heroku tokens are present(all different), it should list all tokens and not just one
### Current Behaviour
Currently, it only list one token per detection and not all the similar type of tokens present.

# this is not about listing same token multiple times but detecting all tokens of a particular detector. ( Github tokens works as expected)

I believe its related to the end of the detector file where "return detector.CleanResults(results), nil" is present. I changed it to just return like how ghp works and it works fine.

### Steps to Reproduce

1. get multiple tokens for a service. Example - heroku API token. 
2. add those tokens to a file and run trufflehog
3. it only lists one token out of the file.
4. Now replace the "return detectors.CleanResults(results),nil" with just "return" and recompile the trufflehog binary
5. run the trufflehog binary again and it will not detect all tokens present in the file for heroku service

## Environment
OS - kali Linux

### References

* #0000

ahrav:
https://github.com/trufflesecurity/trufflehog/releases/tag/v3.16.2 Should resolve this issue.

PR #856, opened 2022-10-21
[THOG-793] - Return all unverified results

ahrav:
Screen Shot 2022-10-30 at 1 32 43 PM
Remove the check to filter and return only a single unverified result.

https://github.com/trufflesecurity/trufflehog/issues/844


trufflesteeeve:
This feels like a change that should be behind a CLI flag if we want to do it. Though `CleanResults`'s comment explains what it does, it doesn't explain why we're doing it - which is essentially to cut down on the noise of detecting secrets that have either already been rotated, or aren't real.

ahrav:
> This feels like a change that should be behind a CLI flag if we want to do it. Though `CleanResults`'s comment explains what it does, it doesn't explain why we're doing it - which is essentially to cut down on the noise of detecting secrets that have either already been rotated, or aren't real.

Yea I agree. So this was actually a common thread in the community slack where users were not getting all the unverified secrets. After talking to Dustin we mainly did this before we had the Dashboard to filter out verified/ non-verified. So with that change in place this was the suggested solution since this would technically be the approach where there we no longer filter out actual unverified results.

That being said I think adding a cmd line flag for this makes sense 👍

dustin-decker:
It was put in place before we grouped credentials, but we do now, so maybe it is no longer needed? It definitely causes some unexpected behavior.

trufflesteeeve:
That seems reasonable. We can always add  functionality back in later if we need it. We might want to look into using more of [the AWS style custom cleaner](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/aws/aws.go#L182) for any key/secret detector types, especially if their secrets also match a standard hash.

ahrav:
> That seems reasonable. We can always add functionality back in later if we need it. We might want to look into using more of [the AWS style custom cleaner](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/aws/aws.go#L182) for any key/secret detector types, especially if their secrets also match a standard hash.

I have actually not seen this before. Yea I like this idea.

ahrav:
OKay, I think I actually found a pretty nice solution here. I move out the call to `CleanResults` into the engine vs within each detector. I also then included a CLI flag if we want to filter unverified results and pass that into the Engine during construction.

README.md
```diff
@@ -180,6 +180,7 @@ Flags:
       --concurrency=1            Number of concurrent workers.
       --no-verification          Don't verify the results.
       --only-verified            Only output verified results.
+      --filter-unverified        Only output first unverified result per chunk per detector if there are more than one results.
```
trufflesteeeve, line 183: __nitpick__: Typo - "more than one results." -> "more than one result."

main.go
```diff
@@ -30,15 +30,16 @@ import (
...
+	cli              = kingpin.New("TruffleHog", "TruffleHog is a tool for finding credentials.")
+	cmd              string
+	debug            = cli.Flag("debug", "Run in debug mode.").Bool()
+	trace            = cli.Flag("trace", "Run in trace mode.").Bool()
+	jsonOut          = cli.Flag("json", "Output in JSON format.").Short('j').Bool()
+	jsonLegacy       = cli.Flag("json-legacy", "Use the pre-v3.0 JSON format. Only works with git, gitlab, and github sources.").Bool()
+	concurrency      = cli.Flag("concurrency", "Number of concurrent workers.").Default(strconv.Itoa(runtime.NumCPU())).Int()
+	noVerification   = cli.Flag("no-verification", "Don't verify the results.").Bool()
+	onlyVerified     = cli.Flag("only-verified", "Only output verified results.").Bool()
+	filterUnverified = cli.Flag("filter-unverified", "Only output first unverified result per chunk per detector if there are more than one results.").Bool()
```
trufflesteeeve, line 42: __nitpick__: Typo - "more than one results." -> "more than one result."

dustin-decker (approved):

trufflesteeeve (approved):
My only comments are about a couple of typos. Looks good!

PR #859, opened 2022-10-24
Run chunker in pipeline

bill-rich:
Split up oversize chunks as part of the engine pipeline instead of handling it in each source.


pkg/sources/chunker.go
```diff
@@ -0,0 +1,41 @@
...
+		reader := bufio.NewReaderSize(bufio.NewReader(r), common.ChunkSize)
+		for {
+			chunkBytes := make([]byte, common.ChunkSize)
+			chunk := *originalChunk
+			n, err := reader.Read(chunkBytes)
+			if err != nil && !errors.Is(err, io.EOF) {
+				logrus.WithError(err).Error("Error chunking reader.")
+				break
+			}
+			peekData, _ := reader.Peek(common.PeekSize)
```
dustin-decker, line 32: nit: We should move the PeekSize and ChunkSize variables

dustin-decker: Do you think we should change chunks to have readers instead of []byte for their data?

dustin-decker (approved):

PR #863, opened 2022-10-24
continue scanning on detector / decoder panic

dustin-decker:


mcastorina (approved):
This means a worker will stop without restarting, which means if all workers panic we could be left with no goroutines scanning chunks.

Probably not a huge concern right now.

dustin-decker:
That's a good point, we can improve that.

zricethezav:
@mcastorina @dustin-decker should this PR be closed or do we still want to introduce this change?

mcastorina:
I think we can merge it and create a separate ticket for monitoring the number of workers we have.

PR #864, opened 2022-10-25
Copy buffer bytes

bill-rich:
buffer.Bytes() gives a pointer to the slice that is being reused. Need to copy the slice to avoid weirdness later.

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -406,7 +406,7 @@ func (s *Git) gitChunk(diff gitparse.Diff, fileName, email, hash, when, urlMetad
 					SourceID:       s.sourceID,
 					SourceType:     s.sourceType,
 					SourceMetadata: metadata,
-					Data:           newChunkBuffer.Bytes(),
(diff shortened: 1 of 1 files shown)


dustin-decker (approved):
Preallocating the slice and using copy would be better here 

https://gist.github.com/xogeny/b819af6a0cf8ba1caaef

trufflesteeeve:
Based on runs of the metrics, it seems like whatever you do, either copy or append, is good so long as you preallocate the destination slice.

bill-rich:
I thought pre-allocating should be the fastest too. For some reason in this case append seems to be slightly faster.

With `append` on my test repo:
32.593306s - 32.818896 seconds

With `make` and `copy`:
32.906427 - 33.12590 seconds

Not a huge difference, but I've been getting those results consistently. I'll merge as is for now to get things fixed and run some tests against a really large repo to see how the two compare.

dustin-decker:
There are probably some external variables there, the clean benchmarks show that pre-allocation with either append or copy is faster.

PR #867, opened 2022-10-26
proposal: SqlServer connection string detector

### LINKED ISSUE #866: SQLServer connection string detector
### Community Note

### Description

At moment trufflehog does not detect SQLServer connection strings at all

## Problem to be Addressed

Here is an example of `Program.cs`

```csharp
using System.Diagnostics;
using System.Text.Json;
using Microsoft.AspNetCore.Mvc;
using Microsoft.Data.SqlClient;
using Microsoft.EntityFrameworkCore;

var builder = WebApplication.CreateBuilder(args);

builder.Services
  .AddDbContext<Database>(optionsBuilder => optionsBuilder
    .UseSqlServer("Server=some.domain.com;Initial Catalog=Demo;User ID=sa;Password=PASSWORD_HERE"));

var app = builder.Build();
```

Here is another more common example from `appsettings.json`

```json
{
  "ConnectionStrings": {
    "Demo": "Server=some.domain.com;Initial Catalog=Demo;User ID=sa;Password=PASSWORD_HERE"
  }
}
```

In both cases trufflehog just not see connection strings which may become a reason for a such an critical leak

Add dedicated detector which will look for something looking like an SQLServer connection string and try to ping it

## Additional Context

There are already all required dependencies in trufflehog to do it and something similar is done in JDBC and MongoDB detectors

### References

- [MongoDB detector](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/mongodb/mongodb.go)
- [JDBC detector](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/jdbc/jdbc.go)

mac2000:
Closes #866 

**The problem:** at moment trufflehog does not see SQL Server connection strings at all 🤷‍♂️

e.g. here is part of `Program.cs`:

```csharp
var builder = WebApplication.CreateBuilder(args);

builder.Services
  .AddDbContext<Database>(optionsBuilder => optionsBuilder
    .UseSqlServer("Server=some.domain.com;Initial Catalog=Demo;User ID=sa;Password=PASSWORD_HERE"));

var app = builder.Build();
```

**Proposal:** this pull request adds dedicated SQLServer connection string detector to prevent such leaks

**How it works**

Because connection string itself is a set of case-insensitive key-value pairs delimited by a semicolon with regexp we are looking for something that may be a connection string

To prevent false positives we then try to parse it with [msdn.Parse](https://github.com/denisenkom/go-mssqldb/blob/master/msdsn/conn_str.go#L110)

And if everything fine as a last step we are trying to connect to server with connection string we have found

Also there is an `TestSQLServer_pattern` with few pattern dedicated tests to see if regexp will match desired strings


dustin-decker (approved):

PR #870, opened 2022-10-26
improve fastly validation endpoint and add extra data

dustin-decker:


mcastorina (approved):
TIL: `%t` for formatting bools!

PR #877, opened 2022-10-28
Add context to handler

bill-rich:
Take a context and break if ctx.Done().

### DIFF
--- a/pkg/handlers/handlers.go
+++ b/pkg/handlers/handlers.go
@@ -35,7 +35,7 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
 				chunk.Data = data
 				chunksChan <- &chunk
 			case <-ctx.Done():
-				break
+				return false
 			}
 			if handlerChan == nil {
 				break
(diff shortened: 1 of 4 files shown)


pkg/handlers/handlers.go
```diff
@@ -27,10 +28,18 @@ func HandleFile(file io.Reader, chunkSkel *sources.Chunk, chunksChan chan (*sour
...
-			chunksChan <- &chunk
+		for {
+			select {
+			case data := <-handlerChan:
+				chunk := *chunkSkel
+				chunk.Data = data
+				chunksChan <- &chunk
+			case <-ctx.Done():
+				break
+			}
```
mcastorina, line 39: **Suggestion:** `break` here actually breaks out of the `select`, not the outter `for` :/

Maybe `return true` or use a labelled `for`?

mcastorina (changes requested):

pkg/handlers/handlers.go
bill-rich: Thanks for catching that!

mcastorina (approved):

pkg/handlers/handlers.go
mcastorina: It's bitten me before 😅

PR #879, opened 2022-10-28
Do local URI verification, while attempting to defuse SSRF 

dustin-decker:
Previously we used the SSRF proxy to mitigate SSRF concerns of potential malicious payloads. However this meant that URI credentials were sent up to the proxy, and verification only worked if the resource was externally accessible.

This change performs verification locally to the scanner, after attempting to defuse http and https payloads. It does so with the following 3 modifications to the detected URI:
- drop the filename on the suffix - basic auth often applies on a directory level and this is one measure to prevent invoking state-changing dynamic content
- drop query parameters
- drop fragments

Additionally it verifies using a GET request.

So if there is a blind SSRF payload on a state-changing GET endpoint with no filepath, it could still fire, but otherwise it should be defused.

The SSRF proxy also previously verified redis and ftp, and these have been split up into new detectors.


pkg/detectors/mongodb/mongodb.go
```diff
@@ -21,7 +21,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 
 var (
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(mongodb(\+srv)?://(?:[^:]+:(?:[^@]+)?@)?(?:[^/]+|/.+.sock?,?)+(?:/([^/."*<>:|?]*))?\?(?:(.+=?=\S+)&?)+)\b`)
+	keyPat = regexp.MustCompile(`\b(mongodb(\+srv)?://[-.%\w{}]{1,50}:([-.%\S]{3,50})@[-.%\w\/:]+)\b`)
```
dustin-decker, line 24: The previous regex was a bit too inclusive and exclusive, sometimes grabbing html nearby and sometimes missing things that were picked up by this pattern, which is a modified version of the URI pattern but only for the mongodb scheme.

dustin-decker:
I'm going to split these up into separate detectors and update some of the verification logic.

pkg/detectors/ftp/ftp.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/jlaffaye/ftp"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct {
+	allowKnownTestSites bool
```
ahrav, line 18: nit: linter says this is unused.

pkg/detectors/ftp/ftp.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	keyPat = regexp.MustCompile(`\bftp:\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+
+	client = common.SaneHttpClient()
```
ahrav, line 27: ditto

pkg/detectors/ftp/ftp.go
```diff
@@ -0,0 +1,103 @@
...
+
+type Scanner struct {
+	allowKnownTestSites bool
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	keyPat = regexp.MustCompile(`\bftp:\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
```
ahrav, line 25: suggestion: You can remove a lot of those redundant escapes `\` as well as the `[]`

ex: `\bftp://\S{3,50}:(\S{3,50})@[-.%\w/:]+\` -> Gives you the same result.

pkg/detectors/ftp/ftp.go
```diff
@@ -0,0 +1,103 @@
...
+			continue
+		}
+
+		results = append(results, s)
+	}
+
+	return results, nil
+}
+
+func verifyFTP(ctx context.Context, u *url.URL) bool {
```
ahrav, line 88: nit: We can remove ctx as an arg since it doesn't look to be used.

pkg/detectors/npmtokenv2/npmtokenv2_test.go
```diff
@@ -1,7 +1,7 @@
 //go:build detectors
 // +build detectors
 
-package npmtoken_new
+package npmtokenv2
```
ahrav, line 4: nit: Linter mentions you have some undefined types. Likely need ot just change the name to use v2 instead of new?

pkg/detectors/redis/redis.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/go-redis/redis"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct {
+	allowKnownTestSites bool
```
ahrav, line 17: nit: same as above, unused.

pkg/detectors/redis/redis.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	keyPat = regexp.MustCompile(`\bredis:\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+
+	client = common.SaneHttpClient()
```
ahrav, line 26: ditto

pkg/detectors/redis/redis.go
```diff
@@ -0,0 +1,101 @@
...
+
+type Scanner struct {
+	allowKnownTestSites bool
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	keyPat = regexp.MustCompile(`\bredis:\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
```
ahrav, line 24: suggestion: This can also be simplified.

`\bredis://\S{3,50}:(\S{3,50})@[-.%\w/:]+\b`

pkg/detectors/redis/redis.go
```diff
@@ -0,0 +1,101 @@
...
+			continue
+		}
+
+		results = append(results, s)
+	}
+
+	return results, nil
+}
+
+func verifyRedis(ctx context.Context, u *url.URL) bool {
```
ahrav, line 87: nit: we can remove ctx as an arg here if unused.

pkg/detectors/uri/uri.go
```diff
@@ -23,7 +21,9 @@ type Scanner struct {
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b[a-zA-Z]{1,10}:?\/\/[-.%\w{}]{1,50}:([-.%\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\b(?:http(?:s)?:)?\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
```
ahrav, line 24: suggestion: simplify this regex and unwrap the extra capture group?

ex: `\b(?:https?:)?//\S{3,50}:(\S{3,50})@[-.%\w/:]+\b`

pkg/detectors/uri/uri.go
```diff
@@ -23,7 +21,9 @@ type Scanner struct {
...
+
+	client = common.SaneHttpClient()
 )
 
 type proxyRes struct {
```
ahrav, line 29: nit: I know you didn't touch this, but looks like we no longer use this struct. We can probably remove?

pkg/detectors/uri/uri.go
```diff
@@ -135,3 +97,53 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		return false
+	}
+	credentialedRes.Body.Close()
+
+	// If the credentialed URL returns a non 2XX code, we can assume it's a false positive.
+	if credentialedRes.StatusCode < 200 || credentialedRes.StatusCode > 299 {
+		return false
+	}
+
+	time.Sleep(time.Millisecond * 10)
```
ahrav, line 128: 🤣

ahrav (approved):
Do we also need to add the two new detectors to the default detectors?

PR #881, opened 2022-10-30
filesystem support for exclude and include filters

### LINKED ISSUE #536: Add --exclude-paths flag to filesystem module
### Description

Like https://github.com/trufflesecurity/trufflehog/issues/420 but for the filesystem module.

## Problem to be Addressed
We frequently use the `filesystem` module to scan large amounts of code. There are many directories we'd like to exclude for a variety of reasons (but the most common being test data that has intentional creds) and we were hoping we could creating a file to ignore these directories just like when using the `git` module.

Unfortunately, it seems the `--exclude-paths` flag is unavailable for filesystem scanning.

Provide `--exclude-paths` using exactly the same interface as the `git` module.

### LINKED ISSUE #865: gitignore
### Community Note

### Description

gitignore support

## Problem to be Addressed

Scanning of filesystem may become a challenging task because there are many folders which are not expected to be scanned, especially folders like node_modules which can contain quadrillion of files

Whenever we are going to scan file or dig into directory we should traverse directories tree to see if there is gitignore somewhere and check if that file or directory isn't there

Something like this one (should be optimized, but describes the idea):

```go
package main

import (
	"fmt"
	"log"
	"os"
	"path/filepath"

	ignore "github.com/sabhiram/go-gitignore"
)

func main() {
	// recursive iteration over `os.Args[1]` path
	filepath.Walk(os.Args[1], func(path string, info os.FileInfo, err error) error {
		if err != nil {
			log.Println(err)
		}

		// path == "/Users/mac/Desktop/demo/obj/Debug/net6.0/logging.AssemblyInfo.cs"
		// info.Name() == "logging.AssemblyInfo.cs"

		// inner, walk up till root or found gitignore
		var gitignore *ignore.GitIgnore
		parent := path
		for {
			gitignore, _ = ignore.CompileIgnoreFile(filepath.Join(parent, ".gitignore"))
			if gitignore != nil {
				// gitignore found, breaking loop
				break
			}
			if parent == filepath.Dir(parent) {
				// we are at root and there is no more parents, breaking loop
				break
			} else {
				parent = filepath.Dir(parent)
			}
		}

		ignored := false
		if gitignore != nil {
			ignored = gitignore.MatchesPath(path)
		}
		fmt.Printf("%v %s\n", ignored, path)
		return nil
	})
}
```

## Additional Context

image

### References

- [go-gitignore](https://github.com/sabhiram/go-gitignore) - library to parse gitignore

marchenko1985:
Closes: #536 and technically closes: #865

What has been done:

- I have moved the git filter creation a little bit lower into its switch case
- And created similar filter creation in the filesystem switch case
- Because of that go asked me to put `err :=` below

For filesystem engine, itself did not want to touch it so just added one more "method" `WithFilter()` so it can be used where needed

In the same way, as in the git source, we are checking if the filter is not nil and the current file passes it

---

Offtopic: while trying to figure out how to do it found one more crappy thing in the filter itself - it does not remove empty lines from the exclusion file, and it may lead to scenarios when by mistake we may exclude all files with space in the name

---

Tests: did not find any tests related to the filter in the git source so have nothing to copy from, and because I'm golang noob all that channel stuff just blows my brain so can not handle that one 🤷‍♂️


main.go
```diff
@@ -174,15 +174,14 @@ func run(state overseer.State) {
 		engine.WithDetectors(!*noVerification, engine.DefaultDetectors()...),
 	)
 
-	filter, err := common.FilterFromFiles(*gitScanIncludePaths, *gitScanExcludePaths)
-	if err != nil {
-		logrus.WithError(err).Fatal("could not create filter")
-	}
-
 	var repoPath string
 	var remote bool
```
ahrav, line 178: nit: If we declare `var err error` here, within the switch we can use regular assignment `=` vs `:=`

ahrav (approved):
Thanks a lot @mac2000 this looks like a great addition. Really appreciate it 😄 

Looks like there is a conflict in main, otherwise LGTM!

mac2000:
Hm, sorry, it was my fault, after messing with the GitHub sync of branches, I broke everything completely 🤷‍♂️

As a result figured out how to reset everything to upstream and cherry pick changes, otherwise my pull request was a mess of all changes from a PR creation

@ahrav if there is a chance to not wait for next conflict will be nice to merge it, seems like other similar changes were merged already (e.g. circleci flag for path exclusion)

> PS: I hope that GitHub will figure out what I'm trying to do, so no need to create another pull request, if I did everything correct, after reopening it we should see desired changes only

mac2000:
@dickc-sg I broke it completely 🤷‍♂️ 

Looking forward to second attempt #1033

@ahrav if you still here will be nice to have you approve there as well so it might be merged

mac2000:
@dickc-sg fyi seems like we have made it 💪

thanks to @dustin-decker 👍 #1033 pull request was approved and merged

Issue #882, opened 2022-10-30
skip update check from github action


mac2000:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Each time github action runs it will check for updates, but because we are using latest image it makes no sense 🤷‍♂️

## Problem to be Addressed

Technically it should speedup actions by few milliseconds 

## Description of the Preferred Solution

Add `--no-update` flag to action

PR #883, opened 2022-10-30
remove updates check from github action

mac2000:
Because action uses the latest image there is no need to check for updates each time

Closes: #882


dustin-decker (approved):

dustin-decker:
Thank you!

PR #896, opened 2022-11-02
add LDAP detector

dustin-decker:
Supports LDAP URIs with separate user and password, and a common MS library.
LDAP supported with non-tls, STARTTLS, and TLS. 
Integration test w/ OpenLDAP included.


mcastorina (approved):
Nice! Those integration tests look familiar :)

dustin-decker:
> Nice! Those integration tests look familiar :)

Yep! They were helpful. We might want to make some of that reusable.

Issue #897, opened 2022-11-03
sqlserver: wrong keywords


mac2000:
because of wrong keyword, sqlserver detector misses majority of files

instead of single `"sqlserver"` string it should be something like: `"Data Source", "Server", "User ID", "User", "UID", "Password"` which are parts of connection string

mnieto:
The sqlserver keyword is still needed. for example, a jdbc connection string looks like:
jdbc:**sqlserver**://;servername=server_name;authentication=SqlPassword;user=username;password=pass@word1

mac2000:
@mnieto there is an dedicated jdbc detector which [does catch this](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/jdbc/sqlserver.go#L24), this detector is more for connections strings outside of jdbc

zricethezav:
https://github.com/trufflesecurity/trufflehog/blob/834163acf5343c103e5d1b8bd88a952f23b85bc8/pkg/detectors/sqlserver/sqlserver.go#L28 keywords have been updated

PR #898, opened 2022-11-03
sqlserver wrong keywords fix for #897

mac2000:
closes: #897 

my mistake, sorry, because of wrong keywords majority of files were missing

I have added parts of connection string which are required for connection instead


dustin-decker:
The keywords are ORed. Can we reduce it to just the most unique one or two that will be present?

mac2000:
Hm probably you are right

At the very end we are hunting for password 

so it does not matter how much of other keywords matched (especially taking into account that they have aliases)

so we can leave only “password”

but unfortunately because of situation whole city now is without electricity and I am not sure if I can do it from phone

So if not will recreate it little bit later

dustin-decker:
I'll look into it more, I think password may be too common and may cause some performance issues.

Stay safe over there, and I hope your electricity is restored soon.

dustin-decker:
Does `Database=` seem like a viable keyword to you?

mac2000:
the problem here is that sql server connection string is an unordered set of key-value pairs 

but even worse some of the options have multiple aliases
 
in the majority of the cases, connection strings may have a bunch of properties but at minimum, we need a hostname, username, and password to perform verification

following are minimal required and valid examples:

```
Data Source=localhost;User ID=sa;Paddword=123
Server=localhost;User=sa;Password=123
Network address=localhost;UID=sa;Password=123
```

[here](https://github.com/denisenkom/go-mssqldb/blob/master/msdsn/conn_str.go#L358-L367) is an list of aliases

in the detector, we want to check only connection strings having passwords (aka I can not imagine someone allowing access to the database without authentication and hunting for secrets with trufflehog)

we are not verifying matches without a password - that's why scan time won't be affected for false-positives

but for keywords, it seems that we are left with hostname and username, and their combinations are the same as in the original commit so I reverted it back

at minimum it seems that it should be something like aliases from golang sql library mentioned above

ps: everything fine we are still standing, but did not find where to put an pull request to come back to "normality" :)

mac2000:
@dustin-decker forgot to mention `Database=` is not enough, aka alternatives are: `Databas           =    MyBlog`, `Initial Catalog=MyBlog`, ... 🤷‍♂️ (markdown eats spaces, we may have `\s*`)

Also, there may be no database property at all (aka connect to default one)

That's why I have mentioned all above about hostname, username, and password

And at the very end, we are going to verify things only if we find something looking like an SQL connection string and it has a password inside so I believe it should not affect overall performance (at least it won't be bigger than any other detector)

pagebrooks (approved):

mac2000:
as I have mentioned here is a list of valid connection string:

```
Data Source=localhost;User ID=sa;Paddword=123
Server=localhost;User=sa;Password=123
Network address=localhost;UID=sa;Password=123
```

Also take a note, the `Database` parameter may be omited, so default one will be used

But in either case it is definitely better than the nothing

dustin-decker:
I think we can include `Server=` and `Network address=` as well.

dustin-decker:
Closed by #1449

PR #899, opened 2022-11-03
Add a bytes chan source to make trufflehog usable as a Go library

hachh:
We want to use trufflehog as a Go library without using an existing source (directory, git, ...) but using bytes from a channel.
I added a small source that allows to use trufflehog directly on a chan of bytes, and example of how to use it as a library.


dxa4481:
Hi @hachh ,

To allow ourselves to continue to support and give back to open source, we sometimes have to make hard choices about features that we can't support. This enables us to sell an enterprise version of TruffleHog and use the revenue from that venture to give back to open source. For that reason, we cannot support this feature at this time, as we believe it may put at risk the enterprise side of our company. We may revisit and reconsider in the future, thank you for the contribution.

PR #901, opened 2022-11-03
[oc-147] - Add context to all git methods

ahrav:
- This will allow the recover to include information about the source that panics.
- Remove logrus and use custom logr.

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -20,8 +20,6 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/google/go-github/v42/github"
-	"github.com/rs/zerolog"
-	log "github.com/sirupsen/logrus"
 	"golang.org/x/oauth2"
 	"golang.org/x/sync/semaphore"
 	"google.golang.org/protobuf/proto"
@@ -217,6 +215,8 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
 		}
 
 	}
+
+	ctx.Logger().V(1).Info("Git source finished scanning", "repo-count", len(s.conn.Repositories))
 	s.SetProgressComplete(len(s.conn.Repositories), len(s.conn.Repositories), fmt.Sprintf("Completed scanning source %s", s.name), "")
 	return nil
 }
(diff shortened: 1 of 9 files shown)


pkg/sources/git/git.go
```diff
@@ -338,11 +336,12 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 
 	var depth int64
 	var reachedBase = false
-	log.WithField("repo", urlMetadata).Debugf("Scanning repo")
+
+	ctx.Logger().V(1).Info("scanning repo", "repo", urlMetadata, "base", scanOptions.BaseHash, "head", scanOptions.HeadHash)
 	for commit := range commitChan {
-		log.Tracef("Scanning commit %s", commit.Hash)
+		ctx.Logger().V(2).Info("scanning commit", "commit", commit.Hash, "message", commit.Message)
```
mcastorina, line 342: I would make this level 5 as it's very verbose.

pkg/sources/git/git.go
```diff
@@ -443,9 +440,8 @@ func (s *Git) gitChunk(diff gitparse.Diff, fileName, email, hash, when, urlMetad
 			}
 		}
 
-		_, err := newChunkBuffer.Write(line)
-		if err != nil {
-			log.WithError(err).Error("Could not write line to git diff buffer.")
+		if _, err := newChunkBuffer.Write(line); err != nil {
+			ctx.Logger().V(1).Info("error writing to chunk buffer", "error", err, "filename", fileName, "commit", hash, "file", diff.PathB)
 		}
```
mcastorina, line 445: Here we could use

```go
ctx.Logger().Error(err, "error writing to chunk buffer", "filename", fileName, ...)
```

which will be an error that gets sent to sentry. I think in this case it makes sense to do it how you have here though.

pkg/sources/git/git.go
```diff
@@ -799,7 +801,7 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
 }
 
 func handleBinary(ctx context.Context, repo *git.Repository, chunksChan chan *sources.Chunk, chunkSkel *sources.Chunk, commitHash plumbing.Hash, path string) error {
-	log.WithField("path", path).Trace("Binary file found in repository.")
+	ctx.Logger().V(1).Info("handling binary file", "path", path)
```
mcastorina, line 804: Maybe this should be level 5 too? Not sure how often it will log this.

pkg/sources/git/git.go
ahrav: Ahh good to know, I was unaware of that actually.

dustin-decker (approved):

PR #902, opened 2022-11-04
Add Unknown visibility level

trufflesteeeve:
Sometimes we don't know the visibility level.


proto/source_metadata.proto
```diff
@@ -154,6 +154,7 @@ enum Visibility {
   public = 0;
   private = 1;
   shared = 2;
+  unknown = 3;
```
ahrav, line 157: thanks for fixing this. I should've made it 0

ahrav (approved):

PR #909, opened 2022-11-10
fixed mailchimp detector

ankushgoel27:
Fixed issue with mailchimp detector where valid result was still showing unverified. Now it works fine


dustin-decker (approved):
Thank you @ankushgoel27 !

ankushgoel27:
the fix is still not working with sane HTTP client. Need to use the http.DefaultClient.Do. Not sure why the sane HTTP client is not working here

PR #910, opened 2022-11-11
Update Scrapfly API Key Format

jjsaunier:
Update Scrapfly API Key detector according to the new format


dustin-decker (approved):
Thank you!

PR #911, opened 2022-11-11
Fixed the brew install tap command

jjasghar:
It seems you need to run the full tap, not just trufflehog.

Signed-off-by: JJ Asghar <awesome@ibm.com>


dustin-decker:
Thank you! Looks like we can do away with the `brew tap` command with this as well.

dustin-decker (approved):

PR #930, opened 2022-11-18
Enable skipping of particular key IDs

dustin-decker:
Ideally we would have separate detect+verify calls to apply this logic, but we don't, and the use case is fairly limited, so just configuring it at initialization time right now.
Another option is just to run this detector twice with verify=false and verify=true.

### DIFF
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -17,10 +17,35 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{}
+type scanner struct {
+	skipIDs map[string]struct{}
+}
 
-// Ensure the Scanner satisfies the interface at compile time.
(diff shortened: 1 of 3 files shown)


pkg/detectors/aws/aws.go
```diff
@@ -72,6 +97,12 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 		resIDMatch := strings.TrimSpace(idMatch[1])
 
+		if s.skipIDs != nil {
```
mcastorina, line 100: **Nit (optional):** I think `s.skipIDs` is always non-nil since it's initialized in `New()`

mcastorina (approved):

pkg/detectors/aws/aws.go
dustin-decker: Yeah, it's probably not needed now since I made the struct not exported.

trufflesteeeve (approved):

PR #933, opened 2022-11-20
Fix error message

thiagola92:
I believe the message is missing the word "not"


ahrav (approved):
Thanks so much for getting this corrected 😄

PR #939, opened 2022-11-22
Add CustomRegex validation

mcastorina:

### DIFF
--- a/pkg/custom_detectors/regex_varstring.go
+++ b/pkg/custom_detectors/regex_varstring.go
@@ -7,7 +7,7 @@ import (
 )
 
 // nameGroupRegex matches `{ name . group }` ignoring any whitespace.
-var nameGroupRegex = regexp.MustCompile(`{\s*([a-zA-Z-_]+)\s*(\.\s*[0-9]*)?\s*}`)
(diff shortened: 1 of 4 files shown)


pkg/custom_detectors/regex_varstring.go
```diff
@@ -0,0 +1,44 @@
...
+// only contain alphanumeric, hyphen, and underscore characters. Group is
+// optional but if provided it must be a non-negative integer. If the group is
+// omitted it defaults to 0.
+type RegexVarString struct {
+	original string
+	// map from name to group
+	variables map[string]int
+}
+
+func NewRegexVarString(original string) RegexVarString {
```
trufflesteeeve, line 22: __Note__: At some point we might want these to error to ensure there isn't any unexpected behavior, but that addition definitely doesn't need to be a part of this PR, and this is great.

trufflesteeeve (approved):

pkg/custom_detectors/regex_varstring.go
mcastorina: Good point. For now we won't be using these function with the focus being on webhooks, but if/when we do implement templates we'll want to be clear with our errors and warnings.

PR #947, opened 2022-11-28
Add logger to context

bill-rich:


ahrav (approved):

main.go
```diff
@@ -208,7 +214,7 @@ func run(state overseer.State) {
 		}
 	case githubScan.FullCommand():
 		if len(*githubScanOrgs) == 0 && len(*githubScanRepos) == 0 {
-			log.Fatal("You must specify at least one organization or repository.")
+			logrus.Error("You must specify at least one organization or repository.")
 		}
```
mcastorina, line 218: Should we exit at this point? I think we'll continue after logging the error.

bill-rich: We should. Muscle memory got me here and I typed `Error` instead of `Fatal`

PR #959, opened 2022-12-06
Change chunker test source

bill-rich:
Switch to a reader with a predetermined size to limit error sources.

### DIFF
--- a/pkg/sources/chunker.go
+++ b/pkg/sources/chunker.go
@@ -37,10 +37,12 @@ func Chunker(originalChunk *Chunk) chan *Chunk {
 			}
 			peekData, _ := reader.Peek(PeekSize)
 			chunk.Data = append(chunkBytes[:n], peekData...)
+			if n > 0 {
+				chunkChan <- &chunk
+			}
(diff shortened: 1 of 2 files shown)


pkg/sources/chunker.go
```diff
@@ -37,10 +37,10 @@ func Chunker(originalChunk *Chunk) chan *Chunk {
 			}
 			peekData, _ := reader.Peek(PeekSize)
 			chunk.Data = append(chunkBytes[:n], peekData...)
-			chunkChan <- &chunk
 			if errors.Is(err, io.EOF) {
 				break
 			}
+			chunkChan <- &chunk
```
mcastorina, line 43: **Question:** I guess this is implementation dependent, but could we have read data _and_ gotten an EOF error? Maybe we should emit the chunk only if `n > 0`.

mcastorina (approved):

pkg/sources/chunker.go
bill-rich: Good call. Better to have the catch there just in case.

PR #960, opened 2022-12-06
Don't scan the --since-commit target

bill-rich:
The commit in `--since-commit` should not be scanned or pre-commit hooks won't work when head of main branch includes a secret.


mcastorina (approved):
Thanks for fixing this!

PR #968, opened 2022-12-09
Add configuration parsing and custom detectors to engine

mcastorina:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -16,9 +16,6 @@ type Config struct {
 
 // Read parses a given filename into a Config.
 func Read(filename string) (*Config, error) {
-	if filename == "" {
-		return &Config{}, nil
-	}
 	input, err := os.ReadFile(filename)
 	if err != nil {
 		return nil, err
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,48 @@
+package config
+
+import (
+	"os"
+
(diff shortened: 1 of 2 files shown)


mcastorina:
Waiting on #950 to be merged.

pkg/config/config.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Config holds user supplied configuration.
+type Config struct {
+	Detectors []detectors.Detector
+}
+
+// Read parses a given filename into a Config.
+func Read(filename string) (*Config, error) {
+	if filename == "" {
+		return &Config{}, nil
```
trufflesteeeve, line 20: __optional__: I feel like this should return an error instead of an empty config. Because an empty filename string would occur anytime `--config` is included but nothing follows it, yeah? (At least I think it would?) Which seems like an accident we should tell the user about, rather than intentionally adding an empty config.

trufflesteeeve (approved):
Looks good. I'd change the empty filename case, but this LGTM.

pkg/config/config.go
mcastorina: Hm.. that's a good point. I'll have to see if there's a way to differentiate between `--config ''` and no flag, because currently I think they both return `""`.

mcastorina: Found there's an `ExistingFile()` method in kingpin which will do the job. Now `--config ''` will error and no flag will not.

PR #977, opened 2022-12-14
Allow using a glob for include list.

ahrav:
- Add `include-repos` and `exclude-repos` flag that can be used with the `--org` flag when scanning Github.
- Allow for glob input for the exclude/include flags.

Screen Shot 2022-12-14 at 2 28 57 PM


pkg/sources/github/github.go
```diff
@@ -744,6 +757,25 @@ func (s *Source) getReposByUser(ctx context.Context, user string) ([]string, err
...
 
+func (s *Source) includeRepo(r string) bool {
+	if len(s.includeRepos) == 0 {
+		return true
+	}
+
+	for _, include := range s.includeRepos {
+		g, err := glob.Compile(include)
+		if err != nil {
+			s.log.WithError(err).Errorf("invalid glob %s", include)
```
bill-rich, line 768: I think it would be better to log it at a different level for enterprise to cut down on the non-dev actionable errors.

main.go
```diff
@@ -220,15 +222,24 @@ func run(state overseer.State) {
 		if len(*githubScanOrgs) == 0 && len(*githubScanRepos) == 0 {
 			logrus.Fatal("You must specify at least one organization or repository.")
 		}
+		var repos []string
+		// If an org is provided, the repos would be the IncludeRepos.
+		// If an org is not provided, the repos would be the repos passed in via the --repo flag.
+		if len(*githubScanOrgs) == 0 {
```
bill-rich, line 228: If a user specifies an `org` and a `repo` they should both be included.

pkg/sources/github/github.go
ahrav: that works with me. I think i copied this logic from below on L783 which is probably why its logged at this level. Might be worth at some point looking through all the logs in here.

main.go
ahrav: oh, interesting. I didn't even realize we supported both during the same command. good to know. Will update this.

pkg/sources/github/github.go
bill-rich: Oh yeah. I think we use error quite a bit. Just had this fresh on my mind after the retro.

mcastorina (approved):
lgtm, though I didn't see any logic for `ExcludeRepos`. Is that already being handled somewhere?

ahrav:
> lgtm, though I didn't see any logic for `ExcludeRepos`. Is that already being handled somewhere?

Yep we already have the logic within the source itself to exclude repos using the `ignoreRepos` field. 
https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/github/github.go#L150

We had implemented that logic for the enterprise version and just needed to add the flag to be used by OSS.

PR #978, opened 2022-12-14
Use Todoist's REST API v2

goncalossilva:
v1 was deprecated on December 5, 2022.

The response is unused, but for reference, here's the migration guide: https://developer.todoist.com/rest/v2/#migrating-from-v1


dustin-decker (approved):
Thank you!

PR #979, opened 2022-12-17
Capture callstack of canceled contexts

mcastorina:


mcastorina:
This is probably not thread safe, but I think worst case is we save the Nth call to `cancel()` instead of first..

dustin-decker (approved):

PR #986, opened 2022-12-22
Remove ctx from source structs

ahrav:


trufflesteeeve (approved):

proto/sources.proto
```diff
@@ -157,6 +157,7 @@ message GitHub {
   string head = 9;
   string base = 10;
   repeated string ignoreRepos = 11;
+  repeated string includeRepos = 12;
```
mcastorina, line 160: I think this shouldn't be part of this PR..

Issue #990, opened 2022-12-29
Buildkite token validation missing tokens without user_read scope


clonsdale-canva:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version

3.21.0

### Expected Behavior

Buildkite token is reported as valid

### Actual Behavior

Buildkite token is not validated as the API call fails due to missing user_read scope

## Additional Context

The logic to check if a buildkite token is valid will send out an API call to the /user endpoint https://github.com/trufflesecurity/trufflehog/blob/009756dce61948a66cf90a8b14018460c91ab4f0/pkg/detectors/buildkite/buildkite.go#L51. This will miss all tokens which do not have the `read_user` scope.

Instead, we can use the `access-token` endpoint, which will return 200 for any valid token, and report on the scopes present / ID of the token - https://buildkite.com/docs/apis/rest-api/access-token.

PR #991, opened 2022-12-29
Use access-token endpoint for validity check

clonsdale-canva:
This PR fixes the issue https://github.com/trufflesecurity/trufflehog/issues/990, it should correctly report keys as valid even if they are missing the user_read scope.


ahrav (approved):
Thanks so much for making this change. Really appreciate the contribution. 😃

dustin-decker (approved):

Issue #992, opened 2023-01-02
TruffleHog version printed to stdout when debug flag is enabled


yilmi:
### TruffleHog Version
This can be reproduced using the latest version of TruffleHog, at the time of writing this bug, version 3.21.0

### Trace Output
```bash
❯ trufflehog filesystem --directory hack --debug
DEBU[0000] running version 3.21.0                       
2023/01/02 15:46:37 [updater parent] run
2023/01/02 15:46:37 [updater parent] checking for updates...
2023/01/02 15:46:38 [updater parent] failed to get latest version: already up to date
2023/01/02 15:46:38 [updater parent] starting /opt/homebrew/bin/trufflehog
DEBU[0000] running version 3.21.0                       
2023/01/02 15:46:38 [updater child#1] run
2023/01/02 15:46:38 [updater child#1] start program
trufflehog 3.21.0 <-- HERE
DEBU[0000] running with up to 10 workers                
DEBU[0000] loaded 2 decoders                            
DEBU[0000] loaded 731 detectors total, 731 with verification enabled. 0 with verification disabled 
INFO[0000] starting pprof and fgprof server on :18066 /debug/pprof and /debug/fgprof 
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2023/01/02 15:46:39 [updater parent] checking for updates...
DEBU[0001] scanned 8 chunks                             
DEBU[0001] scanned 21323 bytes                          
2023/01/02 15:46:39 [updater parent] prog exited with 0
```

### Expected Behavior

Printing out the version in stdout when the -version flag can change the results output format in an unexpected way. When `--debug` is enabled, one would expect logging to be printed in stderr and not change the regular stdout output.

### Actual Behavior

If I do a run without debug and a run with debug, my results files are not the same

### Steps to Reproduce
```bash
❯ trufflehog filesystem --directory hack 2>/dev/null
❯ trufflehog filesystem --directory hack --debug 2>/dev/null
trufflehog 3.21.0
```

## Environment
This should apply to any environment

PR #993, opened 2023-01-02
Removing Debug version Println to logrus debug - Issue #992

yilmi:
Hi,

I created the Issue #992 for this pull request, the problem is that when debug is enabled, we print the trufflehog version in STDOUT instead of STDERR:
```bash
❯ trufflehog filesystem --directory hack 2>/dev/null
❯ trufflehog filesystem --directory hack --debug 2>/dev/null
trufflehog 3.21.0
```

The change is minor and shouldn't required additional testing or have much impact.

Thanks!

### DIFF
--- a/main.go
+++ b/main.go
@@ -150,7 +150,7 @@ func main() {
 
 func run(state overseer.State) {
 	if *debug {
-		fmt.Println("trufflehog " + version.BuildVersion)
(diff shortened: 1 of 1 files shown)


mcastorina (approved):
Nice, thanks for fixing this!

PR #1004, opened 2023-01-08
fix: do not override base parameter with default in GitHub Action

### LINKED ISSUE #1002: Missing argument for since-commit
### Community Note

### TruffleHog Version

v3.23.1

### Trace Output

https://gist.github.com/Huskydog9988/af7f1aeaf7d8cf7639ee426be7203b50

### Expected Behavior

Should provided the necessary commit for the `--since-commit` arg.

### Actual Behavior

No commit was provided

### Steps to Reproduce

Use the github action on v3.23.1

## Environment
Github actions

### References

Haven't found another case yet.

clarkedb:
This reverts #999 which mistakenly makes the `base` parameter for GitHub actions useless because it always sends the default value for `base` into the `--since-commit` flag. Furthermore that default value is  `''` which breaks so `trufflehog` cannot run.

I believe that if `base` is not specified, then the `input.base.default` value will be used automatically when accessing `input.base`. There is no need to add `default`, and doing so only ever gets the default value.

Closes #1002

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -27,7 +27,7 @@ runs:
     - git
     - file://${{ inputs.path }}
     - --since-commit
-    - ${{ inputs.base.default }}
+    - ${{ inputs.base }}
     - --branch
     - ${{ inputs.head }}
     - --fail


mcastorina (approved):
Thanks for finding and fixing this!

PR #1011, opened 2023-01-09
Copy metadata for line number aware sources

bill-rich:
Multiple results from the same chunk could result in line numbers in the metadata being overwritten. Deep copy the chunk and source metadata to avoid clobbering.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -234,8 +234,7 @@ func (e *Engine) detectorWorker(ctx context.Context) {
 							if isGitSource(chunk.SourceType) {
 								copyChunk := *chunk
 								copyMetaDataClone := proto.Clone(chunk.SourceMetadata)
-								copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData)
-								if ok {
+								if copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData); ok {
 									copyChunk.SourceMetadata = copyMetaData
 								}
 								fragStart, mdLine := FragmentFirstLine(&copyChunk)
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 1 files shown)


pkg/engine/engine.go
```diff
@@ -230,9 +230,20 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
 						for _, result := range results {
-							SetResultLineNumber(chunk, &result, fragStart, mdLine)
+							resultChunk := chunk
+							if isGitSource(chunk.SourceType) {
+								copyChunk := *chunk
+								copyMetaDataClone := proto.Clone(chunk.SourceMetadata)
+								copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData)
+								if ok {
+									copyChunk.SourceMetadata = copyMetaData
+								}
```
mcastorina, line 240: **Nit (optional):** These lines can be combined which provides limited scope to the variables

```go
if copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData); ok {
    copyChunk.SourceMetadata = copyMetaData
}
```

pkg/engine/engine.go
```diff
@@ -314,8 +325,6 @@ func FragmentFirstLine(chunk *sources.Chunk) (int64, *int64) {
 
 // SetResultLineNumber sets the line number in the provided result.
 func SetResultLineNumber(chunk *sources.Chunk, result *detectors.Result, fragStart int64, mdLine *int64) {
-	if isGitSource(chunk.SourceType) {
-		offset := FragmentLineOffset(chunk, result)
-		*mdLine = fragStart + offset
-	}
+	offset := FragmentLineOffset(chunk, result)
+	*mdLine = fragStart + offset
 }
```
mcastorina, line 330: I think `SetResultLineNumber` gets called in other places, so we may need to keep this `isGitSource` check.

mcastorina (commented):
Overall looks good, just one concern about `SetResultLineNumber`.

pkg/engine/engine.go
bill-rich: It does also in thog, but that'll need the same changes to fix line numbers as here.

mcastorina (approved):

PR #1013, opened 2023-01-09
Update entrypoint

ahrav:
Tested it on a personal project and it looks to be working.

Screen Shot 2023-01-11 at 5 45 53 PM

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env/bash
 
 # Parse the last argument into an array of extra_args.
 mapfile -t extra_args < <(bash -c "for arg in ${*: -1}; do echo \$arg; done")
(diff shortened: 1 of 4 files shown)


entrypoint.sh
```diff
@@ -2,4 +2,4 @@
 
 # `$*` expands the `args` supplied in an `array` individually
 # or splits `args` in a string separated by whitespace.
```
mcastorina, line 4: **Nit:** please update the comments too

mcastorina (commented):
This is one of those changes that's small but scary 😆 

Have you done any tests to make sure it doesn't completely break everything?

ahrav:
> This is one of those changes that's small but scary 😆
> 
> Have you done any tests to make sure it doesn't completely break everything?

I'm thinking of creating a test release from this branch to see what happens.

mcastorina (approved):

PR #1028, opened 2023-01-17
[chore] - Small cleanup of CircleCi source

ahrav:
- Fix some comments.
- Remove unused args for fxns.
- Remove deprecated ioutil.ReadAll.

### DIFF
--- a/pkg/sources/circleci/circleci.go
+++ b/pkg/sources/circleci/circleci.go
@@ -21,8 +21,8 @@ import (
 const baseURL = "https://circleci.com/api/v1.1/"
 
 type Source struct {
-	name,
-	token string
+	name     string
+	token    string
 	sourceId int64
 	jobId    int64
 	verify   bool
--- a/pkg/sources/circleci/circleci.go
+++ b/pkg/sources/circleci/circleci.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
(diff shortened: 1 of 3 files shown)


pkg/sources/circleci/circleci.go
```diff
@@ -109,7 +107,7 @@ type project struct {
 	RepoName string `json:"reponame"`
 }
 
-func (s *Source) projects(ctx context.Context) ([]project, error) {
+func (s *Source) projects() ([]project, error) {
```
mcastorina, line 110: **Nit (optional):** I don't see an issue with removing this though it might be more work to add it back if we want to log in these functions (or actually use the context).

mcastorina (approved):

pkg/sources/circleci/circleci.go
ahrav: I am going to leave this only because i tend to lean more in the camp of add it when you need it. I fear this could balloon into adding other args that are unused in the hopes they might be used later. I also do acknowledge its context and well in Go you just always pass the context, but it still doesn't sit well with me. There also is this little extra overhead when reading the function to try and map each arg to their usage in the fxn, and when they aren't used it's a little confusing. You tend to re-read the function several times to make sure it's not used.

I suppose a middle ground could be `_ context.Context`?

mcastorina: Yeah, I agree, and would be happy with either. Since they're all unexported functions it shouldn't be an issue to add the context back when we need it.

PR #1029, opened 2023-01-17
Add concurrency to CircleCi source

ahrav:
- Rebased on https://github.com/trufflesecurity/trufflehog/pull/1028 will merge after that is in.


pkg/sources/circleci/circleci.go
```diff
@@ -72,30 +78,46 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
-		}
-
-		for _, bld := range builds {
-			buildSteps, err := s.stepsForBuild(ctx, proj, bld)
+		proj := proj
+		s.jobPool.Go(func() error {
+			builds, err := s.buildsForProject(ctx, proj)
 			if err != nil {
-				return err
+				scanErrs = append(scanErrs, fmt.Errorf("error getting builds for project %s: %w", proj.RepoName, err))
```
mcastorina, line 91: **Suggestion:** we'll need to make this thread safe with a lock or something. Since we're only counting the errors maybe we could use an atomic counter?

mcastorina (changes requested):

pkg/sources/circleci/circleci.go
ahrav: Ahh yea, we also log the errors so i'll make a struct to encapsulate all the errors and add a mutex to ensure thread safety.

mcastorina (approved):
Nice!

PR #1030, opened 2023-01-18
Support file scanning in filesystem source

### LINKED ISSUE #1078: Ability to scan a single file
### Community Note

### Description

I would love for trufflehog to be able to scan a single file.

## Problem to be Addressed

Trufflehog currently "technically" supports being run in a pre-commit hook. However, instead of scanning staged files, it scans for _all files_ that have changed since the last commit, which is NOT the same.

If Trufflehog supports a way to scan a single file, I can run a hook (using something like husky + lint-staged) to run the trufflehog scan on every staged file individually, thus freeing trufflehog from even having to know what "staged files" even are, while limiting the scans strictly to the staged files.

Alternatively, if the Trufflehog team would rather implement a `--staged` (just like gitleaks), where Trufflehog scans only the staged files, it would be an alternate way of achieving the same thing as well, though it would probably be much more work for the team.

mcastorina:


proto/sources.proto
```diff
@@ -111,8 +111,11 @@ message ECR {
   repeated string registries = 2;
 }
 
-message Filesystem{
+message Filesystem {
+  // directories is deprecated and can be removed / renamed to paths when
+  // we no longer depend on the name in enterprise configs
   repeated string directories = 1;
+  repeated string paths = 2;
 }
```
mcastorina, line 119: I'm not sure how to best handle this. We need to keep `directories` for backwards compatibility with enterprise configs (like how we keep `--directories` flag functionality in this PR).

We could technically just have `directories` and pass all paths there, but the name is misleading.

I decided to create a new entry for paths to be clear about the data it contains and we can remove directories in the future if/when config files aren't directly dependent on proto names.

bill-rich: Suggestion: Add something like `// DEPRICATED` for stuff we want to remove in the next major release, so it is easy to grep for it since we probably won't remember when the time comes.

pkg/sources/filesystem/filesystem.go
```diff
@@ -83,100 +82,122 @@ func (s *Source) WithFilter(filter *common.Filter) {
...
-						},
-					},
-				},
-				Verify: s.verify,
-			}
-			return nil
-		})
+		fileInfo, err := os.Stat(cleanPath)
+		if err != nil {
+			logger.Error(err, "unable to scan path")
```
bill-rich, line 94: The error message could better describe what failed. Maybe something like `unable to get file info`

bill-rich (approved):
Looks good. Just two minor suggestions.

0x1 (approved):
lgtm

PR #1033, opened 2023-01-22
filesystem support for exclude and include filters (2nd attemp)

### LINKED ISSUE #536: Add --exclude-paths flag to filesystem module

### LINKED ISSUE #865: gitignore

marchenko1985:
Sorry I have messed up completely and somehow GitHub did not recognize that I want only my 2 commits to be merged and showed up bunch of changes from upstream 🤷‍♂️

Previous "broken" attempt: #881 had approve but were not merged and later on broken by me trying to fix conflicts

---

Closes: https://github.com/trufflesecurity/trufflehog/issues/536 and technically closes: https://github.com/trufflesecurity/trufflehog/issues/865

What has been done:

I have moved the git filter creation a little bit lower into its switch case
And created similar filter creation in the filesystem switch case
Because of that go asked me to put err := below
For filesystem engine, itself did not want to touch it so just added one more "method" WithFilter() so it can be used where needed

In the same way, as in the git source, we are checking if the filter is not nil and the current file passes it


dustin-decker (approved):

PR #1035, opened 2023-01-23
Limit diff size to prevent out of control memory use.

bill-rich:


pkg/gitparse/gitparse.go
```diff
@@ -20,6 +20,9 @@ import (
 // DateFormat is the standard date format for git.
 const DateFormat = "Mon Jan 02 15:04:05 2006 -0700"
 
+// MaxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
```
ahrav, line 23: nit: We can group the two consts together?

const (
    // DateFormat is the standard date format for git.
    DateFormat = "Mon Jan 02 15:04:05 2006 -0700"

    // MaxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
    MaxDiffSize = 1 * 1024 * 1024 * 1024 // 1GB
)

pkg/gitparse/gitparse_test.go
```diff
@@ -181,6 +181,30 @@ func TestMultiCommitContextDiff(t *testing.T) {
 	}
 }
 
+func TestMaxDiffSize(t *testing.T) {
+	bigBytes := bytes.Buffer{}
```
ahrav, line 185: bigBytes !! 😆

ahrav (approved):
Nicee.

One question, is there a plan to handle big diffs, or are we just ignoring them?

bill-rich:
> Nicee.
> 
> One question, is there a plan to handle big diffs, or are we just ignoring them?

For now, just ignoring them. Long term, I think something like a listing of files/repos/orgs etc that were not 100% scanned.

Issue #874, opened 2022-10-26
The tool will not run if the organization has a dash (-) in the name. Example: principal-aaas


principal-aaas:
sudo docker run -it -v "$PWD:/pwd" trufflesecurity/trufflehog:latest github --org=principal-aaas

principal-aaas:
Let me know if anybody is able to scan the principal-aaas organization such as I did in the example above or by another command. Thanks!

ahrav:
Hi @principal-aaas would you be able to post the error you are getting. I tested on a couple orgs with `-` in their names and it looks to work as expected. Ex: `rust-lang` (https://github.com/rust-lang) and `ant-design` (https://github.com/ant-design)

Thanks a lot.

principal-aaas:
Thank you for the message @ahrav. Here is the error that I am receiving:
Unable to find image 'trufflesecurity/trufflehog:latest' locally
latest: Pulling from trufflesecurity/trufflehog
9621f1afde84: Pull complete
c04b1df7ea1e: Pull complete
4f4fb700ef54: Pull complete
684cafaf20e7: Pull complete
d71608a2a2a1: Pull complete
Digest: sha256:4469c599fe8f709cb15531591ff5f0d5dd065f3600d9c6ca70c13ac55a0e6a97
Status: Downloaded newer image for trufflesecurity/trufflehog:latest
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

ERRO[0000] error fetching repos for org or user: principal-aaas  error="could not list repos for org principal-aaas: GET https://api.github.com/orgs/principal-aaas/repos?per_page=100: 404 Not Found []"

principal-aaas:
Does anybody have an idea as to why the tool could not list repos for org principal-aaas ?

dustin-decker:
`principal-aaas` is actually a user, but it should still work. TruffleHog first tries to enumerate the org and should fall back to the user. It works with my username, which also has a dash. 

I was able to reproduce this issue with `principal-aaas`

principal-aaas:
There was an interesting challenge posed with a generous reward by Gloria Brown to the first person that can resolve this issue. Check out her latest post: https://www.linkedin.com/in/gloria-brown-49961b1a8/

dustin-decker:
I just observed this when updating our integration tests to use a new user with a dash in the name. I suspect it's an issue with

- using the org enumeration endpoint for users
AND
[...]

PR #1042, opened 2023-01-25
fix github integration tests

dustin-decker:
A new credential is being used, so I've updated the integration tests.

A few other important things to mention:
- some of the unauthenticated tests have been commented out because of timing out while bumping up against rate limtis
- filtered for only organization account types in `addMembersByApp`
- only use login field for `addOrgsByUser`
- observed #874 with the new `trufflehog-sandbox` account and commented out those failing tests. Thoughts on the problem are in a comment within that issue and mentioned in the commented out code.


0x1 (approved):
lgtm

PR #1052, opened 2023-01-28
Handle errors in a thread safe manner

ahrav:

### DIFF
--- a/pkg/sources/errors.go
+++ b/pkg/sources/errors.go
@@ -2,14 +2,12 @@ package sources
 
 import (
 	"sync"
-	"sync/atomic"
 )
 
 // ScanErrors is used to collect errors encountered while scanning.
 // It ensures that errors are collected in a thread-safe manner.
 type ScanErrors struct {
-	count  uint64
-	mu     sync.Mutex
+	mu     sync.RWMutex
 	errors []error
 }
 
@@ -20,13 +18,14 @@ func NewScanErrors(projects int) *ScanErrors {
 
 // Add an error to the collection in a thread-safe manner.
 func (s *ScanErrors) Add(err error) {
-	atomic.AddUint64(&s.count, 1)
 	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.errors = append(s.errors, err)
-	s.mu.Unlock()
 }
 
 // Count returns the number of errors collected.
 func (s *ScanErrors) Count() uint64 {
-	return atomic.LoadUint64(&s.count)
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return uint64(len(s.errors))
 }
--- a/pkg/sources/errors.go
+++ b/pkg/sources/errors.go
@@ -1,0 +1,32 @@
+package sources
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// ScanErrors is used to collect errors encountered while scanning.
+// It ensures that errors are collected in a thread-safe manner.
+type ScanErrors struct {
+	count  uint64
+	mu     sync.Mutex
+	errors []error
+}
+
+// NewScanErrors creates a new thread safe error collector.
+func NewScanErrors(projects int) *ScanErrors {
+	return &ScanErrors{errors: make([]error, 0, projects)}
+}
+
+// Add an error to the collection in a thread-safe manner.
+func (s *ScanErrors) Add(err error) {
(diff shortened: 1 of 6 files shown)


pkg/sources/errors.go
```diff
@@ -0,0 +1,32 @@
...
+	"sync/atomic"
+)
+
+// ScanErrors is used to collect errors encountered while scanning.
+// It ensures that errors are collected in a thread-safe manner.
+type ScanErrors struct {
+	count  uint64
+	mu     sync.Mutex
+	errors []error
+}
```
mcastorina, line 14: **Nit:** `count` is `len(errors)`, could we just use that? I feel slightly better having a single source of truth for the number of errors. Also I think `mu` could be a `RWMutex`

pkg/sources/gitlab/gitlab.go
```diff
@@ -131,6 +133,55 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
+		log.WithError(repoErr).Warn("error getting repo")
+	}
+
+	// End early if we had errors getting specified repos but none were validated.
+	if len(errs) > 0 && len(repos) == 0 {
+		return errors.New("All specified repos had validation issues, ending scan")
+	}
+
+	// Get all repos if not specified.
+	if repos == nil {
```
mcastorina, line 155: **Question:** Should this check `len(repos) == 0` instead? A `nil` slice is technically different than an empty slice, but I'm not sure if we want the empty slice just not scan anything.

pkg/sources/gitlab/gitlab.go
```diff
@@ -131,6 +133,55 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
+		projects, err := s.getAllProjects(apiClient)
+		if err != nil {
+			return errors.New(err)
```
mcastorina, line 158: **Nit:** Maybe we could wrap the error? Or just return `err`?

pkg/sources/gitlab/gitlab.go
```diff
@@ -131,6 +133,55 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
+			// Ensure the urls are valid before adding them to the repo list.
+			_, err := url.Parse(prj.HTTPURLToRepo)
+			if err != nil {
+				fmt.Printf("could not parse url given by project: %s", prj.HTTPURLToRepo)
+			}
+			repos = append(repos, prj.HTTPURLToRepo)
+		}
+		if repos == nil {
+			return errors.Errorf("unable to discover any repos")
+		}
```
mcastorina, line 175: **Nit:** Same question about `len(repos) == 0`

pkg/sources/s3/s3.go
```diff
@@ -216,19 +218,12 @@ func (s *Source) pageChunker(ctx context.Context, client *s3.S3, chunksChan chan
...
-			continue
-		}
-		wg.Add(1)
-		go func(ctx context.Context, wg *sync.WaitGroup, sem *semaphore.Weighted, obj *s3.Object) {
+		s.jobPool.Go(func() error {
 			defer common.RecoverWithExit(ctx)
-			defer sem.Release(1)
-			defer wg.Done()
 
 			if (*obj.Key)[len(*obj.Key)-1:] == "/" {
```
mcastorina, line 224: **Nit:** if `obj.Key` is a `string` we could do `strings.HasSuffix(*obj.Key, "/")`

mcastorina (approved):
Overall looks good, just some optional nits

pkg/sources/errors.go
ahrav: So the idea here was to encapsulate all details around the errors within this object. Additionally errors is not accessible since it's a private field on this struct. All the sources that use this object should not have direct access to any of the internals. So while the single source of truth is the len(errors) it is not accessible except through the `Count()` method.

Oh, good point, will use RWMutex instead.

pkg/sources/gitlab/gitlab.go
ahrav: yep, seems like it.

note: this was just copy pasta code from what we already had. There could likely be additional things i missed.

pkg/sources/gitlab/gitlab.go
ahrav: sure, will do.

PR #1054, opened 2023-01-29
braintree detector: use production API URL instead of the test sandbo…

swdbo:
As the credentials required to access Braintree's Sandbox and production environments are distinct, this PR focuses solely on identifying and securing the production keys, similar to the logic employed in the [Stripe key detection](https://github.com/trufflesecurity/trufflehog/blob/26afa76e7c800f29afe7257ce41073d3b83c96cb/pkg/detectors/stripe/stripe.go#LL20C20-L20C20)


dustin-decker (approved):
Thank you!

PR #1061, opened 2023-02-01
Update float detector with correct User-Agent and regex

ahrav:
- Float API requires the User-Agent include an email for verification. https://developer.float.com/overview_authentication.html#identify-yourself
- Regex should also include `/`


pkg/common/http.go
```diff
@@ -99,10 +100,13 @@ func PinnedRetryableHttpClient() *http.Client {
...
-			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
-		}).DialContext,
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			d := net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}
+			return d.DialContext(ctx, network, addr)
+		},
```
mcastorina, line 109: Is this functionally the same?

pkg/detectors/float/float.go
```diff
@@ -53,6 +53,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				continue
 			}
 			req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", resMatch))
+			req.Header.Add("User-Agent", "TruffleHog3 (test@test.com)")
```
mcastorina, line 56: **Nit:** I think `example@example.com` is a more secure email to attach to the user agent because it is an [IANA reserved domain](https://www.iana.org/domains/reserved) (in the off-chance the email actually gets used).

mcastorina (approved):

pkg/common/http.go
ahrav: it should be. Although I actually don't need to even update this file. The old version had my linter yelling at me so I updated it, but i'll ignore it.

PR #1065, opened 2023-02-03
Make signable detector regex more specifc

marcinantkiewicz:
add \b to the regex to avoid matching ".. assignable: 123456.."

### DIFF
--- a/pkg/detectors/signable/signable.go
+++ b/pkg/detectors/signable/signable.go
@@ -28,7 +28,7 @@ var (
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"signable"}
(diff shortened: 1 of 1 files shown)


pkg/detectors/signable/signable.go
```diff
@@ -28,7 +28,7 @@ var (
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"signable"}
+	return []string{"\bsignable"}
```
dustin-decker, line 31: This is used for substring matching. You might be looking to modify the regex pattern in the file

dustin-decker (changes requested):

zricethezav:
closing due to inactivity, feel free to reopen with the suggested changes @marcinantkiewicz

PR #1064, opened 2023-02-02
Add TruffleHog version input for GitHub action

mcastorina:
Closes #1040

PR #1066, opened 2023-02-03
Fixes syntax on action.yml

harolhiguera:
Fixes the issue pointed out here:
- https://github.com/trufflesecurity/trufflehog/pull/1064#discussion_r1095242880

Related PR
- https://github.com/trufflesecurity/trufflehog/pull/1064

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -6,7 +6,7 @@ inputs:
   version:
     description: TruffleHog version.
     required: false
-    default: latest
+    default: 'latest'
   path:
     description: Repository path.
     required: true
--- a/action.yml
+++ b/action.yml
@@ -4,6 +4,7 @@ author: Truffle Security Co. <support@trufflesec.com>
 
 inputs:
   version:
+    type: string
     description: TruffleHog version.
     required: false
     default: 'latest'


aldy505:
I don't think just quoting the latest would make any difference. As the yaml specification note that if it contains alphabetical character, it would be parsed as string.

Rather, I'd guess that Github doesn't really support reading the `inputs` arguments for the `image` field. But that's just my guess and I can't prove that I'm 100% correct.

harolhiguera:
> I don't think just quoting the latest would make any difference. As the yaml specification note that if it contains alphabetical character, it would be parsed as string.
> 
> Rather, I'd guess that Github doesn't really support reading the `inputs` arguments for the `image` field. But that's just my guess and I can't prove that I'm 100% correct.

Thank you for your feedback,

Your argument makes sense. Nevertheless, I supposed that would fix it because it is already done in these lines. But, haven’t really proved it. I just looked at the documentation.  Since you mention that, I explicitly added the type field. 

image

mcastorina (changes requested):
If only it were that easy. Unfortunately I think the problem might be related to how we're using the input in `image` (as @aldy505 mentions).

I tested this change locally and it still breaks. For now the `version` input has been reverted.

I'll leave this PR open if you want to iterate on it, otherwise feel free to close it (and thanks for the attempt).

PR #1071, opened 2023-02-04
[chore] - Add tests for errors

ahrav:

### DIFF
--- a/pkg/sources/errors_test.go
+++ b/pkg/sources/errors_test.go
@@ -1,0 +1,149 @@
+package sources
+
+import (
+	"reflect"
+	"sync"
+	"testing"
+)
+
+func TestNewScanErrors(t *testing.T) {
+	testCases := []struct {
+		name     string
+		projects int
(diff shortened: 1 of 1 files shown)


pkg/sources/errors_test.go
```diff
@@ -0,0 +1,149 @@
...
+			want: &ScanErrors{
+				errors: make([]error, 0, 1),
+			},
+		},
+		{
+			name:     "fifty projects",
+			projects: 50,
+			want: &ScanErrors{
+				errors: make([]error, 0, 50),
+			},
```
mcastorina, line 34: Hm, I wonder if these tests don't have enough leniency? For example, does it matter if the underlying slice has exactly 50 allocated here?

I'm wondering too if preallocating the slice is being too pessimistic about scanning. Most of the time there shouldn't be any errors (hopefully).

mcastorina (approved):
I love tests, thanks for adding them!

PR #1072, opened 2023-02-04
Pull gitparse config options out of pkg consts

bill-rich:
Make gitparse options configurable to make testing easier.

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -17,14 +17,6 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
-const (
-	// DateFormat is the standard date format for git.
-	DateFormat = "Mon Jan 02 15:04:05 2006 -0700"
-
-	// MaxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
-	MaxDiffSize = 1 * 1024 * 1024 * 1024 // 1GB
-)
-
 // Commit contains commit header info and diffs.
 type Commit struct {
 	Hash    string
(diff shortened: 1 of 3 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -17,14 +17,6 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
-const (
```
ahrav, line 20: question: Can we actually leave the constants and then just use them in your factory function for `NewParser', but also make them private? I think they are only ever used in this file.

Ex:
```
const (
	// defaultDateFormat is the standard date format for git.
	defaultDateFormat = "Mon Jan 02 15:04:05 2006 -0700"

	// maxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
	maxDiffSize = 1 * 1024 * 1024 * 1024 // 1GB
	// maxCommitSize is the maximum size for a commit. Larger commits will be cut off.
	maxCommitSize = maxDiffSize
)
```

```
// NewParser creates a GitParse config from options and sets defaults.
func NewParser(options ...Option) *Parser {
	parser := &Parser{
		dateFormat:    defaultDateFormat,
		maxDiffSize:   maxDiffSize,   // 1GB
		maxCommitSize: maxCommitSize, // 1GB
	}
	for _, option := range options {
		option(parser)
	}
	return parser
}
```

pkg/gitparse/gitparse.go
```diff
@@ -42,6 +34,46 @@ type Diff struct {
 	IsBinary  bool
 }
 
+// Parser sets values used in GitParse.
+type Parser struct {
+	MaxDiffSize   int
```
ahrav, line 39: nit: I think we can make all the fields in here private to reduce the exposure? If we can then we would just update the reaming functions to account for this change.

pkg/gitparse/gitparse.go
```diff
@@ -42,6 +34,46 @@ type Diff struct {
...
+	MaxCommitSize int
+	DateFormat    string
+}
+
+// MaxDiffSize sets MaxDiffSize option. Diffs larger than MaxDiffSize will
+// be truncated.
+func MaxDiffSize(MaxDiffSize int) Option {
```
ahrav, line 46: super nit: Rename theses to `WithMaxDiffSize` and `WithMaxCommitSize` that way we are consistent with our engine and few other places we use functional variadic args.

Also the param in this function can be camelcased?
`func MaxDiffSize(MaxDiffSize int) Option` -> `func MaxDiffSize(maxDiffSize int) Option`

Same with MaxCommitSize.

I feel like making all the things private, keeps the global namespace a tiny bit cleaner.
```
func WithMaxDiffSize(maxDiffSize int) Option {
	return func(parser *Parser) {
		parser.maxDiffSize = maxDiffSize
	}
}
```

ahrav (approved):
Love this 🙌

bill-rich:
Good call on all the feedback. Got it all fixed up.

PR #1077, opened 2023-02-07
Make archive handler configurable

bill-rich:
Went with making vars in the handler package rather than a configuration that is passed along because of the awkwardness of A) passing the settings through each source type that can scan files, and B) the way handlers are set up to be self-contained units that implement the `Handler` interface.

Also included handling canceled contexts so we can use timeouts along with max size or depth. This will allow us to up the max size without letting extraction time get out of control.

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/mholt/archiver/v4"
 	log "github.com/sirupsen/logrus"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 )
 
 type ctxKey int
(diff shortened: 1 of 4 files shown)


pkg/handlers/archive.go
```diff
@@ -153,23 +170,28 @@ func (d *Archive) ReadToMax(reader io.Reader) (data []byte, err error) {
...
+				fileContent.Write(fileChunk[0:bRead])
+			}
+			if bRead < 512 {
+				return fileContent.Bytes(), nil
+			}
+			if d.size >= maxSize && bRead == 512 {
+				log.Debug("Max archive size reached.")
+				return fileContent.Bytes(), nil
+			}
 		}
```
mcastorina, line 195: **Nit (optional):** I think we can use `common.IsDone(ctx)` at the start of the loop and remove the `select`. That could help de-indent the body.

mcastorina (approved):
Nice! If it's not time-consuming, could we add a test for timing out? Otherwise lgtm

PR #1079, opened 2023-02-07
Add max commit size

bill-rich:
MaxDiffSize was added in a previous PR, but multiple large diffs could still be included in one commit leading to a single commit holding all the diffs in memory while parsing the rest of the commit.

This adds a maximum commit size. If the commit is larger than the max size, it will be dropped into the commitChan and an identical commit (minus existing diffs) will continue to be worked on.

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -194,97 +194,123 @@ func (c *Parser) fromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 	var currentDiff *Diff
 
 	defer common.RecoverWithExit(ctx)
+	defer close(commitChan)
 	for {
-		line, err := outReader.ReadBytes([]byte("\n")[0])
-		if err != nil && len(line) == 0 {
-			break
-		}
-		switch {
-		case isCommitLine(line):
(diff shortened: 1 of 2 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -194,97 +194,123 @@ func (c *Parser) fromReader(ctx context.Context, stdOut io.Reader, commitChan ch
...
-		case isBinaryLine(line):
-			currentDiff.IsBinary = true
-			currentDiff.PathB = pathFromBinaryLine(line)
-		case isLineNumberDiffLine(line):
-			if currentDiff != nil && currentDiff.Content.Len() > 0 {
-				currentCommit.Diffs = append(currentCommit.Diffs, *currentDiff)
-			}
-			newDiff := &Diff{
-				PathB: currentDiff.PathB,
+		select {
```
ahrav, line 199: question: Could we remove this select and replace it w/ `common.IsDone(ctx)` just to reduce the indentation. I think that has the same effect 🤔 
```
if common.IsDone(ctx) {
    cleanupParse(currentCommit, currentDiff, commitChan)
    return
}
```

ahrav (approved):

PR #1084, opened 2023-02-09
Make detectors configurable

### LINKED ISSUE #984: Run certain Detector Type
trufflehog version: trufflehog dev

Currently I am running trufflehog as a pre-commit hook with all possible Detector type. Is it possible to only run few Detector types , say AWS keys, Private keys as such?

ahrav:
Allow for the list of detectors used during verification to be configurable via a CLI flag.

- Add `detectors` CLI flag that accepts a comma separated list of include and exclude detector names.
- If the values for the `detectors` flag are invalid or are names of detectors that we do not support, default to using all detectors.

NOTE: As a benchmark running a scan against the `https://github.com/trufflesecurity/trufflehog.git` repo using all detectors vs only `AWS` resulted in a scan that ran ~16s faster.
All detectors ~ 40s
AWS detector only ~24s

Screenshot 2023-02-10 at 4 18 54 PM


main.go
```diff
@@ -207,13 +208,22 @@ func run(state overseer.State) {
...
+			engine.WithDetectors(!*noVerification, engine.Detectors(ctx, strings.Split(*detectors, ","))...),
+		)
+	} else {
+		e = engine.Start(ctx,
+			engine.WithConcurrency(*concurrency),
+			engine.WithDecoders(decoders.DefaultDecoders()...),
+			engine.WithDetectors(!*noVerification, engine.DefaultDetectors()...),
+			engine.WithDetectors(!*noVerification, conf.Detectors...),
+		)
+	}
```
mcastorina, line 226: **Suggestion:** I think the way this is setup wouldn't allow filters with custom detectors. Maybe the filter should be applied after the list of detectors have been configured?

ahrav: Ah okay. In my head i guess i was thinking if they had custom detectors they wouldn't want any of the non-custom detectors. I can potentially see wanting to use custom with select detectors too.

main.go
```diff
@@ -207,12 +208,20 @@ func run(state overseer.State) {
...
+
 	e := engine.Start(ctx,
 		engine.WithConcurrency(*concurrency),
 		engine.WithDecoders(decoders.DefaultDecoders()...),
-		engine.WithDetectors(!*noVerification, engine.DefaultDetectors()...),
 		engine.WithDetectors(!*noVerification, conf.Detectors...),
 		engine.WithFilterUnverified(*filterUnverified),
+		detectorsOption,
 	)
```
mcastorina, line 225: Nice - so `--config config.yaml --detectors AWS` would use custom detectors and AWS detectors. Does that make sense, or should it be `--config config.yaml --detectors AWS,CustomRegex`?

ahrav: I think that's fine, given the CustomRegex is a bit special due to the fact we pass it via a config. It seems like maybe it should stand on its own? I don't have much of a preference though.

ahrav: oh, i just noticed your comment has a config either way.

mcastorina: Yeah - in the future the config file might have more than one custom detector (or other config options), so being able to filter them by detector type makes sense to me?

Maybe I'm being overly cautious here but it's easier to relax restrictions than to add them so I slightly prefer `--detectors AWS,CustomRegex`.

dustin-decker (approved):

pkg/engine/detectors.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+)
+
+// Detectors only returns a specific set of detectors if they are specified in the
+// detectors list and are valid. Otherwise, it returns the default set of detectors.
```
trufflesteeeve, line 11: __strong suggestion__: I think including a detector in this list that ends up being invalid should be a fatal error. Otherwise a user will do a (potentially very long) scan, and not necessarily know that they didn't actually scan for their githib token or some other typo or misnamed detector.

pkg/engine/detectors_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		{
+			name:     "Valid detectors (2) specified, returns valid set",
+			dts:      []string{"AWS", "Azure"},
+			expected: []detectors.Detector{aws.New(), &azure.Scanner{}},
+		},
+		{
+			name:     "Invalid detector specified, returns only valid set",
+			dts:      []string{"AWS", "InvalidType"},
+			expected: []detectors.Detector{aws.New()},
```
trufflesteeeve, line 38: __related to other comment__: As mentioned above, I think the expectation here should be some variety of fatal error, so accidental misconfigurations can't happen.

trufflesteeeve (commented):
This looks good, though I would strongly suggest we consider misconfigurations here to be an error that we should not continue with the scan, but rather let the user know the name of every expected detector type that didn't match with an actual detector.

dustin-decker:
We might also want to consider an inverse option, like use all detectors except for x,y,z

pkg/engine/detectors.go
ahrav: done ✅

ahrav:
> We might also want to consider an inverse option, like use all detectors except for x,y,z That's the more common use case I think

Will add that functionality in, and update the CLI flags.

pkg/engine/detectors.go
trufflesteeeve: Would it be simple enough to include the misconfigured detectors by putting them in a separate list output from `filterDetectors` or something like that? Just so we can tell the user specifically which ones aren't valid?

PR #1086, opened 2023-02-09
[chore] Remove logrus from github source

mcastorina:


ahrav (approved):
looks like there is a failing test after making `handleRateLimit` a method on the Source, otherwise lgtm

Issue #420, opened 2022-04-17
Add `--exclude-paths` flag to github module


bugbaba:
Hi :)

**Is your feature request related to a problem? Please describe.**
Files like `yarn.lock`, `package-lock.json`, etc are only used to keep track of dependencies of a project. They are getting matched as finding by [signable.go](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/signable/)  regex for example in this below case `-explode-assignable-expression-7`  was matched by the regex in `yarn.lock` file.

![Screenshot_2022-04-17_15-32-17](https://user-images.githubusercontent.com/29960776/163709863-fab68512-e97d-43db-92dc-a6c28fe59427.jpg)

![image](https://user-images.githubusercontent.com/29960776/163709764-ceed660d-a5d9-4cec-80c5-831b30245173.png)

**Describe the solution you'd like**
Add  `--exclude-paths` flag to github module. Currently its available for `git` module only. So that users can exclude files/directories manually as per their use case this will help in reducing the false positive cases and the time taken to scan the repository.

**Describe alternatives you've considered**
Using  the `--only-verified` flag can be used to stop this from showing as finding as the match won't work against the API. But this isn't an efficient/best way to solve this issue in my opinion. `--exclude-paths` flag is more practical specially when the user knows these files need to ignored.

cdavie-weconnect:
I would also like a way to exclude paths when using the Github Action, specifically when testing locally with https://github.com/nektos/act. It would be extra convenient if it could respect my gitignore.

manoelhc:
I like the `.<app>ignore` approach. For example, a `.thignore` file could be more helpful once you can exclude paths by repo and you wouldn't need to create long command lines.

danielfree:
I'd like to have a similar function when scanning the entire org, but exclude several repos, because there're some forked open-source repo that will generate tons of unverified results, e.g., JDBC url in test cases( jdbc:postgresql://localhost:5432/****)

ahrav:
closing this as it looks like a fix for this was merged in on Feb 9.

PR #1087, opened 2023-02-09
Github filter support for exclude and include

MetinSAYGIN:
Hello,

For the ticket #420
- /main.go
    I added the options and descriptions
    edited githubScan.FullCommand() to use the include-paths or exclude-paths according to it.

- /pkg/engine/github.go
    used the filter on ScanGitHub function
    added source.WithScanOptions(scanOptions)

- /sources/github/github.go
    added WithScanOptions (same from the gitlab/gitlab.go file)
    changed scanOptions for s.scanOptions
    removed the scanOptions var because its no longer used

I did these changes after studying the merge for gitlab filter case.

Metin


ahrav (approved):
Thanks a lot Metin 😄 Appreciate the contribution.

PR #1088, opened 2023-02-09
Add Type() to detector interface

trufflesteeeve:
The goal here is to allow the detector type information to be used without the need for reflection. This could possibly allow us to more easily inject information into detectors or filter them out if necessary.


bill-rich (approved):
Looks good other than maybe the linter issue.

ahrav (approved):

dustin-decker (approved):

PR #1089, opened 2023-02-10
[chore] Remove logrus from circleci, filesystem, gitlab, and s3 sources

mcastorina:

### DIFF
--- a/pkg/engine/s3.go
+++ b/pkg/engine/s3.go
@@ -46,6 +46,10 @@ func (e *Engine) ScanS3(ctx context.Context, c sources.Config) error {
 	}
 
 	s3Source := s3.Source{}
+	ctx = context.WithValues(ctx,
+		"source_type", s3Source.Type(),
+		"name", "trufflehog - gitlab",
+	)
 	err = s3Source.Init(ctx, "trufflehog - s3", 0, int64(sourcespb.SourceType_SOURCE_TYPE_S3), true, &conn, runtime.NumCPU())
 	if err != nil {
 		return errors.WrapPrefix(err, "failed to init S3 source", 0)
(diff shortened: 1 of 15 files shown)


pkg/engine/s3.go
```diff
@@ -46,6 +46,10 @@ func (e *Engine) ScanS3(ctx context.Context, c sources.Config) error {
 	}
 
 	s3Source := s3.Source{}
+	ctx = context.WithValues(ctx,
+		"source_type", s3Source.Type(),
+		"name", "trufflehog - gitlab",
```
dustin-decker, line 51: This should be S3

dustin-decker: would it be better to say `<name> source`?

pkg/sources/gitlab/gitlab.go
```diff
@@ -383,35 +384,35 @@ func (s *Source) scanRepos(ctx context.Context, chunksChan chan *sources.Chunk)
...
 	return nil
 }
 
-func (s *Source) ignoreRepo(r string) bool {
+func (s *Source) ignoreRepo(ctx context.Context, r string) bool {
 	for _, ignore := range s.ignoreRepos {
 		g, err := glob.Compile(ignore)
 		if err != nil {
-			log.WithError(err).Errorf("could not compile ignore repo glob %s", ignore)
+			ctx.Logger().V(2).Info("could not compile ignore repo glob", "glob", ignore, "error", err)
```
dustin-decker, line 411: This should be a higher level IMO

dustin-decker: It's error or warning kind of information I think

dustin-decker (approved):
Thanks, Miccah!

pkg/engine/s3.go
mcastorina: Ah, nice catch!

> would it be better to say `<name> source`?

I was just copying the name that we pass to the source, but I think either one is good. Would that be `s3 source` in this case?

mcastorina: I decided to make the key `"source_name"` and the value just `"s3"`.

PR #1090, opened 2023-02-10
Modify Init on Sources interface

ahrav:
Using a config instead of explicit parameters allows for more flexibility when configuring sources.
This also reduces the tight coupling between the sources and how they are initialized.

### DIFF
--- a/pkg/cache/simple.go
+++ b/pkg/cache/simple.go
@@ -1,0 +1,50 @@
+package cache
+
+import (
+	"sync"
+)
+
+// Cache in-memory cache.
+type Cache struct {
+	mu sync.RWMutex
+	m  map[string]string
+}
+
+// Simple constructs a new Simple cache.
+func Simple() *Cache {
+	return &Cache{
+		m: make(map[string]string),
+	}
+}
+
+// Set a value in the cache.
+func (c *Cache) Set(key, value string) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.m[key] = value
+	return nil
+}
+
+// Get a value and a boolean indicating if the value was found.
+func (c *Cache) Get(key string) (string, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	value, ok := c.m[key]
+	return value, ok
+}
+
+// Delete a value from the cache.
+func (c *Cache) Delete(key string) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.m, key)
+	return nil
+}
+
+// Clear the cache.
+func (c *Cache) Clear() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.m = make(map[string]string)
+	return nil
+}
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -5,6 +5,7 @@ import (
 
 	"google.golang.org/protobuf/types/known/anypb"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/cache"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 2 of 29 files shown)


pkg/sources/syslog/syslog.go
```diff
@@ -82,15 +82,15 @@ func (s *Source) InjectConnection(conn *sourcespb.Syslog) {
...
+func (s *Source) Init(_ context.Context, cfg sources.SourceConfig) error {
 
-	s.name = name
-	s.sourceId = sourceId
-	s.jobId = jobId
-	s.verify = verify
+	s.name = cfg.Name
+	s.sourceId = cfg.SourceID
+	s.jobId = cfg.JobID
+	s.verify = cfg.Verify
```
mcastorina, line 90: **Suggestion (optional):** the sources could either embed `sources.SourceConfig` or store `cfg` directly in the source instead of individual entries. Marking as optional because this would touch a lot of code and could be done in another PR.

pkg/sources/sources.go
```diff
@@ -37,7 +103,7 @@ type Source interface {
 	// JobID returns the initialized job ID used for tracking relationships in the DB.
 	JobID() int64
 	// Init initializes the source.
-	Init(aCtx context.Context, name string, jobId, sourceId int64, verify bool, connection *anypb.Any, concurrency int) error
+	Init(aCtx context.Context, cfg SourceConfig) error
```
mcastorina, line 106: **Idea:** it would be nice to have the strictly required information be passed as arguments and the rest as variadic args.

I *think* `connection` is the only strictly required item.

```go
source := git.Source{}
source.Init(ctx, conn,
    sources.WithName("foo"),
    sources.WithJobID(123),
    sources.WithVerify(true),
)
```

Another idea is the same but using the config struct instead of variadic args:

```go
source.Init(ctx, conn, &sources.Config{
    Name:   "foo",
    JobID:  123,
    Verify: true,
})
```

pkg/sources/sources.go
```diff
@@ -28,6 +29,71 @@ type Chunk struct {
...
+	// SourceID is the ID of the source.
+	SourceID int64
+	// Verify specifies whether any secrets in the Chunk should be verified.
+	Verify bool
+	// Connection is the connection information for the source.
+	Connection *anypb.Any
+	// Concurrency is the number of concurrent workers to use for the source.
+	Concurrency int
+	// Cache used by Sources to help reduce duplicate data scans.
+	Cache Cache
```
mcastorina, line 79: Hm.. I'm not convinced this should be part of a general purpose config struct.

pkg/cache/simple.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"sync"
+)
+
+// Cache in-memory cache.
+type Cache struct {
+	mu sync.RWMutex
+	m  map[string]string
+}
```
mcastorina, line 11: As much as I love writing caches, is there an existing library we could use?

pkg/sources/sources.go
mcastorina: Or if not here, then on `NewSourceConfig`

```go
source.Init(ctx, sources.NewConfig(conn,
    sources.WithName("foo"),
    sources.WithJobID(123),
    sources.WithVerify(true),
))
```

pkg/sources/sources.go
```diff
@@ -28,6 +29,71 @@ type Chunk struct {
...
+
+// WithSimpleCache sets the cache for the source.
+func WithSimpleCache() SourceConfigOption {
+	return func(c *SourceConfig) {
+		c.Cache = cache.Simple()
+	}
+}
+
+// SourceConfig provides a way to configure a Source.
+type SourceConfig struct {
```
mcastorina, line 65: **Nit:** Thoughts on renaming this `Config` since it's already in the `source` package?

PR #1091, opened 2023-02-10
[chore] - Remove monolithic config struct

ahrav:

### DIFF
--- a/main.go
+++ b/main.go
@@ -235,15 +235,14 @@ func run(state overseer.State) {
 			defer os.RemoveAll(repoPath)
 		}
 
-		g := func(c *sources.Config) {
-			c.RepoPath = repoPath
-			c.HeadRef = *gitScanBranch
-			c.BaseRef = *gitScanSinceCommit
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -288,35 +285,33 @@ func run(state overseer.State) {
...
 	case s3Scan.FullCommand():
-		s3 := func(c *sources.Config) {
-			c.Key = *s3ScanKey
-			c.Secret = *s3ScanSecret
-			c.Buckets = *s3ScanBuckets
+		cfg := sources.S3Config{
+			Key:       *s3ScanKey,
+			Secret:    *s3ScanSecret,
+			Buckets:   *s3ScanBuckets,
+			CloudCred: *s3ScanCloudEnv,
```
mcastorina, line 300: Looks like `CloudCred` wasn't being used before, good catch.

mcastorina (approved):

main.go
ahrav: oh, I wish I could take credit. My IDE did us a favor. 🤣

PR #1092, opened 2023-02-10
Remove duplicated detectors

trufflesteeeve:
In this case just Heroku and LinearAPI. But this includes the Moonclerck detector, which appears to be a typo that got turned into a separate detector type.


mcastorina (approved):
Nice!

dustin-decker (approved):

PR #1095, opened 2023-02-11
[chore] Remove logrus from trufflehog

mcastorina:

### DIFF
--- a/pkg/common/patterns.go
+++ b/pkg/common/patterns.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-
-	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/common/patterns.go
```diff
@@ -24,7 +22,7 @@ func BuildRegex(pattern string, specialChar string, length int) string {
 
 func BuildRegexJWT(firstRange, secondRange, thirdRange string) string {
 	if RangeValidation(firstRange) || RangeValidation(secondRange) || RangeValidation(thirdRange) {
-		log.Error("Min value should not be greater than or equal to max")
+		panic("Min value should not be greater than or equal to max")
```
ahrav, line 25: 💀

ahrav (approved):

pkg/common/patterns.go
mcastorina: Yeah, this seemed like an appropriate use of `panic` to me since this function is more of a convenience function for some detectors. Previously we just logged an error and built the invalid regex anyway, so I think it's fine?

Issue #1094, opened 2023-02-11
failed to create detectors config


sjgupta19:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version
trufflesecurity/trufflehog:latest

### Expected Behavior

Trufflehog should be running fine

### Actual Behavior

getting an error - failed to create detectors config

### Steps to Reproduce

run trufflehog:latest

## Additional Context
Looks like there's an update thats recently pushed which caused the problem. We started seeing this around that timeframe. and the error msg seems to be related in the latest merge.

arifay1:
based on the error message, seems related to the last commit of the release: https://github.com/trufflesecurity/trufflehog/commit/67784f69283dac04ebc813bbd67e2aa8c45279a5#diff-543d19670426e350a5479664ec75e0805333898df1593f323e02874df118ea52R67

bitsofinfo:
same

aldy505:
This also breaks the GitHub actions workflow if anyone wants to know:

```
Run trufflesecurity/trufflehog@main
  with:
    path: ./
    base: master
    head: HEAD
    extra_args: --debug --only-verified
[...]

PR #1098, opened 2023-02-11
Configure detectors

ahrav:
- Add tests to handle https://github.com/trufflesecurity/trufflehog/issues/1094 when constructing DetectorFilter


main.go
```diff
@@ -211,12 +213,31 @@ func run(state overseer.State) {
...
+	if err != nil {
+		logger.Error(err, "failed to create detectors")
+		os.Exit(1)
+	}
+
+	if len(dts) > 0 {
+		detectorsOption = engine.WithDetectors(!*noVerification, dts...)
+	} else {
+		detectorsOption = engine.WithDetectors(!*noVerification, engine.DefaultDetectors()...)
+	}
```
mcastorina, line 233: **Suggestion:** I think it might be easier to reason about and maintain if we have an option like `engine.WithFilterDetectors` that handles the filtering.

PR #1105, opened 2023-02-13
add more confluence options

dustin-decker:
Adds options to confluence for:
- opting into attachment scanning
- opting out of history scanning


proto/sources.proto
```diff
@@ -97,6 +97,8 @@ message Confluence {
   bool insecure_skip_verify_tls = 6;
   repeated string spaces = 7;
   repeated string ignore_spaces = 8;
+  bool include_attachments = 9;
```
trufflesteeeve, line 100: By default this will not include attachments. Do we want the default to include them or exclude them?

trufflesteeeve (approved):
Looks good, but I could also see the option being `ExcludeAttachments` so they get scanned by default.

proto/sources.proto
dustin-decker: This was intentional since it was a recent addition with major performance implications

dustin-decker: We'll be updating the docs soon to document all of the available options which will help discovery

PR #1106, opened 2023-02-14
Add ability to include and exclude detectors

mcastorina:

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,10 +21,12 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/config"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/decoders"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/engine"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/handlers"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/log"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/output"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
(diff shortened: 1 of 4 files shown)


0x1 (approved):
lgtm

main.go
```diff
@@ -208,11 +212,40 @@ func run(state overseer.State) {
 		handlers.SetArchiveMaxTimeout(*archiveTimeout)
 	}
 
+	// Build include and exclude detector filter sets.
+	var includeDetectorTypes, excludeDetectorTypes map[detectorspb.DetectorType]struct{}
+	{
+		includeList, err := config.ParseDetectors(*includeDetectors)
+		if err != nil {
+			// Exit if there was an error to inform the user of the misconfiguration.
+			logger.Error(err, "invalid detector configuration")
```
trufflesteeeve, line 221: __optional suggestion__: For ease of potentially figuring out later, it might make sense to change the error messages here to something like "invalid include list detector configuration" and "invalid exclude list detector configuration" to make it easier to know which was misconfigured.

main.go
```diff
@@ -208,11 +212,40 @@ func run(state overseer.State) {
...
+		excludeDetectorTypes = detectorTypeToSet(excludeList)
+	}
+	includeFilter := func(d detectors.Detector) bool {
+		_, ok := includeDetectorTypes[d.Type()]
+		return ok
+	}
+	excludeFilter := func(d detectors.Detector) bool {
+		_, ok := excludeDetectorTypes[d.Type()]
+		return !ok
+	}
```
trufflesteeeve, line 240: __optional suggestion__: It might be good to determine if there is any overlap in the exclude vs include and potentially error out, as what should happen in that case is somewhat unknown (though from what I can tell, overlaps would be excluded as excluded filters are applied last).

trufflesteeeve (approved):
LGTM. Just a couple of optional suggestions.

main.go
mcastorina: This is a good call out, but I think that might be more common than not with the way I constructed this.

By default the include list is `all` so any detector in the exclude list will trigger this case. It's subtractive by nature (we start with all possible detectors), then we filter out any detectors not in the include list, then we filter out any detectors in the exclude list.

mcastorina: I've updated the `--exclude-detectors` help text to explain it takes precedence.

PR #1108, opened 2023-02-15
[chore] - Add the unit for max archive size

ahrav:
- Include units to the max-archive-size CLI flag.

Screenshot 2023-02-15 at 9 44 40 AM


mcastorina (approved):
It might be worth noting somewhere that units can be used like `2MB`.

From the kingpin docs:

> Bytes parses numeric byte units. eg. 1.5KB

PR #1111, opened 2023-02-16
[chore] - Small cleanup

ahrav:
- Remove redundant parens.
- Remove select and directly iterate over channel.
- Add commentary to exported methods and struct fields.

### DIFF
--- a/pkg/handlers/handlers.go
+++ b/pkg/handlers/handlers.go
@@ -8,18 +8,28 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
 
+// DefaultHandlers returns all the default handlers.
 func DefaultHandlers() []Handler {
 	return []Handler{
 		&Archive{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/handlers/handlers.go
```diff
@@ -36,23 +47,15 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
...
+		chunk := *chunkSkel
+		chunk.Data = data
+
+		if common.IsDone(ctx) {
 			return false
 		}
+		// Send data on chunksChan.
+		chunksChan <- &chunk
 	}
+	return true
```
mcastorina, line 60: This refactored version operates slightly differently. Previously the function would return any time the context was canceled, but now it will only return if there is data to read and it's not blocked by the `chunksChan`.

Since either reading from `handlerChan` or writing to `chunksChan` can block, I'm not seeing an obvious way to simplify what we have.

Issue #1060, opened 2023-02-01
Tab character added to output filename


diogo-fernan:
## TruffleHog Version

```
$ trufflehog --version
trufflehog 3.26.0
```

## Trace Output

```
$ trufflehog git file://. --since-commit HEAD --json --no-verification --fail --trace
{"level":"debug","msg":"running version 3.26.0","time":"2023-02-01T15:00:10+01:00"}
{"level":"debug","msg":"running version 3.26.0","time":"2023-02-01T15:00:10+01:00"}
{"level":"debug","msg":"running with up to 1 workers","time":"2023-02-01T15:00:10+01:00"}
{"level":"debug","msg":"loaded 2 decoders","time":"2023-02-01T15:00:10+01:00"}
{"level":"debug","msg":"loaded 731 detectors total, 0 with verification enabled. 731 with verification disabled","time":"2023-02-01T15:00:10+01:00"}
2023-02-01T15:00:10+01:00       info-1  trufflehog      cloned repo     {"path": "."}
2023-02-01T15:00:10+01:00       info-1  trufflehog      scanning repo   {"repo": "https://github.com/<org>/<repo>.git", "base": "be9dac16b41dc3e2994a5a6e89828a1cfc12ca72", "head": ""}
2023-02-01T15:00:10+01:00       info-1  trufflehog      reached base commit     {"repo": "https://github.com/<org>/<repo>.git", "commit": "be9dac16b41dc3e2994a5a6e89828a1cfc12ca72"}
2023-02-01T15:00:10+01:00       info-1  trufflehog      scanning unstaged changes       {"path": "."}
2023-02-01T15:00:10+01:00       info-2  trufflehog      scanning unstaged changes from git      {"filename": "filename 2\t", "commit": "", "file": "filename 2\t"}
2023-02-01T15:00:10+01:00       info-1  trufflehog      scanning git repo complete      {"Repo": "https://github.com/<org>/<repo>.git", "path": ".", "time": 62379000}
[...]

PR #1113, opened 2023-02-16
Correctly parse most filenames with ' and '

bill-rich:


mcastorina (approved):
There was some interesting discussion on git filename format [here](https://github.com/trufflesecurity/trufflehog/issues/1060) too.

PR #1120, opened 2023-02-20
fix browserstack detector

raju-kamble:
The old regex is not detecting browserstack keys. They are coming in URI detector. 
The new regex covers all the old and new format of browserstack keys.


dustin-decker (approved):
Thank you!

PR #1127, opened 2023-02-24
Custom regex parallel verify

0x1:


pkg/custom_detectors/custom_detectors.go
```diff
@@ -82,66 +83,98 @@ func (c *customRegexWebhook) FromData(ctx context.Context, verify bool, data []b
 	// ]
 	matches := permutateMatches(regexMatches)
 
+	g := new(errgroup.Group)
```
ahrav, line 86: question: Are we okay not setting an upper bound on the number of goroutines here? Is the idea the len of matches will be low enough for this not to cause issues?

If we did plan on bounding the number of goroutines we can add:
`g.SetLimit(<some-limit>)`

pkg/custom_detectors/custom_detectors.go
```diff
@@ -82,66 +83,98 @@ func (c *customRegexWebhook) FromData(ctx context.Context, verify bool, data []b
...
+
 	// Create result object and test for verification.
+	resultsCh := make(chan detectors.Result, maxTotalMatches)
 	for _, match := range matches {
+		match := match
+		g.Go(func() error {
+			err := c.createResults(ctx, match, verify, resultsCh)
```
ahrav, line 93: nit: We can omit the err here and just return.
`return c.createResults(ctx, match, verify, resultsCh)`

pkg/custom_detectors/custom_detectors.go
```diff
@@ -82,66 +83,98 @@ func (c *customRegexWebhook) FromData(ctx context.Context, verify bool, data []b
...
+			return err
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	close(resultsCh)
```
ahrav, line 101: suggestion: I think we should move this up right after you created the chan and defer the close. Otherwise if we encounter an err in the goroutines w/in the errgroup we are going to leak with this unclosed channel.

so just: `defer close(resultsCh)` at L90

ahrav (approved):
Nice!

pkg/custom_detectors/custom_detectors.go
0x1: aaah! yes that's way cleaner!!!

pkg/custom_detectors/custom_detectors.go
0x1: @mcastorina thoughts?

mcastorina: That's what I was thinking. It's at most 100 goroutines that will be doing "async I/O" work.

pkg/custom_detectors/custom_detectors.go
mcastorina: Unfortunately we need to `close` the channel before iterating over it otherwise the loop on line 103 would never end.

I do think that maybe we can ignore the error returned by `g.Wait()` though, and collect as many results that we got.

Issue #1135, opened 2023-02-28
Debug switch enables profiler by default which can lead to considerable overhead with many workers


yilmi:
### TruffleHog Version
3.28.2

### Trace Output
{"level":"info-0","ts":"2023-02-23T01:04:04Z","logger":"trufflehog","msg":"starting pprof and fgprof server on :18066 /debug/pprof and /debug/fgprof"}

### Expected Behavior
Debug mode should just print debug level messages

The documentation from the [fgprof](https://github.com/felixge/fgprof) project warns about a considerable overhead when using many goroutines:
> The overhead of fgprof increases with the number of active goroutines (including those waiting on I/O, Channels, Locks, etc.) executed by your program. If your program typically has less than 1000 active goroutines, you shouldn't have much to worry about. However, at 10k or more goroutines fgprof might start to cause some noticeable overhead. 

### Actual Behavior

Debug mode sets the logging level AND instantiate the fgprof and pprof listeners

### Steps to Reproduce

 1. Run trufflehog with the `-debug` switch
 2. Notice the log message about the profiling server

### Environment
Any

PR #1136, opened 2023-02-28
Disable profiler in debug mode and add profile switch

yilmi:
This fixes issue #1135 

In its current state, trufflehog will configure a profiler fgprog and pprog using the [fgprof](https://github.com/felixge/fgprof) project.

This same project warns about a considerable overhead when using many goroutines:
> The overhead of fgprof increases with the number of active goroutines (including those waiting on I/O, Channels, Locks, etc.) executed by your program. If your program typically has less than 1000 active goroutines, you shouldn't have much to worry about. However, at 10k or more goroutines fgprof might start to cause some noticeable overhead. 

Changes:
- Separate the profiling mode from debug
- `--debug` is still a switch available and allows for debug logging
- `--profile` is  anew switch available and enables the profiling in an opt-in fashion

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ Flags:
       --help                     Show context-sensitive help (also try --help-long and --help-man).
       --debug                    Run in debug mode.
       --trace                    Run in trace mode.
+      --profile                  Enables profiling and sets a pprof and fgprof server on :18066.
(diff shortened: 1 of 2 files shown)


dustin-decker (approved):
Thank you!

Issue #1140, opened 2023-03-01
Add OpenAI API Token detector


yilmi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add OpenAI API Token detectors

## Problem to be Addressed
Missing the detection and verifications for OpenAI Tokens

## Description of the Preferred Solution
Match the pattern in a precise way, verify the token, and pull extraData for the first organizaton fetched

yilmi:
Preparing a PR

dustin-decker:
Closed by merging #1142

PR #1142, opened 2023-03-01
Add OpenAI API Tokens detector

yilmi:
This feature is related to #1140 

This change adds the detection of OpenAI API Tokens in their current form. It also verifies the token (using an endpoint used by the UI of https://platform.openai.com/ but isn't documented in https://platform.openai.com/docs/api-reference/).

The verification pulls extraData for the first organizaton fetched that can help the user understand which organization can be accessed by the token, how many organizations are accessible with the token,...

**IMPORTANT BEFORE MERGING**
Before merging, please add the following secrets
- OPENAI_UNVERIFIED with a value of `sk-xX1xX2xX3xX4xX5xX6xX7xX8xX9xX0xX1xX2xX3xX456xxxx`
- OPENAI_VERIFIED with a valid token
    - Update the wants part of the verified test to match what is returned and part of the extradata)

### DIFF
--- a/pkg/detectors/openai/openai.go
+++ b/pkg/detectors/openai/openai.go
@@ -1,0 +1,113 @@
+package openai
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strconv"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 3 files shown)


dustin-decker (approved):
Thank you for the contribution! I will make a follow up PR with our required test changes.

PR #1144, opened 2023-03-01
Keyword optimization

zricethezav:
### What this PR does

This PR adds a keyword optimization for string searching using the [Aho-Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm).

Rough benchmarks:

```
# FileSystem Scanning
## Trufflehog repo:
Old: ./trufflehog filesystem . --no-verification --no-update  57.32s user 1.92s system 138% cpu 42.702 total
New: ./trufflehog filesystem . --no-verification --no-update  23.75s user 1.20s system 143% cpu 17.366 total

## Rails repo:
Old: ./trufflehog filesystem ../sandbox/benchmark-repos/rails --no-verification   51.07s user 3.08s system 379% cpu 14.263 total
New: ./trufflehog filesystem ../sandbox/benchmark-repos/rails --no-verification   14.12s user 2.38s system 184% cpu 8.939 total

## GitLab repo:
Old: ./trufflehog filesystem ../sandbox/benchmark-repos/gitlab --no-verification   746.48s user 23.35s system 802% cpu 1:35.89 total
Peak footprint GB (from MacOS activity monitor): 9.1
New: ./trufflehog filesystem ../sandbox/benchmark-repos/gitlab --no-verification   191.10s user 24.91s system 366% cpu 58.878 total
Peak footprint GB (from MacOS activity monitor): 8.0

## GitLab repo with Concurrency limited to 1:
Old: ./trufflehog filesystem ../sandbox/benchmark-repos/gitlab --no-verification    638.34s user 32.47s system 106% cpu 10:32.06 total
Peak footprint GB (from MacOS activity monitor): 6.8
New: ./trufflehog filesystem ../sandbox/benchmark-repos/gitlab --no-verification    175.53s user 32.21s system 117% cpu 2:56.88 total
Peak footprint GB (from MacOS activity monitor): 8.0

---

# Git scanning:
## Trufflehog repo:
Old: ./trufflehog git https://github.com/trufflesecurity/trufflehog.git    28.04s user 1.34s system 365% cpu 8.029 total
New: ./trufflehog git https://github.com/trufflesecurity/trufflehog.git    9.94s user 1.48s system 165% cpu 6.911 total

## Gitleaks repo:
Old: ./trufflehog git https://github.com/zricethezav/gitleaks.git --no-verificatio  24.70s user 0.72s system 420% cpu 6.050 total
New: ./trufflehog git https://github.com/zricethezav/gitleaks.git --no-verificatio  6.62s user 0.67s system 138% cpu 5.284 total

```

Quick Summary: Looks like memory impact is slightly better in some cases and slightly worse in others. I'd consider it a wash (note that I was doing some rough sampling via activity monitor and not using `pprof` which allows for more granular metrics). CPU/Timing on the other hand is improved. The most dramatic improvement came from limiting concurrency to 1 on the GitLab filesystem scan which resulted in a 3x speedup. In all cases we see a reduction in `real` time.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -41,8 +41,6 @@ type Engine struct {
 	// prefilter is a ahocorasick struct used for doing efficient string
 	// matching given a set of words (keywords from the rules in the config)
 	prefilter ahocorasick.AhoCorasick
-
-	keywords []string
 }
 
 type EngineOption func(*Engine)
@@ -136,11 +134,12 @@ func Start(ctx context.Context, options ...EngineOption) *Engine {
 
 	// build ahocorasick prefilter for efficient string matching
 	// on keywords
+	keywords := []string{}
 	for _, d := range e.detectors[false] {
-		e.keywords = append(e.keywords, d.Keywords()...)
+		keywords = append(keywords, d.Keywords()...)
 	}
 	for _, d := range e.detectors[true] {
-		e.keywords = append(e.keywords, d.Keywords()...)
+		keywords = append(keywords, d.Keywords()...)
 	}
 	builder := ahocorasick.NewAhoCorasickBuilder(ahocorasick.Opts{
 		AsciiCaseInsensitive: true,
@@ -148,7 +147,7 @@ func Start(ctx context.Context, options ...EngineOption) *Engine {
 		MatchKind:            ahocorasick.LeftMostLongestMatch,
 		DFA:                  true,
 	})
-	e.prefilter = builder.Build(e.keywords)
+	e.prefilter = builder.Build(keywords)
 
 	ctx.Logger().V(2).Info("loaded decoders", "count", len(e.decoders))
 	ctx.Logger().V(2).Info("loaded detectors",
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -233,7 +233,6 @@ func (e *Engine) detectorWorker(ctx context.Context) {
 	for originalChunk := range e.chunks {
 		for chunk := range sources.Chunker(originalChunk) {
 
-			// init chunk keyword map
 			chunk.Keywords = make(map[string]bool)
 
 			atomic.AddUint64(&e.bytesScanned, uint64(len(chunk.Data)))
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -117,7 +117,6 @@ func Start(ctx context.Context, options ...EngineOption) *Engine {
 	}
 
 	// Set defaults.
-
 	if e.concurrency == 0 {
 		numCPU := runtime.NumCPU()
 		ctx.Logger().Info("No concurrency specified, defaulting to max", "cpu", numCPU)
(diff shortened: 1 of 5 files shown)


pkg/sources/sources.go
```diff
@@ -26,6 +26,8 @@ type Chunk struct {
 	Data []byte
 	// Verify specifies whether any secrets in the Chunk should be verified.
 	Verify bool
+	// Keywords found within the chunk's Data.
+	Keywords map[string]bool
```
dustin-decker, line 30: Why did you add this to the Chunk? Are you thinking that we might want to analyze which keywords get hit?

pkg/engine/engine.go
```diff
@@ -208,6 +232,9 @@ func (e *Engine) DetectorAvgTime() map[string][]time.Duration {
 func (e *Engine) detectorWorker(ctx context.Context) {
 	for originalChunk := range e.chunks {
 		for chunk := range sources.Chunker(originalChunk) {
+
+			chunk.Keywords = make(map[string]bool)
```
dustin-decker, line 236: I'm not sure if it makes as much differences these days, or for this use case, but using struct{} instead of bool has a measurable impact on CPU and memory performance: https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b

pkg/sources/sources.go
zricethezav: @dustin-decker whoops! When I was grokking the code at first I thought we would need `Keywords` as part of `Chunk` for the lookup but now I see we don't. Nice catch. I'll update this tomorrow when I'm on my laptop

dustin-decker (approved):
Huge improvement! Nice job.

dustin-decker (approved):

pkg/engine/engine.go
```diff
@@ -224,21 +246,29 @@ func (e *Engine) detectorWorker(ctx context.Context) {
 				if decoded == nil {
 					continue
 				}
+
 				dataLower := strings.ToLower(string(decoded.Data))
+				matches := e.prefilter.FindAll(dataLower)
```
mcastorina, line 251: **Question:** if the `prefilter` is doing a case-insensitive match, do we still need to lowercase the data here?

mcastorina: We could maybe use [strings.EqualFold](https://pkg.go.dev/strings#EqualFold) when comparing the matches against the keywords.

zricethezav: @mcastorina I believe we do still need to lowercase the data here since we are using `matchedKeywords` as a lookup which expect all keywords to be lowercase. 

> We could maybe use [strings.EqualFold](https://pkg.go.dev/strings#EqualFold) when comparing the matches against the keywords.

I may be misunderstanding but using `strings.EqualFold` would require iterating over all the matched keywords which would be linear time vs a constant time map lookup. Did you have an implementation specifically in mind?

`strings.EqualFold` is nice though, I'm sure we could introduce this in other parts of the code! Found a good article on string comparisons in go: https://hackernoon.com/optimizing-string-comparisons-in-go-7h1b3udm

mcastorina: Ah yeah, that makes sense. I didn't think it through fully, but you're right that `EqualFold` wouldn't work with a map lookup.

Perhaps only lower-casing the keys to the map would work though, instead of the entire chunk of data. That should function the same, but save on memory (and one less pass of the data).

PR #1153, opened 2023-03-04
Add gcs scanning integration

ahrav:
- Add source integration for scanning GCS
- There are multiple ways to provide auth to run the scan the recommend way is using ADC

Screenshot 2023-03-07 at 5 12 03 PM

Note: The notion of "resuming" is currently supported by the `gcsManager`, the logic still needs to be added to the gcs source. As this PR is already very large i'll add it in a follow up PR.


proto/credentials.proto
```diff
@@ -10,6 +10,8 @@ message SSHAuth {}
 
 message CloudEnvironment {}
 
+message ADC{}
```
dustin-decker, line 13: By default GCP SDKs will attempt to load credentials from ADC and from the metadata server (available in GCP runtime environments).  `CloudEnvironment` is used for S3 for the same.

ahrav: ah, that makes sense. I'll use that message instead. Thanks.

dustin-decker (approved):

PR #1154, opened 2023-03-04
Remove period from file extension

ahrav:
- filepath.Ext returns the file extension with a prepended ".". This was resulting in not skipping files that should have been skipped.


mcastorina (approved):
Good catch!

PR #1169, opened 2023-03-10
update integration test excludes

dustin-decker:
Fixes the GCS integration test after adding a public test bucket.


ahrav:
Thanks, i also had these changes in https://github.com/trufflesecurity/trufflehog/pull/1161

ahrav (approved):

PR #1187, opened 2023-03-17
[chore] - log enumeration duration

ahrav:

### DIFF
--- a/pkg/sources/gcs/gcs_manager.go
+++ b/pkg/sources/gcs/gcs_manager.go
@@ -383,6 +383,7 @@ func (g *gcsManager) enumerate(ctx context.Context, bkts []bucket) (*attributes,
 	logger.V(5).Info("enumerating buckets", "numBuckets", len(bkts))
 	stats := newStats(len(bkts))
 
+	now := time.Now()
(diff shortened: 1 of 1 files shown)


pkg/sources/gcs/gcs_manager.go
```diff
@@ -383,6 +383,7 @@ func (g *gcsManager) enumerate(ctx context.Context, bkts []bucket) (*attributes,
 	logger.V(5).Info("enumerating buckets", "numBuckets", len(bkts))
 	stats := newStats(len(bkts))
 
+	now := time.Now()
```
mcastorina, line 386: **Super nit:** If you want to get fancy you can do something like:

```go
defer func(start time.Time){
    logger.Info("finished enumerating buckets", "duration", time.Since(start), ... )
}(time.Now())
```

mcastorina (approved):

pkg/sources/gcs/gcs_manager.go
ahrav: I do. This is beautiful. and makes me very happy that it exists. Thanks.

PR #1189, opened 2023-03-19
Add in-memory caching pkg

ahrav:
- Define a cache interface along w/ an in-memory implementation.
- This will be used by the GCS source to track progress. It can also be used by any other source that needs progress tracking via a cache.
- Adding the interface abstraction allows us to implement other types of caches if needed.

caching library used -> https://pkg.go.dev/github.com/patrickmn/go-cache?utm_source=godoc


pkg/cache/memory/memory.go
```diff
@@ -0,0 +1,77 @@
...
+// Set adds a key-value pair to the cache.
+func (c *Cache) Set(key string, value bool) {
+	c.c.Set(key, value, defaultExpiration)
+}
+
+// Get returns the value for the given key.
+func (c *Cache) Get(key string) (string, bool) {
+	_, ok := c.c.Get(key)
+	return "", ok
+}
```
mcastorina, line 52: We're accidentally always returning `""` here. We probably want `return c.c.Get(key)`.

pkg/cache/memory/memory.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Contents returns all key-value pairs in the cache encodes as a string.
+func (c *Cache) Contents() string {
+	items := c.c.Items()
+	res := make([]string, 0, len(items))
+	for k := range items {
+		res = append(res, k)
+	}
+	return strings.Join(res, ",")
+}
```
mcastorina, line 77: **Question:** why encoded as a single string? Why not a `[]string`?

pkg/cache/memory/memory.go
```diff
@@ -0,0 +1,77 @@
...
+
+// New constructs a new in-memory cache.
+func New() *Cache {
+	c := cache.New(expirationInterval, purgeInterval)
+	return &Cache{c: c}
+}
+
+// NewWithData constructs a new in-memory cache with existing data.
+func NewWithData(ctx context.Context, s string) *Cache {
+	data := strings.Split(s, ",")
```
mcastorina, line 31: **Suggestion:** This interface is a bit too lossy; I think it would be easier to reason about if `NewWithData` took a varargs for the items instead of an encoded string. As an example (and we may never run into this but..) how would one initialize a cache with a comma in the key?

pkg/cache/memory/memory_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if ok {
+		t.Fatalf("Unexpected value for key1 after delete: %v, %v", v, ok)
+	}
+
+	// Test clear.
+	c.Clear()
+	v, ok = c.Get("key10")
+	if ok || v != "" {
+		t.Fatalf("Unexpected value for key10 after clear: %v, %v", v, ok)
+	}
```
mcastorina, line 41: `key10` never appears to be set in this test, so I think this is testing the `Clear` of an empty cache, which may not have been the intention.

mcastorina (changes requested):

pkg/cache/memory/memory.go
ahrav: This is because all the sources that use EncodeResume need that value to be a string. So I figured i'd just return the string.

pkg/cache/memory/memory.go
mcastorina: Yeah, or have another method called `Contains` which returns just the `bool`.

pkg/cache/memory/memory.go
mcastorina: Makes sense, but maybe a method rename would make it clearer? `EncodedContents` maybe?

pkg/cache/memory/memory.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Contents returns all key-value pairs in the cache encodes as a string.
+func (c *Cache) Contents() string {
+	items := c.c.Items()
+	res := make([]string, 0, len(items))
+	for k := range items {
+		res = append(res, k)
+	}
+	return strings.Join(res, ",")
+}
```
mcastorina, line 85: **Suggestion (optional):** If you do want to add convenience methods for actually serializing/deserializing the data, we could use [strings.Replacer](https://pkg.go.dev/strings#Replacer) to escape commas (assuming it has an acceptable benchmark).

For example:
```go
func unescape(s string) string {
	replacer := strings.NewReplacer(`\,`, ",")
	return replacer.Replace(s)
}

func escape(s string) string {
	replacer := strings.NewReplacer(",", `\,`)
	return replacer.Replace(s)
}
```

mcastorina (approved):
Very nice!

PR #1190, opened 2023-03-19
Delete progress tracking from GCS source

ahrav:
- Delete progress tracking
- Progress tracking work is redone -> https://github.com/trufflesecurity/trufflehog/pull/1204


pkg/sources/gcs/gcs.go
```diff
@@ -34,27 +57,58 @@ type Source struct {
...
+func (c *cacheManager) set(key string) {
+	c.cache.Set(key, key)
+}
+
+func (c *cacheManager) shouldPersist() (bool, string) {
+	if !c.shouldCache {
+		return false, ""
+	}
+
+	if c.cache.Count()%c.threshold != 0 {
```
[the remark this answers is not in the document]
ahrav: I figured it would be an easy way to determine if the number of items in the cache is a multiple of our threshold. If it is we can then persist the data in the cache to the DB within the `Progress`'s `EncodeResumeInfo` field.

I'd love to get your thoughts if you had an alternative approach in mind.

pkg/sources/gcs/gcs.go
```diff
@@ -147,142 +152,6 @@ func setGCSManagerOptions(include, exclude []string, includeFn, excludeFn func([
...
-		sort.Strings(processing)
-		info.lastProcessedObject = processing[0]
-		bucketOffset[objName] = info
-	}
-
-	return bucketOffset, nil
-}
-
 // enumerate all the objects and buckets in the source and use the results to
 // set the progress information. This will be used track progression of the scan,
```
rosecodym, line 156: There's no more progress information, right? Does that make this comment out of date now?

ahrav: ah, yep. Updated it to match the new behavior in this PR. Thanks.

rosecodym (approved):
this is pushing the size limit of what i feel comfortable approving, so you might want to get more eyeballs on it, but from what i understand lgtm

PR #1202, opened 2023-03-22
Support for exclude globs at the `git log` level

zricethezav:
### What does this PR do?

Adds a `git` scan mode option to exclude globs at the `git log` level. This is useful if a repo has known files/directories that should be ignored by scans like `/vendor` for Go projects, or `/node_modules ` for node projects.

try it yourself:

```
# This should only report false positives from `go.mod`

git clone -b feature-exclude https://github.com/trufflesecurity/trufflehog.git
cd trufflehog
go build && ./trufflehog  git --no-verification --exclude-globs='**/*.proto,**/*.go' file://$PWD
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,7 +60,7 @@ var (
 	gitScanURI          = gitScan.Arg("uri", "Git repository URL. https://, file://, or ssh:// schema expected.").Required().String()
 	gitScanIncludePaths = gitScan.Flag("include-paths", "Path to file with newline separated regexes for files to include in scan.").Short('i').String()
 	gitScanExcludePaths = gitScan.Flag("exclude-paths", "Path to file with newline separated regexes for files to exclude in scan.").Short('x').String()
-	gitScanExcludeGlobs = gitScan.Flag("exclude-globs", "Comma separated list of globs to exclude in scan.").String()
+	gitScanExcludeGlobs = gitScan.Flag("exclude-globs", "Comma separated list of globs to exclude in scan. This option filters at the `git log` level, resulting in faster scans.").String()
 	gitScanSinceCommit  = gitScan.Flag("since-commit", "Commit to start scan from.").String()
 	gitScanBranch       = gitScan.Flag("branch", "Branch to scan.").String()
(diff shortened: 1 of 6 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -126,9 +126,13 @@ func (c *Parser) RepoPath(ctx context.Context, source string, head string, abbre
 	} else {
 		args = append(args, "--all")
 	}
+	if len(excludedGlobs) != 0 {
```
dustin-decker, line 129: Nit: This length check if unnecessary, the inner iteration over the slice will do the same thing.

main.go
```diff
@@ -60,6 +60,7 @@ var (
 	gitScanURI          = gitScan.Arg("uri", "Git repository URL. https://, file://, or ssh:// schema expected.").Required().String()
 	gitScanIncludePaths = gitScan.Flag("include-paths", "Path to file with newline separated regexes for files to include in scan.").Short('i').String()
 	gitScanExcludePaths = gitScan.Flag("exclude-paths", "Path to file with newline separated regexes for files to exclude in scan.").Short('x').String()
```
dustin-decker, line 62: Do you think that we should deprecate the exclude path regex files longer term? It seems less economic to me than globbing, which is more prevalent among CLI tools that deal with paths and files.

main.go
```diff
@@ -60,6 +60,7 @@ var (
...
+	gitScanExcludeGlobs = gitScan.Flag("exclude-globs", "Comma separated list of globs to exclude in scan.").String()
```
dustin-decker, line 63: Is it worth noting that this option is more performant than the exclude paths option?

dustin-decker (approved):
Approved with some questions and a nit

pkg/gitparse/gitparse.go
zricethezav: whoops, that's a relic. Good catch

main.go
zricethezav: That's a good call

PR #1204, opened 2023-03-23
Use persistable cache for GCS progress tracking

ahrav:
- Simplify progress tracking for the GCS source by levering a cache for objects scanned.
This change no longer supports fine grained enumeration resuming, I think this is a fair tradeoff for the amount of complexity that is removed from the source. On average enumeration takes about ~18s for ~1.3M objects, therefore enumerating on each run is fine.
- This changes how we save progress data within EncodeResumeInfo from objects being map[string]struct{} to just a string. The string is the object name and is saved in an in-memory cache during the scan job to facilitate resuming scans. Note: there will be a follow up PR that uses the md5 hash of the contents vs the object name in the even the same named object is found in multiple buckets.


rosecodym (commented):
is this intended to be merged after #1190?

ahrav:
> is this intended to be merged after #1190?

Yea, I have this rebased off #1190 , i should've left this in Draft until that one was merged.

rosecodym (commented):
**question**: do you feel it's worth adding a test to validate that persistence doesn't occur before the interval has been hit?

ahrav:
> **question**: do you feel it's worth adding a test to validate that persistence doesn't occur before the interval has been hit?

Yea, I like that idea. I'll add a test for it. Thanks for the suggestion.

pkg/sources/gcs/gcs_test.go
```diff
@@ -289,3 +294,57 @@ func TestSourceChunks_ListObjects_Error(t *testing.T) {
...
+	progress := "object0,object1,object2,object3,object4"
+	objs := strings.Split(progress, ",")
+
+	encodeResume := strings.Split(source.Progress.EncodedResumeInfo, ",")
+	sort.Strings(encodeResume)
+
+	assert.Equal(t, objs, encodeResume)
+	assert.Equal(t, int32(5), source.Progress.SectionsCompleted)
+	assert.Equal(t, int64(100), source.Progress.PercentComplete)
+	assert.Equal(t, 0, source.cache.Count())
```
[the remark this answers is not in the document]
rosecodym: Ah, I see it now. I asked "why is the cache cleared" but I probably should have asked "why _was_ the cache cleared," not because it _shouldn't_ be cleared, but because once the run is complete it feels like it shouldn't exist at all! But it has to exist post-run, because it's stored in `Source`, because that's how you persist information from `Init` through to `Chunks`. That implementation constraint forces the cache to be present here, so we might as well manage its state, and verify our management. Got it.

But! As a **suggestion**, could you set the cache up inside `Chunks` instead of `Init`? (It looks like you can as a matter of simply moving the code around, but I'm not familiar with implicit assumptions about those functions, so maybe there's a reason it won't work.) If so, you don't need to store the cache (or its mutex or extra count variable) inside `Source`, so you don't have to worry about clearing or checking it after the run.

pkg/sources/gcs/gcs.go
```diff
@@ -52,7 +57,44 @@ type Source struct {
 	log        logr.Logger
 	chunksCh   chan *sources.Chunk
 
-	sources.Progress
+	processedObjects int32
```
rosecodym, line 60: **question**: What does this do that `Progress.SectionsCompleted` doesn't?

ahrav: I think we can likely use `Progress.SectionsCompleted` here. I was using processedObjects when I had the initial implementation of progress tracking. We no longer need it. Nice rec, thx.

pkg/sources/gcs/gcs_test.go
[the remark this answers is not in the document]
ahrav: I guess one reason why the cache might make sense to still be a part of the source and get initialized in Init would be when we transition to distributed jobs. We would have the job split across instances and they would all need to reference the same cache for progress tracking.

That being said, I might be thinking too far in advance, and we can cross that bridge when we get there.

pkg/sources/gcs/gcs.go
```diff
@@ -225,13 +273,46 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
...
+func (s *Source) setProgress(ctx context.Context, objName string, cache cache.Cache) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	ctx.Logger().V(5).Info("setting progress for object", "object-name", objName)
+	s.SectionsCompleted++
+
+	cache.Set(objName, objName)
+	s.Progress.SectionsRemaining = int32(s.stats.numObjects)
+	s.Progress.PercentComplete = int64(float64(s.SectionsCompleted) / float64(s.stats.numObjects) * 100)
```
rosecodym, line 307: i know this isn't part of this pr so i'm just asking because i'm curious but why does `ProgressComplete` need 64 bits to represent a percentage?

rosecodym (approved):
nice work on this!

pkg/sources/gcs/gcs.go
ahrav: It likely doesn't. I think that was the type when the struct was first created a long time ago and we haven't thought twice about it. I can create a separate PR to update that type along with the related code.

ahrav:
> nice work on this!

Thanks for all the feedback :)

PR #1205, opened 2023-03-26
[chore] - no progress tracking for Jobs w/ ID 0

ahrav:
- We do not need the overhead of a cache or progress tracking for non-stateful runs. This would include all runs using the CLI.


zricethezav:
@ahrav are we still considering this?

ahrav:
> @ahrav are we still considering this?

Nope. Closing now.

PR #1211, opened 2023-03-27
[bug] - Use correct date format for Date posted

ahrav:
Screenshot 2023-03-27 at 7 10 08 AM

### DIFF
--- a/pkg/sources/gcs/gcs_test.go
+++ b/pkg/sources/gcs/gcs_test.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"sort"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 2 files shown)


dustin-decker (approved):

pkg/sources/gcs/gcs_test.go
```diff
@@ -331,9 +334,10 @@ func TestSourceChunks_ListObjects(t *testing.T) {
 		return got[i].SourceMetadata.GetGcs().Filename < got[j].SourceMetadata.GetGcs().Filename
 	})
 
-	for _, c := range got {
-		assert.Equal(t, c.SourceMetadata.GetGcs().Filename, c.SourceMetadata.GetGcs().Filename)
-		assert.Equal(t, c.Data, c.Data)
+	for i, c := range got {
+		assert.Equal(t, want[i].SourceMetadata.GetGcs().Filename, c.SourceMetadata.GetGcs().Filename)
+		assert.Equal(t, want[i].Data, c.Data)
+		assert.Equal(t, want[i].SourceMetadata.GetGcs().CreatedAt, c.SourceMetadata.GetGcs().CreatedAt)
```
rosecodym, line 340: **question**: was this test failing to actually test anything useful before?

ahrav: It didn't seem like it. After I updated the test to use the `want` values, I commented out the part of the code it was supposed to test to make sure it was failing.

I then added it back just to sanity check myself.

I think we missed this during the initial review, likely due to how large that previous PR was. :/

PR #1213, opened 2023-03-28
lint on all branches to catch warnings earlier

zricethezav:
This PR adds linting to all branches. Imo the best time to catch lint warnings would be on developmental branches.


zricethezav:
Looks like the

PR #1215, opened 2023-03-28
add CLI switch to actions config

codevbus:
Adds CLI switch to the action config, should make scan results show up in a PR diff


codevbus:
Confirmed this is working with https://github.com/trufflesecurity/trufflehog-github-actions-example/pull/1

dustin-decker (approved):

PR #1219, opened 2023-03-28
revert to original entrypoint config

codevbus:


dustin-decker (approved):

petegallagher:
Why was saving the results to the GITHUB_OUTPUT variable removed? Can this be reinstated?

This is a really useful feature and I'd like to be able to access the results in a subsequent GitHub Action step.

PR #1220, opened 2023-03-28
Use OAuth2 http client with GCS

ahrav:
- Provide a way to create an OAuth HTTP client that can be use by the Google Cloud Storage library.

### DIFF
--- a/pkg/sources/gcs/gcs.go
+++ b/pkg/sources/gcs/gcs.go
@@ -113,11 +113,11 @@ func configureGCSManager(aCtx context.Context, conn *sourcespb.GCS, concurrency 
 	case *sourcespb.GCS_Unauthenticated:
 		gcsManagerAuthOption = withoutAuthentication()
 	case *sourcespb.GCS_Oauth:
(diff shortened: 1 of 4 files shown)


pkg/sources/gcs/gcs.go
```diff
@@ -107,6 +112,12 @@ func configureGCSManager(aCtx context.Context, conn *sourcespb.GCS, concurrency
 		gcsManagerAuthOption = withDefaultADC(aCtx)
 	case *sourcespb.GCS_Unauthenticated:
 		gcsManagerAuthOption = withoutAuthentication()
+	case *sourcespb.GCS_Oauth:
+		oauth2Client, err := oauth2Client(aCtx, conn.GetOauth())
```
bill-rich, line 116: nit: Having the variable and function name the same could cause confusion.

bill-rich (approved):
Looks good other than one nit on naming.

pkg/sources/gcs/gcs.go
ahrav: ooof, great call.

PR #1235, opened 2023-04-03
fix linting step

dustin-decker:


mcastorina (approved):
Thanks for adding the `exportloopref` here too, I missed that when adding it to the CI

PR #1240, opened 2023-04-05
THOG-920/add oss proto

zubairk14:
Adding support for SharePoint in scanning with source and source_metadata protos for SharePoint.


proto/source_metadata.proto
```diff
@@ -224,6 +224,15 @@ message PublicEventMonitoring {
   }
 }
 
+message SharePoint {
+  string path = 1;
```
dustin-decker, line 228: `link`: URL to take the user to the result

proto/source_metadata.proto
```diff
@@ -224,6 +224,15 @@ message PublicEventMonitoring {
...
+  string timestamp = 2;
```
dustin-decker, line 229: Unix time. time.Time.Unix() (seconds since epoch, a number formatted as a string)

proto/source_metadata.proto
```diff
@@ -224,6 +224,15 @@ message PublicEventMonitoring {
...
+  string author = 3;
```
dustin-decker, line 230: Try to add an `email` field as well

proto/sources.proto
```diff
@@ -270,3 +271,14 @@ message SlackRealtime {
     credentials.SlackTokens tokens = 1;
   }
 }
+
+message Sharepoint {
+  string endpoint = 1 [(validate.rules).string.uri_ref = true];
+  oneof credential {
+    string token = 2;
```
dustin-decker, line 278: refresh_token

ahrav: I actually don't think you need this field at all. refresh_token should be include as part of `credentials.Oauth2`
https://github.com/trufflesecurity/trufflehog/blob/b3d917f9c71789f67e71bed6c3e106a6cbd6d012/proto/credentials.proto#L36-L41

proto/sources.proto
```diff
@@ -270,3 +271,14 @@ message SlackRealtime {
...
+    credentials.ClientCredentials authenticated = 3;
```
dustin-decker, line 279: remove this, and remove ClientCredentials from the credentials proto because it won't be used

proto/sources.proto
```diff
@@ -270,3 +271,14 @@ message SlackRealtime {
...
+    credentials.Oauth2 oauth = 4;
```
dustin-decker, line 280: remove this, we don't need to provide a client secret or ID, just the refresh token.

proto/sources.proto
```diff
@@ -270,3 +271,14 @@ message SlackRealtime {
...
+  }
+  string site_url = 5;
```
dustin-decker, line 282: Not sure if we need this but we can leave it for now. We should enumerate if possible

proto/sources.proto
zubairk14: @ahrav it looks like you are using ClientCredentials as well - are we good to remove? cc: @dustin-decker 

https://github.com/trufflesecurity/trufflehog/blob/b3d917f9c71789f67e71bed6c3e106a6cbd6d012/proto/sources.proto#L260

proto/sources.proto
```diff
@@ -40,6 +40,7 @@ enum SourceType {
   SOURCE_TYPE_SYSLOG = 25;
   SOURCE_TYPE_PUBLIC_EVENT_MONITORING = 26;
   SOURCE_TYPE_SLACK_REALTIME = 27;
+  SOURCE_TYPE_SHAREPOINT = 29;
```
dustin-decker, line 43: Should this be 28?

proto/sources.proto
dustin-decker: Okay, leave it if it's being used elsewhere

proto/sources.proto
zubairk14: 28 is already taken

https://github.com/trufflesecurity/trufflehog/blob/f107e1b49792afcc7c5fe5f034814388fe544ed3/proto/sources.proto#L43

proto/sources.proto
```diff
@@ -213,7 +210,6 @@ message S3 {
     credentials.KeySecret access_key = 1;
     credentials.Unauthenticated unauthenticated = 2;
     credentials.CloudEnvironment cloud_environment = 4;
-    credentials.AWSSessionTokenSecret session_token = 5;
```
dustin-decker, line 216: Why was this removed? The AWS test is failing now

zubairk14: this must have been an accident, i don't recall touching these

zubairk14: i think when i ran the generate protos command it deleted these, restoring now.

dustin-decker (approved):

PR #1249, opened 2023-04-10
Revert "THOG-920/add oss proto"

zubairk14:
Reverts trufflesecurity/trufflehog#1240 because of issues with generated proto files


dustin-decker:
Closed by #1250

PR #1257, opened 2023-04-14
use md5 hash for checking if key exists

ahrav:


mcastorina (approved):

rosecodym (commented):
Is there a reason the test suite doesn't cover this? It seems like an obvious candidate.

PR #1265, opened 2023-04-17
Only add detectors once

bill-rich:
`engine.CustomDetectors()` includes `engine.DefaultDetectors()`

@ahrav Can you take a look to confirm both are not needed?

### DIFF
--- a/main.go
+++ b/main.go
@@ -285,7 +285,6 @@ func run(state overseer.State) {
 	e := engine.Start(ctx,
 		engine.WithConcurrency(*concurrency),
 		engine.WithDecoders(decoders.DefaultDecoders()...),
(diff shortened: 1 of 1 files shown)


mcastorina (approved):
Good catch! What do you think of adding a runtime check in the engine to log that multiple of the same detectors were configured? I'm working on some detector improvements and can make a separate PR for it.

bill-rich:
> Good catch! What do you think of adding a runtime check in the engine to log that multiple of the same detectors were configured? I'm working on some detector improvements and can make a separate PR for it.

I think that is a good call!

mcastorina:
https://github.com/trufflesecurity/trufflehog/pull/1266

PR #1267, opened 2023-04-18
Fix include and exclude detector logic

mcastorina:


mcastorina:
The map was `{detector type} -> {detector type, detector version}` so you couldn't include / exclude a type of two different versions.

pkg/engine/defaults_test.go
```diff
@@ -0,0 +1,50 @@
+package engine
+
+import (
+	"testing"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+func TestDefaultDetectorsHaveUniqueVersions(t *testing.T) {
```
rosecodym, line 10: If we expect all of our default detectors to have unique versions, does this mean that the new logic you added to accommodate different versions of the same detector is intended to apply to non-default detectors?

mcastorina: > If we expect all of our default detectors to have unique versions

I think the name of this test might be misleading. It's checking that for all our default detectors, if there are multiple of the same type, they have unique versions.

As an example, we have two `github` detectors `github.v1` and `github.v2`. If we accidentally made both `v1`, this test would fail.

I don't expect to ever encounter this error since it's rare to have multiple versions, but it's an assumption I have that I wanted to express in code.

mcastorina: I say that before checking the failing tests 🤣 

```
--- FAIL: TestDefaultDetectorsHaveUniqueVersions (0.00s)
    defaults_test.go:30: detector "Gitlab" has duplicate version: 1
    defaults_test.go:30: detector "Github" has duplicate version: 1
    defaults_test.go:30: detector "NpmToken" has duplicate version: 2
```

PR #1269, opened 2023-04-18
Normalize GitHub repos during enumeration

ahrav:
Normalize the repos during enumeration and track the repos,orgs,members using a local cache. This reduces all the slice operations we are doing when populating the orgs, members, and repos.
Preliminary findings:

5000 repos:
BenchmarkEnumerate-8 31 36288521 ns/op 12297210 B/op 55556 allocs/op (OLD)
BenchmarkEnumerate-8 154 7789454 ns/op 11822166 B/op 55523 allocs/op (NEW)

10000 repos:
Screenshot 2023-04-18 at 4 25 32 PM

Address how repos and forks are counted. We used to count repos and forks even if they were ignored, update the logic to increment the counts if we actually include them as part of the scan.


pkg/sources/github/github.go
```diff
@@ -365,16 +472,16 @@ func (s *Source) enumerateWithToken(ctx context.Context, apiEndpoint, token stri
...
 			logger := s.log.WithValues("org", org)
-			if err := s.addRepos(ctx, org, s.getReposByOrg); err != nil {
+			if err := s.getReposByOrg(ctx, org); err != nil {
 				logger.Error(err, "error fetching repos by org")
 			}
 
-			if err := s.addRepos(ctx, ghUser.GetLogin(), s.getReposByUser); err != nil {
-				logger.Error(err, "error fetching repos for user", "user", ghUser.GetLogin())
+			if err := s.getReposByUser(ctx, ghUser.GetLogin()); err != nil {
+				logger.Error(err, "error fetching gists by org")
```
rosecodym, line 480: **issue**: this error message was erroneously changed

pkg/sources/github/github.go
```diff
@@ -577,11 +646,12 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
 	}
 
 	_ = s.jobPool.Wait()
-	if len(scanErrs) == 0 {
-		s.SetProgressComplete(len(s.repos), len(s.repos), "Completed Github scan", "")
+	if scanErrs.Count() > 0 {
+		s.log.V(2).Info("encountered errors while scanning", "count", scanErrs.Count(), "errors", scanErrs)
 	}
+	s.SetProgressComplete(len(s.repos), len(s.repos), "Completed Github scan", "")
```
rosecodym, line 652: was it intentional to change the progress completion behavior of this function? before, completion would only be set if there were no errors, but now completion is set every time.

pkg/sources/github/github.go
```diff
@@ -679,56 +748,42 @@ func (s *Source) getReposByOrg(ctx context.Context, org string) ([]string, error
...
-				continue
-			}
-
-			numRepos++
 			if r.GetFork() {
-				numForks++
 				if !s.conn.IncludeForks {
 					continue
 				}
+				numForks++
```
rosecodym, line 763: is this change to the way forks are counted intentional? (i have the same question about the similar change to the way repos are counted.) the new way _looks_ more correct, but it wasn't called out in your pr description.

pkg/sources/github/github.go
```diff
@@ -128,6 +130,59 @@ func (s *Source) UserAndToken(ctx context.Context, installationClient *github.Cl
...
+		return
+	}
+	c.Cache.Set(key, val)
+}
+
+func (c *filteredRepoCache) ignoreRepo(s string) bool {
+	for _, exclude := range c.exclude {
+		g, err := glob.Compile(exclude)
+		if err != nil {
+			continue
```
rosecodym, line 160: the old code logged glob pattern compilation errors, but this new version silently drops them. are they being logged elsewhere somehow?

pkg/sources/github/github.go
```diff
@@ -742,72 +797,37 @@ func (s *Source) getReposByUser(ctx context.Context, user string) ([]string, err
...
-			return true
-		}
-	}
-	return false
-}
-
-func (s *Source) ignoreRepo(r string) bool {
-	logger := s.log.WithValues("repo", r)
-	for _, ignore := range s.ignoreRepos {
-		g, err := glob.Compile(ignore)
```
rosecodym, line 796: does this recompile every glob pattern every time we find a repo?? that seems really inefficient!

pkg/sources/github/github.go
```diff
@@ -679,56 +748,42 @@ func (s *Source) getReposByOrg(ctx context.Context, org string) ([]string, error
...
-			numRepos++
 			if r.GetFork() {
-				numForks++
 				if !s.conn.IncludeForks {
 					continue
 				}
+				numForks++
+			}
+			repo, err := s.normalizeRepo(r.GetCloneURL())
+			if err != nil {
```
rosecodym, line 766: what could cause repo url normalization failures? are we ok with skipping repos when it happens?

pkg/sources/github/github.go
```diff
@@ -942,8 +962,7 @@ func (s *Source) addAllVisibleOrgs(ctx context.Context) {
 			} else {
 				continue
 			}
-			s.log.V(2).Info("adding organization for repository enumeration", "id", org.ID, "org", name)
-			common.AddStringSliceItem(name, &s.orgs)
```
rosecodym, line 946: what replaced this deleted call?

pkg/sources/github/github.go
```diff
@@ -970,13 +989,9 @@ func (s *Source) addOrgsByUser(ctx context.Context, user string) {
...
 		logger.V(2).Info("Listed orgs", "page", orgOpts.Page, "last_page", resp.LastPage)
 		for _, org := range orgs {
-			var name string
-			if org.Login != nil {
-				name = *org.Login
-			} else {
+			if org.Login == nil {
 				continue
 			}
-			common.AddStringSliceItem(name, &s.orgs)
```
rosecodym, line 979: what replaced this deleted call?

rosecodym (changes requested):
i'm requesting changes specifically for a changed log message (comment inline), but i asked several other questions about things i'm not sure of

pkg/sources/github/github.go
```diff
@@ -259,24 +324,66 @@ func (s *Source) visibilityOf(repoURL string) (visibility source_metadatapb.Visi
...
+	return s.scan(ctx, installationClient, chunksChan)
+}
+
+func (s *Source) enumerate(ctx context.Context, apiEndpoint string) (*github.Client, error) {
+	var installationClient *github.Client
+	var err error
+
+	switch cred := s.conn.GetCredential().(type) {
+	case *sourcespb.GitHub_Unauthenticated:
+		s.enumerateUnauthenticated(ctx)
```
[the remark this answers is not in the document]
ahrav: Ah yea, it looks like it's because all the other cases return the error. `enumerateUnauthenticated` doesn't return an error (it actually returns void) which is why this one doesn't have a return.

pkg/sources/github/github.go
ahrav: Yea, I updated it to match all our other sources. There is usually a chance there is at least 1 error during the scan. If we don't set the progress to complete the FE will interpret it as 'the scan is still running". We've had discussions about this in the past and resorted to the less strict approach. Ideally we have maybe some threshold for errors, that when it gets crossed we consider the source in an unhealthy state. We currently just log most errors and complete the scan if possible.

pkg/sources/github/github.go
ahrav: Ah yea, good callout. I added this to the PR description as I also think the previous way of counting was incorrect.

pkg/sources/github/github.go
ahrav: This was actually the main reason I did this refactor. We were setting the orgs and repos using this `common.AddStringsliceItem` function in order to keep track of repos and orgs. It was wildly inefficient given a number of reasons.
1. Since it was created as a slice we were always checking the slice before adding a new item to the slice. For some large customers we were doing a check to see if we already added the repo or org for each new repo or org. This was replaced by using the cache instead of this slice.
2. Since the slice didn't have a cap, we had to keep growing the slice for large customers and moving old entries into the new slice.
3. This was just confusing, we were doing slice ptr operations that I felt we could get rid of completely by instead using a map where we get all the benefits without all this extra overhead.

pkg/sources/github/github.go
ahrav: same as above, it was replaced with the cache.

pkg/sources/github/github.go
ahrav: Yep, not good. Updated it to online compile on init.

pkg/sources/github/github.go
ahrav: They should be logged now during init.

pkg/sources/github/github.go
```diff
@@ -964,13 +978,9 @@ func (s *Source) addOrgsByUser(ctx context.Context, user string) {
...
 		for _, org := range orgs {
-			var name string
-			if org.Login != nil {
-				name = *org.Login
-			} else {
+			if org.Login == nil {
 				continue
 			}
-			common.AddStringSliceItem(name, &s.orgs)
 		}
```
mcastorina, line 984: It looks like this loop essentially does nothing now. Can it be deleted? Or should it be doing something?

pkg/sources/github/github.go
```diff
@@ -128,6 +130,67 @@ func (s *Source) UserAndToken(ctx context.Context, installationClient *github.Cl
...
+// include and exclude globs.
+func (c *filteredRepoCache) Set(key, val string) {
+	if c.ignoreRepo(key) {
+		return
+	}
+	if !c.includeRepo(key) {
+		return
+	}
+	c.Cache.Set(key, val)
+}
```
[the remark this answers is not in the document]
ahrav: i'm actually not sure on precedence for this source. I left the same logic as we had before. I'm a little hesitant to change it given I don't want to possibly break existing clients. I think we should, as you have alluded to, standardize this logic across all sources. At which point i'd be more comfortable changing any of the existing logic.

pkg/cache/memory/memory_test.go
```diff
@@ -46,12 +46,27 @@ func TestCache(t *testing.T) {
...
+		t.Fatalf("Unexpected keys: %v", k)
+	}
 
+	// Test getting only the values.
+	vals := c.Values()
+	sort.Strings(keys)
+	sort.Strings(vals)
+	if !cmp.Equal(keys, vals) {
+		t.Fatalf("Unexpected values: %v", vals)
+	}
```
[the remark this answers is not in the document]
ahrav: i'd agree.

pkg/sources/github/github.go
ahrav: agreed, I don't think it does anything anymore. NIce find.

ahrav: Oh, wait this means we need an orgCache similar to our repos. I'll get that fixed.

pkg/sources/github/github_test.go
```diff
@@ -201,8 +249,6 @@ func TestAddOrgsByUser(t *testing.T) {
 
 	s := initTestSource(nil)
 	s.addOrgsByUser(context.TODO(), "super-secret-user")
-	assert.Equal(t, 1, len(s.orgs))
-	assert.Equal(t, []string{"sso2"}, s.orgs)
```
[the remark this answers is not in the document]
ahrav: fixed.

mcastorina (approved):

ahrav:
> i'm requesting changes specifically for a changed log message (comment inline), but i asked several other questions about things i'm not sure of

Updated the log and addressed the other comments as well.

rosecodym (approved):
are you going to rebase this on top of #1318 once that one is merged?

PR #1274, opened 2023-04-20
Add utf16 decoder

ahrav:
Screenshot 2023-04-20 at 1 01 45 PM

### DIFF
--- a/pkg/decoders/decoders.go
+++ b/pkg/decoders/decoders.go
@@ -8,6 +8,7 @@ func DefaultDecoders() []Decoder {
 	return []Decoder{
 		&UTF8{},
 		&Base64{},
+		&UTF16{},
 	}
 }
 
--- a/pkg/decoders/utf16.go
+++ b/pkg/decoders/utf16.go
@@ -66,9 +66,9 @@ func guessUTF16Endianness(b []byte) (binary.ByteOrder, error) {
 
 	if evenNullBytes > oddNullBytes {
 		return binary.LittleEndian, nil
-	} else if oddNullBytes > evenNullBytes {
-		return binary.BigEndian, nil
-	} else {
-		return nil, fmt.Errorf("could not determine endianness")
 	}
+	if oddNullBytes > evenNullBytes {
+		return binary.BigEndian, nil
+	}
+	return nil, fmt.Errorf("could not determine endianness")
 }
--- a/pkg/decoders/utf16.go
+++ b/pkg/decoders/utf16.go
@@ -1,0 +1,48 @@
+package decoders
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"unicode/utf16"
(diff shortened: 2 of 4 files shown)


dustin-decker:
Is there a way to better identify UTF16 content? I added this test case, which failed and got a chunk with data:

```go	
                 {
			name:      "Invalid UTF-16 input (it's UTF-8)",
			input:     []byte("Hello World!"),
			expected:  nil,
			expectNil: true,
		},
```
		
I'm concerned that this could increase the data we have to scan by ~50% (assuming 50% chance of even length data)

dustin-decker:
I think we could use this to determine if it's UTF-16BE or UTF-16LE and do the appropriate decoding (binary.LittleEndiantUint16 or binary.BigEndian.Uint16) or nothing based on the first two bytes:

```
// isUTF16 determines if the input byte slice is UTF-16 LE or UTF-16 BE, returning the encoding type.
func isUTF16(b []byte) string {
	if len(b) < 2 {
		return ""
	}
	bom := binary.BigEndian.Uint16(b)
	if bom == 0xFEFF {
		return "UTF-16BE"
	} else if bom == 0xFFFE {
		return "UTF-16LE"
	}
	return ""
}
```

dustin-decker:
I think what I posted won't be compatible with the sliding-window-with-overlap-chunker since later chunks will be missing the BOM  :|

ahrav:
> I think what I posted won't be compatible with the sliding-window-with-overlap-chunker since later chunks will be missing the BOM :|

Could we take an educated guess?
If we look for specific byte patterns that are more likely to appear in one endianness than the other. For example, ASCII characters in UTF-16 are represented with a null byte in either the high or low byte. If you see more null bytes in even positions than odd positions, you might guess that the input is UTF-16LE. Conversely, if there are more null bytes in odd positions, you might guess that it's UTF-16BE. Or is this too wish washy :/

ahrav:
Maybe something like:
```
func guessUTF16Endianness(b []byte) (binary.ByteOrder, error) {
	if len(b) < 2 || len(b)%2 != 0 {
		return nil, errors.New("input length must be even and greater than 2")
	}

	var evenNullBytes, oddNullBytes int

	for i := 0; i < len(b); i += 2 {
		if b[i] == 0 {
			evenNullBytes++
		}
		if b[i+1] == 0 {
			oddNullBytes++
		}
	}

	if evenNullBytes > oddNullBytes {
		return binary.LittleEndian, nil
	} else if oddNullBytes > evenNullBytes {
		return binary.BigEndian, nil
	} else {
		return nil, errors.New("unable to guess endianness")
	}
}
```

ahrav:
nvm, that won't work that well :(

ahrav:
One more small optimization:
Screenshot 2023-04-20 at 2 48 39 PM

dustin-decker (approved):
Nice improvement!

PR #1275, opened 2023-04-20
optimize utf-8 decoder

ahrav:
**small optimization to the utf8 decoder:**
- use the bytes.Buffer type to build the output byte slice instead of creating multiple smaller slices and joining them together at the end.
- replace the bytes.FieldsFunc call with a single loop over the input bytes, appending valid runes to the field slice, and writing it to the buffer when a non-valid rune is encountered or the end of the input is reached.

Screenshot 2023-04-20 at 3 05 37 PM


dustin-decker:
I think this looks good, but could you add a test to confirm? Also, is there any way to avoid the string conversion of the entire chunk?

ahrav:
> I think this looks good, but could you add a test to confirm? Also, is there any way to avoid the string conversion of the entire chunk?

I believe we already have a handful of tests along w/ the benchmark already there: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/decoders/utf8_test.go

ahrav:
> I think this looks good, but could you add a test to confirm? Also, is there any way to avoid the string conversion of the entire chunk?

ahhh nice catch. Yep that should save us another alloc.

ahrav:
Screenshot 2023-04-20 at 3 39 53 PM

dustin-decker (approved):
Very nice!

pkg/decoders/utf8.go
```diff
@@ -26,19 +26,30 @@ func (d *UTF8) FromChunk(chunk *sources.Chunk) *sources.Chunk {
...
+	buf := &bytes.Buffer{}
+	for i, c := range b {
+		if isValidByte(c) {
+			field[fieldLen] = c
+			fieldLen++
+		} else {
+			if fieldLen > 5 {
+				buf.Write(field[:fieldLen])
+			}
+			fieldLen = 0
```
rgmz, line 46: What is the significance of `fieldLen > 5`?

rgmz (commented):
> ...appending valid runes to the field slice, and writing it to the buffer when a non-valid rune is encountered...

@ahrav this seems to be errant behaviour per #3585 and #3223. I think the proper solution would be to append [␚](https://en.wikipedia.org/wiki/Substitute_character) or [�](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character).

PR #1278, opened 2023-04-22
Small optimizations for the base64 decoder

ahrav:
Additional optimizations to the base64 decoder.:
- Remove the `bytes.Contains` and replace with a constant time lookup using a b64 charset mapping using a fixed array.
- Instead of using a bytes.Buffer to store the characters of a substring, track the start and end positions of a substring in `data` during `getSubstringsOfCharacterSet`.
- Handle chunk Data replacement in a single pass.

Screenshot 2023-04-22 at 1 14 19 PM

Screenshot 2023-04-22 at 1 14 31 PM

Screenshot 2023-04-22 at 12 57 18 PM

### DIFF
--- a/pkg/decoders/base64.go
+++ b/pkg/decoders/base64.go
@@ -15,7 +15,6 @@ var (
 )
 
 func (d *Base64) FromChunk(chunk *sources.Chunk) *sources.Chunk {
-
 	encodedSubstrings := getSubstringsOfCharacterSet(chunk.Data, 20)
 	decodedSubstrings := make(map[string][]byte)
 
--- a/pkg/decoders/base64.go
+++ b/pkg/decoders/base64.go
@@ -15,8 +15,9 @@ var (
 )
 
 func (d *Base64) FromChunk(chunk *sources.Chunk) *sources.Chunk {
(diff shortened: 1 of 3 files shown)


pkg/decoders/base64.go
```diff
@@ -26,51 +26,88 @@ func (d *Base64) FromChunk(chunk *sources.Chunk) *sources.Chunk {
...
 	if len(data) == 0 {
 		return nil
 	}
 
+	// Given characters are mostly ASCII, we can use a simple array to map.
+	var b64CharsetMapping [128]bool
+	// Build an array of all the characters in the base64 charset.
+	for _, char := range b64Charset {
+		b64CharsetMapping[char] = true
+	}
```
dustin-decker, line 61: I'm not sure if it will make much difference, but this could be done once with a package var populated with an init() function, which automatically gets called once when the package is imported.

dustin-decker (approved):
Nice! One comment on a potential improvement.

pkg/decoders/base64.go
ahrav: I think that makes sense. Updated.

PR #1284, opened 2023-04-24
Added a new detector for percy.io

shabbirbs:
Added a new detector for percy.io

Application - https://percy.io/

Testing - Tested locally with credentials in .env for positive and negative cases


ahrav (approved):
LGTM!

Thanks so much for the contribution. We really appreciate it.

Issue #144, opened 2019-01-24
Base64 chars missing URL-safe variant


FrugalGuy:
In addition to the standard 64 BASE64 alphabet characters, you should add '_' and '-'.
These are frequently used in place of '+' and '/' to create values that would be safe passed in a URL.
I see secrets encoded with this variant frequently.  I propose you just add the two additional characters.

dustin-decker:
Closed by #1292

PR #1292, opened 2023-04-26
WIP: Add Base64URLSafe decoder

nyanshak:
* Add decoder that can decode base64 strings with '_' and '-' instead of of '+' and '/'.

Addresses #144.

WIP but looking for early feedback:
* should this functionality be merged into the existing base64 decoder or split into a separate Base64URLSafe decoder? There is a lot of shared code and wanted to see if maintainers have preferences towards making separate decoders (like UTF8/UTF16 decoders) or a single `Base64` decoder before polishing it up, wiring into all the places (detectorspb, decoders, engine, detectors.proto, etc.

### DIFF
--- a/pkg/decoders/base64.go
+++ b/pkg/decoders/base64.go
@@ -7,13 +7,20 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
 
-type Base64 struct{}
+type (
+	Base64        struct{}
+	Base64URLSafe struct{}
+)
 
 var (
 	b64Charset  = []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=")
 	b64EndChars = "+/="
 	// Given characters are mostly ASCII, we can use a simple array to map.
 	b64CharsetMapping [128]bool
+
+	b64URLCharset        = []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_=")
+	b64URLEndChars       = "-_="
+	b64URLCharsetMapping [128]bool
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


dustin-decker:
If it's possible to combine the two types of b64 correctly that would be preferred, because there is a decent amount of overhead introduced with each decoder. The UTF8/16 ones are different enough to be separate.

nyanshak:
Okay @dustin-decker - I combined them. I think there's a possible caveat, which is mostly that **combining** them adds potentially a little overhead in attempting to decode strings that would not be valid b64-encoded strings.

For example, this is an invalid b64 encoded string: `abcdef-ghijkl_mnop+qrs/tuvwxyz`. In the separate decoder version, we wouldn't waste time trying to decode this, because once we hit a character outside of the character set, we'd try to decode the substring if it was longer than the threshold, but we wouldn't try to decode the entire string.

If the combined version is good enough, then I think this is ready to go. Otherwise, may need to revert the last commit, then wire up the separate B64URLSafe decoder :woman_shrugging:

pkg/decoders/base64.go
```diff
@@ -24,11 +26,19 @@ func init() {
...
 
 func (d *Base64) FromChunk(chunk *sources.Chunk) *sources.Chunk {
-	encodedSubstrings := getSubstringsOfCharacterSet(chunk.Data, 20)
+	encodedSubstrings := getSubstringsOfCharacterSet(chunk.Data, 20, b64CharsetMapping, b64EndChars)
 	decodedSubstrings := make(map[string][]byte)
 
 	for _, str := range encodedSubstrings {
 		dec, err := base64.StdEncoding.DecodeString(str)
+		if err == nil {
+			if len(dec) > 0 {
```
rosecodym, line 35: this is pretty nitty, but breaking this conjunction into two separate `if` statements caused me to take a second to realize that this is the same logic as is being run on line 42. is it possible to write the two checks the same way?

nyanshak:
```go
	for _, str := range encodedSubstrings {
		dec, err := base64.StdEncoding.DecodeString(str)
		if err == nil {
			if len(dec) > 0 {
				decodedSubstrings[str] = dec
			}
			continue
		}

		dec, err = base64.RawURLEncoding.DecodeString(str)
		if err == nil && len(dec) > 0 {
			decodedSubstrings[str] = dec
		}
	}
```

In context - the first check is intentionally different, because if we successfully decoded with `base64.StdEncoding`, then regardless of whether the string length was 0 or not, we want to `continue` instead of also trying to decode with `base64.RawURLEncoding`. This saves the extra effort of attempting to do the url-safe decoding if we already successfully decoded.

To make these checks the _same_ is possible but either results in:
1. not continuing on the first check, resulting in extra work
2. OR making the second check have an extra nested if block:
```go
		dec, err = base64.RawURLEncoding.DecodeString(str)
		if err == nil {
			if len(dec) > 0 {
				decodedSubstrings[str] = dec
			}
		}
```
But I think both ways of doing this are not ideal.

Do you think we can keep as is or suggest an alternative?

rosecodym: I'm sorry, you're right, I just misread. Thanks for the clarification!

PR #1293, opened 2023-04-26
add scripts to benchmark and plot performance across tags

dustin-decker:
Produces data and a plot for user time spent scanning a given repo for the last N tags, averaged over 5 runs.


hack/bench/versions.sh
```diff
@@ -0,0 +1,83 @@
...
+count=0
+
+
+# Loop over tags and checkout each one in turn, up to the specified number of versions
+for tag in $tags
+do
+  if [ $count -eq $num_versions ]
+  then
+    break
+  fi
```
mcastorina, line 36: **Nit (optional):** double bracket / formatting

```bash
if [[ $count -eq $num_versions ]]; then
    break
fi
```

hack/bench/versions.sh
```diff
@@ -0,0 +1,83 @@
...
+    tmpfile=$(mktemp)
+    /usr/bin/time -o $tmpfile trufflehog git "file://$repo_tmp" --no-verification --no-update >/dev/null 2>&1
+    time_output=$(cat $tmpfile)
+    rm $tmpfile
+
+    # Extract the user time from the output
+    user_time=$(echo $time_output | awk '{print $3}')
+
+    # Add the user time to the sum
+    user_time_sum=$(echo "$user_time_sum + $user_time" | bc)
```
mcastorina, line 72: **Question:** Are you using `bc` because it supports larger numbers than `$(( ))`?

mcastorina (approved):
Nice! It would be cool to have this in CI at some point to catch perf regressions before they get merged.

dustin-decker:
> Nice! It would be cool to have this in CI at some point to catch perf regressions before they get merged.

I think Zach is looking at doing that. Something like comparing the last two tags and exiting 1 if there is a >10% perf decrease.

hack/bench/versions.sh
dustin-decker: Using bc because GPT4 told me to...

non136079 (approved):

Issue #1237, opened 2023-04-05
Customizable object Size for S3 Bucket Scanning


DivyanshuDiwakar:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

## Problem to be Addressed

Right now the maximum object size for a S3 bucket which we can scan is 250 MB. Now suppose, what if someone wants to scan objects having size lesser than 50 MB (not greater than that), so it would be better if we would be able to customize that size using flags like ``` --max-object-size```.

nyanshak:
@dustin-decker I think this issue can be closed since the PR has been merged.

PR #1294, opened 2023-04-26
Add max object size flag for s3 bucket scanning

nyanshak:
Addresses #1237 by adding a `--max-object-size` flag to the s3 scan command. Modeled after the GCS max object size code.


dustin-decker (approved):

dustin-decker:
Thank you!

PR #1296, opened 2023-04-27
Adding --verifier flag, and a method to parse the input.

Machiry:
Please create an issue to collect feedback prior to feature additions. Please also reference that issue in any PRs.
If possible try to keep PRs scoped to one feature, and add tests for new feat


Machiry:
Was a mistake.

PR #1301, opened 2023-04-28
add performance test

zricethezav:
This PR adds a quick performance test to check for any regressions. Currently the threshold is set to fail at a 10% increase in runtime. 

See https://github.com/trufflesecurity/trufflehog/actions/runs/4832402508/jobs/8611119050 for example failure
Screenshot 2023-04-28 at 10 53 12 AM


zricethezav:
@mcastorina @dustin-decker following up on https://github.com/trufflesecurity/trufflehog/pull/1293. I was running into some trouble with the git env so I ended up splitting the script into a few different jobs

dustin-decker (approved):

mcastorina (approved):
lgtm, but had one thought:

Do we want to allow this step to fail so it doesn't block any PRs (at least at first)? Or are we confident enough that it's a stable and consistent test to not be optional?

zricethezav:
@mcastorina I believe we're still able to merge even with a failing test; see `Scan for secrets` failures

mcastorina:
Nice! Ah, I was confusing [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) with required workflows.

PR #1314, opened 2023-05-01
double timeout to reduce test failures stemming from timeouts

zubairk14:
test failures occuring intermittently due to timeout, doubling timeout to reduce failed CI/CD jobs from timeout issues.

### DIFF
--- a/pkg/sources/test_helpers.go
+++ b/pkg/sources/test_helpers.go
@@ -22,8 +22,8 @@ func HandleTestChannel(chunksCh chan *Chunk, cf ChunkFunc) error {
 				return err
 			}
 			return nil
-		case <-time.After(10 * time.Second):
-			return fmt.Errorf("no new chunks received after 10 seconds")
+		case <-time.After(20 * time.Second):
+			return fmt.Errorf("no new chunks received after 20 seconds")
 		}
 	}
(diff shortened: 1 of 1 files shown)


dustin-decker (approved):

dustin-decker:
Can we change this to accept timeout as a parameter? Most sources don't need more than 10 seconds.

pkg/sources/test_helpers.go
```diff
@@ -10,7 +10,7 @@ type ChunkFunc func(chunk *Chunk) error
 
 var MatchError = errors.New("chunk doesn't match")
 
-func HandleTestChannel(chunksCh chan *Chunk, cf ChunkFunc) error {
+func HandleTestChannel(chunksCh chan *Chunk, cf ChunkFunc, timeout int) error {
```
mcastorina, line 13: **Nit (optional):** What you have here is good and I think it can be slightly improved by making `timeout` a `time.Duration`. That forces the user to include the unit at the callsite, which I think is clearer, and also allows for different units (milliseconds maybe?) if wanted.

zubairk14:
fixed underlying issue with https://github.com/trufflesecurity/thog/pull/1029/files

PR #1318, opened 2023-05-02
Extend cache interface

ahrav:
- Github enumeration will require getting either only the keys or only the values at different parts of the source integration. Extend the interface to allow for returning only the keys or values from the cache.

### DIFF
--- a/pkg/cache/memory/memory_test.go
+++ b/pkg/cache/memory/memory_test.go
@@ -46,12 +46,28 @@ func TestCache(t *testing.T) {
 		t.Fatalf("Unexpected value for key10 after clear: %v, %v", v, ok)
 	}
 
-	// Test contents.
+	// Test getting only the keys.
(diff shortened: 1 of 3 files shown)


pkg/cache/memory/memory_test.go
```diff
@@ -46,12 +46,28 @@ func TestCache(t *testing.T) {
...
+		t.Fatalf("Unexpected keys: %v", k)
+	}
+
+	// Test getting only the values.
+	vals := c.Values()
+	sort.Strings(keys)
+	sort.Strings(vals)
+	if !cmp.Equal(keys, vals) {
+		t.Fatalf("Unexpected values: %v", vals)
+	}
```
mcastorina, line 68: **Nit:** I would prefer if this compared values that differed from the keys.

mcastorina (approved):

PR #1320, opened 2023-05-02
Do extraction after decompression

### LINKED ISSUE #1319: Failure to handle certain archive formats
### TruffleHog Version
commit `67972683eaf`

### Expected Behavior

Trufflehog correctly extracts / decompresses, then finds the secret in the archive.

### Actual Behavior

Trufflehog reports `Error unarchiving chunk` and fails to find the secret in the file.

### Steps to Reproduce

Create a folder with secret as test case:
```
ls -la min/
-rw-r--r--  1 user user    2 May  2 11:22 innocuous.txt
-rw-r--r--  1 user user  151 May  2 11:22 secret.txt
```

Make strange archive file (admittedly weird, but I was trying to make this break):

```
zip -r min min
gzip min.zip
mkdir minfail
mv min.zip.gz minfail
```

Run the test:

```
trufflehog filesystem --trace minfail
```

nyanshak:
* Fix #1319 - Fix error where some files do not get properly scanned due to order of extraction / decompression steps. Doing decompression first ensures that a compressed archive (e.g., gzipped zip file), is handled correctly.

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/mholt/archiver/v4"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	logContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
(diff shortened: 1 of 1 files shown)


dustin-decker (approved):
Thank you. Closes #1319

PR #1323, opened 2023-05-02
Output git timestamps as UTC times

### LINKED ISSUE #1312: Unify timestamps
### Description

Currently, results return with the timezone offset of the user who made the commit, like so:

```
Timestamp: 2022-04-05 14:45:07 +0530
...
Timestamp: 2020-06-25 09:48:11 -0700
```

It'd be nice to have an adjusted timestamp field that goes along with this that puts all of them in UTC:

```
Timestamp:     2022-04-05 14:45:07 +0530
UTC Timestamp: 2022-04-05 09:15:07
...
Timestamp:     2020-06-25 09:48:11 -0700
UTC Timestamp: 2020-06-25 16:48:11
```

nyanshak:
Fix #1312 by using UTC timestamps for git result outputs.


mcastorina (approved):
lgtm, thanks!

PR #1324, opened 2023-05-02
[oc-313] - Add GitHub metrics

ahrav:
- Add metric information to our logs around how large each github repo is, total size of all repos, and how long each repo takes to scan.
- Refactored out a lot of the repo related logic into its own file and created an interface for repo listing. We had 3 functions that were almost identical except for the resource we were listing. This resulted in a lot of duplicated code, so I figured i'd clean that up in order to make my life a bit easier when adding in these changes. All the logic around counts and metrics should be isolated only within the `processRepo` fxn vs within the 4 previous functions.


pkg/sources/github/repo.go
```diff
@@ -0,0 +1,232 @@
...
+
+type appListOptions struct {
+	github.ListOptions
+}
+
+func (a *appListOptions) getListOptions() *github.ListOptions {
+	return &a.ListOptions
+}
+
+func (s *Source) addReposByApp(ctx context.Context) error {
```
rosecodym, line 118: this method name starts with `add` but the other two very similar ones start with `get`. is there a reason for this? if not, i find the discrepancy pretty confusing.

rosecodym (commented):
i noticed one potential naming issue and i have some general questions:
* some of the code you moved has TODOs in it, and in particular a `context.TODO()`. is that ok?
* i feel like i've seen more `var` blocks (with multiple declarations) here than i've seen elsewhere. do we have a particular style we prefer?

ahrav:
> i noticed one potential naming issue and i have some general questions:
> 
> * some of the code you moved has TODOs in it, and in particular a `context.TODO()`. is that ok?
> * i feel like i've seen more `var` blocks (with multiple declarations) here than i've seen elsewhere. do we have a particular style we prefer?

The context.TODO() I think can be updated (i just did that), the other's i'm not entirely sure. It might be worth doing a pass  and trying to clean up or remove any TODOs in general. As for the vars, i don't think we have a best practice. If the vars are global (hopefully not a lot) i usually use a single var with parens to group them all together. If they are local I try to put them close to where they are used to help with spatial locality. If the fxn isn't too large and it makes more sense to group them I guess I might also do that. Are you referring to that? or how we declare them:
eg:
```
var foo int
var bar int
```
vs
```
var (
    foo int
    bar int
)
```

pkg/sources/github/repo.go
ahrav: I think this was the original name, but you are correct, this should bet `get`. Thanks.

pkg/sources/github/github_test.go
```diff
@@ -226,7 +226,7 @@ func TestAddReposByApp(t *testing.T) {
 		})
 
 	s := initTestSource(nil)
-	err := s.addReposByApp(context.TODO())
+	err := s.getReposByApp(context.TODO())
```
rosecodym, line 229: here's another `context.TODO()` - did you intend to clean this one up too?

rosecodym (approved):
found another `context.TODO()` but otherwise lgtm

PR #1325, opened 2023-05-02
Add log to track git log size

ahrav:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -254,6 +254,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 				if totalSize > c.maxCommitSize {
 					oldCommit := currentCommit
 					commitChan <- *currentCommit
+					totalLogSize += currentCommit.Size
 					currentCommit = &Commit{
 						Hash:    currentCommit.Hash,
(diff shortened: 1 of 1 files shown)


bill-rich:
The concept looks good, but the `totalLogSize` should also be added to [here](https://github.com/trufflesecurity/trufflehog/blob/1f06a42f32b563c4acc9eaf26b973a15b236e50d/pkg/gitparse/gitparse.go#L256)  since that is what handled especially large commits. You could also add it [here](https://github.com/trufflesecurity/trufflehog/blob/1f06a42f32b563c4acc9eaf26b973a15b236e50d/pkg/gitparse/gitparse.go#L327), but it only deals with the last commit.

ahrav:
> The concept looks good, but the `totalLogSize` should also be added to [here](https://github.com/trufflesecurity/trufflehog/blob/1f06a42f32b563c4acc9eaf26b973a15b236e50d/pkg/gitparse/gitparse.go#L256) since that is what handled especially large commits. You could also add it [here](https://github.com/trufflesecurity/trufflehog/blob/1f06a42f32b563c4acc9eaf26b973a15b236e50d/pkg/gitparse/gitparse.go#L327), but it only deals with the last commit.

Ohhh, nice. I completely missed those.

bill-rich (approved):
Looks good!

Issue #994, opened 2023-01-04
Missing logs and requests timedout not retried


yilmi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version
This can be reproduced using the latest version of TruffleHog, at the time of writing this bug, version 3.21.0

### Trace Output

Current logging of the scan time doesn't include enough information
```bash
info-1    trufflehog    scanning git repo complete    {"path": "/tmp/trufflehog971756979", "time": 18557687886}
```

### Expected Behavior

When running in debug mode, TruffleHog should provide full details with regards to what is happening, including logging http requests.

When logging the time it took to scan a repo, the URL of the repo should be provided as well to help identify repositories that take a lot of time and perhaps exclude them. 

Finally, by default, the GitHub API client doesn't use a `RetryableHttpClient()`, which could help increase requests successful completion when running massive scans

### Actual Behavior

When running trufflehog against a very high number of GitHub repos and at a high concurrency level, we may end up having some cpu, network or storage attrition, which can in turn lead to failed requests. 

TruffleHog doesn't log the HTTP requests that are performed nor their response status, this can make troubleshooting more difficult as it could be cause by multiple reasons, and logging the HTTP requests would sometimes help identify some of them. 

Moreover, network issues can happen sometimes, and a timeout or network error can happen anytime while fetching the list of repos, checking a repo's visibility or cloning it, with the current GitHub API client, this would result in a failure, while this is recoverable.

This may be help with Issue #976 

Finally, we do log the scan time, but we don't provide the repo URL

### Steps to Reproduce

1. Run trufflehog on a GitHub org with a large number of repos (10K)
2. Check the log output

## Environment
 * OS: Linux 8CPU/64GB
 * Version [e.g. 22]

PR #1337, opened 2023-05-10
GitHub basic auth

dustin-decker:
Adds a basic auth credential option to the github source


ahrav (approved):
Do we also need to update our Enterprise docs?

mcastorina (approved):

dustin-decker:
> Do we also need to update our Enterprise docs?

No, this is not intended to be used by end users. It'll be used by thog #994

PR #1345, opened 2023-05-18
Make OpenAI regex more specific

nyanshak:
* OpenAI keys are in the format 'sk-' + 20 alphanumeric chars + the magic string 'OpenAI' base64-encoded + 20 alphanumeric chars.

### DIFF
--- a/pkg/detectors/openai/openai.go
+++ b/pkg/detectors/openai/openai.go
@@ -20,7 +20,8 @@ type Scanner struct{}
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b((?:sk)-[a-zA-Z0-9]{48})\b`)
+	// The magic string T3BlbkFJ is the base64-encoded string: OpenAI
+	keyPat = regexp.MustCompile(`\b((?:sk)-[a-zA-Z0-9]{20}T3BlbkFJ[a-zA-Z-09]{20})\b`)
 )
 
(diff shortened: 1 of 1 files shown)


ahrav (approved):
Nice! Thanks for the improvement here Brendan 😃

pkg/detectors/openai/openai.go
```diff
@@ -20,7 +20,8 @@ type Scanner struct{}
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b((?:sk)-[a-zA-Z0-9]{48})\b`)
+	// The magic string T3BlbkFJ is the base64-encoded string: OpenAI
+	keyPat = regexp.MustCompile(`\b((?:sk)-[a-zA-Z0-9]{20}T3BlbkFJ[a-zA-Z-09]{20})\b`)
```
ahrav, line 24: suggestion: Looks like a typo in the last character class, it should be `[a-zA-Z-0-9]` looks like we are missing a `-`

question: Is there a benefit in keeping the initial non-capture group? Could we remove it and get the same effect.
eg: `\b(sk-[a-zA-Z0-9]{20}T3BlbkFJ[a-zA-Z-0-9]{20})\b`

ahrav (changes requested):

dustin-decker:
The keyword on L45 could be changed to `T3BlbkFJ`, which will be a bit more specific and prevent the regex code path from running as often.

PR #1346, opened 2023-05-18
Check to see if StructuredData exists before attempting to print it

trufflesteeeve:
When `r.Result.StructuredData` is nil, attempting to print parts of it will panic. This fixes that by ensuring it exists before attempting to print those parts.


mcastorina (approved):

trufflesteeeve:
Merging even though performance test didn't pass, because it is using a broken version.

PR #1348, opened 2023-05-18
needed perms for running workflows against forks

codevbus:


mcastorina (approved):

mcastorina:
> Is this test expected to fail? - https://github.com/trufflesecurity/trufflehog/actions/runs/5025916930/jobs/9013512614?pr=1348

No, it looks like it's failing from a 500 error. I'll try rerunning it.

PR #1368, opened 2023-05-29
[chore] - update Float detector regex

ahrav:


rosecodym:
What prompted this change? And are there really no associated tests to add or update?

0x1 (approved):

ahrav:
> What prompted this change? And are there really no associated tests to add or update?

I got a message from someone mentioning this detector was firing often with unverified results. I ran the tests and they were failing. I also went into our Grafana dash to see how many verified results we had using the old regex and there were 0. There were on the other hand 41553 unverified results. I updated the regex and now all the tests actually pass.

PR #1369, opened 2023-05-30
Update sqlserver redaction, deduplication, and URI redaction

dustin-decker:
SQLServer wasn't being properly deduplicated and URI wasn't being properly redacted in some cases

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -2,6 +2,7 @@ package detectors
 
 import (
 	"context"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -152,4 +153,10 @@ func MustGetBenchmarkData() map[string][]byte {
 		"medium": medium,
 		"big":    big,
 	}
+}
+
+func RedactURL(u *url.URL) string {
+	urlSafe := *u
+	urlSafe.User = url.UserPassword(urlSafe.User.Username(), "********")
+	return strings.TrimSpace(strings.Replace(urlSafe.String(), "%2A", "*", -1))
 }
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -155,8 +155,7 @@ func MustGetBenchmarkData() map[string][]byte {
 	}
 }
 
-func RedactURL(u *url.URL) string {
-	urlSafe := *u
-	urlSafe.User = url.UserPassword(urlSafe.User.Username(), "********")
-	return strings.TrimSpace(strings.Replace(urlSafe.String(), "%2A", "*", -1))
+func RedactURL(u url.URL) string {
+	u.User = url.UserPassword(u.User.Username(), "********")
+	return strings.TrimSpace(strings.Replace(u.String(), "%2A", "*", -1))
 }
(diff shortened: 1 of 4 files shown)


pkg/detectors/sqlserver/sqlserver.go
```diff
@@ -31,22 +31,24 @@ func (s Scanner) Keywords() []string {
...
-		if params.Password == "" {
+		if paramsUnsafe.Password == "" {
 			continue
 		}
 
 		detected := detectors.Result{
 			DetectorType: detectorspb.DetectorType_SQLServer,
-			Raw:          []byte(params.Password),
+			Raw:          []byte(paramsUnsafe.Password),
+			RawV2:        []byte(paramsUnsafe.URL().String()),
```
ahrav, line 46: question: Will this potentially lead to duplicate results similar to to the changes we made earlier with the other detectors. Since `Raw` and `RawV2` are used as part of the lookup during deduplication we would create a new secret with potentially the same Raw value.

ahrav (approved):

pkg/detectors/detectors.go
```diff
@@ -153,3 +154,9 @@ func MustGetBenchmarkData() map[string][]byte {
 		"big":    big,
 	}
 }
+
+func RedactURL(u *url.URL) string {
```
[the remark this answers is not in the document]
dustin-decker: Good point, but I think I should probably just require non pointer instead of making a copy within

pkg/detectors/sqlserver/sqlserver.go
dustin-decker: Possibly. I need to sync with @trufflesteeeve  on this since the whole point of introducing RawV2 was to start properly deduping detectors that had an incorrect raw field.

trufflesteeeve: It's been a while since that was added. I'll have to read the code to see what should be happening. We did see a secret duplication issue when RawV2 was added before, though.

trufflesteeeve: So I think I understand where the hashv2 duplication came from. We've always been adding the hashv2 to the secrets - https://github.com/trufflesecurity/thog/blob/46fe1c8d6d5b13737c603e20bafded703fc066c9/scanner/pkg/secrets/secret.go#L81

But if there was no rawv2, then that's just a hash of empty string. But it still gives it a hashv2 value. So when we look it up when we actually have a hash v2 value, it's different. I think we need to adjust our hashv2 lookup logic to include the default hash of empty string as well, and treat it as if there is no hashv2.

dustin-decker: Okay, let's adjust that before merging this or any more changes w/ rawv2

trufflesteeeve: I have the PR for the fix here, nearly done - https://github.com/trufflesecurity/thog/pull/1065

trufflesteeeve: I'm also noticing that with secretsv2 saving, changing the HashV2 will always create a new secret.

trufflesteeeve: I've fully updated but not yet tested the PR above to allow HashV2 to be fine to update. I'm going to try and finish that up early tomorrow.

PR #1370, opened 2023-05-31
fix mockaroo fps

dustin-decker:
Adds a response check instead of just status code

### DIFF
--- a/pkg/detectors/mockaroo/mockaroo.go
+++ b/pkg/detectors/mockaroo/mockaroo.go
@@ -3,6 +3,7 @@ package mockaroo
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


mcastorina:
Thanks for adding this, but it still fails the detector tests we have:

```
[pr/1370●] » go test -tags detectors ./pkg/detectors/mockaroo
--- FAIL: TestMockaroo_FromChunk (0.99s)
    --- FAIL: TestMockaroo_FromChunk/found,_verified (0.45s)
        mockaroo_test.go:100: Mockaroo.FromData() found, verified diff: (-got +want)
             [
              {
               DetectorType: 753,
               DetectorName: "",
               DecoderType: 0,
            -  Verified: false,
            +  Verified: true,
               Raw: [
               ],
               RawV2: [
               ],
               Redacted: "",
               ExtraData: {
               },
               StructuredData: nil,
              },
             ]
FAIL
FAIL    github.com/trufflesecurity/trufflehog/v3/pkg/detectors/mockaroo 1.159s
FAIL
```

Changing the verification URL to `https://my.api.mockaroo.com` works though, from my testing earlier today.

dustin-decker:
It's failing that because the token is expired. The documentation says `https://api.mockaroo.com/api/types`

https://www.mockaroo.com/docs#GET__api_types

ahrav (approved):

dustin-decker:
Actually it was failing because I had the incorrect response type, fixed now and the test is passing

PR #1373, opened 2023-06-01
Check that git meets version requirements

dustin-decker:
Checks that git meets 2.36.6<=x<3.0.0 version requirements.

### DIFF
--- a/pkg/sources/git/cmd_check.go
+++ b/pkg/sources/git/cmd_check.go
@@ -1,0 +1,41 @@
+package git
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


mcastorina (approved):
lgtm; I'm not familiar with git versions, is there a reason it must be less than `v3.0.0`?

dustin-decker:
> lgtm; I'm not familiar with git versions, is there a reason it must be less than `v3.0.0`?

Git 3.0 doesn't exist yet but it'll be a breaking change release that we'll need to test before loosening up the versions.

PR #1375, opened 2023-06-01
add new key pat for mailgun detector

zubairk14:
expanding coverage of our mailgun api key types. 

new key type of 'key-1384019812018' requires token to be placed into username field of Basic Auth.


dustin-decker:
>pending todo: once mailgun support responds, we can switch back our new active key test case to verified: true

Okay, let's wait on this one until our test account is sorted out.

dustin-decker (approved):

PR #1382, opened 2023-06-06
Add DocuSign detector

zubairk14:
Use an undocumented `client_credentials` grant_type to see if we can successfully retrieve an `access_token` from docusign integration key (their version of client id) and secret key. 

Responses will probably always be `200`, so we have an additional check to confirm we receive a valid access token in the response.


pkg/detectors/docusign/docusign.go
```diff
@@ -0,0 +1,128 @@
...
+
+				req.Header.Add("Accept", "application/vnd.docusign+json; version=3")
+				req.Header.Add("Authorization", fmt.Sprintf("Basic %s", encodedCredentials))
+				res, err := client.Do(req)
+
+				// Read the response body
+				body, err := ioutil.ReadAll(res.Body)
+
+				if err != nil {
+					fmt.Println("Error reading response body:", err)
```
dustin-decker, line 90: There should never be prints in a detector

pkg/detectors/docusign/docusign.go
```diff
@@ -0,0 +1,125 @@
...
+				}
+
+				// Close the response body
+				res.Body.Close()
+
+				// Parse the response body into a Response struct
+				var parsedResponse Response
+				err = json.Unmarshal(body, &parsedResponse)
+				if err != nil {
+					return nil, errors.WrapPrefix(err, "Error reading response body", 0)
```
dustin-decker, line 100: Error text is incorrect here

pkg/detectors/docusign/docusign.go
```diff
@@ -0,0 +1,125 @@
...
+				}
+
+				if err == nil {
+					defer res.Body.Close()
```
dustin-decker, line 104: You've already closed the response on L94

dustin-decker:
I forgot to tell you this before, but this needs to be imported in https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/defaults.go to be used

dustin-decker (approved):

PR #1387, opened 2023-06-09
Split files instead of using ReadAll

bill-rich:
Split files into chunks even though chunker would do it later to avoid using huge amounts of memory and causing OOM kill.

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -185,7 +185,6 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 
 	for {
 		chunkBytes := make([]byte, BufferSize)
-		chunk := *chunkSkel
 		reader := bufio.NewReaderSize(reReader, BufferSize)
 		n, err := reader.Read(chunkBytes)
 		if err != nil && !errors.Is(err, io.EOF) {
@@ -193,7 +192,6 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 		}
 		peekData, _ := reader.Peek(PeekSize)
 		if n > 0 {
-			chunksChan <- &chunk
 			chunksChan <- &sources.Chunk{
 				SourceType: s.Type(),
 				SourceName: s.name,
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -207,6 +207,9 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 				Verify: s.verify,
 			}
 		}
+		if errors.Is(err, io.EOF) {
+			break
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -181,23 +182,33 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
...
+				SourceName: s.name,
+				SourceID:   s.SourceID(),
+				Data:       append(chunkBytes[:n], peekData...),
+				SourceMetadata: &source_metadatapb.MetaData{
+					Data: &source_metadatapb.MetaData_Filesystem{
+						Filesystem: &source_metadatapb.Filesystem{
+							File: sanitizer.UTF8(path),
+						},
+					},
 				},
```
mcastorina, line 208: Looks like we're writing to the channel twice here, where the first one is made from the skeleton?

bill-rich: Thanks! Left the mix of original `filesystem` chunking and `chunker` chunking.

mcastorina (approved):

pkg/sources/filesystem/filesystem.go
```diff
@@ -181,23 +182,31 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
...
-		SourceMetadata: &source_metadatapb.MetaData{
-			Data: &source_metadatapb.MetaData_Filesystem{
-				Filesystem: &source_metadatapb.Filesystem{
-					File: sanitizer.UTF8(path),
+
+	for {
+		chunkBytes := make([]byte, BufferSize)
+		reader := bufio.NewReaderSize(reReader, BufferSize)
+		n, err := reader.Read(chunkBytes)
+		if err != nil && !errors.Is(err, io.EOF) {
```
trufflesteeeve, line 190: I believe if reader.Read doesn't read anything, it returns an EOF. So how does this loop terminate? Or am I wrong?

bill-rich: You're correct. Blew through this too late on a Friday and counted on the test too much. Going to get a better filesystem test added.

mcastorina: I read through the [Reader](https://pkg.go.dev/io#Reader) interface docs, which recommends:

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

So that would suggest checking `err != nil` *after* our `n > 0` expression.

I think I would feel safer with something like that and:
```go
if err != nil || n == 0 {
    break
}
```

Which begs the question, should we return the `err` if it's not an EOF error?

bill-rich: I'm ok with switching, but this should work too. The earlier check if the error is not EOF and the one at the end checking if it is EOF is to cover the two behaviors (hits end of reader w/ EOF or hits end of reader and returns EOF on next run iteration).

PR #1396, opened 2023-06-14
Make trace error message so newlines aren't escaped

bill-rich:


mcastorina (approved):

mcastorina:
btw, you should be able to format the JSON to output raw strings with `jq -r '."stack-trace"'`

bill-rich:
Thats an awesome tip! Going to hold on to that one.

PR #1397, opened 2023-06-15
Add Validator interface and example

bill-rich:
Sources can implement the `Validator` interface to test their configuration with error reporting.

Syslog source implements Validator as an example.

### DIFF
--- a/pkg/sources/syslog/syslog.go
+++ b/pkg/sources/syslog/syslog.go
@@ -59,6 +59,44 @@ func NewSyslog(sourceType sourcespb.SourceType, jobID, sourceID int64, sourceNam
 	}
 }
 
+// Validate validates the configuration of the source.
+func (s *Source) Validate() []error {
+	errors := []error{}
+
+	if s.conn.TlsCert != nilString || s.conn.TlsKey != nilString {
+		if s.conn.TlsCert == nilString || s.conn.TlsKey == nilString {
+			errors = append(errors, fmt.Errorf("tls cert and key must both be set"))
+		}
+		if _, err := tls.LoadX509KeyPair(s.conn.TlsCert, s.conn.TlsKey); err != nil {
+			errors = append(errors, fmt.Errorf("error loading tls cert and key: %s", err))
+		}
+	}
+
+	if s.conn.ListenAddress != nilString {
+		switch s.conn.Protocol {
+		case "tcp":
+			_, err := net.Listen(s.conn.Protocol, s.conn.ListenAddress)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/sources/syslog/syslog.go
```diff
@@ -59,6 +59,44 @@ func NewSyslog(sourceType sourcespb.SourceType, jobID, sourceID int64, sourceNam
...
+		}
+		if _, err := tls.LoadX509KeyPair(s.conn.TlsCert, s.conn.TlsKey); err != nil {
+			errors = append(errors, fmt.Errorf("error loading tls cert and key: %s", err))
+		}
+	}
+
+	if s.conn.ListenAddress != nilString {
+		switch s.conn.Protocol {
+		case "tcp":
+			_, err := net.Listen(s.conn.Protocol, s.conn.ListenAddress)
```
mcastorina, line 78: **Question:** Do we need to close the returned socket?

pkg/sources/syslog/syslog.go
```diff
@@ -59,6 +59,44 @@ func NewSyslog(sourceType sourcespb.SourceType, jobID, sourceID int64, sourceNam
...
+			if err != nil {
+				errors = append(errors, fmt.Errorf("error listening on udp socket: %s", err))
+			}
+		default:
+			errors = append(errors, fmt.Errorf("invalid protocol: %s", s.conn.Protocol))
+		}
+	}
+	if s.conn.Protocol != "tcp" && s.conn.Protocol != "udp" {
+		errors = append(errors, fmt.Errorf("invalid protocol: %s", s.conn.Protocol))
+	}
```
mcastorina, line 93: **Nit:** This looks like a duplicate error to the `default` case above.

mcastorina (approved):
lgtm; I think we have to be careful when returning `[]error` since it's valid to check against `nil` but it's always true for `[]error{}`

It also might be nice to specify what we were expecting for the `protocol` or `format` so the user knows how to correct it.

pkg/sources/syslog/syslog.go
bill-rich: They'll automatically be closed when the program terminates, but it'd be better to do it explicitly.

ahrav (approved):

pkg/sources/syslog/syslog.go
```diff
@@ -59,6 +59,44 @@ func NewSyslog(sourceType sourcespb.SourceType, jobID, sourceID int64, sourceNam
 	}
 }
 
+// Validate validates the configuration of the source.
+func (s *Source) Validate() []error {
+	errors := []error{}
```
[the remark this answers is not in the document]
bill-rich: I like it

Issue #1330, opened 2023-05-05
Print which detector is used when a secret is found with custom detectors


Nouman404:
### Description

I think that it would be great if it was possible to see the name of the custom detector used when we create our own regex. For now, we only have "Custom Regex" printed out and it would be better to have the name of the custom regex used.

![image](https://user-images.githubusercontent.com/73934639/236404324-76e4905c-0c4b-4c6a-9c08-2e0b0d3016cf.png)

zricethezav:
@Nouman404 Thanks for opening this issue. I'll take a crack at it.

zricethezav:
Just opened https://github.com/trufflesecurity/trufflehog/compare/main...custom-detector-name where you can track progress. Right now it's just a hacky solution to get the `name` in the output. For example:
```
Found unverified result 🐷🔑❓
Detector Type: CustomRegex
Decoder Type: PLAIN
Raw result: hogs are cool
Name: hog detector
Commit: 6a761551b86e55fd84163f378c12190ce91ea210
Email: Zachary Rice <email>
File: test1.md
Line: 2
Timestamp: 2023-06-09 15:13:12 +0000
```

using this custom config:
```
# config.yaml
detectors:
- name: hog detector
  keywords:
  - hog
  regex:
    adjective: hogs are (\S+)
```

zricethezav:
related issue: https://github.com/trufflesecurity/trufflehog/issues/1063#issue-1567803192

Issue #1063, opened 2023-02-02
Add structured data to custom regex detector's results


fabricat:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Additional information could be added to enrich the results of the custom regex detectors.

## Problem to be Addressed
Regex detector is a very powerful and flexible feature (yet in alpha stage), but its output is too poor currently.

## Description of the Preferred Solution

I'd appreciate if the output results include:
* name of the **detector** (coming from the config file, not `CustomRegex` - in the JSON output this can be added in `ExtraData`)
* name and value of each **regex submatch** (in the JSON output this can be added **as an object** within `StructuredData`)
* **redacted** string (as like as [some detectors](https://github.com/trufflesecurity/trufflehog/blob/80a68b84c2db1d05bab20f85dbb6cd0830adb5af/pkg/detectors/ftp/ftp.go#L59) already do - this requires a _new key_ in the configuration file)

### References

* https://github.com/trufflesecurity/trufflehog/blob/80a68b84c2db1d05bab20f85dbb6cd0830adb5af/pkg/custom_detectors/custom_detectors.go#L96

zricethezav:
Closing with the release of https://github.com/trufflesecurity/trufflehog/releases/tag/v3.41.0

PR #1400, opened 2023-06-15
Custom detector name

zricethezav:
Adds `name` to the extra data map for custom detectors.

Ex output:

```
Found unverified result 🐷🔑❓
Detector Type: CustomRegex
Decoder Type: PLAIN
Raw result: eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
Name: Super custom API detector
File: /Users/zrice/code/trufflesecurity/sandbox/config-demo/test.md
Line: 1
```

Takes care of https://github.com/trufflesecurity/trufflehog/issues/1330 and https://github.com/trufflesecurity/trufflehog/issues/1063


pkg/custom_detectors/custom_detectors.go
```diff
@@ -99,6 +99,10 @@ func (c *customRegexWebhook) FromData(ctx context.Context, verify bool, data []b
 	close(resultsCh)
 
 	for result := range resultsCh {
+		// NOTE: I don't believe this is being set anywhere else, hence the map assignment.
+		result.ExtraData = map[string]string{
```
zricethezav, line 103: for the reviewer: could check for `nil` here but it doesn't look as nice. dont have a strong opinion on this one.

mcastorina (commented):
We are also assigning `c.GetName()` to the `Result.DetectorName` on line 120. Are we not displaying that information anywhere?

mcastorina:
Should we add the entry there instead of including it in `ExtraData`? Maybe with a check to prevent outputting non-empty names?

I have a vague recollection that `DetectorName` was added for this use case.

zricethezav:
> Should we add the entry there instead of including it in ExtraData? Maybe with a check to prevent outputting non-empty names? 

I have a small pushback on that since `DetectorName` is only used for CustomRegex detectors which means we would have to introduce some extra logic to check if DetectorName is set in `PrintPlainOutput`. Not a huge deal but imo a little less clean than relying on the metadata. Happy to discuss further if you feel strongly on this

mcastorina (approved):
No strong opinion either way, but maybe we should remove `DetectorName` if we aren't using it.

PR #1419, opened 2023-06-23
verify response body with expected keywords

zubairk14:
Sometimes, HTTP calls to endpoints with _invalid_ credentials return a `200` status, but without the actual access token (or actual access to the resource). Verifying response body with expected keywords helps mitigate that source of false positives.


pkg/common/utils.go
```diff
@@ -17,3 +23,44 @@ func RemoveStringSliceItem(item string, slice *[]string) {
...
+		if err != nil {
+			return false, err
+		}
+		return found, nil
+	}
+
+	return false, nil
+}
+
+func containsSubstring(reader io.ReadCloser, target string) (bool, error) {
```
ahrav, line 45: suggestion: You can probably even use the `bufio.NewScanner` here. It's a little easier to read imo.
```
func containsSubstring(reader io.Reader, substring string) (bool, error) {
	scanner := bufio.NewScanner(reader)
	for scanner.Scan() {
		if strings.Contains(scanner.Text(), substring) {
			return true, nil
		}
	}
	if err := scanner.Err(); err != nil {
		return false, err
	}
	return false, nil
}
```

pkg/common/utils_test.go
```diff
@@ -1,7 +1,9 @@
 package common
 
 import (
+	"io/ioutil"
```
ahrav, line 4: suggestion: io/ioutil is deprecated, so we should just use io

pkg/common/utils_test.go
```diff
@@ -74,3 +76,31 @@ func TestRemoveItem(t *testing.T) {
 		}
 	}
 }
+
+// Test ParseResponseForKeywords with a reader that contains the keyword and a reader that doesn't.
+func TestParseResponseForKeywords(t *testing.T) {
```
ahrav, line 81: suggestion: Since we have multiple test cases with the exact same setup we might benefit from a table-driven test. We could also maybe throw in some edge cases?

https://google.github.io/styleguide/go/best-practices#tests
```
If inputs are similar, consider unifying them into a table-driven test while keeping the logic inlined in the loop. This helps to avoid repetition while keeping the validation and failure in the Test.
```

Ex:
```
func TestParseResponseForKeywords(t *testing.T) {
	testCases := []struct {
		name     string
		input    string
		keywords []string
		expected bool
	}{
		{
			name:     "Should find keyword",
			input:    "ey: abc",
			keywords: []string{"ey"},
			expected: true,
		},
		{
			name:     "Should not find keyword",
			input:    "fake response",
			keywords: []string{"ey"},
			expected: false,
		},
		{
			name:     "Empty string",
			input:    "",
			keywords: []string{"ey"},
			expected: false,
		},
		{
			name:     "Keyword at end",
			input:    "abc ey",
			keywords: []string{"ey"},
			expected: true,
		},
		{
			name:     "Keyword at start",
			input:    "ey abc",
			keywords: []string{"ey"},
			expected: true,
		},
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			testReader := strings.NewReader(tc.input)
			testReadCloser := io.NopCloser(testReader)
			found, err := ParseResponseForKeywords(testReadCloser, tc.keywords)

			if err != nil {
				t.Errorf("Error: %v", err)
			}

			if found != tc.expected {
				t.Errorf("Expected %v, got %v", tc.expected, found)
			}
		})
	}
}
```

pkg/detectors/docusign/docusign.go
```diff
@@ -83,25 +81,14 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 					return nil, errors.WrapPrefix(err, "Error making request", 0)
 				}
 
-				// Read the response body
-				body, err := io.ReadAll(res.Body)
+				verifiedBodyResponse, err := common.ParseResponseForKeywords(res.Body, []string{"ey"})
```
ahrav, line 84: suggestion: Linter noticed the response body is not closed, looks like it got removed in the refactor. We should throw that back in here to prevent any resource leaks.

ahrav (approved):

pkg/common/utils.go
zubairk14: thanks for the suggestion!

pkg/common/utils_test.go
zubairk14: ah true

pkg/common/utils.go
```diff
@@ -17,3 +23,44 @@ func RemoveStringSliceItem(item string, slice *[]string) {
 		}
 	}
 }
+
+// ParseResponseForKeywords parses the response from detector verification calls for expected keywords in the response.
+func ParseResponseForKeywords(reader io.ReadCloser, keywords []string) (bool, error) {
```
dustin-decker, line 28: This won't perform as expected because reading from the body consumes the reader. That means that once you start reading it, you won't be able to read those parts again. Since we only need to check the response for one substring anyway, I recommend removing this function and just renaming `containsSubstring` to `ResponseContainsSubstring`

dustin-decker (changes requested):

dustin-decker (approved):

PR #1428, opened 2023-06-26
Define SourceUnit enumeration interface

mcastorina:
This is an optional (for now) interface that a source can implement to support creation of source units. An implementation for the filesystem source is added for an example of how it will look.

### DIFF
--- a/pkg/common/context.go
+++ b/pkg/common/context.go
@@ -10,3 +10,15 @@ func IsDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+// CancellableWrite blocks on writing the item to the channel but can be
+// cancelled by the context. If both the context is cancelled and the channel
+// write would succeed, either operation will be performed randomly.
+func CancellableWrite[T any](ctx context.Context, ch chan<- T, item T) error {
+	select {
+	case ch <- item:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -50,6 +50,26 @@ type SourceUnitUnmarshaller interface {
 	UnmarshalSourceUnit(data []byte) (SourceUnit, error)
 }
 
+// SourceUnitEnumerator defines an optional interface a Source can implement to
+// support enumerating an initialized Source into SourceUnits.
+type SourceUnitEnumerator interface {
+	// Enumerate enumerates the initialized Source, outputting units. This
+	// method is synchronous but can be called in a goroutine to support
+	// concurrent enumeration and chunking. An error should only be
+	// returned from this method in the case of context cancellation. All
+	// other errors related to unit enumeration are tracked in the
+	// EnumerationResult.
+	Enumerate(ctx context.Context, units chan<- EnumerationResult) error
+}
+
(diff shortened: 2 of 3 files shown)


pkg/sources/sources.go
```diff
@@ -50,6 +50,26 @@ type SourceUnitUnmarshaller interface {
 	UnmarshalSourceUnit(data []byte) (SourceUnit, error)
 }
 
+// SourceUnitEnumerator defines an optional interface a Source can implement to
+// support enumerating an initialized Source into SourceUnits.
+type SourceUnitEnumerator interface {
+	// Enumerate enumerates the initialized Source, outputting units. This
+	// method is synchronous but can be called in a goroutine to support
```
bill-rich, line 57: Not necessarily disagreeing, but why make it synchronous instead of handling the concurrency inside `Enumerate`? Is the idea that the `units` chan will be shared for enumerating multiple sources?

mcastorina: Mostly thinking that it's easier for a source to implement a synchronous function and have the `SourceManager` (when it exists) handle the concurrency.

For the initial implementation it'll be 1:1 unit channel to enumeration call. I don't anticipate it being shared across sources, since we need to know which source a unit is coming from.

pkg/common/context.go
```diff
@@ -10,3 +10,15 @@ func IsDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+// CancellableWrite blocks on writing the item to the channel but can be
+// cancelled by the context. If both the context is cancelled and the channel
+// write would succeed, either operation will be performed randomly.
+func CancellableWrite[T any](ctx context.Context, ch chan<- T, item T) error {
```
ahrav, line 17: question: Does it make sense to use maybe a custom type instead of any for the the chan. I think that added type security will save us from potentially putting something of the wrong type into the channel. I'm sure the receiver of the item will likely check the type, but this way we can hopefully catch it at compile time.

pkg/sources/sources.go
```diff
@@ -223,3 +243,21 @@ func (p *Progress) GetProgress() *Progress {
 	defer p.mut.Unlock()
 	return p
 }
+
+// CommonEnumerationOk is a helper function to construct an EnumerationResult
+// using a CommonSourceUnit.
+func CommonEnumerationOk(id string) EnumerationResult {
+	return EnumerationResult{
```
ahrav, line 250: nit: We can omit the zero value for Error if you want. You can then inline the thing.
`return EnumerationResult{Unit: CommonSourceUnit{ID: id}}`

Leaving it is also fine if we feel like it adds extra clarity. I'm leaning toward removing it only b/c your comment for `ENumerationResult` mentions the fields being mutually exclusive. Similar comment for EumerationErr.

ahrav (approved):

pkg/common/context.go
mcastorina: I would agree with that, except this is a generic function and the `any` here is a [type constraint](https://go.dev/doc/tutorial/generics#add_generic_function), which means it works with a channel of any type `T`. Additionally, the `item` must also be *the same* `T`.

Calling this function with `chan int` and `string` will not compile.

PR #1431, opened 2023-06-26
Add field for verification indeterminate to detector results

ahrav:
- This will be used to distinguish between credentials being unverified, as opposed to some other failure that occurred during the verification process. (ie 429, timeout, etc)

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -74,6 +74,12 @@ func (r Result) IsVerified() bool {
 	return r.Verified && !r.HasVerificationIndeterminate
 }
 
+// IsUnverified checks if a result is guaranteed to be unverified.
+// This is used to differentiate between secrets that are NOT verified and secrets that COULD NOT be verified.
+func (r Result) IsUnverified() bool {
+	return !r.Verified && !r.HasVerificationIndeterminate
+}
+
 type ResultWithMetadata struct {
 	// SourceMetadata contains source-specific contextual information.
 	SourceMetadata *source_metadatapb.MetaData
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -39,6 +39,8 @@ type EndpointCustomizer interface {
 	DefaultEndpoint() string
 }
 
+// Result is a secret that has been found by a Detector and optionally verified.
+// It contains the raw secret identification data, redacted data, and extra data.
 type Result struct {
 	// DetectorType is the type of Detector.
 	DetectorType detectorspb.DetectorType
@@ -46,7 +48,15 @@ type Result struct {
 	DetectorName string
 	// DecoderType is the type of Decoder.
 	DecoderType detectorspb.DecoderType
-	Verified    bool
+	// Verified is true if the secret has been confirmed to be a valid secret.
+	Verified bool
+	// HasVerificationIndeterminate is true if the secret could NOT be verified.
+	// This is used to differentiate between secrets that are NOT verified and secrets that COULD NOT be verified.
+	// Ex: A 429 response from a service, or a timeout would set this to true.
+	HasVerificationIndeterminate bool
+	// VerificationIndeterminateError holds the error if the secret could NOT be verified.
+	// This is used specifically for cases that resulted in indeterminate resolution, such as a 429 or timeout.
+	VerificationIndeterminateError error
 	// Raw contains the raw secret identifier data. Prefer IDs over secrets since it is used for deduping after hashing.
 	Raw []byte
 	// RawV2 contains the raw secret identifier that is a combination of both the ID and the secret.
@@ -57,6 +67,11 @@ type Result struct {
 	Redacted       string
 	ExtraData      map[string]string
 	StructuredData *detectorspb.StructuredData
+}
+
+// IsVerified checks if a result is guaranteed to be verified.
+func (r Result) IsVerified() bool {
+	return r.Verified && !r.HasVerificationIndeterminate
 }
 
 type ResultWithMetadata struct {
--- a/pkg/detectors/detectors_test.go
+++ b/pkg/detectors/detectors_test.go
@@ -3,7 +3,9 @@
 
 package detectors
 
-import "testing"
+import (
+	"testing"
+)
 
 func TestPrefixRegex(t *testing.T) {
 	tests := []struct {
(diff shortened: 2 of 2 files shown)


pkg/detectors/detectors.go
```diff
@@ -39,14 +39,24 @@ type EndpointCustomizer interface {
...
-	Verified    bool
+	// Verified is true if the secret has been confirmed to be a valid secret.
+	Verified bool
+	// HasVerificationIndeterminate is true if the secret could NOT be verified.
+	// This is used to differentiate between secrets that are NOT verified and secrets that COULD NOT be verified.
+	// Ex: A 429 response from a service, or a timeout would set this to true.
+	HasVerificationIndeterminate bool
+	// VerificationIndeterminateError holds the error if the secret could NOT be verified.
+	// This is used specifically for cases that resulted in indeterminate resolution, such as a 429 or timeout.
+	VerificationIndeterminateError error
```
ahrav, line 59: Do we want this to be an error or a string? I can see an argument for either case.

pkg/detectors/detectors.go
```diff
@@ -59,6 +69,11 @@ type Result struct {
 	StructuredData *detectorspb.StructuredData
 }
 
+// IsVerified checks if a result is guaranteed to be verified.
+func (r Result) IsVerified() bool {
+	return r.Verified && !r.HasVerificationIndeterminate
+}
```
mcastorina, line 75: This is a stricter definition of verified, which I think loosens the definition for unverified assuming `unverified = !verified`, which I don't think is necessarily the case.

Should we also add a `IsUnverified() bool` method?

pkg/detectors/detectors.go
mcastorina, line 59: Thoughts on combining these two attributes into just the `error`, where `nil` indicates `HasVerificationIndeterminate == false`? Or rather, is `(true, nil)` / `(false, non-nil)` valid?

Just throwing it out there for opinions. It might be more confusing tying it together.

ahrav: So my thoughts/push back on that are the following:
1.The dual fields HasVerificationIndeterminate and VerificationIndeterminateError very clearly separate the concern of whether an indeterminate verification state was reached and the specific error that occurred in that case. By collapsing these into one field, i feel like you lose some of that semantic clarity.
2.A nil error is universally understood in Go as "no error occurred". Using nil to signify false for HasVerificationIndeterminate could be counter-intuitive, as this could be interpreted as "the verification was indeterminate, but no error occurred", which may not be the intended meaning. It also feels a bit weird to me, but that's just me.
3. Again me likely being annoying, but the nil check is more verbose,If you use the VerificationIndeterminateError field to signify both the occurrence of an error and the indeterminate verification state, your code will frequently need to check if result.VerificationIndeterminateError != nil to determine whether verification was indeterminate.
This is more verbose and less direct than checking if result.HasVerificationIndeterminate, which clearly and succinctly communicates the intent of the check.
4.Using an error object to store a boolean state can be seen as a misuse of the error type. Errors in Go are generally used to hold information about something that went wrong in the execution of the program, not to hold state information. Using the error in this way in my opinion makes the code less clear, as future readers of the code might expect VerificationIndeterminateError to be used in the conventional way, i.e., to hold error information, not state information.

That being said, I think it makes sense to get additional opinions.

pkg/detectors/detectors.go
ahrav: Yea, that's a valid point. I think we might need that method as well. Thanks.

pkg/detectors/detectors.go
mcastorina: Thanks for the well thought out response to my half baked question :)

Yeah, I like the explicitness of the two separate variables. My original thought was along the lines of something like an `Option` enum where a non-nil error indicates `Some(error)` and a nil error indicates `None`.

It semantically represents two states, rather than four when separated into `bool` and `error`. What would it mean if `VerificationIndeterminateError` is `non-nil` and `HasVerificationIndeterminate` is `false`? Or vice versa?

ahrav: Ahh I see what you mean. Also the alternative also would be a weird state. `HasVerificationIndetermindate` is true and non-nil error.

So a few ideas come to mind:
1. We can simply go off the basis that if `HasVerification...` is not set, we don't care about the error, and vice versa. (i don't like this that much)
2. Make both the fields unexported and use helpers to get and set:
ex:
```
func (r Result) GetVerificationIndeterminate() (bool, error) {
    return r.hasVerificationIndeterminate, r.verificationIndeterminateError
}

func (r Result) SetVerificationIndeterminate(hasIndeterminate bool, err error) {
    if hasIndeterminate && err != nil {
        r.hasVerificationIndeterminate = true
        r.verificationIndeterminateError = err
    } else if !hasIndeterminate && err == nil {
        r.hasVerificationIndeterminate = false
        r.verificationIndeterminateError = nil
    } else {
        panic("Invalid state")
    }
}
```
3. something akin to state enumeration, (maybe what you were referring to?):
ex:
```
type VerificationState int

const (
    VerificationUnknown VerificationState = iota
    VerificationConfirmed
    VerificationIndeterminate
)

func (r Result) GetVerificationState() VerificationState {
    return r.verificationState
}

func (r Result) GetVerificationIndeterminateError() error {
    if r.verificationState != VerificationIndeterminate {
        return nil
    }
    return r.verificationIndeterminateError
}

func (r Result) SetVerificationState(state VerificationState, err error) {
    switch state {
    case VerificationUnknown, VerificationConfirmed:
        if err != nil {
            ...
        }
        r.verificationState = state
        r.verificationIndeterminateError = nil
    case VerificationIndeterminate:
        if err == nil {
            ...
        }
        r.verificationState = state
        r.verificationIndeterminateError = err
    default:
        ...
    }
}
```

mcastorina: Hm.. none of these options seem great and I'm probably over-complicating the design with this theoretical..

dustin-decker: What if `HasVerificationIndeterminate` just became a method that does the nil check?

dustin-decker: I think this should be an error. Errors can be wrapped, and we could use `errors.Is` to check for retryable errors later on, or maybe to just clean up the errors for user presentation.

ahrav:
Just to make sure I don't butcher this, i'm gonna close and let Cody take the reins.

Issue #1198, opened 2023-03-20
errors from client.Do(req) are ignored - leading to potential false negatives


nyanshak:
There is a common pattern in detectors that can lead to undetected false negative results (e.g., result.Verified is set to false even though the issue is valid).

The pattern:

```go
for _, match := range matches {
    ...
    if verify {
        ...
        
        res, err := client.Do(req)
        if err == nil {
            // do all the checks for known false positives and against status codes
        }
        // BUG: in the case where err != nil, the error is ignored
    }

    results = append(results, s1)
}
```

So in the case where some error has occurred while doing the http request, the error from `client.Do(req)` is ignored, and the result will have `Verified` field set to false even though the match may be valid. This may lead to missing valid results and not knowing that an error occurred while scanning that maybe could be retried / re-scanned.

Here are a few examples (though it's in almost every detector):

* https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/cicero/cicero.go#L56-L67
* https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/launchdarkly/launchdarkly.go#L55-L66
* https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/jiratoken/jiratoken.go#L75-L81

The Okta detector appears to do this correctly and would be a good example of how this should be done instead: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/okta/okta.go#L63-L66. As a suggested fix, replacing the broken pattern with the one by okta detector may be ideal.

(Similarly, Okta detector reports the error in constructing the request where other detectors do not, though I'm not sure if this is as much of an issue: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/okta/okta.go#L55-L58)

ManQuiche:
Hey ! 

I will try to fix this quickly.

Is it relevant to work on still adding the results but, for each, telling the user there has been some errors during verification ? Or logs will be enough ?

nyanshak:
I think you could go a few ways on it, but it's essential that the errors get returned in `FromData` calls so that callers know to retry (e.g., that they should re-run trufflehog, or call `scanner.FromData` again).

This could be something like:

```
var err error
...
for _, idMatch := range idMatches {
    ...
    res, doErr := client.Do(req)
    if doErr != nil {
         // Option 1: join the error but continue processing the rest of the results
        err = errors.Join(err, doErr)
        continue

[...]

PR #1432, opened 2023-06-26
Don't return on okta credential failed verification

mcastorina:

### DIFF
--- a/pkg/detectors/okta/okta.go
+++ b/pkg/detectors/okta/okta.go
@@ -63,7 +63,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 				resp, err := common.SaneHttpClient().Do(req)
 				if err != nil {
-					return results, err
+					continue
 				}
 				defer resp.Body.Close()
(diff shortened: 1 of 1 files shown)


ahrav:
hmmm, it seems like according to the discussion here: https://github.com/trufflesecurity/trufflehog/issues/1198, that is what we want?

mcastorina:
Any time an error is returned from `FromChunk` we skip processing all of the results:

https://github.com/trufflesecurity/trufflehog/blob/4a77688097d83259a4ebf1b26d4daec8e130e57b/pkg/engine/engine.go#L299-L311

An error from sending an HTTP request could indicate a network failure and isn't fatal. It's not exactly correct, but I think it's less wrong to `continue` here than to return an error.

dustin-decker:
Yeah that seems most correct for now. Longer term, we should be collecting and surfacing the error for each result without returning early. We might have to rework the approach from #1198 through the bit augmentation proposal, collecting all errors from all results, or migrating to a new interface that separates the detection from verification stages.

dustin-decker (approved):

PR #1435, opened 2023-06-27
scan GitHub PR and issue comments

zubairk14:
Scan PR and issue comments. Support paginated calls to GitHub Issues and Pull Requests service. 

* https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28
* https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28


pkg/sources/github/github.go
```diff
@@ -919,3 +925,141 @@ func (s *Source) setProgressCompleteWithRepo(index int, offset int, repoURL stri
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case chunksChan <- chunk:
+		}
+	}
+	return nil
+}
+
+func removeURLAndSplit(url string) []string {
```
ahrav, line 1059: suggestion: We might also want to ensure repos that are provided using ssh also works. I think we have a similar function that you can likely use as a reference: https://github.com/trufflesecurity/trufflehog/blob/452734adc8f92b723e33ac052c088544de131dc0/pkg/sources/git/git.go#L249

pkg/sources/github/github.go
```diff
@@ -919,3 +925,141 @@ func (s *Source) setProgressCompleteWithRepo(index int, offset int, repoURL stri
...
+		Sort:      sortType,
+		Direction: directionType,
+		ListOptions: github.ListOptions{
+			PerPage: defaultPagination,
+			Page:    1,
+		},
+	}
+
+	for {
+		prComments, _, err := s.apiClient.PullRequests.ListComments(ctx, owner, repo, 0, prOpts)
```
ahrav, line 976: suggestion: We should probably handle rate limits similar to how we do throughout this file.
ex: https://github.com/trufflesecurity/trufflehog/blob/452734adc8f92b723e33ac052c088544de131dc0/pkg/sources/github/github.go#L286

pkg/sources/github/github.go
```diff
@@ -919,3 +925,141 @@ func (s *Source) setProgressCompleteWithRepo(index int, offset int, repoURL stri
...
 }
+
+func (s *Source) scanComments(ctx context.Context, repoPath string, chunksChan chan *sources.Chunk) error {
+	s.log.Info("scanning comments")
+
+	trimmedURL := removeURLAndSplit(repoPath)
+	owner := trimmedURL[1]
+	repo := trimmedURL[2]
+
+	sortType := "created"
```
ahrav, line 936: nit: Grouping related vars tends to read a bit nicer imo.
ex:
```
	var (
		sortType      = "created"
		directionType = "desc"
	)
```

pkg/sources/github/github.go
ahrav, line 976: nit: Can we replace the magic '0' w/ a const and a tiny comment.
ex: `	const defaultIssueNumber = 0 // List all comments for all issues for the repo.`

pkg/sources/github/github.go
```diff
@@ -919,3 +925,141 @@ func (s *Source) setProgressCompleteWithRepo(index int, offset int, repoURL stri
...
+		Sort:      &sortType,
+		Direction: &directionType,
+		ListOptions: github.ListOptions{
+			PerPage: defaultPagination,
+			Page:    1,
+		},
+	}
+
+	for {
+		issueComments, _, err := s.apiClient.Issues.ListComments(ctx, owner, repo, 0, issueOpts)
```
ahrav, line 949: Whoops missed this one, but same comments as below.

ahrav (approved):
Nice job, LGTM!

pkg/sources/github/github.go
zubairk14: ahhh true

pkg/sources/github/github.go
zubairk14: yep for sure

pkg/sources/github/github.go
zubairk14: oooh ill try this - thanks for the code pointer

pkg/sources/github/github.go
zubairk14: good call

pkg/sources/github/github.go
zubairk14: just imported bill's gitURLParse (made it a public func)

PR #1438, opened 2023-06-29
tighten up Shortcut API detector

zubairk14:
fix missing api key, tighten up regex pattern, use response body check


pkg/detectors/shortcut/shortcut.go
```diff
@@ -48,23 +48,33 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.app.shortcut.com/api/v3/projects", nil)
```
zubairk14, line 51: not every key will have a project, but every key will at a minimum have a member

pkg/detectors/shortcut/shortcut_test.go
```diff
@@ -19,7 +19,7 @@ import (
 func TestShortcut_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors2")
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
```
zubairk14, line 22: shortcut api keys were not in detectors2 file

dustin-decker (approved):
Nice improvement

PR #1439, opened 2023-06-29
remove HEAD from git diff command, rename unstaged to staged

zricethezav:
This PR addresses a few things:

1. the function name `Unstaged` should be `Staged` since `--cached` is a synomom for `--staged` according to this [git diff documentation](https://git-scm.com/docs/git-diff):

> git diff [<options>] --cached [--merge-base] [<commit>] [--] [<path>…​]
> This form is to view the changes you staged for the next commit relative to the named <commit>. Typically you would want comparison with the latest commit, so if you do not give <commit>, it defaults to HEAD. If HEAD does not exist (e.g. unborn branches) and <commit> is not given, it shows all staged changes. --staged is a synonym of --cached.
> 

2. Removes `HEAD` from the `git diff -p` arguments since `git diff` will default to `HEAD` if no refs are present. In my testing this change fixes this issue https://github.com/trufflesecurity/trufflehog/issues/1403 

> Typically you would want comparison with the latest commit, so if you do not give <commit>, it defaults to HEAD. If HEAD does not exist (e.g. unborn branches) and <commit> is not given, it shows all staged changes


dustin-decker (approved):
Nice find on the HEAD change

Issue #1453, opened 2023-07-05
GitHub source does not pass apiEndpoint for basicAuth or unautheticated


rgmz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version

3.43.0

### Trace Output

TruffleHog passes the [`githubScanEndpoint`](https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/main.go#L371) for `GitHub_Token` and `GitHub_GithubApp` auth, but not `GitHub_BasicAuth` or `GitHub_Unauthenticated`. This is not really an inconvenience but is nevertheless a bug.

https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/pkg/sources/github/github.go#L353-L366

### Expected Behavior

TruffleHog calls the endpoint specified with `--endpoint` and not `api.github.com`.

```sh
./trufflehog github --endpoint="https://github.example.com/api/v3" --org="example" --only-verified  
2023-07-05T18:10:10-04:00       info-0  trufflehog      loaded decoders {"count": 3}
2023-07-05T18:10:10-04:00       info-0  trufflehog      loaded detectors        {"total": 747, "verification_enabled": 747, "verification_disabled": 0}
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2023-07-05T18:10:20-04:00       info-0  trufflehog      Completed enumeration   {"source_type": "SOURCE_TYPE_GITHUB", "source_name": "github", "num_repos": 1234, "num_orgs": 1, "num_members": 0}
Found verified result 🐷🔑
...
```

### Actual Behavior

TruffleHog ignores `--endpoint` and calls `api.github.com`:

```sh
trufflehog github --endpoint="https://github.example.com/api/v3" --org="example" --only-verified
2023-07-05T18:17:59-04:00	info-0	trufflehog	loaded decoders	{"count": 3}
[...]

PR #1454, opened 2023-07-05
Pass GitHub apiEndpoint for basic or no auth

rgmz:
This is a quick and dirty fix for #1453. 

It may be even easier to just use `github.NewEnterpriseClient` for everything, which is what `enumerateWithApp` seems to do:
https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/pkg/sources/github/github.go#L558

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -356,11 +356,11 @@ func (s *Source) enumerate(ctx context.Context, apiEndpoint string) (*github.Cli
 
 	switch cred := s.conn.GetCredential().(type) {
 	case *sourcespb.GitHub_BasicAuth:
-		if err = s.enumerateBasicAuth(ctx, cred.BasicAuth); err != nil {
+		if err = s.enumerateBasicAuth(ctx, apiEndpoint, cred.BasicAuth); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


pkg/sources/github/github.go
```diff
@@ -377,11 +377,16 @@ func (s *Source) enumerate(ctx context.Context, apiEndpoint string) (*github.Cli
 	return installationClient, nil
 }
 
-func (s *Source) enumerateBasicAuth(ctx context.Context, basicAuth *credentialspb.BasicAuth) error {
-	s.apiClient = github.NewClient(&http.Client{Transport: &github.BasicAuthTransport{
+func (s *Source) enumerateBasicAuth(ctx context.Context, apiEndpoint string, basicAuth *credentialspb.BasicAuth) error {
+	httpClient := &http.Client{Transport: &github.BasicAuthTransport{
```
bill-rich, line 381: Although it isn't in the current code, I think it would be better to use `s.httpClient` and modify the transport. It was set up earlier as a retriable client, so it would be good to continue to use that.

bill-rich (changes requested):
Looks good. One minor change requested.

bill-rich (approved):
Thanks for fixing this!

PR #1461, opened 2023-07-06
Remove the Image4 detector

rgmz:
Image4.io unfortunately shut down in 2022: https://www.indiehackers.com/product/image4io/closed--NA-O_aQ-twkW1nICGZ-

I'm not sure if it's necessary to remove the DetectorType enums as well:
https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/pkg/pb/detectorspb/detectors.pb.go#L828

https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/proto/detectors.proto#L765


dustin-decker (approved):
Thank you!

PR #1462, opened 2023-07-06
additional similarity check for base64 and plain

zricethezav:
This PR reduces the chances a duplicate finding will be reported when the base64 decoder gets run on regular code/text by checking the first 40 (randomly chosen, could/should this be smaller?) characters of a chunk to see if they are the same for decoded vs the original bytes. It's not the best solution but will likely take care of most of these cases. Fixes this issue: https://github.com/trufflesecurity/trufflehog/issues/1450

Other alternative solutions could include keeping track of all the findings and running a deduplication filter during the scan similar to how gitleaks handles duplicates: https://github.com/gitleaks/gitleaks/blob/06db3b91d7495a39caf0d922aef5fee0118e7538/detect/utils.go#L63-L92. Implementing something like that should probably have a dedicated issue associated with it for further discussion.


pkg/engine/engine.go
```diff
@@ -272,8 +272,16 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+				original := chunk.Data
 				decoded := decoder.FromChunk(chunk)
-				if decoded == nil {
+
+				if decoded == nil ||
+					// check if the decoded data is similar "enough" to the original data. If it is, then we can skip scanning the decoded data as
+					// it's likely already picked up by the PLAIN decoder. See related issue: https://github.com/trufflesecurity/trufflehog/issues/1450
+					(decoded != nil &&
+						decoderType == detectorspb.DecoderType_BASE64 &&
+						bytes.Equal(original[0:common.MinInt(40, len(original)-1)], decoded.Data[0:common.MinInt(40, len(decoded.Data)-1)])) {
```
ahrav, line 284: very nit: I could see a small benefit of pulling this into a helper function. Then again there is also the comment so it's probably fine this way.

I was thinking maybe something like:
```
func areFirstNBytesEqual(a, b []byte, numBytes int) bool {
	limit := MinInt(numBytes, MinInt(len(a), len(b)))
	return bytes.Equal(a[:limit], b[:limit])
}
```

ahrav (approved):
I didn't think the solution would've been this intuitive. 😆

pkg/engine/engine.go
zricethezav: Good call, that makes it more readable

dustin-decker:
This change still finds this live GCP credential in the Kubernetes repo:
https://github.com/kubernetes/kubernetes/commit/74bc8a3211620397419f0e4ea996e862b070f9bb#diff-5ef420463c194b54174938185a2059709c9ba9c70a81c93e3b4a7774fb7160beR241

But not this one: https://github.com/kubernetes/kubernetes/commit/89502d3c8cd6601299e003bd8741885affdc9e28#diff-cb41bace1688ea19d6a9e00eebe45be69a4d34efec8d395f3b223e57257f081cR141

PR #1463, opened 2023-07-06
Add new verification error message field

rosecodym:
This PR creates a new field in detector result payloads to contain error messages that signal indeterminate verification status. They are still unused but will support the tri-state verification initiative.


rosecodym:
This will help support the long-term fix alluded to in #1432 and eventually provide a way to address the concerns raised in #1198.

pkg/detectors/detectors.go
```diff
@@ -57,6 +57,10 @@ type Result struct {
 	Redacted       string
 	ExtraData      map[string]string
 	StructuredData *detectorspb.StructuredData
+
+	// This field should only be populated if the verification process itself failed in a way that provides no
+	// information about the verification status of the candidate secret, such as if the verification request timed out.
+	VerificationErrorMessage string
```
mcastorina, line 63: What do you think about using an `error` here? It makes slightly more sense from a types perspective in my opinion, and a `string` can be retrieved from it with the `Error()` method.

mcastorina (approved):

pkg/detectors/detectors.go
rosecodym: seems prudent

PR #1464, opened 2023-07-06
add thog scanner keys detector

zubairk14:
Scan for thog scanner keys. Verify via grpc heartbeat.


scripts/gen_proto.sh
```diff
@@ -37,3 +37,10 @@ protoc -I proto/ \
     --go_out=plugins=grpc:./pkg/pb/custom_detectorspb --go_opt=paths=source_relative \
     --validate_out="lang=go,paths=source_relative:./pkg/pb/custom_detectorspb" \
     proto/custom_detectors.proto
+protoc -I proto/ \
+    -I ${GOPATH}/src \
+    -I /usr/local/include \
+    -I ${GOPATH}/src/github.com/envoyproxy/protoc-gen-validate \
+    --go_out=plugins=grpc:./pkg/detectors/trufflehogenterprisescanner/scannerpb --go_opt=paths=source_relative \
+    --validate_out="lang=go,paths=source_relative:./pkg/detectors/trufflehogenterprisescanner/scannerpb" \
+    proto/api.proto
```
zubairk14, line 46: this is to generate the heartbeart protos

pkg/detectors/trufflehogenterprisescanner/trufflehogenterprisescanner.go
```diff
@@ -0,0 +1,117 @@
...
+	ctx := context.Background()
+	ctx = metadata.NewOutgoingContext(ctx,
+		metadata.New(map[string]string{
+			"name":          strings.TrimSpace(group),
+			"authorization": strings.TrimSpace(token),
+		}))
+
+	systemRoots, err := x509.SystemCertPool()
+	if err != nil {
+		panic(errors.Wrap(err, "cannot load root CA certs"))
```
rosecodym, line 98: Is this a normal way of dealing with failures during verification?

scripts/gen_proto.sh
mcastorina: **Nit (optional):** The filename `api.proto` doesn't sound like it's being used for detector validation but I'm not sure what a better name would be. Could it be added to `detectors.proto` or maybe `detectors_validation.proto`?

pkg/detectors/trufflehogenterprisescanner/trufflehogenterprisescanner.go
zubairk14: @dustin-decker and i had paired on this grpcHeartbeat and come up with this approach. is there an alternative you'd recommend?

rosecodym: i mean panicking. is the thinking here that without the root certs nothing is going to work anyway?

zubairk14: that is my understanding

ahrav: I think instead of panicking here, we should just return the error. That way we still record the cred as unverified since it made it through keyword and regex matching. If we panic here there is a chance the entire application will crash mid scan. We don't really want the application to crash b/c one detector failed to verify a secret.

Actually, I do believe we have a recover mechanism for panics, but nonetheless, we likely just want to return the error here and treat it as an unverified result.

dustin-decker: I think Ahrav is right, this error should be returned. Worth pointing out that currently the errors are swallowed (this will soon change with Cody's work), and that without the root CA store all verification over TLS will fail.

A future improvement we could consider is embedding our own CertPool to fall back to in the default HTTP client if there is no root CA store present. And it's pretty easy to end up in that situation with Docker images.

zricethezav:
Closing due to inactivity. We can always revisit if there is demand.

PR #1466, opened 2023-07-06
Include the line number GitHub & Gitlab links

rgmz:
Both GitHub and Gitlab support linking to a specific line in a file with `#L<number>`. 

This is a small quality of life improvement that makes it easier to find the secret if you're using GitHub or Gitlab, rather than clicking the link and manually having to scroll tens, hundreds, or thousands of lines.

I have not:
- [ ] Confirmed that the test works
- [ ] Confirmed that all results have a line number, or what `line` will be if they don't (since it's an `int64` I guess it will be `0`?)


pkg/sources/git/git.go
```diff
@@ -637,15 +638,23 @@ func normalizeConfig(scanOptions *ScanOptions, repo *git.Repository) (err error)
...
 	if strings.Contains(repo, "bitbucket.org/") {
 		return repo[:len(repo)-4] + "/commits/" + commit
 	}
-	link := repo[:len(repo)-4] + "/blob/" + commit + "/" + file
-
+	var link string
 	if file == "" {
 		link = repo[:len(repo)-4] + "/commit/" + commit
+	} else {
+		link = repo[:len(repo)-4] + "/blob/" + commit + "/" + file
```
ahrav, line 650: nit: I think if the line is 0 (not sure if this would happen, maybe if we are using the zero value), we would append `L#0` to the url. I don't think it matters too much since both github and gitlab appear to just show you the page, but figured i'd mention it. We could also check if line is 0 and return the link before doing the contains check and concatentation.

pkg/sources/git/git_test.go
```diff
@@ -171,6 +172,16 @@ func Test_generateLink(t *testing.T) {
 			},
 			want: "https://github.com/trufflesec-julian/confluence-go-api/blob/047b4a2ba42fc5b6c0bd535c5307434a666db5ec/.gitignore",
```
ahrav, line 173: i believe this test case will fail due to the `L#0` note I mentioned above. If you feel like keeping the L#0 we would just need to update this test as well so it passes.

ahrav (approved):
Thanks a lot @rgmz, I think this is a really useful quality of life improvement. We also appreciate the extra test case.

pkg/sources/git/git.go
rgmz: I've updated the logic accordingly.

I'm also wondering whether it's worth passing in the type of source, rather than relying on the presence of "github" or "gitlab" in the URL. Right now only `github.go` and `gitlab.go` seem to call `git.GenerateLink`, so it should always generate the link.

ahrav: I think it seems okay for now given the points you mentioned. I also feel like we could also potentially refactor this at some point to remove the magic strings throughout the function. Including at the start where we do a check for if it's "bitbucket".

PR #1468, opened 2023-07-07
    Feat: adds a new decoder for compressed data.

joewarrenmeeks:
This is aimed at occasions where people attempt to smuggle secrets out
    by zipping up a text file then base64 encoding it and to detect encoded
    zipped secrets in things like pipeline logs from a CI system.


dustin-decker:
Hey Joe, thanks for the contribution. Can you clean up your commits so that you only need to sign the CLA once?

For clarity for others, we already do archive unpacking, but that occurs before the decoders. TBD whether we will always support recursive decoding, the perf implications can get dicey.

PR #1477, opened 2023-07-11
add role assumption for s3 source

codevbus:
Adds assumeRole for AWS/S3 Source


pkg/sources/s3/s3.go
```diff
@@ -134,76 +175,99 @@ func (s *Source) newClient(region string) (*s3.S3, error) {
...
+	if err != nil {
+		return errors.WrapPrefix(err, "could not create s3 client", 0)
+	}
+
+	if err := processBuckets(client); err != nil {
+		return err
+	}
+
+	// If roles are provided, create additional clients for each role and process
+	for _, roleArn := range s.conn.Roles {
```
mcastorina, line 261: I think we're missing an assignment to the connection's `Roles` attribute in [pkg/engine/s3.go](https://github.com/trufflesecurity/trufflehog/blob/35b332c49860cecd897d4fe5ea39c3eebb08ec81/pkg/engine/s3.go#L21-L23). It should be in the `S3Config` object passed into that function.

https://github.com/trufflesecurity/trufflehog/blob/35b332c49860cecd897d4fe5ea39c3eebb08ec81/pkg/engine/s3.go#L19-L23

mcastorina (commented):
Looking good! Just one more piece to wire up and an ineffectual assignment to clean up.

codevbus:
@mcastorina @dustin-decker - This is ready for review. Some contextual notes and possible discussion points:

- I ended up creating a separate function to handle generating a client for roles. It seemed much easier to break that functionality out vs. nesting a bunch of if/else and iteration inside an auth client
- The `S3_Unauthenticated` type is actually what triggers if I run Trufflehog locally and have AWS credentials set in local files (`~/.aws/credentials`)
- The `S3_Unauthenticated` case defaulted to explicit bucket inputs for generating a list of buckets to scan. I duplicated the functionality under the previous case to enumerate all of the buckets the IAM identity has the ability to make `ListBuckets` API call for.
- Interestingly, if it defaults to the behavior of enumerating all buckets via `ListBuckets`, it will say `scanning bucket` and will not throw an error, if if the IAM policy does not grant the `GetObject` capability.
- If you pass an explicit list of buckets and an explicit list of roles, it will attempted to scan each bucket with both IAM roles.

pkg/sources/s3/s3.go
```diff
@@ -132,79 +134,212 @@ func (s *Source) newClient(region string) (*s3.S3, error) {
...
+	const defaultAWSRegion = "us-east-1"
+
+	if len(s.conn.Roles) > 0 {
+		for _, role := range s.conn.Roles {
+			
+		        client, err := s.newRoleClient(defaultAWSRegion, role)
+		        if err != nil {
+		        	return errors.WrapPrefix(err, "could not create s3 client", 0)
+		        }
+
```
mcastorina, line 196: **Suggestion:** Moving the code in the loop body could help with readability, especially as the steps grow. It also helps to see what is needed for the operation and makes the code generally more testable.

```go
for _, role := range s.conn.Roles {
    // adjust name as desired
    if err := s.chunkWithRole(ctx, chunksChan, defaultAWSRegion, role); err != nil {
        ctx.Logger().V(2).Info("failed chunking with role", "role", role)
        continue
    }
}
```

pkg/sources/s3/s3.go
```diff
@@ -132,79 +134,212 @@ func (s *Source) newClient(region string) (*s3.S3, error) {
...
+		        		if err != nil {
+		        			return fmt.Errorf("could not list s3 buckets: %w", err)
+		        		}
+		        		buckets := res.Buckets
+		        		for _, bucket := range buckets {
+		        			bucketsToScan = append(bucketsToScan, *bucket.Name)
+		        		}
+		        	} else {
+		        		bucketsToScan = s.conn.Buckets
+				}
```
mcastorina, line 225: These two cases look identical and can be combined (or maybe the switch can even be removed?)

pkg/sources/s3/s3.go
```diff
@@ -132,79 +134,212 @@ func (s *Source) newClient(region string) (*s3.S3, error) {
...
+		        	if err != nil {
+		        		s.log.Error(err, "could not get s3 region for bucket", "bucket", bucket)
+		        		continue
+		        	}
+		        	var regionalClient *s3.S3
+		        	if region != defaultAWSRegion {
+		        		regionalClient, err = s.newRoleClient(region, role)
+		        		if err != nil {
+		        			s.log.Error(err, "could not make regional s3 client")
+		        		}
```
mcastorina, line 249: **Question:** Should we `continue` here?

pkg/sources/s3/s3.go
```diff
@@ -132,79 +134,212 @@ func (s *Source) newClient(region string) (*s3.S3, error) {
...
+		        objectCount := uint64(0)
+		        for i, bucket := range bucketsToScan {
+		        	if common.IsDone(ctx) {
+		        		return nil
+		        	}
+
+		        	s.SetProgressComplete(i, len(bucketsToScan), fmt.Sprintf("Bucket: %s", bucket), "")
+
+		        	s.log.Info("Scanning bucket", "bucket", bucket)
+		        	region, err := s3manager.GetBucketRegionWithClient(context.Background(), client, bucket)
```
mcastorina, line 239: **Nit:** We can use the passed in `ctx` instead of making a `context.Background()`.

main.go
```diff
@@ -99,10 +99,13 @@ var (
...
-	s3ScanSessionToken  = s3Scan.Flag("session-token", "S3 session token used to authenticate temporary credentials. Can be provided with environment variable AWS_SESSION_TOKEN.").Envar("AWS_SESSION_TOKEN").String()
+	s3Scan    = cli.Command("s3", "Find credentials in S3 buckets.")
+	s3ScanKey = s3Scan.Flag("key", "S3 key used to authenticate. Can be provided with environment variable AWS_ACCESS_KEY_ID.").Envar("AWS_ACCESS_KEY_ID").String()
+	// TODO: set as repeatable Strings()
+	s3ScanRoleArns     = s3Scan.Flag("role-arn", "Specify the ARN of an IAM role to assume for scanning. You can repeat this flag.").Strings()
+	s3ScanSecret       = s3Scan.Flag("secret", "S3 secret used to authenticate. Can be provided with environment variable AWS_SECRET_ACCESS_KEY.").Envar("AWS_SECRET_ACCESS_KEY").String()
+	s3ScanSessionToken = s3Scan.Flag("session-token", "S3 session token used to authenticate temporary credentials. Can be provided with environment variable AWS_SESSION_TOKEN.").Envar("AWS_SESSION_TOKEN").String()
+	// TODO: add RoleARN in s3
```
mcastorina, line 108: **Nit:** It looks like these `TODO` comments can be removed.

pkg/sources/s3/s3.go
```diff
@@ -132,79 +134,212 @@ func (s *Source) newClient(region string) (*s3.S3, error) {
...
-		s.SetProgressComplete(i, len(bucketsToScan), fmt.Sprintf("Bucket: %s", bucket), "")
+	sess, err := session.NewSession(cfg)
+	if err != nil {
+		return nil, err
+	}
+	
+	stsClient := sts.New(sess)
+	baseCredentials = stscreds.NewCredentialsWithClient(stsClient, roleArn, func(p *stscreds.AssumeRoleProvider) {
+		p.RoleSessionName = "trufflehog"
+	})
```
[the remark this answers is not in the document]
codevbus: Good call out. Functionality seems unaffected by dropping the switch case here.

pkg/sources/s3/s3.go
codevbus: Going to return explicit error so the user knows if role assumption failed(or some other issue)

pkg/sources/s3/s3.go
```diff
@@ -204,6 +234,44 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk) err
...
+
+			bucketsToScan, err := s.getBucketsToScan(client)
+			if err != nil {
+				return err
+			}
+
+			if err := s.scanBuckets(ctx, client, role, bucketsToScan, chunksChan); err != nil {
+				return err
+			}
+		}
```
mcastorina, line 259: I'm not very familiar with role based access, so forgive me if this is a dumb question.

Would it make sense to fallback to non-role based access if the roles were to fail?

mcastorina: Also, this loop calls `scanBuckets` for each role? Does that mean we would scan the same data N times, or does each role have access to different data/buckets?

codevbus: The way that the SDK creates an authentication request makes fallback a bit cumbersome. The user essentially has to start the entire handshake process over again, so I don't think we'd gain much.

The functionality as-is will enumerate all the buckets a given role has the `s3:ListBucket` permission for and scan them, and then repeat for each role that has been passed in by the user.

mcastorina: Ahh that makes a lot more sense! Thanks for clarifying for me.

Unless I'm still missing something, I think it would make sense to `continue` on error instead of returning so we can try to scan the next bucket. What do you think?

dustin-decker (approved):

mcastorina (approved):
lgtm - nice work!

README.md
```diff
@@ -116,25 +116,31 @@ Expected output:
 trufflehog s3 --bucket=<bucket name> --only-verified
 ```
 
-## 5: Scan a Github Repo using SSH authentication in docker
+## 5: Scan S3 buckets using IAM Roles
+
+```bash
+trufflehog s3 --role-arn=<iam role arn>
```
mcastorina, line 122: Should this have the `--bucket` flag?

README.md
```diff
@@ -218,6 +224,30 @@ For example, to scan a  `git` repository, start with
 $ trufflehog git https://github.com/trufflesecurity/trufflehog.git
 ```
 
+## S3 
+
+The S3 source now supports assuming IAM roles for scanning in addition to IAM users. This makes it easier for users to scan multiple AWS accounts without needing to rely on hardcoded credentials for each account.
```
mcastorina, line 229: **Nit:** Remove "now" in the first sentence since the README is more of a long-lived doc.

README.md
```diff
@@ -218,6 +224,30 @@ For example, to scan a  `git` repository, start with
...
+
+The IAM identity that Trufflehog uses initially will need to have `AssumeRole` privileges as a principal in the [trust policy](https://aws.amazon.com/blogs/security/how-to-use-trust-policies-with-iam-roles/) of each IAM role to assume.
```
mcastorina, line 231: **Nit:** `Trufflehog` -> `TruffleHog`

README.md
codevbus: This specific example will enumerate all buckets a role has the authorization to list via the API.

zricethezav (approved):

PR #1479, opened 2023-07-11
Dedupe results

zricethezav:
This PR adds a deduplication feature by collecting all the chunk results _then_ comparing them before sending them off on the channel. We are also removing the decoder similarity check as this is a more general and permanent solution to that problem.


pkg/engine/engine.go
```diff
@@ -233,6 +234,21 @@ func (e *Engine) BytesScanned() uint64 {
...
+func (e *Engine) dedupeAndSend(chunkResults []detectors.ResultWithMetadata) {
+	dedupeMap := make(map[string]struct{})
+	for _, result := range chunkResults {
+		// dedupe by comparing the detector type, raw result, and source metadata
+		// NOTE: in order for the PLAIN decoder to maintain precedence, make sure UTF8 is the first decoder in the
+		// default decoders list
+		if _, ok := dedupeMap[result.DetectorType.String()+string(result.Raw)+fmt.Sprintf("%+v", result.SourceMetadata)]; ok {
+			continue
+		}
+		dedupeMap[result.DetectorType.String()+string(result.Raw)+fmt.Sprintf("%+v", result.SourceMetadata)] = struct{}{}
```
mcastorina, line 246: **Nit (optional):** A single `Sprintf` would be a little easier to read imo:

```go
key := fmt.Sprintf("%s%s%+v", result.DetectorType.String(), string(result.Raw), result.SourceMetadata)
if _, ok := dedupeMap[key]; ok {
    continue
}
dedupeMap[key] = struct{}{}

// ...
```

Also, should we be using `RawV2` instead? or both?

pkg/engine/engine.go
```diff
@@ -257,6 +273,7 @@ func (e *Engine) DetectorAvgTime() map[string][]time.Duration {
 func (e *Engine) detectorWorker(ctx context.Context) {
 	for originalChunk := range e.chunks {
 		for chunk := range sources.Chunker(originalChunk) {
+			var chunkResults = make([]detectors.ResultWithMetadata, 0)
```
mcastorina, line 276: **Nit (optional):** the default value of a slice (`nil`) can be `append`ed, so this is essentially equivalent:

```go
var chunkResults []detectors.ResultWithMetadata
```

Alternatively if you want to keep the `make`, I'd suggest the walrus operator instead of the `var` syntax:

```go
chunkResults := make([]detectors.ResultWithMetadata, 0)
```

mcastorina (commented):
overall looks good to me, just had one question about the data we use to dedupe

pkg/engine/engine.go
zricethezav: probably both. for the tests cases that I was testing with rawv2 was empty string

pkg/engine/engine.go
```diff
@@ -233,6 +234,21 @@ func (e *Engine) BytesScanned() uint64 {
 	return e.bytesScanned
 }
 
+func (e *Engine) dedupeAndSend(chunkResults []detectors.ResultWithMetadata) {
```
ahrav, line 237: I was a bit curious so I did some benchmarking with test with and without the tiny changes i mentioned ad see some pretty nice perf gains.
Screenshot 2023-07-11 at 12 23 53 PM

Test and benchmark code if you were interested:
```
func TestDedupeAndSend(t *testing.T) {
	testCases := []struct {
		name     string
		input    []detectors.ResultWithMetadata
		expected int
	}{
		{
			name: "duplicate detection",
			input: []detectors.ResultWithMetadata{
				{
					Result: detectors.Result{
						DetectorType: detectorspb.DetectorType_AWS,
						Raw:          []byte("test"),
					},
					SourceMetadata: &source_metadatapb.MetaData{
						Data: &source_metadatapb.MetaData_Slack{
							Slack: &source_metadatapb.Slack{
								ChannelName: "test",
							},
						},
					},
				},
				{
					Result: detectors.Result{
						DetectorType: detectorspb.DetectorType_AWS,
						Raw:          []byte("test"),
					},
					SourceMetadata: &source_metadatapb.MetaData{
						Data: &source_metadatapb.MetaData_Slack{
							Slack: &source_metadatapb.Slack{
								ChannelName: "test",
							},
						},
					},
				},
			},
			expected: 1,
		},
		// Add more test cases here...
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			ch := make(chan detectors.ResultWithMetadata, len(tc.input))
			e := Engine{results: ch}

			e.dedupeAndSend(tc.input)

			if len(ch) != tc.expected {
				t.Errorf("Expected %d result(s), got %d", tc.expected, len(ch))
			}
		})
	}
}

func BenchmarkDedupeAndSend(b *testing.B) {
	ch := make(chan detectors.ResultWithMetadata, 5000)
	e := Engine{results: ch}

	results := make([]detectors.ResultWithMetadata, 5000)
	for i := 0; i < 5000; i++ {
		rawValue := fmt.Sprintf("test%d", i)

		// Insert duplicate values every 100th index
		if i%100 == 0 {
			rawValue = "duplicateValue"
		}

		results[i] = detectors.ResultWithMetadata{
			Result: detectors.Result{
				DetectorType: detectorspb.DetectorType_AWS,
				Raw:          []byte(rawValue),
			},
			SourceMetadata: &source_metadatapb.MetaData{
				Data: &source_metadatapb.MetaData_Slack{
					Slack: &source_metadatapb.Slack{
						ChannelName: "test",
					},
				},
			},
		}
	}

	// Start a goroutine to read from the channel concurrently.
	done := make(chan bool)
	go func() {
		for {
			select {
			case <-ch:
				// Read from channel and do nothing.
			case <-done:
				// If done signal is received, stop reading.
				return
			}
		}
	}()

	b.ResetTimer()

	for n := 0; n < b.N; n++ {
		e.dedupeAndSend(results)
	}

	// Send a signal to stop reading from the channel.
	done <- true
}
```

pkg/engine/engine.go
```diff
@@ -233,6 +234,21 @@ func (e *Engine) BytesScanned() uint64 {
...
+	dedupeMap := make(map[string]struct{})
```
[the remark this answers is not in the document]
ahrav: Also i just used 85% off the cuff, I don't have any data that shows this is a good value.

pkg/engine/engine.go
```diff
@@ -233,6 +234,21 @@ func (e *Engine) BytesScanned() uint64 {
...
+	for _, result := range chunkResults {
+		// dedupe by comparing the detector type, raw result, and source metadata
+		// NOTE: in order for the PLAIN decoder to maintain precedence, make sure UTF8 is the first decoder in the
+		// default decoders list
+		if _, ok := dedupeMap[result.DetectorType.String()+string(result.Raw)+fmt.Sprintf("%+v", result.SourceMetadata)]; ok {
```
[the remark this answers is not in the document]
ahrav: oh whoops, i see this was updated. Feel free to ignore.

pkg/decoders/decoders.go
```diff
@@ -6,6 +6,7 @@ import (
 
 func DefaultDecoders() []Decoder {
 	return []Decoder{
+		// UTF8 must be first for duplicate detection
```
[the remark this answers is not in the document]
zricethezav: great idea!

pkg/engine/engine.go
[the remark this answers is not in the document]
zricethezav: Nice optimizations! I think we should leave them off for now and make sure these minimal changes work as intended before implementing any constraints -- could be some odd edge cases. This chunk of code is on a critical path _if_ secrets are detected. Luckily, _most_ chunks don't contain secrets so I think we should leave the minimal changes and make sure they work then optimize over time. WDYT?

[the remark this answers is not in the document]
ahrav: Yep yep, completely agree!

pkg/engine/engine_test.go
```diff
@@ -62,3 +63,12 @@ func TestFragmentLineOffset(t *testing.T) {
...
 }
+
+// Test to make sure that DefaultDecoders always returns the UTF8 decoder first.
+// Technically a decoder test but we want this to run and fail in CI
+func TestDefaultDecoders(t *testing.T) {
+	ds := decoders.DefaultDecoders()
+	if _, ok := ds[0].(*decoders.UTF8); !ok {
+		t.Errorf("DefaultDecoders() = %v, expected UTF8 decoder to be first", ds)
+	}
+}
```
dustin-decker, line 74: Oh good idea!

dustin-decker: Should UTF-16 also be before base64?

PR #1480, opened 2023-07-11
Report indeterminacy in AWS verifier

rosecodym:
This is the first (tentative) implementation of verifier indeterminacy. It modifies the AWS verifier to report verification indeterminacy in two cases:
* Something goes wrong with the HTTP request itself (such as a context timeout)
* The returned status code is outside of [200, 300)

Note that this implementation stores the requested URI in the error that is reported. This is for two reasons. The first is that the context timeout err contains this information by default. The second is that I explicitly added it for unexpected status codes for parity with the previous and to aid with debugging. Presumably AWS wouldn't allow you put anything sensitive in a URI (in this case it's the key ID but not the secret) but it's something to keep in mind going forward.

I also updated a generated proto comment that I forgot to do earlier.

### DIFF
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -190,9 +190,9 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 						// This function will check false positives for common test words, but also it will make sure the key appears "random" enough to be a real key.
 						if detectors.IsKnownFalsePositive(resSecretMatch, detectors.DefaultFalsePositives, true) {
 							continue
-						} else {
-							s1.VerificationError = fmt.Errorf("request to %v returned unexpected status %d", res.Request.URL, res.StatusCode)
 						}
+
+						s1.VerificationError = fmt.Errorf("request to %v returned unexpected status %d", res.Request.URL, res.StatusCode)
 					}
 				} else {
 					s1.VerificationError = err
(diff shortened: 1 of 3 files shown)


pkg/detectors/aws/aws_test.go
```diff
@@ -227,3 +249,8 @@ func BenchmarkFromData(benchmark *testing.B) {
 		})
 	}
 }
+
+func timeoutContext(timeout time.Duration) context.Context {
+	c, _ := context.WithTimeout(context.Background(), timeout)
```
mcastorina, line 254: Generally speaking, you always want to make sure a context gets cancelled to cleanup any resources. Since this is for a test I don't think we need to worry about leaks though.

pkg/detectors/aws/aws.go
```diff
@@ -190,8 +190,12 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 						// This function will check false positives for common test words, but also it will make sure the key appears "random" enough to be a real key.
 						if detectors.IsKnownFalsePositive(resSecretMatch, detectors.DefaultFalsePositives, true) {
 							continue
+						} else {
+							s1.VerificationError = fmt.Errorf("request to %v returned unexpected status %d", res.Request.URL, res.StatusCode)
```
mcastorina, line 194: **Nit (optional):** Since the `if` does a `continue`, I think we can omit the `else` here unless it makes it more readable.

mcastorina (approved):
lgtm

PR #1482, opened 2023-07-11
add envoy api key scanner

brandonjyan:
Motivation: recent Atlassian data leak due to leaked Envoy credentials
Scans for Envoy API keys and validates them: https://developers.envoy.com/hub/docs/getting-an-access-and-refresh
Response has status code 200 for invalid API keys, so need to check for status 401 in response body.


ahrav (approved):
lgtm!
Thanks a lot for the contribution, really appreciate it.

PR #1484, opened 2023-07-12
Define SourceUnit chunking interface

mcastorina:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -232,14 +232,14 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 }
 
 // ChunkUnit implements SourceUnitChunker interface.
-func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, chunks chan<- sources.ChunkResult) error {
+func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
 	path := unit.SourceUnitID()
 	logger := ctx.Logger().WithValues("path", path)
 
 	cleanPath := filepath.Clean(path)
 	fileInfo, err := os.Stat(cleanPath)
 	if err != nil {
-		return common.CancellableWrite(ctx, chunks, sources.ChunkErr(fmt.Errorf("unable to get file info: %w", err)))
+		return reporter.ChunkErr(ctx, fmt.Errorf("unable to get file info: %w", err))
 	}
 
 	ch := make(chan *sources.Chunk)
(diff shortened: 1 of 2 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -229,3 +230,43 @@ func (s *Source) Enumerate(ctx context.Context, units chan<- sources.Enumeration
...
+			// chunks (in the case of archives).
+			err = s.scanFile(ctx, cleanPath, ch)
+		}
+	}()
+
+	for chunk := range ch {
+		if chunk == nil {
+			continue
+		}
+		if err := reporter.ChunkOk(ctx, *chunk); err != nil {
```
rosecodym, line 262: What does it mean if there's an error here? Is terminating the chunking process the correct thing to do if it happens?

pkg/sources/filesystem/filesystem.go
```diff
@@ -229,3 +230,43 @@ func (s *Source) Enumerate(ctx context.Context, units chan<- sources.Enumeration
...
+	if err != nil {
+		return reporter.ChunkErr(ctx, fmt.Errorf("unable to get file info: %w", err))
+	}
+
+	ch := make(chan *sources.Chunk)
+	go func() {
+		defer close(ch)
+		if fileInfo.IsDir() {
+			// TODO: Finer grain error tracking of individual chunks.
+			err = s.scanDir(ctx, cleanPath, ch)
```
rosecodym, line 250: Does this mean that `err` is shared between the goroutine and the calling function? Is this effectively a discard?

rosecodym (commented):
What made you decide to provide a pair of callbacks rather than a pair of channels?

pkg/sources/filesystem/filesystem.go
mcastorina: Any time the callbacks return an error it should be propagated up. Usually this will be a context cancellation.

mcastorina:
> What made you decide to provide a pair of callbacks rather than a pair of channels?

It seems like a cleaner abstraction, especially from the source perspective. It'll also allow us to centralize writing to the channel in a single codepath, or make test interfaces and not have to worry about channels at all.

pkg/sources/filesystem/filesystem.go
rosecodym: Oh I see, I think some of the `err` shadowing confused me.

rosecodym (approved):

pkg/sources/filesystem/filesystem.go
mcastorina: Yes! That was the intention with the design. Any error returned by those callbacks should terminate chunking.

PR #1485, opened 2023-07-12
Rewrite SourceUnitEnumerator to use UnitReporter instead of a channel

mcastorina:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -220,10 +220,10 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 // Enumerate implements SourceUnitEnumerator interface. This implementation simply
 // passes the configured paths as the source unit, whether it be a single
 // filepath or a directory.
-func (s *Source) Enumerate(ctx context.Context, units chan<- sources.EnumerationResult) error {
+func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) error {
 	for _, path := range s.paths {
-		item := sources.CommonEnumerationOk(path)
-		if err := common.CancellableWrite(ctx, units, item); err != nil {
+		item := sources.CommonSourceUnit{ID: path}
+		if err := reporter.UnitOk(ctx, item); err != nil {
 			return err
 		}
 	}
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -53,21 +53,22 @@ type SourceUnitUnmarshaller interface {
 // SourceUnitEnumerator defines an optional interface a Source can implement to
 // support enumerating an initialized Source into SourceUnits.
 type SourceUnitEnumerator interface {
-	// Enumerate enumerates the initialized Source, outputting units. This
-	// method is synchronous but can be called in a goroutine to support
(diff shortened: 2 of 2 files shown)


pkg/sources/sources.go
```diff
@@ -53,21 +53,22 @@ type SourceUnitUnmarshaller interface {
...
-	// EnumerationResult.
-	Enumerate(ctx context.Context, units chan<- EnumerationResult) error
+	// Enumerate creates 0 or more units from an initialized source,
+	// reporting them or any errors to the UnitReporter. This method is
+	// synchronous but can be called in a goroutine to support concurrent
+	// enumeration and chunking. An error should only be returned from this
+	// method in the case of context cancellation, fatal source errors, or
+	// errors returned by the reporter All other errors related to unit
+	// enumeration are tracked by the UnitReporter.
+	Enumerate(ctx context.Context, reporter UnitReporter) error
```
ahrav, line 63: I'm just noticing, but it looks like there are three levels of abstraction here in which all are interface types. Do we think having this much abstraction could get a bit hard to follow? I don't really have an alternative or different approach, but figured i'd mention it.

These all appear to be interfaces (i think):
SourceUnitEnumerator -> UnitReporter -> SourceUnit

mcastorina: Yes, those are all interfaces. I think I'm naturally gravitating towards them for two reasons:

1. It strictly defines the expected API and how parts interact with each other
2. It allows flexibility in its implementation

I think I'm too close to this feature to determine if it's hard to follow, so please let me know if that's the case. I'll brainstorm some alternatives that use less interfaces.

Off the top of my head for the three you mentioned though:

* `SourceUnitEnumerator` - needs to be an interface for a source to implement it. Each source will need to have different implementations.
* `UnitReporter` - could be replaced with a write-only channel that accepts an aggregate struct (containing both a unit and error).
* `SourceUnit` - could be replaced with a general purpose struct (which would be less flexible).

rosecodym: my vote is that the interfaces are good as long as there's a good way to learn about the overall structure as opposed to having to code spelunk. good comments are one possible solution to this; you might also consider writing some documentation somewhere (but then there's the issue of keeping it up to date, of course :( )

pkg/sources/filesystem/filesystem.go
```diff
@@ -220,10 +220,10 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 // Enumerate implements SourceUnitEnumerator interface. This implementation simply
 // passes the configured paths as the source unit, whether it be a single
 // filepath or a directory.
-func (s *Source) Enumerate(ctx context.Context, units chan<- sources.EnumerationResult) error {
+func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) error {
 	for _, path := range s.paths {
-		item := sources.CommonEnumerationOk(path)
-		if err := common.CancellableWrite(ctx, units, item); err != nil {
+		item := sources.CommonSourceUnit{ID: path}
+		if err := reporter.UnitOk(ctx, item); err != nil {
```
rosecodym, line 226: i have the same question about error propagation that i had in #1484 (and i assume the answer is the same)

rosecodym (approved):

PR #1495, opened 2023-07-14
add launch_darkly keyword to launchdarkly scanner

brandonjyan:
Many uses of LaunchDarkly contain the keyword 'launch_darkly' instead of 'launchdarkly', which isn't picked up by the scanner: https://github.com/search?type=code&auto_enroll=true&q=launch_darkly


zricethezav (approved):

zricethezav:
Makes sense to me, thanks @brandonjyan

PR #1496, opened 2023-07-14
add dockerhub scanner

brandonjyan:
add scanner for dockerhub credentials


pkg/detectors/dockerhub/dockerhub.go
```diff
@@ -0,0 +1,91 @@
...
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	usernameMatches := usernamePat.FindAllStringSubmatch(dataStr, -1)
+	emailMatches := emailPat.FindAllString(dataStr, -1)
+	accessTokenMatches := accessTokenPat.FindAllString(dataStr, -1)
+
+	userMatches := emailMatches
+	for _, usernameMatch := range usernameMatches {
+		userMatches = append(userMatches, usernameMatch[1])
```
ahrav, line 48: suggestion: We might want to check the len of `usernameMatch` here to prevent an IOOR panic.

For example if `usernamePat` matches a part of dataStr, but the match does not include any submatches. As a result, `usernameMatches` would be a slice of slice of strings where the inner slices have only 1 string.

We can include a quick check like:
```
if len(usernameMatch) > 1 {
        userMatches = append(userMatches, usernameMatch[1])
    }
```

pkg/detectors/dockerhub/dockerhub.go
```diff
@@ -0,0 +1,91 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	// Can use email or username for login.
+	usernamePat = regexp.MustCompile(`(?im)(?:user|usr|-u)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([a-zA-Z0-9]{4,40})\b[^@]?`)
```
ahrav, line 25: nit (optional): We can replace `{0,1}` -> `?`

pkg/detectors/dockerhub/dockerhub.go
```diff
@@ -0,0 +1,91 @@
...
+	emailPat    = regexp.MustCompile(common.EmailPattern)
+
+	// Can use password or personal access token (PAT) for login, but this scanner will only check for PATs.
+	accessTokenPat = regexp.MustCompile(`\b(?:dckr_pat_)([a-zA-Z0-9_-]){27}\b`)
```
ahrav, line 29: nit (optional): We can remove the capture for `(?:dckr_pat_)` -> `dcker_pat_`. I believe it should have the same behavior.

pkg/detectors/dockerhub/dockerhub_test.go
```diff
@@ -0,0 +1,137 @@
...
+		args    args
+		want    []detectors.Result
+		wantErr bool
+	}{
+		{
+			name: "found, verified",
+			s:    Scanner{},
+			args: args{
+				ctx:    context.Background(),
+				data:   []byte(fmt.Sprintf("docker login -u %s -p %s", username, password)),
```
ahrav, line 48: question: Should we also include the PAT as part of the data here? It looks like the detector only checks for the PAT,  therefore we will need to also include the raw PAT in this []byte.

Ex:
```
pat := testSecrets.MustGetField("DOCKERHUB_PAT")
...
data:   []byte(fmt.Sprintf("docker %s login -u %s -p %s", pat, username, password)),
...
```

pkg/detectors/dockerhub/dockerhub_test.go
```diff
@@ -0,0 +1,137 @@
...
+			want: []detectors.Result{
+				{
+					DetectorType: detectorspb.DetectorType_Dockerhub,
+					Verified:     true,
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "found, verified (email)",
```
ahrav, line 60: question: Were you able to get this to also work? I tried to use an email instead of username and got a 401 from the API. Looking at the docs, they only mention username. https://docs.docker.com/docker-hub/api/latest/#tag/authentication/operation/PostUsersLogin 

It's possible their docs aren't updated, and I setup my test incorrectly, so wanted to double check.

ahrav (commented):
Thanks again for additional contributions. I left a couple questions and one suggestion. Otherwise lgtm! Appreciate it 😄

pkg/detectors/dockerhub/dockerhub_test.go
brandonjyan: The `DOCKERHUB_PASSWORD` field is meant to store the PAT (the PAT is used as a replacement for the password). Sorry for the misleading naming, changed in the commit!

pkg/detectors/dockerhub/dockerhub_test.go
brandonjyan: I don't remember where I found this information 🧐 ... using email works when I send a request via Postman, but it doesn't seem to work on command line. 

Also, I realized the regex was not working as intended which caused the email test to fail, but it should pass now.

pkg/detectors/dockerhub/dockerhub.go
brandonjyan: sounds good!

pkg/detectors/dockerhub/dockerhub_test.go
```diff
@@ -0,0 +1,137 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors2")
+	if err != nil {
+		t.Fatalf("could not get test secrets from GCP: %s", err)
+	}
+	username := testSecrets.MustGetField("DOCKERHUB_USERNAME")
+	email := testSecrets.MustGetField("DOCKERHUB_EMAIL")
+	pat := testSecrets.MustGetField("DOCKERHUB_PAT")
+	inactive_pat := testSecrets.MustGetField("DOCKERHUB_INACTIVE_PAT")
```
ahrav, line 29: nit: Looks like a little python naming might've slipped in here. `inactive_pat` -> `inactivePat`

ahrav (approved):
lgtm!
Thanks again Brandon 😄

PR #1506, opened 2023-07-17
unify JDBC detector ping logic

rosecodym:
Previously, the various JDBC detectors would independently try to verify credentials by a process of trying various permutations of candidates one-by-one. The upcoming tri-state verification work will need to add sophistication to this process in the same way for each one, so this PR first combines all of the logic so it can be upgraded in a single spot.

### DIFF
--- a/pkg/detectors/jdbc/sqlserver_integration_test.go
+++ b/pkg/detectors/jdbc/sqlserver_integration_test.go
@@ -35,7 +35,7 @@ func TestSqlServer(t *testing.T) {
 			wantPing: true,
 		},
 		{
-			input:    "//localhost;database= master;spring.datasource.password=" + sqlServerPass,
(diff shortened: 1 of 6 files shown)


pkg/detectors/jdbc/sqlserver_integration_test.go
```diff
@@ -35,7 +35,7 @@ func TestSqlServer(t *testing.T) {
 			wantPing: true,
 		},
 		{
-			input:    "//localhost;database= master;spring.datasource.password=" + sqlServerPass,
```
rosecodym, line 38: The SQL Server detector does not appear to accept any connection strings with spaces. (This should probably be fixed!)

mcastorina (approved):
nice!

PR #1507, opened 2023-07-18
JDBC indeterminacy

rosecodym:
This PR adds an indeterminacy check to the JDBC verifiers. The way they waterfall through various connection variations makes this a non-trivial thing to do. Ultimately, the only way to determinately decide that credentials are unverified is to receive an authentication error. This raises the hypothetical possibility that credentials can be rotated in a way that Trufflehog will not be able to determinately detect: specifically, if credentials are rotated simultaneously with the database or server in question losing connectivity for some other reason. I think this approach is still best because:
* I think the particular scenario is unlikely
* Manual remediation is still possible
* I can't think of another way to do things :(

### DIFF
--- a/pkg/detectors/jdbc/sqlite.go
+++ b/pkg/detectors/jdbc/sqlite.go
@@ -14,12 +14,18 @@ type sqliteJDBC struct {
 	testing  bool
 }
 
-func (s *sqliteJDBC) ping(ctx context.Context) bool {
+var cannotVerifySqliteError error = errors.New("sqlite credentials cannot be verified")
+
+func (s *sqliteJDBC) ping(ctx context.Context) pingResult {
 	if !s.testing {
 		// sqlite is not a networked database, so we cannot verify
-		return false
+		return pingResult{cannotVerifySqliteError, true}
 	}
-	return ping(ctx, "sqlite3", s.filename)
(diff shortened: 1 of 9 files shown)


pkg/detectors/jdbc/sqlite.go
```diff
@@ -14,12 +14,18 @@ type sqliteJDBC struct {
 	testing  bool
 }
 
-func (s *sqliteJDBC) ping(ctx context.Context) bool {
+var cannotVerifySqliteError error = errors.New("sqlite credentials cannot be verified")
```
mcastorina, line 17: Not something for this PR, but I wonder if we could verify the sqlite credentials if the DB was a local file?

pkg/detectors/jdbc/sqlserver.go
```diff
@@ -3,6 +3,7 @@ package jdbc
 import (
 	"context"
 	"errors"
+	mssql "github.com/denisenkom/go-mssqldb"
 	"strings"
 
 	_ "github.com/denisenkom/go-mssqldb"
```
mcastorina, line 9: **Nit:** Same here about imports

mcastorina (approved):
lgtm!

PR #1510, opened 2023-07-18
Support indeterminacy in alchemy and update detector docs

rosecodym:
This isn't an "important" detector - except that it's the template detector that the detector generator runs on! This PR also:
* Updates the associated documentation
* Changes the method that creates a common testing HTTP client with a timeout to accept a `Duration` directly instead of interpreting a `int64` as a number of seconds
  * Go will silently convert `int64` values to Durations, so we need to be careful when making this change elsewhere

### DIFF
--- a/hack/docs/Adding_Detectors_external.md
+++ b/hack/docs/Adding_Detectors_external.md
@@ -62,14 +62,28 @@ If you think that something should be included outside of these guidelines, plea
    1. Create a [test secrets file, and export the variable](#using-a-test-secret-file)
    2. Update the pattern regex and keywords. Try iterating with [regex101.com](http://regex101.com/).
    3. Update the verifier code to use a non-destructive API call that can determine whether the secret is valid or not.
-   4. Update the tests with these test cases at minimum:
+      * Make sure you understand [verification indeterminacy](#verification-indeterminacy).
+   5. Update the tests with these test cases at minimum:
       1. Found and verified (using a credential loaded from GCP Secrets)
-      2. Found and unverified
-      3. Not found
-      4. Any false positive cases that you come across
+      2. Found and unverified (determinately, i.e. the secret is invalid)
+      3. Found and unverified (indeterminately due to timeout)
+      4. Found and unverified (indeterminately due to an unexpected API response)
+      5. Not found
+      6. Any false positive cases that you come across
    5. Create a pull request for review.
 
 ## Addendum
+
+### Verification indeterminacy
+
+There are two types reasons that secret verification can fail:
+* The candidate secret is not actually a valid secret.
+* Something went wrong in the process unrelated to the candidate secret, such as a transient network error or an unexpected API response.
+
+In Trufflehog parlance, the first type of verification response is called _determinate_ and the second type is called _indeterminate_. Verification code should distinguish between the two by returning an error object in the result struct **only** for indeterminate failures. In general, a verifier should return an error (indicating an indeterminate failure) in all cases that haven't been explicitly identified as determinate failure states. For example, if a verifier makes an HTTP request to a hypothetical authentication endpoint, the response codes could determine what to return:
+* Any `2xx` response would indicate that verification succeeded.
+* A `403` response would indicate that verification failed **determinately** and no error object should be returned.
+* Any other response would indicate that verification failed **indeterminately** and an error object should be returned.
 
 ### Using a test secret file
 
--- a/hack/docs/Adding_Detectors_external.md
+++ b/hack/docs/Adding_Detectors_external.md
@@ -63,7 +63,7 @@ If you think that something should be included outside of these guidelines, plea
    2. Update the pattern regex and keywords. Try iterating with [regex101.com](http://regex101.com/).
    3. Update the verifier code to use a non-destructive API call that can determine whether the secret is valid or not.
       * Make sure you understand [verification indeterminacy](#verification-indeterminacy).
-   5. Update the tests with these test cases at minimum:
+   4. Update the tests with these test cases at minimum:
       1. Found and verified (using a credential loaded from GCP Secrets)
       2. Found and unverified (determinately, i.e. the secret is invalid)
       3. Found and unverified (indeterminately due to timeout)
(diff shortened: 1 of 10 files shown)


hack/docs/Adding_Detectors_external.md
```diff
@@ -62,15 +62,29 @@ If you think that something should be included outside of these guidelines, plea
...
+      4. Found and unverified (indeterminately due to an unexpected API response)
+      5. Not found
+      6. Any false positive cases that you come across
    5. Create a pull request for review.
 
 ## Addendum
 
+### Verification indeterminacy
+
+There are two types reasons that secret verification can fail:
```
ahrav, line 79: nit: two types `of` reasons

ahrav (approved):

samdatkins (approved):
Docs look great to me, it's really nice to have a concrete example

pkg/common/http.go
```diff
@@ -152,9 +176,9 @@ func SaneHttpClient() *http.Client {
 }
 
 // SaneHttpClientTimeOut adds a custom timeout for some scanners
-func SaneHttpClientTimeOut(timeOutSeconds int64) *http.Client {
+func SaneHttpClientTimeOut(timeoutMs int64) *http.Client {
 	httpClient := &http.Client{}
-	httpClient.Timeout = time.Second * time.Duration(timeOutSeconds)
+	httpClient.Timeout = time.Millisecond * time.Duration(timeoutMs)
```
mcastorina, line 181: Since we're changing the units here already, could we change the input parameter to be a `time.Duration` so the user can choose? It also help readability:

```go
client := SaneHttpClientTimeOut(10 * time.Second) // 10s
client := SaneHttpClientTimeOut(10000)            // ???
```

This will also make code that imports this function not compile* when it gets updated, which is better than silently making all timeouts 1000x faster.

*I think.. hopefully Go doesn't silently promote an `int64` to a `time.Duration`

mcastorina (approved):
lgtm with one non-blocking suggestion

pkg/common/http.go
rosecodym: That's a good suggestion but it looks like Go will silently interpret `int64` values as nanoseconds. As you pointed out, this means we should be careful when vendoring the change in.

pkg/detectors/alchemy/alchemy.go
```diff
@@ -47,6 +48,10 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
+			client := s.client
```
trufflesteeeve, line 51: __optional__: You could maybe just check `s.client == nil` and use it on line 59 below, rather than creating a client every time.

hack/docs/Adding_Detectors_external.md
```diff
@@ -62,15 +62,29 @@ If you think that something should be included outside of these guidelines, plea
    1. Create a [test secrets file, and export the variable](#using-a-test-secret-file)
    2. Update the pattern regex and keywords. Try iterating with [regex101.com](http://regex101.com/).
    3. Update the verifier code to use a non-destructive API call that can determine whether the secret is valid or not.
-   4. Update the tests with these test cases at minimum:
+      * Make sure you understand [verification indeterminacy](#verification-indeterminacy).
+   5. Update the tests with these test cases at minimum:
```
trufflesteeeve, line 66: __nit__: This should be `4.`. Though it should render properly regardless.

hack/docs/Adding_Detectors_external.md
```diff
@@ -62,15 +62,29 @@ If you think that something should be included outside of these guidelines, plea
...
+### Verification indeterminacy
+
+There are two types of reasons that secret verification can fail:
+* The candidate secret is not actually a valid secret.
+* Something went wrong in the process unrelated to the candidate secret, such as a transient network error or an unexpected API response.
+
+In Trufflehog parlance, the first type of verification response is called _determinate_ and the second type is called _indeterminate_. Verification code should distinguish between the two by returning an error object in the result struct **only** for indeterminate failures. In general, a verifier should return an error (indicating an indeterminate failure) in all cases that haven't been explicitly identified as determinate failure states. For example, if a verifier makes an HTTP request to a hypothetical authentication endpoint, the response codes could determine what to return:
+* Any `2xx` response would indicate that verification succeeded.
+* A `403` response would indicate that verification failed **determinately** and no error object should be returned.
+* Any other response would indicate that verification failed **indeterminately** and an error object should be returned.
```
trufflesteeeve, line 86: I'm not sure we can say this is true for all apis we test against. e.g. Docker can still be a valid credential and return a 401 - https://github.com/trufflesecurity/trufflehog/blob/8ec5e4916c911144aab39e0602dfa930d3aa2d09/pkg/detectors/dockerhub/dockerhub.go#L79-L82

So maybe adjust the language to be less certain about the indeterminate verification failure in the case of non 2XX or 403 responses.

trufflesteeeve (approved):
Looks good, just some small adjustments.

hack/docs/Adding_Detectors_external.md
rosecodym: I was trying to pick language that made it clear that was a hypothetical case, but I guess I didn't do a good job :(

trufflesteeeve: Maybe a separate paragraph for the start of the hypothetical case? And potentially changing the `would` to `would likely`?

hack/docs/Adding_Detectors_external.md
rosecodym: i wanted to make them all `1.` but goland yelled at me :(

pkg/detectors/alchemy/alchemy.go
rosecodym: do you mean mutating the scanner struct?

trufflesteeeve: Yep. Or set it as a default somewhere? As it is, every time this attempts to verify it will create a new client. Which is fine, but probably not necessary.

PR #1511, opened 2023-07-19
[MongoDB] Detect CosmoDB access keys

rgmz:
The current pattern does not detect [Azure CosmoDB Access Keys](https://learn.microsoft.com/en-us/microsoft-365/compliance/sit-defn-azure-cosmos-db-account-access-key?view=o365-worldwide), which are 88 characters long.

Random example pulled from GitHub:
https://github.com/MichelliBrito/agendalive/blob/e74b7d74cc28140f77e22af1ba372a813dc99aa4/src/main/resources/application.properties#L2

I would recommend generating a live secret and adding that to the test bank, if possible.

### DIFF
--- a/pkg/detectors/mongodb/mongodb.go
+++ b/pkg/detectors/mongodb/mongodb.go
@@ -21,7 +21,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 
 var (
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(mongodb(\+srv)?://[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+)\b`)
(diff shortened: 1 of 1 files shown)


zricethezav:
@rgmz I don't have an issue with this PR per se, but one thought I had, which other team members may take issue with is, should this be a separate detector? From my understanding CosmoDB is an Azure offering by Microsoft  which treats a CosmoDB as a MongoDB. This is a weird one and I don't really have a strong opinion either way.

rgmz:
Given that [CosmoDB for MongoDB supports the MongoDB wire protocol](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/feature-support-36#protocol-support), maintaining a separate detector would probably be unnecessary work (e.g., changes to mongodb would have to be synced).

zricethezav (approved):

PR #1515, opened 2023-07-20
Implement SourceManager basics

mcastorina:
None of this code is used yet, but since it started growing in length I wanted to get some eyes on it before using it in the engine / adding more functionality.

### DIFF
--- a/pkg/engine/filesystem.go
+++ b/pkg/engine/filesystem.go
@@ -28,6 +28,14 @@ func (e *Engine) ScanFileSystem(ctx context.Context, c sources.FilesystemConfig)
 	}
 
 	fileSystemSource := filesystem.Source{}
+	fileSystemSource.WithFilter(c.Filter)
+
+	// TODO: This is how it would be used; everything after these comments would be removed.
+	// handle, err := e.sourceManager.Manage(ctx, "trufflehog - filesystem", fileSystemSource)
+	// if err != nil {
+	// 	return err
+	// }
+	// return e.sourceManager.ScheduleRun(ctx, handle, &conn)
 
 	ctx = context.WithValues(ctx,
 		"source_type", fileSystemSource.Type().String(),
@@ -37,7 +45,6 @@ func (e *Engine) ScanFileSystem(ctx context.Context, c sources.FilesystemConfig)
 	if err != nil {
 		return errors.WrapPrefix(err, "could not init filesystem source", 0)
 	}
-	fileSystemSource.WithFilter(c.Filter)
 	e.sourcesWg.Go(func() error {
 		defer common.RecoverWithExit(ctx)
 		err := fileSystemSource.Chunks(ctx, e.ChunksChan())
--- a/pkg/engine/filesystem.go
+++ b/pkg/engine/filesystem.go
@@ -31,11 +31,19 @@ func (e *Engine) ScanFileSystem(ctx context.Context, c sources.FilesystemConfig)
 	fileSystemSource.WithFilter(c.Filter)
 
 	// TODO: This is how it would be used; everything after these comments would be removed.
-	// handle, err := e.sourceManager.Manage(ctx, "trufflehog - filesystem", fileSystemSource)
+	// handle, err := e.sourceManager.Enroll(ctx, "trufflehog - filesystem", fileSystemSource.Type(),
+	// 	func(ctx context.Context, jobID, sourceID int64) (sources.Source, error) {
+	// 		fileSystemSource := filesystem.Source{}
+	// 		fileSystemSource.WithFilter(c.Filter)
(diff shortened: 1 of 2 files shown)


pkg/engine/filesystem.go
```diff
@@ -37,7 +45,6 @@ func (e *Engine) ScanFileSystem(ctx context.Context, c sources.FilesystemConfig)
 	if err != nil {
 		return errors.WrapPrefix(err, "could not init filesystem source", 0)
 	}
-	fileSystemSource.WithFilter(c.Filter)
```
rosecodym, line 40: Why did you remove this?

pkg/sources/source_manager.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil
+	})
+	return <-ch
+}
+
+// ScheduleRun blocks until a resource is available to run the source, then
+// asynchronously runs it. Error information is lost in this case.
+func (s *SourceManager) ScheduleRun(ctx context.Context, handle handle, conn *anypb.Any) error {
+	if err := ctx.Err(); err != nil {
+		return err
```
rosecodym, line 107: Why isn't this check done in `Run`?

pkg/sources/source_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	headlessJobCounter    int64
+	// Pool limiting the amount of concurrent sources running.
+	pool errgroup.Group
+	// Downstream chunks channel to be scanned.
+	outputChunks chan *Chunk
+}
+
+// sourceTemplate is an aggregate struct to hold the Source and its name.
+type sourceTemplate struct {
+	name   string
```
rosecodym, line 37: Are you expecting to use `name` in logs and the like? I don't see it used anywhere else (but I might just be missing it).

pkg/sources/source_manager.go
```diff
@@ -0,0 +1,168 @@
...
+		outputChunks: outputChunks,
+	}
+	for _, opt := range opts {
+		opt(&man)
+	}
+	return &man
+}
+
+// Manage informs the SourceManager to track and manage a Source.
+func (s *SourceManager) Manage(ctx context.Context, name string, source Source) (handle, error) {
```
rosecodym, line 72: bleh, I am not wild about `Manage` as a verb here. What about `RegisterSource` or something else that alludes to the registration/ownership transfer that's happening?

pkg/sources/source_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	id, err := s.registerSource(ctx, name, source)
+	if err != nil {
+		return 0, err
+	}
+	s.handlesLock.Lock()
+	defer s.handlesLock.Unlock()
+	if _, ok := s.handles[id]; ok {
+		// TODO: smartly handle this?
+		return 0, fmt.Errorf("generated handle ID '%d' already in use locally", id)
```
rosecodym, line 81: I'm wondering if you could avoid having to worry about this weird case by establishing a facade that creates some sort of virtual API for headless managers.

pkg/engine/filesystem.go
mcastorina: I moved it to right after the source is created (line 31). It changes the behavior by setting the filter before calling `Init`, but I checked the source and the order doesn't matter.

pkg/sources/source_manager.go
mcastorina: Ah do you mean like making the `api` required and moving the `headlessJobCounter` etc. into an implementation of that?

I don't see how that would remove this edge case though so maybe I'm misunderstanding.

pkg/sources/source_manager.go
mcastorina: Since `Run` is synchronous, the context cancellation should be caught downstream and returned as an error. Wouldn't hurt to add it there too though.

pkg/sources/source_manager.go
mcastorina: Updated to `Enroll`

pkg/sources/source_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	id, err := s.api.RegisterSource(ctx, name, source.Type())
+	if err != nil {
+		return 0, err
+	}
+	handleID := handle(id)
+	s.handlesLock.Lock()
+	defer s.handlesLock.Unlock()
+	if _, ok := s.handles[handleID]; ok {
+		// TODO: smartly handle this?
+		return 0, fmt.Errorf("generated handle ID '%d' already in use locally", handleID)
```
rosecodym, line 82: is this error message still correct? (is the mention of use "locally" still the best way to describe the problem?)

rosecodym (approved):

rosecodym (approved):
this feels like it's really coming together, and i'm curious to see what it looks like as you flesh out more on the consumer side.

(as an aside: i can't tell whether some of your identifiers are still holdovers from previous implementations, so maybe give them a pass to see if there are any you'd like to update.)

PR #1516, opened 2023-07-20
[chore] Remove parent setting / getting in Context wrapper

mcastorina:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -127,9 +127,10 @@ func Start(ctx context.Context, options ...EngineOption) *Engine {
 	}
 	ctx.Logger().V(2).Info("engine started", "workers", e.concurrency)
 
-	sourcesWg := &errgroup.Group{}
+	sourcesWg, egCtx := errgroup.WithContext(ctx)
 	sourcesWg.SetLimit(e.concurrency)
 	e.sourcesWg = sourcesWg
+	ctx = context.WithLogger(egCtx, ctx.Logger())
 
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -127,10 +127,9 @@ func Start(ctx context.Context, options ...EngineOption) *Engine {
 	}
 	ctx.Logger().V(2).Info("engine started", "workers", e.concurrency)
 
-	sourcesWg, egCtx := errgroup.WithContext(ctx)
+	sourcesWg := &errgroup.Group{}
```
rosecodym, line 130: To my uneducated eye, this looks like a change in functionality - is it? If so, what for? If not, why not?

mcastorina: `errgroup.WithContext` returns a cancellable context, which finishes when any error is returned. This is removing that since `ctx.SetParent` is a destructive behavior.

To maintain behavior, I think we'll want to assign `ctx = egCtx` instead. Thanks for the call out!

ahrav: I think we might need to also update [buildkite](https://github.com/trufflesecurity/thog/blob/847224b30b70a22cbd06a7f94050b43ae96a2786/scanner/pkg/sources/buildkite/buildkite.go#L210) when this gets vendored in.

ahrav (approved):

pkg/engine/engine.go
mcastorina: Ah, that was not there when I looked lol. Good to know!

PR #1524, opened 2023-07-21
Decrease frequency of dependabot alerts to monthly

zricethezav:
Title says it all. Automatically updating all go mod dependencies every week is too much

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,9 @@ updates:
   - package-ecosystem: "gomod" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "weekly"
+      interval: "monthly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       # Check for updates to GitHub Actions every weekday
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -8,9 +8,9 @@ updates:
   - package-ecosystem: "gomod" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "weekly"
+      interval: "monthly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       # Check for updates to GitHub Actions every weekday
```
rosecodym, line 15: Should we update this comment too? Or better yet, delete it?

rosecodym (approved):
While I approve of this in spirit I don't know why the original frequency was picked so you should probably wait for the approval of someone who does :v

mcastorina (approved):
I can't say for certain why a weekly cadence was chosen - I think it was one of those quality of life automations that Dustin setup to make sure our dependencies are up to date.

Monthly cadence is good for the amount of PRs it's generating now imo.

PR #1528, opened 2023-07-23
[chore] - Remove password info from log

ahrav:
- If the clone fails we log the `remotePath` which contains credentials.


mcastorina (approved):
Is that the only place we could log it?

ahrav:
> Is that the only place we could log it?

Just found one more. I think that should be all, but I didn't do a really deep dive. These just popped up while I was working on an on-call issue. I'll create a ticket as a follow up to do a second pass.

PR #1529, opened 2023-07-23
Change port 22 to 3389 for shodankey verification due false positive

angeorgieff:
Updated port 22 to 3389 for the shodankey verification as the port 22 verification check works with an invalid key or anonymously. 

By changing the port to 3389 it actually performs the verification. 

To verify, try this command
`curl -X GET "https://api.shodan.io/shodan/host/count?query=port:22&facets=org,os"` and observe the 200 response with data returned.

Now try the same request with port 3389 without a key and observe the 401 Unauthorized response code `curl -X GET "https://api.shodan.io/shodan/host/count?query=port:3389&f:22=org,os"`

### DIFF
--- a/pkg/detectors/shodankey/shodankey.go
+++ b/pkg/detectors/shodankey/shodankey.go
@@ -48,7 +48,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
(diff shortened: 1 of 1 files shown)


ahrav:
Hey @angeorgieff I think I completely missed this PR prior to the change I made to this detector. Would you mind confirming if the detector now works as expected. I updated the endpoint altogether which appears to return the correct result. If it's still not working for you i'll get this change in.  Appreciate the contribution, and I apologize for missing this earlier.

angeorgieff:
Hello @ahrav I just retested using version 3.48.0 which includes your update and it's working great! No need to merge this PR. Thank you 👍

PR #1533, opened 2023-07-24
Add commitsScanned metrics

bill-rich:
Track which commits have been scanned. Adding to source so that this doesn't need to be implemented in other places.

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -51,7 +51,12 @@ type Git struct {
 	jobID              int64
 	sourceMetadataFunc func(file, email, commit, timestamp, repository string, line int64) *source_metadatapb.MetaData
 	verify             bool
+	metrics            metrics
 	concurrency        *semaphore.Weighted
+}
+
+type metrics struct {
+	commitsScanned map[string]struct{}
 }
 
 func NewGit(sourceType sourcespb.SourceType, jobID, sourceID int64, sourceName string, verify bool, concurrency int,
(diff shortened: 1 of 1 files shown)


rosecodym (approved):
What makes this a "metric"/how is it going to be used? I initially assumed that as a "metric" it would be a counter or something, not a set.

pkg/sources/git/git.go
```diff
@@ -371,6 +383,7 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 			break
 		}
 		depth++
+		s.metrics.commitsScanned[commit.Hash] = struct{}{}
```
mcastorina, line 386: This seems like it would use a lot of memory for large repos. Just `trufflehog` would be 88KB (barring the overhead of maps).

bill-rich: Good point. I used the method from thog without really considering if it was the best. Switched to just counting since thats all that is used.

bill-rich:
> What makes this a "metric"/how is it going to be used? I initially assumed that as a "metric" it would be a counter or something, not a set.

Its used in some places to count the number of commits scanned. I think this method of using a map is left over from when the git log was handled differently. Switched it over to an int since thats all we really use.

rosecodym:
> > What makes this a "metric"/how is it going to be used? I initially assumed that as a "metric" it would be a counter or something, not a set.
> 
> Its used in some places to count the number of commits scanned. I think this method of using a map is left over from when the git log was handled differently. Switched it over to an int since thats all we really use.

I know you already merged, but to follow up on my question: What's going to _consume_ it?

PR #1535, opened 2023-07-24
[chore] - optimize chunker

ahrav:
-  Given the PeekSize is a constant and we are always appending it to the chunkBytes, we can make chunkBytes large enough to hold both ChunkSize and PeekSize from the start. This way, we can eliminate the append call altogether.

Screenshot 2023-07-24 at 11 37 38 AM

### DIFF
--- a/pkg/sources/chunker.go
+++ b/pkg/sources/chunker.go
@@ -12,6 +12,8 @@ const (
 	ChunkSize = 10 * 1024
 	// PeekSize is the size of the peek into the previous chunk.
 	PeekSize = 3 * 1024
+	// TotalChunkSize is the total size of a chunk with peek data.
+	TotalChunkSize = ChunkSize + PeekSize
 )
 
 // Chunker takes a chunk and splits it into chunks of ChunkSize.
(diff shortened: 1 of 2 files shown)


pkg/sources/chunker.go
```diff
@@ -26,20 +26,24 @@ func Chunker(originalChunk *Chunk) chan *Chunk {
 		r := bytes.NewReader(originalChunk.Data)
 		reader := bufio.NewReaderSize(bufio.NewReader(r), ChunkSize)
 		for {
-			chunkBytes := make([]byte, ChunkSize)
+			chunkBytes := make([]byte, ChunkSize+PeekSize)
 			chunk := *originalChunk
+			chunkBytes = chunkBytes[:ChunkSize]
```
mcastorina, line 31: I wonder if the performance would look like with just these changes.

`append` won't need to reallocate if there's enough capacity in the slice.

pkg/sources/chunker.go
```diff
@@ -26,20 +26,24 @@ func Chunker(originalChunk *Chunk) chan *Chunk {
...
 				break
 			}
-			peekData, _ := reader.Peek(PeekSize)
-			chunk.Data = append(chunkBytes[:n], peekData...)
-			if n > 0 {
-				chunkChan <- &chunk
-			}
-			if errors.Is(err, io.EOF) {
+			if n == 0 {
+				if errors.Is(err, io.EOF) {
```
bill-rich, line 37: This is going to `break` either way, so the extra check for `EOF` isn't necessary.

pkg/sources/chunker.go
bill-rich: Would it make sense to have the total `chunkBytes` + `peekBytes` have a different var name to avoid confusion?

ahrav: - append vs copy : very similar.
Screenshot 2023-07-24 at 11 37 38 AM

pkg/sources/chunker.go
ahrav: yea my bad, switched it to a continue for non-eof

ahrav: I think we only want to break for EOF and continue for non-EOF, is that correct? or do we break for both cases.

bill-rich (approved):
Nice improvement!

PR #1538, opened 2023-07-24
Replace aho-corasick library 

zricethezav:
A bug was discovered in the community slack channel when a npm token was not being detected. This was due to the fact that the existing ahocorasick library does not properly implement overlapping matches (see related issue here: https://github.com/petar-dambovaliev/aho-corasick/issues/12).

To test this bug try running an unverified scan on the follow content:

```
npm_token = "00000000-0000-0000-0000-000000000000"
```

You'll see it does not detect a secret. This is because `npm_` is currently taking precedence over the `npm` keyword due to https://github.com/trufflesecurity/trufflehog/blob/93c561f324f57e563d47374e07a29ef49209f27e/pkg/engine/engine.go#L156. However, changing this to `StandardMatch` does not fix it even though the library says it should support overlapping matches. 

Testing this on a few example repos yields pretty similar performance results

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	cloud.google.com/go/secretmanager v1.11.1
 	cloud.google.com/go/storage v1.31.0
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.11
+	github.com/BobuSumisu/aho-corasick v1.0.3
 	github.com/TheZeroSlave/zapsentry v1.17.0
 	github.com/aws/aws-sdk-go v1.44.83
 	github.com/bill-rich/disk-buffer-reader v0.1.7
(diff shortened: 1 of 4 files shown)


bill-rich (approved):

ahrav (approved):

0x1 (approved):

semvis123:
https://github.com/petar-dambovaliev/aho-corasick/pull/13
The previous library should be fixed with the above mentioned pull request.
From my testing [this](https://github.com/petar-dambovaliev/aho-corasick) library will consume way less memory, while the [other library](https://github.com/bobusumisu/aho-corasick) is slightly faster on larger inputs. If you want to prioritize memory usage you might want to consider switching back. But I recommend to run some benchmarks first.

semvis123:
Done 👍

PR #1547, opened 2023-07-25
Make Ahocorasick matching case insensitive

zricethezav:
This PR fixes a bug that prevents secrets from being detected if the keyword or chunk data case varies. 

`strings.ToLower` doesn't seem to have a huge effect on performance either.

```
Benchmark git scan on the rails/rails repo

### NEW
real 27.18s
user 58.14s 
sys  5.47s 
memory:753328KB 
cpu 234%
      62

### OLD
real 27.45s
user 56.43s 
sys  5.50s 
memory:750160KB 
cpu 225%
      62
```

To demonstrate this bug fix, try finding a private key

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -145,10 +145,14 @@ func Start(ctx context.Context, options ...EngineOption) *Engine {
 	// on keywords
 	keywords := []string{}
 	for _, d := range e.detectors[false] {
-		keywords = append(keywords, d.Keywords()...)
+		for _, kw := range d.Keywords() {
+			keywords = append(keywords, strings.ToLower(kw))
+		}
 	}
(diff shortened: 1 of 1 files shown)


mcastorina (approved):
I'm not sure we have any non-lowercase keywords, but good to have this in place.

zricethezav:
@mcastorina I didn't see any, but I agree. Currently, some secrets like private keys are being detected because  `https://github.com/BobuSumisu/aho-corasick` does not have a case insensitive option like `https://github.com/petar-dambovaliev/aho-corasick` and therefore when `BEING PRIVATE KEY` is scanned it fails to match on `private key` (pkey detector keyword). 

Note that the performance hit in the test is a tradeoff for correctness

PR #1550, opened 2023-07-26
Improve MongoDB connection string matching

### LINKED ISSUE #1512: MongoDB pattern does not capture connection string options
### Community Note

### Description

The MongoDB regex only captures the [Connection String](https://www.mongodb.com/docs/manual/reference/connection-string/) protocol, username, password, host, and port. It does not capture the `defaultauthdb` or `?options`, which could lead to valid secrets failing verification.

https://github.com/trufflesecurity/trufflehog/blob/f6aaa4924f576b518b1525b20b434328f9ac1e74/pkg/detectors/mongodb/mongodb.go#L24

For example, the CosmoDB connection string shown in #1511 only partially matches.

https://regex101.com/r/2tl5VU/1

## Problem to be Addressed

Improve false-negative detections of MongoDB connection strings.

The preferred solution would be to improve the regex to capture valid Connection Strings, 

https://www.mongodb.com/docs/manual/reference/connection-string/

## Additional Context

I've created an example regex that matches options. Disclaimer: It's a POC and not well-tested. 

https://regex101.com/r/2tl5VU/7

### References

N/A

rgmz:
This mostly fixes #1512, although it would be valuable to add real secrets to test verification of multiple hosts, connection options, etc.


rgmz:
@zricethezav this should be ready for review now.

pkg/detectors/mongodb/mongodb.go
```diff
@@ -61,9 +68,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			if timeout == 0 {
 				timeout = defaultTimeout
 			}
-			err := verifyUri(resMatch, timeout)
-			s1.Verified = err == nil
-			if !isErrDeterminate(err) {
+			isVerified, verificationErr := verifyUri(ctx, resMatch, timeout)
+			s1.Verified = isVerified
+			if !isErrDeterminate(verificationErr) {
 				s1.SetVerificationError(err, resMatch)
```
ahrav, line 74: nit: I _think_ this should be `verificationErr`.

ahrav (approved):
Thanks @rgmz, sorry for just seeing this PR. Thanks for the detailed test cases. I’ve added some test credentials, and the tests are all passing.

pkg/detectors/mongodb/mongodb.go
rgmz: It absolutely should be.

PR #1551, opened 2023-07-26
Add support for storing handler metadata

mcastorina:
Currently it's a unstructured map[string]string that gets copied into a Result.ExtraData after detection, but we might want to revisit this decision when we get more handlers.


dustin-decker:
Awesome. Could we also surface the compression type being used?

PR #1552, opened 2023-07-26
Fix runtime error when scanning Gist comments

### LINKED ISSUE #1539: `panic: runtime error: index out of range [2] with length 2` when using `--include-members`
### TruffleHog Version
`dev` -- because i manually compiled on windows

### Trace Output

https://gist.github.com/selfagency/1b25d4da9c86e8803d8bbd860a976a46

### Expected Behavior

For the command to have succeeded

### Actual Behavior

```
panic: runtime error: index out of range [2] with length 2

goroutine 130 [running]:
github.com/trufflesecurity/trufflehog/v3/pkg/sources/github.(*Source).scanComments(0xc0001a6c00, {0x2789600?, 0xc001101560}, {0xc001568f80, 0x3c}, 0xc00038b4e0?)
        H:/Development/trufflehog/pkg/sources/github/github.go:945 +0x595
github.com/trufflesecurity/trufflehog/v3/pkg/sources/github.(*Source).scan.func1()
        H:/Development/trufflehog/pkg/sources/github/github.go:675 +0x77f
golang.org/x/sync/errgroup.(*Group).Go.func1()
        C:/Users/dsieradski/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
        C:/Users/dsieradski/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:72 +0xa5
```

### Steps to Reproduce

`trufflehog github --org=$env:GITHUB_ORG --include-members --token=$env:GITHUB_TOKEN --trace`

## Environment
 * OS: Windows 
 * Version: 10

## Additional Context
There are a large number of repos being scanned under a single user (me)

rgmz:
This fixes #1539.

Lingering questions:

- [ ] Are there unit/integration tests for comment scanning?
- [x] Should we fetch information about each Gist, such as the visibility? Doing so will provide more context in results, but will also eat into any rate limits.


pkg/sources/github/github.go
```diff
@@ -966,79 +966,118 @@ func (s *Source) scanComments(ctx context.Context, repoPath string, chunksChan c
...
+		// GitHub Gist URL.
+		var gistId string
+		if len(trimmedURL) == 2 {
+			// https://gist.github.com/<id>
+			gistId = trimmedURL[1]
+		} else if len(trimmedURL) == 3 {
+			// https://gist.github.com/<owner>/<id>
+			gistId = trimmedURL[2]
+		} else {
+			return fmt.Errorf("failed to parse Gist URL: '%s'", repoURL.String())
```
mcastorina, line 979: Returning an error here (or anywhere in this function) will short-circuit scanning the git repository. Not that it's wrong to return an error, but rather we either need to change the order of scanning or change the logic to not return here:

https://github.com/trufflesecurity/trufflehog/blob/d0705c01d8e15db7b9f4b17f77841860d8d60dda/pkg/sources/github/github.go#L700-L708

pkg/sources/github/github.go
```diff
@@ -1109,6 +1148,39 @@ func (s *Source) chunkPullRequestComments(ctx context.Context, repo string, comm
...
+			SourceName: s.name,
+			SourceID:   s.SourceID(),
+			SourceType: s.Type(),
+			SourceMetadata: &source_metadatapb.MetaData{
+				Data: &source_metadatapb.MetaData_Github{
+					Github: &source_metadatapb.Github{
+						Link:     sanitizer.UTF8(comment.GetURL()),
+						Username: sanitizer.UTF8(comment.GetUser().GetLogin()),
+						Email:    sanitizer.UTF8(comment.GetUser().GetEmail()),
+						//Repository: sanitizer.UTF8(gist),
```
mcastorina, line 1165: Any reason `Repository` is commented out here?

pkg/sources/github/github.go
```diff
@@ -1109,6 +1148,39 @@ func (s *Source) chunkPullRequestComments(ctx context.Context, repo string, comm
...
+						Timestamp: sanitizer.UTF8(comment.GetCreatedAt().String()),
+						//Visibility: s.visibilityOf(ctx, repoPath),
```
mcastorina, line 1167: We might be able to address this later by only hydrating some metadata *after* finding a secret in it. The engine doesn't support this yet though.

For now it would be good to leave a comment why we're not getting the visibility (and remove the TODO at the top of the function).

mcastorina (commented):
This looks good to me, though should we also have a flag for configuring scanning gist comments?

pkg/sources/github/github.go
rgmz: An oversight on my part. I've address this and the other points.

rgmz:
> Yes, it should be an opt-in flag because it is sensitive to API rate limits.

Scanning Gist comments or scanning comments in general?

mcastorina:
> > Yes, it should be an opt-in flag because it is sensitive to API rate limits.
> 
> Scanning Gist comments or scanning comments in general?

The other comment scanning code is already opt-in, so we can follow that pattern:

https://github.com/trufflesecurity/trufflehog/blob/0fff65e158adf4c8d10ca44590d645568382c4f9/pkg/sources/github/github.go#L71-L72

rgmz:
> The other comment scanning code is already opt-in, so we can follow that pattern:

Done.

mcastorina (approved):
lgtm, thanks for fixing this!

PR #1562, opened 2023-07-27
Support fatal errors in job reports

mcastorina:


pkg/sources/source_manager.go
```diff
@@ -368,37 +367,31 @@ func (api *headlessAPI) GetJobID(ctx context.Context, id int64) (int64, error) {
 
 // mgrUnitReporter implements the UnitReporter interface.
 type mgrUnitReporter struct {
-	unitCh       chan SourceUnit
-	unitErrs     []error
-	unitErrsLock sync.Mutex
+	unitCh chan SourceUnit
+	report *JobReport
```
rosecodym, line 371: a thought: holding a reference to an entire report here (and also down in `mgrChunkReporter`) might be unnecessarily tight coupling, given all the other stuff in a `JobReport`. you could separate them with an interface on `JobReport` that just accepts errors - at the cost of adding another abstraction. what do you think?

pkg/sources/source_manager.go
```diff
@@ -312,17 +311,18 @@ func (s *SourceManager) runWithUnits(ctx context.Context, handle handle, source
 		unitPool.SetLimit(s.concurrentUnits)
 	}
 	for unit := range reporter.unitCh {
+		unit := unit
 		reporter := &mgrChunkReporter{
-			unitID:  unit.SourceUnitID(),
+			unit:    unit,
```
rosecodym, line 316: why did you need to switch this to be a full unit?

pkg/sources/source_manager.go
mcastorina: yeah, I think that will help define api expectations (which I'm still trying to figure out tbh).

pkg/sources/source_manager.go
mcastorina: The `report` object has a method `AddChunkError` that needs a full unit. I think it logically makes sense even though we'll only use the `SourceUnitID() string`

pkg/sources/job_report.go
```diff
@@ -1,37 +1,269 @@
 package sources
 
+//go:generate mockgen --source=./job_report.go --destination=mock_job_report_test.go --package=sources
```
rosecodym, line 3: what's this?

pkg/sources/source_manager.go
rosecodym: so this still looked weird to me, and i realized that it's because `JobReport` isn't really a job "report" anymore. it's, like, job runtime data? or something? `InProgressJob`?

rosecodym (approved):
some questions, nothing blocking. although i would be very happy with a rename of `JobReport`, either in this PR or as a fast follow.

pkg/sources/source_manager.go
```diff
@@ -23,14 +22,11 @@ type handle int64
 type SourceInitFunc func(ctx context.Context, sourceID int64, jobID int64) (Source, error)
 
 type SourceManager struct {
-	api apiClient
+	api   apiClient
+	hooks []JobReportHook
```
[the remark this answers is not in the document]
mcastorina: Yes, the hooks are set at `SourceManager` initialization and not modified again.

pkg/sources/source_manager.go
```diff
@@ -283,25 +242,36 @@ func (s *SourceManager) runWithoutUnits(ctx context.Context, handle handle, sour
...
+	unitReporter := &mgrUnitReporter{
 		unitCh: make(chan SourceUnit),
+		report: report,
+	}
+	fatalErr := make(chan error, 1)
+	catchFirstFatal := func(err error) {
+		select {
+		case fatalErr <- err:
+		default:
+		}
```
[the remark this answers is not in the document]
mcastorina: Attempts to write the `err` into the `fatalErr` channel, throwing away all but the first one to make it (since `fatalErr` has a buffer size of 1).

I'll add a comment to make it clearer.

pkg/sources/job_report.go
mcastorina: Go supports running commands marked by this directive to generate *stuff*.

In this case it is running the command [mockgen](https://github.com/uber-go/mock) which will create mocks of all the interfaces in the source file and write it to the destination file.

To generate that file, I ran `go generate ./pkg/sources/...`.

More info: [go:generate](https://www.practical-go-lessons.com/post/what-is-go-generate-and-how-to-use-it-ccava8v5toqc70ipi1jg)

I chose to use it because the interface was large and I really just wanted to test whether the methods got called with the expected values, which `mockgen` supports.

PR #1563, opened 2023-07-27
Identify transient AWS verification failures

rosecodym:
It turns out that `GetCallerIdentity` returns a surprising quantity of transient, false-negative 403 responses that carry the `SignatureDoesNotMatch` error reason. I don't know why this is happening, but their transient nature makes them indeterminate verification failures and they should be flagged as such. The AWS detector has therefore been modified to specifically look for the `InvalidClientTokenId` error reason in 403 responses and mark all other responses as indeterminate.

In addition to the functional changes this PR contains some updates to the test code that allow us to test them.

### DIFF
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -18,7 +18,8 @@ import (
 )
 
 type scanner struct {
-	skipIDs map[string]struct{}
+	verificationClient *http.Client
+	skipIDs            map[string]struct{}
 }
 
 func New(opts ...func(*scanner)) *scanner {
@@ -48,7 +49,7 @@ func WithSkipIDs(skipIDs []string) func(*scanner) {
 var _ detectors.Detector = (*scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultVerificationClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	// Key types are from this list https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids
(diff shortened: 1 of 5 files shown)


pkg/detectors/aws/aws.go
```diff
@@ -123,8 +124,9 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				host := "sts.amazonaws.com"
 				region := "us-east-1"
 				endpoint := "https://sts.amazonaws.com"
-				datestamp := time.Now().UTC().Format("20060102")
-				amzDate := time.Now().UTC().Format("20060102T150405Z0700")
+				now := time.Now().UTC()
+				datestamp := now.Format("20060102")
+				amzDate := now.Format("20060102T150405Z0700")
```
rosecodym, line 129: This wasn't related to the actual change. I just fixed it (calling `time.Now()` twice in succession causes people reading the code to do a double-take) while I was here.

pkg/detectors/aws/aws.go
```diff
@@ -245,6 +266,15 @@ func awsCustomCleanResults(results []detectors.Result) []detectors.Result {
...
 }
 
+type awsError struct {
+	Code    string
+	Message string
+}
+
+type awsErrorResponseBody struct {
+	Error awsError
+}
```
mcastorina, line 276: **Nit:** Explicitly adding the `json` tags would help signal that this struct is used for encoding / decoding JSON. It'll also decouple the struct member names from that process.

```go
type awsError struct {
    Code    string `json:"Code"`
    Message string `json:"Message"`
}

type awsErrorResponseBody struct {
    Error awsError `json:"Error"`
}
```

pkg/detectors/aws/aws_test.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
+var unverifiedSecretClient = common.ConstantResponseHttpClient(403, `{"Error": {"Code": "InvalidClientTokenId"} }`)
```
mcastorina, line 20: Are the values here actually capitalized in the wild?

mcastorina (approved):

pkg/detectors/aws/aws_test.go
rosecodym: they sure are!

rosecodym: although it would be a good idea to do the comparison case-insensitively

trufflesteeeve (approved):

PR #1564, opened 2023-07-27
Do not nest transports for Github installation client

rosecodym:
#1454 modified one of the Github enumeration code paths in a way that broke an integration test by causing one client's transport to be used for the construction of a different client, causing authentication failures. This saves the original transport for use, fixing the test.

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -554,28 +554,13 @@ func (s *Source) enumerateWithApp(ctx context.Context, apiEndpoint string, app *
 		return nil, errors.New(err)
 	}
 
-	// This client is used for most APIs.
-	itr, err := ghinstallation.New(
-		s.httpClient.Transport,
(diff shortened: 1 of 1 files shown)


zricethezav (approved):

pkg/sources/github/github.go
```diff
@@ -574,7 +575,7 @@ func (s *Source) enumerateWithApp(ctx context.Context, apiEndpoint string, app *
 	// This client is required to create installation tokens for cloning.
 	// Otherwise, the required JWT is not in the request for the token :/
 	appItr, err := ghinstallation.NewAppsTransport(
```
bill-rich, line 577: Instead of storing the original transport in a `originalTransport`, can we just move this ahead of where `s.httpClient.Transport` is modified?

In either case, a comment of why this var or order of things is necessary could be useful in the future.

PR #1566, opened 2023-07-27
Add path (location of file) to Google Drive proto 

zubairk14:
Add `path` to Google Drive proto to help users identify where the exposed secret lives in their Google Drive.


0x1 (approved):
nice

PR #1570, opened 2023-07-28
Fix gitparse from panicking on a nil-pointer

mcastorina:
Question: why is `currentDiff` a pointer? Can it be made an object so we don't have to worry about initializing it?


bill-rich (approved):

bill-rich:
> Question: why is `currentDiff` a pointer? Can it be made an object so we don't have to worry about initializing it?

We'd need to copy the content of `currentDiff` each time a diff section is finished and a new one started. With it as a pointer, we can just make a new diff and start working with that while leaving the finished one as-is.

PR #1571, opened 2023-07-28
Add test case to gitparse that previously caused a panic

dustin-decker:
Adds a test case to gitparse that caused a panic which was fixed by #1570


bill-rich:
Tested and the #1570 fix handles this. It looked to me like it shouldn't miss anything at first glance. If `currentDiff` is `nil`, then it shouldn't be able to exceed the max diff side, unless I'm overlooking something.

pkg/gitparse/gitparse_test.go
```diff
@@ -2,11 +2,20 @@ package gitparse
...
 	"testing"
 	"time"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/log"
+)
+
+var (
+	//go:embed testdata/panic_diff_1.txt
+	panicDiff1 []byte
```
mcastorina, line 18: Cool, I didn't know you could do this!

pkg/gitparse/gitparse_test.go
```diff
@@ -20,6 +29,20 @@ type testCase struct {
 	function func(bool, ParseState, []byte) bool
 }
 
+func TestParsePanicDiff(t *testing.T) {
+	// setup logger
+	logger, sync := log.New("forager/scanner",
+		log.WithConsoleSink(os.Stderr),
+	)
+	defer func() { _ = sync() }()
+	context.SetDefaultLogger(logger)
```
mcastorina, line 38: This log setup can probably be removed - we now have a default global logger that will be used from `context.Background()`.

pkg/gitparse/gitparse_test.go
```diff
@@ -20,6 +29,20 @@ type testCase struct {
...
+
+	ctx := context.Background()
+	parser := NewParser()
+	commitCh := make(chan Commit, 64)
+	parser.FromReader(ctx, io.NopCloser(bytes.NewReader(panicDiff1)), commitCh, false)
```
mcastorina, line 43: Is the idea of this test that this line *won't* panic? What do you think about using [assert.NotPanics](https://pkg.go.dev/github.com/stretchr/testify/assert#NotPanics) to make that explicit?

mcastorina (approved):

PR #1573, opened 2023-07-28
gitparse: Use an object for currentDiff and currentCommit

mcastorina:
This is safer than relying on state to ensure we don't get a nil pointer dereference.

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -236,7 +236,7 @@ func (c *Parser) executeCommand(ctx context.Context, cmd *exec.Cmd, isStaged boo
 func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan chan Commit, isStaged bool) {
 	outReader := bufio.NewReader(stdOut)
 	var (
-		currentCommit *Commit
+		currentCommit Commit
 		currentDiff   Diff
 
 		totalLogSize int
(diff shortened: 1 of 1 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -309,8 +309,8 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 			if currentCommit == nil {
 				currentCommit = &Commit{}
 			}
-			if currentDiff != nil && currentDiff.Content.Len() > 0 {
-				currentCommit.Diffs = append(currentCommit.Diffs, *currentDiff)
+			if currentDiff.Content.Len() > 0 {
```
0x1, line 312: why does making it an object allow us to avoid nil checking? is it just going to default to an empty struct?

0x1 (approved):
one question, otherwise looks gucci

pkg/gitparse/gitparse.go
bill-rich: Yep. When its initialized it is an empty struct instead of a pointer to `nil`.

ahrav: In this case @bill-rich would it be a pointer to nil or a nil pointer? I think the latter?
This is my understanding:
```
var i *int = nil // i is a nil pointer
var p **int = &i // p is a pointer to a nil pointer || closest thing to pointer to nil lol
```

bill-rich: You're right. I worded that poorly. Should have been "pointing to `nil`".

rosecodym (approved):
this lgtm and i'm _assuming_ it doesn't run afoul of @bill-rich's commentary re: pointers on #1570 (because he's commenting here without objecting) but maybe check with him if you haven't already

mcastorina:
> this lgtm and i'm _assuming_ it doesn't run afoul of @bill-rich's commentary re: pointers on #1570 (because he's commenting here without objecting) but maybe check with him if you haven't already

yep, chatted out-of-band and we were already making a copy when we dereferenced it to put in the channel.

PR #1574, opened 2023-07-28
implement indeterminate LDAP verification

rosecodym:
This PR implements tri-state verification for the LDAP detector. This implementation looks for network errors to explicitly flag as indeterminate, rather than authentication errors to explicitly flag as determinate; this is because the error that occurs from authentication failures doesn't appear to have its own type and I didn't want to have to match on the error message text.

### DIFF
--- a/pkg/detectors/ldap/ldap.go
+++ b/pkg/detectors/ldap/ldap.go
@@ -95,10 +95,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			err := verifyLDAP(username, password, ldapURL)
-			s1.Verified = err == nil
-			if !isErrDeterminate(err) {
-				s1.VerificationError = err
+			verificationError := verifyLDAP(username, password, ldapURL)
+
+			s1.Verified = verificationError == nil
+			if !isErrDeterminate(verificationError) {
+				s1.VerificationError = verificationError
 			}
 		}
 
--- a/pkg/detectors/ldap/ldap.go
+++ b/pkg/detectors/ldap/ldap.go
@@ -65,10 +65,10 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				}
 
 				if verify {
-					err := verifyLDAP(username[1], password[1], ldapURL)
-					s1.Verified = err == nil
-					if !isErrDeterminate(err) {
-						s1.VerificationError = err
+					verificationErr := verifyLDAP(username[1], password[1], ldapURL)
+					s1.Verified = verificationErr == nil
+					if !isErrDeterminate(verificationErr) {
+						s1.VerificationError = verificationErr
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


pkg/detectors/ldap/ldap.go
```diff
@@ -98,7 +108,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		switch e.Err.(type) {
+		case *net.OpError:
+			return false
+		}
+	}
+
+	return true
+}
+
+func verifyLDAP(ctx context.Context, username, password string, ldapURL *url.URL) error {
```
ahrav, line 123: nit: I think we can ignore the `ctx` since it's not used in this fxn.
`verifyLDAP(_, context.Context,...)`

ahrav (approved):

pkg/detectors/ldap/ldap.go
```diff
@@ -98,7 +108,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	switch e := err.(type) {
+	case *ldap.Error:
+		switch e.Err.(type) {
+		case *net.OpError:
+			return false
+		}
+	}
+
+	return true
+}
```
mcastorina, line 121: I tried changing this function to the following:

```go
func isErrDeterminate(err error) bool {
	var target *net.OpError
	return !errors.As(err, &target)
}
```

and it passed the tests, though this function also passed the tests:

```go
func isErrDeterminate(err error) bool {
	return true
}
```

`errors.As` is a bit more flexible since it will `Unwrap` errors.

pkg/detectors/ldap/ldap.go
```diff
@@ -63,7 +65,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				}
 
 				if verify {
-					s1.Verified = verifyLDAP(ctx, username[1], password[1], ldapURL)
+					err := verifyLDAP(username[1], password[1], ldapURL)
+					s1.Verified = err == nil
+					if !isErrDeterminate(err) {
+						s1.VerificationError = err
+					}
```
mcastorina, line 72: **Nit:** The `s1.Verified = err == nil` looks weird, but I'm not sure this is better. 

```go
if err := verifyLDAP(username[1], password[1], ldapURL); err == nil {
    s1.Verified = true
} else if !isErrDeterminate(err) {
    s1.VerificationError = err
}
```

It's *technically* different behavior since we don't explicitly set `s1.Verified = false` for the `err != nil` case.

mcastorina (approved):

pkg/detectors/ldap/ldap.go
rosecodym: err, that second function shouldn't have passed...

mcastorina: sorry for the confusion, I didn't run the integration tests 😕

PR #1575, opened 2023-07-28
add tri-state verification to mongodb detector

rosecodym:


pkg/detectors/mongodb/mongodb.go
```diff
@@ -46,30 +51,47 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-				defer cancel()
-				client, err := mongo.Connect(ctx, options.Client().ApplyURI(resMatch))
-				if err != nil {
-					return
-				}
-				defer func() {
-					if err := client.Disconnect(ctx); err != nil {
-						return
-					}
-				}()
```
rosecodym, line 60: @ahrav i _think_ this was just discarding the result of `client.Disconnect` so i simplified it a bit when i moved it, but give a shout if the new version looks wrong

ahrav: looks good!

pkg/detectors/mongodb/mongodb.go
```diff
@@ -46,30 +51,47 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				s1.VerificationError = err
+			}
 		}
 		results = append(results, s1)
 	}
 
 	return results, nil
 }
 
+func isErrDeterminate(err error) bool {
```
ahrav, line 70: suggestion: I think you can simplify this with:
```
if e, ok := err.(topology.ConnectionError); ok {
		_, ok = e.Unwrap().(*auth.Error)
		return ok
	}
	return false
```

or i guess if we want to use the errors pkg:
```
func isErrDeterminate(err error) bool {
	var connErr topology.ConnectionError
	if errors.As(err, &connErr) {
		var authErr *auth.Error
		return errors.As(connErr.Unwrap(), &authErr)
	}
	return false
}
```

pkg/detectors/mongodb/mongodb.go
```diff
@@ -46,30 +51,47 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+}
+
+func verifyUri(uri string, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	client, err := mongo.Connect(ctx, options.Client().ApplyURI(uri))
+	if err != nil {
+		return err
+	}
+	defer func() {
```
ahrav, line 89: nit: since we are ignoring the err here anyways we can inline it:
`defer client.Disconnect(ctx)`

ahrav (approved):

pkg/detectors/mongodb/mongodb.go
rosecodym: huh, i could have sworn the linter didn't let me do that...

pkg/detectors/mongodb/mongodb.go
rosecodym: i actually specifically wrote this not to be as terse as possible, but to make it easy to extend with other error cases in the future. do you find it hard to read as it is?

pkg/detectors/mongodb/mongodb.go
ahrav: might've complained that you are ignoring the err i suppose.

pkg/detectors/mongodb/mongodb.go
ahrav: ahh okay, i think that makes sense. I sorta had the feeling, but wasn't sure. If we leave the switches might be worth adding the default case to each, no mandatory though. One of the the go linters happens to yell at me when i have a switch with no default. 🤷

pkg/detectors/mongodb/mongodb.go
rosecodym: yeah `golangci-lint` is not about this

PR #1577, opened 2023-07-29
[bug] - fix data races

ahrav:
- Add test to catch all the data races below
- Multiple repos can concurrently use the git source, therefore we need to make the metrics counter atomic.
NOTE: We should probably look for a less hacky solution for the long term. This is just to temporarily fix the issues.

Screenshot 2023-07-29 at 3 08 18 PM

- Modifying the user and token also needs to be protected
Screenshot 2023-07-29 at 7 16 45 PM

- Modifying the scanOptions also need protection
Screenshot 2023-07-29 at 7 29 19 PM

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -90,7 +90,7 @@ func (s *Source) setScanOptions(base, head string) {
 	s.scanOptMu.Lock()
 	defer s.scanOptMu.Unlock()
 	s.scanOptions.BaseHash = base
-	s.scanOptions.BaseHash = head
+	s.scanOptions.HeadHash = head
 }
 
 // Ensure the Source satisfies the interfaces at compile time
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -44,30 +44,38 @@ const (
 )
 
 type Source struct {
-	name                 string
-	githubUser           string
-	githubToken          string
-	sourceID             int64
-	jobID                int64
-	verify               bool
-	repos                []string
-	members              []string
-	orgsCache            cache.Cache
-	filteredRepoCache    *filteredRepoCache
-	memberCache          map[string]struct{}
-	repoSizes            repoSize
-	totalRepoSize        int // total size in bytes of all repos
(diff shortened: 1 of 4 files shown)


pkg/sources/github/github.go
```diff
@@ -78,6 +86,13 @@ func (s *Source) WithScanOptions(scanOptions *git.ScanOptions) {
 	s.scanOptions = scanOptions
 }
 
+func (s *Source) setScanOptions(base, head string) {
+	s.scanOptMu.Lock()
+	defer s.scanOptMu.Unlock()
+	s.scanOptions.BaseHash = base
+	s.scanOptions.BaseHash = head
```
rosecodym, line 93: i'm guessing there's a typo here

pkg/sources/github/github.go
```diff
@@ -678,6 +693,7 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
 			path, repo, err = s.cloneRepo(ctx, repoURL, installationClient)
 			if err != nil {
 				scanErrs.Add(err)
+				return nil
```
rosecodym, line 696: is this related to the data race?

pkg/sources/github/repo.go
```diff
@@ -50,12 +50,15 @@ func (s *Source) cloneRepo(
 
 	case *sourcespb.GitHub_Token:
 		// We never refresh user provided tokens, so if we already have them, we never need to try and fetch them again.
+		s.userMu.Lock()
```
rosecodym, line 53: If `s.userAndToken` panics, this will be locked forever, right? Do you think it's worth breaking out a new function (inline or otherwise) so you can use `defer`?

rosecodym (changes requested):
i think there's a small typo in addition to my other non-blocking questions

pkg/sources/github/github.go
ahrav: This was causing a test to fail. If we can't clone a repo we should not try to scan.

rosecodym (approved):

PR #1578, opened 2023-07-30
[chore] - Do not skip scanning commits if scanning comments errors

ahrav:
- An error while scanning comments should not prevent us from scanning the repo. We should simply log the error and try to scan the repo.


zubairk14:
oooh good idea - did you run into situation where this was causing problems?

ahrav:
> oooh good idea - did you run into a situation where this was causing problems?

I was fixing a separate issue for a data race and then stumbled upon this. The actual chance where we are not able to scan comments should be really really low in practice, but figured i'd just update it while was poking around this part of the code.

mcastorina:
https://github.com/trufflesecurity/trufflehog/pull/1552 also partially addresses this by giving precedence to scanning the repo over the comments.

PR #1579, opened 2023-07-30
[bug] - fix shodan detector

ahrav:
Screenshot 2023-07-29 at 9 43 07 PM

### DIFF
--- a/pkg/detectors/shodankey/shodankey.go
+++ b/pkg/detectors/shodankey/shodankey.go
@@ -2,6 +2,8 @@ package shodankey
 
 import (
 	"context"
+	"encoding/json"
+	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


mcastorina:
If you're up for it, we could add indeterminacy fixes too.

rosecodym:
yeah i'd love for someone other than me to go through the process and see how it feels

ahrav:
> If you're up for it, we could add indeterminacy fixes too.

I'll add the indeterminacy in a separate PR as I probably won't get to it until this Friday during hack time. I was hoping to get this out before then given we have a handful of individuals that have mentioned this detector isn't working correctly.

rosecodym (approved):
what's the problem that this solves?

PR #1582, opened 2023-07-31
create hidden debug flag to disable overseer

zubairk14:
support users to attach a debugger to a local trufflehog scanner


rosecodym (approved):
works great in goland! thanks!

main.go
```diff
@@ -38,6 +38,7 @@ var (
 	debug               = cli.Flag("debug", "Run in debug mode.").Bool()
 	trace               = cli.Flag("trace", "Run in trace mode.").Bool()
 	profile             = cli.Flag("profile", "Enables profiling and sets a pprof and fgprof server on :18066.").Bool()
+	LocalDev            = cli.Flag("local-dev", "Hidden feature to disable overseer for local dev.").Hidden().Bool()
```
trufflesteeeve, line 41: __optional nitpick__: Since this variable is only used in this file, we probably don't need to export it. So renaming to `localDev` would be good.

PR #1584, opened 2023-07-31
Github Oauth2 verification

bill-rich:


pkg/detectors/github/github.go
```diff
@@ -48,13 +56,49 @@ type userRes struct {
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"ghp_", "gho_", "ghu_", "ghs_", "ghr_", "github_pat_"}
+	return []string{"ghp_", "gho_", "ghu_", "ghs_", "ghr_", "github_pat_", "github"}
```
bill-rich, line 59: `github` matches means a lot of matches. Something like `github_client` would cut that down, but could miss creds. I erred on the side of not missing anything, but looking for other opinions.

pkg/detectors/github/github.go
```diff
@@ -48,13 +56,49 @@ type userRes struct {
...
+		}
+		for _, secretMatch := range oauth2ClientSecretMatches {
+			if len(secretMatch) != 2 {
+				continue
+			}
+
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_Github,
+				Raw:          []byte(idMatch[1]),
+			}
```
mcastorina, line 82: If my understanding is correct, we should also have a `RawV2: []byte(idMatch[i] + secretMatch[1])`.

mcastorina (commented):
Do you think this should be a completely separate detector?

pkg/detectors/github/github.go
zricethezav: I'm with you, I think erring on the side of not missing anything makes sense. And to clarify a little bit here, the `github` keyword would be associated with oauth creds right? Would it make sense to use `detectors.PrefixRegex` for the oauth regexes?

bill-rich:
> Do you think this should be a completely separate detector?

I could see either way on this one. I went with adding it to the existing detector to avoid confusion on github cred types, but I could see being more specific being useful also.

pkg/detectors/github/github.go
bill-rich: Good call. Switched.

pkg/detectors/github_oauth2/github_oauth2.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type Scanner struct{ detectors.EndpointSetter }
+
+// Ensure the Scanner satisfies the interfaces at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.Versioner = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+func (Scanner) Version() int            { return 2 }
```
mcastorina, line 21: I don't think this needs to be versioned (yet) since it's the only `GitHubOauth2` detector.

bill-rich: Good call. Definitely not v2 at least

pkg/detectors/github_oauth2/github_oauth2.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		for _, secretMatch := range oauth2ClientSecretMatches {
+			if len(secretMatch) != 2 {
+				continue
+			}
+
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_GitHubOauth2,
+				Raw:          []byte(idMatch[1]),
+				RawV2:        []byte(idMatch[1] + secretMatch[1]),
```
dustin-decker, line 58: You can make `Raw: []byte(idMatch[1] + secretMatch[1])` and filling omit RawV2
The only reason why RawV2 exists is because we did not capture all of the components of some credentials the first time in the Raw field.

cc @mcastorina for your understanding as well

pkg/detectors/github/github.go
dustin-decker: https://github.com/trufflesecurity/trufflehog/pull/1584#discussion_r1281234786

pkg/detectors/github_oauth2/github_oauth2.go
bill-rich: Won't that cause confusion in OSS where the raw value is shown?

dustin-decker: Oh, you're right, I didn't consider that aspect. Keep it how it is then.

pkg/detectors/github_oauth2/github_oauth2.go
```diff
@@ -0,0 +1,84 @@
...
+	"golang.org/x/oauth2/github"
+)
+
+type Scanner struct{ detectors.EndpointSetter }
+
+// Ensure the Scanner satisfies the interfaces at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+func (Scanner) DefaultEndpoint() string { return "https://api.github.com" }
```
mcastorina, line 20: **Nit (non-blocking):** It doesn't look like we're using this endpoint customizer either. I think that can be added in another PR though.

mcastorina (approved):

PR #1586, opened 2023-08-01
Use SourceManager in engine

mcastorina:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -12,7 +12,6 @@ import (
 
 	ahocorasick "github.com/BobuSumisu/aho-corasick"
 	lru "github.com/hashicorp/golang-lru"
-	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
@@ -69,10 +68,9 @@ type Engine struct {
 	prefilter ahocorasick.Trie
 
 	// Engine synchronization primitives.
-	chunks               chan *sources.Chunk
+	sourceManager        *sources.SourceManager
 	results              chan detectors.ResultWithMetadata
 	detectableChunksChan chan detectableChunk
-	sourcesWg            *errgroup.Group
 	workersWg            sync.WaitGroup
 	wgDetectorWorkers    sync.WaitGroup
 	WgNotifier           sync.WaitGroup
@@ -273,10 +271,8 @@ func Start(ctx context.Context, options ...EngineOption) (*Engine, error) {
 	}
 
 	e := &Engine{
-		chunks:               make(chan *sources.Chunk, defaultChannelBuffer),
 		detectableChunksChan: make(chan detectableChunk, defaultChannelBuffer),
 		results:              make(chan detectors.ResultWithMetadata, defaultChannelBuffer),
-		sourcesWg:            &errgroup.Group{},
 		dedupeCache:          cache,
 		printer:              new(output.PlainPrinter), // default printer
 		metrics:              runtimeMetrics{Metrics: Metrics{scanStartTime: time.Now()}},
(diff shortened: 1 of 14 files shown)


mcastorina:
https://github.com/trufflesecurity/trufflehog/pull/1518 should be reviewed / merged first. TIA

rosecodym:
if you set the base branch of this pr to be your other pr then this pr's diff won't include that pr's changes. is that what you're going for here?

pkg/sources/source_manager.go
```diff
@@ -232,7 +249,10 @@ func (s *SourceManager) runWithoutUnits(ctx context.Context, handle handle, sour
 		defer wg.Done()
 		for chunk := range ch {
 			report.ReportChunk(nil, chunk)
-			_ = common.CancellableWrite(ctx, s.outputChunks, chunk)
+			if err := common.CancellableWrite(ctx, s.outputChunks, chunk); err != nil {
+				report.ReportError(Fatal{err})
```
rosecodym, line 253: What makes this fatal?

pkg/sources/job_progress.go
```diff
@@ -163,11 +166,15 @@ func (jp *JobProgress) Finish() {
 }
 func (jp *JobProgress) Done() <-chan struct{} { return jp.ctx.Done() }
 func (jp *JobProgress) ReportUnit(unit SourceUnit) {
-	atomic.AddUint64(&jp.metrics.TotalUnits, 1)
+	jp.metricsLock.Lock()
+	jp.metrics.TotalUnits++
+	jp.metricsLock.Unlock()
```
rosecodym, line 171: Why aren't atomics enough here?

rosecodym (approved):
some non-blocking questions but this looks like a pretty smooth transition

pkg/sources/job_progress.go
rosecodym: ah i see, there's a read elsewhere that this needs to guard against

pkg/engine/engine.go
```diff
@@ -276,8 +272,11 @@ func Start(ctx context.Context, options ...EngineOption) (*Engine, error) {
 	}
 	ctx.Logger().V(3).Info("engine started", "workers", e.concurrency)
 
-	// Limit number of concurrent goroutines dedicated to chunking a source.
-	e.sourcesWg.SetLimit(int(e.concurrency))
+	// Create SourceManager.
+	e.sourceManager = sources.NewManager(
+		sources.WithConcurrency(int(e.concurrency)),
+		sources.WithConcurrentUnits(int(e.concurrency)),
```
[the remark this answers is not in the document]
mcastorina: Great point, I'll update it to `WithConcurrentSources`

pkg/sources/job_progress.go
```diff
@@ -196,7 +205,12 @@ func (jp *JobProgress) EndEnumerating(end time.Time) {
 func (jp *JobProgress) Snapshot() JobProgressMetrics {
 	jp.metricsLock.Lock()
 	defer jp.metricsLock.Unlock()
-	return jp.metrics
+
+	metrics := jp.metrics
+	metrics.Errors = make([]error, len(metrics.Errors))
+	copy(metrics.Errors, jp.metrics.Errors)
+
+	return metrics
```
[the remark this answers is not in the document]
mcastorina: Not that I'm aware of 😕

pkg/sources/source_manager.go
mcastorina: I have somewhat arbitrarily chosen to make context cancellations fatal errors so that they will surface as an error from a run, unlike errors we normally encounter but continue scanning.

pkg/engine/gcs_test.go
```diff
@@ -80,7 +80,10 @@ func TestScanGCS(t *testing.T) {
 				t.Errorf("ScanGCS() got: %v, want: %v", err, nil)
 				return
 			}
-			assert.Nil(t, e.Finish(ctx))
+			if err := e.Finish(ctx); err != nil && !test.wantErr && !strings.Contains(err.Error(), "googleapi: Error 400: Bad Request") {
+				t.Errorf("Finish() got: %v, want: %v", err, nil)
+				return
+			}
```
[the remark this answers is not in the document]
mcastorina:
```
The original was testing that `Init()` errors weren't surfaced in
`Finish()`, but the `SourceManager` changed that behavior.
```

(also included in the commit message for these changes)

pkg/sources/source_manager.go
ahrav: question: What happens if common.CancellableWrite encounters an error?

Looks like the error is reported by calling report.ReportError(Fatal{err}), and then the goroutine is immediately exited with return.

A potential problem here is that runWithoutUnits is not directly aware of this change. The error is reported via report.ReportError, but unless ReportError has side effects that the main function checks (which i hope isn't the case), I think this fxn will continue running.

In which case the source.Chunks function might continue to send chunks to the channel, but there is no longer a consumer for the channel since the goroutine has exited. I think?

ahrav: Maybe use an error channel to communicate errors from the goroutine back to the main function?

PR #1594, opened 2023-08-01
Common chunk reader

ahrav:
### Description:
Create a common chunk reader that can be used when reading large amounts of data. Modify the existing chunker to use the same logic as the new chunker.

### Checklist:

Modified chunker code is ever so slightly more performant.
Screenshot 2023-08-02 at 6 39 43 AM

### DIFF
--- a/pkg/sources/chunker.go
+++ b/pkg/sources/chunker.go
@@ -81,6 +81,9 @@ func WithPeekSize(size int) ConfigOption {
 }
 
 // ChunkReader reads chunks from a reader and returns a channel of chunks and a channel of errors.
+// The channel of chunks is closed when the reader is closed.
+// This should be used whenever a large amount of data is read from a reader.
+// Ex: reading attachments, archives, etc.
 type ChunkReader func(ctx context.Context, reader io.Reader) (<-chan []byte, <-chan error)
 
 // NewChunkReader returns a ChunkReader with the given options.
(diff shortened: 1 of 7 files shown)


pkg/sources/s3/s3.go
```diff
@@ -336,15 +335,27 @@ func (s *Source) pageChunker(ctx context.Context, client *s3.S3, chunksChan chan
...
+				select {
+				case <-ctx.Done(): // priority to context cancellation
+					return ctx.Err()
+				default:
+					select {
+					case <-ctx.Done():
+						return ctx.Err()
+					case chunksChan <- &chunk:
+					}
+				}
```
mcastorina, line 351: Cool -- would you also mind adding this to [common.CancellableWrite](https://github.com/trufflesecurity/trufflehog/blob/v3.46.3/pkg/common/context.go#L14-L24)?

pkg/sources/chunker.go
```diff
@@ -25,28 +27,113 @@ func Chunker(originalChunk *Chunk) chan *Chunk {
...
+	config := applyOptions(opts)
+	return createReaderFn(config)
+}
+
+func applyOptions(opts []ConfigOption) *chunkReaderConfig {
+	// Set defaults.
+	config := &chunkReaderConfig{
+		chunkSize: ChunkSize,      // default
+		totalSize: TotalChunkSize, // default
+	}
```
mcastorina, line 92: `peekSize` should probably be initialized here instead of `totalSize`.

mcastorina (approved):
lgtm!

PR #1596, opened 2023-08-02
Use common chunk reader

ahrav:
### Description:
Use the [common chunk reader](https://github.com/trufflesecurity/trufflehog/pull/1594) in all the sources that read data from files.

### Checklist:

- changing the return type from multiple channels to a single channel with a unified struct that contains the []byte and error. Benchmarks show an almost negligible slow down. The simplicity here outweighs the perf differences imo.
Screenshot 2023-08-03 at 3 49 47 PM


pkg/sources/chunker_test.go
```diff
@@ -84,3 +66,116 @@ func BenchmarkChunker(b *testing.B) {
...
+				if tt.wantErr {
+					assert.Error(t, err, "Expected an error")
+				} else {
+					assert.NoError(t, err, "Unexpected error")
+				}
+			default:
+				if tt.wantErr {
+					assert.Fail(t, "Expected error but got none")
+				}
+			}
```
mcastorina, line 178: Would this code be shaped differently to test the error case?

pkg/sources/circleci/circleci.go
```diff
@@ -229,34 +228,46 @@ func (s *Source) chunkAction(_ context.Context, proj project, bld build, act act
...
+				Data: &source_metadatapb.MetaData_Circleci{
+					Circleci: &source_metadatapb.CircleCI{
+						VcsType:     proj.VCS,
+						Username:    proj.Username,
+						Repository:  proj.RepoName,
+						BuildNumber: int64(bld.BuildNum),
+						BuildStep:   stepName,
+						Link:        linkURL,
+					},
 				},
```
mcastorina, line 252: I have a theory that this code will block if we encounter an error.

mcastorina (changes requested):
Requesting changes because I think this code could block. I will test locally to confirm.

pkg/sources/circleci/circleci.go
mcastorina: Update: I was wrong! I missed that the error channel is buffered

pkg/sources/chunker_test.go
```diff
@@ -84,3 +66,116 @@ func BenchmarkChunker(b *testing.B) {
...
+			wantErr:    false,
+		},
+		{
+			name:       "Equal to totalSize",
+			input:      strings.Repeat("a", 2048),
+			chunkSize:  1024,
+			peekSize:   1024,
+			wantChunks: []string{strings.Repeat("a", 1024), strings.Repeat("a", 1024)},
+			wantErr:    false,
+		},
```
mcastorina, line 142: Shouldn't the first chunk have 2048 bytes (chunk size + peek size)?

pkg/sources/chunker_test.go
```diff
@@ -84,3 +66,116 @@ func BenchmarkChunker(b *testing.B) {
...
+		{
+			name:       "Larger than totalSize",
+			input:      strings.Repeat("a", 4096),
+			chunkSize:  1024,
+			peekSize:   1024,
+			wantChunks: []string{strings.Repeat("a", 1024), strings.Repeat("a", 1024), strings.Repeat("a", 1024), strings.Repeat("a", 1024)},
+			wantErr:    false,
+		},
```
mcastorina, line 150: Same here, wouldn't the chunks have both the chunk data and the peek data?

mcastorina (commented):
After looking at the tests more closely I have a different question.

Also, these tests might be a good candidate for some of the functions in [iotest](https://pkg.go.dev/testing/iotest).

pkg/sources/chunker_test.go
ahrav: fixed

pkg/sources/chunker_test.go
ahrav: fixed

pkg/sources/chunker_test.go
```diff
@@ -137,22 +145,24 @@ func TestNewChunkedReader(t *testing.T) {
...
+			totalSize:  1024 + 1024,
+			wantChunks: []string{strings.Repeat("a", 2048), strings.Repeat("a", 2048), strings.Repeat("a", 2048), strings.Repeat("a", 1024)},
 			wantErr:    false,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			readerFunc := NewChunkReader(WithChunkSize(tt.chunkSize), WithPeekSize(tt.peekSize))
+			readerFunc := NewChunkReader(WithChunkSize(tt.chunkSize), WithPeekSize(tt.peekSize), WithTotalChunkSize(tt.totalSize))
```
mcastorina: It seems kind of odd to me having `WithTotalChunkSize` option. Isn't the total chunk size a function of the chunk size and peek size (namely `ChunkSize + PeekSize`)?

pkg/sources/circleci/circleci.go
```diff
@@ -229,34 +228,46 @@ func (s *Source) chunkAction(_ context.Context, proj project, bld build, act act
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case chunksChan <- chunk:
+			}
+		}
+	}
+	if err := <-errCh; err != nil {
+		ctx.Logger().Error(err, "Error reading chunk.")
+		return nil
 	}
```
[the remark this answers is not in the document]
ahrav: If there were no errors, we would read from a closed channel, which shouldn't block and would return the zero value. Therefore, this should be fine?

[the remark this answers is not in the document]
ahrav: I am considering whether it would be more manageable to return a unified type that includes both the []byte data and the error. For instance:
``` 
type ChunkResult struct {
  data []byte
  err error
}
```
The `ChunkResult` approach makes the code less complex and more intuitive to handle imo. By encapsulating both the []byte data and the error in a single object, it provides a clear and straightforward way to check for errors and access the data without having to manage multiple return values. This can lead to cleaner and easier to maintain code.

Thoughts?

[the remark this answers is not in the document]
mcastorina: Yeah, I like the union approach. I can't imagine a large perf impact, though tests are better than imagination :)

My only recommendation is to use struct objects instead of pointers to avoid the indirection.

pkg/sources/chunker.go
```diff
@@ -101,39 +118,40 @@ func applyOptions(opts []ConfigOption) *chunkReaderConfig {
...
 
 			if err != nil {
 				if !errors.Is(err, io.EOF) {
 					ctx.Logger().Error(err, "error reading chunk")
-					errChan <- err
+					chunkRes.err = err
+					chunkResultChan <- chunkRes
 				}
 				return
 			}
```
mcastorina, line 153: If both `n > 0` and `err != nil` are true, we would send two `ChunkResult` objects:
```go
ChunkResult{data}
ChunkRelust{data, err}
```

Could that cause weird behavior in the caller?

mcastorina (commented):
I like the update to `ChunkResult`, makes it a lot easier to follow. I just had one question about an edge-case, but otherwise lgtm.

pkg/sources/chunker.go
ahrav: hmmm, it could. I'll adjust to only send one object.

mcastorina (approved):
nice!

PR #1597, opened 2023-08-02
synchronize print operations to stdout

ahrav:
### Description:
Ensure print operations to stdout are serialized for all `Printer` implementations. We want to avoid:
Screenshot 2023-08-02 at 8 11 26 AM

### Checklist:

### DIFF
--- a/pkg/output/github_actions.go
+++ b/pkg/output/github_actions.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"sync"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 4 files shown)


mcastorina (approved):
After looking at this example, I wonder if we could make `Lock()` / `Unlock()` methods of the `Printer` interface and embed a `sync.Mutex` in the implementations.

Then wherever we use the `Printer` interface we can use `Lock()` and `Unlock()`.

Just brainstorming ideas - I think what you have here is good.

ahrav:
> After looking at this example, I wonder if we could make `Lock()` / `Unlock()` methods of the `Printer` interface and embed a `sync.Mutex` in the implementations.
> 
> Then wherever we use the `Printer` interface we can use `Lock()` and `Unlock()`.
> 
> Just brainstorming ideas - I think what you have here is good.

Yea, I can see that being useful, but I think it forces the implementation on how to Print.

If we follow the principle of least privilege : Interfaces are typically designed to reflect a contract of what an object can do, not how it should do it. Adding synchronization primitives might violate this principle. 

Then again we are currently using the interface in an environment that requires thread safety....

PR #1602, opened 2023-08-03
Adjust regex and add tests

bill-rich:


pkg/detectors/paypaloauth/paypaloauth_test.go
```diff
@@ -19,14 +19,17 @@ import (
...
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
 	if err != nil {
 		t.Fatalf("could not get test secrets from GCP: %s", err)
 	}
 	secret := testSecrets.MustGetField("PAYPALOAUTH_SECRET")
 	inactiveSecret := testSecrets.MustGetField("PAYPALOAUTH_SECRET_INACTIVE")
 	id := testSecrets.MustGetField("PAYPALOAUTH_CLIENTID")
 
+	newId := testSecrets.MustGetField("PAYPALOAUTH_NEW_INACTIVE_CLIENTID")
+	newSecret := testSecrets.MustGetField("PAYPALOAUTH_NEW_INACTIVE_SECRET")
```
mcastorina, line 31: Are these "new" values a different format to what paypal oauth used to use?

bill-rich: Yes. They're a different length and have a the clientID has a different prefix

mcastorina: In that case, does it make sense to make a v2 detector? Or is the idea to broaden the existing one to support both?

ahrav: question: could this put us in a position where we can no longer reverify/remediate old findings? Do we already have results with the current regex?

bill-rich: The idea is to broaden this one. The old regex was expanded to include more, but the old ones would still match. Not sure if that would qualify for v2.

ahrav: I don't think the new regex captures results from the old regex. For example [new regex](https://regex101.com/r/vYVjmP/1) vs [old regex](https://regex101.com/r/kJ5NPe/1). We want to make sure the new regex also works with old values.

I'd recommend tweaking the regex to include a range:
`\b([A-Za-z0-9_\.]{7}-[A-Za-z0-9_\.]{72}|[A-Za-z0-9_\.]{5}-[A-Za-z0-9_\.]{34,38})\b`

[updated regex](https://regex101.com/r/lB0T1w/1)

ahrav (approved):

PR #1604, opened 2023-08-03
implement tri-state verification in FTP detector

rosecodym:
This PR implements tri-state verification in the FTP detector. The verification timeout was made injectable to support a new test case.

The dogfood/Github action detection of "new" secrets is just picking up new test cases that use the existing test FTP secrets, which are already cleartext elsewhere in the codebase. (They're publicly available.)

### DIFF
--- a/pkg/detectors/ftp/ftp.go
+++ b/pkg/detectors/ftp/ftp.go
@@ -65,7 +65,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			s.Verified = verifyFTP(ctx, parsedURL)
+			verificationErr := verifyFTP(ctx, parsedURL)
+			s.Verified = verificationErr == nil
+			if isErrDeterminate(verificationErr) {
+				s.VerificationError = verificationErr
+			}
 		}
 
 		if !s.Verified {
@@ -85,7 +89,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
-func verifyFTP(ctx context.Context, u *url.URL) bool {
+func isErrDeterminate(e error) bool {
+	return true
+}
+
+func verifyFTP(ctx context.Context, u *url.URL) error {
 	host := u.Host
 	if !strings.Contains(host, ":") {
 		host = host + ":21"
@@ -93,13 +101,11 @@ func verifyFTP(ctx context.Context, u *url.URL) bool {
 
 	c, err := ftp.Dial(host, ftp.DialWithTimeout(5*time.Second))
 	if err != nil {
-		return false
+		return err
 	}
 
 	password, _ := u.User.Password()
-	err = c.Login(u.User.Username(), password)
-
-	return err == nil
+	return c.Login(u.User.Username(), password)
 }
 
 func (s Scanner) Type() detectorspb.DetectorType {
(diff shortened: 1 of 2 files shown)


pkg/detectors/ftp/ftp.go
```diff
@@ -58,48 +70,59 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		results = append(results, r)
 	}
 
 	return results, nil
 }
 
-func verifyFTP(ctx context.Context, u *url.URL) bool {
+func isErrDeterminate(e error) bool {
+	ftpErr := &textproto.Error{}
+	return errors.As(e, &ftpErr) && ftpErr.Code == ftpNotLoggedIn
```
mcastorina, line 110: This is cool

pkg/detectors/ftp/ftp_test.go
```diff
@@ -48,7 +51,7 @@ func TestFTP_FromChunk(t *testing.T) {
 				{
 					DetectorType: detectorspb.DetectorType_FTP,
 					Verified:     true,
-					Redacted:     "ftp://dlpuser:*************************@ftp.dlptest.com",
+					Redacted:     "ftp://dlpuser:********@ftp.dlptest.com",
```
mcastorina, line 54: Curious why some of the tests were adjusted? Did our redaction code update?

pkg/detectors/ftp/ftp_test.go
```diff
@@ -95,9 +135,14 @@ func TestFTP_FromChunk(t *testing.T) {
...
+				if (got[i].VerificationError != nil) != tt.wantVerificationErr {
+					t.Fatalf("wantVerificationError = %v, verification error = %v", tt.wantVerificationErr, got[i].VerificationError)
+				}
 			}
-			if diff := pretty.Compare(got, tt.want); diff != "" {
-				t.Errorf("URI.FromData() %s diff: (-got +want)\n%s", tt.name, diff)
+			opts := cmpopts.IgnoreFields(detectors.Result{}, "VerificationError")
+			if diff := cmp.Diff(got, tt.want, opts); diff != "" {
+				t.Errorf("FTP.FromData() %s diff: (-got +want)\n%s", tt.name, diff)
 			}
```
mcastorina, line 146: **Nit / question:** Could we add `"Raw"` to the `cmpopts.IgnoreFields` and remove setting it to `nil` on line 137? If so, we could convert that loop into iterating over values also.

mcastorina (approved):
Very nice seeing the GH action proving the test is correct :rofl:

pkg/detectors/ftp/ftp_test.go
rosecodym: i guess so? they were failing on `main` and i didn't think that figuring out why was worth it

pkg/detectors/ftp/ftp_test.go
rosecodym: oh yeah probably

rosecodym: good thought moving `Raw` into the ignorelist, but i don't follow the second part of your suggestion

mcastorina: I think the only reason we were iterating over the indices was to clear out the `Raw` value in the object, so now that `Raw` is being ignored we don't need to do that and could update the loop to be:

```go
for _, gotResult := range got {
    if (gotResult.Verification != nil) != tt.wantVerificationErr {
        // ...
    }
}
```

rosecodym: oh, no, it's dumber than that. the root problem is that we know when we want to see a verification error, but we don't always know the exact shape of the error, so there's a top-level "is there an error" flag. but verification errors are per [i]result[/i], not per overall chunk flow, so we loop through and check the expected flag against each individual result object.

there's probably a better way to do it, though - i haven't thought about it too hard.

PR #1605, opened 2023-08-03
Detect API keys without app keys

bill-rich:
DataDog API keys can be used for writing without the app key being present (for writing). Add checking for API key on its own.

### DIFF
--- a/pkg/detectors/datadogtoken/datadogtoken.go
+++ b/pkg/detectors/datadogtoken/datadogtoken.go
@@ -20,7 +20,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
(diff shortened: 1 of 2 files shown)


pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -34,24 +34,28 @@ func (s Scanner) Keywords() []string {
...
 				continue
 			}
-			resApiMatch := strings.TrimSpace(apimatch[1])
+			resAppMatch := strings.TrimSpace(appMatch[1])
 
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_DatadogToken,
-				Raw:          []byte(resMatch),
-				RawV2:        []byte(resMatch + resApiMatch),
+				Raw:          []byte(resAppMatch),
```
ahrav, line 54: This concerns me a little bit. Since we are changing the Raw field we will have un-updatable secrets in the DB. Looking at our results for Datadog we see the following:
Screenshot 2023-08-03 at 2 47 46 PM

This means we have no way to ever update these secrets.

bill-rich: resAppMatch is just resMatch renamed to make the distinction more clear between the two.

ahrav (approved):

PR #1608, opened 2023-08-07
add salesforce detector

zubairk14:
### Description:
Adding salesforce auth token detector

### Checklist:


pkg/detectors/salesforce/salesforce_test.go
```diff
@@ -0,0 +1,167 @@
...
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Salesforce.FromData() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			for i := range got {
+				if len(got[i].Raw) == 0 {
+					t.Fatalf("no raw secret present: \n %+v", got[i])
+				}
+				got[i].Raw = nil
+				if (got[i].VerificationError != nil) != tt.wantVerificationErr {
+					fmt.Printf("where the fuck is the error")
```
rosecodym, line 142: a bold move in a public-facing repository

rosecodym (commented):
what do you think of the new indeterminacy stuff?

also, the detector test template code this pulls from has a couple of small tweaks that should be made. i didn't call them out here because it's just copied code but you might be interested in cleaning them up anyway. i'll open a pr to fix the template so this doesn't keep happening. the two things are:
* one of the format strings is weird (wrong order, one part missing a `=`)
* it should use `cmp.Diff` instead of `pretty.Diff` so that the `Raw` field doesn't need to be deleted before comparison

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,115 @@
...
+				Raw:          []byte(tokenMatch),
+			}
+
+			if verify {
+				client := s.client
+				if client == nil {
+					client = defaultClient
+				}
+
+				req, err := http.NewRequestWithContext(ctx, "GET", instanceMatch+"/services/data/v"+fmt.Sprint(currentVersion)+"/query?q=SELECT+name+from+Account", nil)
```
mcastorina, line 71: **Nit:** `fmt.Sprint` isn't doing anything here and could be removed.

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,115 @@
...
+				req.Header.Set("Authorization", "Bearer "+tokenMatch)
+				if err != nil {
+					continue
+				}
```
mcastorina, line 75: It's rare that creating a request would error out, but maybe we should check the `err` before calling `req.Header.Set`.

rosecodym (changes requested):

rosecodym:
#1609 has the template test code tweaks

pkg/detectors/salesforce/salesforce_test.go
zubairk14: OHHHH whoops my bad

zubairk14: i thought i removed all the debug prints... evidently i missed the most important one to remove 😬

pkg/detectors/salesforce/salesforce.go
zubairk14: good point - i was letting copilot do its thing lol

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,118 @@
...
+
+			if verify {
+				client := s.client
+				if client == nil {
+					client = defaultClient
+				}
+
+				req, err := http.NewRequestWithContext(ctx, "GET", instanceMatch+"/services/data/v"+currentVersion+"/query?q=SELECT+name+from+Account", nil)
+				if err != nil {
+					continue
```
rosecodym, line 73: out of curiosity, do you know how this can fail? (i have no idea myself.) i ask because i want to ask whether continuing is the correct thing to do.

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,118 @@
...
+				}
+				req.Header.Set("Authorization", "Bearer "+tokenMatch)
+				if err != nil {
+					continue
+				}
```
rosecodym, line 78: this looks like duplicate logic

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,118 @@
...
+
+// FromData will find and optionally verify Salesforce secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	instanceMatches := instancePat.FindAllStringSubmatch(dataStr, -1)
+	tokenMatches := accessTokenPat.FindAllStringSubmatch(dataStr, -1)
+
+	for _, instance := range instanceMatches {
+		if len(instance) != 1 {
```
rosecodym, line 47: what is this check for? (i have the same question about `token` below.)

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,118 @@
...
+				if err != nil {
+					// End execution, append Detector Result if request fails to prevent panic on response body checks
+					s1.VerificationError = err
+					results = append(results, s1)
+					continue
+				}
+
+				verifiedBodyResponse, err := common.ResponseContainsSubstring(res.Body, "records")
+				if err != nil {
+					return nil, err
```
rosecodym, line 91: is this written like this because a failure to read the body indicates an underlying problem with the detector, rather than a problem verifying this particular candidate secret?

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,118 @@
...
+				}
+
+				if err == nil {
```
rosecodym, line 94: isn't this always `true` at this point? (if so, that `else` block below should be changed somehow.)

pkg/detectors/salesforce/salesforce.go
```diff
@@ -0,0 +1,118 @@
...
+					defer res.Body.Close()
+					if res.StatusCode >= 200 && res.StatusCode < 300 && verifiedBodyResponse {
+						s1.Verified = true
+					} else if res.StatusCode == 401 {
+						s1.Verified = false
+					} else {
+						s1.VerificationError = fmt.Errorf("request to %v returned unexpected status %d", res.Request.URL, res.StatusCode)
```
rosecodym, line 101: an unexpected status isn't the only way you can get here; a 2xx unaccompanied by "records" in the response body will also do it. in that case, i would find the error message confusing. you might consider breaking out those two cases into different blocks.

pkg/detectors/salesforce/salesforce.go
zubairk14: this is to check that we are picking up regex matches correctly.

if we have capturing groups, then instead of 

`instanceMatches: [[https://trufflesecurity2-dev-ed.develop.my.salesforce.com]]`

it'll return 

`idMatches:  [[id e2e897c7-0887-4523-835d-e75eeedcdc52 e2e897c7-0887-4523-835d-e75eeedcdc52]]`

like we do on the docusign detector - so we expect two results per match 
`id e2e897c7-0887-4523-835d-e75eeedcdc52`
and 
`e2e897c7-0887-4523-835d-e75eeedcdc52` - so we use the second one with index [1]
 https://github.com/trufflesecurity/trufflehog/blob/c9f58b37808647936b3fb5fa230a9547ba96f3e1/pkg/detectors/docusign/docusign.go#L47
 
 tldr: it mostly checks if it's empty or otherwise not what we were expecting the regex to pick up.

pkg/detectors/salesforce/salesforce.go
zubairk14: yes - we could have an error reading the body. 

```
func ResponseContainsSubstring(reader io.ReadCloser, target string) (bool, error) {
	scanner := bufio.NewScanner(reader)
	for scanner.Scan() {
		if strings.Contains(scanner.Text(), target) {
			return true, nil
		}
	}
	if err := scanner.Err(); err != nil {
		return false, err
	}
	return false, nil
}
```

pkg/detectors/salesforce/salesforce.go
zubairk14: oh good point

pkg/detectors/salesforce/salesforce.go
zubairk14: hmm not sure, ill noodle on it

pkg/detectors/salesforce/salesforce.go
rosecodym: but if we do, is bailing on the entire detection the correct action?

zubairk14: hm what's the alternative? 

there's no network calls etc so a retry doesn't make sense afaik... figure it's best to pass on the error and call it quits instead of guessing the secret candidate is not verified & moving on.

rosecodym: you could mark it as an indeterminate failure to verify

zubairk14: hm i thought indeterminate was only for failed network calls? 

if `common.ResponseContainsSubstring(res.Body, "records")` returns an error, it would be a Scanner.Err() that looks like it's basically a parsing error

```
// Err returns the first non-EOF error that was encountered by the Scanner.
func (s *Scanner) Err() error {
	if s.err == io.EOF {
		return nil
	}
	return s.err
}
```

rosecodym: it's just for any failure that isn't known to be caused by the candidate secret being invalid. i don't know what could cause `Scanner.Err()`, but i'm guessing it's not invalid credentials. if you record it as an indeterminate verification, we'll at least have a record of it; but in doing so we might accumulate records that we don't want for some reason. up to you!

rosecodym (approved):
i suppose i will learn another day why we check the size of the regex match sets

PR #1610, opened 2023-08-07
Move commits_scanned to ScanRepo

bill-rich:
Move `commits_scanned` from `Chunks` to `ScanRepo` to show up when the Git source is utilized from another source, such as GitHub.


trufflesteeeve (approved):
So we'll get a "commits_scanned" per repo now, yeah? I think that makes sense.

PR #1612, opened 2023-08-09
Refactor FragmentLineOffset to match multiline secrets

mcastorina:
### Description:
Refactors `FragmentLineOffset` to search for the needle *before* splitting by newlines. This allows matching `Raw` multi-line secrets.

* All tests still pass 
* Added multiline tests with ignore
* Added benchmarks for finding the needle at the start / middle / end of the haystack
  * Before / after runs shown in PR comment

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -599,10 +599,13 @@ func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64, 
 	if !found {
 		return 0, false
 	}
-	lineNumber := int64(len(bytes.Split(before, []byte("\n")))) - 1
-	// if the line contains the ignore tag, we should ignore the result
+	lineNumber := int64(bytes.Count(before, []byte("\n")))
+	// If the line contains the ignore tag, we should ignore the result.
 	endLine := bytes.Index(after, []byte("\n"))
-	if endLine != -1 && bytes.Contains(after[:endLine], []byte(ignoreTag)) {
+	if endLine == -1 {
+		endLine = len(after)
+	}
+	if bytes.Contains(after[:endLine], []byte(ignoreTag)) {
 		return lineNumber, true
 	}
 	return lineNumber, false
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -73,6 +73,17 @@ func TestFragmentLineOffset(t *testing.T) {
 			expectedLine: 3,
 			ignore:       true,
 		},
+		{
+			name: "ignore on last line",
+			chunk: &sources.Chunk{
+				Data: []byte("line1\nline2\nline3\nsecret here // trufflehog:ignore"),
+			},
+			result: &detectors.Result{
+				Raw: []byte("secret here"),
+			},
+			expectedLine: 3,
+			ignore:       true,
+		},
 	}
 
 	for _, tt := range tests {


mcastorina:
Benchmarks before:
```
BenchmarkFragmentLineOffsetStart-8        131709              8180 ns/op           13568 B/op          1 allocs/op
BenchmarkFragmentLineOffsetMiddle-8       127000              9730 ns/op           13568 B/op          1 allocs/op
BenchmarkFragmentLineOffsetEnd-8          107628             10841 ns/op           13568 B/op          1 allocs/op
```
Benchmarks after:
```
BenchmarkFragmentLineOffsetStart-8      12259804                91.74 ns/op           80 B/op          1 allocs/op
BenchmarkFragmentLineOffsetMiddle-8       239596              4868 ns/op            6528 B/op          1 allocs/op
BenchmarkFragmentLineOffsetEnd-8          123247              9600 ns/op           12288 B/op          1 allocs/op
```

pkg/engine/engine.go
```diff
@@ -595,17 +595,17 @@ func SupportsLineNumbers(sourceType sourcespb.SourceType) bool {
...
-			if bytes.Contains(line, []byte(ignoreTag)) {
-				return int64(i), true
-			}
-			return int64(i), false
-		}
+	before, after, found := bytes.Cut(chunk.Data, result.Raw)
+	if !found {
+		return 0, false
+	}
+	lineNumber := int64(len(bytes.Split(before, []byte("\n")))) - 1
```
rosecodym, line 602: i'm not sure how big chunks can get, but if they can get big we could save a bunch of memory by simply counting `\n` characters (rather than splitting them), right?

pkg/engine/engine.go
```diff
@@ -595,17 +595,17 @@ func SupportsLineNumbers(sourceType sourcespb.SourceType) bool {
...
+	// if the line contains the ignore tag, we should ignore the result
+	endLine := bytes.Index(after, []byte("\n"))
+	if endLine != -1 && bytes.Contains(after[:endLine], []byte(ignoreTag)) {
```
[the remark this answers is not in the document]
mcastorina: good catch, I think the ignore tag would be missed if a chunk didn't end in a newline.

pkg/engine/engine.go
mcastorina: soo.. I had that exact same thought and tested it out. for some reason `Split` is way faster than counting:

```
BenchmarkSplitIt-8         14970             77778 ns/op          106525 B/op          1 allocs/op
BenchmarkIterateIt-8        9021            130181 ns/op              46 B/op          0 allocs/op
```

mcastorina: I think in this case it makes sense to prioritize speed over memory consumption, though I could be persuaded otherwise.

It's not *exactly* in the hot path but close to it.

mcastorina: oh, I'm dumb. there's a [bytes.Count](https://pkg.go.dev/bytes#Count) function that `Split` uses internally.

mcastorina:
New benchmarks after switching to `bytes.Count`:

```
BenchmarkFragmentLineOffsetStart-8      25723839                44.66 ns/op            0 B/op          0 allocs/op
BenchmarkFragmentLineOffsetMiddle-8      1544580               774.8 ns/op             0 B/op          0 allocs/op
BenchmarkFragmentLineOffsetEnd-8          796764              1487 ns/op               0 B/op          0 allocs/op
```

rosecodym (approved):
this isn't new so it's not blocking here, but i thought of something else: what if the secret is found twice in the chunk, and only the first one has the ignore tag?

mcastorina:
> this isn't new so it's not blocking here, but i thought of something else: what if the secret is found twice in the chunk, and only the first one has the ignore tag?

If the exact same secret is found twice in the chunk, it would always take the first one and ignore it. I'm not sure it's possible to prevent that the way we're calculating the line number retroactively.

PR #1613, opened 2023-08-09
Support indeterminate verification in Gitlab detector

rosecodym:
This PR adds support for tri-state verification to the Gitlab detector.

### DIFF
--- a/pkg/detectors/gitlab/gitlab.go
+++ b/pkg/detectors/gitlab/gitlab.go
@@ -12,7 +12,11 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{ detectors.EndpointSetter }
+type Scanner struct {
+	client *http.Client
+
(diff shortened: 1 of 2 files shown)


pkg/detectors/gitlab/gitlab.go
```diff
@@ -76,7 +84,13 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 					// 401 is bad key
 					if res.StatusCode == http.StatusOK || res.StatusCode == http.StatusForbidden {
```
ahrav, line 85: Could we switch-ify this?

```
switch res.StatusCode {
case http.StatusOK, http.StatusForbidden:
	secret.Verified = true
case http.StatusUnauthorized:
	// Nothing to do; zero values of Result are the ones we want
default:
	secret.VerificationError = fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
}
```

ahrav (approved):

PR #1617, opened 2023-08-09
fix error message in alchemy test

rosecodym:
The message has an extra comma. This test file isn't important on its own but it is the template for all the new ones so it's good to get it right.


zricethezav (approved):
LGTM

PR #1620, opened 2023-08-12
Fix RubyGems Detector

react-native-mapbox-gl:
### Description:
The verified code for the rubygems was not working correctly. For a valid key and no scope to list gems, the response will be 403 forbidden. For a invalid key, the response will be 401. The scope of rubygems could be limited to not list gems.

### Checklist:


react-native-mapbox-gl:
Fixed myfreshworks detector since it was generating a lot of false positives. Now its more streamlined and give more accurate results.

PR #1624, opened 2023-08-14
Add tri-state verification to sqlserver detector

rosecodym:
This is a different detector than the general JDBC detector.


zubairk14 (approved):
lgtm

PR #1625, opened 2023-08-15
 Integration of SpecializedHandler for Enhanced Archive Processing

ahrav:
---

This description provides an overview of the changes, the rationale behind them, the impact, and mentions the need for comprehensive testing. It's written in a way that should be accessible to other developers reviewing the PR, giving them the necessary context to understand the changes made.

### Description:
This PR introduces the new `SpecializedHandler` interface to our existing archive handling system, enabling the application to recognize and process specialized archives like Debian (.deb) and RPM (.rpm) packages. 

1. **Introduction of SpecializedHandler**: 
   A new interface `SpecializedHandler` has been added, which encapsulates the logic to handle specialized archives, including examining files and determining their specialized nature.

2. **Modification to Existing Handler**:
   The `HandleFile` function now utilizes the `SpecializedHandler`, allowing specialized handling before falling back to regular file type handling. It efficiently leverages both standard and specialized handlers within the existing processing flow.

3. **Support for Debian and RPM Files**:
   Implemented methods `extractDebContent` and `extractRpmContent` to extract the contents of Debian and RPM files, respectively. These functions utilize command-line tools like 'ar', 'rpm2cpio', and 'cpio'.

4. **Tool Verification and MIME Type Handling**:
   A map `mimeTools` has been defined, mapping MIME types to the corresponding tools needed for handling. Functions `isToolInstalled` and `ensureToolsForMimeType` ensure that the required tools are installed. 

5. **Optimization for Nested Files**:
   Specialized handling for nested files has also been implemented, managing temporary files and directories and recursive handling of nested archives.

### Checklist:


pkg/handlers/handlers.go
```diff
@@ -56,3 +63,151 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
...
+
+// HandleSpecializedArchives takes a file path and an io.ReadCloser representing the input file,
+// and processes it based on its extension, such as handling Debian (.deb) and RPM (.rpm) packages.
+// It returns an io.ReadCloser that can be used to read the processed content of the file,
+// and an error if any issues occurred during processing.
+// The caller is responsible for closing the returned reader.
+func HandleSpecializedArchives(ctx logContext.Context, path string, inputFile io.ReadCloser) (io.ReadCloser, error) {
+	var reader io.ReadCloser
+	var err error
+	ext := filepath.Ext(path)
```
bill-rich, line 80: This would work generally, but detecting the mime type would be more reliable.

pkg/handlers/handlers.go
```diff
@@ -56,3 +63,151 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
...
+	if err != nil {
+		return tempEnv{}, fmt.Errorf("unable to create temporary file: %w", err)
+	}
+
+	extractPath, err := os.MkdirTemp("", "tmp_archive")
+	if err != nil {
+		return tempEnv{}, fmt.Errorf("unable to create temporary directory: %w", err)
+	}
+
+	if _, err = io.Copy(tempFile, file); err != nil {
```
bill-rich, line 190: I don't think the file size has been checked, so this could result in copying a very large file.

pkg/sources/filesystem/filesystem.go
```diff
@@ -151,7 +151,13 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 	defer inputFile.Close()
 	logger.V(3).Info("scanning file")
 
-	reReader, err := diskbufferreader.New(inputFile)
```
bill-rich, line 154: If the file has a .deb or .rpm extension, but is not actually a file of that type, we'll end up running through the reader, but not be able to reset it.

pkg/sources/filesystem/filesystem.go
```diff
@@ -151,7 +151,13 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
...
+	reader, err := handlers.HandleSpecializedArchives(ctx, path, inputFile)
```
bill-rich, line 154: Special archive handling in standard archive handling would allow us to get this functionality into all sources that currently support archive handling.

pkg/handlers/handlers.go
```diff
@@ -56,3 +63,151 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
...
+	// List the content of the extraction directory.
+	extractedFiles, err := os.ReadDir(tempEnv.extractPath)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read extracted directory: %w", err)
+	}
+
+	// Determine the correct data archive name. (e.g., data.tar.gz, data.tar.xz)
+	var dataArchiveName string
+	for _, file := range extractedFiles {
+		if strings.HasPrefix(file.Name(), "data.tar.") {
```
bill-rich, line 123: Can this vary?

bill-rich (changes requested):

pkg/sources/filesystem/filesystem.go
ahrav: I think I found a pretty decent way to make this work. Let me know what you think.

pkg/handlers/handlers.go
ahrav: Is the idea here we just read to up `maxSize` or do we skip the file outright?

bill-rich: I think reading up to `maxSize` is the way to go.

bill-rich (approved):
Looks good. Cool use of `io.MultiReader` to deal with the filetyping using the `Reader`.

pkg/handlers/handlers.go
ahrav: So from my reading this is what i've gleaned.

A Debian package file contains two tarballs and a special file, all concatenated together. Here's the structure:
debian-binary: This file contains the version number of the deb package format.
control.tar.gz: This compressed tarball contains metadata about the package, such as dependencies, scripts to run during installation or removal, etc.
data.tar.gz: This compressed tarball contains the actual files of the package. It may be compressed using different methods, leading to extensions like data.tar.xz or data.tar.bz2.

I think it's probably safe to go with this approach. It's the .rpm formats i'm a little concerned about.

PR #1626, opened 2023-08-15
add thog CLI support for GitHub config validate

zubairk14:
### Description:

### Checklist:


pkg/sources/sources.go
```diff
@@ -240,7 +240,7 @@ type Progress struct {
 // Validator is an interface for validating a source. Sources can optionally implement this interface to validate
 // their configuration.
 type Validator interface {
-	Validate() []error
+	Validate(context.Context) []error
```
rosecodym, line 243: Why did you add a context here if your new implementation doesn't use it? (And why doesn't this affect other implementations?)

zubairk14: good catch and apologies - the commit that included the use of ctx didn't get pushed up - but the Github source does need the context to attempt a call on the ghClient to validate that the auth details from the config file worked.

as far as i can tell, syslog is the only other source to use this Validator interface so just updated that.

zubairk14: btw here's the associated change on thog https://github.com/trufflesecurity/thog/pull/1385/files

zubairk14: trufflesecurity/thog#1626

pkg/sources/github/github.go
```diff
@@ -273,6 +273,58 @@ func (s *Source) Init(aCtx context.Context, name string, jobID, sourceID int64,
...
+		ghClient, err = createGitHubClient(s.httpClient, apiEndpoint)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	default:
+		errs = append(errs, errors.Errorf("Invalid configuration given for source. Name: %s, Type: %s", s.name, s.Type()))
+	}
+
+	// Run a simple query to check if the client is valid
+	_, _, err = ghClient.Users.Get(ctx, "")
```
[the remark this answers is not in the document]
rosecodym: ah, ok. do you want to try using the client if there were errors creating it? looking at the implementation of `createGitHubClient` it looks like the returned client will be `nil` in that case, which will cause a panic here (right?)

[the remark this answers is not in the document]
zubairk14: hm yeah i am curious what actually causes nil, err b/c even with the erroneous endpoint and fake token it does generate a 'valid' client - but yes, will gate it to prevent panic!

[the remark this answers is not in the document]
rosecodym: looks like you get `nil` [if there was a failed attempt to create an enterprise client](https://github.com/trufflesecurity/trufflehog/blob/62d359eba4c41e76d43cdca832c275795e8c3477/pkg/sources/github/github.go#L647C1-L647C1)

rosecodym (approved):

pkg/sources/github/github.go
```diff
@@ -273,6 +273,65 @@ func (s *Source) Init(aCtx context.Context, name string, jobID, sourceID int64,
...
+
+	switch cred := s.conn.GetCredential().(type) {
+	case *sourcespb.GitHub_BasicAuth:
+		s.httpClient.Transport = &github.BasicAuthTransport{
+			Username: cred.BasicAuth.Username,
+			Password: cred.BasicAuth.Password,
+		}
+		ghClient, err = createGitHubClient(s.httpClient, apiEndpoint)
+		if err != nil {
+			errs = append(errs, err)
```
bill-rich, line 293: The errors returned when creating the client /might/ not make it clear exactly where the failure is. For validate, it would be better to wrap with a bit of extra context. eg https://github.com/trufflesecurity/trufflehog/blob/948a94792b5a165fc4cb0a22412c96fc3513431e/pkg/sources/syslog/syslog.go#L80

bill-rich (changes requested):
Just the request to wrap errors with extra context and I think its ready to roll.

pkg/sources/github/github.go
zubairk14: ah gotcha

bill-rich (approved):

PR #1633, opened 2023-08-16
[bug] - copy chunk before sending on chunksChan

ahrav:
### Description:
Memory Sharing Issues: I was sharing the chunk variable across iterations. When I update chunk.Data and send the pointer &chunk to the channel, all the chunks I was sending would point to the same memory. The change in one will affect all of them. If another goroutine is reading from the channel and modifying the chunk in parallel, we can run into concurrent read-write conflicts.

Create a new copy before passing it into the chunks chan.
### Checklist:


mcastorina (approved):

dustin-decker (approved):

ahrav:
Added an integration test to keep me honest!

PR #1635, opened 2023-08-17
Docker tar identification from filesystem

joeleonjr:
### Description:

### Checklist:

### DIFF
--- a/pkg/handlers/handlers.go
+++ b/pkg/handlers/handlers.go
@@ -1,10 +1,20 @@
 package handlers
 
 import (
+	"archive/tar"
 	"context"
+	"fmt"
 	"io"
+	"math/rand"
+	"os"
+	"runtime"
 
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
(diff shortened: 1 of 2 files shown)


pkg/handlers/handlers.go
```diff
@@ -56,3 +94,38 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
 		}
 	}
 }
+
+func IsReaderTarball(ctx context2.Context, reader io.Reader) bool {
+	// Check if file is a tarball, if so, check if it is a Docker image.
+	_, err := tar.NewReader(reader).Next()
```
dustin-decker, line 100: I'm not certain, but this might consume some of the reader so that the following usage of the reader is missing some of the beginning data

joeleonjr: I think that's right. Going to close this draft PR and re-work it based on the latest changes to specialized handler. Look for another PR soon.

PR #1642, opened 2023-08-18
Use placeholder as default if field left empty and is required

hxnyk:
### Description:
If nothing's been entered for a required field, populate the value with the placeholder value & bug fixes @mcastorina 

### Checklist:

### DIFF
--- a/pkg/tui/pages/source_configure/trufflehog_configure.go
+++ b/pkg/tui/pages/source_configure/trufflehog_configure.go
@@ -1,6 +1,7 @@
 package source_configure
 
 import (
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -49,7 +50,7 @@ func GetTrufflehogConfiguration() truffleCmdModel {
 		Key:         "concurrency",
 		Required:    false,
 		Help:        "Number of concurrent workers.",
-		Placeholder: "1",
+		Placeholder: strconv.Itoa(runtime.NumCPU()),
 	}
 
 	return truffleCmdModel{textinputs.New([]textinputs.InputConfig{jsonOutput, verification, verifiedResults, excludeDetectors, concurrency}).SetSkip(true)}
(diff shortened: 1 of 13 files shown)


pkg/tui/pages/source_configure/trufflehog_configure.go
```diff
@@ -49,7 +50,7 @@ func GetTrufflehogConfiguration() truffleCmdModel {
 		Key:         "concurrency",
 		Required:    false,
 		Help:        "Number of concurrent workers.",
-		Placeholder: "1",
+		Placeholder: strconv.Itoa(runtime.NumCPU()),
```
rosecodym, line 53: which was the default concurrency before we added the tui stuff?

pkg/tui/pages/source_select/source_select.go
```diff
@@ -73,6 +73,7 @@ func New(c common.Common) *SourceSelect {
 		OssItem("GCS (Google Cloud Storage)", "Scan a Google Cloud Storage instance."),
 		// Enterprise sources.
 		EnterpriseItem("Artifactory", "Scan JFrog Artifactory packages."),
+		// EnterpriseItem("Azure", "Scan Microsoft Azure resources."),
```
rosecodym, line 76: is this still supposed to be commented out?

pkg/tui/sources/github/github.go
```diff
@@ -19,41 +19,69 @@ func GetFields() githubCmdModel {
...
-		command = append(command, "--repo="+inputs["repo"])
+	if inputs["repo"].Value != "" {
+		command = append(command, "--repo="+inputs["repo"].Value)
 	}
 
 	return strings.Join(command, " ")
 }
 
-func (m githubCmdModel) Summary() string {
+func (m githubCmdModel) Test() map[string]textinputs.Input {
```
rosecodym, line 69: is this function still supposed to be here?

pkg/tui/sources/syslog/syslog.go
```diff
@@ -64,10 +64,8 @@ func (m syslogCmdModel) Cmd() string {
 	syslogKeys := [5]string{"address", "protocol", "cert", "key", "format"}
 
 	for _, key := range syslogKeys {
-		if inputs[key] != "" {
```
rosecodym, line 67: what makes this guard no longer necessary?

pkg/tui/pages/source_configure/trufflehog_configure.go
mcastorina: The updated version. We copied it from here: https://github.com/trufflesecurity/trufflehog/blob/ed062178622c74167551eb8f2322c56a78d53a1d/main.go#L47

pkg/tui/sources/syslog/syslog.go
hxnyk: Moved the check to the `GetInputs` func in `textinputs.go`

rosecodym (approved):
sorry, didn't notice that you were waiting on me

PR #1644, opened 2023-08-18
Add flag to get information if trufflehog being ran from TUI

hxnyk:
### Description:
Add flag to get information if trufflehog being ran from TUI @mcastorina 

### Checklist:

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -16,18 +16,19 @@ import (
 	"github.com/jpillora/overseer/fetcher"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/version"
 )
 
-func Fetcher(version string) fetcher.Interface {
+func Fetcher(version string, tui bool) fetcher.Interface {
 	return &OSS{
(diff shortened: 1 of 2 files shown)


pkg/updater/updater.go
```diff
@@ -58,7 +60,8 @@ func (g *OSS) Fetch() (io.Reader, error) {
 	data := &FormData{
 		OS:             runtime.GOOS,
 		Arch:           runtime.GOARCH,
-		CurrentVersion: version.BuildVersion,
+		CurrentVersion: g.CurrentVersion,
```
zubairk14, line 63: curious: not familiar with this part of the code base, why do we need to change `currentVersion` ?

mcastorina: It wasn't being used anywhere, so we figured it was probably meant to be used here. `g.CurrentVersion` is always set to `version.BuildVersion` anyway so it should behave the same.

The alternative is to delete `g.CurrentVersion` and just always use `version.BuildVersion`, which might be better.

rosecodym: if we're not using it anywhere else and `version.BuildVersion` carries the same data i vote delete

mcastorina (approved):

PR #1645, opened 2023-08-18
Update launchdarkly regex, support sdk keys, add tri-state verification

trufflesteeeve:
### Description:
This updates the launch darkly regex to acknowledge that they are using UUIDv4s with a prefix as tokens. It also supports `sdk-` prefixed keys, in addition to the previous `api-` prefixed keys.

It also includes updates to support tri-state verification.

### Checklist:


pkg/detectors/launchdarkly/launchdarkly.go
```diff
@@ -5,28 +5,39 @@ import (
...
-	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"launchdarkly", "launch_darkly"}) + `\b([a-z0-9-]{40})\b`)
+	// Launchdarkly keys are UUIDv4s with either api- or sdk- prefixes.
+	// mob- keys are possible, but are not sensitive credentials.
+	keyPat = regexp.MustCompile(`\b((?:api|sdk)-[a-z0-9]{8}-[a-z0-9]{4}-4[a-z0-9]{3}-[a-z0-9]{4}-[a-z0-9]{12})\b`)
 )
 
-// Keywords are used for efficiently pre-filtering chunks.
-// Use identifiers in the secret preferably, or the provider name.
+// We are not including "mob-" because client keys are not sensitive.
```
zubairk14, line 37: good point - mobile apps and even js apps usually use pkce instead of actual secrets b/c they cannot securely store secrets. 

> Well, traditionally, OAuth 2.0 apps are issued a client_id and client_secret when the developer registers an app. This works fine when the app is running on a web server as the users of the app never have access to the source code and therefore don’t have the opportunity to see the secret. However, this obviously causes problems when we try to use OAuth 2.0 from a JavaScript or native app, since as we’ve seen, they don’t have the ability to keep secrets.

In OAuth 1, a secret was required in order to make every API request, which is one of its major shortcomings, and largely why it was replaced with OAuth 2.0. OAuth 1 was developed before mobile apps were as prevalent as they are today, so it wasn’t developed with these constraints in mind.

That changed with OAuth 2.0, especially once the PKCE (proof key code exchange) extension was introduced. I like to think of PKCE as an “on the fly” client secret. Instead of shipping a client_secret in a mobile app, the way PKCE works is the app creates a new random secret every time it starts the OAuth flow. This way there are no secrets shipped ahead of time, and there is nothing useful for an attacker to steal. You can read more details about how PKCE works in our blog post “[OAuth 2.0 for Native and Mobile Apps](https://developer.okta.com/blog/2018/12/13/oauth-2-for-native-and-mobile-apps)” or on [oauth.com](https://www.oauth.com/oauth2-servers/pkce/).

OAuth still sends access tokens over the wire where they may be visible to you if you’re using something like mitmproxy, but the difference is that this token is issued on the fly and is specific to the user who is using it! This way, there are no secrets shipped in the source code, and if someone wants to intercept the traffic from their own device, all they will see is an access token that was issued just to them! They won’t get access to anything their app can’t already access.

https://developer.okta.com/blog/2019/01/22/oauth-api-keys-arent-safe-in-mobile-apps

pkg/detectors/launchdarkly/launchdarkly.go
```diff
@@ -47,25 +59,48 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+			} else {
+				// This is a server SDK key. Try to initialize using the SDK.
+				_, err := ldclient.MakeCustomClient(resMatch, defaultSDKConfig, defaultSDKTimeout)
+				if err == nil {
+					s1.Verified = true
+				} else if err == ldclient.ErrInitializationFailed {
+					// If initialization fails, the key is not valid, so do nothing.
+				} else if err == ldclient.ErrInitializationTimeout {
+					// If initialization times out, we don't know if the key is valid.
+					s1.VerificationError = err
```
rosecodym, line 94: How do you want to handle errors that are neither of these named cases?

trufflesteeeve: Ah, the documentation claims that those are the only two errors it can return. But reading it immediately contradicts that. So I think all non-ErrInitializationFailed are likely in the "don't know" case. Though it looks like there's a "SDK key contains invalid characters" that definitely means it is not verified.

pkg/detectors/launchdarkly/launchdarkly_test.go
```diff
@@ -81,6 +82,42 @@ func TestLaunchDarkly_FromChunk(t *testing.T) {
...
+				},
+				want: []detectors.Result{
+					{
+						DetectorType: detectorspb.DetectorType_LaunchDarkly,
+						Verified:     true,
+					},
+				},
+				wantErr: false,
+			},
+		*/
```
rosecodym, line 120: idk if this is intended to remain here

rosecodym (approved):
indeterminacy looks good to me! having an indeterminate catch-all is the way i've been doing things too.

dustin-decker (approved):

PR #1646, opened 2023-08-20
[bug] - Only scan gist comments or repo comments

ahrav:
### Description:
We should only scan gist comments or repo comments.
```
{"level":"info-2","ts":"2023-08-18T07:09:55Z","logger":"thog/scanner","msg":"Errors encountered while scanning","pid":"CPTDi","version":"v1.70.18","source_type":"SOURCE_TYPE_GITHUB","source_name":"GitHub_Scheduled_Scan_GTL","job_id":47112,"source_id":212,"error-count":147,"errors":"[error scanning comments in repo https://gist.github.com/19bbb62cea62a831e31b11ca8b6f5d33.git: url missing owner and/or repo: 'https://gist.github.com/19bbb62cea62a831e31b11ca8b6f5d33.git' ... # Repeated for hundreds of gists
```

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -988,44 +988,49 @@ func (s *Source) scanComments(ctx context.Context, repoPath string, chunksChan c
 
 	trimmedURL := removeURLAndSplit(repoURL.String())
 	if repoURL.Host == "gist.github.com" && s.includeGistComments {
-		s.log.Info("scanning github gist comments", "repository", repoPath)
-		// GitHub Gist URL.
-		var gistId string
-		if len(trimmedURL) == 2 {
-			// https://gist.github.com/<id>
-			gistId = trimmedURL[1]
-		} else if len(trimmedURL) == 3 {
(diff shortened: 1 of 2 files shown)


0x1 (approved):
to clarify, the bug was caused by the case where the gist url didn't have the ID or had additional paths that were unhandled?

ahrav:
> to clarify, the bug was caused by the case where the gist url didn't have the ID or had additional paths that were unhandled?

So this was happening b/c we try to scan both gist comments and repo comments. We first check to see if it's a gist comment which has the structure: `https://gist.github.com/19bbb62cea62a831e31b11ca8b6f5d33.git` we successfully scan it. We should then return instead of trying to also scan the repo comments. This is because if we try to scan the repo comment for this same url it wont exist. We currently have a gist url, therefore we just return within the `if repo.URL.Host == "gist.github.com" && s.includeGistComments`. Otherwise we try to process the repo comment. We can't do both b/c the url only belongs to one of the other.

PR #1648, opened 2023-08-21
Fix reversed ordering of arguments

mcastorina:
The source manager initialization function was defined as `sourceID` followed by `jobID`, while the source initialization function is the reverse. This is confusing and easy to mix up since the parameters are the same type.

This commit adds a test to make sure the source manager initializes in the correct order, but it doesn't prevent the library user to make the same mistake. We may want to consider using different types.


rosecodym (approved):
strong :+1: for separate types

PR #1652, opened 2023-08-22
Include the job ID in a chunk

ahrav:
### Description:
Includes the job ID in the chunk struct. This can then be used in the pipeline to lookup jobs as well as potentially terminate jobs that are no longer running.

### Checklist:

### DIFF
--- a/pkg/sources/source_manager.go
+++ b/pkg/sources/source_manager.go
@@ -6,10 +6,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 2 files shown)


mcastorina (commented):
lgtm; we should also add it to the `runWithUnits` method

ahrav:
> lgtm; we should also add it to the `runWithUnits` method

done, not sure if I did it correctly.. but done.

mcastorina (approved):
very nice! I'm glad we can just add it here and not in every place we create a chunk.

PR #1653, opened 2023-08-22
add snowflake detector

### LINKED ISSUE #1607: Request for Snowflake detector

## Description

After reviewing the [list of supported detectors / credential types](https://github.com/trufflesecurity/trufflehog/tree/main/pkg/detectors), there does not appear to be support for Snowflake credentials. Specifically, identifying if a given set of credentials are from a managed user in Snowflake. 

### Preferred Solution

Similar to other detectors, identifying credentials (keypair or uname/pw) and detecting whether they're still active against a benign endpoint would be preferred. 

### Additional Context

While snowflake supports SSO authN/Z, there are still scenarios where programmatic access via service accounts is desirable. These SAs will have their own credentials associated with them, and service owners will be required to manage these credentials in their own environments. Authenticating with credentials can either by via [basic auth](https://docs.snowflake.com/en/user-guide/api-authentication#overview), or via [keypairs](https://docs.snowflake.com/en/user-guide/key-pair-auth).

#### References

https://www.snowflake.com/blog/snowflake-service-account-securitypart-1/
https://www.snowflake.com/blog/snowflake-service-account-security-part-2/

zubairk14:
Resolves #1607 

### Description:
* Detect Snowflake secrets (compound URI of account, username, password)
* Enrich Secret Result with account and databases that the secret has access to.

### Checklist:


pkg/detectors/snowflake/snowflake.go
```diff
@@ -0,0 +1,196 @@
...
+
+				if verify {
+					config := &gosnowflake.Config{
+						Account:  resAccountMatch,
+						User:     resUsernameMatch,
+						Password: resPasswordMatch,
+						Database: database,
+					}
+
+					fmt.Println("config: ", config)
```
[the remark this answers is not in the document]
zubairk14: resolved

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/felixge/fgprof v0.9.3
 	github.com/getsentry/sentry-go v0.22.0
 	github.com/go-errors/errors v1.4.2
-	github.com/go-git/go-git/v5 v5.8.1
+	github.com/go-git/go-git/v5 v5.8.0
```
dustin-decker, line 34: looks like you downgraded go-git, can you revert that?

pkg/detectors/snowflake/snowflake.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+	"net/http"
+	"regexp"
+	"strings"
+	"unicode"
+)
+
+type Scanner struct {
+	client *http.Client
```
dustin-decker, line 18: Lint issue here (unused)

pkg/detectors/snowflake/snowflake.go
```diff
@@ -0,0 +1,161 @@
...
+				if verify {
+					_ = &gosnowflake.Config{
+						Account:  resAccountMatch,
+						User:     resUsernameMatch,
+						Password: resPasswordMatch,
+						Database: database,
+					}
+
+					// Open a connection to Snowflake
+					db, err := sql.Open("snowflake", uri) // Needs the snowflake driver from gosnowflake
```
dustin-decker, line 114: Is there a way to provide a TCP timeout through the driver config or the connection string?

pkg/detectors/snowflake/snowflake.go
```diff
@@ -0,0 +1,161 @@
...
+						var databases []string
+						for rows.Next() {
+							var name, createdOn, isDefault, isCurrent, origin, owner, comment, option, retentionTime, kind string
+							err := rows.Scan(&createdOn, &name, &isDefault, &isCurrent, &origin, &owner, &comment, &option, &retentionTime, &kind)
+							if err != nil {
+								s1.VerificationError = fmt.Errorf("unable to finish querying Snowflake to enrich secret ExtraData %+v", err)
+							}
+							databases = append(databases, name)
+						}
+						s1.ExtraData["databases"] = strings.Join(databases, ", ")
```
dustin-decker, line 142: Neat!

pkg/detectors/snowflake/snowflake.go
```diff
@@ -0,0 +1,161 @@
...
+						Database: database,
+					}
+
+					// Open a connection to Snowflake
+					db, err := sql.Open("snowflake", uri) // Needs the snowflake driver from gosnowflake
+
+					if err != nil {
+						s1.VerificationError = fmt.Errorf("unable to open a connection to Snowflake %+v", err)
+					}
+					defer db.Close()
```
dustin-decker, line 119: Question: Is it okay to Close this if there was an error? With net/http requests the request would be nil and have a nil pointer exception

go.mod
zubairk14: oh idk how that happened

pkg/detectors/snowflake/snowflake.go
zubairk14: when i tried testing it with bad credentials, it errors without any NPE/panic

pkg/detectors/snowflake/snowflake.go
zubairk14: The sql.Open function from the database/sql package doesn't provide a direct option to set a timeout for the connection.

Is there a way to disrupt the connection externally in the test?

dustin-decker: Look at the connection string parameters: https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Connection_Parameters

Also, I don't think you need to define the config struct, from the looks of it you can just do an unnamed import of the driver and then construct the URI like you're already doing, based on the example at the top of https://pkg.go.dev/github.com/snowflakedb/gosnowflake#section-documentation

pkg/detectors/snowflake/snowflake.go
```diff
@@ -0,0 +1,158 @@
...
+						for rows.Next() {
+							var name, createdOn, isDefault, isCurrent, origin, owner, comment, option, retentionTime, kind string
+							err := rows.Scan(&createdOn, &name, &isDefault, &isCurrent, &origin, &owner, &comment, &option, &retentionTime, &kind)
+							if err != nil {
+								s1.VerificationError = fmt.Errorf("unable to finish querying Snowflake to enrich secret ExtraData %+v", err)
+							}
+							databases = append(databases, name)
+						}
+						s1.ExtraData["databases"] = strings.Join(databases, ", ")
+						s1.Verified = true
```
rosecodym, line 140: The presence of a verification error (which is possible above) signals that the detector wasn't able to determine whether the candidate credentials are valid. That's not really consistent with setting `Verified` to true (which means that we know that the credentials are valid) - the framework assumes that you won't combine those two things, so you shouldn't. I'm not sure what the best solution here is - maybe putting those error messages somewhere else in the `Result` payload?

(I know that making this fourth "verification state" mathematically possible but forbidding its use is kind of weird, but it's the price we pay for being able to migrate detectors gradually.)

rosecodym (dismissed):
in addition to my in-line comment, an indeterminate failure test would be nice if it's feasible

pkg/detectors/snowflake/snowflake.go
zubairk14: ah just noticed the loginTimeout param - thanks for the callout.

> loginTimeout: Specifies the timeout, in seconds, for login. The default is 60 seconds. The login request gives up after the timeout length if the HTTP response is success.

pkg/detectors/snowflake/snowflake.go
zubairk14: ahh - good point

pkg/detectors/snowflake/snowflake.go
zubairk14: if i malform the connection string like 
> uri := fmt.Sprintf("%s://^#%s@%s/%s", resUsernameMatch, resPasswordMatch, resAccountMatch, database)

it will still generate non-nil `db`

example:

```
db:  &{{{} {} 0} {zubairkhan14://^#fakepassword1!@tuacoip-zt74995/SNOWFLAKE 0x104642948} {{} {} 0} {0 0} [] map[] 0 0 0x140001f07e0 false map[] map[] 0 0 0 0 <nil> 0 0 0 0 0x10278a0f0}
```

zubairk14: the only condition in which sql.Open returns an empty response with error is if the driver does not exist - which we are hard setting to snowflake

```
func Open(driverName, dataSourceName string) (*DB, error) {
	driversMu.RLock()
	driveri, ok := drivers[driverName]
	driversMu.RUnlock()
	if !ok {
		return nil, fmt.Errorf("sql: unknown driver %q (forgotten import?)", driverName)
	}
```

dustin-decker (approved):

PR #1655, opened 2023-08-23
Test S3 role assumption

rosecodym:
This PR adds a test of the S3 role assumption functionality. It currently only tests role assumption within a single account.

### DIFF
--- a/pkg/sources/s3/s3_test.go
+++ b/pkg/sources/s3/s3_test.go
@@ -10,13 +10,12 @@ import (
 
 	"github.com/kylelemons/godebug/pretty"
 	"github.com/stretchr/testify/assert"
-	"google.golang.org/protobuf/types/known/anypb"
-
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/credentialspb"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/sourcespb"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
 func TestSource_Chunks(t *testing.T) {
@@ -35,6 +34,7 @@ func TestSource_Chunks(t *testing.T) {
 		name       string
 		verify     bool
 		connection *sourcespb.S3
+		setEnv     map[string]string
 	}
 	tests := []struct {
 		name          string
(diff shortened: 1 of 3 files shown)


pkg/sources/s3/s3_test.go
```diff
@@ -58,13 +60,34 @@ func TestSource_Chunks(t *testing.T) {
...
+				connection: &sourcespb.S3{
+					Roles: []string{"arn:aws:iam::619888638459:role/s3-test-assume-role"},
+				},
+				setEnv: map[string]string{
+					"AWS_ACCESS_KEY_ID":     s3key,
+					"AWS_SECRET_ACCESS_KEY": s3secret,
+				},
+			},
+			wantErr:       false,
+			wantChunkData: `W2RlZmF1bHRdCmF3c19zZWNyZXRfYWNjZXNzX2tleSA9IFF5OTlTK1pCL0NXbEZOdHhQQWlkN2dHenZzclhoQkI3dXJBQ0FMcFoKYXdzX2FjY2Vzc19rZXlfaWQgPSBBS0lBMzVPSFgyRFNPWkc2NDdOSApvdXRwdXQgPSBqc29uCnJlZ2lvbiA9IHVzLWVhc3QtMg==`,
```
rosecodym, line 78: This is the same secret content as in the other test, just with the order of the access key ID and secret key reversed.

pkg/sources/s3/s3_test.go
```diff
@@ -58,13 +60,34 @@ func TestSource_Chunks(t *testing.T) {
...
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
 			var cancelOnce sync.Once
 			defer cancelOnce.Do(cancel)
 
+			for k, v := range tt.init.setEnv {
+				env.Patch(t, k, v)
```
mcastorina, line 88: [env.Patch](https://pkg.go.dev/gotest.tools/env#Patch) returns a function that resets the environment variable back to its original value. Should we use that here in a `defer` or something?

mcastorina (approved):
lgtm, thanks for adding the test! just had one question

pkg/sources/s3/s3_test.go
rosecodym: The documentation for v3 says this:
> When used with Go 1.14+ the unpatch function will be called automatically when the test ends, unless the TEST_NOCLEANUP env var is set to true.
Buuuut I'm also seeing now that it also says that `env.Patch` is deprecated in favor of `t.Setenv` so I'll switch it over to use that.

PR #1662, opened 2023-08-24
Support azure git links

ahrav:
### Description:
Extend the GenerateLinks function to support link generation for Azure repositories.
Refactored the function to use consts instead of magic strings.
Moved the function to the giturl package.

### Checklist:

### DIFF
--- a/pkg/giturl/giturl.go
+++ b/pkg/giturl/giturl.go
@@ -2,6 +2,7 @@ package giturl
 
 import (
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


pkg/giturl/giturl.go
```diff
@@ -88,3 +110,35 @@ func NormalizeOrgRepoURL(provider provider, repoURL string) (string, error) {
...
 	return parsed.String(), nil
 }
+
+// GenerateLink crafts a link to the specific file from a commit.
+// Supports GitHub, GitLab, Bitbucket, and Azure Repos.
+// If the provider supports hyperlinks to specific lines, the line number will be included.
+func GenerateLink(repo, commit, file string, line int64) string {
+	switch determineProvider(repo) {
+	case providerBitbucket:
+		return repo[:len(repo)-4] + "/commits/" + commit
```
mcastorina, line 120: Is the `len(repo)-4` to strip the `.com/` part out of the repo? Does that mean `determineProvider` should use `strings.HasSuffix` instead of `strings.Contains`?

mcastorina (approved):
just had one non-blocking question

pkg/giturl/giturl.go
ahrav: I believe so. Could you elaborate a bit, i'm not sure I quite follow. `determineProvider` is called prior to stripping the TLD, so I think it's operating on the entire url. I think I might be missing something 🤣

mcastorina: No it's more likely I'm missing something. I'm confused what the `len(repo)-4` is doing. Are we expecting `repo` to end in a TLD? If so, why isn't that reflected in `determineProvider`?

ahrav: ahh i get you now. I am not sure, I believe repository is the link to the repo, or at least that's what it appears as. I probably overlooked that piece completely since I just moved the function over here and added the new case. That is a bit odd, but looks like both Github and Gitlab use it that way. Yea so looks like that is the case: https://github.com/trufflesecurity/trufflehog/blob/f2bfcc7ac6d8ca915c18f66b9ab1ccb0adaa9b14/pkg/sources/git/git.go#L504

ahrav: Might be something we want to address... but it will likely need a little more exploration just to make sure we don't break anything.

PR #1665, opened 2023-08-25
Add AvailableCapacity method to SourceManager

mcastorina:
### Description:
Add `AvailableCapacity` method to `SourceManager` to expose the approximate number of goroutines available to run jobs.

### Checklist:

### DIFF
--- a/pkg/sources/source_manager.go
+++ b/pkg/sources/source_manager.go
@@ -36,7 +36,10 @@ type SourceManager struct {
 	handles     map[handle]sourceInfo
 	handlesLock sync.Mutex
 	// Pool limiting the amount of concurrent sources running.
-	pool            errgroup.Group
+	pool                errgroup.Group
+	poolLimit           int
+	currentRunningCount int32
+	// Max number of units to scan concurrently per source.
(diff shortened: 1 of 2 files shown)


pkg/sources/source_manager.go
```diff
@@ -207,6 +215,16 @@ func (s *SourceManager) ScanChunk(chunk *Chunk) {
 	s.outputChunks <- chunk
 }
 
+// AvailableCapacity returns the number of concurrent jobs the manager can
+// accommodate at this time. If there is no limit, -1 is returned.
+func (s *SourceManager) AvailableCapacity() int {
```
ahrav, line 220: suggestion: Could we avoid in-band errors here and return multiple values?
maybe:
```
func (s *SourceManager) AvailableCapacity() (int, bool) {
	if s.poolLimit == 0 {
		return 0, false
	}
	runCount := atomic.LoadInt32(&s.currentRunningCount)
	return s.poolLimit - int(runCount), true
}
```
ref: https://google.github.io/styleguide/go/decisions#in-band-errors

ahrav (approved):

PR #1708, opened 2023-08-26
[chore] - update Docker source

ahrav:
### Description:
- Add common chunker to handle tar files.
- Add ability to scan images concurrently.

### Checklist:


pkg/sources/docker/docker.go
```diff
@@ -71,164 +72,202 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
+					return err
+				}
+				dockerLayersScanned.WithLabelValues(s.name).Inc()
+			}
 
-		layers, err := img.Layers()
-		if err != nil {
-			return err
-		}
+			dockerImagesScanned.WithLabelValues(s.name).Inc()
```
rosecodym, line 121: what's the additional increment for?

pkg/sources/docker/docker.go
```diff
@@ -71,164 +72,202 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
 
-		for _, layer := range layers {
+			return nil
+		})
+	}
+	_ = workers.Wait()
```
rosecodym, line 126: is it generally considered ok to discard this result? (without even, say, logging it?)

pkg/sources/docker/docker.go
```diff
@@ -71,164 +72,202 @@ func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, ver
...
 			}
-		}
 
-		logger = logger.WithValues("image", base, "tag", tag)
-		logger.V(2).Info("scanning image")
+			for _, layer := range layers {
+				if err := s.processLayer(ctx, layer, imgInfo, chunksChan); err != nil {
+					return err
+				}
+				dockerLayersScanned.WithLabelValues(s.name).Inc()
```
rosecodym, line 118: you also increment this inside `processLayer` - was that on purpose?

rosecodym (commented):
are there any tests you can add for the new tar functionality?

ahrav:
> are there any tests you can add for the new tar functionality?

new functionality? I don't believe there should be any new functionality in terms of the reading. The new functionality should only be the concurrency. 🤔

pkg/sources/docker/docker.go
ahrav: whoops. Removed the increment in processLayer.

rosecodym:
oh, am i misunderstanding
> Add common chunker to handle tar files.

?

pkg/sources/docker/docker.go
ahrav: Added some logic to collect errors from the goroutines.

pkg/sources/docker/docker.go
ahrav: I think this is the only increment for images. I can't find the other increment.

ahrav:
> oh, am i misunderstanding
> 
> > Add common chunker to handle tar files.
> 
> ?

Oh, that shouldn't be new functionality. I am just using the common chunker to handle the chunking instead of doing it manually. The existing test should cover this behavior.

pkg/sources/docker/docker.go
rosecodym: oh whoops i missed that images has its own metric

rosecodym (approved):

PR #1711, opened 2023-08-27
Filter unique detectors by keywords in chunk

ahrav:
### Description:
The main idea is to prevent iterating overing every detector for every chunk Even though we have the keywords in the chunk we iterate over all the detectors to check if they keyword is in the map we pre-populated. I believe this overhead can be eliminated by creating a mapping of keywords to detectors during engine initialization. Given this direct mapping when scanning each chunk we can isolate only the detectors with the given keywords.

- Since the keyword to detector, and detector type to detector info map are only written to during engine initialization. 
- Each detector worker will include a re-usable map for unique detectors. This reduces the pressure on the GC since the same few maps can be reused over the run.
- By changing the return type of FromChunk for the decoders to include the decoder type we can also remove the slightly costly interface type assertions along with the switch statement.

We get a small speedup on the tiny repo used in the benchmark, but the space savings is pretty substantial due to the new changes.

Screenshot 2023-08-27 at 7 57 20 PM

### Checklist:


pkg/engine/engine.go
```diff
@@ -433,64 +517,46 @@ type detectableChunk struct {
...
+				for k, detector := range uniqueDetectors {
+					decoded.Chunk.Verify = detector.shouldVerify
+					wgDetect.Add(1)
+					e.detectableChunksChan <- detectableChunk{
+						chunk:    *decoded.Chunk,
+						detector: detector,
+						decoder:  decoded.DecoderType,
+						wgDoneFn: wgDetect.Done,
 					}
+					delete(uniqueDetectors, k)
```
rosecodym, line 552: Oh neat, I wouldn't have guessed that removing a map entry during iteration is ok to do.

pkg/engine/engine.go
```diff
@@ -433,64 +517,46 @@ type detectableChunk struct {
 func (e *Engine) detectorWorker(ctx context.Context) {
 	var wgDetect sync.WaitGroup
 
+	// Reuse the same map to avoid allocations.
+	const avgDetectorsPerChunk = 2
+	uniqueDetectors := make(map[detectorspb.DetectorType]detectorInfo, avgDetectorsPerChunk)
```
rosecodym, line 522: To reprise our previous conversation, I still find this identifier confusing as a reader. What about `chunkSpecificDetectors` or `detectorsForThisChunk`?

pkg/engine/git_test.go
```diff
@@ -120,9 +121,10 @@ func BenchmarkGitEngine(b *testing.B) {
 	defer cancel()
 
 	e, err := Start(ctx,
-		WithConcurrency(1),
+		WithConcurrency(uint8(runtime.NumCPU())),
```
rosecodym, line 124: What effect did this have on the benchmarks?

pkg/engine/engine.go
```diff
@@ -305,46 +350,84 @@ func Start(ctx context.Context, options ...EngineOption) (*Engine, error) {
 		e.detectors[true] = DefaultDetectors()
 		e.detectors[false] = []detectors.Detector{}
 	}
+	ctx.Logger().V(4).Info("default engine options set")
+}
+
+// buildLookups prepares maps for fast detector lookups. Instead of scanning
+// through an array of detectors for every chunk, this lookup optimization
+// provides rapid access to relevant detectors using keywords.
+func (e *Engine) buildLookups(ctx context.Context) {
```
rosecodym, line 359: This function and its callees do a lot of manipulation of private `Engine` fields that are only used by this function and its callees. That suggests to me that it be helpful those parts of engine `Engine` into its own type, carving part of the big ball of mud out into a more well-manicured dirt patch: something that turns a range of detectors into an Aho-Corasick trie and the `keyword -> []detector` lookup used in `detectorWorker`. Is that something you considered?

pkg/engine/git_test.go
ahrav: I believe the only difference was that running w/ 1 was slower but the proportions were about the same. I was also trying to simulate normal use case where we use the `runtime.NumCPU` to determine concurrency unless overridden by the user.

pkg/engine/engine.go
ahrav: Took a stab at this. Let me know if this was what you had in mind. Moved the ahocorasick logic into a separate file. The main engine file was getting a bit unwieldy. Happy to move it back if we think this is more confusing.

pkg/engine/ahocorasickcore.go
```diff
@@ -0,0 +1,124 @@
...
+			key := ac.classifyDetector(d, verify)
+			keywords = ac.extractAndMapKeywords(d, key, keywords)
+		}
+	}
+
+	// Implementing a trie aids in substring searches among the keywords.
+	// This is crucial when you have large sets of strings to search through.
+	ac.buildTrie(keywords)
+	ctx.Logger().V(4).Info("ahoCorasickCore lookups built")
+}
```
rosecodym, line 74: this function is only called by one other function, and that function only does one thing (call this function). i would find it more readable to just inline it unless you had some other reason to break it out.

pkg/engine/ahocorasickcore.go
```diff
@@ -0,0 +1,124 @@
...
+// classifyDetector assigns a unique key for each detector. This key based on type and version,
+// ensures faster lookups and reduces redundancy in our main detector store.
+func (ac *ahoCorasickCore) classifyDetector(d detectors.Detector, shouldVerify bool) detectorKey {
+	detectorType := d.Type()
+	var version int
+	if v, ok := d.(detectors.Versioner); ok {
+		version = v.Version()
+	}
+	key := detectorKey{detectorType: detectorType, version: version}
+	ac.detectorTypeToDetectorInfo[key] = detectorInfo{Detector: d, shouldVerify: shouldVerify}
```
rosecodym, line 85: if you move this assignment to the caller you can both simplify this function (by removing `shouldVerify` and the receiver) and give it a more descriptive name like `createDetectorKey`

pkg/engine/ahocorasickcore.go
```diff
@@ -0,0 +1,124 @@
...
+		ac.keywordsToDetectors[kwLower] = append(ac.keywordsToDetectors[kwLower], key)
+	}
+	return keywords
+}
+
+// buildTrie uses the Ahocorasick algorithm to create a trie structure for efficient keyword matching.
+// This ensures that we can rapidly match against a vast set of keywords without individually comparing each one.
+func (ac *ahoCorasickCore) buildTrie(keywords []string) {
+	ac.prefilter = *ahocorasick.NewTrieBuilder().AddStrings(keywords).Build()
+}
```
rosecodym, line 104: this is another candidate for inlining - it's one line long and is only called in one place.

pkg/engine/ahocorasickcore.go
```diff
@@ -0,0 +1,124 @@
...
+
+// matchString performs a string match using the Aho-Corasick algorithm, returning an array of matches.
+// Designed for internal use within the ahoCorasickCore component.
+func (ac *ahoCorasickCore) matchString(input string) []*ahocorasick.Match {
+	return ac.prefilter.MatchString(strings.ToLower(input))
+}
```
rosecodym, line 110: i'll also suggest inlining this - the one thing it does (call `ToLower`) is also done by its only caller, so clearly the abstraction wasn't clear enough to help!

rosecodym (commented):
i like the `ahoCorasickCore` breakout, but its method configuration was hard for me to follow. i left some specific suggestions for things that would help me understand it, but they might not be exhaustive. let me know if there's some reason for the specific implementation that i missed though.

pkg/engine/ahocorasickcore.go
```diff
@@ -0,0 +1,124 @@
...
+
+// setup initializes the internal state of ahoCorasickCore to prepare it for keyword matching.
+// This involves pre-filtering setup and lookup optimization, critical for the engine's performance.
+func (ac *ahoCorasickCore) setup(ctx context.Context) {
+	ac.buildLookups(ctx)
+	ctx.Logger().V(4).Info("ahoCorasickCore setup complete")
+}
+
+// buildLookups prepares maps for fast detector lookups. Instead of scanning through an array of
+// detectors for every chunk, this lookup optimization  provides rapid access to relevant detectors using keywords.
```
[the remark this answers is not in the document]
ahrav: I have no idea how you caught that. Even after you mentioned it, it took me a minute to find. 😓

pkg/engine/ahocorasickcore.go
ahrav: The main reason I didn't inline this was because I didn't want the engine to have knowledge about the `prefilter` trie data structure. I'll fix the multiple `ToLower` so it's only called in here.

pkg/engine/ahocorasickcore.go
ahrav: ahhh i like that.

rosecodym (approved):
👍

pkg/engine/ahocorasickcore.go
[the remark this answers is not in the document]
rosecodym: that's wild, it's probably the most obvious thing in this pr to me. it jumped out _immediately_.

PR #1715, opened 2023-08-28
Validate s3 source

rosecodym:
### Description:
This PR adds S3 source validation. This is accomplished by factoring out common "bucket visiting" logic to be used by both scanning and validation.

### DIFF
--- a/pkg/sources/s3/s3.go
+++ b/pkg/sources/s3/s3.go
@@ -186,18 +186,24 @@ func (s *Source) getBucketsToScan(client *s3.S3) ([]string, error) {
 func (s *Source) scanBuckets(ctx context.Context, client *s3.S3, role string, bucketsToScan []string, chunksChan chan *sources.Chunk) error {
 	objectCount := uint64(0)
 
+	logger := s.log
+	if role != "" {
+		logger = logger.WithValues("roleArn", role)
+	}
+
 	for i, bucket := range bucketsToScan {
+		logger = logger.WithValues("bucket", bucket)
+
 		if common.IsDone(ctx) {
 			return nil
 		}
 
 		s.SetProgressComplete(i, len(bucketsToScan), fmt.Sprintf("Bucket: %s", bucket), "")
-		s.log.Info("Scanning bucket", "bucket", bucket)
+		logger.Info("Scanning bucket")
 
 		regionalClient, err := s.getRegionalClientForBucket(ctx, client, role, bucket)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/sources/s3/s3.go
```diff
@@ -163,34 +182,30 @@ func (s *Source) getBucketsToScan(client *s3.S3) ([]string, error) {
 	return bucketsToScan, nil
 }
 
-func (s *Source) scanBuckets(ctx context.Context, client *s3.S3, role string, bucketsToScan []string, chunksChan chan *sources.Chunk) error {
-	const defaultAWSRegion = "us-east-1"
+func (s *Source) scanBuckets(ctx context.Context, client *s3.S3, role string, bucketsToScan []string, chunksChan chan *sources.Chunk) {
```
rosecodym, line 185: This method signature used to return an `error` but there was no logic that ever actually returned one, so I removed it. (The lack of return isn't an oversight - the scanning code logs errors without bailing out.)

pkg/sources/s3/s3.go
```diff
@@ -396,6 +412,65 @@ func (s *Source) pageChunker(ctx context.Context, client *s3.S3, chunksChan chan
...
 
+func (s *Source) validateBucketAccess(ctx context.Context, client *s3.S3, roleArn string, buckets []string) []error {
+	shouldHaveAccessToAllBuckets := roleArn == ""
+	wasAbleToListAnyBucket := false
+	var errors []error
+
+	for _, bucket := range buckets {
+		if common.IsDone(ctx) {
+			return nil
+		}
```
mcastorina, line 423: **Suggestion:** Returning an error here makes more sense in the context of validation imo.

```go
return append(errors, ctx.Err())
```

pkg/sources/s3/s3.go
```diff
@@ -163,34 +182,30 @@ func (s *Source) getBucketsToScan(client *s3.S3) ([]string, error) {
...
+func (s *Source) scanBuckets(ctx context.Context, client *s3.S3, role string, bucketsToScan []string, chunksChan chan *sources.Chunk) {
 	objectCount := uint64(0)
 
+	logger := s.log
+	if role != "" {
+		logger = logger.WithValues("roleArn", role)
+	}
+
 	for i, bucket := range bucketsToScan {
+		logger = logger.WithValues("bucket", bucket)
```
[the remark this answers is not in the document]
rosecodym: oh interesting, i assumed that keys weren't repeatable

mcastorina (approved):

PR #1717, opened 2023-08-28
Use common chunker for archive handler

ahrav:
### Description:
Leverage the common chunker within the archive handler. This should also make sure the chunks include peek data.

### Checklist:


rosecodym (approved):
i don't really know how the archive stuff works so take this approval with a grain of salt

ahrav:
> i don't really know how the archive stuff works so take this approval with a grain of salt

I'll take that approval as an initial okay that the code looks not too bad. I'll wait for another review to make sure using the sliding window w/ overlap chunker in the archive handler looks okay.

Issue #1712, opened 2023-08-28
Add Tailscale Detector


marwan-at-work:
## Description
Hi there,
We at Tailscale would love to add a detector for secrets created by the Tailscale API or admin control as described here: https://tailscale.com/kb/1215/oauth-clients/

### Preferred Solution
Following the Trufflehog docs here https://github.com/trufflesecurity/trufflehog/blob/main/hack/docs/Adding_Detectors_external.md

PR #1719, opened 2023-08-29
Add Tailscale detector

marwan-at-work:
### Description:

This PR adds a new Tailscale detector for API and Oauth tokens

Fixes https://github.com/trufflesecurity/trufflehog/issues/1712

### Checklist:


zricethezav:
@marwan-at-work Thanks for submitting this PR! I love Tailscale :) 

One of the engineers will review this PR. In the meantime, would you mind resolving the merge conflicts?

pkg/detectors/tailscaleapi/tailscaleapi.go
```diff
@@ -0,0 +1,85 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`^tskey-(api|oauth)-[0-9A-Za-z_]+-[0-9A-Za-z_]+$`)
```
willnorris, line 24: we probably don't want to just limit to these prefixes.  There are quite a few: https://tailscale.com/kb/1277/key-prefixes/.  But we also may add more later, so maybe just a regex for the key type as well?

tylersmalley, line 24: The comment above makes me believe that they want to use word boundaries and not start/end anchors.

```suggestion
	keyPat = regexp.MustCompile(`\btskey-(api|oauth)-[0-9A-Za-z_]+-[0-9A-Za-z_]+\b`)
```

marwan-at-work: We ended up with a new endpoint that should verify everything?

marwan-at-work:
@zricethezav thanks for the kind words. Resolved the conflicts and the PR is ready for review :)

pkg/detectors/tailscale/tailscale.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// FromData will find and optionally verify Tailscaleapi secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+
+	for _, match := range matches {
+		if len(match) != 2 {
```
ahrav, line 41: suggestion: We con forego this len check because the regex directly targets and captures the desired key using boundary characters. Since the regex captures the entire key, the `len(match)` should only contain a single element in contrast to some of the other detectors which contain they key + context. For example [twitters regex](https://github.com/trufflesecurity/trufflehog/blob/8d66fde6de03816e6f2ecf09a3d0429ef60fe7c4/pkg/detectors/twitter/twitter.go#L24) uses the keyword [twitter](https://github.com/trufflesecurity/trufflehog/blob/8d66fde6de03816e6f2ecf09a3d0429ef60fe7c4/pkg/detectors/twitter/twitter.go#L30) which isn't part of the key. We also include the helper [detectors.prefixRegex](https://github.com/trufflesecurity/trufflehog/blob/8d66fde6de03816e6f2ecf09a3d0429ef60fe7c4/pkg/detectors/detectors.go#L121) which is why `match` in those detectors have multiple items.

If we use `resMatch := strings.TrimSpace(match[0])` we should be good.

ahrav (approved):
Thanks a lot @marwan-at-work, I know you spent a decent amount of time getting this ironed out and we really appreciate the contribution. Similar to Zach, I am also a huge Tailscale fan!

pkg/detectors/tailscale/tailscale.go
marwan-at-work: @ahrav thanks for the explanation, changed to `match[0]`

bugbaba:
Hi @marwan-at-work, 


Just wanted to double-check about the `/secret-scanning/verify` endpoint. I noticed that it isn't [documented](https://github.com/tailscale/tailscale/blob/main/api.md) and it appears to be a unique endpoint created specifically for trufflehog, which seems a bit unusual. 

It would be better to have a more general endpoint that could be utilized to verify the validity of the token.

marwan-at-work:
@bugbaba this is going to be a general endpoint for all secret scanning products that don't require a specific implementation besides validating a token (as opposed to GitHub [for example](https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partner-program#joining-the-secret-scanning-program-on-github), which requires specific public key verification, request/response body shapes, etc).

A more generic endpoint that I saw in other implementations is to hit a `/profiles` endpoint but being overly cautious from a security prospective, we'd like to not return any profile or user identifiers besides just a boolean about the requested token. Of course, if the token is valid, actors can hit other endpoints that retrieves more information but some implementation might permit us to revoke the token before that can happen and we currently don't have a /profile endpoint anyway without knowing more information like the tailnet that this token belongs to. 

As for documentation, this endpoint is end customers would care about, but it's something we'll discuss for sure. 

Thanks

PR #1720, opened 2023-08-29
update jira detector to match new variable tokens

codevbus:
### Description:
Updates the Jira detector to handle the newer token format

### Checklist:


ahrav:
I think we can leverage our [versioner interface](https://github.com/trufflesecurity/trufflehog/blob/e18cfd5e0af1469a9f05b8d5732bcc94c39da49c/pkg/detectors/detectors.go#L30). [here](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/github_old/github_old.go#L22) is an example w/ Github.

0x1 (approved):
lgtm

codevbus:
Local detector tests for both versions of jira are currently failing. I've updated `detectors5` to have both sets of test values.

PR #1730, opened 2023-08-30
Add Sourcegraph Token Detection

shivasurya:
### Description:

- Detect Sourcegraph API Token secret and verify using graphql endpoint

### Checklist:


ahrav (approved):
LGTM 👍 
Thanks a lot for the contribution @shivasurya , this looks great!

pkg/detectors/sourcegraph/sourcegraph.go
```diff
@@ -0,0 +1,94 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"sourcegraph", "sg_token"}) + `\b(sgp_[a-z0-9]{40})\b`)
```
ahrav, line 25: question: Do you know if tokens allow for characters from g-z? I've only ever seen sg tokens that are `[a-f0-9]` in which case we could tighten the regex a bit more. Otherwise, this works just as well.

zricethezav (approved):

shivasurya:
Thank you so much team 🥇

pkg/detectors/sourcegraph/sourcegraph.go
shivasurya: I'll verify that 👍🏻

rgmz, line 25: I don't think the PrefixRegex is necessary here. SourceGraph's token format in unique, so checking for `sourcegraph` or `sg_token` will miss valid results.

We should follow-up to this.

rgmz: From my testing it does not appear to use `g-z`.

PR #1731, opened 2023-08-30
add tri state verification to slack (not slack webhook)

zubairk14:
### Description:

### Checklist:

### DIFF
--- a/pkg/detectors/slack/slack.go
+++ b/pkg/detectors/slack/slack.go
@@ -13,13 +13,16 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
 
 // Check that the Slack scanner implements the SecretScanner interface at compile time.
 var _ detectors.Detector = Scanner{}
 
 var (
-	tokenPats = map[string]*regexp.Regexp{
+	defaultClient = common.SaneHttpClient()
+	tokenPats     = map[string]*regexp.Regexp{
 		"Slack Bot Token":               regexp.MustCompile(`xoxb\-[0-9]{10,13}\-[0-9]{10,13}[a-zA-Z0-9\-]*`),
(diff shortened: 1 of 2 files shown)


pkg/detectors/slack/slack.go
```diff
@@ -53,34 +56,54 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 					defer res.Body.Close()
 					var authResponse authRes
 					if err := json.NewDecoder(res.Body).Decode(&authResponse); err != nil {
 						continue
 					}
-					s.Verified = authResponse.Ok
+
+					if authResponse.Ok {
+						s1.Verified = true
+						// Slack API returns 200 even if the token is invalid. We need to check the error field.
```
rosecodym, line 87: do you think it's worth adding some code to check for non-2xx responses?

rosecodym (changes requested):
you're a machine! (only change request is to remove the commented stuff)

pkg/detectors/slack/slack.go
zubairk14: this API seems to return 200 always so imo our best bet is to depend on the else block for whenever the `authResponse` is not OK or the else to catch when the client response catches an error

```
 else {
						s1.VerificationError = fmt.Errorf("unexpected error auth response %+v", authResponse.Error)
					}
					```

Comment (rosecodym) [reply pkg/detectors/slack/slack.go:87]: What about checking for a response body that can't be successfully decoded as JSON?

File: pkg/detectors/slack/slack.go [hunk 2]
```diff
@@ -53,34 +56,54 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					} else if authResponse.Error == "invalid_auth" {
+						// The secret is determinately not verified (nothing to do)
+					} else {
+						s1.VerificationError = fmt.Errorf("unexpected error auth response %+v", authResponse.Error)
```
rosecodym, line 91: Can you add a test for this? One of the fake HTTP clients available allows you to inject a fake response body - lmk if you want me to dig up a specific example.

pkg/detectors/slack/slack.go
dustin-decker: Added

PR #1735, opened 2023-08-31
updating sendgrid detector to use tri-state verification

0x1:
### Description:
Updating the Sendgrid detector to use tri-state verification. 
Also added new tokens for the tests.

### Checklist:

### DIFF
--- a/pkg/detectors/sendgrid/sendgrid.go
+++ b/pkg/detectors/sendgrid/sendgrid.go
@@ -13,13 +13,15 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"sendgrid"}) + `(SG\.[\w\-_]{20,24}\.[\w\-_]{39,50})\b`)
 )
@@ -42,7 +44,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 		res := strings.TrimSpace(match[1])
 
-		s := detectors.Result{
+		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_SendGrid,
 			Raw:          []byte(res),
 		}
@@ -51,6 +53,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			// there are a few endpoints we can check, but templates seems the least sensitive.
 			// 403 will be issued if the scope is wrong but the key is correct
 			baseURL := "https://api.sendgrid.com/v3/templates"
+
+			client := s.client
+			if client == nil {
+				client = defaultClient
+			}
 
 			// test `read_user` scope
 			req, err := http.NewRequestWithContext(ctx, "GET", baseURL, nil)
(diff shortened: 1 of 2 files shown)


pkg/detectors/sendgrid/sendgrid.go
```diff
@@ -67,16 +74,18 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				// 403 means good key but not the right scope
 				// 401 is bad key
 				if res.StatusCode == http.StatusOK || res.StatusCode == http.StatusForbidden {
-					s.Verified = true
+					s1.Verified = true
```
ahrav, line 77: question: How come we are setting verified to true if the response has a 403 status code. Is it b/c we are assuming the credential is valid, and it just doesn't have access to this particular resource? Are we also certain that the `/templates` endpoint is the least restrictive, while still requiring auth? This might also just also just be a normal pattern with detectors where we consider 403s verified.

ahrav (commented):
General question for detectors, do we consider 403s as verified?

pkg/detectors/sendgrid/sendgrid.go
zubairk14: this 403 being valid but incorrect scope might be a sendgrid specific thing. i've gotten 403 on completely invalid API keys before.

pkg/detectors/sendgrid/sendgrid_test.go
```diff
@@ -18,7 +19,7 @@ import (
 func TestSendgrid_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors2")
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
```
zubairk14, line 22: did we add a new secret?

0x1: i did yeah

pkg/detectors/sendgrid/sendgrid.go
0x1: > How come we are setting verified to true if the response has a 403 status code. Is it b/c we are assuming the credential is valid, and it just doesn't have access to this particular resource? 
correct
> Are we also certain that the /templates endpoint is the least restrictive, while still requiring auth? 
no not certain, it's what we had before - there might potentially be a better endpoint but i didn't put in the time to research that  😅

rosecodym:
> General question for detectors, do we consider 403s as verified?

I am unaware of a general pattern - I think it's going to be verifier-specific. Generally speaking, I would consider a secret verified if the verification endpoint indicated that the request was _authenticated_ but not _authorized to perform the action we're using to verify_, but I can see myself convinced by a persuasive counterargument.

pkg/detectors/sendgrid/sendgrid_test.go
```diff
@@ -76,22 +80,57 @@ func TestSendgrid_FromChunk(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			s := Scanner{}
-			got, err := s.FromData(tt.args.ctx, tt.args.verify, tt.args.data)
+			got, err := tt.s.FromData(tt.args.ctx, tt.args.verify, tt.args.data)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("Sendgrid.FromData() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
 			for i := range got {
 				got[i].Raw = nil
```
rosecodym, line 130: with the switch to `cmp.Diff` you don't need to clear `Raw`

0x1:
> General question for detectors, do we consider 403s as verified?

it depends on the detector and how the api responds. if it's certain that for a given API, 403 will always result from a lack of permissions && not from failed authentication then i'd say yes.

rosecodym (approved):
nice!

PR #1737, opened 2023-08-31
updating sendbird detector to use tri-state verification

0x1:
### Description:
updating Sendbird detector to use tri-state verification

### Checklist:

### DIFF
--- a/pkg/detectors/sendbird/sendbird.go
+++ b/pkg/detectors/sendbird/sendbird.go
@@ -2,6 +2,7 @@ package sendbird
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"regexp"
@@ -12,18 +13,24 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat   = regexp.MustCompile(detectors.PrefixRegex([]string{"sendbird"}) + `\b([0-9a-f]{40})\b`)
 	appIdPat = regexp.MustCompile(detectors.PrefixRegex([]string{"sendbird"}) + `\b([0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12})\b`)
 )
+
+type userResp struct {
+	Code int `json:"code"`
+}
 
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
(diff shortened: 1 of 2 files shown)


pkg/detectors/sendbird/sendbird_test.go
```diff
@@ -97,7 +151,8 @@ func TestSendbird_FromChunk(t *testing.T) {
 				}
 				got[i].Raw = nil
```
rosecodym, line 152: with the switch to `cmp.Diff` you don't need to clear `Raw` here anymore

pkg/detectors/sendbird/sendbird_test.go
```diff
@@ -79,14 +82,65 @@ func TestSendbird_FromChunk(t *testing.T) {
...
+					DetectorType: detectorspb.DetectorType_Sendbird,
+					Verified:     false,
+				},
+			},
+			wantErr:             false,
+			wantVerificationErr: true,
+		},
+		{
+			name: "found, verified but unexpected api surface",
+			s:    Scanner{client: common.ConstantResponseHttpClient(404, "")},
```
rosecodym, line 108: There's a comment in the verification code that indicates that Sendbird returns a 400 for all errors. Does that include what we'd normally expect a 404 for? If so, this test should probably test a 500 or 502 or 503 or something (and be renamed).

pkg/detectors/sendbird/sendbird.go
```diff
@@ -61,20 +68,42 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-					} else {
-						// This function will check false positives for common test words, but also it will make sure the key appears 'random' enough to be a real key.
-						if detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
-							continue
+					} else if res.StatusCode == 400 { // Sendbird returns 400 for all errors
+						var userResp userResp
+						err := json.NewDecoder(res.Body).Decode(&userResp)
+						if err != nil {
+							s1.VerificationError = fmt.Errorf("error decoding json response body: %w", err)
+						}
```
rosecodym, line 87: It looks like you keep on truckin' even if the JSON couldn't be decoded - was that intentional?

rosecodym (commented):
nice digging! i have two nits and one question about intentional functionality.

pkg/detectors/sendbird/sendbird_test.go
0x1: no, we can still get a 404. for example if we called an endpoint that didn't exist (maybe bc the api changed) we can get a 404. in that case, we just don't get an extra error code in the response (which we mostly get with 400, but also a few other status codes: 403, 429, 500, 503)

pkg/detectors/sendbird/sendbird.go
0x1: yes, i assumed if the json decode failed that would still count as an indeterminate case. do you have a different opinion?

0x1: i guess we can potentially update the next if condition to be `else if userResp.Code != 400401` such that it doesn't run if decode fails?

rosecodym: yeah, that's where i was going. as written right now, you set the error information, but then immediately check `userResp.Code`, which will always be zero, meaning you'll immediately overwrite the error information you just set with `fmt.Errorf("unexpected response code: %d", userResp.Code)`, which is inaccurate.

rosecodym (approved):
:heart: the tests

PR #1740, opened 2023-09-01
Refactor SourceManager to remove Enrollment

mcastorina:
### Description:
Initializing the Source will be the responsibility of the caller. The
SourceManager exposes a GetIDs method for getting a source and job ID.

### Checklist:

### DIFF
--- a/pkg/engine/circleci.go
+++ b/pkg/engine/circleci.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/sourcespb"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources/circleci"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/engine/circleci.go
```diff
@@ -27,17 +26,13 @@ func (e *Engine) ScanCircleCI(ctx context.Context, token string) error {
...
-			if err := circleSource.Init(ctx, "trufflehog - Circle CI", jobID, sourceID, true, &conn, runtime.NumCPU()); err != nil {
-				return nil, err
-			}
-			return &circleSource, nil
-		})
-	if err != nil {
+	sourceName := "trufflehog - Circle CI"
+	sourceID, jobID, _ := e.sourceManager.GetIDs(ctx, sourceName, new(circleci.Source).Type())
+
+	circleSource := &circleci.Source{}
```
rosecodym, line 32: this is definitely just a weird little nit but if you're going to create a new source anyway, why not use that one to get a type to pass to `GetIDs`?

rosecodym: also, is discarding the error here the right thing to do?

rosecodym (commented):
this looks really, really clean! i only have one question about discarding errors

ahrav (approved):

pkg/engine/circleci.go
mcastorina: yeah, it's definitely weird. I have it this way because I only want to get the source type to pass to `GetIDs`. I'll make a followup PR to add constants to each source and use that instead of creating an object just to call `Type()`.

> also, is discarding the error here the right thing to do?

For now, the engine always uses the default API implementation (i.e. no API) which never fails. I could add error checks in case that ever changes though.

mcastorina: followup PR: https://github.com/trufflesecurity/trufflehog/pull/1768

Issue #1460, opened 2023-07-06
Ignore known false-positives for the Parseur detector


rgmz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The Parseur detector is prone to false positives in `package.json`, `yarn.lock`, `yarn-error.log`, and `package-lock.json` (I haven't personally seen this yet but I would assume so) files due to the [`parseuri`](https://www.npmjs.com/package/parseuri) and [`parseurl`](https://www.npmjs.com/package/parseurl) packages, which are common dependencies.

Examples:

- node_modules/parseuri/package.json
  ```json5
  {
    // ...
    "dist": {
      "shasum": "80204a50d4dbb779bfdc6ebe2778d90e4bce320a",
      "tarball": "https://registry.npmjs.org/parseuri/-/parseuri-0.0.5.tgz"
    },
    "gitHead": "792c9a63162a4484eb6b4f95fc611ccf224a24b6",
    // ...
  }

  // Raw result: 792c9a63162a4484eb6b4f95fc611ccf224a24b6
  ```
- yarn.lock
  ```yml
  parseurl@~1.3.1:
    version "1.3.1"
    resolved "https://registry.yarnpkg.com/parseurl/-/parseurl-1.3.1.tgz#c8ab8c9223ba34888aa64a297b28853bec18da56"

  # Raw result: c8ab8c9223ba34888aa64a297b28853bec18da56
  ``` 

## Problem to be Addressed

Ignore known false positives to reduce requests to Parseur's API and noise in `--no-verification` mode.

## Description of the Preferred Solution

There are two solutions that come to mind:

1. Improve TruffleHog's PrefixRegex/keyword logic so that it's possible to specify a match for "parseur" and not word collisions like "parseurl" or "parseuri". This would also help reduce false positives for other patterns like #1456.
2. Implement special logic for `package.json`, `package-lock.json`, and `yarn.lock` files to either:
  a. ignore them entirely
  b. ignore known bad matches like `parseurl/-/parseurl-x.x.x.tgz#c8ab8c9223ba34888aa64a297b28853bec18da56`.

[...]

PR #1741, opened 2023-09-01
Skip detectors for known bad chunks

### LINKED ISSUE #1517: Pass additional `Chunk` information to detectors
### Community Note

### Description

Presently, detectors have no knowledge of the source (e.g., "Git") or metadata (e.g., ["file: package-lock.json"](https://github.com/trufflesecurity/trufflehog/blob/20b77938285b82bc80531ba176989b7f8bae8c4b/pkg/engine/git.go#L50-L63)), and only receive a stream of bytes.

https://github.com/trufflesecurity/trufflehog/blob/20b77938285b82bc80531ba176989b7f8bae8c4b/pkg/detectors/detectors.go#L20-L21

While this design makes sense given TruffleHog's goal of scanning a multitude of sources (e.g., Git, Confluence, Slack), the lack of contextual information limits the power/usefulness of the detectors. For example, you cannot skip known bad filetypes like `yarn.lock` (#1460)[^1], nor can you write filetype/language-specific rules like [checking for JDBC credentials in .java/JVM code](https://github.com/dbeaver/dbeaver/blob/a66e69a30cf5dec82c9fec488e7991cb6f15aca9/plugins/org.jkiss.dbeaver.ext.mysql/src/MySQLErrorsTest.java#L24)[^2].

[^1]: As far as I can tell
[^2]: You can write that rule, however, it seems like it would run on every chunk which could adversely affect performance.

## Problem to be Addressed

Provide more context to Decoders so that it's possible to ignore known bad files/filetypes and write file/filetype-specific rules.

Two potential solutions come to mind:

1. Replace the `FromData(ctx context.Context, verify bool, data []byte) ([]Result, error)` function with `FromChunk(ctx context.Context, chunk Chunk) ([]Result, error)`
  https://github.com/trufflesecurity/trufflehog/blob/20b77938285b82bc80531ba176989b7f8bae8c4b/pkg/sources/sources.go#L14C1-L29
2. Alter the signature of `FromData` to include `SourceType` as well as `SourceMetadata` (presumably you'd want `SourceType` to make pulling relevant metadata easier).

## Additional Context

N/A

### References

* #1460 
* #1456

rgmz:
### Description:

This is a POC to fix #1517. Skipping detectors on chunks that are known to be problematic (e.g., #1460) should improve performance by reducing the number of false-positives and extraneous network requests.

Any feedback and suggestions are welcome. (Also, I have yet to test whether this specific code works; emphasis on "concept".)

### Checklist:

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"math/big"
 	"net/url"
+	"regexp"
 	"strings"
 	"unicode"
 
@@ -42,6 +43,27 @@ type CustomResultsCleaner interface {
 	// user-provided configuration that controls whether results are cleaned. (User-provided configuration is not the
 	// only factor that determines whether the engine runs cleaning logic.)
 	ShouldCleanResultsIrrespectiveOfConfiguration() bool
+}
+
+// ConditionalDetector is an optional interface that a detector can implement to
+// skip chunks based on specific criteria.
+type ConditionalDetector interface {
+	// ShouldScanChunk determines whether the detector should run.
+	ShouldScanChunk(chunk sources.Chunk) bool
+}
+
+var lockFilePat = regexp.MustCompile(`(^|/)(package(-lock)?\.json|yarn\.lock)$`)
+
+// Conditions is a set of common conditions to be used by ConditionalDetector.
+// (Using anonymous structs is weird, but Go has no concept of static members... https://stackoverflow.com/a/55390104)
+var Conditions = struct {
+	// LockFiles are a common source of false-positives.
+	// https://github.com/trufflesecurity/trufflehog/issues/1460
+	IsLockFile func(path string) bool
+}{
+	IsLockFile: func(path string) bool {
+		return lockFilePat.MatchString(path)
+	},
 }
 
 // Versioner is an optional interface that a detector can implement to
--- a/pkg/detectors/parseur/parseur.go
+++ b/pkg/detectors/parseur/parseur.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
@@ -18,6 +20,7 @@ type Scanner struct {
 
 // Ensure the Scanner satisfies the interface at compile time
 var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.ConditionalDetector = (*Scanner)(nil)
 
 var (
 	defaultClient = common.SaneHttpClient()
@@ -30,6 +33,13 @@ var (
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
 	return []string{"parseur"}
+}
+
+func (s Scanner) ShouldScanChunk(chunk sources.Chunk) bool {
+	if m, ok := sources.NewGitSourceMetadata(chunk.SourceType, chunk.SourceMetadata); ok {
+		return !detectors.Conditions.IsLockFile(m.File)
+	}
+	return true
 }
 
 // FromData will find and optionally verify Parseur secrets in a given set of bytes.
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -45,6 +45,42 @@ type Chunk struct {
 	Verify bool
 }
 
+// GitSourceMetadata defines a common struct for Git-based source metadata.
+type GitSourceMetadata struct {
+	Repository string
+	Commit     string
+	File       string
+}
+
+func NewGitSourceMetadata(source sourcespb.SourceType, data *source_metadatapb.MetaData) (*GitSourceMetadata, bool) {
+	if data == nil {
+		return nil, false
+	}
+
+	switch source {
+	case sourcespb.SourceType_SOURCE_TYPE_GIT:
+		md := data.GetGit()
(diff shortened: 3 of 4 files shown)


pkg/detectors/parseur/parseur.go
```diff
@@ -30,6 +33,14 @@ func (s Scanner) Keywords() []string {
 	return []string{"parseur"}
 }
 
+func (s Scanner) ScanChunk(chunk sources.Chunk) bool {
+	// TODO: Can |chunk.SourceMetadata| be nil?
+	if m, ok := chunk.SourceMetadata.GetData().(sources.GitSourceMetadata); ok {
```
rgmz, line 38: Just tested this. Sadly, it does not work — life is not so easy.

rosecodym: Do you mean that you need to explicitly attempt to unmarshal the metadata to determine whether it's from Git?

pkg/detectors/detectors.go
```diff
@@ -27,6 +28,23 @@ type Detector interface {
 	Type() detectorspb.DetectorType
 }
 
+// ConditionalDetector is an optional interface that a detector can implement to
+// skip chunks based on specific criteria.
+type ConditionalDetector interface {
+	// ScanChunk determines whether the detector should run.
+	ScanChunk(chunk sources.Chunk) bool
```
rosecodym, line 35: this is minor, but i double-took at the verb `Scan` here. i expected something like `ShouldScanChunk` instead.

pkg/sources/sources.go
```diff
@@ -38,6 +38,14 @@ type Chunk struct {
...
 }
 
+// GitSourceMetadata defines a common interface for Git-based source metadata.
+// For example, this should match Git, Azure, Bitbucket, GitHub, and Gitlab.
+type GitSourceMetadata interface {
+	GetRepository() string
+	GetCommit() string
+	GetFile() string
+}
+
```
rosecodym, line 48: I really like this idea, assuming we can find a performant way to actually use it.

pkg/detectors/detectors.go
```diff
@@ -27,6 +28,23 @@ type Detector interface {
...
+
+// FilenameConditions is a set of common conditions to be used by ConditionalDetector.
+// (Using anonymous structs is weird, but Go has no concept of static members... https://stackoverflow.com/a/55390104)
+var FilenameConditions = struct {
+	// LockFiles are a common source of false-positives.
+	// https://github.com/trufflesecurity/trufflehog/issues/1460
+	LockFiles *regexp.Regexp
+}{
+	LockFiles: regexp.MustCompile(`(^|/)(package(-lock)?\.json|yarn\.lock)$`),
+}
```
rosecodym, line 46: What do you think of hiding this behind an `IsLockFile` function?

rosecodym (commented):
I like smuggling the interface to avoid changing the entire `Detector` signature, and I like the idea of using chunk metadata to help make the decision about whether a given detector should scan a given chunk - assuming we can find a more performant way to do the metadata type check than repeated attempts at unmarshaling.

As for the implied question of whether this is even worth doing: You've convinced me, but I'll leave space for others to weigh in if they have other opinions.

pkg/detectors/parseur/parseur.go
rgmz: I thought that `GetData()` returned a top-level object, however, it actually returns an object like `{"bitbucket":{...}} | {"github":{...}} | {"gitlab":{...}}`. Only calling `GetX()` returns the inner object.

We'd need to construct an ugly switch statement to handle each SourceType, similar to: https://github.com/trufflesecurity/trufflehog/blob/b03cc302630ead9afbce8c16c74bf2c29123c0c1/pkg/engine/engine.go#L605-L609.

pkg/detectors/detectors.go
rgmz: It was originally `ShouldScanChunk`, I'm not sure why I removed `Should`; naming things is hard.

Do you prefer `ShouldScanChunk` or `MatchesChunk`?

pkg/sources/sources.go
rgmz: What are the performance implications of this right now?

pkg/sources/sources.go
```diff
@@ -38,6 +38,42 @@ type Chunk struct {
 	Verify bool
 }
 
+// GitSourceMetadata defines a common struct for Git-based source metadata.
+type GitSourceMetadata struct {
+	Repository string
+	Commit     string
+	File       string
```
rgmz, line 45: Should these be pointers?

rgmz:
@rosecodym I've updated the implementation based on feedback and testing. Let me know your thoughts on this.

pkg/detectors/detectors.go
rosecodym: My concern with `MatchesChunk` is that the term "matches" is used elsewhere in detector code and novices might be tempted to conflate the two senses. I think `ShouldScanChunk` is fine and the least likely to confuse.

rgmz: Makes sense. I've changed the name.

pkg/sources/sources.go
rgmz: Ah, you have to cast the `struct` to `interface{}` first.
```
md := data.GetGit()
if d, ok := interface{}(md).(GitSource); ok {
   d.GetFile() // do something
}
```

rosecodym:
@rgmz we just had an internal conversation about this and a question came up: How much of this problem is associated with lockfiles specifically? If the answer is "a lot," does it make sense to either use file exclusions or to consider skipping them at the engine level rather than adding complexity to detectors?

rgmz:
I wouldn't say the problem is specific to lock files. There are lots of files that create a high volume of false detections for specific detectors, that doesn't mean the files should be skipped altogether — [even lock files can contain secrets](https://github.com/antinomy-studio/size/blob/62c63809029385347c3c67205ceadc30b3115823/package-lock.json#L7-L9).

PR #1743, opened 2023-09-01
[bug] - Correctly create azure git links

ahrav:
### Description:
Correctly generate the link for an azure commit.

### Checklist:


rosecodym (approved):
did they change something or was this never correct?

ahrav:
> did they change something or was this never correct?

Was never correct. Found out when i tried to navigate to the link of a secret after vendored into enterprise.

PR #1746, opened 2023-09-04
[chore] - Sentry detector update

ahrav:
### Description:
Add test case for account deactivated and when the API returns an empty result array. `[]`

### Checklist:

### DIFF
--- a/pkg/detectors/sentrytoken/sentrytoken_test.go
+++ b/pkg/detectors/sentrytoken/sentrytoken_test.go
@@ -124,6 +124,23 @@ func TestSentryToken_FromChunk(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name: "found, account deactivated",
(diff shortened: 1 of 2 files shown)


pkg/detectors/sentrytoken/sentrytoken_test.go
```diff
@@ -174,6 +209,9 @@ const responseBody403 = `
   }
 ]
 `
+	reponseAccountDeactivated = `{"datail": "Authentication credentials were not provided"}`
```
0x1, line 212: should this be `detail`?

0x1 (approved):
lgtm other than clarification on `datail`

pkg/detectors/sentrytoken/sentrytoken_test.go
ahrav: ah yea, it shouldn't affect the test since we are expecting a completely different structure. Nonetheless, fixed. TY.

PR #1747, opened 2023-09-05
Add optional param to Chunks

ahrav:
### Description:
Provide an optional ChunkgTarget to Chunks which can be used when only targeting specific data from a source.

### Checklist:


ahrav:
@rosecodym what's your opinion on this interface and method name related to secret reverification? I aimed for naming consistency with our existing Chunks method due to their similarity. However, the naming for both doesn't clearly indicate the logic—querying data and then chunking. Do you think it aligns with our intent?

ahrav:
As per a discussion w/ @rosecodym we have decided to forego the additional interface and use an "optional" variadic parameter instead.

rosecodym (approved):

PR #1749, opened 2023-09-05
Fix pagerdutyapikey Detector

### LINKED ISSUE #1658: PagerDuty detector does not detect secrets
### TruffleHog Version
```
$ trufflehog --version
trufflehog 3.53.0
```

### Expected Behavior

Trufflehog should detect my PagerDuty API Key

### Actual Behavior

Trufflehog did not detect my PagerDuty API Key

### Steps to Reproduce

1. Generate a PagerDuty API key (or several General Access keys and User Tokens, to get a variety of keys/formats)
2. Create a file called `secrets` where each line contains `pagerduty <api_key_here>`
3. `trufflehog filesystem ./secrets`
4. Keys aren't detected (the regex doesn't find them)

## Environment
 * OS: MacOS
 * Version 13.4.1

## Additional Context

On further investigation, it looks like neither of PagerDuty's General Access tokens or User Tokens are detected by Trufflehog's regex. 

It seems like some assumptions are made about the location/existence of certain special characters (`-`, `_`, and `+`) that aren't always correct. Most tokens that I generate don't have special characters in the locations that Trufflehog expects, and some don't have those special characters at all.

I've reached out to PagerDuty and asked if they can provide regex patterns for their keys. It seems like a catch-all pattern that could be used is `\b([a-zA-Z0-9\-\_\+]{20})\b`, but I can wait to hear back from PagerDuty for something more accurate.

I'd also like to add `pd` to the list of PagerDuty keywords, as I've seen that more commonly than `pagerduty`.

Happy to assign myself to fix this

harmonherring-pro:
### Description:
Resolves #1658 by fixing `pagerdutyapikey` detector

### Checklist:

### DIFF
--- a/pkg/detectors/pagerdutyapikey/pagerdutyapikey.go
+++ b/pkg/detectors/pagerdutyapikey/pagerdutyapikey.go
@@ -21,13 +21,13 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"pagerduty"}) + `\b([a-zA-Z0-9_+-]{20})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"pagerduty", "pd"}) + `\b([a-zA-Z0-9_+-]{20})\b`)
 )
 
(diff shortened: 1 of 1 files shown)


ahrav (approved):
Thanks a lot @harmonherring-pro, appreciate you updating this detector.

PR #1750, opened 2023-09-06
Add log verbosity

codevbus:
### Description:
Adds and updates some verbosity settings

### Checklist:


pkg/sources/s3/s3.go
```diff
@@ -278,7 +278,7 @@ func (s *Source) pageChunker(ctx context.Context, client *s3.S3, chunksChan chan
 
 		// ignore large files
 		if *obj.Size > s.maxObjectSize {
-			s.log.V(3).Info("Skipping %d byte file (over maxObjectSize limit)", "object", *obj.Key)
+			s.log.V(5).Info("Skipping %d byte file (over maxObjectSize limit)", "object", *obj.Key)
```
rosecodym, line 281: i guess we're seeing a lot of these?

rosecodym (approved):

pkg/sources/s3/s3.go
codevbus: Doesn't look like it with a cursory log search but I think this was a specific ask from @dustin-decker

PR #1754, opened 2023-09-07
cache dupes w/ different decoders

ahrav:
### Description:
To manage recurring findings among various decoders, we cached secrets to eliminate repetitive listings. However, this unintentionally omitted repeated results within the same decoder. Our goal: Detail EVERY secret location if existing.

Use case:
In TH with pre-commit, if a secret appears in several spots, only one gets marked. Users remove the marked secret and attempt another commit. This fails since the secret still lurks elsewhere. Displaying all secret locations enables users to address all instances prior to committing.

Also moved up the initial print to before we complete a scan as that was bothering me 😃 
### Checklist:


rosecodym:
is it tractable to add an automated test for this?

ahrav:
added test

rosecodym (approved):

PR #1755, opened 2023-09-08
Include flag for `--fail-fast`

### LINKED ISSUE #1287: Feature Request: Have the "fail" flag die immediately if any results are found
### Description

I want a flag that stops scanning if any results are found (pursuant to other filters).

## Problem to be Addressed

We have hundreds of repositories, and on average, scanning one repository takes about half an hour. I have a work project right now that wants me to get _whether_ there are secrets, not what the secrets _are_.

The easiest way would be to make a change in `main.go` to add something after line 442 like: "if fail is true, die right now"

(I could try a pull request, but I've never programmed in Go.)

ahrav:
### Description:
When supplied, the --fail-flag causes the scanner to exit immediately upon discovering a single result. It acknowledges and works in tandem with other flags. So, if paired with --only-verified, the scanner stops at the first verified credential.

Also make notifying thread-safe since we are printing results we don't need multiple notifier workers.

Screenshot 2023-09-07 at 7 14 50 PM

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -433,22 +433,15 @@ func Start(ctx context.Context, options ...EngineOption) (*Engine, error) {
 		}()
 	}
 
-	// We want 1/4th of the notifier workers as the number of scanner workers.
-	const notifierWorkerRatio = 4
-	maxNotifierWorkers := 1
-	if numWorkers := e.concurrency / notifierWorkerRatio; numWorkers > 0 {
-		maxNotifierWorkers = int(numWorkers)
-	}
-	ctx.Logger().V(2).Info("starting notifier workers", "count", maxNotifierWorkers)
-	for worker := 0; worker < maxNotifierWorkers; worker++ {
-		e.WgNotifier.Add(1)
-		go func() {
-			ctx := context.WithValue(ctx, "notifier_worker_id", common.RandomID(5))
-			defer common.Recover(ctx)
-			defer e.WgNotifier.Done()
-			e.notifyResults(ctx)
-		}()
-	}
+	// We want only one notifier worker.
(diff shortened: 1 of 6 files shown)


pkg/engine/engine.go
```diff
@@ -392,6 +458,17 @@ func Start(ctx context.Context, options ...EngineOption) (*Engine, error) {
 // more sources may be scanned by the engine.
 func (e *Engine) Finish(ctx context.Context) error {
 	defer common.RecoverWithExit(ctx)
+
+	// Handle the fail-fast termination signal.
+	e.termSignal.mutex.Lock()
+	isFailFast := e.termSignal.isFailFast
```
0x1, line 464: is there a reason why `isFailFast` is stored in a variable first instead of checking it directly in the conditional? and is there some chance that ` e.termSignal.isFailFast` could change between when it's unlocked and when we check `isFailFast`?

ahrav: Correct. isFailFast can be modified within `notifyResults` which is run by multiple workers. This way we are ensuring the state hasn't been changed.

On a slightly related note, if we remove the concurrent notification, we could probably do away with the mutex altogether.

ahrav: Although, actually we would still need the mutex in order to use the condition variable :(

mcastorina (commented):
I'm curious if you explored using context cancellation as a simpler implementation, and if it didn't work, why (for my own edification).

zricethezav:
@ahrav is this a feature we still want to pursue?

ahrav:
> @ahrav is this a feature we still want to pursue?

I don't think so. I'll close this.

PR #1757, opened 2023-09-08
Retry AWS verification 403s

rosecodym:
### Description:
This PR introduces retries on 403s in the AWS detector in attempt to work around erroneous `SignatureDoesNotMatch` errors. As part of the work, the detector has been refactored somewhat, resulting in two minor semantic changes:
* Errors crafting the verification HTTP request no longer result in the candidate secret being discarded.
* The known-words-based false positive check now runs (and potentially discards candidate secrets) even if verification is disabled. This unifies its behavior with the hash-based false positive check.

### Checklist:

### DIFF
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -104,8 +104,6 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 		}
 
-		//fmt.Printf("=========== the number of secret matches %d", len(secretMatches))
-
 		for _, secretMatch := range secretMatches {
 			if len(secretMatch) != 2 {
 				continue
@@ -124,7 +122,6 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 
 			results = append(results, s1)
-			//fmt.Printf("%+v\n", results)
 			// If we've found a verified match with this ID, we don't need to look for any more. So move on to the next ID.
 			if s1.Verified {
 				break
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -177,54 +177,12 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				if client == nil {
 					client = defaultVerificationClient
 				}
-				res, err := client.Do(req)
-				if err == nil {
 
-					if res.StatusCode >= 200 && res.StatusCode < 300 {
-						identityInfo := identityRes{}
-						err := json.NewDecoder(res.Body).Decode(&identityInfo)
-						if err == nil {
-							s1.Verified = true
-							s1.ExtraData = map[string]string{
-								"account": identityInfo.GetCallerIdentityResponse.GetCallerIdentityResult.Account,
-								"user_id": identityInfo.GetCallerIdentityResponse.GetCallerIdentityResult.UserID,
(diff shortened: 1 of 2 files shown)


pkg/detectors/aws/aws.go
```diff
@@ -242,6 +145,130 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return awsCustomCleanResults(results), nil
 }
 
+func (s scanner) verifyMatch(ctx context.Context, resIDMatch, resSecretMatch string, retryOn403 bool) (verified bool, extraData map[string]string, verificationErr error) {
```
rosecodym, line 148: i do not like this signature but i can't think of a version i like better

ahrav (approved):

trufflesteeeve (approved):

pkg/detectors/aws/aws.go
```diff
@@ -242,6 +145,125 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+
+	client := s.verificationClient
+	if client == nil {
+		client = defaultVerificationClient
+	}
+
+	res, err := client.Do(req)
+	if err == nil {
+
+		if res.StatusCode >= 200 && res.StatusCode < 300 {
```
ahrav, line 212: Don't hate me too much, but what if we split this up into something like:
```
...
res, err := client.Do(req)
	if err != nil {
		return false, nil, err
	}
	defer res.Body.Close()

	switch {
	case res.StatusCode >= http.StatusOK && res.StatusCode < http.StatusMultipleChoices:
		return s.handleSuccessResponse(res)
	case res.StatusCode == http.StatusForbidden:
		return s.handleForbiddenResponse(ctx, resIDMatch, resSecretMatch, retryOn403, res)
	default:
		return false, nil, fmt.Errorf("request to %v returned unexpected status %d", res.Request.URL, res.StatusCode)
	}
...


func (s scanner) handleSuccessResponse(res *http.Response) (bool, map[string]string, error) {
	identityInfo := identityRes{}
	if err := json.NewDecoder(res.Body).Decode(&identityInfo); err != nil {
		return false, nil, err
	}
	extraData := map[string]string{
		"account": identityInfo.GetCallerIdentityResponse.GetCallerIdentityResult.Account,
		"user_id": identityInfo.GetCallerIdentityResponse.GetCallerIdentityResult.UserID,
		"arn":     identityInfo.GetCallerIdentityResponse.GetCallerIdentityResult.Arn,
	}
	return true, extraData, nil
}

func (s scanner) handleForbiddenResponse(ctx context.Context, resIDMatch, resSecretMatch string, retryOn403 bool, res *http.Response) (bool, map[string]string, error) {
	if retryOn403 {
		return s.verifyMatch(ctx, resIDMatch, resSecretMatch, false)
	}
	var body awsErrorResponseBody
	if err := json.NewDecoder(res.Body).Decode(&body); err != nil {
		return false, nil, fmt.Errorf("couldn't parse the sts response body (%v)", err)
	}
	if strings.EqualFold(body.Error.Code, "InvalidClientTokenId") {
		return false, nil, nil
	}
	return false, nil, fmt.Errorf("request to %v returned status %d with an unexpected reason (%s: %s)", res.Request.URL, res.StatusCode, body.Error.Code, body.Error.Message)
}
```

ahrav (approved):

pkg/detectors/aws/aws.go
ahrav: Plus all the comments ofc

ahrav (approved):

PR #1760, opened 2023-09-09
Improve private key detector

dustin-decker:
### Description:
- Surfaces verification information in extra_data rather than structured_data
- Verifies directly against GitHub and GitLab rather than the stale dataset of GitHub accounts in Driftwood

### Checklist:


pkg/detectors/privatekey/privatekey.go
```diff
@@ -53,28 +55,76 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
...
+			} else {
+				secret.VerificationError = err
+			}
+
+			user, err := verifyGitHubUser(parsedKey)
+			if user != nil {
+				secret.Verified = true
+				secret.ExtraData["github_user"] = *user
+			}
+			if err != errPermissionDenied {
```
ahrav, line 98: suggestion: `!errors.Is(err, errPermissionDenied)`

pkg/detectors/privatekey/privatekey.go
```diff
@@ -53,28 +55,76 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
...
+				secret.VerificationError = err
+			}
+
+			user, err = verifyGitLabUser(parsedKey)
+			if user != nil {
+				secret.Verified = true
+				secret.ExtraData["gitlab_user"] = *user
+			}
+			if err != errPermissionDenied {
```
ahrav, line 107: ditto: `!errors.Is(err, errPermissionDenied)`

pkg/detectors/privatekey/privatekey.go
```diff
@@ -100,35 +150,12 @@ func lookupFingerprint(publicKeyFingerprintInHex string, includeExpired bool) (d
...
-		if data.GithubSshKey == nil {
-			data.GithubSshKey = make([]*detectorspb.GitHubSSHKey, 0)
-		}
-		data.GithubSshKey = append(data.GithubSshKey, &detectorspb.GitHubSSHKey{
-			User: r.Username,
-		})
-		seen[r.Username] = struct{}{}
-	}
-
 	return
```
ahrav, line 159: nit: I know you didn't touch this but could we get rid of the naked return.
`return data, nil`

pkg/detectors/privatekey/ssh.go
```diff
@@ -0,0 +1,123 @@
...
+		return "", err
+	}
+	_ = session.Wait()
+
+	return output.String(), err
+}
+
+var errPermissionDenied = errors.New("permission denied")
+
+func verifyGitHubUser(parsedKey any) (*string, error) {
```
ahrav, line 88: question: How come you decided to return `*string` instead of just `string` for the first return value?

pkg/detectors/privatekey/privatekey.go
```diff
@@ -53,28 +55,76 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
...
-				secret.StructuredData = data
 				if data != nil {
 					secret.Verified = true
+					secret.ExtraData["certificate_urls"] = strings.Join(data.CertificateURLs, ", ")
 				}
+			} else {
+				secret.VerificationError = err
+			}
+
+			user, err := verifyGitHubUser(parsedKey)
```
ahrav, line 93: question: How come here and below you chose to do the `user != nil` prior to the err check?

ahrav (approved):

pkg/detectors/privatekey/cracker_test.go
```diff
@@ -68,11 +68,15 @@ func Test_crack(t *testing.T) {
...
-			_, err := crack(tt.in)
+			_, passphrase, err := crack(tt.in)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("crack() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
+
+			if passphrase != string(testEncryptedKeyCorrectPassword) {
+				t.Errorf("crack() passphrase = %v, want %v", passphrase, string(testEncryptedKeyCorrectPassword))
+			}
```
rosecodym, line 79: I would expect this new check to require updating the test case itself. Does that need to be done?

pkg/detectors/privatekey/privatekey.go
```diff
@@ -53,28 +55,76 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
...
 		if verify {
 			data, err := lookupFingerprint(fingerprint, s.IncludeExpired)
 			if err == nil {
-				secret.StructuredData = data
 				if data != nil {
 					secret.Verified = true
+					secret.ExtraData["certificate_urls"] = strings.Join(data.CertificateURLs, ", ")
 				}
+			} else {
+				secret.VerificationError = err
```
rosecodym, line 90: This field isn't just informational - setting it indicates that `Verified` isn't trustworthy. We haven't yet sat down and figured out what that means when `Verified` is also set to true, which you've done later in the function. We've just been proscribing the combination entirely. (Put another way: We've added tri-state verification, but this combination is a _fourth_ state.)

The fact that this error is compatible with verifying the candidate key suggests to me that it's not actually a verification error, but possibly just another thing to put into `ExtraData`, but I could be misunderstanding what this is doing, in which case we should figure out how to update our conceptual model of indeterminacy.

(This isn't the only place in this function where you set this field in a code path that could also end up setting `Verified` to `true`, so pretend I left this comment in those other places too.)

rosecodym (changes requested):
If I read correctly, the updated detector can set a verification state that we haven't defined anywhere: verified but indeterminate. If that's really what needs to happen here, we need to figure out what that means semantically. (If not, I made a suggestion for a different thing to do inline.)

pkg/detectors/privatekey/cracker_test.go
dustin-decker: It didn't, but I've added the wanted passphrase as a parameter for if we add more cases.

pkg/detectors/privatekey/ssh.go
dustin-decker: Optional value

pkg/detectors/privatekey/privatekey.go
dustin-decker: Let me know what you think of the change. I've collected the errors into a slice and join them if they exists and verified is false.

rosecodym (approved):
verification indeterminacy looks good

PR #1761, opened 2023-09-10
[chore] - fix slackwebhook detector

ahrav:
### Description:
When an account is no longer active the response body contains `missing_text_or_fallback_or_attachments` which should be treated as an unverified result. Also update the secret in GSM, as the existing verified test case was failing.

### Checklist:


pkg/detectors/slackwebhook/slackwebhook.go
```diff
@@ -69,11 +70,10 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				if err != nil {
 					continue
 				}
-				body := string(bodyBytes)
 
 				defer res.Body.Close()
 
-				if res.StatusCode >= 200 && res.StatusCode < 300 || (res.StatusCode == 400 && (strings.Contains(body, "no_text") || strings.Contains(body, "missing_text"))) {
+				if res.StatusCode >= 200 && res.StatusCode < 300 || (res.StatusCode == 400 && (bytes.Equal(bodyBytes, []byte("no_text")) || bytes.Equal(bodyBytes, []byte("missing_text")))) {
```
rosecodym, line 76: What situations do these body checks actually handle?

ahrav: Depending on the version of slackwebhooks that were setup we get 400s with either "no_text" or "missing_text". These usually indicate the webhook is active but does not support empty text in the payload. At least that is my current understanding.

rosecodym (approved):
while you're here, do you mind adding tests for the 400 success situations?

rosecodym:
We're seeing this logic break in some circumstances - `missing_text_or_fallback_or_attachments` is apparently coming back for live webhooks. Additionally, [Slack's documentation](https://slack.com/help/articles/360000446446-Manage-deactivated-members-apps-and-integrations) says that incoming webhooks remain active when their creating user is deactivated, so I'm not sure what the actual expected behavior should be here.

PR #1762, opened 2023-09-11
update gitlabv2 to tri-state

0x1:
### Description:
updating gitlab v2 detector to use tri-state verification and fixing some of the tests in v1 and v2

### Checklist:


pkg/detectors/gitlab/gitlab.go
```diff
@@ -53,7 +53,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			resMatch = keyString[len(keyString)-1]
 		}
 
-		secret := detectors.Result{
+		s1 := detectors.Result{
```
rosecodym, line 56: Did you rename this for similarity to other detectors?

pkg/detectors/gitlab/gitlabv2_test.go
```diff
@@ -1,115 +0,0 @@
...
-	"testing"
-	"time"
-
-	"github.com/kylelemons/godebug/pretty"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
-)
-
-func TestGitlabv2_FromChunk(t *testing.T) {
```
rosecodym, line 15: Why did you delete this test?

pkg/detectors/gitlabv2/gitlab.go
```diff
@@ -42,7 +46,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			continue
 		}
 
-		secret := detectors.Result{
+		s1 := detectors.Result{
```
rosecodym, line 49: same question as v1 rename

pkg/detectors/gitlabv2/gitlab_test.go
```diff
@@ -31,18 +32,19 @@ func TestGitlab_FromChunk(t *testing.T) {
...
+		wantErr             bool
+		wantVerificationErr bool
 	}{
 		{
 			name: "found, verified",
 			s:    Scanner{},
 			args: args{
 				ctx:    context.Background(),
-				data:   []byte(fmt.Sprintf("You can find a aws secret %s within", secret)),
+				data:   []byte(fmt.Sprintf("You can find a gitlab secret %s within", secret)),
```
rosecodym, line 47: ❤️

rosecodym (commented):
my only blocking question is about the deleted test

pkg/detectors/gitlab/gitlab.go
0x1: yes, any detectors that i updated

pkg/detectors/gitlabv2/gitlab.go
0x1: yes, i updated this for any detectors i touched

pkg/detectors/gitlab/gitlabv2_test.go
0x1: because it was testing v2. i figured we should just have a test for v1 and a test for v2, not both. but i'm open to thoughts on if we should keep it

pkg/detectors/gitlabv2/gitlab.go
0x1: there's a few other detectors that still use `secret`. i was planning on updating them after all the tristate detectors have been updated (if they don't get updated during the process).

rosecodym:
per out-of-band conversation: this change is good, but we suspect it might not resolve the underlying flip-flopping issue. @0x1 is going to do some research to try to settle that question.

pkg/detectors/gitlab/gitlabv2_test.go
rosecodym: oh, i totally missed that it was a v2 test in the v1 directory. now that i see that, i completely agree.

0x1:
we will be consolidating the gitlab detectors instead of adding tri-state to v2. PR here: https://github.com/trufflesecurity/trufflehog/pull/1776

PR #1774, opened 2023-09-13
Update Source interface to use SourceID and JobID types

mcastorina:
### Description:
The previous implementation used int64 for both, which can be mixed up easily. Using distinct types adds a layer of type safety checked by the compiler.

### Checklist:

### DIFF
--- a/pkg/sources/gcs/gcs.go
+++ b/pkg/sources/gcs/gcs.go
@@ -45,12 +45,12 @@ func (s *Source) Type() sourcespb.SourceType {
 }
 
 // SourceID number for GCS Source.
-func (s *Source) SourceID() int64 {
+func (s *Source) SourceID() sources.SourceID {
 	return s.sourceId
 }
 
 // JobID number for GCS Source.
-func (s *Source) JobID() int64 {
+func (s *Source) JobID() sources.JobID {
 	return s.jobId
 }
 
@@ -62,8 +62,8 @@ type objectManager interface {
 // Source represents a GCS source.
 type Source struct {
 	name        string
-	jobId       int64
-	sourceId    int64
+	jobId       sources.JobID
+	sourceId    sources.SourceID
 	concurrency int
 	verify      bool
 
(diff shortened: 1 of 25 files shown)


pkg/sources/syslog/syslog.go
```diff
@@ -123,7 +123,7 @@ func (s *Source) InjectConnection(conn *sourcespb.Syslog) {
 }
 
 // Init returns an initialized Syslog source.
-func (s *Source) Init(_ context.Context, name string, jobId, sourceId int64, verify bool, connection *anypb.Any, concurrency int) error {
+func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourceId sources.SourceID, verify bool, connection *anypb.Any, concurrency int) error {
```
0x1, line 126: nit: any reason sometimes it's `jobID` like in line 52 and sometimes it's `jobId`?

pkg/sources/gcs/gcs.go
```diff
@@ -111,7 +111,7 @@ func (c *persistableCache) shouldPersist() (bool, string) {
 }
 
 // Init returns an initialized GCS source.
-func (s *Source) Init(aCtx context.Context, name string, id int64, sourceID int64, verify bool, connection *anypb.Any, concurrency int) error {
+func (s *Source) Init(aCtx context.Context, name string, id sources.JobID, sourceID sources.SourceID, verify bool, connection *anypb.Any, concurrency int) error {
```
0x1, line 114: nit: same question as jobID - looks like sometimes we use `sourceID` and sometimes it's `sourceId`. should this be standardized or do we use them for different purposes?

0x1 (approved):
when you say the implementation of int64 can be mixed up easily, do you mean by humans - or is there some Go thing that can happen? and what kind of additional type safety do we get beyond when it's just int64?


(other than the question^ and nits looks solid)

pkg/sources/syslog/syslog.go
mcastorina: No, just ambiguity in what "camel case" means. I asked Ahrav at one point and the official Google style guide recommends `jobID`, so I tend to use that.

https://google.github.io/styleguide/go/decisions#initialisms

I'll make a followup PR to update the variable names, since this PR is just updating the types.

mcastorina:
> when you say the implementation of int64 can be mixed up easily, do you mean by humans - or is there some Go thing that can happen? and what kind of additional type safety do we get beyond when it's just int64?

Yeah, by humans. Making them distinct types will allow the compiler to catch accidentally using a job ID as a source ID or vice versa.

```go
func Foo(sourceID int64, jobID int64)
func Bar(sourceID SourceID, jobID JobID)

Foo(jobID, sourceID) // no problemo (jobID and sourceID are both int64)
Bar(jobID, sourceID) // problemo (jobID and sourceID are distinct types)
```

rosecodym (approved):

PR #1780, opened 2023-09-18
added regex to detect the old format for dockerhub tokens

ankushgoel27:
This PR added to the dockerhub regex detecting the old format of the dockerhub token which is still active. Also added another parameter for username detection "id".

Tested and it works


ankushgoel27:
anything else needed on this PR?

PR #1781, opened 2023-09-18
fixed rubygems detector

ankushgoel27:
Fixed rubygems detector. If the token doesn't have enough privs, it could also give an HTTP forbidden error even on a valid token.

### DIFF
--- a/pkg/detectors/rubygems/rubygems.go
+++ b/pkg/detectors/rubygems/rubygems.go
@@ -47,7 +47,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			req, err := http.NewRequestWithContext(ctx, "GET", "https://rubygems.org/api/v1/gems", nil)
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://rubygems.org/api/v1/gems.json", nil)
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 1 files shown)


pkg/detectors/rubygems/rubygems.go
```diff
@@ -47,7 +47,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			req, err := http.NewRequestWithContext(ctx, "GET", "https://rubygems.org/api/v1/gems", nil)
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://rubygems.org/api/v1/gems.json", nil)
```
dustin-decker, line 50: Maybe we should use `GET - /api/v1/profile/me.json` instead?

It returns a body like this ```[
  {
    "id": 1,
    "handle": "qrush",
    "mfa": "enabled"
  }
]
```

We could surface `username` and `mfa_enabled` in the extra data field

dustin-decker: https://guides.rubygems.org/rubygems-org-api/

dustin-decker: It probably doesn't ever return forbidden either

ankushgoel27: GET - /api/v1/profile/me.(json|yaml)
View basic user information for your account, including Multi-factor authentication status. Requires username and password to be passed.

it requires username and password to be passed

dustin-decker (approved):

PR #1784, opened 2023-09-18
Update sonarcloud.go

ankushgoel27:
[Making the detector more braod. I have seen instances where the sonarcloud token mentions only sonar and not the full sonarcloud string.](https://github.com/trufflesecurity/trufflehog/pull/1783)


dustin-decker (approved):
Thank you

PR #1785, opened 2023-09-18
updating browserstack detector to use tri-state verification

0x1:
### Description:
updating browserstack detector regex to remove http and https prefixes, and the userpat regex to use a ? quantifier (bc some tokens may not contain `._-`
updating browserstack detector to use tri-state verification

### Checklist:

### DIFF
--- a/pkg/detectors/browserstack/browserstack.go
+++ b/pkg/detectors/browserstack/browserstack.go
@@ -2,6 +2,7 @@ package browserstack
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/detectors/browserstack/browserstack.go
```diff
@@ -56,29 +59,39 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 
 			if verify {
+				client := s.client
```
ahrav, line 62: Same here albeit this if/else isn't as deeply nested as the other PR. I don't feel too strongly about the refactor here, but if you feel like it, we can do something similar: 100% optional

```
for _, userMatch := range userMatches {
...
			if verify {
				client := s.client
				if client == nil {
					client = defaultClient
				}

				isVerified, verificationErr := verifyBrowserStackCredentials(ctx, client, resUserMatch, resMatch)
				s1.Verified = isVerified
				s1.VerificationError = verificationErr
			}

			if !s1.Verified && detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
				continue
			}
			results = append(results, s1)
..
}

const browserStackAPIURL = "https://www.browserstack.com/automate/plan.json"

func verifyBrowserStackCredentials(ctx context.Context, client *http.Client, username, accessKey string) (bool, error) {
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, browserStackAPIURL, nil)
	if err != nil {
		return false, err
	}
	req.SetBasicAuth(username, accessKey)

	res, err := client.Do(req)
	if err != nil {
		return false, err
	}
	defer res.Body.Close()

		return true, nil
	} else if res.StatusCode != 401 {
		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
	}

	return false, nil
}
```

ahrav (approved):

pkg/detectors/browserstack/browserstack.go
0x1: i like this approach - will continue on future detectors!

PR #1790, opened 2023-09-20
Add ability to dynamically scale concurrently running sources

mcastorina:
### Description:
Refactor SourceManager to use a counting semaphore to allow for dymanically changing limits. This complicated `Wait() error` which needs to return the first error encountered. We previously got that for free using `errgroup.Group`, however now we need to handle that ourselves. `Wait()` needs to return an error for use in the engine to set the correct exit code.

### Checklist:

### DIFF
--- a/pkg/sources/source_manager.go
+++ b/pkg/sources/source_manager.go
@@ -7,9 +7,9 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/marusama/semaphore/v2"
 	"golang.org/x/sync/errgroup"
 
-	"github.com/marusama/semaphore/v2"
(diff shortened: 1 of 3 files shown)


pkg/sources/source_manager.go
```diff
@@ -2,12 +2,14 @@ package sources
...
 import (
 	"fmt"
+	"runtime"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"golang.org/x/sync/errgroup"
 
+	"github.com/marusama/semaphore/v2"
```
dustin-decker, line 12: nit: group this with other third party dependencies

dustin-decker (approved):

PR #1792, opened 2023-09-20
updating microsoft teams webhook detector to use tri-state verification

0x1:
### Description:
updating microsoft teams webhook detector to use tri-state verification

### Checklist:

### DIFF
--- a/pkg/detectors/microsoftteamswebhook/microsoftteamswebhook.go
+++ b/pkg/detectors/microsoftteamswebhook/microsoftteamswebhook.go
@@ -50,37 +50,16 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			DetectorType: detectorspb.DetectorType_MicrosoftTeamsWebhook,
 			Raw:          []byte(resMatch),
 		}
+
 		if verify {
 			client := s.client
 			if client == nil {
 				client = defaultClient
 			}
 
-			payload := strings.NewReader(`{'text':''}`)
-			req, err := http.NewRequestWithContext(ctx, "POST", resMatch, payload)
-			if err != nil {
-				s1.VerificationError = err
-			} else {
-				req.Header.Add("Content-Type", "application/json")
-				res, err := client.Do(req)
-				if err != nil {
-					s1.VerificationError = err
-				} else {
-					body, err := io.ReadAll(res.Body)
-					res.Body.Close()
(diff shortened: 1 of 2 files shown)


pkg/detectors/microsoftteamswebhook/microsoftteamswebhook.go
```diff
@@ -48,19 +51,33 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			Raw:          []byte(resMatch),
 		}
 		if verify {
+			client := s.client
```
ahrav, line 54: Suggestion: All the logic looks great to me. The nested if/else is definitely a little hard to follow. Do you think we could do something like the following to reduce the nesting and maybe even make the flow a little more clear:
```
...
		if verify {
			client := s.client
			if client == nil {
				client = defaultClient
			}

			isVerified, verificationErr := verifyWebhook(ctx, client, resMatch)
			s1.Verified = isVerified
			s1.VerificationError = verificationErr
		}

		if !s1.Verified && detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, false) {
			continue
		}

		results = append(results, s1)
...
}

func verifyWebhook(ctx context.Context, client *http.Client, webhookURL string) (bool, error) {
	payload := strings.NewReader(`{'text':''}`)
	req, err := http.NewRequestWithContext(ctx, "POST", webhookURL, payload)
	if err != nil {
		return false, err
	}

	res, err := client.Do(req)
	if err != nil {
		return false, err
	}
	defer res.Body.Close()

	body, err := io.ReadAll(res.Body)
	if err != nil {
		return false, err
	}

	switch {
	case res.StatusCode == http.StatusBadRequest:
		if strings.Contains(string(body), "Text is required") {
			return true, nil
		}
		return false, fmt.Errorf("unexpected response body: %s", string(body))
	case res.StatusCode < 200 || res.StatusCode >= 500:
		return false, fmt.Errorf("unexpected HTTP response status: %d", res.StatusCode)
	default:
		return false, nil
	}
}
```

ahrav (approved):

pkg/detectors/microsoftteamswebhook/microsoftteamswebhook.go
0x1: love it, that's way cleaner!

PR #1803, opened 2023-09-21
use ptr for chunking target

ahrav:
### Description:
The ChunkingTarget contains a proto which has a mutex. In order to correctly handle passing the proto we need to use a ptr.

### Checklist:


mcastorina (approved):
Did you consider using a pointer in the definition of `ChunkingTarget`? Not sure if that would be beneficial or not over these changes.

ahrav:
> Did you consider using a pointer in the definition of `ChunkingTarget`? Not sure if that would be beneficial or not over these changes.

yea that's a great idea and I only realized it after i started. Going to close this PR and have it resolved in https://github.com/trufflesecurity/trufflehog/pull/1804

PR #1804, opened 2023-09-21
Github partial scan

ahrav:
### Description:
Include the ability for the Github source to handle partial scans of the source. This allows us to use the metadata for a secret to only scan the chunk the secret was found in. This can be used for efficient revivification of the secret.

### Checklist:

### DIFF
--- a/pkg/sources/github/repo.go
+++ b/pkg/sources/github/repo.go
@@ -258,6 +258,11 @@ func (s *Source) getDiffForFileInCommit(ctx context.Context, query commitQuery) 
 		return "", fmt.Errorf("error fetching commit %s: %w", query.sha, err)
 	}
 
+	if len(commit.Files) == 0 {
+		return "", fmt.Errorf("commit %s does not contain any files", query.sha)
+	}
+
+	res := new(strings.Builder)
 	// Only return the diff if the file is in the commit.
 	for _, file := range commit.Files {
 		if *file.Filename != query.filename {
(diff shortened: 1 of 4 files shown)


pkg/sources/github/repo.go
```diff
@@ -239,6 +239,41 @@ func (s *Source) processRepos(ctx context.Context, target string, listRepos repo
...
+		if *file.Filename != query.filename {
+			continue
+		}
+
+		if file.Patch == nil {
+			return "", fmt.Errorf("commit %s file %s does not have a diff", query.sha, query.filename)
+		}
+
+		return *file.Patch, nil
+	}
```
mcastorina, line 272: I think a commit could have multiple patches in the same file..

mcastorina (approved):
super cool - just saw one possible edge case

pkg/sources/github/repo.go
ahrav: Oh interesting, I didn't realize that. Do you know of an easy way go about accessing/finding those other patches. When i looked at the docs for the library we use to connect to Github the `CommitFile` struct looks like:
```
// CommitFile represents a file modified in a commit.
type CommitFile struct {
	SHA              *string `json:"sha,omitempty"`
	Filename         *string `json:"filename,omitempty"`
	Additions        *int    `json:"additions,omitempty"`
	Deletions        *int    `json:"deletions,omitempty"`
	Changes          *int    `json:"changes,omitempty"`
	Status           *string `json:"status,omitempty"`
	Patch            *string `json:"patch,omitempty"`
	BlobURL          *string `json:"blob_url,omitempty"`
	RawURL           *string `json:"raw_url,omitempty"`
	ContentsURL      *string `json:"contents_url,omitempty"`
	PreviousFilename *string `json:"previous_filename,omitempty"`
}
```
🤔

Issue #1806, opened 2023-09-22
Implement a installation script with CheckSum Validation.


varmakarthik12:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description
Looks like there is no version-specific home brew formula to install a specific version of Trufflehog. This makes a installation like `brew install trufflesecurity/trufflehog/trufflehog@3.56.0` fail. And also there are no built in checksum validations when a user is installing trufflehog. For folks who want to pin a specific version of trufflehog for obvious reasons. The only way now is to have a custom script to Download and unpack from the release. Having an install script natively with checksum validation can hugely improve the user experience & promote secure use of trufflehog. 

### Preferred Solution

There is an installation shell script that checks the platform & arch of the installation host. Download & unpack the respective binary. Does the checksum validations built in?

Installing trufflehog would be a simple ` curl <install-scrtip> | sh`

### Additional Context

The description says it all.

varmakarthik12:
I see this is merged now. Thanks @hibare  for the PR. Thanks, @mcastorina & @ahrav for reviewing this. Closing this as done.

PR #1808, opened 2023-09-22
Implement an installation script with CheckSum Validation

hibare:
### Description:
PR addresses issue https://github.com/trufflesecurity/trufflehog/issues/1806.

Added an installation script that downloads a specific version of Trufflehog from release page. If no version is specified, it defaults to the latest release, performing checksum validation before installation.
Additionally, the default installation directory is ./bin, but users can specify an alternative location using the -b flag, followed by the absolute path of their preferred destination.

### Checklist:

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,368 @@
+#!/bin/sh
+set -e
+
+usage() {
+  this=$1
+  cat <<EOF
+$this: download go binaries for trufflesecurity/trufflehog
+
+Usage: $this [-b] bindir [-d] [tag]
+  -b sets bindir or installation directory, Defaults to ./bin
+  -d turns on debug logging
+   [tag] is a tag from
+   https://github.com/trufflesecurity/trufflehog/releases
+   If tag is missing, then the latest will be used.
+
+EOF
+  exit 2
+}
+
+parse_args() {
+  #BINDIR is ./bin unless set be ENV
+  # over-ridden by flag below
+
+  BINDIR=${BINDIR:-./bin}
+  while getopts "b:dh?x" arg; do
+    case "$arg" in
+      b) BINDIR="$OPTARG" ;;
+      d) log_set_priority 10 ;;
+      h | \?) usage "$0" ;;
+      x) set -x ;;
+    esac
+  done
+  shift $((OPTIND - 1))
+  TAG=$1
+}
+
+# this function wraps all the destructive operations
+# if a curl|bash cuts off the end of the script due to
+# network, either nothing will happen or will syntax error
+# out preventing half-done work
+execute() {
+  tmpdir=$(mktemp -d)
+  log_debug "downloading files into ${tmpdir}"
+  http_download "${tmpdir}/${TARBALL}" "${TARBALL_URL}"
+  http_download "${tmpdir}/${CHECKSUM}" "${CHECKSUM_URL}"
+  hash_sha256_verify "${tmpdir}/${TARBALL}" "${tmpdir}/${CHECKSUM}"
+  srcdir="${tmpdir}"
+  (cd "${tmpdir}" && untar "${TARBALL}")
+  test ! -d "${BINDIR}" && install -d "${BINDIR}"
+
+  binexe=${BINARY}
+  if [ "$OS" = "windows" ]; then
+    binexe="${binexe}.exe"
+  fi
+  install "${srcdir}/${binexe}" "${BINDIR}/"
+  log_info "installed ${BINDIR}/${binexe}"
+
+  rm -rf "${tmpdir}"
+}
+
+get_binary() {
+  case "$PLATFORM" in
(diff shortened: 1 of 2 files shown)


ahrav:
Thanks a lot @hibare for this contribution. It looks really neat. @mcastorina would you mind taking a look at this as well. Your bash-fu is definitely a lot better than mine, which would make your review a lot more helpful 😄

scripts/install.sh
```diff
@@ -0,0 +1,368 @@
...
+# out preventing half-done work
+execute() {
+  tmpdir=$(mktemp -d)
+  log_debug "downloading files into ${tmpdir}"
+  http_download "${tmpdir}/${TARBALL}" "${TARBALL_URL}"
+  http_download "${tmpdir}/${CHECKSUM}" "${CHECKSUM_URL}"
+  hash_sha256_verify "${tmpdir}/${TARBALL}" "${tmpdir}/${CHECKSUM}"
+  srcdir="${tmpdir}"
+  (cd "${tmpdir}" && untar "${TARBALL}")
+  test ! -d "${BINDIR}" && install -d "${BINDIR}"
```
mcastorina, line 49: Is the `test` necessary here? It looks like `install -d dir` is non-destructive and exits 0 if the directory already exists.

Alternatively we could use `mkdir -p` which has the same behavior afaict.

scripts/install.sh
```diff
@@ -0,0 +1,368 @@
...
+  esac
+}
+
+tag_to_version() {
+  if [ -z "${TAG}" ]; then
+    log_info "checking GitHub for latest tag"
+  else
+    log_info "checking GitHub for tag '${TAG}'"
+  fi
+  REALTAG=$(github_release "$OWNER/$REPO" "${TAG}") && true
```
mcastorina, line 82: Is the intention of `&& true` to prevent exiting on error? If so, it should probably be `|| true`.

scripts/install.sh
```diff
@@ -0,0 +1,368 @@
+#!/bin/sh
+set -e
```
mcastorina, line 2: **Nit:** In general I'd prefer using long options when available in scripts that are saved to a file. It helps with readability, especially for those less practiced in the shell.

As an example, instead of `set -e` I would prefer `set -o errexit`.

scripts/install.sh
```diff
@@ -0,0 +1,368 @@
...
+
+  BINDIR=${BINDIR:-./bin}
+  while getopts "b:dh?x" arg; do
+    case "$arg" in
+      b) BINDIR="$OPTARG" ;;
+      d) log_set_priority 10 ;;
+      h | \?) usage "$0" ;;
+      x) set -x ;;
+    esac
+  done
```
mcastorina, line 32: My second request for long options :)

I know `getopt` supports it but not sure about `getopts`. What do you think about supporting long options in the installation script?

mcastorina (commented):
Overall looks good to me.

scripts/install.sh
hibare: Makes sense, changed to set -o errexit

scripts/install.sh
hibare: Removed directory existence test.

As for the choice between `install -d` and `mkdir -p`, I initially used install -d in the installation script due to personal preference as its in an installation script and has a scope to set different permissions than the default ones in the single command.

Please let me know if this approach is acceptable, or if you prefer using mkdir -p. I'm open to making the change if needed.

scripts/install.sh
hibare: Changed it. Thank you!!

scripts/install.sh
hibare: getops does not support long options, its limited to single character options only. There are workaround for getopts to support long options but its not error proof requiring frequent maintenance. Other option is to switch over to getopt. I'm not sure how frequently long options are requested. 

Please let me know if you think long options are necessary. Happy to change it.

mcastorina: Hm, I think in this case the short options are fine. The install script will most often be used in a copy-paste situation and there aren't too many of them.

mcastorina (approved):

mcastorina:
Tested it out and works great!

```
/tmp » mkdir bin
/tmp » curl -sSfL https://raw.githubusercontent.com/trufflesecurity/trufflehog/e5deec77464eb76d731b0c21ac173074977ca7a8/scripts/install.sh | sh -s -- -b ./bin
trufflesecurity/trufflehog info checking GitHub for latest tag
trufflesecurity/trufflehog info found version: 3.57.0 for v3.57.0/darwin/arm64
trufflesecurity/trufflehog info installed ./bin/trufflehog
/tmp » ls -lh bin
total 108M
-rwxr-xr-x 1 miccah wheel 108M Sep 25 12:43 trufflehog
/tmp » ./bin/trufflehog --version
trufflehog 3.57.0
```

ChronosMasterOfAllTime:
#1809

PR #1816, opened 2023-09-25
added PR and Issue body scanning

joeleonjr:
### Description:
TruffleHog's `github` module currently has `--pr-comments` and `--issue-comments` flags. Importantly, those flags only scan the comments and not the initial post to the `pr` or `issue`. Expected user behavior when TruffleHog scans PRs and Issues is to include both the initial PR/Issue post and all related comments. Not scanning the initial post will lead to incomplete results. 

This PR changes the CLI args from:
`--pr-comments --issue-comments` to `--prs --issues (--exclude-comments)`.

The github `pkg` logic has been updated so that the `--pr` flag scans all PR posts and related comments and the `--issues` flag scans all Issue posts and related comments. If a user is rate-limit conscious and does not want to scan all comments they can use the `--exclude-comments` flag. 

### Checklist:


main.go
```diff
@@ -73,20 +73,21 @@ var (
...
-	githubIncludeRepos      = githubScan.Flag("include-repos", `Repositories to include in an org scan. This can also be a glob pattern. You can repeat this flag. Must use Github repo full name. Example: "trufflesecurity/trufflehog", "trufflesecurity/t*"`).Strings()
-	githubExcludeRepos      = githubScan.Flag("exclude-repos", `Repositories to exclude in an org scan. This can also be a glob pattern. You can repeat this flag. Must use Github repo full name. Example: "trufflesecurity/driftwood", "trufflesecurity/d*"`).Strings()
-	githubScanIncludePaths  = githubScan.Flag("include-paths", "Path to file with newline separated regexes for files to include in scan.").Short('i').String()
-	githubScanExcludePaths  = githubScan.Flag("exclude-paths", "Path to file with newline separated regexes for files to exclude in scan.").Short('x').String()
-	githubScanIssueComments = githubScan.Flag("issue-comments", "Include issue comments in scan.").Bool()
-	githubScanPRComments    = githubScan.Flag("pr-comments", "Include pull request comments in scan.").Bool()
-	githubScanGistComments  = githubScan.Flag("gist-comments", "Include gist comments in scan.").Bool()
```
mcastorina, line 89: The CLI flags and arguments are considered to be our public facing API, so unfortunately we can't change them so easily.

We'll need to keep the existing flags and their functionality and optionally add new flags that have some overlapping but preferred usage.

mcastorina (changes requested):
I'm requesting changes for the CLI flags, but I will continue reviewing.

main.go
joeleonjr: Returned flags back to their original state. Removed the `exclude-comments` flag based on Dylan's comments. Added two new ones `prs` and `issues`, which we can highlight in the documentation, but that don't break the `issue-comments` and `pr-comments` flags.

joeleonjr: Also, I added in logic so that if a user runs `issue-comments` and `issues`, it'll only review the comments once, so as to not waste API queries. Same for PRs. This allows us to keep all 4 flags without any downside to the user.

pkg/sources/github/github.go
```diff
@@ -1141,16 +1147,75 @@ func (s *Source) processRepoComments(ctx context.Context, repoPath string, trimm
...
+	if s.includePRs {
+		if err := s.processPRs(ctx, repoInfo, chunksChan); err != nil {
+			return err
+		}
+		if !s.includePRComments {
+			if err := s.processPRComments(ctx, repoInfo, chunksChan); err != nil {
+				return err
+			}
+		}
 	}
```
mcastorina, line 1178: **Nit (optional):** I like that we're not repeating work here, but it's easier for me to read the code if its laid out like so:

```go
if s.includeIssues || s.includeIssueComments {
    // processIssueComments
}

if s.includeIssues {
    // processIssues
}

if s.includePRs || s.includePRComments {
    // processPRComments
}

if s.includePRs {
    // processPRs
}
```

Another idea - if we make `includeIssues` only for scanning issues (and not comments), we could make the CLI flag configure both for the same overall functionality. This would make the `Source` slightly more flexible in case we ever want to scan only one or the other.. which is probably an unlikely configuration? That's why this is an optional nit comment 😅

pkg/sources/github/github.go
```diff
@@ -1221,6 +1322,45 @@ func (s *Source) processPRComments(ctx context.Context, info repoInfo, chunksCha
 	return nil
 }
 
+func (s *Source) chunkIssues(ctx context.Context, repo, repoPath string, issues []*github.Issue, chunksChan chan *sources.Chunk) error {
+	for _, issue := range issues {
+
+		// Skip pull requests since covered by processPRs.
+		if issue.IsPullRequest() {
+			continue
+		}
```
mcastorina, line 1331: Is this something we can filter out in the `Issues.ListByRepo` call?

mcastorina (approved):
lgtm! If you're up for it, it'd be nice to have some tests (in another PR).

pkg/sources/github/github.go
joeleonjr: Surprisingly, it doesn't seem possible. Here's the official API docs:

https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues

My guess is GH's GraphQL endpoint would let us do this, but I'm not sure.

pkg/sources/github/github.go
joeleonjr: Updated codebase for the first suggestion. 

For the second suggestion, would users have to add two flags to the CLI? `--issues --issue-comments` ? The general idea (at least in my mind) is to encourage users to only use the `--issues` flag so that Issue Body text and Issue Comments are all scanned. As it stands in production right now, I think users would assume the initial Issue body text is being scanned.

PR #1819, opened 2023-09-26
Separate gitlab detectors

0x1:
### Description:
Update GitlabV1 and GitlabV2 detectors to handle a unique version of secret
Update Gitlab to tri-state verification
### Checklist:


pkg/detectors/gitlab/gitlab.go
```diff
@@ -14,21 +14,20 @@ import (
...
 var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+var _ detectors.Versioner = (*Scanner)(nil)
 
 func (Scanner) Version() int            { return 1 }
 func (Scanner) DefaultEndpoint() string { return "https://gitlab.com" }
 
 var (
 	defaultClient = common.SaneHttpClient()
-	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"gitlab"}) + `\b((?:glpat|)[a-zA-Z0-9\-=_]{20,22})\b`)
+	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"gitlab"}) + `\b[^glpat-]([a-zA-Z0-9\-=_]{20,22})\b`)
```
rosecodym, line 30: this regex looks really suspicious. `[^glpat-]` means "exactly one character that is not a `g`, `l`, `p`, `a`, `t`, or `-`", which i'm guessing is not what you want. if you want to exclude a prefix, one solution is use something called a "negative lookbehind" but i've never done that before so you'd have to become the expert yourself. (it might also be possible to put the prefix in a capture group, make it optional, and exclude matches that contain it later, but i don't know if it's possible to do this without additionally matching the un-prefixed key, which would defeat the purpose.)

pkg/detectors/gitlab/gitlab_v1_test.go
```diff
@@ -6,17 +6,18 @@ package gitlab
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-func TestGitlab_FromChunk(t *testing.T) {
+func TestGitlab_FromChunk_WithV2Secrets(t *testing.T) {
```
rosecodym, line 20: i _think_ you meant to rename the tests in the _other_ v1 test file

pkg/detectors/gitlab/gitlab_v2_test.go
```diff
@@ -6,13 +9,14 @@ import (
...
-	"github.com/kylelemons/godebug/pretty"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-func TestGitlabv2_FromChunk(t *testing.T) {
+// This test ensures gitlab v1 detector does not work on gitlab v2 secrets
```
rosecodym, line 18: ❤️

rosecodym (changes requested):

pkg/detectors/gitlabV2/gitlab_v1_test.go
```diff
@@ -1,64 +1,72 @@
 //go:build detectors
 // +build detectors
 
-package gitlabv2
+package gitlabV2
```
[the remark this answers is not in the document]
0x1: hmmm... now that you mention it. go's [docs](https://go.dev/blog/package-names) do say to use mixedCaps :(

pkg/detectors/gitlab/gitlab_v1_test.go
0x1: 👀

pkg/detectors/gitlab/gitlab.go
0x1: hmm... unfortunately lookbehinds aren't supported in Go, but it probably makes more sense to explicitly filter out the results with `glpat`.

rosecodym (approved):
ok well i still don't quite understand why the v1 detector doesn't pick up v2 secrets but the tests pass so that's more of a homework project for me than you. nice work!

PR #1820, opened 2023-09-26
[chore] add figmav2 to defaults

0x1:
### Description:
add figmav2 to defaults

### Checklist:


mcastorina (commented):
`figmapersonalaccesstokenV2` should be replaced with `figmapersonalaccesstoken_v2` since the package name should match the directory it's in. Typically Go uses underscores for packages instead of camel case since some filesystems are case insensitive (my theory).

0x1:
> `figmapersonalaccesstokenV2` should be replaced with `figmapersonalaccesstoken_v2` since the package name should match the directory it's in. Typically Go uses underscores for packages instead of camel case since some filesystems are case insensitive (my theory).

agreed w/ package name should match directory name. but the [docs](https://go.dev/blog/package-names#Package-names:~:text=They%20are%20lower%20case%2C%20with%20no%20under_scores%20or%20mixedCaps.) say package names should be lowercase and no mixedCaps or underscores. thoughts?

0x1:
> > `figmapersonalaccesstokenV2` should be replaced with `figmapersonalaccesstoken_v2` since the package name should match the directory it's in. Typically Go uses underscores for packages instead of camel case since some filesystems are case insensitive (my theory).
> 
> agreed w/ package name should match directory name. but the [docs](https://go.dev/blog/package-names#Package-names:~:text=They%20are%20lower%20case%2C%20with%20no%20under_scores%20or%20mixedCaps.) say package names should be lowercase and no mixedCaps or underscores. thoughts?

apparently the [right way](https://go.dev/blog/v2-go-modules) to do this is by adding the version to the import path, but keeping the original package name. it's actually pretty clean.

ahrav (approved):
very nice!

mcastorina (approved):

PR #1822, opened 2023-09-26
Update Adding_Detectors_external.md

zricethezav:
### Description:
Introduces a dedicated testing section to the "Adding a New Secret Detector" section. I _think_ this makes a little more sense but I'm open to suggestions

### Checklist:


hack/docs/Adding_Detectors_external.md
```diff
@@ -116,8 +95,49 @@ For example, consider a hypothetical authentication endpoint that returns `200 O
...
+```
+
+Now that a `.env` file is present, the test file can load secrets locally.
+
+Next, Update the tests with these test cases at minimum:
+1. Found and verified (using a credential loaded from the .env file)
+2. Found and unverified (determinately, i.e. the secret is invalid)
+3. Found and unverified (indeterminately due to timeout)
+4. Found and unverified (indeterminately due to an unexpected API response)
+5. Not found
```
rosecodym, line 113: Do you think it's worth calling out that the generator will create these cases by default, and "adding" them really just means making any necessary changes, of which there might not be any?

zricethezav: @rosecodym great suggestion! I think that's worth calling out for sure

rosecodym (approved):
do you want to update the internal version of this doc in parallel? i'm not really sure if there are rules about keeping them synchronized

PR #1824, opened 2023-09-27
[bug] - correctly check err

ahrav:
### Description:
We need to reuse the same err initially declared if we want to cleanup after we try to clone.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -312,12 +312,19 @@ func GitURLParse(gitURL string) (*url.URL, error) {
 	return parsedURL, nil
 }
 
-// CloneRepo clones a given Git repository and returns its local path along with a git.Repository object for further operations.
-// The function utilizes a nested function (closure) to centralize error handling.
-// The outer function sets up a deferred call to CleanOnError, which captures the address of a single 'err' variable.
-// The inner function (closure) is responsible for the core logic and returns an error that sets the outer 'err' variable.
-// This ensures that CleanOnError always has the most up-to-date error information for conditional cleanup.
-func CloneRepo(ctx context.Context, userInfo *url.Userinfo, gitUrl string, args ...string) (string, *git.Repository, error) {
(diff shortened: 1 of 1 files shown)


rosecodym (commented):
to make sure i understand this: would it be logically equivalent to write this as
```
func CloneRepo(...) (string, *git.Repository, error) {
    // setup
    path, repo, err := newExtractedFunction(url)
    if err != nil {
        CleanOnError(err, path)
        return "", nil, err
    }
    return path, repo, nil
}
```
that is, you don't _need_ the nested function or the `defer` - you just felt this implementation was the cleanest?

pkg/sources/git/git.go
```diff
@@ -312,63 +312,80 @@ func GitURLParse(gitURL string) (*url.URL, error) {
...
+			return fmt.Errorf("could not open cloned repo: %w", err)
+		}
+		logger.V(1).Info("successfully cloned repo")
+
+		return nil
+	}()
 	if err != nil {
-		return "", nil, fmt.Errorf("could not open cloned repo: %w", err)
+		return "", nil, err
 	}
```
mcastorina, line 387: Instead of using a `defer`, we could call `CleanOnError` here. The only considerable difference is it wouldn't run if there was a `panic` in the closure.

mcastorina (approved):
good find!

ahrav:
> to make sure i understand this: would it be logically equivalent to write this as
> 
> ```
> func CloneRepo(...) (string, *git.Repository, error) {
>     // setup
>     path, repo, err := newExtractedFunction(url)
>     if err != nil {
>         CleanOnError(err, path)
>         return "", nil, err
>     }
>     return path, repo, nil
> }
> ```
> 
> that is, you don't _need_ the nested function or the `defer` - you just felt this implementation was the cleanest?

Yep identical. Definitely not the cleanest, just for some reason it kinda made sense at the time. Will use an extra fxn instead.

PR #1830, opened 2023-09-27
add tristate verification to twitch

0x1:
### Description:
add tristate verification to twitch

### Checklist:

### DIFF
--- a/pkg/detectors/twitch/twitch.go
+++ b/pkg/detectors/twitch/twitch.go
@@ -21,9 +21,10 @@ type Scanner struct {
 // Ensure the Scanner satisfies the interface at compile time
 var _ detectors.Detector = (*Scanner)(nil)
 
+const verifyURL = "https://id.twitch.tv/oauth2/token"
+
 var (
 	defaultClient = common.SaneHttpClient()
-	verifyURL     = "https://id.twitch.tv/oauth2/token"
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"twitch"}) + `\b([0-9a-z]{30})\b`)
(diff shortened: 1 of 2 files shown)


pkg/detectors/twitch/twitch_test.go
```diff
@@ -47,14 +49,87 @@ func TestTwitch_FromChunk(t *testing.T) {
...
 					Verified:     true,
 				},
+				{
+					DetectorType: detectorspb.DetectorType_Twitch,
+					Verified:     false,
+				},
+				{
+					DetectorType: detectorspb.DetectorType_Twitch,
+					Verified:     false,
+				},
```
rosecodym, line 71: is the order of these results consistent? i think that the test will flake if not.

0x1: yes it should always consistent bc the data string doesn't change.

pkg/detectors/twitch/twitch.go
```diff
@@ -58,37 +62,59 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Add("Content-Length", strconv.Itoa(len(data.Encode())))
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+
+	defer res.Body.Close()
+	if res.StatusCode == http.StatusOK {
```
ahrav, line 109: nit (optional): You could switchify this and the default case handles the final return.
```
switch res.StatusCode {
case http.StatusOK:
	return true, nil
case http.StatusBadRequest, http.StatusForbidden:
	return false, nil
default:
	return false, fmt.Errorf("unexpected http response status %d", res.StatusCode)
```

ahrav (approved):

pkg/detectors/twitch/twitch.go
```diff
@@ -13,13 +14,16 @@ import (
...
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
+	verifyURL     = "https://id.twitch.tv/oauth2/token"
```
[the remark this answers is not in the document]
0x1: u right, prob should be const

Issue #1476, opened 2023-07-11
Ignore scanning GitHub gists with a flag


CameronLonsdale:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When scanning with --include-members, it will include scanning gists. It would be really nice to have the option to exclude gists as this could see your rate limit exceed the 5000 requests per hour limit which GitHub enforces.

## Problem to be Addressed

Ignore scanning Gists to avoid GitHub rate limits.

## Description of the Preferred Solution

`--ignore-gists` or `--include-gists` option

### References

* #0000

zricethezav:
Hi @CameronLonsdale, thanks for opening this issue! I can see this as being useful. Can't make any promises when we'll get to this but we're happy review community PRs implementing this enhancement. I think this would be a good "first issue" for anyone out there who wants to contribute.

pi1814:
Hi All,
I would like to get involved in the truffleHog community. Can I take this up?
I will keep you posted regarding the progress and blockers if any.

CameronLonsdale:
I think there's a workaround for this using the `--include-repos` and `--exclude-repos` flags. Since gist URLs are not scoped to your user, if you run `trufflehog github --org CameronLonsdale --exclude-repos "CameronLonsdale/*"` this should exclude all my repositories and only scan my public gists.

kashifkhan0771:
Hi @CameronLonsdale would you be okay with closing this issue since there’s a workable solution using the `--include-repo` and `--exclude-repo` flags? Or do you think it’s still worth implementing a separate --ignore-gists flag?

CameronLonsdale:
Happy to close with the workaround, the only benefit to an --ignore-gists flag is that it might save API tokens when querying github

kashifkhan0771:
Okay, I’ll check if someone on our team has bandwidth to take this on.

PR #1853, opened 2023-10-02
Add new argument to ignore/include github gist

za:
### Description:
To address issue #1476 

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -87,6 +87,8 @@ var (
 	githubScanIssueComments = githubScan.Flag("issue-comments", "Include issue descriptions and comments in scan.").Bool()
 	githubScanPRComments    = githubScan.Flag("pr-comments", "Include pull request descriptions and comments in scan.").Bool()
 	githubScanGistComments  = githubScan.Flag("gist-comments", "Include gist comments in scan.").Bool()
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -87,6 +87,8 @@ var (
 	githubScanIssueComments = githubScan.Flag("issue-comments", "Include issue descriptions and comments in scan.").Bool()
 	githubScanPRComments    = githubScan.Flag("pr-comments", "Include pull request descriptions and comments in scan.").Bool()
 	githubScanGistComments  = githubScan.Flag("gist-comments", "Include gist comments in scan.").Bool()
+	githubScanIgnoreGist    = githubScan.Flag("ignore-gist", "Ignore gist comments in scan.").Bool()
+	githubScanIncludeGist   = githubScan.Flag("include-gist", "Include gist comments in scan.").Bool()
```
rgmz, line 91: I think it would be better to have a single flag `--include-gists`, which would fit in with the existing convention of `--include-forks` and `--include-members.

za: Addressed in the following commit

zricethezav:
This is an incomplete PR and doesn't implement logic that ignores/includes github gists. Closing for now. Feel free to reopen when implemented fully.

PR #1858, opened 2023-10-03
Add generic glob filter

mcastorina:
### Description:
Adds a generic glob filter for arbitrary include and ignore lists.

### Checklist:

### DIFF
--- a/pkg/common/glob/glob.go
+++ b/pkg/common/glob/glob.go
@@ -12,9 +12,6 @@ import (
 type Filter struct {
 	exclude []glob.Glob
 	include []glob.Glob
-	// If an object is not found in either of the filters, return this
-	// value in ShouldInclude.
-	notFoundShouldIncludeBehavior bool
 }
 
 type globFilterOpt func(*Filter) error
@@ -43,26 +40,6 @@ func WithIncludeGlobs(includes ...string) globFilterOpt {
 			}
 			f.include = append(f.include, g)
 		}
-		return nil
-	}
-}
-
-// WithDefaultDeny configures the filter to deny objects that are not found
-// in either exclude or include filter (in the ambiguous case where both are
-// configured).
-func WithDefaultDeny() globFilterOpt {
-	return func(f *Filter) error {
-		f.notFoundShouldIncludeBehavior = false
-		return nil
-	}
-}
-
-// WithDefaultAllow configures the filter to allow objects that are not found
-// in either exclude or include filter (in the ambiguous case where both are
-// configured).
-func WithDefaultAllow() globFilterOpt {
-	return func(f *Filter) error {
-		f.notFoundShouldIncludeBehavior = true
 		return nil
 	}
 }
@@ -96,7 +73,7 @@ func (f *Filter) ShouldInclude(object string) bool {
 			return ok
 		}
 		// Ambiguous case.
-		return f.notFoundShouldIncludeBehavior
+		return false
 	}
 }
 
--- a/pkg/common/glob/glob.go
+++ b/pkg/common/glob/glob.go
@@ -22,7 +22,7 @@ func WithExcludeGlobs(excludes ...string) globFilterOpt {
 		for _, exclude := range excludes {
 			g, err := glob.Compile(exclude)
 			if err != nil {
-				return err
+				return fmt.Errorf("invalid exclude glob %q: %w", exclude, err)
 			}
 			f.exclude = append(f.exclude, g)
 		}
@@ -36,7 +36,7 @@ func WithIncludeGlobs(includes ...string) globFilterOpt {
 		for _, include := range includes {
 			g, err := glob.Compile(include)
 			if err != nil {
-				return err
+				return fmt.Errorf("invalid include glob %q: %w", include, err)
 			}
 			f.include = append(f.include, g)
 		}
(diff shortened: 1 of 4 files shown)


pkg/common/glob/glob.go
```diff
@@ -0,0 +1,123 @@
...
+
+type globFilterOpt func(*Filter) error
+
+// WithExcludeGlobs adds exclude globs to the filter.
+func WithExcludeGlobs(excludes ...string) globFilterOpt {
+	return func(f *Filter) error {
+		for _, exclude := range excludes {
+			g, err := glob.Compile(exclude)
+			if err != nil {
+				return err
```
rosecodym, line 25: Does this error include the pattern that failed? When I was refactoring the globbing in the Gitlab source I think I had to add that explicitly to the error output.

pkg/common/glob/glob.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, err
+		}
+	}
+	return filter, nil
+}
+
+// Pass returns whether the object is not in the include list or in the exclude
+// list.
+func (f *Filter) Pass(object string) bool {
+	if f == nil {
```
rosecodym, line 61: i didn't even know that go let you call methods on a nil receiver!

pkg/common/glob/glob.go
```diff
@@ -0,0 +1,123 @@
...
+		return f.passInclude(object)
+	} else {
+		if pass, err := f.passExcludeInclude(object); err == nil {
+			return pass
+		}
+		// Ambiguous case. Should we skip? Should we not skip?
+		// Let's skip since the user indicated they want *some* form of
+		// filtering.
+		// TODO: Maybe this could be configurable.
+		return true
```
rosecodym, line 79: Would adding a default case to the filter resolve this easily? It seems like it would take little effort but provide a lot more clarity. (Maybe make it three-state, where the third state is "return an error?")

pkg/common/glob/glob.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// passExclude checks for explicitly excluded paths. This should only be called
+// when the include list is empty.
```
rosecodym, line 84: implementing all this via four very-similar-yet-occasionally-inverted blocks, some of which have sneaky preconditions, dispatched via a four-clause if statement, is triggering my spidey sense. does the complexity come from differentiating between glob sets that are empty and glob sets that are non-empty but reject the candidate object?

rosecodym (commented):
what motivated this?

pkg/common/glob/glob.go
mcastorina: Yeah, it made the most sense in my head to differentiate the use cases.

* If I provide only an include list, I expect only matches to be scanned and everything else to be skipped.
* Likewise, if I only provide an exclude list, I expect the opposite behavior.
* If I provide both.. I'm not sure what to expect? Maybe an example would be overlapping globs. My expectation for the below is to scan everything under `/home/miccah/` except the tmp dir.
  * Include `/home/miccah/**`
  * Exclude `/home/miccah/tmp/**`

To maybe appease your spidey senses, the only exposed method is `Pass`. The other functions are not accessible outside the package and cannot be called.

pkg/common/glob/glob.go
mcastorina: Yes! I usually think of method receivers as syntax sugar for `func Pass(f *Filter, object string) bool`. I think it's possible because `nil` is typed in Go.

It's not often I rely on that functionality, but I thought it would be a nice touch here.

mcastorina:
> what motivated this?

I'm adding include / ignore filters to another source and decided to try and standardize the behavior in single package. If there's already something like this I can use, even better!

pkg/common/glob/glob.go
rosecodym: oh no i mean something like
```
type Filter struct {
	exclude []glob.Glob
	include []glob.Glob
        conflictBehavior bool
}
```
or
```
type Filter struct {
	exclude []glob.Glob
	include []glob.Glob
        conflictBehavior SomeNewTriStateType
}
```

pkg/common/glob/glob.go
rosecodym: yeah, my brain is just trying to find a more compact mathematical representation of the logic. i _think_ what i want is this, which isn't actually possible and wouldn't perform optimally if it were:
1. adjust empty sets
  * if only the include set is empty, add a universal acceptance glob to it
  * if only the exclude set is empty, add a universal rejection glob to it <- this is the impossible part
2. test the candidate against both sets
  * if the exclude set accepts, reject the candidate
  * otherwise, iff the include set accepts, accept the candidate

this makes step (2) simpler, but at the cost of running unnecessary globs, and needing to use an undefinable "reject everything" glob. since what i want is impossible and slow, what you've got is probably best :)

pkg/common/glob/glob.go
rosecodym: fun trivia fact: this capability was in .net framework 1 (way back in the day) and isn't precluded by the implementation but the designers intentionally forbade it because it confused too many people

rosecodym:
oh, also, is "no filters at all means reject everything" what we do now? it surprised me.

mcastorina:
> oh, also, is "no filters at all means reject everything" what we do now? it surprised me.

"no filters at all" should accept everything (i.e. `Pass` always returns `false`). Should we invert `Pass` so it's easier to reason about? I did that in the tests.. open to name suggestions :)

pkg/common/glob/glob.go
mcastorina: Oh I like it! If we do the tristate type, I think we would also have to return `(bool, error)` since that would be the only way to return an error on conflict.

rosecodym:
"Pass" means _reject_? Yeah, I didn't catch that and I find it really surprising. I expected the opposite.

PR #1860, opened 2023-10-04
[bug] - use Repositories field from conn

ahrav:
### Description:
We should be using the `Repositories` field from the Github source proto for the include portion of the include/ignore filtering.

### Checklist:


0x1 (approved):
👍🏽

PR #1866, opened 2023-10-05
Detector-Competition-Fix: fix notion.so false negative verification

lc:
### Description:
This pull request fixes the notion detector's verification process which likely leads to inaccurate results.

Notion allows developers to create API tokens with fine-tuned permissions:

CleanShot 2023-10-05 at 07 38 30

The original verification process attempts to hit `/v1/users` but expects a `200 OK` status code. 

However, if the key is valid, but does not have permission to list users, Notion responds with `403 Forbidden`:

```json
{"object":"error","status":403,"code":"restricted_resource","message":"Insufficient permissions for this endpoint."}
``` 

This indicates the secret **is valid**, but it must be used on another endpoint.

### Checklist:


zricethezav (approved):
@lc great improvement!

PR #1862, opened 2023-10-05
Detector-Competition-New: add ramp.com client id & secret detector

lc:
### Description:
This pull request implements detection for ramp.com's `client_id`  and `client_secret` fields.

I followed the [API documentation](https://docs.ramp.com/developer-api/v1/authorization) from Ramp:

CleanShot 2023-10-04 at 22 57 52

I've verified them using my own credentials.

Ramp allows applications to fine-tune the scope they allow, but it appears that no validation is done on the `scope` **after** issuance (any further API requests with the newly generated token)

### Checklist:
* [x] Tests passing (`make test-community`)?
* [x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #1868, opened 2023-10-05
Detector-Competition-Fix: fix foursquare client id & secret formatting

lc:
### Description:
This fixes the client id & secret formatting for the FourSquare detector.

`client_id:client_secret` is easier to read compared to `client_idclient_secret`

### Checklist:


lc:
This likely isn't enough to qualify for points, but still an improvement to be made

zricethezav:
@lc I should've added more context to this comment: https://github.com/trufflesecurity/trufflehog/pull/1862#issuecomment-1748002692.

Either is fine, but when creating a _new_ detector, semi colons are preferred. Problems start to pop up when we change `RawV2` for _existing_ detectors.

> This likely isn't enough to qualify for points, but still an improvement to be made

: / you're right, and because this is _changing_ the format I'm gonna close the PR since it would cause issues

PR #1870, opened 2023-10-05
Detector-Competition-Feat: Add Klaviyo API Secret Detector

lc:
### Description:
This pull request implements the detection and verification of Klaviyo API keys. 

These private API keys may have fine-tuned permissions: 

CleanShot 2023-10-05 at 12 42 39

Verification is done by requesting `/api/profiles` and checking the response.

If the key is invalid, Klaviyo returns `401 Unauthorized` with `"code":"authentication_failed",`

If the key is valid, it either responds with:
* `200 OK` (should the scope have `profiles:read`)
* `403 Forbidden` & `"code":"permission_denied"` – indicating the key is valid, but the user must to enumerate which API endpoint it has access to.

The Klaviyo documentation can be found [here](https://developers.klaviyo.com/en/docs/get_started). 

### Checklist:


rosecodym (commented):
Thank you for this contribution! I see you've got some semi-complicated 403 logic and there are some unhandled cases that should be handled. In particular, it doesn't look like you're explicitly handling these cases with a 403 response status:
* The error code is not `permission_denied`
* There are no error codes returned (not expected, but maybe the server is malfunctioning)
* The response body is not parseable as JSON (maybe the server is malfunctioning)

The last case should result in an error being stored in `s1.VerificationError` so that we know that the failure to verify doesn't necessarily have anything to do with the candidate credential. I'm not sure what the correct behavior is for the other cases - do they mean that the credentials are invalid or that something else has gone wrong? If the latter, an error should also be stored in `s1.VerificationError`.

lc:
Hey @rosecodym, 

Thanks for the review, I've added the correct error handling (I believe).

From Klaviyo:

403 | Forbidden | The request contains valid authentication information, but does not have permissions to perform the specified action.See API key scopes for more information.
-- | -- | --


So if we:
1. Get a 403
2. Can decode an error into the `response` struct JSON 
3. And the response struct contains an error

Then the key is verified.

rosecodym (approved):
The new checks look great, thanks!

PR #1873, opened 2023-10-06
[chore] Add SourceUnitEnumChunker filesystem tests

mcastorina:
### Description:
Add tests for the filesystem enumeration and `SourceUnit` chunking methods.

### Checklist:

### DIFF
--- a/pkg/sources/filesystem/filesystem_test.go
+++ b/pkg/sources/filesystem/filesystem_test.go
@@ -143,6 +143,9 @@ func TestEnumerate(t *testing.T) {
 	for _, unit := range reporter.Units {
 		assert.Contains(t, units, unit.SourceUnitID())
 	}
+	for _, unit := range units {
+		assert.Contains(t, reporter.Units, sources.CommonSourceUnit{ID: unit})
+	}
 }
 
 func TestChunkUnit(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/sourcestest/sourcestest.go
```diff
@@ -0,0 +1,61 @@
+package sourcestest
```
mcastorina, line 1: I wasn't sure where to put these so I made a `sourcestest` package akin to [iotest](https://pkg.go.dev/testing/iotest). Open to suggestions!

pkg/sources/filesystem/filesystem_test.go
```diff
@@ -120,3 +113,192 @@ func TestScanFile(t *testing.T) {
...
+
+	reporter := sourcestest.TestReporter{}
+	err = s.Enumerate(ctx, &reporter)
+	assert.NoError(t, err)
+
+	assert.Equal(t, len(units), len(reporter.Units))
+	assert.Equal(t, 0, len(reporter.UnitErrs))
+	for _, unit := range reporter.Units {
+		assert.Contains(t, units, unit.SourceUnitID())
+	}
```
rosecodym, line 145: Do you think it's worth checking the opposite of this as well? (I'm presuming that there's no handy collection comparison function available.)

rosecodym (approved):
lgtm but it's probably worth getting more votes for your module organization question

pkg/sources/filesystem/filesystem_test.go
mcastorina: Yeah, since one is a `[]string` and the other is a `[]SourceUnit`, it was easier to check this way. The longer more correct way would be to convert the `[]SourceUnit` to `[]string` and check that they match.. I was being lazy.

pkg/sourcestest/sourcestest.go
dustin-decker: Oh I have a test implementation https://github.com/trufflesecurity/trufflehog/pull/1877 but open to this too.

PR #1875, opened 2023-10-06
Add an option to filter unverified results using shannon entropy

dustin-decker:
Adds an option to filter out unverified results using shannon entropy.

### Checklist:

### DIFF
--- a/pkg/detectors/falsepositives.go
+++ b/pkg/detectors/falsepositives.go
@@ -92,20 +92,18 @@ func bytesToCleanWordList(data []byte) []string {
 	return words
 }
 
-// DefaultMinEntropy setting inspired by the original TruffleHog's default value for strings
-const DefaultMinEntropy = 3.0
-
 func StringShannonEntropy(input string) float64 {
 	chars := make(map[rune]float64)
+	inverseTotal := 1 / float64(len(input)) // precompute the inverse
 
 	for _, char := range input {
 		chars[char]++
 	}
 
 	entropy := 0.0
-	for _, value := range chars {
-		value /= float64(len(input))
-		entropy += value * math.Log2(value)
+	for _, count := range chars {
+		probability := count * inverseTotal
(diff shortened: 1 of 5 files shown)


pkg/detectors/falsepositives.go
```diff
@@ -90,3 +91,41 @@ func bytesToCleanWordList(data []byte) []string {
 	}
 	return words
 }
+
+// DefaultMinEntropy setting inspired by the original TruffleHog's default value for strings
+const DefaultMinEntropy = 3.0
+
+func StringShannonEntropy(input string) float64 {
```
ahrav, line 98: suggestion: If pre-compute the inverse of the length we can use multiplication instead of division within the entropy calculation.
ex:
```
func StringShannonEntropy(input string) float64 {
	chars := make(map[rune]float64)
	inverseTotal := 1 / float64(len(input)) // precompute the inverse

	for _, char := range input {
		chars[char]++
	}

	entropy := 0.0
	for _, count := range chars {
		probability := count * inverseTotal
		entropy += probability * math.Log2(probability)
	}

	return -entropy
}
```

This gives us a nice ~40% perf gain:
Screenshot 2023-10-08 at 11 08 41 AM

ahrav (approved):

pkg/detectors/falsepositives.go
```diff
@@ -90,3 +91,41 @@ func bytesToCleanWordList(data []byte) []string {
 	}
 	return words
 }
+
+// DefaultMinEntropy setting inspired by the original TruffleHog's default value for strings
+const DefaultMinEntropy = 3.0
```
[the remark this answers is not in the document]
dustin-decker: Not anymore, I'll remove it.

pkg/detectors/falsepositives.go
dustin-decker: Oh nice. Added test and changed it out.

PR #1878, opened 2023-10-09
Add temp directory management

codevbus:
### Description:
Adds logic to create unique temp directories for repository cloning. Detects unused/orphaned temp directories and deletes them on a new scan run.

### Checklist:


codevbus:
Temp directories are created here: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L332

We want to change the tempdir name to something like: `trufflehog-<pid>-<random>`

Assuming you initiate a scan of a "git" repo, the call is passed like this:

[run](https://github.com/trufflesecurity/trufflehog/blob/main/main.go#L215) -> [case gitScan.FullCommand():](https://github.com/trufflesecurity/trufflehog/blob/main/main.go#L383) -> [ScanGit](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/git.go#L17) -> [Run returns](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L101-L102) [asyncRun](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L106)  -> [run](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L218) -> [runWithoutUnits](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L243) -> [Chunks](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L158) -> [scanRepos](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L158) -> [CloneRepo (returned from a parent function depending on the type of git auth)](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L425-L440)

I've got a couple of implementation questions:

- By default there are at least two scanner PIDs running for a given scan job. Depending on concurrency values, it may be more. Is there a rule/best-practice for which PID should actually be interpolated into the directory name?

- What's the cleanest way to pass the PID value through to the tempdir creation? Attach it to the `ctx` object?

main.go
```diff
@@ -529,6 +574,12 @@ func run(state overseer.State) {
 		}
 	}
 
+	pids, _ := getScannerPIDs()
```
codevbus, line 577: Is this the right timing to get the PID values and instantiate a cleanup?

dustin-decker:
> Temp directories are created here: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L332
> 
> 
> 
> We want to change the tempdir name to something like: `trufflehog-<pid>-<random>`
> 
> 
> 
> Assuming you initiate a scan of a "git" repo, the call is passed like this:
> 
> 
> 
> [run](https://github.com/trufflesecurity/trufflehog/blob/main/main.go#L215) -> [case gitScan.FullCommand():](https://github.com/trufflesecurity/trufflehog/blob/main/main.go#L383) -> [ScanGit](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/git.go#L17) -> [Run returns](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L101-L102) [asyncRun](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L106)  -> [run](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L218) -> [runWithoutUnits](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/source_manager.go#L243) -> [Chunks](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L158) -> [scanRepos](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L158) -> [CloneRepo (returned from a parent function depending on the type of git auth)](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git.go#L425-L440)
> 
> 
> 
> I've got a couple of implementation questions:
> 
> 
> 
> - By default there are at least two scanner PIDs running for a given scan job. Depending on concurrency values, it may be more. Is there a rule/best-practice for which PID should actually be interpolated into the directory name?

There is only the launcher and the scanner. It does not depend on concurrency values.

> - What's the cleanest way to pass the PID value through to the tempdir creation? Attach it to the `ctx` object?

You should be able to get it at site of the tempdir creation using something like os.Getpid()

mcastorina:
> What's the cleanest way to pass the PID value through to the tempdir creation? Attach it to the `ctx` object?

It depends on if we want to use the actual PID assigned by the OS, or if we want to generate an internal one using [common.RandomID(5)](https://github.com/trufflesecurity/trufflehog/blob/1642f45c7972e44274354ff40b98f55534e7dce6/pkg/common/utils.go#L58). If the OS PID, I agree with Dustin's suggestion. If not, I think using the context is a reasonable approach.

What is the purpose of using the PID in the name?

codevbus:
Getting this test failure:
```
--- FAIL: TestGitEngine (0.04s)
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 75 [running]:
testing.tRunner.func1.2({0x106b25780, 0x14000b4a108})
        /Users/mike.vanbuskirk/.gvm/gos/go1.21.0/src/testing/testing.go:1545 +0x1c8
testing.tRunner.func1()
        /Users/mike.vanbuskirk/.gvm/gos/go1.21.0/src/testing/testing.go:1548 +0x360
panic({0x106b25780?, 0x14000b4a108?})
        /Users/mike.vanbuskirk/.gvm/gos/go1.21.0/src/runtime/panic.go:914 +0x218
github.com/trufflesecurity/trufflehog/v3/pkg/sources/git.CloneRepo({0x106c936b0, 0x140015c2d50}, 0x0, {0x14001568000, 0x34}, {0x0, 0x0, 0x0})
        /Users/mike.vanbuskirk/truffle-build/trufflehog/pkg/sources/git/git.go:380 +0x280
github.com/trufflesecurity/trufflehog/v3/pkg/sources/git.CloneRepoUsingUnauthenticated(...)
        /Users/mike.vanbuskirk/truffle-build/trufflehog/pkg/sources/git/git.go:483
github.com/trufflesecurity/trufflehog/v3/pkg/sources/git.PrepareRepo({0x106c936b0, 0x140015c2d50}, {0x106048767, 0x34})
        /Users/mike.vanbuskirk/truffle-build/trufflehog/pkg/sources/git/git.go:942 +0x4fc
github.com/trufflesecurity/trufflehog/v3/pkg/engine.TestGitEngine(0x14001868000)
        /Users/mike.vanbuskirk/truffle-build/trufflehog/pkg/engine/git_test.go:34 +0xe0
testing.tRunner(0x14001868000, 0x106c604e8)
        /Users/mike.vanbuskirk/.gvm/gos/go1.21.0/src/testing/testing.go:1595 +0xe8
created by testing.(*T).Run in goroutine 1
        /Users/mike.vanbuskirk/.gvm/gos/go1.21.0/src/testing/testing.go:1648 +0x33c
FAIL    github.com/trufflesecurity/trufflehog/v3/pkg/engine     2.527s
ok      github.com/trufflesecurity/trufflehog/v3/pkg/gitparse   (cached)
ok      github.com/trufflesecurity/trufflehog/v3/pkg/giturl     (cached)
ok      github.com/trufflesecurity/trufflehog/v3/pkg/handlers   (cached)
ok      github.com/trufflesecurity/trufflehog/v3/pkg/log        (cached)
ok      github.com/trufflesecurity/trufflehog/v3/pkg/sanitizer  (cached)
FAIL
make: *** [test-community] Error 1
```

rosecodym (commented):
It's a bummer that this solution precludes future parallelization of this test, but threading the executable name down through the call tree would be even more of a bummer :(

pkg/sources/git/git.go
```diff
@@ -319,6 +322,46 @@ type cloneParams struct {
...
+	tempDir := os.TempDir()
+	files, err := os.ReadDir(tempDir)
+	if err != nil {
+		return err
+	}
+
+	pidStr := strconv.Itoa(pid)
+
+	for _, file := range files {
+		if file.IsDir() && strings.Contains(file.Name(), pidStr) && !strings.Contains(file.Name(), pidStr+"-") {
```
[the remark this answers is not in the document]
codevbus: I think I mistakenly left in logic from two separate iterations of this. Will fix.

pkg/sources/git/git.go
```diff
@@ -319,6 +322,48 @@ type cloneParams struct {
...
+
+	pidStr := strconv.Itoa(pid)
+
+	for _, file := range files {
+		if file.IsDir() && strings.Contains(file.Name(), pidStr) {
+			dirPath := fmt.Sprintf("%s/%s", tempDir, file.Name())
+			if err := os.RemoveAll(dirPath); err != nil {
+				ctx.Logger().Error(err, "Error deleting temp directory", "directory path", dirPath)
+			}
+			logger.V(1).Info("Deleted directory", dirPath)
```
rosecodym, line 361: either use `ctx.Logger()` or `logger` in both places imo

pkg/sources/git/git.go
```diff
@@ -319,6 +322,48 @@ type cloneParams struct {
...
+	for _, proc := range procs {
+		if strings.Contains(proc.Executable(), ExecutableName) {
+			pids = append(pids, proc.Pid())
+		}
+	}
+
+	return pids, nil
+}
+
+func cleanTempDir(ctx context.Context, pid int) error {
```
rosecodym, line 345: this function is in kind of a weird place: some of its errors it returns, but others it logs. you could get consistency by either logging all the errors and returning nothing or using [errors.Join](https://pkg.go.dev/errors#Join) to combine deletion errors into a single error and returning that. @ahrav what's your opinion on this?

pkg/sources/git/git.go
```diff
@@ -329,7 +374,21 @@ func CloneRepo(ctx context.Context, userInfo *url.Userinfo, gitURL string, args
 	if err = GitCmdCheck(); err != nil {
 		return "", nil, err
 	}
-	clonePath, err := os.MkdirTemp(os.TempDir(), "trufflehog")
+
+	pids, _ := getScannerPIDs(ctx)
+
+	for _, pid := range pids {
+		err := cleanTempDir(ctx, pid)
```
rosecodym, line 381: isn't this going to delete all the temp dirs for active trufflehog processes? isn't that the opposite of what we want?

codevbus: I might just leave a comment to explain it; it will only delete dirs that do *not* match those PID values.

go.sum
```diff
@@ -1013,8 +1016,13 @@ honnef.co/go/tools v0.0.0-20190106161140-3f1c8253044a/go.mod h1:rf3lG4BRIbNafJWh
 honnef.co/go/tools v0.0.0-20190418001031-e561f6794a2a/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt0JzvZhAg=
+<<<<<<< HEAD
```
rosecodym, line 1019: ruh roh

dustin-decker:
This doesn't belong in the git package because this problem is not unique to git and we need to call it from the scanner level.

rosecodym:
@dustin-decker 

> You should be able to get it at site of the tempdir creation using something like os.Getpid()

> This doesn't belong in the git package because this problem is not unique to git and we need to call it from the scanner level.

The git package makes its own tempdirs, right? So these seem to be in conflict to me. Or am I misunderstanding your meaning somewhere? (I can see a case for extracting the temp dir path inside the git module but bubbling the cleanup logic all the way back to the scanner, if that's what you're going for?)

codevbus:
Refactored per feedback, ty!

Remaining questions:

- Do I have logging vs. returning errors in the right places?
- Do I need to bother defining interfaces for these functions?
- Cleanup logic is in `main.go`, PID-grabbing is in `git.go`. Should these be in the `common` module and imported so they can be re-used elsewhere?

Confirmed this works in local testing with git. If a scan is terminated non-gracefully, there are temp dirs created in `$TEMPDIR` with the following schema: `trufflehog-<pid>-<rand>`. On a successful scan, directories not containing live PID values are removed.

main.go
```diff
@@ -176,6 +176,34 @@ func init() {
...
+	tempDir := os.TempDir()
+	files, err := os.ReadDir(tempDir)
+	if err != nil {
+		return fmt.Errorf("Error reading temp dir: %w", err)
+	}
+
+	pidStr := strconv.Itoa(pid)
+
+	for _, file := range files {
+		if file.IsDir() && strings.Contains(file.Name(), dirName) && !strings.Contains(file.Name(), pidStr) {			dirPath := fmt.Sprintf("%s/%s", tempDir, file.Name())
```
mcastorina, line 197: **Nit:** [filepath.Join](https://pkg.go.dev/path/filepath#Join) can be used to handle OS-specific separators.

main.go
```diff
@@ -176,6 +176,34 @@ func init() {
...
+			if err := os.RemoveAll(dirPath); err != nil {
+				return fmt.Errorf("Error deleting temp directory: %s", dirPath)
+			}
+			ctx.Logger().V(1).Info("Deleted directory", dirPath)
```
mcastorina, line 201: The logger we're using uses optional key/values at the end of an `Info` argument list. In other words, it's expecting:

```go
Info(message, key1, value1, key2, value2, ...)
```

Since it's a vararg, it's syntactically valid to call it like `Info(message, key)` like we have here, and it won't surface until runtime as a weird looking log.

These sorts of mistakes are easy to both make and miss, but I'm not sure how to prevent them. 😕

pkg/sources/git/git.go
```diff
@@ -319,6 +321,43 @@ type cloneParams struct {
...
+
+	for _, proc := range procs {
+		if strings.Contains(proc.Executable(), executable) {
+			pids = append(pids, proc.Pid())
+		}
+	}
+
+	if len(pids) < 2 {
+		return pids[0], nil
+	} else {
```
mcastorina, line 347: **Suggestion:** This will panic if `len(pids) == 0`. Maybe something like this:

```go
if len(pids) == 0 {
    return 0, fmt.Errorf("PID for %q not found", executable)
} else if len(pids) == 1 {
    return pids[0], nil
} else {
    // ...
}
```

pkg/sources/git/git.go
```diff
@@ -319,6 +321,43 @@ type cloneParams struct {
...
+
+	        // Sort the PIDs in ascending order
+	        sort.Ints(pids)
+
+	        // Return the lowest value of the first two PIDs
+	        if pids[0] < pids[1] {
+	        	return pids[0], nil
+	        }
+	        return pids[1], nil
```
mcastorina, line 356: I don't understand why we're explicitly comparing the first two in the list if it's already sorted? Wouldn't `pids[0]` always be the lowest value?

pkg/sources/git/git.go
```diff
@@ -319,6 +321,43 @@ type cloneParams struct {
 	clonePath string
 }
 
+//Defines the interface for returning a single PID value given an executable name
+type PID interface {
+	getPIDs(ctx context.Context, executable string) (int, error)
+}
+
+var ExecutableName = "trufflehog"
```
mcastorina, line 329: I think this is true in most cases, but not always. Could we use `os.Args[0]` instead?

mcastorina (commented):
Miscellaneous suggestion: running `gofmt -w .` should fix any formatting / whitespace issues.

pkg/sources/git/git.go
```diff
@@ -319,6 +321,36 @@ type cloneParams struct {
...
+	procs, err := ps.Processes()
+	if err != nil {
+		return 0, fmt.Errorf("error getting jobs PIDs: %w", err)
+	}
+
+	for _, proc := range procs {
+		if strings.Contains(proc.Executable(), executable) {
+			pids = append(pids, proc.Pid())
+		}
+	}
```
dustin-decker, line 343: Use os.Getpid()

pkg/sources/git/git.go
```diff
@@ -329,7 +361,15 @@ func CloneRepo(ctx context.Context, userInfo *url.Userinfo, gitURL string, args
...
 		return "", nil, err
 	}
-	clonePath, err := os.MkdirTemp(os.TempDir(), "trufflehog")
+
+	pid, err := GetScannerPIDs(ctx, ExecutableName)
+	if err != nil {
+		return "", nil, err
+	}
+
+	tmpdir := fmt.Sprintf("%s-%d-", ExecutableName, pid)
```
dustin-decker, line 370: I recommend adding a MkdirTemp to a `cleantemp` package that takes care of all of this, using os.Getpid instead of GetScannerPIDs

dustin-decker:
> be in conflict to me. Or am I misunderstanding your meaning somewhere? (I can see a case for extracting the temp dir path inside the git module but bubbling the cleanup logic all the way back to the scanner, if that's what you're going for.)

I meant,

- the scanner should clean orphaned temp directories at a higher level (probably a goroutine ticker kicked off in main.go)
- the logic should be contained in a separate package (suggesting `cleantemp`)
- the `cleantemp` package should also contain a MkdirTemp function that returns a temp directory for use by Sources, without each Source being concerned with prefixes or PIDs. os.Getpid() can be used for fetching the current process PID.

main.go
```diff
@@ -213,6 +215,24 @@ func main() {
...
+	var execName = "trufflehog"
+
+	ticker := time.NewTicker(900 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			pid := os.Getpid()
+			err = cleantemp.CleanTempDir(ctx, execName, pid)
```
dustin-decker, line 230: This is now removing all temp directories except for the currently running PID. What about other instances of trufflehog? It should only clean up orphaned directories if their trufflehog PID is not running

main.go
```diff
@@ -213,6 +215,24 @@ func main() {
 	if err != nil {
 		logFatal(err, "error occurred with trufflehog updater 🐷")
 	}
+
+	ctx := context.Background()
+
+	var execName = "trufflehog"
+
+	ticker := time.NewTicker(900 * time.Second)
```
dustin-decker, line 223: 15 * time.Minute is more readable

main.go
```diff
@@ -213,6 +232,20 @@ func main() {
...
+	ctx := context.Background()
+
+	var execName = "trufflehog"
+
+	// Inital orphaned dir cleanup when the scanner is invoked
+	pid := os.Getpid()
+	err = cleantemp.CleanTempDir(ctx, execName, pid)
+	if err != nil {
+		ctx.Logger().Error(err, "Error cleaning up orphaned directories ")
+	}
```
dustin-decker, line 245: This can go into the function you made before it goes into the ticker loop.

pkg/cleantemp/cleantemp.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return fmt.Errorf("Error reading temp dir: %w", err)
+	}
+
+	// Current PID
+	pidStr := strconv.Itoa(pid)
+
+	for _, file := range files {
+		// Make sure we don't delete the working dir of the current PID
+		if file.IsDir() && strings.Contains(file.Name(), dirName) && !strings.Contains(file.Name(), pidStr) {
```
dustin-decker, line 62: It would be good to do a format check on the name to ensure that we don't accidentally delete another folder containing 'trufflehog' in the name. I've seen people unpack the distribution there for testing.

codevbus: Would `dirName+"-"` suffice or should it be a full regex check?

dustin-decker (approved):

PR #1891, opened 2023-10-13
Fix wrong line number

shreyas-sriram:
### Description:
This fixes the wrong line numbers that are reported. See https://github.com/trufflesecurity/trufflehog/issues/1876.

- line number is to be calculated as `count("\n") + 1`

### Checklist:


shreyas-sriram:
Hey @dustin-decker! Can I get a review of this?

bill-rich (approved):
Thanks for the fix!

sxlijin:
This doesn't appear to be a "fix wrong line number" PR - this changes trufflehog's JSON output from returning 0-indexed line numbers to 1-indexed line numbers and **is a backwards breaking change**.

bill-rich:
Turns out that the issue is only in the `filesystem` source. The fix is to add line-aware chunking there.

PR #1893, opened 2023-10-16
Add ShannonEntropy test for an empty string

mcastorina:
### Description:
I thought I noticed a divide by 0 error in `ShannonEntropy`, so I added a test for it and it passed! I forgot `1 / 0.0` is `+inf` in floating point instead of an error. Adding the test in case the implementation ever changes.

### Checklist:


dustin-decker (approved):
TY

PR #1895, opened 2023-10-17
Add Coinbase Wallet-as-a-Service detector

rgmz:
### Description:
This adds a detector for Coinbase's [Wallet as a Service (WaaS)](https://docs.cloud.coinbase.com/waas/docs/auth) API.

The detector logic is implemented using Coinbase's SDK, which adds just over 1MiB to the binary. My rationale for doing so is that reverse-engineering the SDK might result in brittle/unreliable logic, as the endpoints are not publicly documented (to my knowledge).

### Checklist:


zricethezav:
@rgmz do you by chance have a screenshot of the tests passing?

rgmz:
Sure, here you go: 
image

I had to disable two test cases because this does not use the provided `http.Client`, as mentioned above.

rosecodym:
> I had to disable two test cases because this does not use the provided `http.Client`, as mentioned above.

[I believe you can inject a configured `httpClient`](https://pkg.go.dev/github.com/coinbase/waas-client-library-go@v1.0.8/clients#WaaSClientOption) and use it in those tests. If not, you might be able to implement the timeout test by using a context with a timeout (I can find an example elsewhere in the codebase if that'd be helpful), and you can just delete the other test. You might as well delete the `httpClient` from the `Scanner` struct in that case.

rgmz:
> [I believe you can inject a configured httpClient](https://pkg.go.dev/github.com/coinbase/waas-client-library-go@v1.0.8/clients#WaaSClientOption) and use it in those tests

You're absolutely right. Works like a charm!

pkg/detectors/coinbase_waas/coinbase_waas.go
```diff
@@ -0,0 +1,142 @@
...
+		"\\r\\n", "\n",
+		"\\n", "\n",
+		"\\r", "\n",
+	)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"organizations", "apiKeys", "ec"}
```
zricethezav, line 47: @rgmz should `ec` be `begin ec`?

rgmz: Probably. I wasn't sure how spaces or letters within other words would be handled (i.e., would `begin ec` catch both `BEGIN EC ` and `BEGIN ECDSA`)?

Though I suppose you could just add both as separate keywords.

rgmz: `begin ec` seems to work for both cases.

image

dustin-decker (approved):
Looks great, thank you!

PR #1899, opened 2023-10-17
Scan GitHub issue & pr titles

rgmz:
### Description:

This change appends issue and pr titles to the scan data for `--issue-comments` and `--pr-comments` respectively. People apparently share secrets in titles, for instance: https://github.com/coinbase/waas-client-library-go /issues/52 (intentional space to avoid linking).

It would be ideal to write test cases for this, however, off the top of my head I'm not sure the best way to test `chunkIssues` and `chunkPullRequests`. Unit tests would seemingly require either mocking `s.visibilityOf()` or rewriting the code.

https://github.com/trufflesecurity/trufflehog/blob/6ea3a7da4a868b9e1375f038a65da0e8d1112a9f/pkg/sources/github/github.go#L1313-L1335

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -1356,7 +1356,7 @@ func (s *Source) chunkIssues(ctx context.Context, repoInfo repoInfo, issues []*g
 					},
 				},
 			},
-			Data:   []byte(sanitizer.UTF8(issue.GetBody())),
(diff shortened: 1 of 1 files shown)


ahrav (approved):
This makes sense to me. Goes to show, secrets literally are everywhere 😅 Thanks for the addition.

rgmz:
It occurs to me that it might be better to separate them with a newline, in case any patterns rely on word boundaries.

PR #1900, opened 2023-10-18
Add Voiceflow detector

rgmz:
### Description:

This adds a detector for https://voiceflow.com Dialog Manager (`VF.DM.xxx.yyy`) tokens. A valid token can be created by [signing up for a Voiceflow account and following these instructions](https://developer.voiceflow.com/reference/project).

### Checklist:


zricethezav:
@rgmz same goes for this PR, do you have a screenshot of the tests passing?

zricethezav:
Thanks for including the screenshot!

@rgmz 
> On that note, I should probably rewrite my tests to fold TestVoiceflow_Pattern cases into TestVoiceflow_FromChunk. The former was copied from elsewhere and doesn't seem necessary in this case.

Eh, I don't think that's necessary. `TestVoiceflow_Pattern` is a nice way to test the regex without calling out to the service. Hell, we might include something like that in the default template in the future.

zricethezav (approved):

PR #1916, opened 2023-10-18
Detector-Competition-Fix: Fix plaid.com API key detection

lc:
### Description:
This pull request fixes the Plaid.com detector. The detector should work for the **production** API rather than the development. 

Arguably, the integrity, confidentiality, and availability of a business's `plaid.com` is not affected in the sandbox/development environment.

CleanShot 2023-10-18 at 14 42 13

### Checklist:


rgmz:
> Arguably, the integrity, confidentiality, and availability of a business's `plaid.com` is not affected in the sandbox/development environment.

I feel like it should check both development and production. The former still has access to real data, apparently.

> Plaid has three environments: Sandbox, Development, and Production. Items, once created, cannot be moved to another environment. The Sandbox environment supports only test Items. The Development environment supports up to 100 live Items using real data.

https://plaid.com/docs/api/#api-host

lc:
> > Arguably, the integrity, confidentiality, and availability of a business's `plaid.com` is not affected in the sandbox/development environment.
> 
> I feel like it should check both development and production. The former still has access to real data, apparently.
> 
> > Plaid has three environments: Sandbox, Development, and Production. Items, once created, cannot be moved to another environment. The Sandbox environment supports only test Items. **The Development environment supports up to 100 live Items using real data.**
> 
> https://plaid.com/docs/api/#api-host
> 
> That being said, I'm not familiar with how Plaid's API model works. If a valid client ID and secret can be used against any environment, then it doesn't make sense to check both dev and prod.

The client_id remains the same, but the client_secret is different for dev/prod – added a check to see if we've found a development or production key. The environment verified is returned in `s1.ExtraData`

zricethezav (approved):
@lc nice work!

PR #1917, opened 2023-10-18
Detector-Competition-Fix: Fix currencycloud.com API key

lc:
### Description:
This pull request fixes the currencycloud.com detector. The detector should work for the production API rather than the development.

### Checklist:


zricethezav:
> This pull request fixes the currencycloud.com detector. The detector should work for the production API rather than the development.

@lc Similar to the fix for plaid: https://github.com/trufflesecurity/trufflehog/pull/1916/files, what about adding a check for dev and prod?

lc:
@zricethezav I am unsure if the devapi contains any live information. I can add the check just in case.

zricethezav (approved):

PR #1923, opened 2023-10-19
Improve ScrapingBee detections

rgmz:
### Description:

### Checklist:


pkg/detectors/scrapingbee/scrapingbee.go
```diff
@@ -53,15 +53,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 			res, err := client.Do(req)
 			if err == nil {
-				defer res.Body.Close()
-				if res.StatusCode >= 200 && res.StatusCode < 300 {
+				if res.StatusCode == http.StatusOK {
```
zricethezav, line 56: defer statement should remain and anything in between 200 and 300 should be considered acceptable. I think the verification block should remain the same

zricethezav (changes requested):
Happy to accept the scrapingbee keyword permutations but the rest should be removed imo

pkg/detectors/detectors.go
```diff
@@ -121,7 +121,7 @@ func CleanResults(results []Result) []Result {
 func PrefixRegex(keywords []string) string {
 	pre := `(?i)(?:`
 	middle := strings.Join(keywords, "|")
-	post := `)(?:.|[\n\r]){0,40}`
+	post := `)(?:.|[\n\r]){0,120}`
```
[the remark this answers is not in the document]
rgmz: I think it's something worth investigating or making configurable — perhaps as a separate change. Based on my tests, the current limit fails to catch certain live secrets.

pkg/detectors/scrapingbee/scrapingbee.go
rgmz: > defer statement should remain

GoLang was yelling at me about potential resource leaks from using `defer` in a for loop. Is that not a concern in this context?

> anything in between 200 and 300 should be considered acceptable. 

ScrapingBee seems to return `200` for all successful responses (2xx and 3xx) and passes the origin response in the `Spb-initial-status-code` header.

```sh
$ curl 'https://app.scrapingbee.com/api/v1?api_key=$API_KEY&url=https://httpbin.org/status/209&render_js=False'  -i
HTTP/1.1 200 OK
...
Spb-initial-status-code: 209
Spb-resolved-url: https://httpbin.org/status/209
```

pkg/detectors/detectors.go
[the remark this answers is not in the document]
ankushgoel27: i have seen it on many ocassions where the secret is missed if the keyword is not before but after the secret or the keyword is beyond 40 chars of the secret.
Also, when there are multiple matches to happen such as domain, id, secret etc. using a prefix regex doesn't makes sense due to the 40- char limit as that will be used up by the regex matches itself.

pkg/detectors/scrapingbee/scrapingbee.go
ahrav: @rgmz yea the defer thing still bugs me to this day. It hasn't been an issue "yet", but i'm sure give it enough time, it's gonna bite us.

rgmz: From what I've read, to properly handle response bodies we would need to:

1. Move the logic into a separate function so that defer is called when expected (I think? This is my own conjecture)
2. `defer res.Body.Close()`
3. Call `io.Copy(io.Discard, res.Body)` so that the connection can be reused


Sources:
- https://old.reddit.com/r/golang/comments/13fphyz/til_go_response_body_must_be_closed_even_if_you/jjynmj7/
- https://andrii-kushch.medium.com/is-it-necessary-to-close-the-body-in-the-http-response-object-in-golang-171c44c9394d

ahrav: Correct, or use an anonymous function directly at the callsite. Imo, it's a little cleaner to have a separate function.

zricethezav:
@rgmz closing this one out as you opened another PR for the response draining. Happy to accept the PR with these changes though: `Compound detector names like scrapingbee should have keywords handling permutations like scraping_bee and scraping-bee`. Feel free to reuse this PR or open a new one if you want 👍🏻

PR #1930, opened 2023-10-20
Add UnitHook and NoopHook implementations

mcastorina:
### Description:
The UnitHook tracks metrics per unit of a job, and safely makes them available to the caller. It also works for sources that do not support source units.

### Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,7 @@ require (
 	github.com/h2non/filetype v1.1.3
 	github.com/hashicorp/go-retryablehttp v0.7.4
 	github.com/hashicorp/golang-lru v0.5.1
+	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/jlaffaye/ftp v0.2.0
 	github.com/joho/godotenv v1.5.1
 	github.com/jpillora/overseer v1.1.6
--- a/pkg/sources/job_progress_hook.go
+++ b/pkg/sources/job_progress_hook.go
@@ -6,19 +6,41 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
 // UnitHook implements JobProgressHook for tracking the progress of each
 // individual unit.
 type UnitHook struct {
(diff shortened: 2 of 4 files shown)


pkg/sources/job_progress_hook.go
```diff
@@ -0,0 +1,214 @@
...
+	// lru.NewWithEvict can only fail if the size is < 0.
+	cache, _ := lru.NewWithEvict(1024, func(key string, value *UnitMetrics) {
+		if value.IsFinished() {
+			return
+		}
+		ctx.Logger().Error(fmt.Errorf("eviction"), "dropping unit metric",
+			"id", key,
+			"metric", value,
+		)
+	})
```
mcastorina, line 38: This eviction function is called even when we manually call `Remove`, which is I added a check for `value.IsFinished()`.

However, it's possible for a value to be finished and evicted without ever being consumed. In fact, that's probably more common than evicting an in-progress metric. In this case, we wouldn't see the log.

TODO: Refactor this callback to log for unhandled evictions.

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/h2non/filetype v1.1.3
 	github.com/hashicorp/go-retryablehttp v0.7.4
 	github.com/hashicorp/golang-lru v0.5.1
+	github.com/hashicorp/golang-lru/v2 v2.0.7
```
rosecodym, line 50: @ahrav i thought we already had an lru cache somewhere? does it just not use this package?

rosecodym (approved):
is your approach to these metrics documented anywhere? i don't recall anything about them in the original design doc (maybe i'm just forgetting them) and i would like a high level overview of your strategy

go.mod
ahrav: https://github.com/trufflesecurity/trufflehog/blob/68f28a0e34f9f95bbfc533595756c6525c7b508f/go.mod#L49C7-L49C7 yep.

ahrav: maybe not v2?

mcastorina:
> is your approach to these metrics documented anywhere? i don't recall anything about them in the original design doc (maybe i'm just forgetting them) and i would like a high level overview of your strategy

No, this approach came about mostly through exploratory coding.

Thanks for the callout, I'll write up a doc!

PR #1933, opened 2023-10-21
Detector-Competition-Fix: Fix AppFollow Detection & Verification

lc:
### Description:
This fixes AppFollow detection and verification.  AppFollow uses JWTs for API keys and has changed the header you're required to send:

CleanShot 2023-10-21 at 09 09 35@2x

Furthermore, the `https://api.appfollow.io/test` endpoint has been removed and 404's when you send verification. 

Changed the regular expression to match JWTs and now perform validation by requesting `https://api.appfollow.io/api/v2/account/users`

CleanShot 2023-10-21 at 09 07 45@2x

### Checklist:


pkg/detectors/appfollow/appfollow.go
```diff
@@ -22,7 +20,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appfollow"}) + `\b([0-9A-Za-z]{20})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appfollow"}) + `\b([\w-]*\.[\w-]*\.[\w-]*)\b`)
```
zricethezav, line 23: @lc since this is a JWT I think we can tighten up the regex a bit. See https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/formio/formio.go#L23 for an example

rgmz: I can't speak to AppFollow's format, but based on my testing you can really only set a minimum for each fragment. The payload and signature can be quite long (>1,000 chars), and [as mentioned elsewhere](https://github.com/gitleaks/gitleaks/pull/1214#issuecomment-1685136432) the regexp package only allows you to specify <=1,000 chars.

lc: tightened it up @zricethezav:

CleanShot 2023-10-23 at 12 24 03

pkg/detectors/appfollow/appfollow.go
```diff
@@ -22,7 +20,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appfollow"}) + `\b([0-9A-Za-z]{20})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appfollow"}) + `\b(eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9\.[0-9A-Za-z]{74}\.[0-9A-Z-a-z\-_]{43})\b`)
```
rgmz, line 23: Do we know for certain that the JWT header will always be exactly:

```
{"typ":"JWT","alg":"HS256"}
```

lc: Yes, I believe so.

zricethezav (approved):

Blissful304 (approved):

Blissful304 (approved):

PR #1934, opened 2023-10-21
Detector-Competition-Fix: Fix UrlScan verification

lc:
### Description:
The verification for urlscan is broken as the endpoint at `https://urlscan.io/user/quotas` always returns `200 OK`. 

When a request does not contain an API key, the `source` field  in the JSON response is `"ip-address"`

CleanShot 2023-10-21 at 09 19 59@2x

When a valid API key **is** sent in the request (and is valid), the `source` field in the JSON response is `"team"`:

CleanShot 2023-10-21 at 09 20 51@2x

### Checklist:

CleanShot 2023-10-21 at 09 18 11@2x


rosecodym:
Thanks for taking a look at this. I see what you mean about urlscan requests that do not include an API key, but this detector always provides an API key, so I can't tell when the new logic will be exercised. (If the API key is provided but invalid, urlscan appears to return a `400` or a `401` depending on whether the key is formatted correctly.) What prompted you to make this change?

lc:
> Thanks for taking a look at this. I see what you mean about urlscan requests that do not include an API key, but this detector always provides an API key, so I can't tell when the new logic will be exercised. (If the API key is provided but invalid, urlscan appears to return a `400` or a `401` depending on whether the key is formatted correctly.) What prompted you to make this change?

Oh good point – I was thinking mostly of the case where the API Key was empty, but that should never happen. For some reason, I was under the assumption that `200 OK` would be a response that URLscan would respond to the trufflehog client with

zricethezav:
closing this for reasons @rosecodym mentioned

PR #1935, opened 2023-10-21
Detector-Competition-Fix: Update formio regex to match Jwt token

fumblehool:
### Description:
The Regex for Formio (https://portal.form.io) JWT token fails to match token as length of token is less.
It has a hard limit of 310 characters but while testing, my token only had 228 characters.
So this PR sets this length to a variable b/w 200 and 310 to avoid detection failures.

### Checklist:

### Running Tests Locally
Screenshot 2023-10-21 at 10 15 31 PM

### Running formio locally:
In order to run formio locally using Docker, follow the official guide: 
https://help.form.io/deployments/deployment-guide/local-deployment


zricethezav:
@fumblehool 
> So this PR sets this length to a variable b/w 200 and 310 to avoid detection failures.


Any reason why 200? Any risk of this being lower than 200?

fumblehool:
@zricethezav the length depends on the payload data.
Currently, it follows this structure:
```
{
  "user": {
    "_id": XXXXX
  },
  "iss": "https://api.form.io",
  "sub": XXXXX,
  "jti": XXXXX,
  "iat": XXXXX,
  "exp": XXXXXX
}
```
For this structure, the length is around 310.
If I convert user object into a user_id string, the length comes out to be 210 (close to original 228 limit).

I've converted this into a range to avoid updating the regex frequently.

zricethezav:
> For this structure, the length is around 310.
If I convert user object into a user_id string, the length comes out to be 210 (close to original 228 limit).
I've converted this into a range to avoid updating the regex frequently.

Makes sense, thanks for the explanation

zricethezav (approved):

PR #1949, opened 2023-10-22
Detector-Competition-Fix: Fix ScreenshotAPI Verification

lc:
### Description:
This fixes the Screenshot API verification. The endpoint at `https://shot.screenshotapi.net/screenshot?token=KEY&url=https://google.com&width=1920&height=1080&output=image` returns a `400` when an invalid key is sent. However, when a valid key is sent, the API responds with a `302` to the screenshot captured. The current validator does not catch this. 

So, I've updated the verifier, removing the `output=image` parameter, which causes the API to respond with `200 OK` and with the JSON object:

```json
{"screenshot":"https://storage.screenshotapi.net/google_com_42a262253e9b.png","url":"https://google.com","created_at":"2023-10-22T15:26:04.155Z","is_fresh":false,"token":"REDACTED","width":"1920","height":"1080","ttl":"2023-11-21T16:19:54.724Z"}
```

It now validates correctly.

### Checklist:

image


zricethezav:
Nice work

zricethezav (approved):

PR #1952, opened 2023-10-23
Upgrade gocb and gocbcore

nyanshak:
* gocb has been updated to use newer version of gocbcore, so the replace directive to use an older gocbcore is no longer needed.


dustin-decker (approved):
Thank you!

PR #1957, opened 2023-10-24
Detector-Competition-Fix: Fix redis to now support SSL, and look for azure redis connection strings

dylanTruffle:
### Description:

### Checklist:

### DIFF
--- a/pkg/detectors/redis/redis.go
+++ b/pkg/detectors/redis/redis.go
@@ -2,9 +2,9 @@ package redis
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"regexp"
-	"fmt"
 	"strings"
 
 	"github.com/go-redis/redis"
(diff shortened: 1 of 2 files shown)


pkg/detectors/redis/redis.go
```diff
@@ -18,7 +19,8 @@ type Scanner struct{}
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\bredis://[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\bredi(s{1,2})://[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
```
dustin-decker, line 22: It looks like this added a new capturing group - so I think the match index is probably wrong for the existing detector now. Did you run the test?

Also, please run go fmt

zricethezav: formatted and fixed the test, might've been a copy/paste error

zricethezav (approved):

PR #1958, opened 2023-10-24
Detector-Competition-Feat: Adding Azure Container Registry Password Detector

dylanTruffle:
### Description:

### Checklist:


pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+    url = regexp.MustCompile(`([a-zA-Z0-9-]{1,100})\.azurecr\.io`)
+    password = regexp.MustCompile(`[A-Za-z0-9+/=]{52}`)
```
dustin-decker, line 26: This needs boundary characters or this part can get a lot of matches.

dustin-decker: Also looks like you need to run `go fmt -w`. Your editor should do this for you.

pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			if verify {
+				client := s.client
+				if client == nil {
+					client = defaultClient
+				}
+
+				auth := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", username, password)))
+				url := fmt.Sprintf("https://%s/v2/", endpoint)
```
dustin-decker, line 63: Is it possible that endpoint would contain a trailing slash?

pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
```diff
@@ -0,0 +1,97 @@
...
+				req, err := http.NewRequest("GET", url, nil)
+				if err != nil {
+					continue
+				}
+
+				req.Header.Set("Authorization", fmt.Sprintf("Basic %s", auth))
+				resp, err := client.Do(req)
+				if err != nil {
+					continue
```
dustin-decker, line 72: An indeterminate error should be set here. This would happen because of a network issue.

pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
dylanTruffle: I don't think so, because the regex doesn't match a trailing slash

zricethezav (approved):

PR #1964, opened 2023-10-24
Remove docker container after make protos finishes

fumblehool:
### Description:
Running `make protos` makes proto files using docker container but the container is not removed.
This PR adds the `--rm` flag to make sure user won't get a bunch of stopped containers hogging up space

### Checklist:


dustin-decker (approved):
Thank you

PR #1976, opened 2023-10-24
Update databricks pat domain regex

4390c336:
### Description:
The existing regex doesn’t cover instances hosted on azure and GCP
the prefix can also be a number or a name and can contain one or two dots.
this change makes it cover all the known cases
 
I also fixed the url that lacked the scheme and change it to point to the token API which make more sense here since we are trying to validate tokens.

### Checklist:


ankushgoel27:
shouldn't you also change the 

req, err := http.NewRequestWithContext(ctx, "GET", resDomainMatch + "/api/2.0/clusters/list", nil)

to this?

req, err := http.NewRequestWithContext(ctx, "GET", "https://" + resDomainMatch + "/api/2.0/clusters/list", nil)

PR #1990, opened 2023-10-25
Competition-Detector-Fix: fixed zulipchat detector

ankushgoel27:
The zulipchat detector was not working correctly. Made changes to regex to make it work. Also fixed the latest testing scenarios. 

### Checklist:

zulipchat

### DIFF
--- a/pkg/detectors/zulipchat/zulipchat.go
+++ b/pkg/detectors/zulipchat/zulipchat.go
@@ -21,9 +21,9 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"zulipchat"}) + common.BuildRegex(common.AlphaNumPattern, "", 32))
-	idPat     = regexp.MustCompile(detectors.PrefixRegex([]string{"zulipchat"}) + common.EmailPattern)
-	domainPat = regexp.MustCompile(detectors.PrefixRegex([]string{"zulipchat", "domain"}) + common.SubDomainPattern)
+	keyPat    = regexp.MustCompile(common.BuildRegex(common.AlphaNumPattern, "", 32))
+	idPat     = regexp.MustCompile(`\b(([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9_\-]+)(\.[a-zA-Z]{2,5}){1,2})\b`)
+	domainPat = regexp.MustCompile(`\b([a-z0-9-]+\.zulipchat\.com)\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
@@ -48,7 +48,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 		for _, idMatch := range idMatches {
 			// getting the last word of the string
-			resIdMatch := strings.TrimSpace(idMatch[0][strings.LastIndex(idMatch[0], " ")+1:])
+			resIdMatch := strings.TrimSpace(idMatch[1])
 
 			for _, domainMatch := range domainMatches {
 				if len(domainMatch) != 2 {
@@ -60,10 +60,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				s1 := detectors.Result{
 					DetectorType: detectorspb.DetectorType_ZulipChat,
 					Raw:          []byte(resMatch),
+					RawV2:        []byte(fmt.Sprintf("%s:%s:%s",resMatch, resIdMatch, resDomainMatch)),
 				}
 
 				if verify {
-					req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://%s.zulipchat.com/api/v1/users", resDomainMatch), nil)
+					req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://%s/api/v1/users", resDomainMatch), nil)
 					if err != nil {
 						continue
 					}
--- a/pkg/detectors/zulipchat/zulipchat.go
+++ b/pkg/detectors/zulipchat/zulipchat.go
@@ -12,13 +12,15 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{}
+type Scanner struct{
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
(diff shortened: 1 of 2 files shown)


rosecodym (commented):
Thanks for this! I left a comment about the test secrets management. Once you update your branch with `main` we can get this in.

pkg/detectors/zulipchat/zulipchat_test.go
```diff
@@ -19,13 +20,13 @@ import (
 func TestZulipChat_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors3")
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
 	if err != nil {
 		t.Fatalf("could not get test secrets from GCP: %s", err)
 	}
 	secret := testSecrets.MustGetField("ZULIPCHAT")
+	domain := testSecrets.MustGetField("ZULIPCHAT_DOMAIN")
```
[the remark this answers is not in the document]
ankushgoel27: please check my new commit. I made the changes recommended by you.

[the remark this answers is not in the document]
ankushgoel27: @rosecodym

pkg/detectors/zulipchat/zulipchat_test.go
```diff
@@ -19,13 +20,13 @@ import (
 func TestZulipChat_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors3")
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors2")
```
rosecodym, line 23: Now it's been changed to `detectors2`. `detectors3` was, and still is, correct - you can just leave it the way it was.

pkg/detectors/zulipchat/zulipchat.go
```diff
@@ -12,18 +12,20 @@ import (
...
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"zulipchat"}) + common.BuildRegex(common.AlphaNumPattern, "", 32))
-	idPat     = regexp.MustCompile(detectors.PrefixRegex([]string{"zulipchat"}) + common.EmailPattern)
-	domainPat = regexp.MustCompile(detectors.PrefixRegex([]string{"zulipchat", "domain"}) + common.SubDomainPattern)
+	keyPat    = regexp.MustCompile(common.BuildRegex(common.AlphaNumPattern, "", 32))
```
ahrav, line 26: suggestion: For `keyPat` and `idPat` I think it's safe to leave the `detectors.PrefixRegex` since neither of them have the keyword or prefix directly within the regex. `domainPat` looks good since `zulipchat` is embedded within the regex itself.

ahrav (commented):
Just one small comments otherwise LGTM! Thanks @ankushgoel27

pkg/detectors/zulipchat/zulipchat.go
ankushgoel27: Due to prefixregex 40 char limit, i prefer to leave them out because then the zulipchat keyword needs to present in the line containing the idpat and keypat. If not present, detector wont work. 
Also , being 3 regex's in work and one containing zulipchat domain itself, i think it wont generate any false positives if we leave the prefix regex out as all 3 regexes need to match for this detector to work.

rgmz:
Might I suggest adding `zulip`, as well as `zulip.com` and `zulip.org`?

https://github.com/zulip/python-zulip-api/blob/5199c140772cfde93b437a0a190891299db79ca7/zulip/integrations/svn/zulip_svn_config.py#L4C1-L5

ankushgoel27:
> Might I suggest changing the keyword to `zulip`, as well as adding `zulip.com` and `zulip.org` as domains?
> 
> https://github.com/zulip/python-zulip-api/blob/5199c140772cfde93b437a0a190891299db79ca7/zulip/integrations/svn/zulip_svn_config.py#L4C1-L5

Can you clarify more regarding zulip.com and zulip.org? When i created an account on zulipchat, it gave me a domain as mykeyword.zulipchat.com. One can also have a his own domain redirecting to [keyword].zulipchat.com so having a common domain regex makes sense but not sure about the zulip.org and zulip.com

rgmz:
The site used to be zulip.org at some point: https://github.com/zulip/zulip.github.io

ankushgoel27:
> The site used to be zulip.org at some point: https://github.com/zulip/zulip.github.io

the github link says that zulip.org is not used for anything anymore.

rgmz:
If someone has a token configured to point to zulip.org, wouldn't it still be valid?

For instance, https://chat.zulip.org/ still resolves.

pkg/detectors/zulipchat/zulipchat.go
```diff
@@ -12,24 +12,26 @@ import (
...
+	idPat     = regexp.MustCompile(`\b(([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9_\-]+)(\.[a-zA-Z]{2,5}){1,2})\b`)
```
rgmz, line 27: This needs to be tweaked a bit. At the moment, it wouldn't match:
```
email="bot-name@site.zulipchat.com"
```

pkg/detectors/zulipchat/zulipchat.go
```diff
@@ -12,24 +12,26 @@ import (
...
+	domainPat = regexp.MustCompile(`\b([a-z0-9-]+\.zulipchat\.com)\b`)
```
rgmz, line 28: IMO, this should include both `zulip.com` and `zulip.org` to handle all potential cases.

```suggestion
       // I think this is right. Ought to be double-checked though.
	domainPat = regexp.MustCompile(`\b([a-z0-9-]+\.zulip(?:(?:chat)?\.com|\.org))\b`)
```

https://github.com/alfonsrv/impf-botpy/blob/2eb10199984b3a575f645891003ba4b7afe3ecae/docs/ALERTS.md?plain=1#L52
https://github.com/zulip/zulip-js/blob/db9b0e4f7535552e768d6ecc7b3f4fbf4ea910fe/examples/zuliprc#L3

pkg/detectors/zulipchat/zulipchat.go
ankushgoel27: yes, You are correct. I have modified the regex. 

([a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})  should work.

Thanks for pointing this out.

zricethezav (approved):

PR #1991, opened 2023-10-25
Detector-Competition-Fix: fixed the regex for the packagecloud

ankushgoel27:
Removed the prefix regex to account for variations i have seen in the wild such as https://token:@packagecloud.io where packagecloud appears after and not before the token.

### Checklist:


pkg/detectors/packagecloud/packagecloud.go
```diff
@@ -20,7 +20,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"packagecloud"}) + `\b([0-9a-f]{48})\b`)
+	keyPat = regexp.MustCompile(`\b([0-9a-f]{48})\b`)
```
rgmz, line 23: This will generate way too many false-positives. Perhaps we could have two patterns, one with prefix and one with suffix.

rgmz: Hmm, fair enough.

ahrav: @ankushgoel27 what if instead we use:
`keyPat = regexp.MustCompile(`https://([0-9a-f]{48}):@packagecloud\.io`)`
We can then use the capture group to pull out the actual token. I'm using the regex based on the convo in slack in terms of the structure.

ankushgoel27: @ahrav, its a common problem with many detectors even when there is basic auth being used such as curl -u apikey:secretkey https://site.com or with github such as https://username:token@github.com where the prefix keyword comes after the secrets. this is not just limited to packagecloud.

zricethezav: @ankushgoel27 @ahrav 
We could do something like this:

```
(?i)(?:packagecloud)(?:.|[\n\r]){0,40}\b([0-9a-f]{48}|)\b|(\b([0-9a-f]{48})\b)(?:.|[\n\r]){0,40}(?:packagecloud\.com)
```
https://regex101.com/r/PVokH8/1

I don't think this change should qualify for the detector competition for the following reasons:
1. It's not fixing an endpoint or updating the regex to match a new token but rather it's just expanding the capture scope.
2. This change could be applied to many different detectors. I.e, easy to game the competition.

That said, I think opening an issue on expanding the scope of regex's for basic auth detectors is a valid one. The regex I posted above has the issue of two different capture groups which would need to be addressed within the detector.

PR #2001, opened 2023-10-25
Detector-Competition-Fix: Fix LiveAgent Detector & Verifier

lc:
### Description:
This fixed the LiveAgent detector & verifier. Each `ladesk.com` gets a unique subdomain, so I've updated the code to match a subdomain and verifies detected API Keys against the API on the subdomain. 

image

### Checklist:


pkg/detectors/liveagent/liveagent.go
```diff
@@ -20,7 +21,8 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"liveagent"}) + `\b([a-zA-Z0-9]{32})\b`)
+	domainPat = regexp.MustCompile(`\b(https?://[A-Za-z0-9-]+\.ladesk\.com)\b`)
+	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"liveagent"}) + `\b([a-zA-Z0-9]{32})\b`)
```
rgmz, line 25: We should include `apiKey` as a prefix as it's likely to be more closely associated with the secret than the string "liveagent".

**Example**

```sh
$ curl --location --request GET 'https://your.ladesk.com/api/v3/tickets' \
--header 'apikey: 0865evv883j6auflvlt7zrwgqn32qo74' \
--header 'Content-Type: application/json'
```

https://support.liveagent.com/741982-API-key

lc: Hmm, we could do `apikey|liveagent` so that `LIVEAGENT_APIKEY=KEY_HERE` and that curl command would both be captured...Since there's a check for the domain too there shouldn't be a problem with false positives.

rgmz: Makes sense. Changing it here would produce `liveagent|apikey` in the pattern:
```diff
-	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"liveagent"}) + `\b([a-zA-Z0-9]{32})\b`)
+	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"liveagent", "apikey"}) + `\b([a-zA-Z0-9]{32})\b`)
```

And then it would also make sense to add `ladesk` to the list of keywords to ensure the curl command would be selected for scanning in the absence of `liveagent`.
```diff
func (s Scanner) Keywords() []string {
-	return []string{"liveagent"}
+	return []string{"liveagent", "ladesk"}
}
```

lc:
I've updated the regular expressions

zricethezav (approved):

PR #2010, opened 2023-10-27
Remove verify flag from Aho-Corasick core

rosecodym:
### Description:
The Aho-Corasick wrapper we have tracks information about whether verification should be enabled on an individual detector basis, but that functionality isn't related to the matching functionality of Aho-Corasick, and including it complicates the implementation. This PR removes it to simplify some things.

This PR removes some code that supported a potential future implementation of detector-specific verification settings, but that feature has not actually been implemented yet, so there's no loss of functionality. If we want that feature we can add it back on top of this in a more separated way.

### Checklist:


ahrav (approved):

pkg/engine/engine.go
```diff
@@ -53,7 +53,7 @@ type Engine struct {
 	// CLI flags.
 	concurrency uint8
 	decoders    []decoders.Decoder
-	detectors   map[bool][]detectors.Detector
+	detectors   []detectors.Detector
```
mcastorina, line 56: o7 `map[bool][]detectors.Detector`, you served us well.

mcastorina (approved):
lgtm!

PR #2016, opened 2023-10-28
Detector-Competition-Feat: Added Budibase API token detector

fumblehool:
### Description:
PR to add budibase detector:

links
Budibase login: https://moz123.budibase.app/builder/portal/apps
How to get API token: https://docs.budibase.com/docs/public-api
API used for verification: https://docs.budibase.com/reference/appsearch

### Checklist:

### Verification:

Screenshot 2023-10-28 at 2 26 03 PM


pkg/detectors/budibase/budibase.go
```diff
@@ -0,0 +1,97 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"budibase"}) + `\b([a-z0-9]{32}-[a-z0-9]{78})\b`)
```
ahrav, line 25: question: How were you able to determine that the group after "-" should be 78 characters long? I created a handful of test tokens and they all appeared to be 80 characters long. I was unable to get the tests to pass until I modified the regex. Do you know if there is a difference between maybe the hosted vs on-prem api tokens?

Additionally I think we can tighten the regex by limiting the alpha characters to just `a-f`, i'm pretty sure these are hex represented.

ahrav (changes requested):
One clarifying question around the regex length. Otherwise the rest of the code LGTM!

pkg/detectors/budibase/budibase.go
fumblehool: I created the tokens from hosted budibasea and everytime it come out with 78 chars 🤔
Should I keep it as `{78,80}` to include both tokens?

Also, I'll limit the alpha characters to just `a-f` 👍

ahrav: Lets go with `{78,80}`, that should hopefully catch both types of tokens.

fumblehool:
@ahrav Updated 👍 
Please merge 😄

ahrav (approved):

PR #2018, opened 2023-10-28
Detector-Competition-Feat: Added Stripo API token detector

fumblehool:
### Description:
Add Detector for Stripo (stripo.com)

### Links:
Docs: https://api.stripo.email/reference/findemails

### Checklist:

### Verification:

Screenshot 2023-10-28 at 2 54 02 PM


pkg/detectors/stripo/stripo.go
```diff
@@ -0,0 +1,88 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// JWT token
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"stripo"}) + `\b(eyJhbGciOiJIUzI1NiJ9\.[0-9A-Za-z]{130}\.[0-9A-Z-a-z\-_]{43})\b`)
```
ahrav: nit: Just for future regexes I noticed that when you create this pattern `[0-9A-Z-a-z\-_]` it includes an extra `-` and a redundant escape. We can just get away with `[0-9A-Za-z-_]`. No need to adjust this one, I can update it as I merge in main and resolve the conflicts, just for future reference.

ahrav (approved):
LGTM! thanks again @fumblehool

pkg/detectors/stripo/stripo.go
rgmz: I'd recommend either doing `[0-9A-Z-az\-_]` or `[0-9A-Za-z_-]` out of an abundance of caution. Putting just `-` between characters can inadvertently create a range. For instance, `[a-z0-9_-~]` matches any symbol with a character value between `_` (95) and `~` (126).

PR #2024, opened 2023-10-28
Detector-Competition-Feat: Added Ngrok API token detector

fumblehool:
### Description:
PR for ngrok detector

### Docs:
https://ngrok.com/docs/api/resources/agent-ingresses/#list-agent-ingresses

### Checklist:

### Verification:

Screenshot 2023-10-28 at 9 29 56 PM


pkg/detectors/ngrok/ngrok.go
```diff
@@ -0,0 +1,87 @@
...
+	client *http.Client
+}
+
+
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"ngrok"}) + `\b(2XOm[0-9A-Za-z]{23}_[0-9A-Za-z-_]{21})\b`)
```
ahrav, line 25: question: Can you confirm that tokens should begin with 2X0m? I created a handful of test tokens and they all began with `2XQ` but the similarity ended after that. Also i'm not sure the second portion of characters after the `_` contain `-` or `_`. All the tokens I created were only alphanumeric. Does your token contain those characters?

I was able to get the tests to pass using the following regex: `\b(2XQ[0-9A-Za-z]{24}_[0-9A-Za-z]{21})\b`

ahrav (dismissed):
Couple questions around the regex, otherwise this LGTM.

pkg/detectors/ngrok/ngrok.go
ahrav: Actually, i think the regex can be a little more specific. It appears like the token begins with 2XQ[0-9] so:
`\b(2XQ[0-9][0-9A-Za-z]{23}_[0-9A-Za-z]{21})\b` should work.

rgmz: It seems that the token prefix is not static. My existing authtoken started with `2TXSe...`, however, after refreshing it the prefixes were `2XQMsu`->`2XQMuJ`->`2XQN4s`->`2XQN62`->`2XQN7`. If I had to guess, there is some sort of globally incrementing ID, similar to [how Reddit IDs work](https://old.reddit.com/r/redditdev/comments/n624n/submission_ids_question/).

As far as I can tell, the "best" pattern would be something like:
```
# Every token I've generated has had a digit as the first character of the second segment. This may have been a coincidence.
\b2[a-zA-Z0-9]{26}_\d[a-zA-Z0-9]{20}\b
```

fumblehool:
Thanks for the review, I've updated the regex 😄

zricethezav (approved):

zricethezav (approved):

PR #2032, opened 2023-10-29
Competition-Detector-Fix: fixed zendesk token regex

ankushgoel27:
### Description:
Removed the prefix regex "zendesk" from the token regex as it was hindering with detection. Sometimes, the zendesk keyword will not be in the 40 char range. Either way, it shouldn't matter removing it because there are 3 regexes in place for this to work so no chance for false positives. Also the domain part contains the word "zendesk".

I have seen in the wild where this detector didn't work but valid credentials were present. 
zendesk

### Checklist:

### DIFF
--- a/pkg/detectors/zendeskapi/zendeskapi.go
+++ b/pkg/detectors/zendeskapi/zendeskapi.go
@@ -21,8 +21,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	token  = regexp.MustCompile(detectors.PrefixRegex([]string{"zendesk"}) + `([A-Za-z0-9_-]{40})`)
-	email  = regexp.MustCompile(`\b([a-zA-Z-0-9-]{5,16}\@[a-zA-Z-0-9]{4,16}\.[a-zA-Z-0-9]{3,6})\b`)
+	token  = regexp.MustCompile(`([A-Za-z0-9_-]{40})`)
+	email  = regexp.MustCompile(`\b(\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,7}\b)\b`)
 	domain = regexp.MustCompile(`\b([a-zA-Z-0-9]{3,16}\.zendesk\.com)\b`)
 )
 


pkg/detectors/zendeskapi/zendeskapi.go
```diff
@@ -21,8 +21,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	token  = regexp.MustCompile(detectors.PrefixRegex([]string{"zendesk"}) + `([A-Za-z0-9_-]{40})`)
-	email  = regexp.MustCompile(`\b([a-zA-Z-0-9-]{5,16}\@[a-zA-Z-0-9]{4,16}\.[a-zA-Z-0-9]{3,6})\b`)
+	token  = regexp.MustCompile(`([A-Za-z0-9_-]{40})`)
+	email  = regexp.MustCompile(`\b(\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,7}\b)\b`)
```
zricethezav, line 25: @ankushgoel27 any reason why we need back to back `\b`s here?

rgmz: Also, why `[A-Z|a-z]`? Is that meant to be uppercase OR lowercase `(A-Z|a-z)`? Regardless, I think `[A-Za-z]` would be fine.

ankushgoel27: Fixed email regex. I believe it's ready for merge.
Thanks for pointing it out.

pkg/detectors/zendeskapi/zendeskapi.go
```diff
@@ -21,8 +21,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	token  = regexp.MustCompile(detectors.PrefixRegex([]string{"zendesk"}) + `([A-Za-z0-9_-]{40})`)
-	email  = regexp.MustCompile(`\b([a-zA-Z-0-9-]{5,16}\@[a-zA-Z-0-9]{4,16}\.[a-zA-Z-0-9]{3,6})\b`)
+	token  = regexp.MustCompile(`([A-Za-z0-9_-]{40})`)
+	email  = regexp.MustCompile(`\b([a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})\b`)
```
zricethezav, line 25: @ankushgoel27 I think the domain range should have an upper bound. Something like `[a-zA-Z]{2,10}`.

I'm removing the label as there is no material fix here, it's just expanding the regex capture. This change could be applied to lots of our detectors.

ankushgoel27: Hi @zricethezav, the min chars for the subdomain part are 3 when creating a new account on zendesk. so i think 3,16 is still good. it does have upper bound - `[a-zA-Z-0-9]{3,16}\.zendesk\.com`
Also, the current implementation of email regex is broken. its very weak and it missed valid credentials. So i think it should be included in the competition. the current implementation of email doesn't capture - zach@custom.com

pkg/detectors/zendeskapi/zendeskapi.go
ankushgoel27: @rgmz can we merge this please

rgmz: I'm just a random guy.

pkg/detectors/zendeskapi/zendeskapi.go
zricethezav: @ankushgoel27 
In effort to reduce false positives, `detectors.PrefixRegex([]string{"zendesk"})` should be added back in. The email local section needs a range as does the email domain. Something like: 
```
([a-zA-Z0-9._%+-]{3,30}@[a-zA-Z0-9.-]{3,100}\.[a-zA-Z]{2,10})
```

pkg/detectors/zendeskapi/zendeskapi.go
ankushgoel27: @zricethezav can we merge this please

pkg/detectors/zendeskapi/zendeskapi.go
ankushgoel27: hi zach, i removed the zendesk prefix because the domain part already contains the zendesk string. i believe its unnecessary to include the zendesk string in reach regex since one occurrence is enough.
I saw valid live credentials that were missed because of this because the email or token doesn't contain the zendesk string. adding the string will create lots of false negatives.

pkg/detectors/zendeskapi/zendeskapi.go
ankushgoel27: @zricethezav can we merge this please

PR #2033, opened 2023-10-29
Competition-Detector-Fix: fixed Clickup regex

ankushgoel27:
### Description:
I have seen a valid token in the wild where pk_[0-9] have 7 digits instead of 8. So modified the regex to include the same. Also, i don't see a need to have a prefix of "clickup" as the token format is very unique.

### Checklist:


zricethezav:
> I have seen a valid token in the wild where pk_[0-9] have 7 digits instead of 8. So modified the regex to include the same. Also, i don't see a need to have a prefix of "clickup" as the token format is very unique.

I haven't been able to generate tokens with 7 digits. Unless you can provide an example of one with a token of 7 digits via a private channel I don't see this as a fix. Additionally the `pk_`  prefix is used by a couple different providers so it's not unique. Closing this PR. If you can provide proof that 7 digits is used then we can reopen and consider this for hacktoberfest points.

rgmz:
>  Additionally the pk_ prefix is used by a couple different providers so it's not unique. 

Wouldn't the keywords filter out chunks that don't contain `clickup`?

https://github.com/trufflesecurity/trufflehog/blob/4a44a746a2d4d886994d535352493210ba6458e0/pkg/detectors/clickuppersonaltoken/clickuppersonaltoken.go#L28-L30

zricethezav:
> Wouldn't the keywords filter out chunks that don't contain clickup?

Yes it would.

@ankushgoel27 reopening as we confirmed `{7,8}` is an acceptable range

zricethezav (approved):

PR #2035, opened 2023-10-29
increased prefix regex length

ankushgoel27:
### Description:

i have seen many cases where a detector doesn't work because the prefix word is beyond 40 chars so increasing the range makes sense to me. Increasing it from 40 chars to 60 chars

### Checklist:


rgmz:
See also: https://github.com/trufflesecurity/trufflehog/pull/1923#discussion_r1365729811

PR #2037, opened 2023-10-29
adding 'token' keyword to regex for github_old

ankushgoel27:
### Description:
Seen the keyword token in the wild for the old format of GitHub token so adding that to the regex.

### Checklist:

### DIFF
--- a/pkg/detectors/github_old/github_old.go
+++ b/pkg/detectors/github_old/github_old.go
@@ -43,7 +43,7 @@ type userRes struct {
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"github", "gh", "pat"}
+	return []string{"github", "gh", "pat", "token"}
 }
 
 // FromData will find and optionally verify GitHub secrets in a given set of bytes.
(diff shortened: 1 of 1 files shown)


rgmz:
This might lead to more false-positives, but I support it. I've seen numerous examples of GitHub tokens defined as `vcs_token`, `git_token`, `api_token`, or just `token`, most of which wouldn't be picked up by the current detector.

pkg/detectors/github_old/github_old.go
```diff
@@ -25,7 +25,7 @@ func (Scanner) DefaultEndpoint() string { return "https://api.github.com" }
 var (
 	// Oauth token
 	// https://developer.github.com/v3/#oauth2-token-sent-in-a-header
-	keyPat = regexp.MustCompile(`(?i)(?:github|gh|pat)[^\.].{0,40}[ =:'"]+([a-f0-9]{40})\b`)
+	keyPat = regexp.MustCompile(`(?i)(?:github|gh|pat|token)[^\.].{0,40}[ =:'"]+([a-f0-9]{40})\b`)
```
ahrav, line 28: suggestion: We might also need it to be added to the list of keywords below on L46. Otherwise i'm not sure this will have the intended behavior.

ahrav (approved):

pkg/detectors/github_old/github_old.go
ankushgoel27: i made the changes you mentioned

ankushgoel27:
can this be considered for the competition?

rgmz:
This helped me find a valid token that otherwise would not have been detected. Great change, @ankushgoel27. :)

```
apiVersion: v1
data:
  token: "***"
kind: Secret
metadata:
  name: github-token
```

PR #1912, opened 2023-10-18
Handle secondary GitHub ratelimits

rgmz:
### Description:

This fixes #1911, which turned out to be caused by [GitHub's secondary rate limit](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#secondary-rate-limits).

```sh
# debug print statements removed in the final commit
$ ./trufflehog github --org coinbase --issue-comments --token=github_pat_xxx --only-verified
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2023-10-18T14:06:11-04:00       info-0  trufflehog      Completed enumeration   {"num_repos": 109, "num_orgs": 1, "num_members": 0}
2023-10-18T14:06:12-04:00       info-0  trufflehog      scanning github issue comments  {"repository": "https://github.com/coinbase/code-of-conduct.git"}
2023-10-18T14:06:12-04:00       info-0  trufflehog      scanning github issue comments  {"repository": "https://github.com/coinbase/.github.git"}
...
Waiting for: 1m0s
Waiting for: 1m0s
2023-10-18T14:06:25-04:00       info-0  trufflehog      scanning github issue comments  {"repository": "https://github.com/coinbase/rosetta-ethereum.git"}
Waiting for: 1m0s
2023-10-18T14:06:25-04:00       info-0  trufflehog      scanning github issue descriptions      {"repository": "https://github.com/coinbase/coinbase-pro-node.git"}
Waiting for: 1m0s
2023-10-18T14:06:25-04:00       info-0  trufflehog      scanning github issue comments  {"repository": "https://github.com/coinbase/rosetta-cli.git"}
Waiting for: 1m0s
Waiting for: 1m0s
Waiting for: 1m0s
Waiting for: 1m0s
2023-10-18T14:07:05-04:00       info-0  trufflehog      scanning github issue comments  {"repository": "https://github.com/coinbase/salus.git"}
2023-10-18T14:07:12-04:00       info-0  trufflehog      scanning github issue descriptions      {"repository": "https://github.com/coinbase/salus.git"}
...
```

It also contains a minor change where, instead of calling `s.visibilityOf()` in the individual processing methods, it is called once in `processRepoComments()` and passed that along in `repoInfo`.

(However, this doesn't actually reduce the number of API calls because `visibilityOf` caches results.)

### Checklist:
* [ ] Tests passing (`make test-community`)?
[...]

PR #2041, opened 2023-10-29
Fix handling of GitHub ratelimit information

rgmz:
### Description:
This is a follow-up to #1912, which used the headers from the response to determine rate-limiting information, instead of using the values from `RateLimitError.Rate`:

https://github.com/trufflesecurity/trufflehog/blob/5058cadc441296d1297f8531c16af4f5727e4a9b/pkg/sources/github/github.go#L838-L841

Although that logic seemed solid, I discovered that it did not work in some circumstances. This lead to the "unexpected" path more often than intended, and periodic instances where requests would be made before the ratelimit was refreshed.

https://github.com/trufflesecurity/trufflehog/blob/5058cadc441296d1297f8531c16af4f5727e4a9b/pkg/sources/github/github.go#L872

```
2023-10-29T17:28:25-04:00       error   trufflehog      Failed to parse ratelimit-remaining     {"error": "strconv.Atoi: parsing \"\": invalid syntax"}
2023-10-29T17:28:25-04:00       error   trufflehog      Failed to parse ratelimit-reset {"error": "strconv.Atoi: parsing \"\": invalid syntax"}
2023-10-29T17:28:08-04:00       error   trufflehog      unexpected rate limit error     {"retry_after": "5m0s", "retry_after_orig": "0s", "remaining": 0, "knownWait": false, "resume_time": "2023-10-29 17:33:08.06715002 -0400 EDT m=+6232.418916613", "headers": {}, "type": "*github.RateLimitError", "error": "GET https://api.github.com/repos/xxx/yyyl: 403 API rate limit of 5000 still exceeded until 2023-10-29 17:49:12 -0400 EDT, not making remote request. [rate reset in 21m04s]"}
```

### Checklist:


rgmz:
I will note that this can still result in workers concurrently hitting the API and failing:
```
2023-10-29T19:43:18-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m47.724523981s", "resume_time": "2023-10-29 19:50:06.000004355 -0400 EDT m=+520.724369060"}
2023-10-29T19:43:18-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m47.474734374s", "resume_time": "2023-10-29 19:50:06.000003887 -0400 EDT m=+520.724368589"}
2023-10-29T19:43:22-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m43.351695671s", "resume_time": "2023-10-29 19:50:06.000003626 -0400 EDT m=+520.724368327"}
2023-10-29T19:43:23-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m42.607366169s", "resume_time": "2023-10-29 19:50:06.000004116 -0400 EDT m=+520.724368819"}
2023-10-29T19:43:25-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m40.810664635s", "resume_time": "2023-10-29 19:50:06.000003248 -0400 EDT m=+520.724367948"}
2023-10-29T19:43:27-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m38.59667703s", "resume_time": "2023-10-29 19:50:06.000003904 -0400 EDT m=+520.724368599"}
2023-10-29T19:43:28-04:00       info-0  trufflehog      rate limited    {"retry_after": "6m37.454843428s", "resume_time": "2023-10-29 19:50:06.000003603 -0400 EDT m=+520.724368302"}
...
```

I don't know if you get penalized for trying to make requests when you've hit your rate-limit. It may be worth adding a mutex to prevent extraneous requests from being dispatched.

pkg/sources/github/github.go
```diff
@@ -850,25 +856,33 @@ func (s *Source) handleRateLimit(errIn error, res *github.Response) bool {
 	// https://docs.github.com/en/rest/overview/resources-in-the-rest-api#exceeding-the-rate-limit
 	if retryAfter <= 0 && res != nil {
 		var err error
-		remaining, err = strconv.Atoi(res.Header.Get("x-ratelimit-remaining"))
-		if err != nil {
-			knownWait = false
+
+		remainingHeader := res.Header.Get("x-ratelimit-remaining")
+		if remainingHeader != "" {
+			if remaining, err = strconv.Atoi(remainingHeader); err != nil {
```
rosecodym, line 862: If the parsing of this header (or the other one) fails, do you think it's worth logging out the unparseable value?

rosecodym (commented):
Thanks for digging into this, it looks like gnarly stuff. The changes seem non-obvious and I want to ensure I've got them right. Tell me if this is correct:

There are two functional changes here. The first applies to when we're handling an error that turns out to be a primary rate limit error with a `Remaining` property that's in the future and no requests remaining. Before, we'd ignore `Remaining` and try to parse the rate limit response headers ourselves, but now we ignore the headers and just wait until that specified time.

The second case is in the header parsing logic. Before, if `x-ratelimit-remaining` was missing, we'd fall into the "unexpected" case. That behavior is no longer present - if the header is missing, we still attempt to process `x-ratelimit-reset`. (The code to parse _that_ header has been changed to match the other one, but the effective logic hasn't changed. If the retry interval from the original error is non-positive, but `x-ratelimit-reset` is missing, we still fall into the "unexpected" case because we're supposed to wait but we can't figure out how long to wait for.)

There is also the change of moving `Now()` to the start of the function - was that an important logic fix or just a nice cleanup you did while you were here?

pkg/sources/github/github.go
```diff
@@ -850,25 +856,33 @@ func (s *Source) handleRateLimit(errIn error, res *github.Response) bool {
...
 
-	resumeTime := time.Now().Add(retryAfter).String()
+	var resumeTime string
 	if knownWait && remaining == 0 && retryAfter > 0 {
+		resumeTime = now.Add(retryAfter).String()
 		s.log.V(2).Info("rate limited", "retry_after", retryAfter.String(), "resume_time", resumeTime)
 	} else {
-		// TODO: Use exponential backoff instead of static retry time.
 		retryAfter = time.Minute * 5
+		resumeTime = now.Add(retryAfter).String()
```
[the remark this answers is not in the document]
rgmz: The previous logic would print the incorrect `resumeTime` in the `else` condition. I suppose it doesn't need to be assigned in both blocks, only the `else` one.

```golang
resumeTime := time.Now().Add(retryAfter).String() // retryAfter is 1 minute, resumeTime is now+1m
if knownWait && remaining == 0 && retryAfter > 0 {
  // ...
} else {
  retryAfter = time.Minute * 5 // retryAfter is now 5 minutes, resumeTime is now+1m
  s.log.V(2).Error(errIn, "unexpected rate limit error", "retry_after", retryAfter.String(), "resume_time", resumeTime)
}
```

pkg/sources/github/github.go
rgmz: I don't think so. The only failure I've encountered in testing is when the header is empty (`""`), which should no longer be an issue.

rgmz:
Coming back to this after a while, I went on a bit of a wild goose chase because it started returning a generic `github.ErrorResponse`. It turns out that GitHub changed their documentation URL which broke the response handling (https://github.com/google/go-github/commit/182859f01734d19cfa11a09dccede8864820b344).

I will push a change shortly that includes upgrading `go-github`.

rgmz:
> The second case is in the header parsing logic...

After fixing the aforementioned issue and experimenting with this further, I don't think there's a point to keeping this part. The `go-github` package should always translate response into a usable error, and if it can't then whatever the issue would likely impact us as well.

---

I've refactored the logic to use a `sync.RWMutex`. The goal is to make the logs more user-friendly/useful by only logging one message per given rate limit event. Any feedback or potential issues with this approach are welcome.

```
2023-12-10T15:01:29-05:00       info-0  trufflehog      rate limited    {"resume_time": "2023-12-10 15:02:37.14123656 -0500 EST m=+796.659151165"}
2023-12-10T15:03:12-05:00       info-0  trufflehog      rate limited    {"resume_time": "2023-12-10 15:04:20.218740308 -0500 EST m=+899.736654911"}
2023-12-10T15:04:43-05:00       info-0  trufflehog      rate limited    {"resume_time": "2023-12-10 15:05:51.38309714 -0500 EST m=+990.901011742"}
2023-12-10T15:06:25-05:00       info-0  trufflehog      rate limited    {"resume_time": "2023-12-10 15:07:30.166662804 -0500 EST m=+1089.684577403"}
```

> There is also the change of moving Now() to the start of the function - was that an important logic fix or just a nice cleanup you did while you were here?

Calling `time.Now()` in multiple places bugged me and also resulted in minor inconsistencies. It's for stylistic and consistency purposes (albeit mostly the former).

pkg/sources/github/github.go
```diff
@@ -824,53 +825,70 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
+			rateLimitMu.Unlock()
+			return false
 		}
 
-		resetTime, err := strconv.Atoi(res.Header.Get("x-ratelimit-reset"))
-		if err != nil || resetTime == 0 {
-			knownWait = false
-		} else if resetTime > 0 {
-			retryAfter = time.Duration(int64(resetTime)-time.Now().Unix()) * time.Second
+		jitter := time.Duration(rand.Intn(10)+1) * time.Second
```
rgmz, line 874: GitHub's API is buggy and will periodically reject requests that should have succeeded.
```
GET https://api.github.com/repos/org/repo: 403 API rate limit exceeded for user ID 32133502. If you reach out to GitHub Support for help, please include the request ID C8F0:12C9:34AADD1:6CC2F9A:653FC128. [rate limit was reset 0s ago]
GET https://api.github.com/repos/org/repo: 403 API rate limit exceeded for user ID 32133502. If you reach out to GitHub Support for help, please include the request ID C8F0:12C9:34AAE96:6CC3126:653FC128. [rate limit was reset 1s ago]
```

My solution to this is to add jitter, however...

1. I'm not sure if this is the best way to accomplish that (this specific implementation may be flawed)
2. It might be best to calculate jitter dynamically before calling `time.Sleep`

pkg/sources/github/github.go
```diff
@@ -824,53 +825,70 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
+		if retryAfter > 0 {
+			retryAfter = retryAfter + jitter
+			rateLimitResumeTime = now.Add(retryAfter)
+			s.log.V(0).Info(fmt.Sprintf("exceeded %s rate limit", limitType), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
+		} else {
+			retryAfter = (5 * time.Minute) + jitter
+			rateLimitResumeTime = now.Add(retryAfter)
```
rgmz, line 881: Repeating this definition is awkward but I wanted to cleanly differentiate between `retryAfter` being provided.

pkg/sources/github/github.go
```diff
@@ -951,11 +964,8 @@ func (s *Source) addAllVisibleOrgs(ctx context.Context) {
 		},
 	}
 	for {
-		orgs, resp, err := s.apiClient.Organizations.ListAll(ctx, orgOpts)
-		if err == nil {
-			resp.Body.Close()
-		}
```
rgmz, line 957: Closing the response body isn't necessary because `go-github` does it for us.

https://github.com/google/go-github/blob/c135be9459ed40979be841f065badb7de19d467b/github/github.go#L902

pkg/sources/github/github.go
```diff
@@ -824,53 +823,70 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
-	} else if errors.As(errIn, &abuseLimit) {
-		retryAfter = abuseLimit.GetRetryAfter()
-	} else {
+// handleRateLimit returns true if a rate limit was handled
+//
+// Unauthenticated users have a rate limit of 60 requests per hour.
+// Authenticated users have a rate limit of 5,000 requests per hour,
+// however, certain actions are subject to a stricter "secondary" limit.
+// https://docs.github.com/en/rest/overview/rate-limits-for-the-rest-api
+func (s *Source) handleRateLimit(errIn error) bool {
```
rgmz, line 837: The response is no longer necessary if we aren't attempting to parse any of the headers.

pkg/sources/github/github.go
rgmz: The specific message `rate limit was reset 1s ago` [comes from the go-github package](https://github.com/google/go-github/blob/2cc47f9fa6b4d4997260c590388eecd19045c145/github/github.go#L1495-L1497) but I'm not sure if this is a bug with go-github. That error indicates that a 403 was received from GitHub with the header `x-remaining: 0`.

I suspect that it's a rare issue with GitHub's API that most people don't encounter because they don't frequently exhaust their ratelimit. In typical circumstances you'd have to wait an hour before resuming, however, scanning issues/prs triggers the secondary rate limit every minute.

In any case, adding a bit of jitter before retrying is probably something we should be doing regardless.

ahrav (approved):
This looks good to me, thanks for taking the time to get this all cleaned up and updated.

rgmz:
@ahrav I've resolved the conflicts.

pkg/sources/github/github.go
```diff
@@ -869,53 +869,70 @@ func (s *Source) cloneAndScanRepo(ctx context.Context, client *github.Client, re
...
-		remaining, err = strconv.Atoi(res.Header.Get("x-ratelimit-remaining"))
-		if err != nil {
-			knownWait = false
+	rateLimitMu.RLock()
+	resumeTime := rateLimitResumeTime
+	rateLimitMu.RUnlock()
+
+	var retryAfter time.Duration
+	if resumeTime.IsZero() || time.Now().After(resumeTime) {
+		rateLimitMu.Lock()
```
rosecodym, line 894: It's possible for a bunch of workers to get blocked here, right? And if they entered this function because of the secondary rate limit, won't they therefore all end up waiting for their rate limit period _in sequence_?

I'm having trouble juggling these locks in my head so I might be completely misunderstanding this. (Or maybe this is a theoretically possibility that's unlikely in practice?)

rosecodym (commented):
I have a question about the currency, and unrelatedly, are any of these linting errors low-hanging fruit you can fix? If not, would you mind line-level disabling them using `// staticcheck:nolint` and opening a new issue to clean them up?

rgmz:
I guess bumping the go-github version resulted in several deprecations. I can take a look at this later. 😅

pkg/sources/github/github.go
rgmz: While it's possible that multiple workers will hit this, the actual work performed in this block is minimal and the lock should be released fairly quickly. `time.Sleep` is called after `rateLimitMu.Unlock()`.

https://github.com/trufflesecurity/trufflehog/blob/7bef8df565aac02575208595fced48551e54c0b7/pkg/sources/github/github.go#L930-L936

rosecodym (approved):
@ahrav i know you already approved this, but is it going to conflict with anything you're currently working on?

ahrav:
> @ahrav i know you already approved this, but is it going to conflict with anything you're currently working on?

We should be good. Thanks for double checking 🙇

PR #2050, opened 2023-10-30
Competition-Detector-Fix: Salesforce - fixed regex and protocol change

ankushgoel27:
### Description:
Seen instance url without "my" in the name so removed that. It shouldn't matter because [0-9a-zA-Z-\.]{1,100} should also cover any instances of keyword "my" in the instance name.
Moved https:// protocol to the verifier so that the instance URL's can be captured even if protocol is not present.
Added RawV2 to print both instance url and token


ankushgoel27:
salesforce

ankushgoel27:
May I get a review on this. It's been almost a year

PR #2051, opened 2023-10-30
Competition-Detector-Fix: fixed verifier code - hubspot

ankushgoel27:
### Description:
IF the correct scope is not set on the API key, a 403 could be returned.

This is such a small change - not sure if its even worthy for the competition. If not, don't add the label to it.

Hubspot released a new API format and 2 methods to authenticate. Older format is still supported. I will submit the new v2 detector soon.

### Checklist:


zricethezav:
@ankushgoel27 rather than accepting a 403, could we use a different endpoint that always returns a 200 or 300 if a valid api key is present?

ankushgoel27:
> @ankushgoel27 rather than accepting a 403, could we use a different endpoint that always returns a 200 or 300 if a valid api key is present?

Well, if the key is invalid, a 401 is returned. 403 means, key is valid but scope doesn't contain this endpoint.
Not sure how to create a key without any scope, when i go into hubspot account, i need to select at least one scope to create a key. also,this detector is using the old format so cant create new keys using the old format to test. i had one laying around who gave 403 when i tested this endoint.

https://legacydocs.hubspot.com/docs/overview
https://developers.hubspot.com/docs/api/private-apps

zricethezav:
> also,this detector is using the old format so cant create new keys using the old format to test. i had one laying around who gave 403 when i tested this endoint.

Could be a case of a lapsed TTL key. Gonna close this as the evidence isn't sufficient imo

ankushgoel27:
i have enough evidence here -

hubspot


As you can see with the error msg. if the token is not in scope, i get a 403 and if the token is invalid, i get a 401 error msg

PR #2056, opened 2023-10-30
Competition-detector-fix: fixed helpscout detector regex and verifier

ankushgoel27:
### Description:
I created a helpscout account and regenerated token many times and it always gave a [a-z0-9]{40} char long token and not the [A-Za-z0-9]{56} one so i replaced the regex. Maybe that was old format and not valid anymore. i have no way to check if the old format style tokens are supported or not. if they are, we can have a OR style regex match or i can close this PR and create a V2 style detector for the helpscout.

Verifier:
https://developer.helpscout.com/docs-api/

It suggests a basic auth-style API call so added that logic as well
helpscout

### Checklist:

### DIFF
--- a/pkg/detectors/helpscout/helpscout.go
+++ b/pkg/detectors/helpscout/helpscout.go
@@ -2,7 +2,6 @@ package helpscout
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


zricethezav (approved):

zricethezav:
confirmed token change

PR #2058, opened 2023-10-30
Refactor Git source to use SourceUnits

mcastorina:
### Description:
This led to a lot of changes to replace code that expects a channel with
a ChunkReporter. Luckily, a reporter can have the same functionality as
a channel, so this commit also adds a ChanReporter to maintain
functionality for the other codepaths.

### Checklist:

### DIFF
--- a/pkg/handlers/handlers.go
+++ b/pkg/handlers/handlers.go
@@ -37,7 +37,7 @@ type Handler interface {
 // packages them in the provided chunk skeleton, and sends them to chunksChan.
 // The function returns true if processing was successful and false otherwise.
 // Context is used for cancellation, and the caller is responsible for canceling it if needed.
-func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, chunksChan chan *sources.Chunk) bool {
+func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, reporter sources.ChunkReporter) bool {
 	aCtx := logContext.AddLogger(ctx)
 	for _, h := range DefaultHandlers() {
 		h.New()
@@ -51,7 +51,7 @@ func HandleFile(ctx context.Context, file io.Reader, chunkSkel *sources.Chunk, c
 			return false
 		}
 
-		if success := processHandler(aCtx, h, reReader, chunkSkel, chunksChan); success {
+		if success := processHandler(aCtx, h, reReader, chunkSkel, reporter); success {
 			return true
 		}
 	}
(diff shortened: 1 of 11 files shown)


pkg/handlers/handlers.go
```diff
@@ -37,7 +37,7 @@ type Handler interface {
 // packages them in the provided chunk skeleton, and sends them to chunksChan.
```
rosecodym, line 37: comment needs an update

pkg/sources/git/git.go
```diff
@@ -174,111 +177,98 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 	totalRepos := len(s.conn.Repositories) + len(s.conn.Directories)
-	// TODO: refactor to remove duplicate code
+	for i, repoURI := range s.conn.Repositories {
+		s.SetProgressComplete(i, totalRepos, fmt.Sprintf("Repo: %s", repoURI), "")
+		if len(repoURI) == 0 {
+			continue
+		}
+		if err := s.scanRepo(ctx, repoURI, reporter); err != nil {
+			ctx.Logger().Info("error scanning repository", "repo", repoURI, "error", err)
+			continue
```
rosecodym, line 192: is this necessary or is it just here for symmetry?

rosecodym (commented):
ok i'm seeing three parts of this pr, tell me if i've got this right:
* reorganizing the git source
* modifying every source _except_ the git source to use a new channel-wrapping reporter
* modifying the git source to use the new units system

if i do have that right, can i talk you into breaking this into two or even three prs?
1. reorganize the git source
1. create and use `ChanReporter` _everywhere_
1. modify the git source to use units

or

1. create and use `ChanReporter` _everywhere_
1. reorganize the git source and modify it to use units

mcastorina:
> ok i'm seeing three parts of this pr, tell me if i've got this right:
> * reorganizing the git source
> * modifying every source _except_ the git source to use a new channel-wrapping reporter
> * modifying the git source to use the new units system

For the most part, yes, but the `git` source is also using the channel wrapping reporter in the `Chunks` method.


> if i do have that right, can i talk you into breaking this into two or even three prs?
> 
> 1. reorganize the git source
> 2. create and use `ChanReporter` _everywhere_
> 3. modify the git source to use units
> 
> or
> 
> 1. create and use `ChanReporter` _everywhere_
> 2. reorganize the git source and modify it to use units

Yeah, good idea. The latter two PR approach makes more sense to me, so I'll go with that.

pkg/sources/git/git.go
mcastorina: Not necessary - it's similar to the `return` I did at the end of a function. No functional difference, but marginally better if we add code after it.

mcastorina:
The changes in this PR have been split up into #2082 and #2083

PR #2059, opened 2023-10-30
[chore] Fix SourceManager flaky test

mcastorina:
### Description:
Sorting by EndTime is not deterministic, however sorting by StartTime should be. StartTime is set in a goroutine that's limited by WithConcurrentUnits, so it should happen in order that the units are received.

### Checklist:


rosecodym (approved):
do the `unitChunk` objects forward anything even more deterministic to the metrics that you could use?

mcastorina:
> do the `unitChunk` objects forward anything even more deterministic to the metrics that you could use?

The unit ID is part of the metrics. Do you think checking the start time is too strict of a test?

PR #2061, opened 2023-10-30
[chore] - add binutils dep to dockerfile

ahrav:
### Description:
This allows us to correctly extract some archive file types on unix-like systems.

### Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN  --mount=type=cache,target=/go/pkg/mod \
      GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o trufflehog .
 
 FROM alpine:3.18
-RUN apk add --no-cache bash git openssh-client ca-certificates rpm2cpio binutils \
+RUN apk add --no-cache bash git openssh-client ca-certificates rpm2cpio binutils cpio \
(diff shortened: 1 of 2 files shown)


rgmz:
Is it possible to warn or fail if these dependencies aren't present? I didn't have `rpm2cpio` installed on my machine and didn't realize this was an issue until I increased the log verbosity.

ahrav:
> Is it possible to warn or fail if these dependencies aren't present? I didn't have `rpm2cpio` installed on my machine and didn't realize this was an issue until I increased the log verbosity.

I thought we did return an error [here](https://github.com/trufflesecurity/trufflehog/blob/1cf74b1fff14f10a05e285da58e984e8ddc86e29/pkg/handlers/archive.go#L262-L262). Wondering if it's worth setting this up in a way such that if a user doesn't want to install those deps, we still scan everything else? Or do we want to make it a required dep?

rgmz:
> thought we did return an error [here](https://github.com/trufflesecurity/trufflehog/blob/1cf74b1fff14f10a05e285da58e984e8ddc86e29/pkg/handlers/archive.go#L262-L262).

It's possible, but I swear I've never seen that error.

dustin-decker (approved):

PR #2062, opened 2023-10-30
use rawv2 for pubnubpublish

rosecodym:
### Description:
We're seeing secrets of this type flap between verified and unverified, which is expected behavior for multipart secrets without `RawV2` defined. This PR adds `RawV2` for secrets of this type.

### Checklist:

### DIFF
--- a/pkg/detectors/pubnubpublishkey/pubnubpublishkey.go
+++ b/pkg/detectors/pubnubpublishkey/pubnubpublishkey.go
@@ -55,6 +55,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_PubNubPublishKey,
 				Raw:          []byte(resMatch),
+				RawV2:        []byte(resMatch + "/" + ressubMatch),
 			}
 
 			if verify {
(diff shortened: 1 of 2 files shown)


pkg/detectors/pubnubpublishkey/pubnubpublishkey.go
```diff
@@ -55,6 +55,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_PubNubPublishKey,
 				Raw:          []byte(resMatch),
+				RawV2:        []byte(resMatch + "/" + ressubMatch),
```
rosecodym, line 58: Most multipart secret detectors just splice the two parts together, which seems unwise to me, so I used `/` as a separator because that's how the secrets are combined in the URL during verification. A few other detectors use `:` as a separator; if that's a convention we're trying to follow I'm happy to change it.

dustin-decker: For getting a password hash of the value it doesn't really matter, as long as we include all the parts, but if we ever wanted to parse the field it does.  I think the ship has sailed on making this a parseable information. What I wish we did was something like a map[string]string, which could be sorted, serialized, and hashed.

dustin-decker (approved):

PR #2063, opened 2023-10-30
Centralize logic for checking archive extraction tools

ahrav:
### Description:
Only check for archive extraction tools once during init. Reuse the cache during scans to prevent unnecessary filesystem lookups.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -268,7 +268,7 @@ func ensureToolsForMimeType(mimeType mimeType) error {
 	}
 
 	for _, tool := range tools {
-		if installed, ok := extractToolCache[tool]; !ok || !installed {
+		if installed := extractToolCache[tool]; !installed {
 			return fmt.Errorf("required tool %s is not installed", tool)
 		}
 	}
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -251,11 +251,12 @@ var mimeTools = map[mimeType][]string{
 var extractToolCache map[string]bool
 
 func init() {
-	extractToolCache = make(map[string]bool, len(mimeTools))
 	// Preload the extractToolCache with the availability status of each required tool.
+	extractToolCache = make(map[string]bool)
 	for _, tools := range mimeTools {
 		for _, tool := range tools {
-			extractToolCache[tool] = isToolInstalled(tool)
+			_, err := exec.LookPath(tool)
+			extractToolCache[tool] = err == nil
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/handlers/archive.go
```diff
@@ -233,39 +233,60 @@ func (a *Archive) ReadToMax(ctx context.Context, reader io.Reader) (data []byte,
...
+func isToolInstalled(tool string) bool {
+	if installed, ok := extractToolCache[tool]; ok {
+		return installed
+	}
+
+	// If the tool isn't in the cache, check its availability and cache it.
+	_, err := exec.LookPath(tool)
+	installed := err == nil
+	// Cache the result to avoid future filesystem checks for this tool.
+	extractToolCache[tool] = installed
```
mcastorina, line 286: Writing to the map here makes this method not thread safe.

mcastorina (commented):
Are we using `isToolInstalled` for any tool besides the ones we already know about? I'm not sure it's worth the added complexity to support that use case.

pkg/handlers/archive.go
ahrav: But this is in the init method there is only a single thread? This is the only place writes occur, so I think this should be fine.

ahrav: or you mean during the run of the program?

mcastorina: Yeah, during the run of the program. Is the method only going to be ran from `init()`? In that case I think we should inline it or add a comment so no one tries to re-use it as it's not obvious it could panic.

rgmz: > In that case I think we should inline it or add a comment so no one tries to re-use it as it's not obvious it could panic.

Call it [`secretIsToolInstalledDoNotUseOrYouWillBeFired`](https://github.com/facebook/react/blob/80bff5397bf854750dbe7c286f61654ea58938c5/src/umd/ReactUMDEntry.js). :P

ahrav: done

pkg/handlers/archive.go
```diff
@@ -233,36 +233,45 @@ func (a *Archive) ReadToMax(ctx context.Context, reader io.Reader) (data []byte,
...
 	tools, exists := mimeTools[mimeType]
 	if !exists {
-		return fmt.Errorf("unsupported mime type")
+		return fmt.Errorf("unsupported mime type: %s", mimeType)
 	}
 
 	for _, tool := range tools {
-		if !isToolInstalled(tool) {
-			return fmt.Errorf("Required tool " + tool + " is not installed")
+		if installed, ok := extractToolCache[tool]; !ok || !installed {
```
mcastorina, line 271: **Nit:** `extractToolCache` value is `bool`, so we can use the default value:

```go
if installed := extractToolCache[tool]; !installed {
    // ...
}
```

mcastorina (approved):
Have you considered setting some configuration flag in `Archive` for control flow in `HandleSpecialized`, for example?

ahrav:
> Have you considered setting some configuration flag in `Archive` for control flow in `HandleSpecialized`, for example?

What were you thinking of? I haven't really thought too hard about this, other than noticing we checked if we had tools installed for every file 😓

mcastorina:
> What were you thinking of?

Something like..

```go
func init() {
    if !archiveToolsInstalled {
        disableArchiveHandler()
    }
}
```

or the inverse?

ahrav:
> > What were you thinking of?
> 
> Something like..
> 
> ```
> func init() {
>     if !archiveToolsInstalled {
>         disableArchiveHandler()
>     }
> }
> ```
> 
> or the inverse? I also haven't thought too deeply on it.

Ahhh yea I hadn't thought of that. I guess it would involve a little more logic b/c what I touched was a very specific part of the archive handler that relates only to .deb and .rpm files. The archive handler also handles all other archive file types out of the box using the archiver library. We just have to use `rpm2cpio` and `ar` to extract certain archives and then feed it into the archiver lib.

So your idea would work, we'd just have to exclude certain archive types but not necessarily all.

PR #2070, opened 2023-10-31
Competition-Detector-Fix: Improving GitHub oauth Detector

ankushgoel27:
### Description:
having  a prefix regex of 40 chars for keyword `github` is limited as with multiple regexes and the length of the secret itself, the prefix regex 40 char limit is not sufficient. So i added another keywords that are most common seen in the wild for the oauth config files. ClientID will have the keyword `ID` and clientsecret will have the keyword `secret`. Keyword `client` will also work for both the regexes.

### Checklist:


zricethezav:
This would introduce too many false positives on unverified scans

ankushgoel27:
> This would introduce too many false positives on unverified scans

@zricethezav Currently this is generating a lot of false negatives. Like i said before, 40 chars limit in prefixregex client secret and client id is used up so a match is not happening. 

I guess an alternate would be to improve the prefix regex logic and increase its range

PR #2077, opened 2023-10-31
[chore] - correctly handle input shorter than 512 bytes

ahrav:
### Description:
Check for EOF error as well as err != nil so that input shorter than 512 bytes is also correctly processed.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -418,7 +418,7 @@ func determineMimeType(reader io.Reader) (mimeType, io.Reader, error) {
 	// If fewer bytes are read, MIME type detection may still succeed.
 	buffer := make([]byte, 512)
 	n, err := reader.Read(buffer)
-	if err != nil && err != io.EOF {
+	if err != nil && !errors.Is(err, io.EOF) {
 		return "", nil, fmt.Errorf("unable to read file for MIME type detection: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/sources/chunker_test.go
```diff
@@ -7,8 +7,8 @@ import (
 	"testing"
 	"testing/iotest"
 
-	diskbufferreader "github.com/trufflesecurity/disk-buffer-reader"
 	"github.com/stretchr/testify/assert"
+	diskbufferreader "github.com/trufflesecurity/disk-buffer-reader"
```
ahrav, line 11: Note: not sure how/why this got updated, but i'm going to leave it since its just a import reorder.

pkg/handlers/archive.go
```diff
@@ -414,9 +414,11 @@ func (a *Archive) handleNestedFileMIME(ctx logContext.Context, tempEnv tempEnv,
 // determineMimeType reads from the provided reader to detect the MIME type.
 // It returns the detected MIME type and a new reader that includes the read portion.
 func determineMimeType(reader io.Reader) (mimeType, io.Reader, error) {
+	// A buffer of 512 bytes is used since many file formats store their magic numbers within the first 512 bytes.
+	// If fewer bytes are read, MIME type detection may still succeed.
 	buffer := make([]byte, 512)
 	n, err := reader.Read(buffer)
-	if err != nil {
+	if err != nil && err != io.EOF {
```
mcastorina, line 421: **Nit:** `error.Is(err, io.EOF)`

mcastorina (approved):

PR #2082, opened 2023-11-01
[chore] Replace chunks channel with ChunkReporter in git based sources

mcastorina:
### Description:
ChunkReporter is more flexible and will allow code reuse for unit chunking. ChanReporter was added as a way to maintain the original channel functionality, so this PR should not alter existing behavior.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -25,7 +25,6 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/cleantemp"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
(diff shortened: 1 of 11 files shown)


pkg/sources/git/git.go
```diff
@@ -552,13 +555,18 @@ func (s *Git) gitChunk(ctx context.Context, diff gitparse.Diff, fileName, email,
 					Data:           append([]byte{}, newChunkBuffer.Bytes()...),
 					Verify:         s.verify,
 				}
+				if err := reporter.ChunkOk(ctx, chunk); err != nil {
+					// TODO: Return error.
+					return
+				}
```
rosecodym, line 561: What can cause `ChunkOk` to return an error?

rosecodym (approved):
what a beautiful, boring pr

pkg/sources/git/git.go
mcastorina: Depends on implementation, but so far only context cancellations.

PR #2083, opened 2023-11-01
Refactor git source to support scanning units

mcastorina:
### Description:
Adds functions for scanning a single repo or directory as well as all the required methods for unit chunking.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -80,8 +80,11 @@ func NewGit(sourceType sourcespb.SourceType, jobID sources.JobID, sourceID sourc
 }
 
 // Ensure the Source satisfies the interfaces at compile time.
-var _ sources.Source = (*Source)(nil)
-var _ sources.SourceUnitUnmarshaller = (*Source)(nil)
+var _ interface {
+	sources.Source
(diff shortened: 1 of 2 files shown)


pkg/sources/git/git.go
```diff
@@ -80,8 +80,11 @@ func NewGit(sourceType sourcespb.SourceType, jobID sources.JobID, sourceID sourc
 }
 
 // Ensure the Source satisfies the interfaces at compile time.
-var _ sources.Source = (*Source)(nil)
-var _ sources.SourceUnitUnmarshaller = (*Source)(nil)
+var _ interface {
+	sources.Source
+	sources.SourceUnitEnumChunker
+	sources.SourceUnitUnmarshaller
+} = (*Source)(nil)
```
rosecodym, line 87: whoa i didn't know you could smoosh interfaces together like this

rosecodym (approved):

pkg/sources/git/git.go
mcastorina: It's an anonymous interface composed of other interfaces! Similar to `interface{}` (aka `any`) but with stuff in it.

PR #2087, opened 2023-11-02
added resource type mapping to extraData in AWS

joeleonjr:
### Description:
AWS IDs contain a 4 character prefix that map to a specific resource type. The mapping is located here: 
[https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids).

This PR adds a new field to the "ExtraData" attribute for AWS results. As long as the AWS ID contains a valid 4 character prefix, it will return the related resource type (verified or not).

### Checklist:


pkg/detectors/aws/aws.go
```diff
@@ -115,12 +131,19 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				Raw:          []byte(resIDMatch),
 				Redacted:     resIDMatch,
 				RawV2:        []byte(resIDMatch + resSecretMatch),
+				ExtraData: map[string]string{
+					"resource_type": resourceTypes[resIDMatch[0:4]],
```
ahrav, line 135: question: Instead of this slice indexing did you consider adding a capture group to the prefix and just using that directly. That should prevent any potential IOOR panics.
For example:
New regex - `idPat     = regexp.MustCompile(`\b(AKIA|ABIA|ACCA|ASIA)([0-9A-Z]{16})\b`)`
Then within the code around matching:
```
for _, idMatch := range idMatches {
		if len(idMatch) != 3 {
			continue
		}
		resIDMatch := strings.TrimSpace(idMatch[0])

		....

		for _, secretMatch := range secretMatches {
			if len(secretMatch) != 2 {
				continue
			}
			resSecretMatch := strings.TrimSpace(secretMatch[1])

			s1 := detectors.Result{
				DetectorType: detectorspb.DetectorType_AWS,
				Raw:          []byte(resIDMatch),
				Redacted:     resIDMatch,
				RawV2:        []byte(resIDMatch + resSecretMatch),
				ExtraData: map[string]string{
					"resource_type": resourceTypes[idMatch[1]],  // HERE WE JUST USE THE CAPTURE
				},
			}
...
```

pkg/detectors/aws/aws.go
```diff
@@ -115,12 +131,19 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				},
 			}
 
 			if verify {
 				verified, extraData, verificationErr := s.verifyMatch(ctx, resIDMatch, resSecretMatch, true)
 				s1.Verified = verified
-				s1.ExtraData = extraData
+				//Append the extraData to the existing ExtraData map.
+				// This will overwrite with the new verified values.
+				for k, v := range extraData {
```
ahrav, line 144: question: Did you consider passing `s1` or even just the `ExtraData` map to `verifyMatch` so we set the values when we are setting all the other ExtraData value within that function?

ahrav (approved):
Couple questions otherwise LGTM!

pkg/detectors/aws/aws.go
joeleonjr: We definitely could. I have another PR coming re: account number mapping. We can access AWS account numbers just based on the ID (AKIA*****) value. However, if we verify the secret, then we'll want to have the value from AWS override the locally computed one. (It should be correct the vast majority of the time, if not all the time, but just in case it's not, we can overwrite with the official number). For that reason, I though this would be simplest + involve changing fewer functions. If for style purposes, you'd like me to change it, I'm happy to.

ahrav: Nope, this is just fine. Thanks for the explanation.

pkg/detectors/aws/aws.go
joeleonjr: Just updated with latest commit.

PR #2090, opened 2023-11-03
[chore] Speedup IsKnownFalsePositive using sets

mcastorina:
### Description:
Update our false positive checks to use a set instead of iterating over a slice each time.
Also adds a check that the match is a valid UTF-8 string.

Benchmark before (worst case where `IsKnownFalsePositive` returns false):

```
» go test -tags=detectors -bench FalsePositive ./pkg/detectors
goos: darwin
goarch: arm64
pkg: github.com/trufflesecurity/trufflehog/v3/pkg/detectors
BenchmarkDefaultIsKnownFalsePositive-8             24123             49376 ns/op
PASS
ok      github.com/trufflesecurity/trufflehog/v3/pkg/detectors  1.922s
```

Benchmark after:

```
» go test -tags=detectors -bench FalsePositive ./pkg/detectors
goos: darwin
goarch: arm64
pkg: github.com/trufflesecurity/trufflehog/v3/pkg/detectors
BenchmarkDefaultIsKnownFalsePositive-8           5667696               201.6 ns/op
PASS
ok      github.com/trufflesecurity/trufflehog/v3/pkg/detectors  1.537s
```

### Checklist:

### DIFF
--- a/pkg/detectors/falsepositives.go
+++ b/pkg/detectors/falsepositives.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 )
 
 var DefaultFalsePositives = []FalsePositive{"example", "xxxxxx", "aaaaaa", "abcde", "00000", "sample", "www"}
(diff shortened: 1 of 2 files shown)


bill-rich (approved):
Very cool way to skip the append and checking each slice element!

PR #2097, opened 2023-11-06
Improve Azure Storage detection

rgmz:
### Description:
~This PR tightens up the regex for `AccountName`+`AccountKey` and fixes a potential runtime panic when constructing the request.~

This PR expands the the Azure storage detector to find many common configurations, not just connection strings. It also adds a bunch of unit tests.

### Checklist:


spydx:
I would love to see this reach a release. The existing detector, for me, missed many AzureStorage keys that this work here does detect.

pkg/detectors/azure_storage/storage.go
```diff
@@ -0,0 +1,190 @@
...
+			results = append(results, s1)
+		}
+	}
+
+	return results, nil
+}
+
+func (s Scanner) verifyMatch(ctx context.Context, client *http.Client, name string, key string, extraData map[string]string) (bool, error) {
+	// https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key
+	now := time.Now().UTC().Format(http.TimeFormat)
```
ahrav, line 130: TIL about `http.TimeFormat`.

pkg/detectors/azure_storage/storage.go
```diff
@@ -0,0 +1,190 @@
...
+	switch res.StatusCode {
+	case http.StatusOK:
+		// parse response
+		response := storageResponse{}
+		if err := xml.NewDecoder(res.Body).Decode(&response); err != nil {
+			return false, err
+		}
+
+		// update the extra data with container names only
+		var containerNames []string
```
ahrav, line 169: suggestion (optional): We can build the final value directly.
```
var b strings.Builder
for i, c := range response.Containers.Container {
	if i > 0 {
		b.WriteString(", ")
	}
	b.WriteString(c.Name)
}
extraData["container_names"] = b.String()
```

pkg/detectors/azure_storage/storage_integration_test.go
ahrav, line 1: I think this file uses a different package name. Should we update it to `azure_storage`?

pkg/engine/defaults/defaults.go
```diff
@@ -2,6 +2,7 @@ package defaults
 
 import (
 	"bytes"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors/azure_storage"
```
ahrav, line 5: nit: Can we move this import into the other group?

ahrav (approved):

pkg/detectors/azure_storage/storage_integration_test.go
rgmz: Yeah. This is technically an error but it doesn't surface because of the flag.

pkg/detectors/azure_storage/storage.go
rgmz: Good idea. Done.

rgmz:
@ahrav Done.

PR #2102, opened 2023-11-08
update protos so we can use the git source for CI

ahrav:
### Description:
These changes allow us to specify the fields necessary to scan Git in CI via a config.yaml.

### Checklist:


bill-rich (commented):
Is this so we can use `Chunks` rather than doing all the `ScanOptions` setup manually?

ahrav:
> Is this so we can use `Chunks` rather than doing all the `ScanOptions` setup manually?

That wasn't the initial intent but that would work. This was so that we can provide --since-commit and `--branch` options in a config.yaml for enterprise so we can use the `scan` subcommand vs the `git` subcommand. This then allows for results to be displayed in their dashboard as well. Given the `git` subcommand doesn't connect to the api, but `scan` does.

bill-rich (approved):

PR #2104, opened 2023-11-09
add proto fields for Git

ahrav:
### Description:
Add support for Git scanning configuration fields to the proto message.

### Checklist:


proto/sources.proto
```diff
@@ -172,6 +172,11 @@ message Git {
   string head = 6;
   string base = 7;
   bool bare = 8;
+  string include_paths_file = 9; // path to file containing newline separated list of paths
+  string exclude_paths_file = 10; // path to file containing newline separated list of paths
+  string exclude_globs = 11; // comma separated list of globs
+  int64  max_depth = 12;
+  string uri = 13; // repository URL. https://, file://, or ssh://
```
rosecodym, line 179: How does this field interact with `repeated string repositories`?

ahrav: The `repositories` message is specifically utilized by the enterprise version of our product, which allows for scanning multiple repositories simultaneously. On the other hand, the `uri` message is employed in both the CLI and CI environments for scanning a single resource. Additionally, in CI contexts, the `uri` message is compatible with various optional flags such as `--since-commit`, `--bare`, `--branch`, and others, enhancing its functionality in these scenarios.

In [this PR](https://github.com/trufflesecurity/trufflehog/pull/2105), my approach focused on streamlining the implementation to avoid complex additional code. I found that introducing a separate message facilitated this goal. If, however, we later determine that merging this with the repositories message leads to easier maintenance, we can revisit and make that change. Such a modification would require us to adjust some of the underlying logic in the source, which I initially intended to minimize.

rosecodym: Yeah, I can buy this as the best option. Would you mind explaining it in comments though? I can definitely see it confusing a future maintainer.

mcastorina (approved):

rosecodym (approved):

PR #2105, opened 2023-11-09
move all Git setup into Init method

ahrav:
### Description:
Remove all Git setup via the CLI into the Git source's Init method. This will allow the enterprise pipeline to reuse all the functionality currently available in OSS via the config file. This should also allow for CI scans using the `scan` cmd with a Git source with all the current flags. ex (--since-commit, --branch, etc)

### Checklist:


pkg/sources/git/git.go
```diff
@@ -45,9 +47,6 @@ type Source struct {
 	sources.Progress
 	conn        *sourcespb.Git
 	scanOptions *ScanOptions
-	// Kludge to preserve engine.ScanGit functionality which doesn't expect
-	// the scanning to clean up the directory.
-	preserveTempDirs bool
```
rosecodym, line 50: Is this removed because we never need to do it anymore?

rosecodym (commented):
i really like this; i'm only holding off on approving because of a question i have on the base branch pr

pkg/sources/git/git.go
ahrav: We do need to still do it, it's just now it's handled within the source [here](https://github.com/trufflesecurity/trufflehog/blob/dca78d25650e3c5d07350440f655426dbe7b0812/pkg/sources/git/git.go#L286-L286). We had the kludge in place b/c the caller in `main` used to handle the cleanup, but since i removed all that logic in main and in the engine, the source can correctly handle it.

ahrav: This should also hopefully more closely align with all the other functionality where the source is responsible for cleaning up any temp dirs.

rosecodym: oh, by "this" i meant the kludge. i should have been more clear!

ahrav: Right we don't need to do this anymore. It was there for just to preserve the weird behavior we had with the engine and the source. The source takes care of that functionality now.

mcastorina: Thanks for cleaning up my kludge. I'm happy to see it go!

mcastorina (approved):

PR #2106, opened 2023-11-09
[thog-1548] add auto redaction for verification errors

0x1:
### Description:
VerificationError has been updated to a private variable.
SetVerificationError is now the recommended way to set verification errors, and any sensitive values should be passed-in to be automatically redacted.
All detectors and tests previously using VerificationError has also been updated to use the new logic.

### Checklist:

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -3,12 +3,12 @@ package detectors
 import (
 	"context"
 	"crypto/rand"
-	"errors"
-	"fmt"
 	"math/big"
 	"net/url"
 	"strings"
 	"unicode"
+
+	"errors"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/source_metadatapb"
@@ -83,11 +83,12 @@ func redactSecrets(err error, secrets ...string) error {
 	for _, secret := range secrets {
 		errStr = strings.Replace(errStr, secret, "[REDACTED]", -1)
 	}
-	// return redactedError{err, errStr}
-	return fmt.Errorf("%s", errStr)
+	return errors.New(errStr)
 }
 
 // unwrapToLast returns the last error in the chain of errors.
+// This is added to exclude non-essential details (like URLs) for brevity and security.
+// Also helps us optimize performance in redaction and enhance log clarity.
 func unwrapToLast(err error) error {
 	for {
 		unwrapped := errors.Unwrap(err)
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -3,6 +3,8 @@ package detectors
 import (
 	"context"
 	"crypto/rand"
+	"errors"
+	"fmt"
 	"math/big"
 	"net/url"
 	"strings"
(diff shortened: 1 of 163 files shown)


pkg/detectors/detectors.go
```diff
@@ -59,7 +61,42 @@ type Result struct {
 
 	// This field should only be populated if the verification process itself failed in a way that provides no
 	// information about the verification status of the candidate secret, such as if the verification request timed out.
-	VerificationError error
+	verificationError error
+}
+
+// SetVerificationError is the only way to set a verification error. Any sensetive values should be passed-in as secrets to be redacted.
+func (r *Result) SetVerificationError(err error, secrets ...string) {
```
0x1, line 68: the main changes are here

pkg/detectors/detectors.go
```diff
@@ -59,7 +61,42 @@ type Result struct {
...
+
+// Public accessors for the fields could also be provided if needed.
+func (r *Result) VerificationError() error {
+	return r.verificationError
+}
+
+// redactSecrets replaces all instances of the given secrets with [REDACTED] in the error message.
+func redactSecrets(err error, secrets ...string) error {
+	lastErr := unwrapToLast(err)
+	errStr := lastErr.Error()
```
rosecodym, line 82: Since you're throwing away all the intermediate errors eventually anyway, could you just stringify `err` without worrying about unwrapping it yourself, and then do the string replacement? I'm sure what you've got works, but verifying that requires knowing the details of how error unwrapping works, whereas if you just stringify from the outset you don't need to think about any of that.

0x1: that's what i did first. but in testing this with @trufflesteeeve we realized that sometimes the errors are super long and contain information that isn't pertinent - URLs and such, for example. And since it's generally the last error that has the most value, this ensures the logs output is compact, reduces chances of errors containing secrets which helps with performance somewhat in redaction.

rosecodym: Ah, ok, that makes sense. Would you mind putting it in a code comment?

rosecodym (approved):
i didn't review each individual detector, because i assume that the changes are the same for all of them. it would be nice to examine test coverage, though - did you do that? it's hard because so many detector tests already don't pass but there's probably a way we can leverage the xml test report we use in CI to at least see if any detector tests that used to pass started failing.

0x1:
> i didn't review each individual detector, because i assume that the changes are the same for all of them. it would be nice to examine test coverage, though - did you do that? it's hard because so many detector tests already don't pass but there's probably a way we can leverage the xml test report we use in CI to at least see if any detector tests that used to pass started failing.

ya it's hard bc there could potentially be problems in tests that already don't pass bc of nonexistant tokens bc they panic at the top. but one crude measurement i did do is running the test and doing a line count on how many 'ok' and 'FAIL' we get. 

**main**: _494_ ok, _1646_ FAIL
**thog-1548**: _491_ ok, _1656_ FAIL

it's not perfect bc there are some duplicates, but overall pretty close. i'm still trying to find the remaining ones.

pkg/detectors/detectors.go
```diff
@@ -59,7 +61,42 @@ type Result struct {
...
+func unwrapToLast(err error) error {
+	for {
+		unwrapped := errors.Unwrap(err)
+		if unwrapped == nil {
+			// We've reached the last error in the chain
+			return err
+		}
+		err = unwrapped
+	}
 }
```
mcastorina, line 100: I'm curious if [errors.Cause](https://pkg.go.dev/github.com/pkg/errors#hdr-Retrieving_the_cause_of_an_error) could be used instead of this function?

> errors.Cause will recursively retrieve the topmost error that does not implement causer, which is assumed to be the original cause.

pkg/detectors/detectors.go
```diff
@@ -59,7 +61,42 @@ type Result struct {
...
+
+// redactSecrets replaces all instances of the given secrets with [REDACTED] in the error message.
+func redactSecrets(err error, secrets ...string) error {
+	lastErr := unwrapToLast(err)
+	errStr := lastErr.Error()
+	for _, secret := range secrets {
+		errStr = strings.Replace(errStr, secret, "[REDACTED]", -1)
+	}
+	// return redactedError{err, errStr}
+	return fmt.Errorf("%s", errStr)
```
mcastorina, line 87: **Nit (optional):** `errors.New(errStr)` would work too since we're not formatting the string.

mcastorina (approved):
nice!

pkg/detectors/detectors.go
0x1: **errors.Cause** still gives a verbose output for some errors:

**with Cause()**
image

**with unwrapToLast()**
image

mcastorina: Ah, thanks for looking into that

zricethezav:
@0x1 changes LGTM, mind resolving the AWS detector? then we can get this merged

PR #2108, opened 2023-11-10
adding postgres detector

dylanTruffle:
…password match

### Description:

### Checklist:


0x1:
`patterns_test.go` needs to be updated to handle the updated pattern. also need to add a testfile for the postgres detector.

0x1:
`patterns_test.go` needs to be changed to handle the updated pattern. also need to add a testfile for the postgres detector.

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,170 @@
...
+		if pgURL.User != nil {
+			username := pgURL.User.Username()
+			password, _ := pgURL.User.Password()
+			hostname := pgURL.Hostname()
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_Postgres,
+				Raw:          []byte(strings.Join([]string{hostname, username, password}, "\t")),
+			}
+            if verify {
+                verificationErr:= verifyPostgres(pgURL)
```
deckarep, line 104: Since `ctx` is provided within the `FromData`, maybe that should be passed along to the `verifyPostgres()` call?

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"context"
+    "database/sql"
+	"net/url"
+	"regexp"
+	"strings"
+    "fmt"
+    "time"
+
+    "github.com/trufflesecurity/trufflehog/v3/pkg/common"
```
deckarep, line 12: Looks like gofmt should be ran.

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,170 @@
...
+    if pgURL.User != nil {
+        username = pgURL.User.Username()
+        password, _ = pgURL.User.Password()
+    }
+    hostname := pgURL.Hostname()
+
+    // Handle custom port
+    port := pgURL.Port()
+    if port == "" {
+        port = "5432" // Default PostgreSQL port
```
deckarep, line 49: Nit: Utilize string constant for default port

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,170 @@
...
+    connStr := fmt.Sprintf("user=%s password=%s host=%s port=%s sslmode=%s", username, password, hostname, port, sslmode)
+
+    // Open a connection to the database
+    db, err := sql.Open("postgres", connStr)
+    if err != nil {
+        return err
+    }
+    defer db.Close()
+
+    ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
deckarep, line 69: Is there support for providing default time.Durations and allowing this to be configured at runtime with the cli?

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,170 @@
...
+                combinedLength := len(username) + len(password) + len(hostname[1])
+                if combinedLength > 255 {
+                    continue // Skip if the combined length is too long
+                }
+                s1.Raw = []byte(strings.Join([]string{hostname[1], username, password}, "\t"))
+                result = true
+                postgresURL := url.URL{
+                    Scheme: "postgresql",
+                    User:   url.UserPassword(username, password),
+                    Host:   fmt.Sprintf("%s:%s", hostname[1], "5432"),
```
deckarep, line 147: In the verify function there is handling for the default port if nothing else was specified, but it looks like this code always assumes a default port of "5432"

pkg/detectors/postgres/postgres.go
rgmz: No, but that would be a useful enhancement across detectors. I find that 3s for private key lookup can easily time out if there's a lot of network traffic.

pkg/common/patterns.go
```diff
@@ -58,14 +58,14 @@ func (r RegexState) Matches(data []byte) []string {
...
+	raw := fmt.Sprintf(`(?im)(?:user|usr)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:%+v'"\s]{4,40})`, pattern)
 
 	return RegexState{regexp.MustCompile(raw)}
 }
 
-// PasswordRegexCheck constructs an username usernameRegex pattern from a given pattern of excluded characters.
+// PasswordRegexCheck constructs a password passwordRegex pattern from a given pattern of excluded characters.
 func PasswordRegexCheck(pattern string) RegexState {
-	raw := fmt.Sprintf(`(?im)(?:pass|password)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:%+v]{4,40})`, pattern)
+	raw := fmt.Sprintf(`(?im)(?:pass)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:%+v'"\s]{4,40})`, pattern)
```
0x1, line 68: @dylanTruffle any reason for adding `'` and `"` to the excluded chars in password? for username that makes sense, but for password those are valid - no?

pkg/common/patterns.go
```diff
@@ -58,14 +58,14 @@ func (r RegexState) Matches(data []byte) []string {
 
 // UsernameRegexCheck constructs an username usernameRegex pattern from a given pattern of excluded characters.
 func UsernameRegexCheck(pattern string) RegexState {
-	raw := fmt.Sprintf(`(?im)(?:user|usr)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:%+v]{4,40})\b`, pattern)
+	raw := fmt.Sprintf(`(?im)(?:user|usr)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:%+v'"\s]{4,40})`, pattern)
```
0x1, line 61: @dylanTruffle to clarify, the reason for removing `/b` in username was to be able to detect it in the middle of URL, right?

pkg/common/patterns.go
dustin-decker: I'm guessing he was adding that for some boundary condition, but there are better ways to do that. Right now it doesn't look like there is any sort of boundary.

0x1: makes sense. i still have concerns that `'` or `"` may be valid chars. 

for example, Couchbase, the detector for which uses this library, allows both as valid special characters: [couchbase password policy](https://docs.couchbase.com/server/current/rest-api/rest-set-password-policy.html#curl-syntax:~:text=Acceptable%20special%20characters)

perhaps we need to add additional mechanisms to catch or filter passwords within quotes

rgmz: In my experience this can't be done in a satisfying way without defining the pattern *n* times. Otherwise it is not exhaustive and is prone to errors.

E.g.,
```regexp
\s([^\s]+)\s|"([^"]+)"|'([^']+)'
```

pkg/detectors/postgres/postgres_test.go
```diff
@@ -0,0 +1,336 @@
...
+			want: func() []detectors.Result {
+				r := detectors.Result{
+					DetectorType: detectorspb.DetectorType_Postgres,
+					Verified:     false,
+				}
+				return []detectors.Result{r}
+			}(),
+			wantErr: false,
+		},
+		// This test seems take a long time to run (70s+) even with the timeout set to 1s. It's not clear why.
```
0x1, line 226: inactive/incorrect hosts always seem to take a long time even with timeout (70s+). not sure why. this could present a problem when running. any ideas why this might happen or how to mitigate?

rosecodym: which call specifically is timing out?

rosecodym:
Can any of this code be unified with the postgres capabilities of the JDBC detector?

pkg/detectors/postgres/postgres_test.go
0x1: [this](https://github.com/trufflesecurity/trufflehog/pull/2108/files#diff-655d62de1dd6db263b159e4695761b8052314875331af5a1e97e5c95e2216f64R218) one

0x1:
> Can any of this code be unified with the postgres capabilities of the JDBC detector?

not without a lot more work. i think we could pull out some of the logic (like checking if error is determinate) into a common package, but i think it makes sense to do that in a separate PR.

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,239 @@
...
+	for _, hostname := range hostnames {
+		for _, port := range ports {
+			hosts = append(hosts, fmt.Sprintf("%s:%s", hostname, port))
+		}
+	}
+
+	if verify {
+		var verifiedHosts []string
+		for _, host := range hosts {
+			hostname, port := strings.Split(host, ":")[0], strings.Split(host, ":")[1]
```
mcastorina, line 148: **Nit:** I don't think the Go compiler is smart enough to know this should be one call to `strings.Split`

Suggestion:
```go
parts := strings.Split(host, ":")
hostname, port = parts[0], parts[1]
```

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,239 @@
...
+	} else if strings.Contains(err.Error(), "password authentication failed") || // incorrect username or password
+		strings.Contains(err.Error(), "connection refused") { // inactive host
+		return false, nil
+	}
+
+	return false, err
+}
+
+func determineSSLMode(pgURL *url.URL) string {
+	sslmode := "disable"
```
mcastorina, line 230: The default sslmode is `prefer`. See: https://www.postgresql.org/docs/current/libpq-ssl.html

0x1: lol "this makes no sense from a security point of view, and it only promises performance overhead if possible. It is only provided as the default for backward compatibility, and is not recommended in secure deployments."

0x1: looks like prefer is [not implemented](https://github.com/lib/pq/issues/1006) yet

`unsupported sslmode "prefer"; only "require" (default), "verify-full", "verify-ca", and "disable" supported`

pkg/detectors/postgres/postgres_test.go
0x1: yep, good call. added.

pkg/detectors/postgres/postgres.go
mcastorina: Hm, I was thinking we should use the default in the case where the original connection string doesn't specify the `sslmode`. Since we're using `libpq` (right?), is there a different default value?

My main concern is that we don't know the network we are using to connect to the DB, and we shouldn't change the default behavior. Especially to a less secure option.

Maybe `determineSSLMode` should return `(string, bool)` to differentiate when it's not specified?

0x1: i've updated it to use 'require' as default and fall-back to 'disable' _only_ in the case where we get an error saying ssl is not enabled.

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,264 @@
...
+					Host:   fmt.Sprintf("%s:%s", hostname, port),
+				}
+				pgURLs = append(pgURLs, postgresURL)
+			}
+		}
+	}
+	return pgURLs
+}
+
+// if verification is turned on, and we are able to verify at least one host, return only verified hosts
```
rosecodym, line 136: you're overloading "verify" here in that you're using it to mean "postgres is running at all" rather than "we found valid credentials". i'd find it easier to read if you reworded the comment to not use the word "verify", e.g. `if verification is turned on, and we can confirm that postgres is running on at least one host, return only hosts where it's running. otherwise return all hosts.`

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,264 @@
...
+		setOfMatches[match] = struct{}{}
+	}
+	var results []string
+	for match := range setOfMatches {
+		results = append(results, match)
+	}
+	return results
+}
+
+func dedupMatches2(matches [][]string) []string {
```
rosecodym, line 186: this is gross but i can't think of any less-gross alternatives :/ maybe include a doc comment explaining why it exists?

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,264 @@
...
+		return verifyPostgres(pgURL, timeoutInSeconds)
+	}
+	return false, err
+}
+
+func determineSSLMode(pgURL *url.URL) string {
+	// default ssl mode is "prefer" per https://www.postgresql.org/docs/current/libpq-ssl.html
+	// but is currently not implemented in the driver per https://github.com/lib/pq/issues/1006
+	// default to "disable" for now as it is the least restrictive
+	sslmode := "require"
```
rosecodym, line 255: these don't seem to match!

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		if strings.Contains(err.Error(), "connection refused") {
+			// inactive host
+			return false, nil
+		}
+		return false, err
+	}
+	defer db.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
rosecodym, line 232: how did you pick this timeout?

pkg/common/patterns.go
rosecodym, line 61: I'm pretty uncomfortable changing these "common" patterns in a PR that also adds a new detector, because one change is very general and one is very targeted. That suggests they should be in different PRs.

_However_, these "common" patterns are only used in three places. I'm not sure why they were made common in the first place - couchbase, snowflake, and postgres don't actually have any inherent relationship. This reads like it was purely mechanical code deduplication, which isn't usually a good idea. This, in turn, suggests re-duplicating the patterns into the specific locations they're used and just changing the ones that need changing. In this PR, that would just mean putting these patterns directly into the postgres detector and leaving these "general" changes to a future PR.

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,264 @@
...
+
+	err = db.PingContext(ctx)
+	if err == nil {
+		return true, nil
+	} else if strings.Contains(err.Error(), "password authentication failed") || // incorrect username or password
+		strings.Contains(err.Error(), "connection refused") { // inactive host
+		return false, nil
+	}
+
+	// if ssl is not enabled, retry with sslmode=disable
```
rosecodym, line 243: as being discussed in slack: why not just use `allow` from the jump?

pkg/detectors/postgres/postgres.go
0x1: i think this was leftover from dylan's changes.

pkg/detectors/postgres/postgres.go
0x1: allow isn't supported by driver: https://github.com/lib/pq/issues/1006

pkg/detectors/postgres/postgres.go
```diff
@@ -0,0 +1,264 @@
...
+		if pgURL.User == nil {
+			continue
+		}
+		username := pgURL.User.Username()
+		password, _ := pgURL.User.Password()
+		hostport := pgURL.Host
+		result := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Postgres,
+			Raw:          []byte(username + password),
+			RawV2:        []byte(hostport + username + password),
```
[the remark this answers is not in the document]
0x1: i believe it's supposed to be unique, and hostport+username wouldn't be unique if we find multiple potential passwords.

pkg/common/patterns.go
0x1: done.

pkg/detectors/postgres/postgres.go
0x1: do you think it should be smaller?

rosecodym: i expected _larger_ but i haven't really thought about it too much and i don't think it's a big deal either way

ahrav:
Awesome job @0x1 this has been a pretty huge undertaking. Appreciate you. ❤️

rosecodym (approved):

rosecodym:
Thanks for taking this on!

PR #2115, opened 2023-11-17
fixed gist direct link generation

joeleonjr:
### Description:
GitHub Gists are scanned using the `trufflehog github --repo <gist_url>` command, since a gist is just a git repo. At the moment, the `GenerateLink` function used to craft links to specific files from commit in GitHub repos containing secrets does not work properly for gists. The formatting does not include "blob" or "commit" and the fragment identifier works slightly differently (ex: `/#file-filename-txt-L5`). 

I adjusted the logic so that it generates accurate direct links to gists containing secrets.

Here are two gists to test this out on:

https://gist.github.com/joeleonjr/be68e34b002e236160dbb394bbda86fb (note two different AWS canary tokens leaked at different commits)

https://gist.github.com/joeleonjr/21b3d4460c54d4f8e18a7f71f5d4cb21 (note the file name with multiple extensions...those are converted from "." into "-" by GH)

### Checklist:

### DIFF
--- a/pkg/giturl/giturl.go
+++ b/pkg/giturl/giturl.go
@@ -135,7 +135,25 @@ func GenerateLink(repo, commit, file string, line int64) string {
 		fallthrough
 	default:
 		var baseLink string
-		if file == "" {
+
+		//Gist links are formatted differently
+		if strings.HasPrefix(repo, "https://gist.github.com") {
+			baseLink = repo[:len(repo)-4] + "/"
+			if commit != "" {
+				baseLink += commit + "/"
+			}
+			if file != "" {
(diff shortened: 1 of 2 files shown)


mcastorina:
Thanks! Could you also add a test case [here](https://github.com/trufflesecurity/trufflehog/blob/v3.63.0/pkg/giturl/giturl_test.go#L105)?

joeleonjr:
> Thanks! Could you also add a test case [here](https://github.com/trufflesecurity/trufflehog/blob/v3.63.0/pkg/giturl/giturl_test.go#L105)?

Added! Please let me know if you need more.

mcastorina (approved):
💯

PR #2118, opened 2023-11-20
Fix forks and repos counter, add metric for orgs enumerated

dustin-decker:
Fixes a counter and adds a new metric for orgs enumerated. If we used the orgs cache for enumeration with app I would have gotten the count there instead. The metric is needed to confirm that a customer integration migration is performing as expected.

### Checklist:

### DIFF
--- a/pkg/sources/github/repo.go
+++ b/pkg/sources/github/repo.go
@@ -200,6 +200,8 @@ func (s *Source) processRepos(ctx context.Context, target string, listRepos repo
 		numRepos, numForks int
 	)
 
+	uniqueOrgs := map[string]struct{}{}
+
 	for {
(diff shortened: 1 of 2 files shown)


rgmz:
`num_members` also seems to be broken, only showing 0.

pkg/sources/github/repo.go
```diff
@@ -220,7 +222,16 @@ func (s *Source) processRepos(ctx context.Context, target string, listRepos repo
 			if r.GetFork() && !s.conn.IncludeForks {
 				continue
 			}
-			numForks++
+
+			if r.GetFork() {
```
ahrav, line 226: Thanks for fixing this.

ahrav (approved):

PR #2126, opened 2023-11-27
added ci scanning info to readme

joeleonjr:
### Description:
The README doesn't detail the standard process of running TruffleHog in CI. Updated to include instructions.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ trufflehog docker --image trufflesecurity/secrets --only-verified
 
 ## 11: Scan in CI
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -196,6 +196,14 @@ Use the `--image` flag multiple times to scan multiple images.
 trufflehog docker --image trufflesecurity/secrets --only-verified
 ```
 
+## 11: Scan in CI
+
+Set the `--since-commit` flag to your default branch that people merge into (ex: "main"). Set the `--branch` flag to your PR's branch name (ex: "feature-1"). The `--fail` flag will return an 183 error code if valid credentials are found.
```
rosecodym, line 201: In a CI workflow, the branch name will probably be exposed via a variable rather than something that can be hard-coded. What do you think of mentioning this and possibly linking to some documentation of those variables for common CI platforms?

rosecodym (approved):
thanks for this!

PR #2130, opened 2023-11-28
Add disk buffer tempfile cleanup

codevbus:
### Description:
Extends automatic cleanup to orphaned temp files from DiskBufferReader

### Checklist:


codevbus:
Setting this to draft, will need to merge this: https://github.com/trufflesecurity/disk-buffer-reader/pull/9 before this functionality can be implemented.

pkg/cleantemp/cleantemp.go
```diff
@@ -85,3 +102,45 @@ func CleanTempDir(ctx logContext.Context, dirName string, pid int) error {
 	}
 	return nil
 }
+
+func CleanTempFile(ctx logContext.Context, fileName string, pid int) error {
```
rosecodym, line 106: Does this method differ from `CleanTempDir` in any way other than inverting the `file.IsDir()` check? (Other than its log messages.) If not, and it's always called alongside `CleanTempFile`, then can they just be collapsed into one function that deletes things irrespective of dir-ness?

Put another way: Instead of adding a new function to delete files and always calling it alongside the old function that deletes directories, could you modify the old function to delete files and directories together? That would result in less code to maintain without creating any new coupling.

pkg/cleantemp/cleantemp.go
```diff
@@ -27,84 +41,62 @@ func MkdirTemp() (string, error) {
...
-			// Mark these directories initially as ones that should be deleted
+	for _, file := range files {
+		// Match files or directories excluding the current PID
+		if re.MatchString(file.Name()) && !strings.Contains(file.Name(), pidStr) {
 			shouldDelete := true
-			// If they match any live PIDs, mark as should not delete
+			// Check if the name matches any live PIDs
 			for _, pidval := range pids {
-				if strings.Contains(dir.Name(), fmt.Sprintf("-%s-", pidval)) {
+				if strings.Contains(file.Name(), pidval) {
```
dustin-decker, line 78: can you revert this change?

dustin-decker:
Looks like there is a merge conflict still

codevbus:
@dustin-decker - this looks like a net new merge conflict after additional changes were made to the underlying pre-merge changes: https://github.com/trufflesecurity/trufflehog/pull/2136

pkg/cleantemp/cleantemp.go
```diff
@@ -27,27 +26,33 @@ func MkdirTemp() (string, error) {
 	return dir, nil
 }
 
+// Unlike MkdirTemp, we only want to generate the filename string.
+// The tempfile creation in trufflehog we're interested in
+// is generally handled by "github.com/trufflesecurity/disk-buffer-reader"
+func MkFilename() string {
+	pid := os.Getpid()
+	filename := fmt.Sprintf("%s-%d-", "trufflehog", pid)
```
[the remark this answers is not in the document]
codevbus: Could you elaborate? Currently `MkFilename()` gets called whenever a diskbuffer file needs to be generated.

pkg/cleantemp/cleantemp.go
```diff
@@ -14,11 +14,16 @@ import (
 	logContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
+const (
```
codevbus, line 17: @rosecodym - moved the string/formatting to a global const. I think leaving https://github.com/trufflesecurity/trufflehog/pull/2130/files#diff-60ea9ceab29f368c394ac493e6abe8a9195a1fbf7ea2dedbe420f94b8e84cfcfR78 as is may still make sense as it should catch anything with a live pid value in it even if, for whatever reason, it didn't originate from this module.

rosecodym (approved):

PR #2133, opened 2023-11-28
Simplify temp dir cleaning

dustin-decker:
Simplifies the temporary directory cleaning.

### Checklist:

### DIFF
--- a/pkg/cleantemp/cleantemp.go
+++ b/pkg/cleantemp/cleantemp.go
@@ -57,7 +57,7 @@ func CleanTempDir(ctx logContext.Context) error {
 	}
 
 	tempDir := os.TempDir()
-	files, err := os.ReadDir(tempDir)
+	dirs, err := os.ReadDir(tempDir)
 	if err != nil {
 		return fmt.Errorf("Error reading temp dir: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/cleantemp/cleantemp.go
```diff
@@ -44,38 +51,35 @@ func CleanTempDir(ctx logContext.Context, dirName string, pid int) error {
 	}
 
 	for _, proc := range procs {
-		if strings.Contains(proc.Executable(), dirName) {
+		if proc.Executable() == execName {
```
mcastorina, line 54: This is stricter check - have you tested it's still working as intended?

pkg/cleantemp/cleantemp.go
```diff
@@ -44,38 +51,35 @@ func CleanTempDir(ctx logContext.Context, dirName string, pid int) error {
...
-	files, err := os.ReadDir(tempDir)
+	dirs, err := os.ReadDir(tempDir)
 	if err != nil {
 		return fmt.Errorf("Error reading temp dir: %w", err)
 	}
 
-	// Current PID
-	pidStr := strconv.Itoa(pid)
-
 	pattern := `^trufflehog-\d+-\d+$`
```
mcastorina, line 65: This pattern is hard-coded to use `trufflehog`, should it be using `execName`? Sorry I'm not familiar with the naming convention we decided on.

mcastorina (approved):

pkg/cleantemp/cleantemp.go
dustin-decker: I did not but it says ```	// Executable name running this process. This is not a path to the
	// executable.```

dustin-decker: I added a rough test to verify that.

pkg/cleantemp/cleantemp.go
dustin-decker: No, MkdirTemp has "trufflehog" hardcoded to create the temp directories.

PR #2137, opened 2023-11-30
Fix azure storage panic when invalid URL is constructed

dustin-decker:
### Description:
Fix azure panic when invalid URL is constructed

### Checklist:


rgmz:
The pattern can be further refined to only detect valid characters. This is a part of #2097, which I haven't finalized yet:

https://github.com/trufflesecurity/trufflehog/blob/bd0600b7c7a416ee2c219f5c2dc5f5264fe92043/pkg/detectors/azurestorage/azurestorage.go#L26-L27

rosecodym (approved):
👍 as a short-term fix

PR #2139, opened 2023-11-30
Check private keys concurrently

rgmz:
### Description:
This uses a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) to look up private keys concurrently, rather than in sequence.

~Edit: I somehow forgot to make the actual functions go routines. I will need to correct that tomorrow.~

### Checklist:

### DIFF
--- a/pkg/detectors/privatekey/privatekey.go
+++ b/pkg/detectors/privatekey/privatekey.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors"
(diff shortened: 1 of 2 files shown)


rosecodym (approved):
thanks for this!

dustin-decker:
I think this is okay because we own the service being checked, but otherwise we should not do this for other providers. If anything, we need to consider host-based throttling to ensure we're not slamming providers.

rgmz:
That is correct, and it will still iterate through each result sequentially as the WaitGroup is inside the loop.

I suppose you could argue that if the detector is faster it will lead to more requests, however, I don't forsee that being a meaningful increase.

rgmz:
I've resolved the conflicts.

PR #2142, opened 2023-12-01
[chore] - fix error comparisons

ahrav:
### Description:
Correctly handle error comparisons.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -3,6 +3,7 @@ package git
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/sources/git/git.go
```diff
@@ -791,7 +790,7 @@ func stripPassword(u string) (string, error) {
 
 	repoURL, err := url.Parse(u)
 	if err != nil {
-		return "", errors.WrapPrefix(err, "repo remote cannot be sanitized as URI", 0)
+		return "", fmt.Errorf("repo remote cannot be sniffed as URI: %w", err)
```
mcastorina, line 793: **Nit:** I had to re-read this error message a couple of times and look at the context before it clicked what it is conveying. What about `repo remote is not a URI`?

mcastorina (approved):

PR #2143, opened 2023-12-01
[chore] - fix import name clashes

ahrav:
### Description:
Make sure var names don't clash with imported packages.

### Checklist:


mcastorina (approved):

rosecodym (approved):
this doesn't seem bad or anything but what prompted you to do it?

ahrav:
> this doesn't seem bad or anything but what prompted you to do it?

Noticed the issue in a file i randomly had open and then decided I needed a break from on-call work. So figured i'd knock some of these things out.

PR #2144, opened 2023-12-01
[chore] - update empty slice declaration

ahrav:
### Description:
Update empty slice declarations to use `var`.

### Checklist:


pkg/detectors/tailscale/tailscale_test.go
```diff
@@ -21,7 +21,7 @@ func TestTailscaleapi_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
 	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors4")
-	if err != nil && false {
+	if err != nil {
```
mcastorina, line 24: Was the `&& false` put there to disable the test case? If so, why is it being removed in this PR?

ahrav: I don't think so. I removed it because the test passes. It might've been added until we added the token to our GSM.

mcastorina (approved):

PR #2154, opened 2023-12-02
[chore] - Increase pagination limit

ahrav:
### Description:
- Some cleanup.
- Set pagination limit to the max value allowed (100).

### Checklist:

### DIFF
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -293,45 +293,62 @@ func (s *Source) basicAuthSuccessful(apiClient *gitlab.Client) bool {
 func (s *Source) getAllProjects(ctx context.Context, apiClient *gitlab.Client) ([]*gitlab.Project, error) {
 	// Projects without repo will get user projects, groups projects, and subgroup projects.
 	user, _, err := apiClient.Users.CurrentUser()
-
 	if err != nil {
-		msg := fmt.Sprintf("unable to authenticate using: %s", s.authMethod)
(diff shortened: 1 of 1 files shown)


pkg/sources/gitlab/gitlab.go
```diff
@@ -293,45 +293,62 @@ func (s *Source) basicAuthSuccessful(apiClient *gitlab.Client) bool {
...
+		return nil, fmt.Errorf("unable to authenticate using %s, %w", s.authMethod, err)
+	}
+
+	uniqueProjects := make(map[int]*gitlab.Project)
+	var (
+		projects              []*gitlab.Project
+		projectsWithNamespace []string
+	)
+
+	// Used to filter out duplicate projects.
```
rosecodym, line 306: do we see these? (i guess we could if the project list is updated during the operation)

rosecodym (approved):
thanks for looking at this! i still think there was some other cause of the slowdown though - 10m isn't 80% too long, it's, like, 99.9% too long

pkg/sources/gitlab/gitlab.go
ahrav: Yea, i was initially thinking the same thing, but i think the same project can appear under the user's list as well as the group's list.

ahrav:
> thanks for looking at this! i still think there was some other cause of the slowdown though - 10m isn't 80% too long, it's, like, 99.9% too long

Yea there has to be something else going on here, this is by no means "THE FIX".

PR #2174, opened 2023-12-05
[feat] - Remove go-git dependency

ahrav:
### Description:
Replace go-git with git cat-file for reading binary files

The go-git library loads entire file contents into memory when reading files from a commit. This causes high memory usage for large binaries.

This PR shells out to the native git cat-file command instead to retrieve file contents, which reduces memory consumption by streaming directly from the git object database rather than loading into memory.

By replacing go-git usage with git cat-file, memory usage is reduced significantly when processing large files in older commits.

Here are some profiling statistics:
Memory before:
Screenshot 2023-12-04 at 5 20 35 PM

CPU before:
Screenshot 2023-12-04 at 5 21 01 PM

Memory after:
Screenshot 2023-12-04 at 5 24 24 PM

CPU after.
Screenshot 2023-12-04 at 5 24 59 PM

This entire codepath is gone:
Screenshot 2023-12-04 at 5 03 23 PM

### Checklist:

### DIFF
--- a/pkg/engine/git_test.go
+++ b/pkg/engine/git_test.go
@@ -106,12 +106,7 @@ func TestGitEngine(t *testing.T) {
 
 func BenchmarkGitEngine(b *testing.B) {
 	ctx := context.Background()
-	repoUrl := "https://github.com/dustin-decker/secretsandstuff.git"
-	path, _, err := git.PrepareRepo(ctx, repoUrl)
-	if err != nil {
-		b.Error(err)
-	}
-	defer os.RemoveAll(path)
+	repoUrl := "https://github.com/trufflesecurity/trufflehog.git"
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
@@ -132,11 +127,9 @@ func BenchmarkGitEngine(b *testing.B) {
 		}
 	}()
 
+	cfg := sources.GitConfig{URI: repoUrl}
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		// TODO: this is measuring the time it takes to initialize the source
-		// and not to do the full scan
-		cfg := sources.GitConfig{URI: path}
 		if err := e.ScanGit(ctx, cfg); err != nil {
 			return
 		}
--- a/pkg/engine/git_test.go
+++ b/pkg/engine/git_test.go
@@ -106,7 +106,7 @@ func TestGitEngine(t *testing.T) {
 
 func BenchmarkGitEngine(b *testing.B) {
 	ctx := context.Background()
-	repoUrl := "https://github.com/trufflesecurity/trufflehog.git"
+	repoUrl := "https://github.com/dustin-decker/secretsandstuff.git"
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -982,9 +982,9 @@ func handleBinary(ctx context.Context, gitDir string, reporter sources.ChunkRepo
 		catFileArg = "cat-file"
 		blobArg    = "blob"
 
-		maxSize = 100 * 1024 * 1024 // 100 MB
+		maxSize = 1 * 1024 * 1024 * 1024 // 1GB
 	)
-	cmd := exec.Command("git", "--git-dir="+gitDir, catFileArg, blobArg, commitHash.String()+":"+path)
+	cmd := exec.Command("git", "-C", gitDir, catFileArg, blobArg, commitHash.String()+":"+path)
 
 	var stderr bytes.Buffer
 	cmd.Stderr = &stderr
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -978,13 +978,8 @@ func handleBinary(ctx context.Context, gitDir string, reporter sources.ChunkRepo
 		return nil
 	}
 
-	const (
-		catFileArg = "cat-file"
-		blobArg    = "blob"
-
-		maxSize = 1 * 1024 * 1024 * 1024 // 1GB
-	)
-	cmd := exec.Command("git", "-C", gitDir, catFileArg, blobArg, commitHash.String()+":"+path)
+	const maxSize = 1 * 1024 * 1024 * 1024 // 1GB
+	cmd := exec.Command("git", "-C", gitDir, "cat-file", "blob", commitHash.String()+":"+path)
 
 	var stderr bytes.Buffer
 	cmd.Stderr = &stderr
@@ -999,11 +994,11 @@ func handleBinary(ctx context.Context, gitDir string, reporter sources.ChunkRepo
 	}
 	defer func() {
 		if err := fileReader.Close(); err != nil {
-			ctx.Logger().Error(err, "Error closing fileReader")
+			ctx.Logger().Error(err, "error closing fileReader")
 		}
 		if err := cmd.Wait(); err != nil {
 			ctx.Logger().Error(
-				err, "Error waiting for command",
+				err, "error waiting for command",
 				"command", cmd.String(),
 				"stderr", stderr.String(),
 				"commit", commitHash,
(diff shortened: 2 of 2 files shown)


pkg/sources/git/git.go
```diff
@@ -964,30 +970,66 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
+		ctx.Logger().V(5).Info("skipping binary file", "path", path)
+		return nil
 	}
 
-	file, err := commit.File(path)
+	const (
+		catFileArg = "cat-file"
+		blobArg    = "blob"
+
+		maxSize = 100 * 1024 * 1024 // 100 MB
```
bill-rich, line 985: My opinion is that this should match the max diff size (`pkg/gitparse/gitparse.go`).

pkg/sources/git/git.go
```diff
@@ -964,30 +970,66 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
+	)
+	cmd := exec.Command("git", "--git-dir="+gitDir, catFileArg, blobArg, commitHash.String()+":"+path)
```
bill-rich, line 987: Are `--git-dir` and `-C` doing the same thing?

pkg/sources/git/git.go
```diff
@@ -964,30 +970,66 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
-	if err != nil {
+	var fileContent bytes.Buffer
+	// Create a limited reader to ensure we don't read more than the max size.
+	lr := io.LimitReader(fileReader, int64(maxSize))
+
+	// Using io.CopyBuffer for performance advantages. Though buf is mandatory
+	// for the method, due to the internal implementation of io.CopyBuffer, when
+	// *bytes.Buffer implements io.WriterTo or io.ReaderFrom, the provided buf
+	// is simply ignored. Thus, we can pass nil for the buf parameter.
+	_, err = io.CopyBuffer(&fileContent, lr, nil)
```
bill-rich, line 1022: Does this end up with the file content in memory twice (for `lr` and `fileContent`)? If so, could we use dbr instead?

pkg/engine/git_test.go
```diff
@@ -106,12 +106,7 @@ func TestGitEngine(t *testing.T) {
 
 func BenchmarkGitEngine(b *testing.B) {
 	ctx := context.Background()
-	repoUrl := "https://github.com/dustin-decker/secretsandstuff.git"
-	path, _, err := git.PrepareRepo(ctx, repoUrl)
-	if err != nil {
-		b.Error(err)
-	}
-	defer os.RemoveAll(path)
+	repoUrl := "https://github.com/trufflesecurity/trufflehog.git"
```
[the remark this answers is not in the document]
ahrav: Ahh whoops I changed it to get better benchmarks but I'll change this back.

pkg/sources/git/git.go
ahrav: Will do!

pkg/sources/git/git.go
ahrav: That's a great question, I think so. I can use -C instead.

pkg/sources/git/git.go
ahrav: Oh no, it only loads the data into memory once. When we populate fileContents. The limit reader shouldn't load anything into memory, I believe it only sets a limit on the data that can be read.

ahrav: You can also see it in the profile there is only one alloc present.

pkg/sources/git/git.go
ahrav: TIL: https://stackoverflow.com/a/42299983

rosecodym (commented):
you're gonna _hate_ me for this but have we tested this on all of our supported oses?

ahrav:
> you're gonna _hate_ me for this but have we tested this on all of our supported oses?

Just linux and Mac.

pkg/sources/git/git.go
```diff
@@ -964,30 +970,66 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
+	if err != nil && !errors.Is(err, io.EOF) {
 		return err
 	}
-	defer fileReader.Close()
 
-	reader, err := diskbufferreader.New(fileReader)
+	if fileContent.Len() == maxSize {
+		ctx.Logger().V(2).Info("Max archive size reached.", "path", path)
+		ctx.Logger().Info("Max archive size reached.", "path", path)
```
rosecodym, line 1029: What's the purpose of this double logging?

rosecodym:
> > you're gonna _hate_ me for this but have we tested this on all of our supported oses?
> 
> Just linux and Mac.

Unfortunately, I ask because I've seen stdout handling like this blow up on Windows. What's the lift to test it there? I'd hate for this to break the entire Windows version.

rosecodym:
> > Member
> 
> I believe this pattern of handling Git output should work properly on Windows as well, since we use a similar approach in other areas of our Git integration code. However, you have raised a fair point that there may be nuances to account for. I don't recall us running into Windows-specific issues with piping Git stdout before, specifically [here](https://github.com/trufflesecurity/trufflehog/blob/f967fd8c17afb4f6aee949aa36f903a6d539a1de/pkg/gitparse/gitparse.go#L220C24-L220C24) and
> 
> I'm not certain on the lift to test on Windows, i'm not sure if anyone else has advice here.

Oh, if we've done this before without trouble then I'm satisfied.

pkg/sources/git/git.go
```diff
@@ -964,30 +970,65 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
 	}
 
-	file, err := commit.File(path)
+	const (
+		catFileArg = "cat-file"
+		blobArg    = "blob"
+
+		maxSize = 1 * 1024 * 1024 * 1024 // 1GB
+	)
+	cmd := exec.Command("git", "-C", gitDir, catFileArg, blobArg, commitHash.String()+":"+path)
```
mcastorina, line 987: **Extreme nit:** I know you like consts, but I think `catFileArg` and `blobArg` can be inlined and it would be more readable.

pkg/sources/git/git.go
```diff
@@ -964,30 +970,65 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
 
-	if common.SkipFile(file.Name) {
-		ctx.Logger().V(5).Info("skipping binary file", "path", path, "file_name", file.Name)
-		return nil
+	if err := cmd.Start(); err != nil {
+		return err
 	}
+	defer func() {
+		if err := fileReader.Close(); err != nil {
+			ctx.Logger().Error(err, "Error closing fileReader")
```
mcastorina, line 1002: **Question:** I've been starting log messages with lowercase - should I be using capitalized messages?

pkg/sources/git/git.go
```diff
@@ -964,30 +970,65 @@ func getSafeRemoteURL(repo *git.Repository, preferred string) string {
...
+		}
+		if err := cmd.Wait(); err != nil {
+			ctx.Logger().Error(
+				err, "Error waiting for command",
```
mcastorina, line 1006: Same question here.

mcastorina (commented):
This doesn't appear to remove the entire `go-git` dependency since it is still imported. What else are we using it for?

mcastorina (approved):

rgmz:
Anecdotally, on WSL/Ubuntu `git cat-file` seems to be failing quite frequently. It's possible that this is more _visibile_ than go-git but not actually more frequent, I haven't had a chance to look deeper into this yet.

```
2023-12-28T16:01:09-05:00       error   trufflehog      error waiting for command       {"source_manager_worker_id": "gRIb9", "command": "/usr/bin/git -C /tmp/trufflehog-526236-1916300294/.git cat-file blob eccf0ea1efa268bccbd4131b0c23e5e168adb5c4:hphp/test/frameworks/perf/wordpress/wordpress-3.9.1.tar.gz", "stderr": "", "commit": "eccf0ea1efa268bccbd4131b0c23e5e168adb5c4", "error": "signal: broken pipe"}
```

PR #2175, opened 2023-12-05
remove unnecessary Git cmd check

ahrav:
### Description:
We don't need to run this check for each commit and clone, it's already handled during init. Each invocation can add up.
Screenshot 2023-12-04 at 5 56 28 PM

### Checklist:


0x1 (approved):
✂️

dustin-decker (approved):

PR #2178, opened 2023-12-05
[fixup] - Skip trying to determine MIME type for directories

ahrav:
### Description:
When handling files within an archive, we should avoid trying to determine the MIME type for directories.

Error:
```
unable to read file for MIME type detection: read /var/folders/1m/m4kl5q2102bfr4k4566nmf800000gn/T/tmp_archive3228926043/usr: is a directory
```

### Checklist:


mcastorina (approved):
We unarchive to the filesystem? It's not done in memory?

ahrav:
> We unarchive to the filesystem? It's not done in memory?

Yea I believe for .rpm and .deb archives.

PR #2179, opened 2023-12-05
[chore] - Refactor common code into a separate function

ahrav:
### Description:
Refactor out code that handles commit resolution for both head and base.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -721,66 +721,73 @@ func (s *Git) ScanRepo(ctx context.Context, repo *git.Repository, repoPath strin
 		}
 	}
 
-	// We're logging time, but the repoPath is usually a dynamically generated folder in /tmp
-	// To make this duration logging useful, we need to log the remote as well
+	// We're logging time, but the repoPath is usually a dynamically generated folder in /tmp.
+	// To make this duration logging useful, we need to log the remote as well.
 	remotes, _ := repo.Remotes()
-	repoUrl := "Could not get remote for repo"
+	repoURL := "Could not get remote for repo"
 	if len(remotes) != 0 {
-		repoUrl = getSafeRemoteURL(repo, remotes[0].Config().Name)
+		repoURL = getSafeRemoteURL(repo, remotes[0].Config().Name)
 	}
 
 	scanTime := time.Now().Unix() - start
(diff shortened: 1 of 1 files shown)


pkg/sources/git/git.go
```diff
@@ -721,68 +721,96 @@ func (s *Git) ScanRepo(ctx context.Context, repo *git.Repository, repoPath strin
...
-				return fmt.Errorf("unable to resolve base ref: %w", err)
-			}
-		}
+// normalizeConfig updates scanOptions with the resolved base and head commit hashes.
+// It's designed to handle scenarios where BaseHash and HeadHash in scanOptions might be branch names or
+// other non-hash references. This ensures that both the base and head commits are resolved to actual commit hashes.
+// If either commit cannot be resolved, it returns early.
+// If both are resolved, it finds and sets the merge base in scanOptions.
+func normalizeConfig(scanOptions *ScanOptions, repo *git.Repository) error {
+	baseCommit, baseSet, err := resolveAndSetCommit(repo, &scanOptions.BaseHash)
```
rosecodym, line 749: `BaseHash` is an unfortunate name for a field that might not be a hash :/

pkg/sources/git/git.go
```diff
@@ -721,68 +721,96 @@ func (s *Git) ScanRepo(ctx context.Context, repo *git.Repository, repoPath strin
...
 	}
 
 	// If baseCommit is an ancestor of headCommit, update c.BaseRef to be the common ancestor.
-	if headCommit != nil && baseCommit != nil {
-		mergeBase, err := headCommit.MergeBase(baseCommit)
-		if err != nil || len(mergeBase) < 1 {
-			return fmt.Errorf("unable to resolve merge base: %w", err)
-		}
-		scanOptions.BaseHash = mergeBase[0].Hash.String()
+	mergeBase, err := headCommit.MergeBase(baseCommit)
```
rosecodym, line 764: i think `headCommit` can be `nil` here (if `len(scanOptions.HeadHash) == 0`)

rosecodym (commented):
nice cleanup but i think there's a possible `nil` dereference in there

pkg/sources/git/git.go
ahrav: Nice catch. Did I introduce that bug, it feels like even with the existing code this could've been an issue.

ahrav: Thinking about this a bit more, i think we need to have both headCommit and baseCommit not be nil. If either are nil we get panics. I'll change the above check for either being nil instead of both being nil.

rosecodym (approved):

PR #2182, opened 2023-12-05
allow targets for the source manager

ahrav:
### Description:
Add arguments to the source manager so that it can handle chunking targets needed for secret reverification.

### Checklist:


ahrav:
Not sure if this is the best way to handle this, so feel free to provide alternatives that are better. :)

mcastorina:
This looks reasonable to me, however it's not actually using the targets. Is that intentional?

ahrav:
> This looks reasonable to me, however it's not actually using the targets. Is that intentional?

Oh whoops, missed the important part. Will add.

mcastorina (approved):

PR #2187, opened 2023-12-07
Detectors Updates 1 for Tristate Verification

0x1:
### Description:
Updates to some detectors to handle verification errors. 
includes:
- handling verification error
- ensuring correct handling of status codes
- pulling out verification logic
- updates to tests for verification errors
- updates to tokens (all tests for detectors updates should be passing)
- update to aeroworkflow detector pattern (previously was not accounting for additional symbols)

### Checklist:

### DIFF
--- a/pkg/detectors/abbysale/abbysale.go
+++ b/pkg/detectors/abbysale/abbysale.go
@@ -17,10 +17,12 @@ type Scanner struct {
 	client *http.Client
 }
 
-// Ensure the Scanner satisfies the interface at compile time.
-var _ detectors.Detector = (*Scanner)(nil)
+const abbysaleURL = "https://api.abyssale.com"
 
 var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector = (*Scanner)(nil)
+
 	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
@@ -76,7 +78,8 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 }
 
 func verifyAbbysale(ctx context.Context, client *http.Client, resMatch string) (bool, error) {
-	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.abyssale.com/ready", nil)
+	// https://developers.abyssale.com/rest-api/authentication
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, abbysaleURL+"/ready", nil)
 	if err != nil {
 		return false, err
 	}
@@ -86,13 +89,15 @@ func verifyAbbysale(ctx context.Context, client *http.Client, resMatch string) (
 		return false, err
 	}
 	defer res.Body.Close()
-	if res.StatusCode == http.StatusOK {
+
+	switch res.StatusCode {
+	case http.StatusOK:
 		return true, nil
-	} else if res.StatusCode != http.StatusForbidden {
+	case http.StatusForbidden:
+		return false, nil
+	default:
 		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
 	}
-
-	return false, nil
 }
 
 func (s Scanner) Type() detectorspb.DetectorType {
--- a/pkg/detectors/abbysale/abbysale.go
+++ b/pkg/detectors/abbysale/abbysale.go
@@ -2,22 +2,26 @@ package abbysale
 
 import (
 	"context"
-	regexp "github.com/wasilibs/go-re2"
+	"fmt"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"abbysale"}) + `\b([a-z0-9A-Z]{40})\b`)
@@ -27,6 +31,13 @@ var (
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
 	return []string{"abbysale"}
+}
+
+func (s Scanner) getClient() *http.Client {
+	if s.client != nil {
+		return s.client
+	}
+	return defaultClient
 }
 
 // FromData will find and optionally verify Abbysale secrets in a given set of bytes.
@@ -47,29 +58,41 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.abyssale.com/ready", nil)
-			if err != nil {
-				continue
-			}
-			req.Header.Add("x-api-key", resMatch)
-			res, err := client.Do(req)
-			if err == nil {
-				defer res.Body.Close()
-				if res.StatusCode >= 200 && res.StatusCode < 300 {
-					s1.Verified = true
-				} else {
-					// This function will check false positives for common test words, but also it will make sure the key appears 'random' enough to be a real key.
-					if detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
-						continue
-					}
-				}
-			}
+			client := s.getClient()
+			isVerified, verificationErr := verifyAbbysale(ctx, client, resMatch)
+			s1.Verified = isVerified
+			s1.SetVerificationError(verificationErr, resMatch)
+		}
+
+		// This function will check false positives for common test words, but also it will make sure the key appears 'random' enough to be a real key.
+		if !s1.Verified && detectors.IsKnownFalsePositive(resMatch, detectors.DefaultFalsePositives, true) {
+			continue
 		}
 
 		results = append(results, s1)
 	}
 
 	return results, nil
+}
+
+func verifyAbbysale(ctx context.Context, client *http.Client, resMatch string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.abyssale.com/ready", nil)
+	if err != nil {
+		return false, err
+	}
+	req.Header.Add("x-api-key", resMatch)
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode != http.StatusForbidden {
+		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
+
+	return false, nil
 }
 
 func (s Scanner) Type() detectorspb.DetectorType {
--- a/pkg/detectors/abstract/abstract_test.go
+++ b/pkg/detectors/abstract/abstract_test.go
@@ -68,7 +68,7 @@ func TestAbstract_FromChunk(t *testing.T) {
 					DetectorType: detectorspb.DetectorType_Abstract,
 					Verified:     false,
 				}
-				r.SetVerificationError(fmt.Errorf("context deadline exceeded"))
+				r.SetVerificationError(context.DeadlineExceeded)
 				return []detectors.Result{r}
 			}(),
 			wantErr: false,
--- a/pkg/detectors/abstract/abstract_test.go
+++ b/pkg/detectors/abstract/abstract_test.go
@@ -9,7 +9,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kylelemons/godebug/pretty"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
@@ -19,7 +20,7 @@ import (
 func TestAbstract_FromChunk(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancel()
-	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors1")
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
 	if err != nil {
 		t.Fatalf("could not get test secrets from GCP: %s", err)
 	}
@@ -43,7 +44,7 @@ func TestAbstract_FromChunk(t *testing.T) {
 			s:    Scanner{},
 			args: args{
 				ctx:    context.Background(),
-				data:   []byte(fmt.Sprintf("You can find a abstract secret %s within", secret)),
+				data:   []byte(fmt.Sprintf("You can find a abstract secret %s within but verified", secret)),
 				verify: true,
 			},
 			want: []detectors.Result{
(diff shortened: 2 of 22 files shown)


pkg/detectors/abuseipdb/abuseipdb.go
```diff
@@ -49,41 +58,52 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		}
+
 		results = append(results, s1)
 	}
 
 	return results, nil
 }
 
+func verifyAbuseIPDB(ctx context.Context, client *http.Client, resMatch string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.abuseipdb.com/api/v2/check?ipAddress=118.25.6.39", nil)
```
rosecodym, line 79: do you have any idea why this ip address is the one we use? (i know this isn't new code, i'm just curious)

pkg/detectors/aeroworkflow/aeroworkflow.go
```diff
@@ -84,6 +84,29 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	req.Header.Add("apikey", resMatch)
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode >= 200 && res.StatusCode < 300 {
+		return true, nil
+	} else if res.StatusCode < http.StatusUnauthorized || res.StatusCode > http.StatusForbidden {
```
rosecodym, line 102: i don't think this is any less typing than a version that doesn't inadvertently include a 402

pkg/detectors/agora/agora_test.go
```diff
@@ -48,13 +49,61 @@ func TestAgora_FromChunk(t *testing.T) {
...
 					Verified:     true,
 				},
+				{
+					DetectorType: detectorspb.DetectorType_Agora,
+					Verified:     false,
+				},
+				{
+					DetectorType: detectorspb.DetectorType_Agora,
+					Verified:     false,
+				},
```
rosecodym, line 67: why did these cases get added?

pkg/detectors/aha/aha.go
```diff
@@ -74,6 +83,30 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+
+	// https://www.aha.io/api
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode < http.StatusUnauthorized || res.StatusCode > http.StatusNotFound {
```
rosecodym, line 103: same comment as above re: 402 (also are both 403 and 404 expected here?)

pkg/detectors/alibaba/alibaba.go
```diff
@@ -87,49 +107,75 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	var alibabaResp alibabaResp
+	if err = json.NewDecoder(res.Body).Decode(&alibabaResp); err != nil {
+		return false, err
+	}
+
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode < http.StatusBadRequest || res.StatusCode > http.StatusNotFound {
+		// 400 used for most of error cases
+		// 404 used if the AccessKeyId is not valid
```
rosecodym, line 167: what about 401, 402, and 403?

rosecodym (commented):
in addition to the structural changes, you've changed some of these verification response checks and even regexes in a couple of cases. i'm a little concerned about combining these multiple types of changes into a single pr like this - is it possible to split them up?

pkg/detectors/abuseipdb/abuseipdb.go
0x1: nope, i'm not sure.

pkg/detectors/alibaba/alibaba.go
0x1: i didn't see 401, 402, and 403 in the my testing. but it was easier to add the range

pkg/detectors/abuseipdb/abuseipdb.go
```diff
@@ -49,41 +58,52 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		return false, err
+	}
+	defer res.Body.Close()
+	if res.StatusCode >= 200 && res.StatusCode < 300 {
+		bodyBytes, err := io.ReadAll(res.Body)
+		if err != nil {
+			return false, err
+		}
+		bodyString := string(bodyBytes)
+		validResponse := strings.Contains(bodyString, `ipAddress`)
```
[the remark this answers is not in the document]
0x1: you'd have to bytify the string tho. would that still be preferred?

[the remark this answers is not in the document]
rosecodym: you mean `ipAddress`? yeah, it's shorter. and if you wanted to go hog-wild you could just do it once in a global `var`.

pkg/detectors/aha/aha.go
```diff
@@ -12,16 +12,19 @@ import (
...
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aha"}) + `\b([0-9a-f]{64})\b`)
+	URLPat = regexp.MustCompile(`\b([A-Za-z0-9](?:[A-Za-z0-9\-]{0,61}[A-Za-z0-9])\.aha\.io)`)
```
rosecodym, line 27: how did you discover this (and other) regex changes?

pkg/detectors/alibaba/alibaba.go
rosecodym: i see two problems with doing the "easier" thing here. the smaller one is that you're assuming that the external system will continue to behave in the way you've seen it behave, which is always a risk. now in this particular case, that's unlikely to cause real problems - 401s and 403s are probably determine verification errors if they ever do show up, and 402s will probably never happen - but it's not a great happen to get into.

the bigger problem is that your code doesn't match your comment, so the next person to read this is going to be confused and probably ask the same question i did. but you won't be around to answer it! so they'll have to spend time chasing down a ghost.

0x1: fair

pkg/detectors/abbysale/abbysale.go
```diff
@@ -72,6 +74,26 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyAbbysale(ctx context.Context, client *http.Client, resMatch string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.abyssale.com/ready", nil)
```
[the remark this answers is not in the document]
0x1: why is it baffling? looks like a normal url, no?

pkg/detectors/aha/aha.go
0x1: i was trying to fix the test by generating new tokens. realized that the url we used to verify this was the generic `https://company.aha.io/api/v1/me` from the docs, and not the actual subdomain that we needed (no the requests don't work with the generic one). so i added this to capture the url as well (a format we use for other detectors that require URLs).

i'll add comments on the other regexes i updated separately.

pkg/detectors/artifactory/artifactory.go
```diff
@@ -11,16 +12,18 @@ import (
...
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b([a-zA-Z0-9]{73})`)
+	keyPat = regexp.MustCompile(`\b([a-zA-Z0-9]{64,73})`)
```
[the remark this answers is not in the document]
rosecodym: Interesting. If you can't find any concrete documentation, I can see arguments both ways. This might be a broader question for the group.

pkg/detectors/abuseipdb/abuseipdb.go
```diff
@@ -49,41 +59,51 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		}
+		validResponse := bytes.Contains(bodyBytes, []byte("ipAddress"))
+		if validResponse {
+			return true, nil
+		} else {
+			return false, nil
+		}
+	} else if res.StatusCode != http.StatusUnauthorized {
+		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
```
rosecodym, line 103: Was this based on any documentation you read, or just reading the original implementation? (Or testing?)

pkg/detectors/accuweather/accuweather.go
```diff
@@ -71,6 +74,24 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		return false, err
+	}
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode != http.StatusUnauthorized {
```
rosecodym, line 89: Was this based on documentation or testing?

pkg/detectors/adzuna/adzuna.go
```diff
@@ -81,6 +83,27 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode != http.StatusUnauthorized {
```
rosecodym, line 100: Was this based on documentation or testing?

pkg/detectors/aeroworkflow/aeroworkflow.go
```diff
@@ -84,6 +84,29 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	req.Header.Add("apikey", resMatch)
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode != http.StatusUnauthorized && res.StatusCode != http.StatusForbidden {
```
rosecodym, line 102: I'm guessing this was based on documentation or testing?

pkg/detectors/aha/aha.go
```diff
@@ -74,6 +83,30 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+
+	// https://www.aha.io/api
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode != http.StatusUnauthorized && res.StatusCode != http.StatusNotFound {
```
rosecodym, line 103: i didn't realize i'd be asking the same question five times until i was already in too deep :(

rosecodym (commented):
I asked this inline where it was relevant, but in general, how did you decide which response status codes constituted determinate versus indeterminate errors? (I know we discussed some of them when the original ranges you used were broader, but I realize now I'm wondering what your general strategy was.) I just want to get it into my head so that I don't ask about every single detector in the future changes :)

We can chat about this synchronously too if that would be easier.

pkg/detectors/abuseipdb/abuseipdb.go
0x1: this was original implementation. i only tested to ensure it still worked, but didn't manually hit the endpoints myself.

0x1:
> I asked this inline where it was relevant, but in general, how did you decide which response status codes constituted determinate versus indeterminate errors? (I know we discussed some of them when the original ranges you used were broader, but I realize now I'm wondering what your general strategy was.) I just want to get it into my head so that I don't ask about every single detector in the future changes :)
> 
> We can chat about this synchronously too if that would be easier.

determinate when failure is due to incorrect secrets (wrong token, wrong user/pass combo, wrong url, etc). these can yield different codes: unauthorized, forbidden, 200 w/ some custom response, etc. (different apis have different implementations).

indeterminate when failure cannot be used to assess token validity.
caveat here is that we can't know every iteration of how the api works, so there might be some edge cases - but we try to ensure the known cases of incorrect secret, and set indeterminacy for everything else. 
for example, i've seen a case where 404 is the response when tokens are incorrect, so in that situation it will be hard to differentiate between incorrect token and if the api url changes at some point.
there are other edgecases also, for example 'unauthorized' is used if token is wrong, but can also be used if token is valid but lacks access/permissions.

pkg/detectors/aeroworkflow/aeroworkflow.go
0x1: yes, this one i manually checked

pkg/detectors/aha/aha.go
0x1: also manually checked

pkg/detectors/adzuna/adzuna.go
0x1: original implementation

pkg/detectors/accuweather/accuweather.go
0x1: original implementation

pkg/detectors/artifactory/artifactory.go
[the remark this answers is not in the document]
0x1: ok i'll dig a bit deeper to see if i can find examples between the two

pkg/detectors/abuseipdb/abuseipdb.go
0x1: added docs now

rosecodym:
> > I asked this inline where it was relevant, but in general, how did you decide which response status codes constituted determinate versus indeterminate errors? (I know we discussed some of them when the original ranges you used were broader, but I realize now I'm wondering what your general strategy was.) I just want to get it into my head so that I don't ask about every single detector in the future changes :)
> > We can chat about this synchronously too if that would be easier.
> 
> determinate when failure is due to incorrect secrets (wrong token, wrong user/pass combo, wrong url, etc). these can yield different codes: unauthorized, forbidden, 200 w/ some custom response, etc. (different apis have different implementations).
> 
> indeterminate when failure cannot be used to assess token validity. caveat here is that we can't know every iteration of how the api works, so there might be some edge cases - but we try to ensure the known cases of incorrect secret, and set indeterminacy for everything else. for example, i've seen a case where 404 is the response when tokens are incorrect, so in that situation it will be hard to differentiate between incorrect token and if the api url changes at some point. there are other edgecases also, for example 'unauthorized' is used if token is wrong, but can also be used if token is valid but lacks access/permissions.

sorry, i didn't ask my question well. i'm looking for more for your personal strategy: do you just hammer on each of these apis and see what comes out? check documentation? both? (if so, in what order?)

0x1:
> > > I asked this inline where it was relevant, but in general, how did you decide which response status codes constituted determinate versus indeterminate errors? (I know we discussed some of them when the original ranges you used were broader, but I realize now I'm wondering what your general strategy was.) I just want to get it into my head so that I don't ask about every single detector in the future changes :)
> > > We can chat about this synchronously too if that would be easier.
> > 
> > 
> > determinate when failure is due to incorrect secrets (wrong token, wrong user/pass combo, wrong url, etc). these can yield different codes: unauthorized, forbidden, 200 w/ some custom response, etc. (different apis have different implementations).
> > indeterminate when failure cannot be used to assess token validity. caveat here is that we can't know every iteration of how the api works, so there might be some edge cases - but we try to ensure the known cases of incorrect secret, and set indeterminacy for everything else. for example, i've seen a case where 404 is the response when tokens are incorrect, so in that situation it will be hard to differentiate between incorrect token and if the api url changes at some point. there are other edgecases also, for example 'unauthorized' is used if token is wrong, but can also be used if token is valid but lacks access/permissions.
> 
> sorry, i didn't ask my question well. i'm looking for more for your personal strategy: do you just hammer on each of these apis and see what comes out? check documentation? both? (if so, in what order?)
yes both. check documentation if available first, hammer if documentation isn't clear.

pkg/detectors/abstract/abstract_test.go
```diff
@@ -54,12 +55,48 @@ func TestAbstract_FromChunk(t *testing.T) {
...
+				ctx:    context.Background(),
+				data:   []byte(fmt.Sprintf("You can find a abstract secret %s within but verified", secret)),
+				verify: true,
+			},
+			want: func() []detectors.Result {
+				r := detectors.Result{
+					DetectorType: detectorspb.DetectorType_Abstract,
+					Verified:     false,
+				}
+				r.SetVerificationError(context.DeadlineExceeded)
```
rosecodym, line 71: til about `context.DeadlineExceeded`!

rosecodym (approved):

rosecodym (approved):
phew, finally over the finish line! nice work doing all this research

PR #2196, opened 2023-12-09
[fixup] - Avoid reading decompressed data into memory

ahrav:
### Description:
Update archive handling process to avoid redundant decompression and improve overall efficiency. Currently, the code decompresses the archive data and then reads the decompressed content into memory before processing it. This approach is inefficient, especially for large archives, as it leads to unnecessary memory usage and performance overhead.

Instead of reading the decompressed content into memory, directly feed the decompressed reader back into the archive handler. This eliminates the need for intermediate buffering and allows for streaming processing, reducing memory usage and improving performance.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -84,13 +84,6 @@ func (a *Archive) FromFile(originalCtx context.Context, data io.Reader) chan []b
 	return archiveChan
 }
 
-type decompressorInfo struct {
-	depth       int
-	reader      io.Reader
-	archiveChan chan []byte
-	archiver    archiver.Decompressor
-}
-
 // openArchive takes a reader and extracts the contents up to the maximum depth.
 func (a *Archive) openArchive(ctx context.Context, depth int, reader io.Reader, archiveChan chan []byte) error {
 	if depth >= maxDepth {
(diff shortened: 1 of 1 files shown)


pkg/handlers/archive.go
```diff
@@ -108,11 +101,14 @@ func (a *Archive) openArchive(ctx context.Context, depth int, reader io.Reader,
...
-		return a.handleDecompressor(ctx, info)
+		// Decompress tha archive and feed the decompressed data back into the archive handler to extract any nested archives.
+		compReader, err := archive.OpenReader(arReader)
+		if err != nil {
+			return err
+		}
+		return a.openArchive(ctx, depth+1, compReader, archiveChan)
 	case archiver.Extractor:
-		return archive.Extract(context.WithValue(ctx, depthKey, depth+1), reader, nil, a.extractorHandler(archiveChan))
+		return archive.Extract(context.WithValue(ctx, depthKey, depth+1), arReader, nil, a.extractorHandler(archiveChan))
```
rosecodym, line 111: Why did this change?

ahrav: The extract should be using reader from `archiver.Identify` otherwise `Extract` is using the wrong reader. We should only be using the `arReader` once it's available.

ahrav: The reason why it currently works is because we are reading all the data out of the compressed file and passing the  `bytes.NewReader` back to `openArchive`. If we pass the reader directly, we need to make sure to use the new reader as a result of `archiver.Identify`.

mcastorina (approved):

PR #2197, opened 2023-12-09
[fixup] - skip binary files.

ahrav:
### Description:
Avoid scanning binary and non-text files within the archive handler.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gabriel-vasile/mimetype"
 	"github.com/h2non/filetype"
 	"github.com/mholt/archiver/v4"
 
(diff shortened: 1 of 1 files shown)


rgmz:
I do think that skipping binary files would be a mistake. I recently discovered credentials leaked in an large binary file inside of an archive.

Perhaps it could be toggleable?

0x1:
implementation looks good. what was the reason for skipping binary? we can pull out strings from them to scan, no?

ahrav:
Scanning binaries provides significant value, so having it on by default makes sense in most cases. However, scanning very large binary repositories can sometimes cause the scanner to get stuck processing an impractical amount of data. 

We recently had an instance where 100+ GBs of binaries across a monorepo caused scans to halt. To avoid these edge cases, we should implement adaptive limits that adjust scan scope dynamically in response to the data encountered. This would restrict binary scanning if specific thresholds around repository size, file count or other heuristics are exceeded.

For example, logic could restrict binary scanning if thresholds are exceeded around repository size, file count, file types, or other heuristics. This would allow scans to remain practical, while still extracting insights from source code and typical binaries.
  
Given this, I have converted the PR to draft for now and want to rethink the current implementation and approach before submitting. Please let me know if you have any other suggestions or ideas around effectively balancing default binary analysis with practical safeguards.

PR #2205, opened 2023-12-11
Check for SourceUnit support dynamically in the SourceManager

mcastorina:
### Description:
Calls a function to determine whether we should use source units at runtime to allow for feature flagging. This can be expanded in the future to allow logic based on the source being ran, however until the need arises it has no knowledge of it.

### Checklist:

### DIFF
--- a/pkg/sources/source_manager.go
+++ b/pkg/sources/source_manager.go
@@ -236,8 +236,8 @@ func (s *SourceManager) run(ctx context.Context, source Source, report *JobProgr
 		"source_type", source.Type().String(),
 	)
 	// Check for the preferred method of tracking source units.
-	useSourceUnits := len(targets) == 0 && s.useSourceUnitsFunc != nil && s.useSourceUnitsFunc()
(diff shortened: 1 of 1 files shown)


pkg/sources/source_manager.go
```diff
@@ -24,7 +24,8 @@ type SourceManager struct {
 	// Max number of units to scan concurrently per source.
 	concurrentUnits int
 	// Run the sources using source unit enumeration / chunking if available.
-	useSourceUnits bool
+	// Checked at runtime to allow feature flagging.
+	useSourceUnitsFunc func() bool
```
ahrav, line 28: question: This is mainly for future proofing?

ahrav (approved):

pkg/sources/source_manager.go
mcastorina: No, it's a function so it can be called at runtime / updated dynamically, instead of only at initialization.

PR #2206, opened 2023-12-11
Update GitParse logic to handle edge case.

### LINKED ISSUE #2109: GitParse Error: `failed to parse line "--- /dev/null\n" after state "ModeLine"`

### Description

I encountered the following error in the wild:
```
2023-11-12T10:29:01-05:00       error   trufflehog      Recovering at the latest commit or diff...
        {"job_id": 1, "source_manager_worker_id": "navoz", "source_type": "SOURCE_TYPE_GITHUB", "source_name": "trufflehog - github", "error": "failed to parse line \"--- /dev/null\n\" after state \"ModeLine\" (commit=fd6e99e7a80199b76a694603be57c5ade1de18e7)"}
```

I don't know the specific repository this belongs to. I will investigate and see what the issue might be.

Edit:

`--- /dev/null` after ModeLine suggests that it went directly `ModeLine->FromLine` without an IndexLine. This wasn't accounted for in #1486, and hasn't been encountered before as far as I can tell.

https://github.com/trufflesecurity/trufflehog/blob/737d6b764d8cf827d39d6dff5e881fa0b15e5734/pkg/gitparse/gitparse.go#L614-L615

rgmz:
### Description:

This PR has a few changes.

- Fixes #2109.
- Passes the repository origin URL to gitparse (if present) . This was necessary to identify the cause of #2109 and is probably a good change that should be implemented.
- Minor refactoring/code clean up

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -18,11 +18,12 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/google/go-github/v42/github"
-	diskbufferreader "github.com/trufflesecurity/disk-buffer-reader"
 	"golang.org/x/oauth2"
 	"golang.org/x/sync/semaphore"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
+
+	diskbufferreader "github.com/trufflesecurity/disk-buffer-reader"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/cleantemp"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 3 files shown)


pkg/sources/git/git.go
```diff
@@ -615,10 +620,16 @@ func (s *Git) gitChunk(ctx context.Context, diff gitparse.Diff, fileName, email,
...
+	remoteURL := getSafeRemoteURL(repo, "origin")
+	var repoCtx context.Context
+	if remoteURL != "" {
+		repoCtx = context.WithValue(ctx, "repo", remoteURL)
+	} else {
+		repoCtx = ctx
+	}
 
-	commitChan, err := gitparse.NewParser().Staged(ctx, path)
+	commitChan, err := gitparse.NewParser().Staged(repoCtx, path)
```
rgmz, line 632: Idk if there's value in including the remote url when scanning staged changes. Presumably this will only be called in the context of a single repository, so the culprit should be obvious.

ahrav:
This one is a little out of my wheelhouse, so i'll defer to @zricethezav and @bill-rich

pkg/sources/git/git.go
```diff
@@ -459,22 +459,27 @@ func (s *Git) CommitsScanned() uint64 {
 const gitDirName = ".git"
 
 func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string, scanOptions *ScanOptions, reporter sources.ChunkReporter) error {
-	commitChan, err := gitparse.NewParser().RepoPath(ctx, path, scanOptions.HeadHash, scanOptions.BaseHash == "", scanOptions.ExcludeGlobs, scanOptions.Bare)
+	// Get the remote URL for reporting (may be empty)
+	remoteURL := getSafeRemoteURL(repo, "origin")
+	var repoCtx context.Context
+	if remoteURL != "" {
+		repoCtx = context.WithValue(ctx, "repo", remoteURL)
```
bill-rich, line 466: We could use the path if there is no `remoteURL` so that the context always has some repo info.

bill-rich (changes requested):
Just one minor request/question.

bill-rich (approved):

pkg/sources/git/git.go
rgmz: I've been meaning to update this. Not sure when I marked it as resolved.

rgmz: Done.

PR #2207, opened 2023-12-12
Fix azurestorage detector

0x1:
### Description:
Refactoring Azure Storage detector and fixing bug
- bugfix: accountID (which is not unique) was used as raw, leading to secrets being overwritten and flipflopping; updated to use key as raw instead
- some accountIDs may no longer exist. leading to host not being found err in request. this should denote an invalid key. added check and test case for this.
- updated test secrets. all tests should be passing now
- refactor: standard refactor, pulling out verification logic, fixing verification error test cases, using http values instead of hard coded values, etc.

### Checklist:

### DIFF
--- a/pkg/detectors/azurestorage/azurestorage.go
+++ b/pkg/detectors/azurestorage/azurestorage.go
@@ -98,13 +98,15 @@ func verifyAzureStorageKey(ctx context.Context, client *http.Client, accountName
 	}
 	defer res.Body.Close()
 
-	if res.StatusCode == http.StatusOK {
+	switch res.StatusCode {
+	case http.StatusOK:
 		return true, nil
-	} else if res.StatusCode != http.StatusForbidden { // 403 if account id or key is invalid, or if account is disabled
+	case http.StatusForbidden:
+		// 403 if account id or key is invalid, or if the account is disabled
+		return false, nil
+	default:
 		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
 	}
-
-	return false, nil
 }
 
 func (s Scanner) Type() detectorspb.DetectorType {
--- a/pkg/detectors/azurestorage/azurestorage.go
+++ b/pkg/detectors/azurestorage/azurestorage.go
@@ -51,7 +51,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_AzureStorage,
-			Raw:          []byte(accountName),
+			Raw:          []byte(accountKey),
 		}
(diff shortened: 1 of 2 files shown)


pkg/detectors/azurestorage/azurestorage.go
```diff
@@ -44,47 +51,15 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_AzureStorage,
-			Raw:          []byte(accountName),
+			Raw:          []byte(accountKey),
```
0x1, line 54: this is the main fix for bug

rosecodym: is the account name sensitive information? if not, i think we should add it to `ExtraData` (like we do for aws keys). that could be helpful in support cases and just generally useful to users.

pkg/detectors/azurestorage/azurestorage_test.go
```diff
@@ -112,6 +119,25 @@ func TestAzurestorage_FromChunk(t *testing.T) {
...
+			want: func() []detectors.Result {
+				r := detectors.Result{
+					DetectorType: detectorspb.DetectorType_AzureStorage,
+					Verified:     false,
+				}
+				r.SetVerificationError(fmt.Errorf("unexpected HTTP response status 404"), secret)
+				return []detectors.Result{r}
+			}(),
+			wantErr:             false,
+			wantVerificationErr: true,
```
rosecodym, line 131: same question as that other pr: is this flag necessary?

pkg/detectors/azurestorage/azurestorage.go
0x1: no it's not sensitive, that's a good idea

pkg/detectors/azurestorage/azurestorage_test.go
0x1: nope

ahrav (approved):

pkg/detectors/azurestorage/azurestorage.go
```diff
@@ -93,6 +71,42 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	if err != nil {
+		// If the host is not found, we can assume that the accountName is not valid
+		if strings.Contains(err.Error(), "no such host") {
+			return false, nil
+		}
+		return false, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusOK {
```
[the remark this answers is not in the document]
0x1: agreed, that is a bit nicer. done.

PR #2208, opened 2023-12-12
fix and refactor browserstack detector

0x1:
### Description:
Fixing a few issues with Browserstack Detector:
- standard refactor
- Previously `[._-]+` in the user pattern was updated to `[._-]?` because some userIDs do not contain this. However it could be the case that some userIDs contain multiple such symbols in the middle, and this could be causing some users to not detect tokens. updating to use `[._-]*` instead
- Browserstack (via Cloudflare) has been blocking verification requests. Go doesn't handle cookies by default. Adding cookie jar. Adding test case to check this in the future.
- removing `common.SaneHttpClient()` bc for some reason it yields requests getting blocked even with cookiejar.
image

### Checklist:

### DIFF
--- a/pkg/detectors/browserstack/browserstack.go
+++ b/pkg/detectors/browserstack/browserstack.go
@@ -102,7 +102,6 @@ func verifyBrowserStackCredentials(ctx context.Context, client *http.Client, use
 	}
 	req.Header.Add("Content-Type", "application/json")
 	req.SetBasicAuth(username, accessKey)
-	fmt.Println(username, accessKey)
 
 	res, err := client.Do(req)
 	if err != nil {
--- a/pkg/detectors/browserstack/browserstack.go
+++ b/pkg/detectors/browserstack/browserstack.go
@@ -3,13 +3,15 @@ package browserstack
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
+	"net/http/cookiejar"
 	"regexp"
 	"strings"
 
-	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+	"golang.org/x/net/publicsuffix"
 )
 
 type Scanner struct {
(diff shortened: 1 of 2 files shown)


pkg/detectors/browserstack/browserstack_test.go
```diff
@@ -101,13 +104,37 @@ func TestBrowserStack_FromChunk(t *testing.T) {
...
 					Verified:     false,
 					RawV2:        []byte(fmt.Sprintf("%s%s", secret, secretUser)),
-				},
+				}
+				r.SetVerificationError(fmt.Errorf("unexpected HTTP response status 404"), secret)
+				results := []detectors.Result{r}
+				return results
+			}(),
+			wantErr:             false,
+			wantVerificationErr: true,
```
rosecodym, line 118: It looks like you can delete the flag in this test case because you're setting the verification error explicitly in the test case - can you delete the flag for all of them?

pkg/detectors/browserstack/browserstack.go
```diff
@@ -22,11 +24,9 @@ var _ detectors.Detector = (*Scanner)(nil)
 const browserStackAPIURL = "https://www.browserstack.com/automate/plan.json"
 
 var (
-	defaultClient = common.SaneHttpClient()
```
rosecodym, line 25: As a future maintainer, I'd find it helpful to see a comment (somewhere) explaining that `SaneHttpClient` doesn't appear to work for unknown reasons (even with the cookie jar attached).

pkg/detectors/browserstack/browserstack_test.go
0x1: yep. nice catch

pkg/detectors/browserstack/browserstack.go
```diff
@@ -22,11 +24,9 @@ var _ detectors.Detector = (*Scanner)(nil)
...
-
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"hub-cloud.browserstack.com", "accessKey", "\"access_Key\":", "ACCESS_KEY", "key", "browserstackKey", "BS_AUTHKEY", "BROWSERSTACK_ACCESS_KEY"}) + `\b([0-9a-zA-Z]{20})\b`)
-	userPat = regexp.MustCompile(detectors.PrefixRegex([]string{"hub-cloud.browserstack.com", "userName", "\"username\":", "USER_NAME", "user", "browserstackUser", "BS_USERNAME", "BROWSERSTACK_USERNAME"}) + `\b([a-zA-Z\d]{3,18}[._-]?[a-zA-Z\d]{6,11})\b`)
+	userPat = regexp.MustCompile(detectors.PrefixRegex([]string{"hub-cloud.browserstack.com", "userName", "\"username\":", "USER_NAME", "user", "browserstackUser", "BS_USERNAME", "BROWSERSTACK_USERNAME"}) + `\b([a-zA-Z\d]{3,18}[._-]*[a-zA-Z\d]{6,11})\b`)
```
ahrav, line 29: Nice find!

pkg/detectors/browserstack/browserstack.go
```diff
@@ -89,16 +102,26 @@ func verifyBrowserStackCredentials(ctx context.Context, client *http.Client, use
 	}
 	req.Header.Add("Content-Type", "application/json")
 	req.SetBasicAuth(username, accessKey)
+	fmt.Println(username, accessKey)
```
ahrav, line 105: We should probably remove this. 🤣

ahrav (changes requested):
Requesting changes to remove the print statement. Otherwise LGTM!

pkg/detectors/browserstack/browserstack.go
0x1: rip. thanks for catching that

ahrav (approved):

PR #2211, opened 2023-12-13
Convert Shortcut detector to tri-state verification

rgmz:
### Description:
This converts the Shortcut detector to use tri-state verification. It also (implicitly) removes `FromData` returning errors directly, which periodically caused issues scanning chunks.

```
2023-11-07T03:44:34-05:00	error	trufflehog	error scanning chunk	{"detector_worker_id": "JeY04", "timeout": 10, "error": "Error making request: Get \"https://api.app.shortcut.com/api/v3/member\": dial tcp: lookup api.app.shortcut.com: i/o timeout"}
```

### Checklist:

### DIFF
--- a/pkg/detectors/shortcut/shortcut.go
+++ b/pkg/detectors/shortcut/shortcut.go
@@ -2,7 +2,6 @@ package shortcut
 
 import (
 	"context"
-	"github.com/go-errors/errors"
 	"net/http"
 	"regexp"
 	"strings"
@@ -35,7 +34,6 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	dataStr := string(data)
 
 	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
-
 	for _, match := range matches {
 		if len(match) != 2 {
 			continue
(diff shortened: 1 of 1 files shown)


pkg/detectors/shortcut/shortcut.go
```diff
@@ -48,39 +46,47 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-		results = append(results, s1)
+	verifiedBodyResponse, err := common.ResponseContainsSubstring(res.Body, "name")
+	_ = res.Body.Close()
+	if err != nil {
+		return false, err
+	}
+
+	if res.StatusCode >= 200 && res.StatusCode < 300 && verifiedBodyResponse {
+		return true, nil
+	} else {
```
ahrav, line 87: nit: We can likely remove the `else` here and just `return false, nil`

ahrav (approved):

pkg/detectors/shortcut/shortcut.go
rgmz: Updated that.

PR #2214, opened 2023-12-14
Encode '%' when generating Git URLs

rgmz:
### Description:
This fixes an issue where file paths that contain `%` would cause `UpdateLinkLineNumber` to break.

Example:
```
2023-11-03T07:40:37-04:00       error   trufflehog      unable to parse link to update line number      {"detector_worker_id": "PSF2d", "timeout": 10, "link": "https://github.com/GeekMasher/tree-sitter-hcl/blob/a7f23cc5795769262f5515e52902f86c1b768994/example/real_world_stuff/coreos/coreos%tectonic-installer%installer%frontend%ui-tests%output%metal.tfvars#L1", "error": "parse \"https://github.com/GeekMasher/tree-sitter-hcl/blob/a7f23cc5795769262f5515e52902f86c1b768994/example/real_world_stuff/coreos/coreos%tectonic-installer%installer%frontend%ui-tests%output%metal.tfvars\": invalid URL escape \"%te\""}
```

### Checklist:

### DIFF
--- a/pkg/giturl/giturl.go
+++ b/pkg/giturl/giturl.go
@@ -118,6 +118,10 @@ func NormalizeOrgRepoURL(provider provider, repoURL string) (string, error) {
 // Supports GitHub, GitLab, Bitbucket, and Azure Repos.
 // If the provider supports hyperlinks to specific lines, the line number will be included.
 func GenerateLink(repo, commit, file string, line int64) string {
+	// Some paths contain '%' which breaks |url.Parse| if not encoded.
+	// https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding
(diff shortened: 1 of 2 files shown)


ahrav (approved):
TIL, thanks.

mcastorina:
Nice! I think this solution can be improved to cover all escapes via [url.PathEscape](https://pkg.go.dev/net/url#PathEscape).

rgmz:
> Nice! I think this solution can be improved to cover all escapes via [url.PathEscape](https://pkg.go.dev/net/url#PathEscape).

That's a good point. I wasn't sure if additional escapes would break anything as I only encountered the issue with `%` specifically.

PR #2218, opened 2023-12-14
Update GitLab v1 verification to check for valid JSON response

### LINKED ISSUE #2099: GitLab detector should check response body content

#### Description

The [GitLab (old) detector](https://github.com/trufflesecurity/trufflehog/blob/703e158648c4eaf61de59a70a8f2e81fd7f48e6f/pkg/detectors/gitlab/gitlab.go#L90) should verify response headers and body content to avoid false-positives. 

Certain values for the authorization header will result in a `200` status code, however, the body is text/html and not JSON.

```sh
$ curl -i "https://gitlab.com/api/v4/user" -H "Authorization: Bearer Continuous-deployment"
HTTP/2 200
date: Mon, 06 Nov 2023 22:16:50 GMT
content-type: text/html; charset=utf-8
...

<!DOCTYPE html><html><head><meta charset=utf-8>...
```

rgmz:
### Description:
This fixes #2099.

### Checklist:

### DIFF
--- a/pkg/detectors/gitlab/gitlab.go
+++ b/pkg/detectors/gitlab/gitlab.go
@@ -2,7 +2,9 @@ package gitlab
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/detectors/gitlab/gitlab.go
```diff
@@ -18,9 +20,11 @@ type Scanner struct {
 }
 
 // Ensure the Scanner satisfies the interfaces at compile time.
-var _ detectors.Detector = (*Scanner)(nil)
-var _ detectors.EndpointCustomizer = (*Scanner)(nil)
-var _ detectors.Versioner = (*Scanner)(nil)
+var (
```
ahrav, line 23: Out of curiosity, do you prefer the current way of doing the compile time interface check, or
```
var _ interface {
	detectors.Detector
	detectors.EndpointCustomizer
	detectors.Versioner
} = (*Scanner)(nil)
```

I only ask, b/c we've seen all 3 forms in the codebase, and I have an urge to standardize them 🤣

ahrav (approved):

pkg/detectors/gitlab/gitlab.go
rgmz: Honestly, I hate all of them. Implicit interfaces feel like a mistake. 

With that being said, `_ detectors.Versioner = (*Scanner)(nil)` seems to be the most common pattern, but your alternative seems to be the most concise way to write it.

PR #2223, opened 2023-12-14
[feat] - leverage workers to handle commits concurrently

ahrav:
### Description:
Use concurrent workers to process commits from the commitChan to improve performance. This prevents blocking when scanning large binary files that could stall the current single-threaded approach.

### Checklist:


pkg/sources/git/git.go
```diff
@@ -466,78 +483,110 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
-			if commit.Hash == scanOptions.BaseHash {
-				logger.V(1).Info("reached base commit", "commit", commit.Hash)
-				break
+	return nil
+}
+
+func (s *Git) startWorkers(ctx context.Context, meta scanMetadata, worker workerFunc, commitChan <-chan gitparse.Commit) {
+	var wg sync.WaitGroup
+
+	for i := 0; i < runtime.NumCPU(); i++ {
```
rgmz, line 498: Should this respect the concurrency flag?

ahrav: Ahh yes, I think that makes sense. Good call.

pkg/sources/git/git.go
```diff
@@ -466,78 +483,110 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
 
-			fileName := diff.PathB
-			if fileName == "" {
-				continue
+			var depth int64
+
+			if len(meta.scanOptions.BaseHash) > 0 {
+				if commit.Hash == meta.scanOptions.BaseHash {
+					ctx.Logger().V(1).Info("reached base commit", "commit", commit.Hash)
+					break
```
mcastorina, line 525: I *think* this breaks out of the `select` and not the `for`. Is that intentional? Maybe some labels will help here.

pkg/sources/git/git.go
```diff
@@ -466,78 +483,110 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
+				}
+			}
+			if meta.scanOptions.MaxDepth > 0 && depth >= meta.scanOptions.MaxDepth {
+				ctx.Logger().V(1).Info("reached max depth", "depth", depth)
+				break
```
mcastorina, line 530: Same here.

pkg/sources/git/git_test.go
```diff
@@ -251,7 +251,8 @@ func TestSource_Chunks_Integration(t *testing.T) {
 				}
 
 				if expectedData, exists := tt.expectedChunkData[key]; !exists {
-					t.Errorf("A chunk exists that was not expected with key %q", key)
+					continue
+					// t.Errorf("A chunk exists that was not expected with key %q", key)
```
mcastorina, line 255: Accidental removal?

mcastorina (commented):
Just a few questions on the first pass. Going to take a closer look.

pkg/sources/git/git_test.go
ahrav: Oh no, i actually need to continue here since the original test was dependent on the ordering. Since we are using go routines we just need to check the final count.

pkg/sources/git/git.go
ahrav: Oh... actually i think we can just return since the initial implementation we broke out of the outer for and returned nil.

ahrav: I actually think we can simply return nil? Since [this break](https://github.com/trufflesecurity/trufflehog/blob/b0fab16ad4b5cc42c38ee4e2ee362aa3c47dea24/pkg/sources/git/git.go#L479C10-L479C10) would also return from the function, or is that not correct?

mcastorina: Yeah, returning `nil` looks like it would preserve the original functionality.

mcastorina (changes requested):
I think the original commit parsing relied on its sequential nature for control flow that doesn't easily translate to workers.

I'm only requesting changes because I'm confused by the control flow and not convinced it's correct, but I could be missing something.

pkg/sources/git/git.go
```diff
@@ -466,78 +483,110 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
-			if fileName == "" {
-				continue
+			var depth int64
+
+			if len(meta.scanOptions.BaseHash) > 0 {
+				if commit.Hash == meta.scanOptions.BaseHash {
+					ctx.Logger().V(1).Info("reached base commit", "commit", commit.Hash)
+					return nil
+				}
+			}
```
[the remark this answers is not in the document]
ahrav: Ahh yep, that's true. If we hit this we should actually signal to all the workers we are done.... wrap it up, call it a day. I'll adjust.

pkg/sources/git/git.go
```diff
@@ -466,78 +483,110 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
+			if meta.scanOptions.MaxDepth > 0 && depth >= meta.scanOptions.MaxDepth {
+				ctx.Logger().V(1).Info("reached max depth", "depth", depth)
+				return nil
```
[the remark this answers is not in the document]
ahrav: I guess we used to increment the depth for every commit pulled from the commit chan. That actually means, this approach might not work. I should look over the initial implementation to double check. Will convert to Draft in the meantime.

pkg/sources/git/git.go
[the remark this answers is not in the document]
mcastorina: Wouldn't doing it in the workers introduces a race condition, and we might scan more commits than asked?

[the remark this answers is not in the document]
ahrav: Yea, I think this might not be the best approach. I think we should have the workers simply handle the binary diffs. Leave the rest to be handled serially.

PR #2226, opened 2023-12-15
[feat] - Make skipping binaries configurable

ahrav:
### Description:
Add option to disable scanning of binary files from Git sources

### Checklist:


pkg/handlers/archive.go
```diff
@@ -115,6 +119,33 @@ func (a *Archive) openArchive(ctx logContext.Context, depth int, reader io.Reade
...
+		}
+
+		// Create a new reader that starts with the buffer we just read
+		// and continues with the rest of the original reader.
+		reader = io.MultiReader(bytes.NewReader(buffer[:n]), reader)
+
+		mime := mimetype.Detect(buffer)
+		mimeT := mimeType(mime.String())
+
+		fmt.Println(mimeT)
```
mcastorina, line 136: Debug print?

pkg/handlers/archive.go
```diff
@@ -115,6 +119,33 @@ func (a *Archive) openArchive(ctx logContext.Context, depth int, reader io.Reade
...
+
+		if mimeT == machOType || mimeT == octetStream {
+			ctx.Logger().V(5).Info("skipping binary file", "ext", mimeT)
+			return nil
+		}
+
+		if common.SkipFile(mime.Extension()) {
+			ctx.Logger().V(5).Info("skipping file", "ext", mimeT)
+			return nil
+		}
```
mcastorina, line 146: Why do we do this only if `skipBinaries` is set?

pkg/handlers/archive.go
```diff
@@ -161,9 +192,11 @@ func (a *Archive) extractorHandler(archiveChan chan []byte) func(context.Context
...
 
-		if common.SkipFile(f.Name()) {
-			lCtx.Logger().V(5).Info("skipping file", "filename", f.Name())
-			return nil
+		if a.skipBinaries {
+			if common.SkipFile(f.Name()) {
+				lCtx.Logger().V(5).Info("skipping file", "filename", f.Name())
+				return nil
+			}
 		}
```
mcastorina, line 200: Same question here, why was this changed to only skip files if `skipBinaries` is set?

pkg/handlers/archive.go
```diff
@@ -222,6 +255,8 @@ type mimeType string
 const (
 	arMimeType  mimeType = "application/x-unix-archive"
 	rpmMimeType mimeType = "application/x-rpm"
+	machOType   mimeType = "application/x-mach-binary"
+	octetStream mimeType = "application/octet-stream"
```
mcastorina, line 259: What about ELF or EXE files? Are those covered by `application/octet-stream`?

pkg/sources/git/git.go
```diff
@@ -58,14 +58,15 @@ type Git struct {
...
+	skipBinaries       bool
 }
 
 type metrics struct {
 	commitsScanned uint64
 }
 
 func NewGit(sourceType sourcespb.SourceType, jobID sources.JobID, sourceID sources.SourceID, sourceName string, verify bool, concurrency int,
-	sourceMetadataFunc func(file, email, commit, timestamp, repository string, line int64) *source_metadatapb.MetaData,
+	sourceMetadataFunc func(file, email, commit, timestamp, repository string, line int64) *source_metadatapb.MetaData, skipBinaries bool,
```
mcastorina, line 69: This function just keeps growing 🤣 

Not something to fix in this PR of course, but I'm looking forward to refactoring this.

pkg/handlers/archive.go
ahrav: Oh i forgot, we had added a bunch of binary extensions to that function, but that probably doesn't make much sense anymore. I'll update that function to remove binary extensions.

pkg/handlers/archive.go
ahrav: updated

pkg/handlers/archive.go
ahrav: Some of the reading i did they showed up as octet-stream, but i also read that they can be `application-x-elf`. Just to be safe, I added additional extensions to the list of known binary extensions. There might be a more robust way of determining if the file is a binary format.

pkg/sources/git/git.go
ahrav: Yea... it was a bit painful adding another param :(

pkg/common/vars.go
```diff
@@ -58,3 +90,13 @@ func SkipFile(filename string) bool {
...
+
+// IsBinary returns true if the file extension is in the binaryExtensions list.
+func IsBinary(filename string) bool {
+	for _, ext := range binaryExtensions {
+		if strings.TrimPrefix(filepath.Ext(filename), ".") == ext {
+			return true
+		}
+	}
+	return false
+}
```
mcastorina, line 102: **Nit:** In another PR we should do a map lookup (or even better?, a switch case).

pkg/common/vars.go
```diff
@@ -58,3 +90,13 @@ func SkipFile(filename string) bool {
 	}
 	return false
 }
+
+// IsBinary returns true if the file extension is in the binaryExtensions list.
+func IsBinary(filename string) bool {
+	for _, ext := range binaryExtensions {
+		if strings.TrimPrefix(filepath.Ext(filename), ".") == ext {
```
mcastorina, line 97: **Nit:** case-insensitive comparison

```go
fileExt := strings.TrimPrefix(filepath.Ext(filename), ".")
if strings.EqualFold(ext, fileExt) {
    // ...
}
```

pkg/handlers/archive.go
```diff
@@ -115,6 +119,31 @@ func (a *Archive) openArchive(ctx logContext.Context, depth int, reader io.Reade
...
 func (a *Archive) handleNonArchiveContent(ctx logContext.Context, reader io.Reader, archiveChan chan []byte) error {
+	buffer := make([]byte, 512)
+	n, err := reader.Read(buffer)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return fmt.Errorf("unable to read file for MIME type detection: %w", err)
+	}
+
+	// Create a new reader that starts with the buffer we just read
+	// and continues with the rest of the original reader.
+	reader = io.MultiReader(bytes.NewReader(buffer[:n]), reader)
```
mcastorina, line 130: Would using [bufio.Peek](https://pkg.go.dev/bufio#Reader.Peek) be equivalent here? I find that a little easier to understand than reading a buffer then resetting it with a multireader.

```go
bufReader := bufio.NewReaderSize(reader, 512)
buffer, err := bufReader.Peek(512)
if err != nil && !errors.Is(err, io.EOF) {
    // ...
}
```

pkg/handlers/archive.go
```diff
@@ -115,6 +119,31 @@ func (a *Archive) openArchive(ctx logContext.Context, depth int, reader io.Reade
...
+
+	mime := mimetype.Detect(buffer)
+	mimeT := mimeType(mime.String())
+
+	if common.SkipFile(mime.Extension()) {
+		ctx.Logger().V(5).Info("skipping file", "ext", mimeT)
+		return nil
+	}
```
mcastorina, line 138: Does this actually work? `SkipFile` expects a full filename and we're just passing the extension.

mcastorina (approved):

pkg/handlers/archive.go
ahrav: According to the documentation `.Extension()` returns:
```
// Extension returns the file extension associated with the MIME type.
// It includes the leading dot, as in ".html". When the file format does not
// have an extension, the empty string is returned.
```

pkg/common/vars.go
ahrav: nah, i'm fix this right now lol. I like the map too. I'll also post benchies.

ahrav: Screenshot 2023-12-15 at 10 41 03 AM

#microptimizations 🥳

pkg/handlers/archive.go
ahrav: beautiful, thank you!

ahrav: Ahh weird... with that change the tests fail. I'll dig into why.

mcastorina: Hm.. did you replace all the `reader` variables with `bufReader`? That was the only downside I could think of with this approach.

ahrav:
```
// Peek returns the next n bytes without advancing the reader. The bytes stop
// being valid at the next read call
```

mcastorina: I take that to mean the slice returned by `Peek` becomes invalid (since it's probably returning the internal buffer to avoid a copy). Do we need the buffer to remain valid after the next read?

proto/sources.proto
```diff
@@ -226,6 +229,7 @@ message GitHub {
   bool includePullRequestComments = 14;
   bool includeIssueComments = 15;
   bool includeGistComments = 16;
+  bool skip_binaries = 17;
```
rgmz, line 232: Slightly off topic but what is the convention for protos? Snake case, camel case, both?

ahrav: Argh, i should've probably just used camel case, even though the style guide does recommend snake.
https://protobuf.dev/programming-guides/style/#message-field-names

ahrav: I think I can actually update all of them to be snake_case. I'm fairly certain that wouldn't be a breaking change. I'll double check though.

mcastorina: I thought the idea with protos was that the name doesn't matter, just the number. So I agree changing them shouldn't be a breaking change. The generated Go code would probably be the same too (hopefully).

rgmz: Fair enough. I was curious because I was about to add a new field and realized there was an inconsistency.

mcastorina: Oh, yeah. I would also like to know (and stick to) whatever the convention is! Just commenting on the renaming idea.

Thanks for pointing out the inconsistency!

PR #2191, opened 2023-12-08
Propagate TruffleHog context to handlers

rgmz:
### Description:

Handlers currently use the built-in `context` package rather than `github.com/trufflesecurity/trufflehog/v3/pkg/context` (logContext). This results in vague errors that are impossible to trace, for instance:

```
2023-12-07T17:19:25-05:00       error   trufflehog      error unarchiving chunk.    {"error": "handling file: file1.txt: runtime error: index out of range [0] with length 0"}
```
https://github.com/trufflesecurity/trufflehog/blob/2a7813929b3a4d17c2bc6aab09ae1a95a066b4de/pkg/handlers/archive.go#L72-L81

This PR changes the code to pass logContext instead, resulting in traceable errors. Notably, it was not possible to figure out what file was causing a runtime panic (crashing TruffleHog) without this change.
```
2023-12-07T17:19:25-05:00       error   trufflehog      error unarchiving chunk.        {"source_manager_worker_id": "V1eMi", "job_id": 1, "source_id": 1, "source_name": "trufflehog - github", "source_type": "SOURCE_TYPE_GITHUB", "commit": "37ef02d1818ae263956b7c8bc702b85cdbc83d20", "path": "chromium/chrome/test/data/safe_browsing/rar/passwd.rar", "timeout": 30, "error": "handling file: file1.txt: runtime error: index out of range [0] with length 0"}
```

Two thoughts:
1. It may be easier to remove all imports of `context` and replace them with `github.com/trufflesecurity/trufflehog/v3/pkg/context`. This awkwardly changes `Handler.FromFile` but not `Handler.IsFileType`.
2. This does make the errors more verbose. Is it necessary to include all the fields or could we create pass a new context that only contains important information?

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #2228, opened 2023-12-15
Pass logging context to `Detector.FromData`

rgmz:
### Description:

This PR touches a lot of files, however, the sole change is replacing `import ... "context"` with `import ... "github.com/trufflesecurity/trufflehog/v3/pkg/context"`.

**Why?**
While the logging context is not used in any detectors (as far as I'm aware), it is incredibly useful for local debugging, similar to #2191. Right now it is difficult to figure out what specific chunk triggered a detector, and in many instances logging is just done with `fmt.Printf` which is devoid of any context. Perhaps there is an easier or better way to accomplish this — let me know.

An alternative is to change `Detector.FromData` to  `Detector.FromChunk` (#1517). However, that's a larger change that could be accomplished via less invasive means (#1741).

### Checklist:


mcastorina:
First off, thanks for taking the time and effort for this change.

> Perhaps there is an easier or better way to accomplish this — let me know.

Would using `logContext.AddLogger` accomplish what you're doing?

https://github.com/trufflesecurity/trufflehog/blob/b0fab16ad4b5cc42c38ee4e2ee362aa3c47dea24/pkg/context/context.go#L162-L170

In the detector you could do something like the below when needed for debugging. Since we already pass a log context into the detectors, it should preserve the key/value pairs.

```go
logger := logContext.AddLogger(ctx).Logger()
```

rgmz:
That's bizarre: I swear that didn't work before but it does now. Perhaps I accidentally mixed in a call to `context` instead of the TruffleHog context which doesn't preserve values ([Slack thread](https://trufflehog-community.slack.com/archives/CK3UYAQF2/p1701891895819369)). 😵

mcastorina:
Gah, sorry I missed that thread! Yeah, I think the log info would get lost if it's wrapped in a stdlib context. I wonder if there's a way to traverse the ancestry of contexts instead of only looking at the child..

PR #2220, opened 2023-12-14
Improve GitHub scan logging

rgmz:
### Description:
This fixes #2066.

It is an opinionated change to remove `scanning repo %d/%d`. I experimented with a few potential solutions but could not find anything that created a desirable result. I think having the count when a repository has _finished_ scanning is the most comprehensible option.

- Logging "scanning 30/150" (iteration #) is confusing because loop order != order things are processed. You can end up with "scanning 110/150", "scanning "97/150", "scanning "97/150", "scanning 30/150", etc.
- I also tried creating a second atomic `uint64` to track started vs. scanned separately results. This was _better_ but still confusing as you'd end up with "scanning repo 101/150", "scanned repo 136/150", "scanned repo 137/150".

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #2233, opened 2023-12-16
Scan GitHub wikis

rgmz:
### Description:
GitHub wikis are another potential source for secrets. 

This PR adds the functionality to scan wikis. It is enabled by default because wikis are just repositories that you can clone, and aren't subject to any rate limits.

### Checklist:


pkg/sources/github/github.go
```diff
@@ -770,46 +777,36 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
-				logger.V(2).Info(fmt.Sprintf("scanned %d/%d repos", scanned, len(s.repos)), "repo_size", repoSize, "duration_seconds", time.Since(start).Seconds())
-			}(now)
-
-			if err = s.git.ScanRepo(ctx, repo, path, s.scanOptions, sources.ChanReporter{Ch: chunksChan}); err != nil {
-				scanErrs.Add(fmt.Errorf("error scanning repo %s: %w", repoURL, err))
-				return nil
+				_, err := s.cloneAndScanRepo(wikiCtx, installationClient, wikiURL, chunksChan)
+				if err != nil {
+					scanErrs.Add(err)
+					return nil
```
rosecodym, line 797: is there any chance that the repo comments are still scannable even if wiki scanning fails?

pkg/engine/github.go
```diff
@@ -21,9 +21,11 @@ func (e *Engine) ScanGitHub(ctx context.Context, c sources.GithubConfig) error {
 		ScanUsers:                  c.IncludeMembers,
 		IgnoreRepos:                c.ExcludeRepos,
 		IncludeRepos:               c.IncludeRepos,
+		IncludeForks:               c.IncludeForks,
 		IncludeIssueComments:       c.IncludeIssueComments,
 		IncludePullRequestComments: c.IncludePullRequestComments,
 		IncludeGistComments:        c.IncludeGistComments,
+		IncludeWikis:               c.IncludeWikis,
```
rosecodym, line 28: where is `IncludeWikis` actually checked?

rosecodym (commented):
Thanks for this! I have some minor questions about logging, but my big question is my silliest: Where is the new flag actually checked? I feel like I'm missing something dumb.

pkg/engine/github.go
rgmz: Oh dear.

pkg/sources/github/github.go
rgmz: That's a good point.

pkg/engine/github.go
rgmz: An oversight on my part. It should be checked now in `repo.go`.

pkg/sources/git/git.go
```diff
@@ -474,12 +475,21 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
+	logger := ctx.Logger()
+	if ctx.Value("repo") == nil {
+		logger = logger.WithValues("repo", urlMetadata)
+	}
+	if scanOptions.BaseHash != "" {
+		logger = logger.WithValues("base", scanOptions.BaseHash)
+	}
+	if scanOptions.HeadHash != "" {
+		logger = logger.WithValues("head", scanOptions.HeadHash)
+	}
```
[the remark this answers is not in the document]
rgmz: Quite the opposite, I was trying to exclude those values unless they're actually specified. Evidently I was tired when I pushed this.

pkg/sources/git/git.go
```diff
@@ -722,18 +735,22 @@ func (s *Git) ScanRepo(ctx context.Context, repo *git.Repository, repoPath strin
...
 	}
 
+	logger := ctx.Logger()
 	// We're logging time, but the repoPath is usually a dynamically generated folder in /tmp.
 	// To make this duration logging useful, we need to log the remote as well.
-	remotes, _ := repo.Remotes()
-	repoURL := "Could not get remote for repo"
-	if len(remotes) != 0 {
-		repoURL = getSafeRemoteURL(repo, remotes[0].Config().Name)
+	// However, the context may already contain a key for "repo".
```
rosecodym, line 741: What are the circumstances under which this happens? _Should_ it be happening?

pkg/sources/github/github.go
```diff
@@ -342,6 +343,12 @@ func checkGitHubConnection(ctx context.Context, client *github.Client) error {
 }
 
 func (s *Source) visibilityOf(ctx context.Context, repoURL string) (visibility source_metadatapb.Visibility) {
+	// It isn't possible to get the visibility of a wiki.
+	// We must use the visibility of the corresponding repository.
+	if strings.HasSuffix(repoURL, ".wiki.git") {
+		repoURL = strings.TrimSuffix(repoURL, ".wiki.git") + ".git"
```
rosecodym, line 349: Does this happen under "normal" usage? As far as I can tell, this logic will only be executed if someone ties to scan a wiki repo directly - it won't run for wikis picked up as part of regular repo scans. Do I have that right?

pkg/sources/github/github.go
```diff
@@ -770,46 +777,36 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
-			if err != nil {
-				scanErrs.Add(fmt.Errorf("error cloning repo %s: %w", repoURL, err))
-				return nil
-			}
-
-			s.setScanOptions(s.conn.Base, s.conn.Head)
-
-			repoSize := s.repoSizes.getRepo(repoURL)
-			logger.V(2).Info(fmt.Sprintf("scanning repo %d/%d", i, len(s.repos)), "repo_size_kb", repoSize)
+			// Scan the repository's wiki, if it has one.
```
rosecodym, line 789: The wiki is only scanned if it exists _and the scan was configured to look for it_. I can see myself being misled by this comment down the road. I'm not sure if the comment should be clarified or some logic changing elsewhere.

pkg/sources/github/github.go
```diff
@@ -59,6 +58,7 @@ type Source struct {
 	repos             []string
 	orgsCache         cache.Cache
 	filteredRepoCache *filteredRepoCache
+	repoWikiCache     map[string]struct{} // repositories that have `"has_wiki": true`
```
rosecodym, line 61: This is repositories that have `"has_wiki":true` _and_ the wiki was resolvable without redirecting, right? I can see myself being misled by this comment down the road - I think that just "repositories with wikis" would be sufficient, but maybe you want to be more precise.

rosecodym (commented):
please forgive my persnicketiness about comments - i'm trying to do my future self some favors

pkg/sources/github/github.go
rgmz: I believe it would get triggered in the following places via `github.cloneAndScanRepo` > `git.ScanRepo` > `git.ScanCommits`.

https://github.com/trufflesecurity/trufflehog/blob/28212c9a82221b1947674d5b5557486a7077d0c5/pkg/sources/git/git.go#L508-L510

https://github.com/trufflesecurity/trufflehog/blob/28212c9a82221b1947674d5b5557486a7077d0c5/pkg/sources/git/git.go#L529

pkg/sources/git/git.go
rgmz: The GitHub source passes a context with `repo` because (a) we already know the remoteURL, and (b) including that information in pre-scan logs might be useful. The GitLab source does not include this information but it's arguably useful to have. The Git source is written with the assumption that the remoteURL is attached (if available) in the `ScanRepo` function.

I don't know if I can say whether it _should_ be happening, but having duplicate keys is a bit annoying.

rgmz:
> please forgive my persnicketiness about comments - i'm trying to do my future self some favors

No need to apologize, these are all valid points. Some of these are remnants from prior iterations, such as when I thought that `"has_wiki": true` meant the repository has a wiki (😂🔫), or when I realized that it makes more sense to check if a wiki exists in `repo.go`.

rgmz:
Hey @rosecodym, do you have any other questions or feedback for this change?

pkg/sources/github/github.go
```diff
@@ -772,50 +780,35 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
+					wikiURL := strings.TrimSuffix(repoURL, ".git") + ".wiki.git"
+					wikiCtx := context.WithValue(ctx, "repo", wikiURL)
+
+					_, err := s.cloneAndScanRepo(wikiCtx, installationClient, wikiURL, chunksChan)
+					if err != nil {
+						scanErrs.Add(err)
+						// Don't return, it still might be possible to scan comments.
+					}
 				}
-				logger.V(2).Info(fmt.Sprintf("scanned %d/%d repos", scannedCount, len(s.repos)), "duration_seconds", time.Since(start).Seconds())
```
rosecodym, line 806: Does this no longer happen if the scan panics? That's a fine change (and your version is more readable), I just want to ensure I understand the diff.

pkg/sources/github/github.go
```diff
@@ -833,6 +826,41 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
+func (s *Source) cloneAndScanRepo(ctx context.Context, client *github.Client, repoURL string, chunksChan chan *sources.Chunk) (time.Duration, error) {
+	var duration time.Duration
+
+	ctx.Logger().V(2).Info("attempting to clone repo")
+	path, repo, err := s.cloneRepo(ctx, repoURL, client)
+	defer func(path string) {
+		if path != "" {
+			_ = os.RemoveAll(path)
+		}
+	}(path)
```
rosecodym, line 838: Thank you for aggressively preserving the order of operations here, but I think you can just move `defer os.RemoveAll(path)` to after the error check :) (@mcastorina that's what you did elsewhere for something, right?)

rosecodym (commented):
Ah, apologies for letting this get away from me, and thanks for the reminder. The only non-clarifying question I have is about a new flag check (marked inline).

pkg/sources/github/github.go
rosecodym: ah yes, a similar refactor for gitlab in #2345

pkg/sources/github/github.go
```diff
@@ -772,50 +780,35 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
-				scanFailed = true
-				return nil
 			}
 
-			if err = s.scanComments(ctx, repoURL, chunksChan); err != nil {
-				scanErrs.Add(fmt.Errorf("error scanning comments in repo %s: %w", repoURL, err))
-				scanFailed = true
-				return nil
+			// Scan comments, if enabled.
+			if s.includeGistComments || s.includeIssueComments || s.includePRComments {
```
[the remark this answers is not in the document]
rgmz: I added an explicit check because I think it's more consistent and readable to have `if includeWiki { doWiki } ... if includeComments { doComments }`.

The `s.includeGistComments`, `s.includeIssueComments`, and `s.includePRComments` weren't being _ignored_, but they were implemented in a way that's awkward and difficult to understand (imo). I also don't think it makes sense to enter a function that can return errors when the flags are disabled.

https://github.com/trufflesecurity/trufflehog/blob/9867ce8eb8366405f4ff97670a897da6139966c7/pkg/sources/github/github.go#L1104-L1107

https://github.com/trufflesecurity/trufflehog/blob/9867ce8eb8366405f4ff97670a897da6139966c7/pkg/sources/github/github.go#L1179-L1182

pkg/sources/github/github.go
rgmz: > Does this no longer happen if the scan panics?

These changes were made based on how the code looked before #2220 was merged. Now that #2220 has been merged, the diff is a bit awkward.

Your understanding is correct. In fact, since @dustin-decker moved the metric and `scannedCount` increment to the end of the function, I should move this logging statement as well.

pkg/sources/github/github.go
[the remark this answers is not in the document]
rosecodym: Oof, strong agree. Thanks for this.

rosecodym (approved):
This looks great, thanks for doing it!

PR #2235, opened 2023-12-16
[chore] - add additional binary extensions to skip

ahrav:
### Description:
Add additional binary extensions to skip and pdf.

### Checklist:

### DIFF
--- a/pkg/common/vars.go
+++ b/pkg/common/vars.go
@@ -28,6 +28,7 @@ var (
 		"wav",
 		"flac",
 		"webp",
+		"pdf",
 
 		// images
 		"png",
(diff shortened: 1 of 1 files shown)


pkg/common/vars.go
```diff
@@ -79,6 +80,10 @@ var (
 		"vxd":    {}, // Virtual device driver in Windows
 		"sfx":    {}, // Self-extracting archive
 		"bundle": {}, // Mac OS X application bundle
+		"pyo":    {}, // Compiled Python file
+		"pyc":    {}, // Compiled Python file
+		"whl":    {}, // Python wheel file
```
rgmz, line 85: Wheels are just zip files.

ahrav: whoooops, one auto-complete too many. 😓

pkg/common/vars.go
```diff
@@ -28,6 +28,7 @@ var (
 		"wav",
 		"flac",
 		"webp",
+		"pdf",
```
rosecodym, line 31: how sure are we about this? these can presumably be opened up and inspected, right? do we get false positives from them or something?

dustin-decker: I've never seen a result in a PDF unless it has gone through a PDF handler (not yet released)

dustin-decker (approved):

PR #2237, opened 2023-12-16
[chore] - leverage bufio.Peek to determine mimetype

ahrav:
### Description:
@mcastorina suggested using `bufio.Peek` to determine mimetype for an earlier PR which I found to be a great solution. So I've updated the existing mimetype detection function to also use bufio.Peek. Consolidated the code into that single function.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -14,7 +14,6 @@ import (
 	"time"
 
 	"github.com/gabriel-vasile/mimetype"
-	"github.com/h2non/filetype"
 	"github.com/mholt/archiver/v4"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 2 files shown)


pkg/handlers/archive.go
```diff
@@ -434,25 +436,26 @@ func (a *Archive) handleNestedFileMIME(ctx logContext.Context, tempEnv tempEnv,
 
 // determineMimeType reads from the provided reader to detect the MIME type.
 // It returns the detected MIME type and a new reader that includes the read portion.
-func determineMimeType(reader io.Reader) (mimeType, io.Reader, error) {
+func determineMimeType(reader io.Reader) (*mimetype.MIME, io.Reader, error) {
+	bufReader := bufio.NewReaderSize(reader, mimeTypeBufferSize)
```
rosecodym, line 440: is creating a new reader so expensive that re-using this one makes sense? the downside of doing it this way is that now all readers have to use the mime-detection buffer size, which might not always be the best choice.

ahrav:
> @ahrav is this still relevant?

Nope. Will close, thanks.

PR #2238, opened 2023-12-16
[chore] - use snake_case for naming

ahrav:
### Description:
Use snake_casing for protobuf message names.
https://protobuf.dev/programming-guides/style/#message-field-names

### Checklist:


rosecodym:
Isn't renaming proto fields a thing you're basically never supposed to do?

rgmz:
Context: https://github.com/trufflesecurity/trufflehog/pull/2226#discussion_r1428564939

ahrav:
> Isn't renaming proto fields a thing you're basically never supposed to do?

I believe in this case since the change is only the casing the generated files are identical.
https://github.com/trufflesecurity/trufflehog/blob/928cd4413c487e4b81eab2f30e2f3c0e2a87aaa7/proto/sources.proto#L209-L209
https://github.com/trufflesecurity/trufflehog/blob/928cd4413c487e4b81eab2f30e2f3c0e2a87aaa7/pkg/pb/sourcespb/sources.pb.go#L1619-L1619

proto/sources.proto
```diff
@@ -224,11 +224,11 @@ message GitHub {
   bool includeForks = 8;
```
dustin-decker, line 224: one more

dustin-decker (approved):

dustin-decker:
>Isn't renaming proto fields a thing you're basically never supposed to do?

The potential impact here is with the generated code and unmarshaling/marshaling data. Here it looks like it impacts the JSON field names but we are not using those yet. The slot numbers and the data within them remain the same, so data will still go over the wire in the same way.

PR #2222, opened 2023-12-14
Enhance HuggingFace extra data

rgmz:
### Description:
This adds more meaningful data to the HuggingFace detector.

Example:
```
Found verified result 🐷🔑
Detector Type: HuggingFace
Decoder Type: BASE64
Raw result: hf_...
Username: jdoe
Email: jdoe@example.com
Token: deploy (write)
Organizations: contoso:write, TruffleSecurity:read
File: /tmp/example/data.yml
Line: 26

```

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #2240, opened 2023-12-17
[chore] - Add regex and keyword for api_org tokens

ahrav:
### Description:
The HuggingFace API also has tokens that begin with api_org, so i've updated the regex to detect both.

### Checklist:


rgmz:
I considered this. They're technically being deprecated and are read-only, but they still exist in the wild.

It does complicate #2222 because the auth struct is different for org tokens.

ahrav:
> I considered this. They're technically being deprecated and are read-only, but they still exist in the wild.
> 
> It does complicate #2222 because the auth struct is different for org tokens.

That's a good point. I think i'm going to merge your PR, and come back to adjust this.

pkg/detectors/huggingface/huggingface.go
```diff
@@ -22,13 +22,13 @@ var _ detectors.Detector = (*Scanner)(nil)
...
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\bhf_[a-zA-Z0-9]{34}\b`)
+	keyPat = regexp.MustCompile(`\b(?:hf_|api_org_)[a-zA-Z0-9]{34}\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"huggingface", "hugging_face", "hf"} // Huggingface docs occasionally use "hf" instead of "huggingface"
+	return []string{"huggingface", "hugging_face", "hf_", "api_org_"} // Huggingface docs occasionally use "hf" instead of "huggingface"
```
rgmz, line 31: While we're here, since `hf_` and `api_org_` are both distinct prefixes there isn't really a point in having these two keywords. A regex with `hugginface` but no `hf_` will never match.
```suggestion
	return []string{"hf_", "api_org_"}
```

rosecodym (approved):
lgtm once you make that keyword change

PR #2241, opened 2023-12-18
Update workflow to check out refs from forks

rgmz:
### Description:
The `test` check currently fails on PRs from forks because it attempts to fetch the branch from this repository:
```
Run actions/checkout@v4
Syncing repository: trufflesecurity/trufflehog
...
# https://github.com/trufflesecurity/trufflehog/pull/2234/checks
Fetching the repository
Determining the checkout info
  /usr/bin/git branch --list --remote origin/patch-1
  /usr/bin/git tag --list patch-1
  Error: A branch or tag with the name 'patch-1' could not be found
```

This seems to be a known problem. The fix appears to be specifying a `repository`, though I'm not sure the best way to specify it for both PRs and main.
https://github.com/actions/checkout/issues/551

### Checklist:


dustin-decker:
Thanks for taking a look at this. I approved the run, but it doesn't look like it fixed it.

rgmz:
Seems like a mistake on my part. The failing workflow is actually scanning for secrets — the stage just happens to be called "test."

rgmz:
@dustin-decker I believe this should work now.

zricethezav:
I would feel better updating this PR so that we have two different jobs, one for forks and one for the trufflesecurity org. The forked job would use the trufflehog action [as defined in the readme](https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#advanced-usage). 

Reason being for this request is a malicious actor could put some nasty code in their fork of trufflehog which would then have access to our GHA runners. A simple `fmt.Println(os.Environ())` for example. I think there is a reason GitHub doesn't encourage checking out forks by default.

joeleonjr:
> I would feel better updating this PR so that we have two different jobs, one for forks and one for the trufflesecurity org. The forked job would use the trufflehog action [as defined in the readme](https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#advanced-usage).
> 
> Reason being for this request is a malicious actor could put some nasty code in their fork of trufflehog which would then have access to our GHA runners. A simple `fmt.Println(os.Environ())` for example. I think there is a reason GitHub doesn't encourage checking out forks by default.
> 
> edit: for reference https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests

I agree with @zricethezav. At the moment, the PR code doesn't appear exploitable. However, a couple small modifications in the future and we would potentially leak internal secrets from the GH runner. This isn't worth the risk IMO. Two separate jobs seems more secure.

zricethezav:
Going to close this PR for now to just be extra cautious. Feel free to reopen when appropriate changes have been made

rgmz:
> I would feel better updating this PR so that we have two different jobs, one for forks and one for the trufflesecurity org. The forked job would use the trufflehog action [as defined in the readme](https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#advanced-usage).
> 
> Reason being for this request is a malicious actor could put some nasty code in their fork of trufflehog which would then have access to our GHA runners. A simple `fmt.Println(os.Environ())` for example. I think there is a reason GitHub doesn't encourage checking out forks by default.
> 
> edit: for reference https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests

I'm indifferent towards the solution, though I will point out that risk isn't exclusive to the secrets workflow. Hence why workflow runs from external contributors require approval.

That blog post predates GitHub adding basic abuse protections like workflow approvals.
https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

zricethezav:
"approve and run" is hardly protection. People make honest mistakes or don't read changes carefully all the time

rgmz:
No argument there.

PR #2244, opened 2023-12-19
Adds basic if/else check if pid slice is empty

codevbus:
### Description:
Makes sure that there are detected trufflehog pids before attempting to clean up tempdirs.

### Checklist:

### DIFF
--- a/pkg/cleantemp/cleantemp.go
+++ b/pkg/cleantemp/cleantemp.go
@@ -70,31 +70,35 @@ func CleanTempArtifacts(ctx logContext.Context) error {
 	}
 
 	for _, artifact := range artifacts {
-		if trufflehogRE.MatchString(artifact.Name()) {
-			// Mark these artifacts initially as ones that should be deleted.
-			shouldDelete := true
-			// Check if the name matches any live PIDs.
-			for _, pidval := range pids {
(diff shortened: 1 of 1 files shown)


pkg/cleantemp/cleantemp.go
```diff
@@ -70,31 +70,35 @@ func CleanTempArtifacts(ctx logContext.Context) error {
...
-			// Mark these artifacts initially as ones that should be deleted.
-			shouldDelete := true
-			// Check if the name matches any live PIDs.
-			for _, pidval := range pids {
-				if strings.Contains(artifact.Name(), fmt.Sprintf("-%s-", pidval)) {
-					shouldDelete = false
-					break
+		if len(pids) == 0 {
+			ctx.Logger().Info("No trufflehog processes were found")
+		} else {
```
ahrav, line 75: suggestion: If we add a continue after the log, we can avoid the else and reduce the indentation.
```
if len(pids) == 0 {
    ctx.Logger()....
    continue
}
if trufflehogRE.... { ..}
```

ahrav (approved):

pkg/cleantemp/cleantemp.go
ahrav: [style guide ref](https://google.github.io/styleguide/go/decisions#indent-error-flow)

PR #2254, opened 2023-12-22
Update to Sourcegraph Access token format

shivasurya:
### Description:

We at Sourcegraph have updated the token format recently. Check out the detailed [secret format documentation](https://docs.sourcegraph.com/dev/security/secret_formats).

There are three format supported as of now, here are the ways you could test it.

1. Legacy `SOURCEGRAPH_V1` - generate the access token and use the `last part` alone from the format `sgp_identifier_lastpart`. Remove the prefix sgp & identifier for the sake of testing.
2. Legacy `SOURCEGRAPH_v2` - generate the access token and use the format as `sgp_lastpart`. Remove the identifier for the sake of testing.
3. Current format `SOURCEGRAPH_V3` - generate the access token and use as it is. `sgp_identifer_lastpart`

Let me know if you have any questions. Thank you.

### Checklist:


pkg/detectors/sourcegraph/sourcegraph.go
```diff
@@ -22,7 +22,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	defaultClient = common.SaneHttpClient()
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(sgp_[a-f0-9]{40})\b`)
+	keyPat = regexp.MustCompile(`\b(sgp_(?:[a-fA-F0-9]{16}|local)_[a-fA-F0-9]{40}|sgp_[a-fA-F0-9]{40}|[a-fA-F0-9]{40})\b`)
```
rgmz, line 25: This new detector matches the following patterns, right?
```
sgp_0000000000000000_0000000000000000000000000000000000000000
sgp_local_0000000000000000000000000000000000000000
sgp_0000000000000000000000000000000000000000
0000000000000000000000000000000000000000
```

If so, it might be good to split the last one out and add a [prefix](https://github.com/trufflesecurity/trufflehog/blob/a98a4de2fb5c74eeecf8438ff1b7e42b2d3021df/pkg/detectors/tru/tru.go#L25). I don't think it would get flagged right now unless the chunk also happens to include the `sgp_` keyword.

ahrav (approved):
Thank you @shivasurya for taking the time to up this. Really appreciated. Additionally, thanks for building such awesome products as well 😃 I'm a huge fan.

PR #2256, opened 2023-12-22
Skip all binaries

bill-rich:
Skip all handling of binary files when skipBinaries is set.

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -1005,6 +1005,11 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 		return nil
 	}
 
+	if s.skipBinaries {
+		fileCtx.Logger().V(5).Info("skipping binary file", "path", path)
+		return nil
+	}
+
 	var handlerOpts []handlers.Option
 	if s.skipBinaries {
 		handlerOpts = append(handlerOpts, handlers.WithSkipBinaries(true))
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -1011,13 +1011,6 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 	}
 
 	var handlerOpts []handlers.Option
-	if s.skipBinaries {
-		handlerOpts = append(handlerOpts, handlers.WithSkipBinaries(true))
-		if common.IsBinary(path) {
-			fileCtx.Logger().V(5).Info("skipping binary file")
-			return nil
-		}
-	}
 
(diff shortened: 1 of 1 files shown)


pkg/sources/git/git.go
```diff
@@ -1005,6 +1005,11 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 		return nil
 	}
 
+	if s.skipBinaries {
+		fileCtx.Logger().V(5).Info("skipping binary file", "path", path)
+		return nil
+	}
+
 	var handlerOpts []handlers.Option
 	if s.skipBinaries {
```
rosecodym, line 1014: can you delete this block now that it's a no-op?

rosecodym (changes requested):

pkg/sources/git/git.go
bill-rich: Probably. I'm just not sure of why it was done this way originally. Wanted check with @ahrav, but he is out today, and I was trying to get this fix in fast.

rosecodym: i spoke with him about this yesterday and he said it was just an oversight, so i'm trying to not compound the damage :)

dustin-decker: I'm guessing that he just wanted to skip binaries and considered archives separately. I think the filtering is flawed though because it is based on file extension and unix binaries don't usually have one.

dustin-decker (approved):
LGTM after removing the no-op block below

pkg/sources/git/git.go
rosecodym: oh, sorry, i'm not trying to resolve the question functionality in this pr. (it's a good question, i just wasn't trying to do that here.) my comment was: i'm pretty sure this block is dead, and i don't want to merge dead code. (maybe i'm just missing something that makes it non-dead though.)

rosecodym (approved):

pkg/sources/git/git.go
```diff
@@ -1005,15 +1005,13 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
-		handlerOpts = append(handlerOpts, handlers.WithSkipBinaries(true))
-		if common.IsBinary(path) {
-			fileCtx.Logger().V(5).Info("skipping binary file")
-			return nil
-		}
+		fileCtx.Logger().V(5).Info("skipping binary file", "path", path)
+		return nil
 	}
 
+	var handlerOpts []handlers.Option
```
trufflesteeeve, line 1013: __optional nitpick__: Probably best to delete this and it's only usage on line 1052/1050, just so a future reader won't think we're actually using handlerOpts at the moment.

trufflesteeeve (approved):

PR #2260, opened 2023-12-22
Use directory iterator instead of walkdir

dustin-decker:
### Description:
Walkdir was recursively walking tmp, this iterates /tmp and nothing further

### Checklist:

### DIFF
--- a/pkg/cleantemp/cleantemp.go
+++ b/pkg/cleantemp/cleantemp.go
@@ -76,7 +76,7 @@ func CleanTempArtifacts(ctx logContext.Context) error {
 	defer dir.Close()
 
 	for {
-		entries, err := dir.Readdir(1) // read only one entry
+		entries, err := dir.ReadDir(1) // read only one entry
 		if err != nil {
 			if err == io.EOF {
 				break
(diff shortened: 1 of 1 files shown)


pkg/cleantemp/cleantemp.go
```diff
@@ -69,25 +69,39 @@ func CleanTempArtifacts(ctx logContext.Context) error {
 	}
 
 	tempDir := os.TempDir()
-	err = filepath.WalkDir(tempDir, func(path string, d fs.DirEntry, err error) error {
+	dir, err := os.Open(tempDir)
+	if err != nil {
+		return fmt.Errorf("error opening temp dir: %w", err)
```
codevbus, line 74: should this be set to a log-level for `--debug`

codevbus (approved):

pkg/cleantemp/cleantemp.go
trufflesteeeve: I think because it's reading just the temp directory itself that it's fine to have a full error here. If the top level directory can't be read, we likely have a bit of a problem.

trufflesteeeve (approved):
Just a question and a suggestion.

pkg/cleantemp/cleantemp.go
```diff
@@ -69,25 +69,39 @@ func CleanTempArtifacts(ctx logContext.Context) error {
...
+	}
+	defer dir.Close()
+
+	for {
+		entries, err := dir.Readdir(1) // read only one entry
```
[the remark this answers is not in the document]
dustin-decker: Oh, thanks for the call out. Looks like ReadDir is recommended and has the same functionality here.

PR #2261, opened 2023-12-24
Update stripe detector regex

NikhilPanwar:
### Description:
The regex secretKey = regexp.MustCompile(`[rs]k_live_[a-zA-Z0-9]{20,30}`) works only for keys generated before 2019, keys created after 2019 can be of legth upto 255 characters  

https://support.stripe.com/questions/new-api-keys-are-longer-than-existing-keys#:~:text=As%20per%20our%20API%20guidelines,to%20255%20characters%20in%20length.


ahrav (approved):
Thank for updating this ❤️

PR #2264, opened 2023-12-28
Overhaul npm detector

### LINKED ISSUE #1455: NpmToken detector should verify against the registry URL, if present
### Community Note

### Description

The NpmToken v1 & v2 detectors are hard-coded to verify tokens against `https://registry.npmjs.org/-/whoami`:

https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/pkg/detectors/npmtoken/npmtoken.go#L53

https://github.com/trufflesecurity/trufflehog/blob/a99d89d71110b1111bb7f7724067fa8a43a2319e/pkg/detectors/npmtokenv2/npmtokenv2.go#L53

However, npm can be used with registries other than the default (`registry.npmjs.org`), meaning that valid tokens can inadvertently be marked as unverified. This is especially likely in enterprise environments which often have internal instances of Nexus Repository, Artifactory, or other artifact repository managers, and do not use `registry.npmjs.org` directly..

## Problem to be Addressed

The NpmToken detectors should verify the tokens against the relevant registry/registries, where possible.

NpmTokens are _often_ found in the `.npmrc` configuration file, where they are required to be scoped to a specific registry. 

For example, the following NpmTokens have enough contextual information for TruffleHog to determine what registries they should be checked against instead of `registry.npmjs.org`:
``` 
//artifactory.example.com/:_authToken=NpmToken.18b533de-b347-46bc-8d2e-ae6550c6a16a
//nexus.example.com/repository/npm-proxy/:_authToken=NpmToken.105d5de9-17d0-4a08-a5e9-3e5d7070225b
```
https://docs.npmjs.com/cli/v9/configuring-npm/npmrc#auth-related-configuration

### References

N/A

rgmz:
### Description:
This fixes #1455. It matches new npm tokens (`npm_xxx...`), old npm tokens (`NpmToken.0000-...`, `0000-...`), and "non-standard" tokens such as Artifactory using a JWT or GitHub packages using a PAT. Basic auth is a separate can of worms, so is out of scope for this PR.

The high-level flow for all detectors is:
```mermaid
flowchart TD
    A[Got token] -->|Search for associated URL| B{URL found?}
    B -->|Yes| C[Verify against URL]
    B -->|No| D[Search for any potential URLs]
    D --> E[Verify against all URLs]
    
    C-->Z[END]
    E-->Z
```

### Checklist:

### DIFF
--- a/pkg/detectors/npm/common.go
+++ b/pkg/detectors/npm/common.go
@@ -1,0 +1,161 @@
+package npm
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+)
+
+var defaultClient = common.SaneHttpClient()
+
+type npmScanner struct {
+	client *http.Client
+}
+
+// verifyToken attempts to verify a |token| by finding the associated registry URL in |data|.
+// It returns three values:
+//  1. whether the token is valid
+//  2. data associated with the token
+//  3. any errors encountered during verification
+func (s npmScanner) verifyToken(ctx context.Context, data string, token string) (bool, map[string]string, error) {
+	if s.client == nil {
+		s.client = defaultClient
+	}
+
+	registry := findTokenRegistry(data, token)
+	if registry != nil {
+		// A high confidence match was found, attempt to verify the token against it.
+		// e.g., |token|="s3cret" and |data| contains "//npm.company.com/:_authToken=s3cret".
+		// TODO: Handle multiple high confidence matches
+		return doVerification(ctx, s.client, registry, token)
+	} else {
+		// A high confidence match was not found.
+		// Attempt to verify the token against any registries we can find.
+		var (
+			registries = findAllRegistryURLs(data)
+			errs       = make([]error, 0, len(registries))
+
+			verified  bool
+			extraData map[string]string
+			err       error
+		)
+		for _, registry := range registries {
+			verified, extraData, err = doVerification(ctx, s.client, registry, token)
+			if verified {
+				return true, extraData, err
+			}
+			if err != nil {
+				errs = append(errs, err)
+			}
+		}
+		return false, nil, errors.Join(errs...)
+	}
+}
+
+// Most repositories implement a "whoami" endpoint
+// that returns the username of the authenticated user.
+type whoamiResponse struct {
+	Username string `json:"username"`
+}
+
+// doVerification checks whether |token| is valid for the given |registry|.
+func doVerification(ctx context.Context, client *http.Client, registry *registryInfo, token string) (bool, map[string]string, error) {
+	// Construct and send request.
+	scheme := registry.Scheme.Prefix()
+	if registry.Scheme == unknown {
+		scheme = isHttps.Prefix()
+	}
+	reqUrl := fmt.Sprintf("%s%s/-/whoami", scheme, registry.Uri)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqUrl, nil)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to construct request: %s", err)
+	}
+
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+	res, err := client.Do(req)
+	if err != nil {
+		// A |tls.RecordHeaderError| likely means that the server is using HTTP, not HTTPS.
+		// TODO: Is it possible to handle the reverse case?
+		var tlsErr tls.RecordHeaderError
+		if errors.As(err, &tlsErr) && registry.Scheme == isHttps {
+			r := *registry
+			r.Scheme = isHttp
+			return doVerification(ctx, client, &r, token)
+		}
+		return false, nil, fmt.Errorf("request to %s failed: %w", reqUrl, err)
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	// Handle the response.
+	if res.StatusCode == http.StatusOK {
+		body, _ := io.ReadAll(res.Body)
+		whoamiRes := whoamiResponse{}
+		if err := json.Unmarshal(body, &whoamiRes); err != nil {
+			if json.Valid(body) {
+				return false, nil, fmt.Errorf("failed to decode response %s: %w", reqUrl, err)
+			} else {
+				// If the response isn't JSON it's highly unlikely to be valid.
+				return false, nil, nil
+			}
+		}
+
+		// It is possible for the response to be `{"username": null}`, `{"username":""}`, etc.
+		// While a valid token _can_ return an empty username, the registry is likely returning 200 for invalid auth.
+		// TODO: Write a test for this.
+		if whoamiRes.Username == "" ||
+			(registry.RegistryType == nexusRepo3 && strings.HasPrefix(whoamiRes.Username, "anonymous")) ||
+			(registry.RegistryType == jetbrains && whoamiRes.Username == "internal") {
+			req.Header.Del("Authorization")
+			res2, err := client.Do(req)
+			if err != nil {
+				return false, nil, fmt.Errorf("request failed for %s: %w", reqUrl, err)
+			}
+			_, _ = io.Copy(io.Discard, res.Body)
+			_ = res2.Body.Close()
+
+			if res2.StatusCode == http.StatusOK {
+				return false, nil, nil
+			}
+		}
+
+		data := map[string]string{
+			"registry_type":  registry.RegistryType.String(),
+			"registry_url":   registry.Uri,
+			"username":       whoamiRes.Username,
+			"rotation_guide": "https://howtorotate.com/docs/tutorials/npm/",
+		}
+		return true, data, nil
+	} else if res.StatusCode == http.StatusUnauthorized ||
+		(registry.RegistryType == github && res.StatusCode == http.StatusForbidden) {
+		// Token is not valid.
+		return false, nil, nil
+	} else {
+		// Here be dragons.
+		return false, nil, fmt.Errorf("unexpected response status %d for %s", res.StatusCode, reqUrl)
+	}
+}
+
+// firstNonEmptyMatch returns the index and value of the first non-empty match.
+// If no non-empty match is found, it will return: 0, "".
+func firstNonEmptyMatch(matches []string, skip int) (int, string) {
+	if len(matches) < skip {
+		return 0, ""
+	}
+	// The first index is the entire matched string.
+	for i, val := range matches[skip:] {
+		if val != "" {
+			return i + skip, val
+		}
+	}
+	return 0, ""
+}
--- a/pkg/detectors/npm/npm_token_generic.go
+++ b/pkg/detectors/npm/npm_token_generic.go
@@ -1,0 +1,77 @@
+package npm
+
+import (
+	"context"
+	"regexp"
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type ScannerGeneric struct {
+	npmScanner
+}
+
+// Ensure the Scanner satisfies the interfaces at compile time.
+var _ interface {
+	detectors.Detector
+	detectors.Versioner
+} = (*ScannerGeneric)(nil)
+
+func (s ScannerGeneric) Version() int { return 0 }
+
+// genericKeyPat should match all possible values for .npmrc auth tokens.
+// TODO: Ensure this works with Yarn and UPM configurations.
+var genericKeyPat = regexp.MustCompile(`(?:_authToken|(?i:npm[_\-.]?token))['"]?[ \t]*[=:]?(?:[ \t]*['"]?)?([a-zA-Z0-9\-_.+=/]{5,})`)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s ScannerGeneric) Keywords() []string {
+	return []string{"_authToken", "npm_token", "npm-token", "npm.token"}
+}
+
+// FromData will find and optionally verify secrets in a given set of bytes.
+func (s ScannerGeneric) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	// Deduplicate results for more efficient handling.
+	tokens := make(map[string]struct{})
+	for _, match := range genericKeyPat.FindAllStringSubmatch(dataStr, -1) {
+		t := match[1]
+		// Ignore results that can be handled by the v1 or v2 detectors.
+		if strings.HasPrefix(t, "NpmToken.") || strings.HasPrefix(t, "npm_") {
+			continue
+		}
+		tokens[t] = struct{}{}
+	}
+	if len(tokens) == 0 {
+		return
+	}
+
+	// Iterate through results.
+	for token := range tokens {
+		s1 := detectors.Result{
+			DetectorType: s.Type(),
+			Raw:          []byte(token),
+		}
+
+		if verify {
+			verified, extraData, vErr := s.verifyToken(ctx, dataStr, token)
+			s1.Verified = verified
+			s1.ExtraData = extraData
+			s1.SetVerificationError(vErr)
+		}
+
+		results = append(results, s1)
+	}
+	return
+}
+
+func (s ScannerGeneric) Type() detectorspb.DetectorType {
+	return detectorspb.DetectorType_NpmToken
+}
+
+func (s ScannerGeneric) Description() string {
+	return "NPM tokens are used to authenticate and publish packages to the npm registry."
+}
--- a/pkg/detectors/npm/npm_token_generic_test.go
+++ b/pkg/detectors/npm/npm_token_generic_test.go
@@ -1,0 +1,256 @@
+//go:build detectors
+// +build detectors
+
+package npm
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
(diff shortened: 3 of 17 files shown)


pkg/detectors/npm/npm_token_v1_test.go
```diff
@@ -0,0 +1,202 @@
...
+				got[i].Raw = nil
+			}
+			if diff := pretty.Compare(got, tt.want); diff != "" {
+				t.Errorf("NpmTokenV1.FromData() %s diff: (-got +want)\n%s", tt.name, diff)
+			}
+		})
+	}
+}
+
+func BenchmarkFromDataV1(benchmark *testing.B) {
```
rgmz, line 188: The main downside of consolidating everything under one package is that you can't have `BenchmarkFromData` multiple times.

pkg/detectors/npm/registry.go
```diff
@@ -0,0 +1,300 @@
...
+func (scheme scheme) Prefix() string {
+	return [...]string{
+		"",
+		"http://",
+		"https://",
+	}[scheme]
+}
+
+// A collection of known registry implementations.
+type registryType int
```
rgmz, line 42: These are all that came to mind. I'm probably missing a few mainstream options.

pkg/detectors/npm/registry.go
```diff
@@ -0,0 +1,300 @@
...
+func parseKnownRegistryURI(data string, registryUri string) *registryInfo {
+	matches := knownRegistryPat.FindStringSubmatch(registryUri)
+	if len(matches) == 0 {
+		return nil
+	}
+
+	// Skip the first two indices: 1 is the entire string, 2 is the protocol.
+	index, uri := firstNonEmptyMatch(matches, 2)
+	info := &registryInfo{
+		RegistryType: registryType(index - 1),
```
rgmz, line 217: This is spooky black magic that relies on `knownRegistryPat` having the same number of capture groups as `registryType` has values. Unfortunately, it doesn't seem possible to programmatically get enum values so this can't be verified at compile time.

pkg/engine/ahocorasick/ahocorasickcore.go
```diff
@@ -1,9 +1,10 @@
-package engine
+package ahocorasick
```
rgmz, line 1: This had to be moved out of engine, otherwise it was not possible to add keyword detection to the detector tests.

pkg/detectors/npm/npm_token_generic.go
```diff
@@ -0,0 +1,78 @@
...
+	npmScanner
+}
+
+// Ensure the Scanner satisfies the interfaces at compile time.
+var _ interface {
+	detectors.Detector
+	detectors.Versioner
+} = (*ScannerGeneric)(nil)
+
+func (s ScannerGeneric) Version() int { return 0 }
```
rgmz, line 22: Version 0 might not be the most idiomatic way to approach this.

rgmz (commented):
This went through several iterations so there may be some leftover cruft or confusing bits.

pkg/detectors/npm/registry.go
rosecodym: Would it be feasible to create some sort of `registry` struct that contains a pattern and a name and a const int, and then smoosh them all together into a list in one place? That way there wouldn't have to be these multiple lists maintained across multiple files. (If not, then I think some loud comments in all the places that are magically correlated are warranted.)

pkg/engine/engine.go
```diff
@@ -581,11 +582,11 @@ func (e *Engine) notifyResults(ctx context.Context) {
...
-		key := fmt.Sprintf("%s%s%s%+v", r.DetectorType.String(), r.Raw, r.RawV2, r.SourceMetadata)
-		if val, ok := e.dedupeCache.Get(key); ok {
-			if res, ok := val.(detectorspb.DecoderType); ok && res != r.DecoderType {
-				continue
-			}
+		key := fmt.Sprintf("%s%s%s", r.DetectorType.String(), r.Raw, r.RawV2)
+		if _, ok := e.dedupeCache.Get(key); ok {
+			//if res, ok := val.(detectorspb.DecoderType); ok && res != r.DecoderType {
+			continue
+			//}
```
rosecodym, line 589: What are these changes?

pkg/detectors/npm/common.go
```diff
@@ -0,0 +1,161 @@
...
+		for _, registry := range registries {
+			verified, extraData, err = doVerification(ctx, s.client, registry, token)
+			if verified {
+				return true, extraData, err
+			}
+			if err != nil {
+				errs = append(errs, err)
+			}
+		}
+		return false, nil, errors.Join(errs...)
```
rosecodym, line 58: Tell me if I'm reading this right: Any verification errors returned from any registry will propagate to the final result, even if any (other) individual registry returns a determinate failure (i.e. no error.) This means that if _any_ registry reports indeterminate verification, then either the token will be verified or indeterminately unverified.

If I'm right, this is an issue, because it means that tokens can plausibly be impossible to determinately revoke. (A similar problem exists with the JDBC detector.)

rosecodym (commented):
This is a ton of work, thanks for diving into it. I added some questions inline.

pkg/detectors/npm/npm_token_generic.go
```diff
@@ -0,0 +1,78 @@
...
+
+// genericKeyPat should match all possible values for .npmrc auth tokens.
+// TODO: Ensure this works with Yarn and UPM configurations.
```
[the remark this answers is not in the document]
rgmz: I've done some further testing and having dedicated scanners for each pattern type (`.npmrc`, `.yarnrc`, `.upmconfig.toml`, etc.) seems to produce fewer false positives. However, I'm not sure how to best organize them. Are "versions" meant to have any semantic meaning/order? 

i.e., is it confusing to have one detector type with numerical versions that are completely unrelated (npm version 1 & 2 are siblings, version 3 is yarn, version 4 is rennovate, version 5 is a hypothetical update to the yarn config, etc.)

pkg/engine/engine.go
rgmz: Whoops, didn't mean to include this. I prefer only getting a single match when testing, it's easier to tell what matched vs didn't match.

pkg/detectors/npm/registry.go
rgmz: I considered doing that but was concerned that looping through _n_ patterns would be much less efficient. In the worst case, you'd need to loop through all the "known" patterns, find all "generic" registry matches, and do a secondary lookup to determine the scheme for each registry.

pkg/detectors/npm/common.go
```diff
@@ -0,0 +1,161 @@
...
+	reqUrl := fmt.Sprintf("%s%s/-/whoami", scheme, registry.Uri)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqUrl, nil)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to construct request: %s", err)
+	}
+
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+	res, err := client.Do(req)
+	if err != nil {
+		// A |tls.RecordHeaderError| likely means that the server is using HTTP, not HTTPS.
```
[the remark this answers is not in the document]
rgmz: In some circumstances it isn't possible to determine the scheme of the registry URI. For example, `//registry.host.com/npm/:_authToken=...` does not contain a scheme but we can do a secondary lookup to tell that it should be HTTP.
```
//registry.host.com/npm/:_authToken=...
registry=http://registry.host.com/npm/
```

If that information isn't present in the chunk then the scheme is `unknown`. HTTPS is a reasonable default but may not necessarily be correct. Trying to connect to a HTTP server with HTTPS seems to reliably result in a `tls.RecordHeaderError`; conversely, I haven't found a reliable indicator that you're to connect to a HTTPS port with HTTP.

We could limit the automation fallback to cases where the scheme is `unknown`. Personally, I don't think there's any risk in having it.

pkg/detectors/npm/common.go
```diff
@@ -0,0 +1,161 @@
...
+			req.Header.Del("Authorization")
+			res2, err := client.Do(req)
+			if err != nil {
+				return false, nil, fmt.Errorf("request failed for %s: %w", reqUrl, err)
+			}
+			_, _ = io.Copy(io.Discard, res.Body)
+			_ = res2.Body.Close()
+
+			if res2.StatusCode == http.StatusOK {
+				return false, nil, nil
```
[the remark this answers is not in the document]
rgmz: > If I got that correctly, then it took me a bit to understand it. (And if I didn't, then I definitely didn't understand it!) I think I would have figured it out faster if this logic were factored out into a named function, but maybe there's another approach that would clarify it as well.

Your understanding is correct, but the verification logic is probably going to get even more convoluted, so it would make sense to extract this into its own function.

Some registries do not implement the _whoami_ endpoint and return a 404. Others, like `npm.fontawesome.com`, have a no-op whoami endpoint that always returns `200 {"username": null}`. I have encountered dozens of Azure and Google registry URLs but they all return 404. I don't know if that means the tokens are invalid, the registry doesn't implement the endpoint, or the registry no longer exists.

pkg/detectors/npm/common.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+
+		data := map[string]string{
+			"registry_type":  registry.RegistryType.String(),
+			"registry_url":   registry.Uri,
```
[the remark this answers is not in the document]
rgmz: I don't think so. 

I _can_ see someone feeling uncomfortable with that information being stored on Truffle's servers, but other internal URLs (GitHub, Jenkins, Artifactory, etc.) are already necessarily included in scan result data.

pkg/detectors/npm/common.go
rgmz: Sort of. The 'happy path' is that a token is associated with a single known registry(1) and we get either a determinate verification or failure. If the token is associated with a single registry but it isn't a 'known' pattern(2), it's possible to encounter an indeterminate failure caused by implementation-specific differences.

```
# 1
//artifactory.corp.com/artifactory/api/npm/npm/:_authToken=*abcdef123456789*

# 2 .npmrc
//packages.corp.net/custom_layout/:_authToken=...
# 2 CLI flag
npm config set registry https://packages.corp.net/custom_layout/
```

The worst-case scenario is that a token cannot be correlated with a single registry, and there are multiple potential registries. In that case, it can be difficult to tell whether something determinately or indeterminately failed. Some implementations are extremely sensitive, with things as small as a trailing slash being the difference between a 200 API response and a generic 400 HTML error page.

So you can either report failures for all attempted URLs, or report none of them and just mark the token as unverified[^1]. Honestly, the detector that's most likely to trigger this right now is the existing npm v1 one that just searches for `npm.+{uuid}`. It's prone to false positives — even GitHub Advanced Security flags random UUIDs as npm tokens.


[^1]: prior versions had more logic to handle invalid registries (e.g., being redirected to https://landing.jfrog.com/reactivate-server/test or a HTML 404 page). That was a bit complicated to refactor so I ended up dropping it.

pkg/detectors/npm/npm_token_generic.go
mcastorina: iirc the way we implemented filtering depends on the version being non-zero, so you wouldn't be able to exclude/ignore just the `v0` detector.

mcastorina: I've added a test [here](https://github.com/trufflesecurity/trufflehog/pull/2272) to codify that assumption.

pkg/detectors/npm/registry.go
rosecodym: Hm, I'm either misunderstanding something or being unclear. I'm just proposing merging the three lists of registries you have (const ints, string names, and patterns) into one list of structs, each of which contains three fields. Why would that add any additional loops?

rgmz: Perhaps I'm misunderstanding. Can you provide an example of what you mean?

pkg/detectors/npm/common.go
```diff
@@ -0,0 +1,161 @@
...
+		// TODO: Write a test for this.
+		if whoamiRes.Username == "" ||
+			(registry.RegistryType == nexusRepo3 && strings.HasPrefix(whoamiRes.Username, "anonymous")) ||
+			(registry.RegistryType == jetbrains && whoamiRes.Username == "internal") {
+			req.Header.Del("Authorization")
+			res2, err := client.Do(req)
+			if err != nil {
+				return false, nil, fmt.Errorf("request failed for %s: %w", reqUrl, err)
+			}
+			_, _ = io.Copy(io.Discard, res.Body)
```
0x1, line 123: should this be `res2.Body`?

pkg/detectors/npm/registry.go
```diff
@@ -0,0 +1,300 @@
...
+// - Yarn (TODO)
+// - Unity Package Manager (TODO)
+func findTokenRegistry(data string, token string) *registryInfo {
+	// .npmrc stores auth as `//registry.com/path/:authToken=$TOKEN
+	// Therefore, we should be able to correlate a token to a registry with a high degree of confidence.
+	registryAuthPat := regexp.MustCompile(fmt.Sprintf(
+		// language=regexp
+		`(?i)(//%s(?:/[a-z0-9._-]+)*)/:(?:_auth(?:Token)?|_password).{1,20}%s`, hostPat, token))
+	matches := registryAuthPat.FindStringSubmatch(data)
+	if len(matches) == 0 {
```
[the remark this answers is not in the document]
rgmz: I don't believe so. If a regex pattern matches, it always returns a fixed number of match groups. If you have mutually exclusive match groups like `(?:(foo)|(bar))` then the non-matching group will just be an empty string.


```go
func main() {
	pat := regexp.MustCompile(`(?i)hello (?:(john)|(anne)|(world)|(.+))`)
	matches := pat.FindStringSubmatch("Hello Anne")
	fmt.Printf("Matches return %d groups: %s\n", len(matches), strings.Join(matches, ", "))
}

// Matches return 5 groups: Hello Anne, , Anne, , 
```
https://go.dev/play/p/cdPwcqZwDcT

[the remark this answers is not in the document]
rgmz: I suppose you could make the argument that it prevents failure, but if your pattern changes I think it's better to explicitly fail than silently break.

e.g., stuff like this will silently break if the pattern is updated to add another match group.
https://github.com/trufflesecurity/trufflehog/blob/5e78ad402e7f98817b18a93e4e5807ada53bad4d/pkg/detectors/aha/aha.go#L40-L42

pkg/detectors/npm/common.go
rgmz: Absolutely. 😅

pkg/detectors/npm/registry.go
rosecodym:
```
type registry struct {
    // maybe an int too, i don't know if it's necessary
    name    string
    pattern string
}

// somewhere, maybe at a top level
registries := []registry {
    {
        name: "npm",
        pattern: `(registry\.(?:npmjs\.org|yarnpkg\.com))`
    },
    {
        name: "artifactoryCloud",
        pattern: `([a-z0-9-]+\.jfrog\.io/(?:artifactory|[a-z0-9._-]+)/api/npm/[a-z][a-z0-9._-]+)`
    {,
    // etc
}
```

(This could also be some kind of map. I see that some of the patterns are pretty long and decomposable, so maybe they get built out of a function somehow.) Building the big monster pattern is a matter of looping through the slice and just joining all the patterns together. This would eliminate the spooky magic that ties together the big pattern and the registry list.

pkg/detectors/npm/npm_token_generic_test.go
```diff
@@ -0,0 +1,256 @@
...
+registry = https://npm.company.com/
+always-auth = true`,
+			expected: "dL4dfTOJSL8pijHrBFPKqp2bUwLGkVotezEn8dfTPe-Qa1cP",
+		},
+		"_authToken/scoped/artifactory": {
+			input: `registry=https://artifactory.example.com/artifactory/api/npm/npm/
+//artifactory.example.com/artifactory/api/npm/npm/:_authToken=eyJ2ZXIiOiIyIiwidHlwIjoiSldUIiwiYWxnIjoiUlMyNTYiLCJraWQiOiJSZ25DcnBEVXlKOV9yNElVRnNSU2hqU0E0aGpibEpjZ0M2bnJhN3ZqcTNNIn0.eyJzdWIiOiJqZnJ0QDAxY2pwdDc0N3ZyNHo0MTU4MHNiN3MxYW14XC91c2Vyc1wvYXJ0dXJvLmNhbXBvcyIsInNjcCI6ImFwcGxpZWQtcGVybWlzc2lvbnNcL2dyb3VwczpyZWFkZXJzLGRlcGxveS1kZXYtbnBtLGRlcGxveS1sb2NhbCIsImF1ZCI6ImpmcnRAMDFjanB0NzQ3dnI0ejQxNTgwc2I3czFhbXgiLCJpc3MiOiJqZnJ0QDAxY2pwdDc0N3ZyNHo0MTU4MHNiN3MxYW14XC91c2Vyc1wvYXJ0dXJvLmNhbXBvcyIsImV4cCI6MTY1NjAwNDUxOSwiaWF0IjoxNjU2MDAwOTE5LCJqdGkiOiJjOWZhM2VhNS01MzE0LTQwMzUtYjNiYy03OGNjMDJmYmM1NWMifQ.EIEPLzz9H2oQ3rKLKI_t1qtxi-G9ym6P6J5z7xWLsq79aHK3XV_E8_yuUK8giaOOdl_CaITOjl8Bt2aUVXTZKFKIOiQMLscBM4B6qGj_n4Qq8jiMwcKjnD_iWx8Wo-aNHHxgjvrRdWf-2UPIm6lSc77oZbUNAjhA5Q-W3uQRG7d50FGZpq_EEZsfbOcD7EMU2ZnvfYNTgTtmhZWfLefzB6xUF8WHgiDAVHJKQ2fKLX45Z9trc2SkKQmPBxaS-pBtKBhK15kQZ3x625KtLRr2ZwgOaJKHcg4SwuGOpyTF48nTk53SDorSj6fqlypTavVQRi-5cuSGTPrqLObk6lwpRg`,
```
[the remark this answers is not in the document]
rgmz: To answer your question: most of these were sourced from GitHub+Sourcegraph search. I did my best to 'mask' them by arbitrarily changing the host + secret.

PR #2272, opened 2024-01-05
[chore] Add test to check all versioned detectors are non-zero

mcastorina:
### Description:
Adds a test to check all versioned detectors return a non-zero value. This assumption was used to simplify detector filtering in the engine.

### Checklist:

### DIFF
--- a/pkg/engine/defaults_test.go
+++ b/pkg/engine/defaults_test.go
@@ -48,3 +48,20 @@ func TestDefaultDetectorTypesImplementing(t *testing.T) {
 		)
 	}
 }
+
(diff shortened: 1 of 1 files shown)


rosecodym:
Why is the secret scan finding things? Did the semi-recent update the action cause a misconfiguration?

ahrav (approved):

mcastorina:
> Why is the secret scan finding things? Did the semi-recent update the action cause a misconfiguration?

I couldn't figure that out either. My guess is it's doing a full git repo scan instead of just the changes in this PR.

PR #2276, opened 2024-01-07
[feat] - cache duplicate creds during detection

ahrav:
### Description:
This PR introduces caching for AWS credentials in the detector, aiming to minimize redundant API calls with identical credentials. The cache uses the rawV2 value as its key and a bespoke cacheItem for its value, enabling the reconstruction of complete metadata for each result, such as VerificationErr, verified, and ExtraData.

In practical testing with the [trufflehog repository](https://github.com/trufflesecurity/trufflehog), this implementation reduced AWS API calls significantly, from 707 to 256, while maintaining the same data processing volume, as indicated by unchanged chunk numbers. This approach holds potential for broader application across our most frequently used detectors, enhancing efficiency and performance.

### Checklist:

### DIFF
--- a/pkg/detectors/aws/aws.go
+++ b/pkg/detectors/aws/aws.go
@@ -9,9 +9,12 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/cache"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/cache/memory"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
@@ -20,6 +23,8 @@ import (
 type scanner struct {
 	verificationClient *http.Client
 	skipIDs            map[string]struct{}
+
+	credsCache cache.Cache
 }
 
 // resourceTypes derived from: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids
@@ -46,6 +51,11 @@ func New(opts ...func(*scanner)) *scanner {
 
 		opt(scanner)
 	}
+
+	scanner.credsCache = memory.New(
+		memory.WithExpirationInterval(1*time.Hour),
+		memory.WithPurgeInterval(2*time.Hour),
+	)
 
 	return scanner
 }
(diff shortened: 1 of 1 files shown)


pkg/detectors/aws/aws.go
```diff
@@ -148,11 +184,15 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 				}
+				cacheItem.extra = s1.ExtraData
 				if verificationErr != nil {
 					s1.SetVerificationError(verificationErr, resSecretMatch)
+					cacheItem.verificationErr = verificationErr
 				}
 			}
 
+			// Cache the result.
+			s.credsCache.Set(rawV2, cacheItem)
```
rgmz, line 195: Should this be a pointer?

https://github.com/patrickmn/go-cache#usage (last example)

ahrav: Indeed, using a pointer is the right approach. I'm considering implementing a constructor to ensure that we consistently return a pointer. While this approach doesn't offer substantial protection since it's within the same package, it sets the groundwork for potential future refactoring. This could involve moving it into a separate package, from which it can be imported by all other detectors

pkg/detectors/aws/aws.go
```diff
@@ -47,6 +51,11 @@ func New(opts ...func(*scanner)) *scanner {
 		opt(scanner)
 	}
 
+	scanner.credsCache = memory.New(
+		memory.WithExpirationInterval(1*time.Hour),
```
ahrav, line 55: These are just placeholders for now, we could consider passing these in as args to `New`.

ahrav: Or use some default value.

pkg/detectors/aws/aws.go
```diff
@@ -126,16 +155,28 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				RawV2:        []byte(rawV2),
 				ExtraData: map[string]string{
 					"resource_type": resourceTypes[idMatch[2]],
 				},
 			}
 
+			if val, ok := s.credsCache.Get(rawV2); ok {
+				item, ok := val.(cacheItem)
+				if !ok {
+					continue
```
ahrav, line 173: Not sure if this is the right behavior, i'd like to maybe log this as an error? Not sure if constructing our custom `logcontext.AddLogger` make sense so we can do that?

pkg/detectors/aws/aws.go
codevbus: How big could the cache conceivably get in a given time frame?

ahrav: That's a good question. It would really depend on what was being scanned. We can run some tests to scan some repos and record some metrics to see what the size of this cache grows to be. Alternatively, we could use a LRU cache where we fix the size if needed.

rgmz:
Still relevant?

PR #2298, opened 2024-01-12
Add flag to write job reports to disk

mcastorina:
### Description:
Adds a hook to the `SourceManager` to track unit progress and write the finished metrics to a file.

**Example output of source that supports units:**

```json
{
  "data": {
    "unit": {
      "source_unit_id": "/tmp/secretsandstuff"
    },
    "parent": {
      "job_id": 1,
      "source_id": 1,
      "source_name": "trufflehog - filesystem"
    },
    "start_time": "2024-01-11T16:16:27.684967-08:00",
    "end_time": "2024-01-11T16:16:27.687357-08:00",
    "total_chunks": 3,
    "total_bytes": 365,
    "errors": []
  },
  "version": 1
}
{
  "data": {
    "unit": {
      "source_unit_id": "/tmp/not-found"
    },
    "parent": {
      "job_id": 1,
      "source_id": 1,
      "source_name": "trufflehog - filesystem"
    },
    "start_time": "2024-01-11T16:16:27.684982-08:00",
    "end_time": "2024-01-11T16:16:27.685008-08:00",
    "total_chunks": 0,
    "total_bytes": 0,
    "errors": [
      "error chunking unit \"/tmp/not-found\": unable to get file info: stat /tmp/not-found: no such file or directory"
    ]
  },
  "version": 1
}
```

**Example output of source that does not support units:**

```json
{
  "data": {
    "parent": {
      "job_id": 1,
      "source_id": 1,
      "source_name": "trufflehog - github"
    },
    "start_time": "2024-01-11T16:17:09.050788-08:00",
    "end_time": "2024-01-11T16:17:26.966082-08:00",
    "total_chunks": 63609,
    "total_bytes": 386489349,
    "errors": []
  },
  "version": 1
}
```

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,6 +64,7 @@ var (
 	archiveTimeout       = cli.Flag("archive-timeout", "Maximum time to spend extracting an archive.").Duration()
 	includeDetectors     = cli.Flag("include-detectors", "Comma separated list of detector types to include. Protobuf name or IDs may be used, as well as ranges.").Default("all").String()
 	excludeDetectors     = cli.Flag("exclude-detectors", "Comma separated list of detector types to exclude. Protobuf name or IDs may be used, as well as ranges. IDs defined here take precedence over the include list.").String()
+	jobReportFile        = cli.Flag("output-report", "Write a scan report to the provided path.").Hidden().OpenFile(os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
 
 	gitScan             = cli.Command("git", "Find credentials in git repositories.")
 	gitScanURI          = gitScan.Arg("uri", "Git repository URL. https://, file://, or ssh:// schema expected.").Required().String()
@@ -413,6 +414,7 @@ func run(state overseer.State) {
 		engine.WithPrinter(printer),
 		engine.WithFilterEntropy(*filterEntropy),
 		engine.WithVerificationOverlap(*allowVerificationOverlap),
+		engine.WithJobReportWriter(*jobReportFile),
 	)
 	if err != nil {
 		logFatal(err, "error initializing engine")
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
@@ -399,6 +400,10 @@ func run(state overseer.State) {
 		fmt.Fprintf(os.Stderr, "🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷\n\n")
 	}
 
+	var jobReportWriter io.WriteCloser
+	if *jobReportFile != nil {
+		jobReportWriter = *jobReportFile
+	}
 	e, err := engine.Start(ctx,
 		engine.WithConcurrency(uint8(*concurrency)),
 		engine.WithDecoders(decoders.DefaultDecoders()...),
@@ -414,7 +419,7 @@ func run(state overseer.State) {
 		engine.WithPrinter(printer),
 		engine.WithFilterEntropy(*filterEntropy),
 		engine.WithVerificationOverlap(*allowVerificationOverlap),
-		engine.WithJobReportWriter(*jobReportFile),
+		engine.WithJobReportWriter(jobReportWriter),
 	)
 	if err != nil {
 		logFatal(err, "error initializing engine")
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -2,8 +2,10 @@ package engine
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"runtime"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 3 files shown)


pkg/sources/job_progress_hook.go
```diff
@@ -25,6 +26,10 @@ func WithUnitHookCache(cache *lru.Cache[string, *UnitMetrics]) UnitHookOpt {
 	return func(hook *UnitHook) { hook.metrics = cache }
 }
 
+func OnUnitHookFinish(f func(UnitMetrics)) UnitHookOpt {
+	return func(hook *UnitHook) { hook.finishFunc = f }
+}
```
mcastorina, line 31: I'm not fully convinced this is a good solution and am open to alternatives.

Another idea is to spawn a goroutine to periodically flush the hook and look for metrics that are finished to write out.

main.go
```diff
@@ -62,6 +62,7 @@ var (
 	archiveTimeout       = cli.Flag("archive-timeout", "Maximum time to spend extracting an archive.").Duration()
 	includeDetectors     = cli.Flag("include-detectors", "Comma separated list of detector types to include. Protobuf name or IDs may be used, as well as ranges.").Default("all").String()
 	excludeDetectors     = cli.Flag("exclude-detectors", "Comma separated list of detector types to exclude. Protobuf name or IDs may be used, as well as ranges. IDs defined here take precedence over the include list.").String()
+	jobReportFile        = cli.Flag("output-report", "Write a scan report to the provided path.").Hidden().OpenFile(os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
```
mcastorina, line 65: This is `Hidden` for now so we can change it without breaking users.

pkg/engine/engine.go
```diff
@@ -320,6 +330,40 @@ func (e *Engine) initialize(ctx context.Context, options ...Option) error {
...
+			}
+			mu.Lock()
+			defer mu.Unlock()
+			if _, err := e.jobReportWriter.Write(details); err != nil {
+				ctx.Logger().Error(err, "error writing to file")
+				return
+			}
+			if _, err := e.jobReportWriter.Write([]byte{'\n'}); err != nil {
+				ctx.Logger().Error(err, "error writing to file")
+			}
```
rosecodym, line 360: You could avoid this by creating `JobReportWriter.WriteLn` but I'm not sure how worth it that is

pkg/sources/job_progress_hook.go
rosecodym: What don't you like about it?

pkg/sources/job_progress_hook.go
```diff
@@ -143,6 +148,9 @@ func (u *UnitHook) Finish(ref JobProgressRef) {
 			metric.EndTime = snap.EndTime
 			metric.Errors = snap.Errors
 		}
+		if u.finishFunc != nil {
+			go u.finishFunc(*metric)
```
rosecodym, line 152: Does this _need_ to be done in a goroutine, or is it just for performance? And if it's just for performance, what's your sense of the magnitude of the expected benefit?

pkg/sources/job_progress_hook.go
mcastorina: It's a specific option for this specific hook. If there are other hooks that similarly need to be flushed, the solution doesn't scale. I think that's acceptable for now since there's just this one hook, but a system that knows about a `Flush` method and calls it might be nice.

pkg/engine/engine.go
mcastorina: Is `WriteLn` a standard interface or would that be custom? The nice thing about `io.Writer` is that a lot of structs already implement it (like `os.File`).

pkg/sources/job_progress_hook.go
mcastorina: It's for performance, and my hunch says it could be significant since it's doing IO, but it depends on how frequently this gets called (i.e. how many units are getting scanned).

I'll look into gathering some data.

pkg/engine/engine.go
rosecodym: oh, i missed that `jobReportWriter` is an interface. never mind. it kinda sucks that you have to check the error twice but i guess this is the best option.

mcastorina: Oh maybe I could append a newline to `details` before writing it.

pkg/sources/job_progress_hook.go
mcastorina: > additional implementation complexity

Are you referring to the locks needed for writing the file? I'm not sure I understand. I was thinking goroutines are light enough that they can be used for async operations like this.

rosecodym (dismissed):
i have some questions/suggestions but nothing blocking

pkg/engine/engine.go
```diff
@@ -373,6 +383,47 @@ func (e *Engine) initialize(ctx context.Context, options ...Option) error {
...
+		unitHook, finishedMetrics := sources.NewUnitHook(ctx)
+		opts = append(opts, sources.WithReportHook(unitHook))
+		e.wgDetectorWorkers.Add(1)
+		go func() {
+			defer e.wgDetectorWorkers.Done()
+			defer func() {
+				e.jobReportWriter.Close()
+				// Add a bit of extra information if it's a *os.File.
+				if namer, ok := e.jobReportWriter.(interface{ Name() string }); ok {
+					ctx.Logger().Info("report written", "path", namer.Name())
```
rosecodym, line 403: i love this!

rosecodym (approved):

PR #2299, opened 2024-01-12
[feat] - Provide CLI flag to only use custom verifiers

ahrav:
### Description:
This PR introduces a new `custom-verifiers-only` CLI flag that allows the scanner to only verify against the provided detector endpoints.
### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -366,7 +366,7 @@ func run(state overseer.State) {
 				"detector", id,
 			)
 		}
-		if !*customVerifiersOnly {
+		if !*customVerifiersOnly || len(urls) == 0 {
 			urls = append(urls, customizer.DefaultEndpoint())
 		}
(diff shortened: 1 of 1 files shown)


zricethezav (approved):

main.go
```diff
@@ -365,8 +366,9 @@ func run(state overseer.State) {
 				"detector", id,
 			)
 		}
-		// TODO: Add flag to ignore the default endpoint.
-		urls = append(urls, customizer.DefaultEndpoint())
+		if !*customVerifiersOnly {
+			urls = append(urls, customizer.DefaultEndpoint())
+		}
```
mcastorina, line 371: Thoughts on appending the default if `len(urls) == 0`? Might be a better user experience if you want to overwrite one endpoint of only one of the possible detectors that can be configured.

PR #2301, opened 2024-01-12
Fix Postgres detector

rosecodym:
### Description:
* [x] Use all found URIs in a chunk, rather than just the first one
* [x] Test locally on a port other than 5433, because that one can have local dev problems
* [ ] Stop manufacturing candidate secrets with "localhost" as a candidate host
  * [ ] ignore if it it's in the chunk
  * [ ] don't conjure it up as a default
* [ ] Make it so that if "localhost" is found in a chunk, the "host" part of it doesn't trigger the host detection pattern

### Checklist:


rosecodym:
closed in favor of #2314

PR #2302, opened 2024-01-12
Draft: fix for iad format when DN is in uri, I think...

zricethezav:
### Description:

### Checklist:

### DIFF
--- a/pkg/detectors/ldap/ldap.go
+++ b/pkg/detectors/ldap/ldap.go
@@ -65,7 +65,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				}
 
 				if verify {
-					verificationErr := verifyLDAP(username[1], password[1], ldapURL)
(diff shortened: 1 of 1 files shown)


rgmz:
> This probably needs some more polish and LDAP is out of my wheelhouse but it looks like the Bind:

Bind usually requires the full DistinguishedName OR distinct username attribute.

Active Directory used to use `sAMAccountName` (e.g., `jdoe`) as the username, however, if you have multiple forests then you'd need to do `jdoe@domain`. Now `UserPrincipalName` is preferred. https://pleasantpasswords.com/info/pleasant-password-server/i-ldap-and-ad/active-directory-and-samaccountnames

Not sure about other LDAP implementations.

zricethezav:
> Bind usually requires the full DistinguishedName OR distinct username attribute.

@rgmz anyone to determine which one to use? This change attempts both if a username returns an err

PR #2304, opened 2024-01-14
added azure protos

roxanne-tampus:
### Description:
To add detectors for azure services

### Checklist:


ahrav (approved):

proto/detectors.proto
```diff
@@ -977,6 +977,17 @@ enum DetectorType {
   Ngrok = 966;
   Replicate = 967;
   Postgres = 968;
+  AzureActiveDirectoryApplicationSecret = 969;
+  AzureCacheForRedisAccessKey = 970;
+  AzureCosmosdbKeyIdentifiable = 971;
```
ahrav, line 982: nit: I think we should probably make the `db` capitalized so it's similar to other detectors with the same name. (ex `MongoDB`
`AzureCosmosdbKeyIdentifiable` - >  `AzureCosmosDBKeyIdentifiable`

proto/detectors.proto
```diff
@@ -977,6 +977,17 @@ enum DetectorType {
...
+  AzureDevopsPersonalAccessToken = 972;
+  AzureFunctionKey = 973;
+  AzureMlWebServiceClassicIdentifiableKey = 974;
+  AzureSasToken = 975;
+  AzureSearchAdminKey = 976;
+  AzureSearchQueryKey = 977;
+  AzureManagementCertificate = 978;
+  AzureSql = 979;
```
ahrav, line 990: nit: Similarly here, lets make `Sql` -> `SQL` so it matches our other detector's casing. (ex: `SQLServer`)

proto/detectors.proto
```diff
@@ -977,6 +977,17 @@ enum DetectorType {
   Ngrok = 966;
   Replicate = 967;
   Postgres = 968;
+  AzureActiveDirectoryApplicationSecret = 969;
+  AzureCacheForRedisAccessKey = 970;
+  AzureCosmosdbKeyIdentifiable = 971;
+  AzureDevopsPersonalAccessToken = 972;
+  AzureFunctionKey = 973;
+  AzureMlWebServiceClassicIdentifiableKey = 974;
```
ahrav, line 985: nit: Lets also do the same for `Ml` -> `ML`.
`AzureMlWebServiceClassicIdentifiableKey` -> `AzureMLWebServiceClassicIdentifiableKey`

ahrav (commented):
I know we aren't really following the [style guide](https://protobuf.dev/programming-guides/style/#enums) for naming protobufs, so I think making sure the convention is consistent across the codebase is likely more important at this point. I think common acronyms like DB, SQL, ML, CI, etc can likely be upper-cased.

roxanne-tampus:
Got it.  Thanks! I just pushed the changes.

PR #2306, opened 2024-01-16
[feat] - tmp file diffs

ahrav:
### Description:
This PR aims to reduce memory usage when scanning large repositories containing diffs that exceed 10MB. Diffs now write to temporary disk files instead of allocating in-memory buffers.

The 10MB threshold is configurable and could be adjusted based on gathering metrics on average diff size. The bytes.Buffer could also be pre-allocated to the average diff length to optimize memory allocation.

By streaming large diffs to disk rather than buffering fully in memory, we can alleviate memory pressure and OutOfMemory errors that occur when scanning extremely large repositories.

Screenshot 2024-01-14 at 7 25 52 PM

### Checklist:


pkg/gitparse/gitparse.go
```diff
@@ -53,6 +131,51 @@ type Parser struct {
...
+		// Data is in a file, read from the file.
+		file, err := os.Open(d.streamDestination.Name())
+		if err != nil {
+			return nil, err
+		}
+		return newAutoDeletingFileReader(file), nil
+	}
+	// Data is in memory.
+	return &noOpCloser{Reader: bytes.NewReader(d.Content.Bytes())}, nil
+}
```
mcastorina, line 158: Can this function only be called after we finish parsing the diff (i.e. after `finalize()` is called)?

pkg/gitparse/gitparse.go
```diff
@@ -40,10 +42,86 @@ type Commit struct {
...
 // Diff contains the info about a file diff in a commit.
 type Diff struct {
-	PathB     string
-	LineStart int
-	Content   bytes.Buffer
-	IsBinary  bool
+	PathB             string
+	LineStart         int
+	Content           bytes.Buffer // Keep in-memory buffer for smaller diffs
+	streamDestination *os.File     // File destination for larger diffs
```
mcastorina, line 48: Should `Content` be unexported?

pkg/gitparse/gitparse.go
```diff
@@ -53,6 +131,51 @@ type Parser struct {
...
+
+// newAutoDeletingFileReader creates a new autoDeletingFileReader
+func newAutoDeletingFileReader(file *os.File) *autoDeletingFileReader {
+	return &autoDeletingFileReader{file: file}
+}
+
+// Read implements the io.Reader interface
+func (r *autoDeletingFileReader) Read(p []byte) (int, error) {
+	return r.file.Read(p)
+}
```
mcastorina, line 171: **Nit (optional):** If we embed `*os.File` in `autoDeletingFileReader` we would get this method for free. Up to you on whether that's too magical or you prefer the explicit pass-through.

pkg/gitparse/gitparse.go
ahrav: I tried, but apparently it's used elsewhere in the code :(. Tried not to touch too many more things with this change, but I think it's something we can look to change.

pkg/gitparse/gitparse.go
ahrav: I did contemplate this and decided against it. Happy to embed it if we find that approach a little cleaner. Always happy for less code.

pkg/gitparse/gitparse.go
ahrav: yea that sounds accurate. 👍

pkg/gitparse/gitparse.go
mcastorina: `Content` doesn't always have the diff contents though, so if it's being used elsewhere that sounds like a bug waiting to happen.

ahrav: My thinking is, if it's currently an issue, this won't make it worse. 😅

pkg/gitparse/gitparse.go
```diff
@@ -42,10 +44,45 @@ type Commit struct {
...
+	for _, opt := range opts {
+		opt(d)
+	}
+	d.contentWriter = bufferedfilewriter.New(bufferedfilewriter.WithThreshold(defaultThreshold))
+
+	return d
+}
+
+// Len returns the length of the storage.
+func (d *Diff) Len() int { return d.contentWriter.Len() }
```
ahrav, line 70: These methods probably can be removed, but I was trying to avoid reaching into the contentWriter directly. This is also exported since we need access to this in the git source. Open to changing it if we feel these are providing any additional encapsulation.

pkg/gitparse/gitparse.go
```diff
@@ -40,10 +42,86 @@ type Commit struct {
...
+	d := &Diff{threshold: defaultThreshold}
+	for _, opt := range opts {
+		opt(d)
+	}
+
+	return d
+}
+
+// write handles writing diff data to either an in-memory buffer or a file, depending on the size.
+func (d *Diff) write(ctx context.Context, p []byte) error {
```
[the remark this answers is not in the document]
ahrav: Moved it to its own pkg.

pkg/sources/git/git.go
```diff
@@ -1048,9 +1089,6 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 		return err
 	}
 	defer func() {
-		if err := fileReader.Close(); err != nil {
```
ahrav, line 1051: removed, since StdoutPipe clsoes the pipe after seeing the command exit:
```
// StdoutPipe returns a pipe that will be connected to the command's
// standard output when the command starts.
//
// Wait will close the pipe after seeing the command exit....
```

mcastorina (commented):
I have a general question about the bufferedfilewriter package. What is the intended API for using it?

For example, is it meant to be written to all at once and then read from all at once? Can the contents be peeked at before reading? I think answering how it will be used (or needs to be used) will help shape the interface.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -0,0 +1,189 @@
...
+		opt(w)
+	}
+	return w
+}
+
+// Len returns the number of bytes in the buffer.
+func (w *BufferedFileWriter) Len() int { return w.buf.Len() }
+
+// String returns the contents of the buffer as a string.
+func (w *BufferedFileWriter) String() string { return w.buf.String() }
```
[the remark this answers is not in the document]
ahrav: Oh yea good catch. This should be the size, since that's what we will use to determine if we want to chunk the diff in the `git` source.

ahrav:
> I have a general question about the bufferedfilewriter package. What is the intended API for using it?
> 
> For example, is it meant to be written to all at once and then read from all at once? Can the contents be peeked at before reading? I think answering how it will be used (or needs to be used) will help shape the interface.

Main use case is for our git handling of diff/commits. So multiple writes, but only a single read for now. No peek, no extra fancy stuff for now.

pkg/gitparse/gitparse.go
```diff
@@ -326,12 +363,15 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 			if currentCommit == nil {
 				currentCommit = &Commit{}
 			}
-			if currentDiff.Content.Len() > 0 || currentDiff.IsBinary {
-				currentCommit.Diffs = append(currentCommit.Diffs, currentDiff)
+			if currentDiff.Len() > 0 || currentDiff.IsBinary {
+				currentCommit.Diffs = append(currentCommit.Diffs, *currentDiff)
+				if err := currentDiff.finalize(); err != nil {
+					ctx.Logger().Error(err, "failed to finalize diff")
```
rosecodym, line 369: are there any extra kvps we can add here to help us out if we ever run into this?

pkg/gitparse/gitparse.go
```diff
@@ -152,14 +189,13 @@ func (c1 *Commit) Equal(c2 *Commit) bool {
 			return false
 		case d1.LineStart != d2.LineStart:
 			return false
-		case d1.Content.String() != d2.Content.String():
+		case d1.contentWriter.String() != d2.contentWriter.String():
```
rosecodym, line 192: doesn't `bufferedFileWriter.String()` only stringify the in-memory buffer? i might have missed something.

rosecodym (commented):
I have one concrete question about stringification, but taking a step back, I'm pretty concerned about big-banging this into the codebase. It's a pretty sophisticated I/O structure, used in many of our most popular sources, that interacts with the temp file system - something we've historically (and currently!) had nasty trouble with. It's also probably got a pretty complicated performance profile in which a bunch of variables we don't entirely control (disk latency, diff size, flush-to-disk threshold) combine to create what's most likely a non-monotonic performance curve. (I'm guessing that it's good for small diffs and good for large diffs but pretty bad for diffs right around the cutoff.)

Is there _any_ way we can roll this out gradually? I know that it would make the transitional code itself kind of gnarly but this thing is big, and deep, and complicated, and touching risky things. This might be a good topic for a synchronous chat, if you want.

pkg/gitparse/gitparse.go
ahrav: err whoops, it should be the whole diff contents. Will add a test and update.

pkg/gitparse/gitparse.go
```diff
@@ -22,12 +23,96 @@ const (
...
+func (b *buffer) String(_ context.Context) string { return b.Buffer.String() }
+
+// Diff contains the information about a file diff in a commit.
+// It abstracts the underlying content representation, allowing for flexible handling of diff content.
+// The use of contentWriter enables the management of diff data either in memory or on disk,
+// based on its size, optimizing resource usage and performance.
+type Diff struct {
+	PathB         string
+	LineStart     int
+	contentWriter contentWriter
```
ahrav, line 79: This is the only line that changed in this struct + the comment. Moved it so it's grouped closer to its constructor and methods.

pkg/gitparse/gitparse.go
```diff
@@ -38,19 +123,44 @@ type Commit struct {
...
+
+	for i := range c1.Diffs {
+		d1 := c1.Diffs[i]
+		d2 := c2.Diffs[i]
+		switch {
+		case d1.PathB != d2.PathB:
+			return false
+		case d1.LineStart != d2.LineStart:
+			return false
+		case d1.contentWriter.String(ctx) != d2.contentWriter.String(ctx):
```
ahrav, line 149: Similarly, this is the only line that changed.

ahrav:
failing tests look to be due to account lockout with Gitlab.

pkg/gitparse/gitparse.go
```diff
@@ -13,19 +12,22 @@ import (
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/writers/buffered_file_writer"
 )
 
 const (
 	// defaultDateFormat is the standard date format for git.
 	defaultDateFormat = "Mon Jan 02 15:04:05 2006 -0700"
 
 	// defaultMaxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
-	defaultMaxDiffSize = 1 * 1024 * 1024 * 1024 // 1GB
+	defaultMaxDiffSize = 2 * 1024 * 1024 * 1024 // 1GB
```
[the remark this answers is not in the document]
mcastorina: **Nit:** The comment should be updated too

pkg/gitparse/gitparse.go
```diff
@@ -22,12 +23,96 @@ const (
...
 
 	// defaultMaxCommitSize is the maximum size for a commit. Larger commits will be cut off.
-	defaultMaxCommitSize = 1 * 1024 * 1024 * 1024 // 1GB
+	defaultMaxCommitSize = 2 * 1024 * 1024 * 1024 // 1GB
```
mcastorina, line 29: Same here

pkg/gitparse/gitparse.go
```diff
@@ -22,12 +23,96 @@ const (
...
+	// Write appends data to the content storage.
+	Write(ctx context.Context, data []byte) (int, error)
+	// ReadCloser provides a reader for accessing stored content.
+	ReadCloser() (io.ReadCloser, error)
+	// Close finalizes the content storage, performing cleanup if necessary.
+	Close() error
+	// Len returns the current size of the content.
+	Len() int
+	// String returns the content as a string.
+	String(ctx context.Context) string
```
mcastorina, line 46: What do you think of changing this to also return an error?

```go
String(ctx context.Context) (string, error)
```

If the backend storage can change, I think it's reasonable to assume that some backends cannot return the data in a single string.

pkg/gitparse/gitparse.go
ahrav: thank you!

rosecodym (commented):
sorry, surprise plumber arrival, but i wanted to get my one comment in!

pkg/gitparse/gitparse.go
```diff
@@ -22,12 +23,96 @@ const (
...
 
+// contentWriter defines a common interface for writing, reading, and managing diff content.
+// It abstracts the underlying storage mechanism, allowing flexibility in how content is handled.
+// This interface enables the use of different content storage strategies (e.g., in-memory buffer, file-based storage)
+// based on performance needs or resource constraints, providing a unified way to interact with different content types.
+type contentWriter interface { // Write appends data to the content storage.
+	// Write appends data to the content storage.
+	Write(ctx context.Context, data []byte) (int, error)
+	// ReadCloser provides a reader for accessing stored content.
+	ReadCloser() (io.ReadCloser, error)
```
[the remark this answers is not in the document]
ahrav: I believe it's only an `io.Reader`. We need the `Closer` specifically for the `filebufferedwriter` implementation in which we need to close the file. The `Close` on the buffer is a no-op.

pkg/sources/git/git.go
```diff
@@ -539,30 +539,54 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
-				Data:           diff.Content.Bytes(),
-				Verify:         s.verify,
+
+			chunkData := func(d *gitparse.Diff) error {
+				metadata := s.sourceMetadataFunc(fileName, email, hash, when, remoteURL, int64(diff.LineStart))
+
+				reader, err := diff.ReadCloser()
+				if err != nil {
+					ctx.Logger().Error(err, "error creating reader for commits", "filename", fileName, "commit", hash, "file", diff.PathB)
+					return nil
```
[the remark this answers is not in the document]
ahrav: I think the idea is to prevent stopping a scan for an error for a single diff. Do we think this should warrant stopping the scan?

pkg/gitparse/gitparse.go
```diff
@@ -38,19 +123,60 @@ type Commit struct {
...
+		switch {
+		case d1.PathB != d2.PathB:
+			return false
+		case d1.LineStart != d2.LineStart:
+			return false
+		case d1.IsBinary != d2.IsBinary:
+			return false
+		default:
+			equal, err := isEqualContentString(d1.contentWriter, d2.contentWriter)
+			if err != nil || !equal {
```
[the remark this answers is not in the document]
ahrav: I can pass in a ctx and log it.

pkg/gitparse/gitparse.go
rosecodym: Do you want to guard them behind the state like `ReadCloser()`? I find it a bit surprising that some of the read methods are guarded and some aren't.

ahrav: I like that idea. Will guard them.

rosecodym (approved):

PR #2308, opened 2024-01-16
Assume unauthenticated github scans have public visibility

mcastorina:
### Description:
Checks whether the github scan is unauthenticated and assumes the visibility of the repo is public. If this is a bad assumption, we could switch to a non-rate-limited way of checking the visibility.

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -357,6 +357,10 @@ func (s *Source) visibilityOf(ctx context.Context, repoURL string) (visibility s
 		s.mu.Unlock()
 	}()
 	logger := s.log.WithValues("repo", repoURL)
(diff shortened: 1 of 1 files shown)


rgmz:
This is sensible but I think it makes most sense to record the repository's visibility during initial enumeration. There's no need to make an API call for visibility at all, regardless of auth.

dustin-decker (approved):

mcastorina:
> This is sensible but I think it makes most sense to record the repository's visibility [during initial enumeration](https://github.com/trufflesecurity/trufflehog/blob/a1dc660f41350e4696b1fc5f691c7ef48e858a54/pkg/sources/github/repo.go#L221). There's no need to make an API call for visibility at all, regardless of auth.

That's a great point! I'm going to merge this for now since recording that during enumeration will require a larger refactor.

PR #2309, opened 2024-01-18
Refactor UnitHook to block the scan if finished metrics aren't handled

mcastorina:
### Description:
`UnitHook` was originally designed to drop metrics when the producer out-paced the consumer. This changes it so the hook will block the scan until there is room to write the finished metrics to the output channel.

### Checklist:

### DIFF
--- a/pkg/sources/job_progress_hook.go
+++ b/pkg/sources/job_progress_hook.go
@@ -16,6 +16,7 @@ type UnitHook struct {
 	metrics         *lru.Cache[string, *UnitMetrics]
 	mu              sync.Mutex
 	finishedMetrics chan UnitMetrics
+	logBackPressure func()
 	NoopHook
 }
 
@@ -43,9 +44,15 @@ func NewUnitHook(ctx context.Context, opts ...UnitHookOpt) (*UnitHook, <-chan Un
 			"metric", value,
 		)
 	})
+	var once sync.Once
 	hook := UnitHook{
 		metrics:         cache,
 		finishedMetrics: make(chan UnitMetrics, 1024),
+		logBackPressure: func() {
+			once.Do(func() {
+				ctx.Logger().Info("back pressure detected in unit hook")
+			})
+		},
 	}
 	for _, opt := range opts {
 		opt(&hook)
@@ -60,6 +67,18 @@ func (u *UnitHook) id(ref JobProgressRef, unit SourceUnit) string {
 		unitID = unit.SourceUnitID()
 	}
 	return fmt.Sprintf("%d/%d/%s", ref.SourceID, ref.JobID, unitID)
+}
+
+func (u *UnitHook) ejectFinishedMetric(metric UnitMetrics) {
+	// Intentionally block the hook from returning to supply back-pressure
+	// to the source.
+	select {
+	case u.finishedMetrics <- metric:
+		return
+	default:
+		u.logBackPressure()
+	}
+	u.finishedMetrics <- metric
 }
 
 func (u *UnitHook) StartUnitChunking(ref JobProgressRef, unit SourceUnit, start time.Time) {
@@ -82,9 +101,7 @@ func (u *UnitHook) EndUnitChunking(ref JobProgressRef, unit SourceUnit, end time
 		return
 	}
 	metrics.EndTime = &end
-	// Intentionally block the hook from returning to supply back-pressure
-	// to the source.
-	u.finishedMetrics <- *metrics
+	u.ejectFinishedMetric(*metrics)
 }
 
 func (u *UnitHook) finishUnit(id string) (*UnitMetrics, bool) {
@@ -157,9 +174,7 @@ func (u *UnitHook) Finish(ref JobProgressRef) {
 	metrics.StartTime = snap.StartTime
 	metrics.EndTime = snap.EndTime
 	metrics.Errors = snap.Errors
-	// Intentionally block the hook from returning to supply back-pressure
-	// to the source.
-	u.finishedMetrics <- *metrics
+	u.ejectFinishedMetric(*metrics)
 }
 
 // InProgressSnapshot gets all the currently active metrics across all jobs.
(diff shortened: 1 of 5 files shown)


pkg/sources/job_progress_hook.go
```diff
@@ -122,51 +146,40 @@ func (u *UnitHook) ReportError(ref JobProgressRef, err error) {
...
 	output := make([]UnitMetrics, 0, u.metrics.Len())
 	for _, id := range u.metrics.Keys() {
-		metric, ok := u.metrics.Get(id)
-		if !ok {
-			continue
-		}
-		output = append(output, *metric)
-		if metric.IsFinished() {
-			metric.handled = true
-			u.metrics.Remove(id)
```
rosecodym, line 164: Was this being done here before because there wasn't really a better place for it? But now, there is, so you moved it? Or am I misunderstanding?

pkg/sources/source_manager_test.go
```diff
@@ -316,7 +316,7 @@ func TestSourceManagerAvailableCapacity(t *testing.T) {
 }
 
 func TestSourceManagerUnitHook(t *testing.T) {
-	hook := NewUnitHook(context.TODO())
+	hook, ch := NewUnitHook(context.TODO())
```
rosecodym, line 319: what's the difference between `context.TODO()` and `context.Background()`?

pkg/sources/source_manager_test.go
```diff
@@ -366,14 +370,10 @@ func TestSourceManagerUnitHook(t *testing.T) {
...
-// TestSourceManagerUnitHookNoBlock tests that the UnitHook drops metrics if
-// they aren't handled fast enough.
-func TestSourceManagerUnitHookNoBlock(t *testing.T) {
-	var evictedKeys []string
-	cache, _ := lru.NewWithEvict(1, func(key string, _ *UnitMetrics) {
-		evictedKeys = append(evictedKeys, key)
-	})
-	hook := NewUnitHook(context.TODO(), WithUnitHookCache(cache))
+// TestSourceManagerUnitHookBackPressure tests that the UnitHook blocks if the
+// finished metrics aren't handled fast enough.
```
rosecodym, line 374: ❤️

rosecodym (commented):
This code looks good (to my relatively inexpert eye), but do we have a monitoring plan yet? I'd like to avoid getting too out over our skies on the code itself.

pkg/sources/job_progress_hook.go
mcastorina: This was the previous mechanism to drain the cache, which is now in `EndUnitChunking` and `Finish`. The idea was that you would periodically call `UnitMetrics()` and get the list of in progress and finished metrics, and the finished ones would be considered "handled" and removed from the cache.

pkg/sources/source_manager_test.go
mcastorina: Functionally they're equivalent, but the docs say this:

> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

ahrav (approved):
This looks overall good to me. Do we have any way to know if we every actually run into the situation where the back pressure actually gets applied? It would be surprising, but nonetheless might be useful.

ahrav:
> > Do we have any way to know if we every actually run into the situation where the back pressure actually gets applied?
> 
> I could add some code where we write to the channel that would look like:
> 
> ```go
> select {
>     case u.finishedMetrics <- metric:
>         // no backpressure
>     default:
>         // backpressure; do something / log?
>         u.finishedMetrics <- metric
> }
> ```
> 
> What do you think? Or maybe add another prometheus metric to track the size of the channel?

Yea I think even just a log honestly would be great. I'm pretty curious to see if we ever execute that code path. I'd actually maybe want to get a profile if that did happen.

mcastorina:
@rosecodym @ahrav I added both a log and a metric, and we can decide to keep one or the other later. Let me know what you think of the implementation.

pkg/sources/job_progress_hook.go
```diff
@@ -122,51 +175,38 @@ func (u *UnitHook) ReportError(ref JobProgressRef, err error) {
...
+		if metrics, ok := u.metrics.Get(id); ok {
+			output = append(output, *metrics)
 		}
 	}
 	return output
 }
 
+func (u *UnitHook) Close() error {
```
rosecodym, line 205: Is this used anywhere?

pkg/sources/job_progress_hook.go
```diff
@@ -52,6 +81,18 @@ func (u *UnitHook) id(ref JobProgressRef, unit SourceUnit) string {
 	return fmt.Sprintf("%d/%d/%s", ref.SourceID, ref.JobID, unitID)
 }
 
+func (u *UnitHook) ejectFinishedMetric(metric UnitMetrics) {
```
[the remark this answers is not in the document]
mcastorina: Indecisiveness lol. I tend to use plural for slices of objects, but for some reason I named the type plural `UnitMetrics`. 🤷 

I'll make it plural for consistency, but open to suggestions in general for naming things.

pkg/sources/job_progress_hook.go
mcastorina: Yes, it gets called when the `SourceManager` is waited:

https://github.com/trufflesecurity/trufflehog/blob/fb230aa0ae379066cd6dd348d723a0eb7ba0fb47/pkg/sources/source_manager.go#L173-L177

mcastorina: Some synchronization is required to make the goroutine reading the channel finish before everything gets shut down, but that doesn't happen in this PR since `UnitHook` isn't used yet.

I figured it made sense for the source manager to close out the hooks since the hook initialized it.

rosecodym: > Some synchronization is required to make the goroutine reading the channel finish before everything gets shut down, but that doesn't happen in this PR since UnitHook isn't used yet.

You read my mind :)

rosecodym (approved):

rosecodym (approved):

PR #2310, opened 2024-01-18
[fixup] - save 8 bytes per chunk

ahrav:
### Description:
Reduce chunk byte size from 88 -> 80.

### Checklist:

### DIFF
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -18,6 +18,9 @@ type (
 
 // Chunk contains data to be decoded and scanned along with context on where it came from.
 type Chunk struct {
+	// Data is the data to decode and scan.
+	Data []byte
+
 	// SourceName is the name of the Source that produced the chunk.
 	SourceName string
 	// SourceID is the ID of the source that the Chunk originated from.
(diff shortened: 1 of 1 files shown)


dustin-decker:
Can you share your reference on struct ordering? I'm surprised that the byte slice would be first in the struct

ahrav:
> Can you share your reference on struct ordering? I'm surprised that the byte slice would be first in the struct

Sure. [here is an example](https://go.dev/play/p/Azf4a7O-DhC) and a decent[blog](https://go101.org/article/memory-layout.html). TLDR go slice is a struct with 3 words, and when it comes to alignment we take the total of the types fields, thus it is the largest of all the fields in our struct:
```
type sliceHeader struct {
    Length        int
    Capacity      int
    ZerothElement *byte
}
```

ahrav:
Also for some more context, our metrics indicate that instances of the Chunk struct are extremely prevalent - on the order of billions created for certain use cases. Even though the size optimization from 88 to 80 bytes seems minor for an individual Chunk, this 8 byte reduction per instance has the potential to significantly scale in the aggregate across billions of Chunks.

dustin-decker (approved):
Thanks for the example and the resource. LGTM!

rosecodym:
How on earth did you notice this??

Also, I know you already merged, but I think it would be wise to add to the `Chunk` doc comment that the struct layout matters and why.

ahrav:
> How on earth did you notice this??
> 
> Also, I know you already merged, but I think it would be wise to add to the `Chunk` doc comment that the struct layout matters and why.

Yea great call. I read a blog, our Chunk struct came to mind. Then realized there was a [tool](https://github.com/mdempsky/maligned), and then realized there is also a way to have the linter catch it. Chatted with Miccah to see if reordering the fields hindered readability, we concluded it didn't and thus this PR. 🥳 

[PR](https://github.com/trufflesecurity/trufflehog/pull/2317#issuecomment-1900641902) with updated comment.

PR #2312, opened 2024-01-18
Add prometheus metrics to measure hook execution time

mcastorina:
### Description:
Add prometheus metrics to measure hook execution time.

### Checklist:

### DIFF
--- a/pkg/sources/job_progress.go
+++ b/pkg/sources/job_progress.go
@@ -8,20 +8,6 @@ import (
 	"fmt"
 	"sync"
 	"time"
-
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
-)
-
-var (
-	hooksExecTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
-		Namespace: common.MetricsNamespace,
-		Subsystem: common.MetricsSubsystem,
-		Name:      "hooks_exec_time_ms",
-		Help:      "Time spent executing hooks (ms)",
-		Buckets:   []float64{5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000},
-	}, nil)
 )
 
 type JobProgressHook interface {
--- a/pkg/sources/job_progress.go
+++ b/pkg/sources/job_progress.go
@@ -8,6 +8,20 @@ import (
 	"fmt"
 	"sync"
 	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+)
+
+var (
+	hooksExecTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "hooks_exec_time_ms",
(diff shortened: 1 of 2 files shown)


ahrav (commented):
How do you feel about putting this in a `metrics.go` file similar to how we currently do it for other parts of the code. (ie [github](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/github/metrics.go), [gitlab](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/gitlab/metrics.go), etc)?

pkg/sources/job_progress.go
```diff
@@ -8,6 +8,20 @@ import (
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+)
+
+var (
+	hooksExecTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "hooks_exec_time_ms",
+		Help:      "Time spent executing hooks (ms)",
+		Buckets:   []float64{5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000},
```
ahrav, line 23: question: have you tested this in a dev/local env? 

Since this needs to be sent over the network even with compression there is a chance this exceeds our current gRPC payload limit. I only mention this because this one of the main reasons I had to change our metrics for detectors to only include 3-4 bucket values due to the enormous size of the payload.

ahrav: I added compression and also increased the payload limit on the API server, but i'm not sure if that would be enough for these metrics.

mcastorina: Ah, good callout. I wasn't aware that the number of buckets affected the size of the payload. These buckets were copied from another metric in thog (but on the API side, so no GRPC messaging). I'll reduce them.

mcastorina:
> How do you feel about putting this in a `metrics.go` file similar to how we currently do it for other parts of the code. (ie [github](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/github/metrics.go), [gitlab](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/gitlab/metrics.go), etc)?

I considered it, but felt like I was adding too many files to the `sources` package. Should `SourceManager` et al. be in it's own package?

pkg/sources/job_progress.go
ahrav: Funnily enough I did the exact same thing and copied the exact same metrics when I added them for our detectors 🤣 .

ahrav (approved):

ahrav:
> > How do you feel about putting this in a `metrics.go` file similar to how we currently do it for other parts of the code. (ie [github](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/github/metrics.go), [gitlab](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/gitlab/metrics.go), etc)?
> 
> I considered it, but felt like I was adding too many files to the `sources` package. Should `SourceManager` et al. be in it's own package?

I personally don't mind either approach tbh. Gonna leave that up to you, or anyone else that feels strongly. You can also do that in a separate PR if that's easier.

pkg/sources/job_progress.go
rosecodym: Ooh, I didn't know that either.

rosecodym (approved):
I weakly vote for moving it into its own metrics file - I've found it nice to be able to see how metrics work by looking at just one file - but I really don't think it matters much either way.

PR #2313, opened 2024-01-18
Walk directories in filesystem source enumeration

mcastorina:
### Description:
Previously the unit could be a directory, but now it is individual files. This leverages the SourceManager's concurrent units to speed up scanning of large directories.

```
» find /tmp/files -type f | wc -l
  100000

# before
» ./trufflehog filesystem /tmp/files
2024-01-18T12:48:05-08:00       info-0  trufflehog      finished scanning       {"chunks": 100000, "bytes": 588895, "verified_secrets": 0, "unverified_secrets": 0, "scan_duration": "27.942921208s"}
# 27.94s

# after
» ./trufflehog filesystem /tmp/files
2024-01-18T12:47:04-08:00       info-0  trufflehog      finished scanning       {"chunks": 100000, "bytes": 588895, "verified_secrets": 0, "unverified_secrets": 0, "scan_duration": "14.342492916s"}
# 14.34s
```

### Checklist:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -233,7 +233,10 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 			continue
 		}
 		err = fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
-			if err != nil || d.IsDir() {
+			if err != nil {
+				return reporter.UnitErr(ctx, err)
+			}
+			if d.IsDir() {
 				return nil
 			}
 			fullPath := filepath.Join(path, relativePath)
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -223,6 +223,26 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
 // filepath or a directory.
 func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) error {
 	for _, path := range s.paths {
+		fileInfo, err := os.Stat(filepath.Clean(path))
+		if err != nil {
+			if err := reporter.UnitErr(ctx, err); err != nil {
+				return err
+			}
+			continue
+		}
+		if fileInfo.IsDir() {
+			return fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
+				if err != nil || relativePath == "." {
+					return nil
+				}
+				fullPath := filepath.Join(path, relativePath)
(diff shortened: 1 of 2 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -223,6 +223,26 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
...
+			continue
+		}
+		if fileInfo.IsDir() {
+			return fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
+				if err != nil || relativePath == "." {
+					return nil
+				}
+				fullPath := filepath.Join(path, relativePath)
+				item := sources.CommonSourceUnit{ID: fullPath}
+				if err := reporter.UnitOk(ctx, item); err != nil {
```
ahrav, line 240: nit: Can we just `return reporter.UnitOK(ctx, item)` ?

ahrav (approved):

pkg/sources/filesystem/filesystem.go
```diff
@@ -223,9 +218,32 @@ func (s *Source) scanFile(ctx context.Context, path string, chunksChan chan *sou
...
+		}
+		if !fileInfo.IsDir() {
+			item := sources.CommonSourceUnit{ID: path}
+			if err := reporter.UnitOk(ctx, item); err != nil {
+				return err
+			}
+			continue
+		}
+		err = fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
+			if err != nil || d.IsDir() {
```
rosecodym, line 236: Does this risk throwing away any errors that we were previously capturing?

rosecodym (approved):
LGTM, other than my one question.

It is occurring to me that we might need to all come together sometime soonish to talk about our philosophy of concurrency - I think that different parts of the codebase are treating the topic differently and we might need to do some reconciliation.

pkg/sources/filesystem/filesystem.go
mcastorina: For reference, the [WalkDirFunc](https://pkg.go.dev/io/fs#WalkDirFunc) describes the `err` argument:

> The err argument reports an error related to path, signaling that WalkDir will not walk into that directory. The function can decide how to handle that error; as described earlier, returning the error will cause WalkDir to stop walking the entire tree.

Perhaps there's an argument that we should record the error in the job report, but as it stands we currently ignore any errors given by `WalkDir`: https://github.com/trufflesecurity/trufflehog/blob/8380e1713ea5d3fc21c2f1256ab703f2af30b0de/pkg/sources/filesystem/filesystem.go#L112-L116

dustin-decker:
Looks like this caused a regression
https://github.com/trufflesecurity/trufflehog/issues/2350

PR #2314, opened 2024-01-18
Narrow Postgres detector to only look for URIs

rosecodym:
### Description:
This PR re-enables the Postgres detector and modifies it to only look for connection string URIs. If we're happy that it's not generating a bunch of false hits we can later add separate-credential support.

### Checklist:


pkg/detectors/postgres/postgres.go
```diff
@@ -69,186 +139,87 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
...
-				}
-				postgresURL := url.URL{
-					Scheme: "postgresql",
-					User:   url.UserPassword(username, password),
-					Host:   fmt.Sprintf("%s:%s", hostname, port),
-				}
-				pgURLs = append(pgURLs, postgresURL)
-			}
+		params := make(map[string]string)
+		parts := connStrPartPattern.FindAllStringSubmatch(connStr, -1)
```
ahrav, line 151: nit: If we construct `params` after parts, we can pre-alloc the map. it'll make it go SOOOO much faster :)

`params := make(map[string]string, len(parts))`

pkg/detectors/postgres/postgres.go
```diff
@@ -69,186 +139,87 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
...
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
-	defer cancel()
-
-	err = db.PingContext(ctx)
+	err = db.Ping()
 	if err == nil {
 		return true, nil
-	} else if strings.Contains(err.Error(), "password authentication failed") || // incorrect username or password
-		strings.Contains(err.Error(), "connection refused") { // inactive host
+	} else if strings.Contains(err.Error(), "password authentication failed") {
```
ahrav, line 207: nit: A switch is a lot easier to read here imo. We can also capture the final `return false, nil` in the default block:
```
switch {
	case err == nil:
		return true, nil
	case strings.Contains(err.Error(), "password authentication failed"):
		return false, nil
	case errors.Is(err, pq.ErrSSLNotSupported) && params[pg_sslmode] == "":
		// If the sslmode is unset, then either it was unset in the candidate secret, or we've intentionally unset it
		// because it was specified as 'allow' or 'prefer', neither of which pq supports. In all of these cases, non-SSL
		// connections are acceptable, so now we try a connection without SSL.
		params[pg_sslmode] = pg_sslmode_disable
		defer delete(params, pg_sslmode) // We want to return with the original params map intact (for ExtraData)
		return verifyPostgres(params)
	case isErrorDatabaseNotFound(err, params[pg_dbname]):
		return true, nil // If we know this, we were able to authenticate
	default:
		return false, nil
	}
```

pkg/detectors/postgres/postgres_test.go
```diff
@@ -24,9 +25,9 @@ var postgresDockerHash string
 
 const (
 	postgresUser = "postgres"
-	postgresPass = "23201dabb56ca236f3dc6736c0f9afad"
+	postgresPass = "23201da=b56ca236f3dc6736c0f9afad"
 	postgresHost = "localhost"
-	postgresPort = "5433"
+	postgresPort = "5434" // Do not use 5433, as local dev environments can use it for other things
 
 	inactiveUser = "inactive"
```
ahrav, line 32: I think this is unused.

ahrav: same with inactivePort. I know you didn't touch these, but figured while i'm here i'd call it out.

ahrav (approved):
Thanks for reworking this ❤️

pkg/detectors/postgres/postgres_test.go
```diff
@@ -62,195 +78,245 @@ func TestPostgres_FromChunk(t *testing.T) {
...
 		},
 		{
-			name: "found, unverified due to error - inactive port",
-			s:    Scanner{},
-			args: args{
-				ctx:    context.Background(),
-				data:   []byte(fmt.Sprintf(`postgresql://%s:%s@%s:%s/postgres`, postgresUser, postgresPass, postgresHost, inactivePort)),
-				verify: true,
-			},
+			name: "found connection URI, unverified due to error - wrong port",
```
0x1, line 242: nit: probably should be "no port" instead of "wrong port"?

0x1 (approved):
nice, thanks for cleaning this up!

rosecodym:
the failing test was caused by a new gitlab project - i'll fix in a follow-up pr

pkg/detectors/postgres/postgres.go
```diff
@@ -3,63 +3,133 @@ package postgres
...
+// happen to run into a case where this matters we can address it then.
 var (
-	_               detectors.Detector = (*Scanner)(nil)
-	uriPattern                         = regexp.MustCompile(`\b(?i)postgresql://[\S]+\b`)
-	hostnamePattern                    = regexp.MustCompile(`(?i)(?:host|server|address).{0,40}?(\b[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\b)`)
-	portPattern                        = regexp.MustCompile(`(?i)(?:port|p).{0,40}?(\b[0-9]{1,5}\b)`)
-	usernamePattern                    = regexp.MustCompile(`(?im)(?:user|usr)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:'"\s]{4,40})`)
-	passwordPattern                    = regexp.MustCompile(`(?im)(?:pass)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:'"\s]{4,40})`)
+	_                  detectors.Detector = (*Scanner)(nil)
+	uriPattern                            = regexp.MustCompile(`\b(?i)postgres(?:ql)?://\S+\b`)
```
rgmz, line 48: This won't find things like `"postgresql+asyncpg://postgres:secret@localhost/testdb"`. Not sure what the + syntax is called.
https://github.com/agronholm/apscheduler/blob/b4ceea0ed300545a27bb8dbbbfb382a46d8ea90f/examples/web/asgi_starlette.py#L59

rosecodym: Hm, I was just going off the [official connection URI docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS), and I don't see that mentioned anywhere. Do you have any references you can point me to?

rgmz: It appears to be a Python-specific quirk:
```
dialect+driver://username:password@host:port/database
```
https://www.tutorialspoint.com/sqlalchemy/sqlalchemy_dialects.htm

rosecodym: Ah, ok. For this first pass we're aiming for correctness over completeness, because the previous implementation was yielding a ton of false positives. The eventual plan is to add other types of Postgres secrets (non-URI connection strings, at the very least), so we can add SQLAlchemy-extended URIs to that list.

PR #2316, opened 2024-01-19
updating detector logic for zenscrape

joeleonjr:
### Description:
Zenscrape verification logic currently does a string compare on the body of a proxied request looking for "Not enough requests". I think their API changed b/c there is now [a /status endpoint that can be used for verification](https://app.zenscrape.com/documentation#:~:text=Authentification%20%26%20Apikey%20Information) instead.

Logic has been updated to use status codes instead of results.

### Checklist:
* [ x] Tests passing (`make test-community`)?
* [ x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

### DIFF
--- a/pkg/gitparse/gitparse_test.go
+++ b/pkg/gitparse/gitparse_test.go
@@ -741,16 +741,6 @@ func TestIndividualCommitParsing(t *testing.T) {
 			}
 			j++
 		}
-		//for _, pass := range test.passes {
-		//	if !test.function(false, pass.latestState, pass.line) {
-		//		t.Errorf("%s: Parser did not recognize correct line. (%s)", name, string(pass.line))
-		//	}
-		//}
-		//for _, fail := range test.fails {
-		//	if test.function(false, fail.latestState, fail.line) {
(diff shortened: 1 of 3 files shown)


rosecodym:
What made you decide not to use the new /status endpoint?

joeleonjr:
The existing logic was showing false positives.

> What made you decide not to use the new /status endpoint?

The existing logic was showing false positives. Any uuid near zenscrape was showing verified for me.

dustin-decker (approved):

ahrav (approved):

joeleonjr:
> > The existing logic was showing false positives.
> > > What made you decide not to use the new /status endpoint?
> > 
> > 
> > The existing logic was showing false positives. Any uuid near zenscrape was showing verified for me.
> 
> Gotcha. Would you mind putting that in a comment so that some bright-eyed maintainer doesn't come along in six months and "fix" it?

I misread your question @rosecodym. I thought you asked why *did* I use the /status endpoint. I swore I committed my update using the /status endpoint. Just reviewed and I definitely didn't. Looks like I only committed the status code logic updates. I'm updating this PR to use the status endpoint, which is what I intended to do this whole time.

PR #2317, opened 2024-01-19
[chore] - Update Chunk struct comment

ahrav:
…ering

### Description:
Update the comment of the `Chunk` struct to inform users the ordering of the fields are significant and have memory implications.

### Checklist:

### DIFF
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -2,7 +2,6 @@ package sources
 
 import (
 	"sync"
-	"unsafe"
 
 	"google.golang.org/protobuf/types/known/anypb"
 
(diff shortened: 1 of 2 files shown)


rgmz:
If you want to be paranoid, you could hard-code a compile time check that will panic if the size fails.

rosecodym (approved):
❤️

rosecodym:
> If you want to be paranoid, you could hard-code a compile time check that will panic if the size changes.

I think @ahrav thought you were joking, but...I kinda like this? [It doesn't look too weird.](https://commaok.xyz/post/compile-time-assertions/)

ahrav:
what about a test as well? Something like:
```
// TestChunkSize ensures that the Chunk struct does not exceed 80 bytes.
func TestChunkSize(t *testing.T) {
	assert.Equal(t, unsafe.Sizeof(Chunk{}), uintptr(80), "Chunk struct size exceeds 80 bytes")
}
```

ahrav:
> If you want to be paranoid, you could hard-code a compile time check that will panic if the size changes.

Is there an easy way to do this without having to import C as in the example Cody mentioned?

rgmz:
Oh my. I was naively thinking there was a way to evaluate expressions at compile-time. I think I confused it with `go generate` or runtime init panics (e.g., placing it inside `init {}`).

Go makes me sad.

ahrav:
> Oh my. I was naively thinking there was a way to evaluate expressions at compile-time. I think I confused it with `go generate` or runtime init panics (e.g., placing it inside `init {}`).
> 
> Go makes me sad.
> 
> > oooo maybe something cute like:
> 
> This is both amazing and terrifying.

Hahaha, this was fun, but I might just let the test catch changes 🤣

dustin-decker (approved):

PR #2324, opened 2024-01-22
[feat] - Replace regexp pkg w/ go-re2 in detectors

ahrav:
### Description:
Replace the standard library regexp with [go-re2](https://github.com/wasilibs/go-re2), a drop-in regexp library that benchmarks significantly faster for our typical detector input size (100 bytes - 10kB). go-re2 shows close to 10x performance gains on large test data, which matches the [~10kB chunk size](https://github.com/trufflesecurity/trufflehog/blob/4afc224c635d10e732119f715f93788af1502ce4/pkg/sources/chunker.go#L14) our detectors operate on. By swapping the regexp library, we stand to see major efficiency improvements in our detectors without changing any logic. go-re2 is optimized for input in the 100 byte to 10kB range, making it a natural fit for our ~10kB chunk data. This change will enable faster and more scalable detectors as our input grows over time.

Note: The go-re2 authors state the library performs worse than standard regexp on tiny (~10 byte) input data, as shown in benchmarks. However, given our typical chunk size of ~10kB, we expect go-re2 to provide better performance despite this tradeoff. Our input rarely drops to 10 byte sizes, so we are making an informed decision to optimize for mid-sized data at the small cost of edge case efficiency.

Note: We cannot use a go mod replace directive since that is not allowed on standard library packages. The regexp import path is thus updated for all detectors. [ref thread](https://github.com/golang/go/issues/35283#issuecomment-550458470) in which Russ explains why this is not possible.

Screenshot 2024-01-21 at 4 11 45 PM

### Checklist:


zricethezav (approved):
Amazing! This looks like the best regex replacement lib out there:

> go-re2 is a drop-in replacement for the standard library [regexp](https://pkg.go.dev/regexp) package which uses the C++ [re2](https://github.com/google/re2) library for improved performance with large inputs or complex expressions. By default, re2 is packaged as a WebAssembly module and accessed with the pure Go runtime, [wazero](https://wazero.io/). This means that it is compatible with any Go application, regardless of availability of cgo.

very cool

trufflesteeeve:
It's potentially not relevant, but I wanted to check with y'all on how we handle non-utf-8 strings currently. Mostly because of this quote from go-re2:

>Invalid utf-8 strings are treated differently. The standard library silently replaces invalid utf-8 with the unicode replacement character. This library will stop consuming strings when encountering invalid utf-8.

dustin-decker:
> It's potentially not relevant, but I wanted to check with y'all on how we handle non-utf-8 strings currently. Mostly because of this quote from go-re2:
> 
> > Invalid utf-8 strings are treated differently. The standard library silently replaces invalid utf-8 with the unicode replacement character. This library will stop consuming strings when encountering invalid utf-8.

Great catch. Sounds like something we need to address first.

ahrav:
I added some tests in the huggingface_test.go file to test with [invalid utf-8](https://go.dev/play/p/H7HDto9mHvR) sequences.

ahrav:
> It's potentially not relevant, but I wanted to check with y'all on how we handle non-utf-8 strings currently. Mostly because of this quote from go-re2:
> 
> > Invalid utf-8 strings are treated differently. The standard library silently replaces invalid utf-8 with the unicode replacement character. This library will stop consuming strings when encountering invalid utf-8.

I added some tests in the [huggingface_test.go](https://github.com/trufflesecurity/trufflehog/pull/2324/files#diff-b37c766759dc4658efc5d8c6bf6e982ea5cc19ed076adadca153a6b1d9507e08R131-R179) file to test with [invalid utf-8](https://go.dev/play/p/H7HDto9mHvR) sequences. So I think we might be okay? I'm wondering if the authors of the library were referring strictly to strings and not []byte? since we use []byte in the detectors it appears it still detects the credentials correctly.

rosecodym (approved):

rgmz:
I would test the base64 decoder as well. It seems to produce invalid (?) unicode sequences if the entire input isn't valid base64.

PR #2329, opened 2024-01-22
updating doppler logic

joeleonjr:
### Description:
Doppler has [6 secret types](https://docs.doppler.com/reference/auth-token-formats). TruffleHog only previously supported 2. This adds support for all 6. Additionally, doppler's API has a [/me](https://docs.doppler.com/reference/auth-me) endpoint that can be used for verification. This endpoint also returns basic info about the token; those details have been added to the `ExtraData` field.

Tests have been updated to reflect the `ExtraData` field.

### Checklist:
* [ x] Tests passing (`make test-community`)?
* [ x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?


pkg/detectors/doppler/doppler.go
```diff
@@ -44,20 +55,44 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 				defer res.Body.Close()
 				if res.StatusCode >= 200 && res.StatusCode < 300 {
 					s1.Verified = true
+
+					body, err := io.ReadAll(res.Body)
+					if err != nil {
+						body = []byte{}
+					}
+
+					var data map[string]interface{}
```
zricethezav, line 79: **suggestion**: The API spec indicates that `type` will be a `string`. We could `var data map[string]string` instead of `interface`. This way we can avoid casting to strings later in the code.

Screenshot 2024-01-23 at 8 51 21 AM

Alternatively, and thinking about this a bit more, we could just shove the entire body into a response struct (only defining the data we care about) rather than relying on multiple maps. https://mholt.github.io/json-to-go/ is a useful tool for converting json to go.

zricethezav: https://github.com/trufflesecurity/trufflehog/pull/1945/files is a good example on how to load the response into a struct

joeleonjr: updates added.

dustin-decker (approved):
Nice improvement, thanks Joe.

PR #2107, opened 2023-11-09
[thog-1548] log verification errors

0x1:
### Description:
Update plaintext logging to include verification error

#### Current Output
image

#### Updated Output
image

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #2335, opened 2024-01-24
updates to plain and json printing to include verification error

0x1:
### Description:
Same as this [PR](https://github.com/trufflesecurity/trufflehog/pull/2107) but only printer changes

Update plaintext logging to include verification error

#### Current Output
image

#### Updated Output
image

### Checklist:

### DIFF
--- a/pkg/output/json.go
+++ b/pkg/output/json.go
@@ -17,6 +17,13 @@ import (
 type JSONPrinter struct{ mu sync.Mutex }
 
 func (p *JSONPrinter) Print(_ context.Context, r *detectors.ResultWithMetadata) error {
+	verificationErr := func(err error) string {
+		if err != nil {
+			return err.Error()
+		}
+		return ""
+	}(r.VerificationError())
+
 	v := &struct {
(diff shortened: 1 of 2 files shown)


rosecodym (approved):
Great call breaking this out!

I don't want to design by committee here, but it's a pretty noticeable UI change and I think it would be a good idea to at least get one more pair of eyes on it (maybe @zricethezav ?)

0x1:
> Great call breaking this out!
> 
> I don't want to design by committee here, but this is a pretty noticeable UI change and I think it would be a good idea to at least get one more pair of eyes on it (maybe @zricethezav ?)

Agreed. On that point, would it be worth having a flag that can be used to revert back to old format? I can imagine some people may have gotten used to it. Or there could be case where people output the log into a file and use diff to see changes, this could break that.

dustin-decker (approved):
Looks great to me. Side note - I think the error shown for AWS could be considered determinately not verified and we should probably handle that case. I saw a few of those in our dataset as well. I think it comes from invalid combinations of matches.

PR #2337, opened 2024-01-25
added azurefunctionkey detector

roxanne-tampus:
### Description:
Added AzureFunctionKey detector

### Checklist:


pkg/detectors/azurefunctionkey/azurefunctionkey.go
```diff
@@ -0,0 +1,89 @@
...
+// FromData will find and optionally verify azure secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+	urlMatches := azureUrlPat.FindAllStringSubmatch(dataStr, -1)
+	for _, match := range matches {
+		resMatch := strings.TrimSpace(match[0])
+		for _, urlMatch := range urlMatches {
+			resUrl := strings.TrimSpace(urlMatch[0])
+			fmt.Println(resUrl)
```
dustin-decker, line 44: Make sure not to have any print functions in the detectors. We should add a linter to help prevent this.

pkg/detectors/azurefunctionkey/azurefunctionkey.go
```diff
@@ -0,0 +1,89 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat      = regexp.MustCompile(`\b[a-zA-Z0-9_-]{54,56}==`)
```
dustin-decker, line 25: It looks like this might be base64, so the ending might not always be `==`. Can you check a few more keys to confirm? Also, we should probably have a PrefixRegex here since it is not a very unique pattern.

dustin-decker: Also is this for all access key types? Is there any way to tell the difference?

https://learn.microsoft.com/en-us/azure/azure-functions/security-concepts?tabs=v4#function-access-keys

roxanne-tampus: I just updated the guide for this https://github.com/trufflesecurity/how-to-rotate/pull/16

dustin-decker (approved):

dustin-decker (approved):

PR #2340, opened 2024-01-26
Add Personio detector

nyanhp:
### Description:
This test adds support for Personio (<https://personio.com>), an HR tool containing sensitive data where a leaked key could have enormous impact. The provider expects two components, a GUID prefixed with `papi-` as well as a 48 character string prefixed with `papi-`. While API keys can be scoped to specific fields, what we can test is the authentication, which returns a bearer token. Regardless of the key's scopes the authentication will work if the credentials are still valid.

### Checklist:

### DIFF
--- a/pkg/detectors/personio/personio.go
+++ b/pkg/detectors/personio/personio.go
@@ -3,9 +3,10 @@ package personio
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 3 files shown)


zricethezav:
@nyanhp lgtm, just need to resolve merge conflicts

PR #2342, opened 2024-01-26
Update the template detector

rgmz:
### Description:
This introduces following changes to the template detector + generate script:

- Fix a bug where `generate.go` creates `DetectorType_Detectorname` instead of `DetectorType_DetectorName`
  This could be further improved by searching for the exact DetectorType casing. Matching the input name is a good enough solution for now, imo.
- De-duplicate matches prior to attempting verification, that way we don't waste time+io repeatedly verifying the same data 
- Moves verification logic out of the for loop in `FromData` and drains + closes the response body (context in #2086)
- Added a keyword+pattern test so that basic issues can be caught without having to run tests with a live secret

### Checklist:

### DIFF
--- a/pkg/detectors/alchemy/alchemy.go
+++ b/pkg/detectors/alchemy/alchemy.go
@@ -3,9 +3,10 @@ package alchemy
 import (
 	"context"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
+	"io"
 	"net/http"
-	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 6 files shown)


pkg/engine/ahocorasick/ahocorasickcore.go
```diff
@@ -1,9 +1,10 @@
-package engine
+package ahocorasick
```
rgmz, line 1: This had to be moved out of the engine package, per https://github.com/trufflesecurity/trufflehog/pull/2264#discussion_r1437829049.

pkg/detectors/alchemy/alchemy.go
```diff
@@ -31,57 +32,69 @@ func (s Scanner) Keywords() []string {
...
 
-// FromData will find and optionally verify Alchemy secrets in a given set of bytes.
+// FromData will find and optionally verifyMatch Alchemy secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
 
-	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
-
-	for _, match := range matches {
-		if len(match) != 2 {
```
rgmz, line 41: This doesn't actually do what it's meant to, and will cause silent failures if the number of match groups is changed.

https://github.com/trufflesecurity/trufflehog/pull/2264#discussion_r1443382434

dustin-decker (approved):
Thank you!

PR #2348, opened 2024-01-29
added azuresearchadminkey detector

roxanne-tampus:
### Description:
Added AzureSearchAdminKey detector

### Checklist:


pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
```diff
@@ -0,0 +1,100 @@
...
+		resMatch := strings.TrimSpace(match[1])
+		for _, serviceMatch := range serviceMatches {
+			if len(serviceMatch) != 2 {
+				continue
+			}
+			resServiceMatch := strings.TrimSpace(serviceMatch[1])
+
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_AzureSearchAdminKey,
+				Raw:          []byte(resMatch + resServiceMatch),
```
dustin-decker, line 56: Small correction to what I told you previously: Raw should contain the key match, and RawV2 should contain all components. This is because we use part of the Raw field to provide a redacted value to help people find the leak.
In this case it starts with the key value, so it will be fine as-is.

dustin-decker (approved):

PR #2356, opened 2024-01-30
Polite Verification

ahrav:
### Description:
This PR changes the default behavior of verifying secrets to be more polite. If a single secret is detected by two or more different detectors in a single chunk, then verification of that secret is disabled.

You can override this behavior by using the `--allow-verification-overlap` option.

### Checklist:


main.go
```diff
@@ -49,6 +49,7 @@ var (
 	concurrency         = cli.Flag("concurrency", "Number of concurrent workers.").Default(strconv.Itoa(runtime.NumCPU())).Int()
 	noVerification      = cli.Flag("no-verification", "Don't verify the results.").Bool()
 	onlyVerified        = cli.Flag("only-verified", "Only output verified results.").Bool()
+	forceReverification = cli.Flag("allow-verification-overlap", "Allow verification of similar credentials across detectors").Bool()
```
dustin-decker, line 52: Please update this flag naming

pkg/engine/engine.go
```diff
@@ -92,6 +97,20 @@ type Engine struct {
 
 	// verify determines whether the scanner will attempt to verify candidate secrets
 	verify bool
+
+	// Note: bad hack only used for testing
+	reverificationTracking *reverificationTracking
```
dustin-decker, line 102: Nit: consider if we should rename this and other related variables to be more close to the flag name

main.go
zricethezav: Should we rename all the `reverification` elements? `reverificationWorkers`, etc

pkg/engine/engine.go
```diff
@@ -492,10 +584,135 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+// in that chunk - which is expected and not malicious. Those intra-detector dupes are still verified.
+type chunkSecretKey struct {
+	secret       string
+	detectorInfo ahocorasick.DetectorInfo
+}
+
+func likelyDuplicate(ctx context.Context, val chunkSecretKey, dupes map[chunkSecretKey]struct{}) bool {
+	const similarityThreshold = 0.9
+
+	valStr := val.secret
```
dustin-decker, line 605: What happened to the exact match case? We can do a very efficient exact match lookup in the map without iteration.

pkg/engine/engine.go
```diff
@@ -492,10 +584,135 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+				detectorsWithResult[detector] = struct{}{}
+			}
+
+			for _, res := range results {
+				var val []byte
+				if res.RawV2 != nil {
+					val = res.RawV2
+				} else {
+					val = res.Raw
+				}
```
dustin-decker, line 658: Should we track and check these both separately? Could we potentially have a raw v2 in a different order, or a detector with a raw but not a rawv2 and it doesn't seem like a likely duplicate?

pkg/engine/engine.go
ahrav: We are using a new key type:
```
// chunkSecretKey ties secrets to the specific detector that found them. This allows identifying identical
// credentials extracted by multiple different detectors processing the same chunk. Or duplicates found
// by the same detector in the chunk. Exact matches on lookup indicate a duplicate secret for a detector
// in that chunk - which is expected and not malicious. Those intra-detector dupes are still verified.
type chunkSecretKey struct {
	secret       string
	detectorInfo ahocorasick.DetectorInfo
}
```

So an exact match is actually okay given it means we found the same cred multiple times in the chunk with the same detector.

zricethezav: @ahrav https://github.com/trufflesecurity/trufflehog/pull/2356/commits/96c6cd518d8cc8a28f77ce5e0caebc6692bb2b57 any reason the exact match case was removed in this commit? Should we add it back in?

ahrav: So the idea is if there is an identical match that means they DetectorKey AND secret are identical. If that's true then they are the same cred in the same detector. The same secret across multiple detectors would not be identical matches, but they would get caught by the Levenshtein similarity check.

pkg/engine/engine.go
```diff
@@ -492,10 +584,135 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+		}
+
+		similarity := strutil.Similarity(valStr, dupe, metrics.NewLevenshtein())
+
+		// close enough
+		if similarity > similarityThreshold {
+			ctx.Logger().V(2).Info(
+				"found similar duplicate",
+				"val", val,
+				"dupe", dupe,
```
rosecodym, line 620: Logging raw secrets feels like an accident waiting to happen. I know that they ultimately get printed anyway, but I'm worried about someone accidentally copying this somewhere it shouldn't be. Is it necessary to do here?

pkg/engine/engine.go
```diff
@@ -492,10 +584,135 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+					val = res.Raw
+				}
+
+				// Use levenstein distance to determine if the secret is likely the same.
+				// Ex:
+				// - postman api key: PMAK-qnwfsLyRSyfCwfpHaQP1UzDhrgpWvHjbYzjpRCMshjt417zWcrzyHUArs7r
+				// - malicious detector "api key": qnwfsLyRSyfCwfpHaQP1UzDhrgpWvHjbYzjpRCMshjt417zWcrzyHUArs7r
+				key := chunkSecretKey{secret: string(val), detectorInfo: detector}
+				if _, ok := chunkSecrets[key]; ok {
+					chunkSecrets[key] = struct{}{}
```
rosecodym, line 666: Why do you rewrite the map entry here?

pkg/engine/engine.go
```diff
@@ -492,10 +584,135 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+					continue
+				}
+
+				if likelyDuplicate(ctx, key, chunkSecrets) {
```
rosecodym, line 670: Isn't this always going to be true because `chunkSecrets` always contains `key`?

pkg/engine/engine.go
```diff
@@ -492,10 +584,135 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+					}
+					e.processResult(ctx, detectableChunk{
+						chunk:    chunk.chunk,
+						detector: detector,
+						decoder:  chunk.decoder,
+						wgDoneFn: wgDetect.Done,
+					}, res)
+
+					// Remove the detector from the list of detectors with results.
+					delete(detectorsWithResult, detector)
```
rosecodym, line 684: And won't this only remove later detectors in a collision set, but not the first? (Assuming my previous concern about `likelyDuplicate` always returning true is misplaced, anyway.)

rosecodym (commented):
I have a few questions about the overlap detection logic itself

pkg/engine/engine.go
```diff
@@ -300,10 +334,25 @@ func (e *Engine) initialize(ctx context.Context, options ...Option) error {
 	if err != nil {
 		return fmt.Errorf("failed to initialize LRU cache: %w", err)
 	}
+	const (
+		// detectableChunksChanMultiplier is set to accommodate a high number of concurrent worker goroutines.
+		// This multiplier ensures that the detectableChunksChan channel has sufficient buffer capacity
+		// to hold messages from multiple worker groups (detector workers/ verificationOverlap workers) without blocking.
+		// A large buffer helps accommodate for the fact workers are producing data at a faster rate
+		// than it can be consumed.
```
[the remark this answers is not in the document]
ahrav: You raise a good point. If the producers are generating data faster than the consumers can process it, there is a risk of the consumers becoming a bottleneck, which could limit overall throughput. This would be true regardless of buffer size.

Simply increasing the buffer capacity provides only a temporary solution, as the buffer will eventually fill up if the consumers can't keep pace. A more robust approach may be to scale the number of consumer goroutines to match the production rate. This helps ensure the consumers have enough total capacity to handle all of the data being produced in parallel. This would likely require a bit more testing and fine-tuning so I took the faster approach in the interim.

pkg/engine/engine.go
ahrav: Err... great question... no reason to re-write this entry. I have removed it. thanks for catching it.

pkg/engine/engine.go
ahrav: It should not since `likelyDuplicate` only compares the secret values across different detectors. The chunkSecretKey struct allows efficient lookup to handle two cases:

1.  Identical secret + detector pairs indicate a duplicate secret found by the same detector. In this case, we verify the result since it flags duplicate secrets within a chunk.
2. Different secret values trigger the likelyDuplicate check. If secrets are similar enough across different detectors, it likely indicates a global secret so we skip verification.

Using the custom struct avoids extra map lookups and clearly separates the two cases - verifying identical intra-chunk dupes while skipping inter-detector global secrets. The goal is to catch duplicate issues for a single detector, while allowing broadly matched secrets across detectors to avoid unnecessary verifications.

pkg/engine/engine.go
ahrav: Yep, you are correct. The `detectorsWithResult` should be a `map[ahocorasick.DetectorInfo]struct{}` instead.

pkg/engine/engine.go
zricethezav: @dustin-decker rip it out entirely or log a heavily redacted value? Something like the first 5 characters or so?

pkg/engine/engine.go
rosecodym: Oh, I just...misread? Weird. Ok!

pkg/engine/engine.go
zricethezav: Probably worth to do something like 
```
		if valStr == dupe && val.detectorInfo != dupeKey.detectorInfo {
			ctx.Logger().V(2).Info(
				"found exact duplicate",
			)
			return true
		}
```

In the loop before hitting it with a similarity calculation?

rgmz:
> If a single secret is detected by two or more different detectors in a single chunk, then verification of that secret is disabled.

What are the potential implications of this? Would versioned detectors that match the same secret, such as [GitLab](https://github.com/trufflesecurity/trufflehog/blob/c2ae31d06077155dda16eec3dd85a0e46c431629/pkg/detectors/gitlab/gitlab.go#L53-L55) , result in verification being disabled?

dustin-decker:
> > If a single secret is detected by two or more different detectors in a single chunk, then verification of that secret is disabled.
> 
> What are the potential implications of this? Would versioned detectors that match the same secret, such as [GitLab](https://github.com/trufflesecurity/trufflehog/blob/c2ae31d06077155dda16eec3dd85a0e46c431629/pkg/detectors/gitlab/gitlab.go#L53-L55) , result in verification being disabled?

Richard brings up a good point here. I think the answer is yes with the current code. Using the proto name of the detector instead within detector info might resolve that.

ahrav:
> > > If a single secret is detected by two or more different detectors in a single chunk, then verification of that secret is disabled.
> > 
> > 
> > What are the potential implications of this? Would versioned detectors that match the same secret, such as [GitLab](https://github.com/trufflesecurity/trufflehog/blob/c2ae31d06077155dda16eec3dd85a0e46c431629/pkg/detectors/gitlab/gitlab.go#L53-L55) , result in verification being disabled?
> 
> Richard brings up a good point here. I think the answer is yes with the current code. Using the proto name of the detector instead within detector info might resolve that.

Well for the Gitlab detector this works correctly because the regexes are different enough that the `likelyDuplicate` will return `false`. If they regexes were similar, then yes we'd be marking them as invalid. I don't believe we currently have any versioned detectors that fall into this category, but this could be an issue with custom detectors with regexes that are very similar.

I think [this test](https://github.com/trufflesecurity/trufflehog/pull/2356/files#diff-b16f55bb6289c9df88deaedc99a45db09154ba13ee3a186b7c5636d67f5b345dR223-R256) is an example.

ahrav:
> > > > If a single secret is detected by two or more different detectors in a single chunk, then verification of that secret is disabled.
> > > 
> > > 
> > > What are the potential implications of this? Would versioned detectors that match the same secret, such as [GitLab](https://github.com/trufflesecurity/trufflehog/blob/c2ae31d06077155dda16eec3dd85a0e46c431629/pkg/detectors/gitlab/gitlab.go#L53-L55) , result in verification being disabled?
> > 
> > 
> > Richard brings up a good point here. I think the answer is yes with the current code. Using the proto name of the detector instead within detector info might resolve that.
> 
> Well for the Gitlab detector this works correctly because the regexes are different enough that the `likelyDuplicate` will return `false`. If the regexes were similar, then yes we'd be marking them as unverified. I don't believe we currently have any versioned detectors that fall into this category, but this could be an issue with custom detectors with regexes that are very similar.
> 
> I think [this test](https://github.com/trufflesecurity/trufflehog/pull/2356/files#diff-b16f55bb6289c9df88deaedc99a45db09154ba13ee3a186b7c5636d67f5b345dR223-R256) is an example.

Added some logic to handle the edge case, and included a [test](https://github.com/trufflesecurity/trufflehog/pull/2356/files#diff-b16f55bb6289c9df88deaedc99a45db09154ba13ee3a186b7c5636d67f5b345dR265-R317)

pkg/engine/engine.go
```diff
@@ -492,10 +584,144 @@ func (e *Engine) detectorWorker(ctx context.Context) {
...
+
+		similarity := strutil.Similarity(valStr, dupe, metrics.NewLevenshtein())
+
+		// close enough
+		if similarity > similarityThreshold {
+			ctx.Logger().V(2).Info(
+				"found similar duplicate",
+			)
+			return true
+		}
```
dustin-decker, line 634: I think the last consideration I have for this PR is that we need to let the user know that we've encountered the same/similar finding across multiple different detectors and will not be verifying the result for their safety, and they can override the behavior with --allow-verification-overlap. The current messages are not adequate. It's probably best surfaced through VerificationError, where it will show in the output.

rgmz: This could always be release in audit/warn mode to gauge impact before enabling it in a later release.

dustin-decker: Added verification erro

dustin-decker (approved):

pkg/engine/engine_test.go
```diff
@@ -214,6 +222,140 @@ func TestEngine_DuplicatSecrets(t *testing.T) {
 	assert.Equal(t, want, e.GetMetrics().UnverifiedSecretsFound)
 }
 
+// TestEngine_VersionedDetectorsVerifiedSecrets is a test that detects ALL verified secrets across
+// versioned detectors.
+func TestEngine_VersionedDetectorsVerifiedSecrets(t *testing.T) {
```
rgmz, line 227: BTW this breaks `make test-community` since the secrets aren't accessible.

```
--- FAIL: TestEngine_VersionedDetectorsVerifiedSecrets (0.01s)
    engine_test.go:231:
                Error Trace:    /tmp/trufflehog/pkg/engine/engine_test.go:231
                Error:          Received unexpected error:
                                failed to create secretmanager client: google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information
                                github.com/trufflesecurity/trufflehog/v3/pkg/common.GetSecret
                                        /tmp/trufflehog/pkg/common/secrets.go:55
```

skeweredlogic: Likely - yes. The beginning of the trace looks the same. The aftermath is a panic and then segfault.

PR #2364, opened 2024-02-01
[feat] - concurently scan the filesystem source

ahrav:
### Description:
concurrently scan files within the fileystem source.

### Checklist:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -96,7 +96,7 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 		fileInfo, err := os.Stat(cleanPath)
 		if err != nil {
 			logger.Error(err, "unable to get file info")
-			return nil
+			continue
 		}
 
 		if fileInfo.IsDir() {
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/go-logr/logr"
 	diskbufferreader "github.com/trufflesecurity/disk-buffer-reader"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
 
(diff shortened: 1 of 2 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -93,7 +96,7 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 		fileInfo, err := os.Stat(cleanPath)
 		if err != nil {
 			logger.Error(err, "unable to get file info")
-			continue
+			return nil
```
rosecodym, line 99: This seems like a really big change! Is it related to concurrency? Should it be in this PR?

pkg/sources/filesystem/filesystem_test.go
```diff
@@ -71,16 +72,23 @@ func TestSource_Scan(t *testing.T) {
...
 					return
 				}
 			}()
-			gotChunk := <-chunksCh
-			if diff := pretty.Compare(gotChunk.SourceMetadata, tt.wantSourceMetadata); diff != "" {
-				t.Errorf("Source.Chunks() %s diff: (-got +want)\n%s", tt.name, diff)
+			var counter int
+			for chunk := range chunksCh {
+				counter++
+				if chunk.SourceMetadata.GetFilesystem().GetFile() == "filesystem.go" {
```
rosecodym, line 85: Do you think it's worth checking that this ever happened?

rosecodym (commented):
The concurrency looks as straightforward as concurrency can ever be (not much) but I'm nervous about the added early return elsewhere.

pkg/sources/filesystem/filesystem.go
ahrav: Whoops this was left behind when we had first put the concurrency within this level. reverted it.

pkg/sources/filesystem/filesystem.go
```diff
@@ -126,9 +135,13 @@ func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sour
 			return nil
 		}
 
-		if err = s.scanFile(ctx, fullPath, chunksChan); err != nil {
-			ctx.Logger().Info("error scanning file", "path", fullPath, "error", err)
-		}
+		workerPool.Go(func() error {
+			if err = s.scanFile(ctx, fullPath, chunksChan); err != nil {
+				ctx.Logger().Info("error scanning file", "path", fullPath, "error", err)
```
mcastorina, line 140: Should this be an `Error` log instead of `Info`?

ahrav: yea probably.

pkg/sources/filesystem/filesystem_test.go
ahrav: I can use the counter to capture this happening.

mcastorina (approved):

PR #2366, opened 2024-02-02
[bug] - use DetectorKey as the key in the detectorKeysWithResults map

ahrav:
### Description:
The current map key contains unhashable fields. Use the `DetectorKey` instead as the key in the map.
[
Screenshot 2024-02-02 at 12 37 54 PM
](url)

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -644,7 +644,7 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 
 	// Reuse the same map and slice to avoid allocations.
 	const avgSecretsPerDetector = 8
-	detectorsWithResult := make(map[ahocorasick.DetectorInfo]struct{}, avgSecretsPerDetector)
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -694,18 +694,18 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
...
 		}
 
-		for detector := range detectorsWithResult {
+		for key := range detectorKeysWithResults {
 			wgDetect.Add(1)
 			chunk.chunk.Verify = e.verify
 			e.detectableChunksChan <- detectableChunk{
 				chunk:    chunk.chunk,
-				detector: detector,
+				detector: e.ahoCorasickCore.GetDetectorByKey(key),
```
zricethezav, line 708: pretty sure this is not gonna happen but might be worth it to check for nil before sending the chunk since `GetDetectorByKey` could return nil

zricethezav (approved):

PR #2367, opened 2024-02-03
Implement SourceUnitEnumChunker for GitLab

mcastorina:
### Description:
Implements `SourceUnitEnumChunker` in GitLab for enumerating and chunking units.

### Checklist:

### DIFF
--- a/pkg/engine/gitlab_integration_test.go
+++ b/pkg/engine/gitlab_integration_test.go
@@ -36,6 +36,6 @@ func TestGitLab(t *testing.T) {
 
 	// Check the output provided by metrics.
 	metrics := e.GetMetrics()
-	assert.GreaterOrEqual(t, metrics.ChunksScanned, uint64(36390))
-	assert.GreaterOrEqual(t, metrics.BytesScanned, uint64(92662889))
+	assert.GreaterOrEqual(t, metrics.ChunksScanned, uint64(36312))
+	assert.GreaterOrEqual(t, metrics.BytesScanned, uint64(91618854))
 }
--- a/pkg/engine/gitlab_integration_test.go
+++ b/pkg/engine/gitlab_integration_test.go
@@ -1,0 +1,41 @@
+//go:build integration
+// +build integration
+
+package engine
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+func TestGitLab(t *testing.T) {
+	// Run the scan.
+	ctx := context.Background()
+	e, err := Start(ctx,
+		WithDetectors(DefaultDetectors()...),
+		WithVerify(false),
+	)
+	assert.NoError(t, err)
+
+	secret, err := common.GetTestSecret(ctx)
+	if err != nil {
+		t.Fatal(fmt.Errorf("failed to access secret: %v", err))
+	}
+	err = e.ScanGitLab(ctx, sources.GitlabConfig{
+		Token: secret.MustGetField("GITLAB_TOKEN"),
+	})
+	assert.NoError(t, err)
+
+	err = e.Finish(ctx)
+	assert.NoError(t, err)
+
+	// Check the output provided by metrics.
+	metrics := e.GetMetrics()
+	assert.GreaterOrEqual(t, metrics.ChunksScanned, uint64(36390))
+	assert.GreaterOrEqual(t, metrics.BytesScanned, uint64(92662889))
+}
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -190,7 +190,7 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 				return ctx.Err()
 			},
 		}
-		if err := s.getReposFromGitlab(ctx, apiClient, ignoreRepo, reporter); err != nil {
+		if err := s.getAllProjectRepos(ctx, apiClient, ignoreRepo, reporter); err != nil {
 			return err
 		}
 	}
@@ -251,21 +251,24 @@ func (s *Source) Validate(ctx context.Context) []error {
 		errs = append(errs, fmt.Errorf("could not compile ignore repo pattern %q: %w", pattern, err))
 	})
 
-	// getAllProjects only uses the reporter for reporting errors, so the
-	// empty VisitorReporter is okay to use.
-	projects, err := s.getAllProjects(ctx, apiClient, sources.VisitorReporter{})
-	if err != nil {
+	// Query GitLab for the list of configured repos.
+	var repos []string
+	visitor := sources.VisitorReporter{
+		VisitUnit: func(ctx context.Context, unit sources.SourceUnit) error {
+			repos = append(repos, unit.SourceUnitID())
+			return nil
+		},
+	}
+	if err := s.getAllProjectRepos(ctx, apiClient, ignoreProject, visitor); err != nil {
 		errs = append(errs, err)
 		return errs
 	}
 
-	for _, p := range projects {
-		if !ignoreProject(p.PathWithNamespace) {
-			return errs
-		}
+	if len(repos) == 0 {
+		errs = append(errs, fmt.Errorf("ignore patterns excluded all projects"))
 	}
 
-	return append(errs, fmt.Errorf("ignore patterns excluded all projects"))
+	return errs
 }
 
 func (s *Source) newClient() (*gitlab.Client, error) {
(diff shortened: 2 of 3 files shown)


pkg/sources/gitlab/gitlab.go
```diff
@@ -406,8 +428,17 @@ func (s *Source) getAllProjects(ctx context.Context, apiClient *gitlab.Client) (
...
+// using the provided API client. Ignored repositories will be skipped. The
+// reporter, if provided, is used for reporting errors that would only
+// otherwise be logged, and reporting repositories as they are found.
+func (s *Source) getReposFromGitlab(
+	ctx context.Context,
+	apiClient *gitlab.Client,
+	ignoreRepo func(string) bool,
+	reporter sources.UnitReporter,
+) ([]string, error) {
+	projects, err := s.getAllProjects(ctx, apiClient, reporter)
```
mcastorina, line 441: It would be nice if this could stream the projects so we can start getting the repos early, but I'll leave that for a future optimization.

pkg/engine/gitlab_integration_test.go
```diff
@@ -0,0 +1,41 @@
...
+	})
+	assert.NoError(t, err)
+
+	err = e.Finish(ctx)
+	assert.NoError(t, err)
+
+	// Check the output provided by metrics.
+	metrics := e.GetMetrics()
+	assert.GreaterOrEqual(t, metrics.ChunksScanned, uint64(36390))
+	assert.GreaterOrEqual(t, metrics.BytesScanned, uint64(92662889))
```
rosecodym, line 40: Are you using `GreaterOrEqual` to guard against future stuff being added to Gitlab?

pkg/sources/gitlab/gitlab.go
```diff
@@ -344,7 +355,11 @@ func (s *Source) getAllProjects(ctx context.Context, apiClient *gitlab.Client) (
 	for {
 		userProjects, res, err := apiClient.Projects.ListUserProjects(user.ID, projectQueryOptions)
 		if err != nil {
-			return nil, fmt.Errorf("received error on listing user projects: %w", err)
+			err = fmt.Errorf("received error on listing user projects: %w", err)
+			if err := reportErr(ctx, err); err != nil {
+				return nil, err
+			}
+			break
```
rosecodym, line 362: Does using `break` instead of `return` mean that this error is now non-fatal when running with units?

pkg/sources/gitlab/gitlab.go
```diff
@@ -368,7 +383,11 @@ func (s *Source) getAllProjects(ctx context.Context, apiClient *gitlab.Client) (
 	for {
 		groupList, res, err := apiClient.Groups.ListGroups(&listGroupsOptions)
 		if err != nil {
-			return nil, fmt.Errorf("received error on listing groups, you probably don't have permissions to do that: %w", err)
+			err = fmt.Errorf("received error on listing groups, you probably don't have permissions to do that: %w", err)
+			if err := reportErr(ctx, err); err != nil {
+				return nil, err
+			}
+			break
```
rosecodym, line 390: same question about `break`

pkg/sources/gitlab/gitlab.go
```diff
@@ -564,3 +606,77 @@ func normalizeRepos(repos []string) ([]string, []error) {
...
+				return err
+			}
+		}
+		return nil
+	}
+
+	// Otherwise, enumerate all repos.
+	ignoreRepo := buildIgnorer(s.ignoreRepos, func(err error, pattern string) {
+		ctx.Logger().Error(err, "could not compile ignore repo glob", "glob", pattern)
+		// TODO: Handle error returned from UnitErr.
```
rosecodym, line 649: Any reason you can't just return it? An un-compilable ignore glob sounds pretty fatal to me.

rosecodym (commented):
this mostly looks good but i have a small question about some control flow

pkg/sources/gitlab/gitlab.go
mcastorina: Yes, it's non-fatal for both units and non-units. I thought it made sense to do so instead of quitting early.

pkg/sources/gitlab/gitlab.go
mcastorina: Yep, this was intentional. Do you think it's okay to do this? I'll double check the flow.

pkg/sources/gitlab/gitlab.go
mcastorina: It's in a callback right now, so it's just adding the error to the report. This comment is referring to the error when adding the error to the report, which can happen from context cancellation. If that's the case, it'll throw the same error when returning later in this function, so I don't think it's worth the effort to escape the error(s) out of the closure.

But you bring up a good point about un-compilable globs - should we make it fatal? Right now it appears they are not.

pkg/sources/gitlab/gitlab.go
rosecodym: Ah right, both cases, derp. I think it's fine to do, just wanted to ensure I understood.

pkg/sources/gitlab/gitlab.go
rosecodym: Oh, right, derp again. It's probably a good thing to handle at some point but I don't think you need to do it as part of this PR. (I believe it's written that way so that `Validate` can detect it, so it's not like it's completely undiscoverable.)

rosecodym (approved):
Looks pretty straightforward. I see that you're accommodating the legacy `Chunks` workflow by passing a `nil` reporter and then checking for reporter nil-ness. One idea I'm playing around with in the Confluence sources is creating a local reporter that does the "legacy" logging, rather than just passing `nil`. Is that something you considered?

mcastorina:
> Looks pretty straightforward. I see that you're accommodating the legacy `Chunks` workflow by passing a `nil` reporter and then checking for reporter nil-ness. One idea I'm playing around with in the Confluence sources is creating a local reporter that does the "legacy" logging, rather than just passing `nil`. Is that something you considered?

Ooh interesting. So calling `ChunksOk` on the reporter would send the chunk on the channel, and `ChunksErr` would log the error? I like it.

That reminded me of the [ChanReporter](https://github.com/trufflesecurity/trufflehog/blob/v3.67.3/pkg/sources/chan_reporter.go#L10-L13), and it turns out it already does exactly that lol. I should use it.

What about for `UnitOk` and `UnitErr`? `UnitOk` appends to a slice maybe? 🤔

pkg/sources/gitlab/gitlab.go
```diff
@@ -246,19 +251,24 @@ func (s *Source) Validate(ctx context.Context) []error {
...
 		errs = append(errs, err)
 		return errs
 	}
 
-	for _, p := range projects {
-		if !ignoreProject(p.PathWithNamespace) {
-			return errs
-		}
+	if len(repos) == 0 {
+		errs = append(errs, fmt.Errorf("ignore patterns excluded all projects"))
```
rosecodym, line 268: lol this is way clearer than my old code that you changed, thanks!

pkg/sources/gitlab/gitlab.go
```diff
@@ -310,28 +320,54 @@ func (s *Source) basicAuthSuccessful(apiClient *gitlab.Client) bool {
...
 
 	// Used to filter out duplicate projects.
-	processProjects := func(projList []*gitlab.Project) {
+	processProjects := func(projList []*gitlab.Project) error {
 		for _, proj := range projList {
-			if _, exists := uniqueProjects[proj.ID]; !exists {
-				uniqueProjects[proj.ID] = proj
-				projects = append(projects, proj)
-				projectsWithNamespace = append(projectsWithNamespace, proj.NameWithNamespace)
+			// Skip projects we've already seen.
```
rosecodym, line 345: Any idea what causes this?

pkg/sources/gitlab/gitlab.go
```diff
@@ -564,3 +587,74 @@ func normalizeRepos(repos []string) ([]string, []error) {
...
+	ignoreRepo := buildIgnorer(s.ignoreRepos, func(err error, pattern string) {
+		ctx.Logger().Error(err, "could not compile ignore repo glob", "glob", pattern)
+		// TODO: Handle error returned from UnitErr.
+		_ = reporter.UnitErr(ctx, fmt.Errorf("could not compile ignore repo glob: %w", err))
+	})
+	return s.getAllProjectRepos(ctx, apiClient, ignoreRepo, reporter)
+}
+
+// ChunkUnit downloads and reports chunks for the given GitLab repository unit.
+func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
```
rosecodym, line 637: Did you consider trying to unify this code path with the one in `scanRepos`? I ask because it includes some non-trivial logic that seems easy to forget about during later maintenance. (But I can see the argument against it: that some error handling and logic makes unifying them non-trivial itself.)

rosecodym (approved):
lgtm!

pkg/sources/gitlab/gitlab.go
mcastorina: There was a bit too much logic around progress / concurrency / error handling that it was easier to copy/paste and delete.

I'll try to unify in a followup PR.

pkg/sources/gitlab/gitlab.go
mcastorina: Not sure what you're referring to here. If you're referring to seeing projects we've already seen, then I _think_ it's because we're paginating so there might be some overlap in rare cases.

PR #2368, opened 2024-02-03
fixes calendly api key regex

ankushgoel27:
### Description:
This PR fixes the regex for the calendly api key. The current regex doesn't detect the JWT key properly and misses. This regex is more open and detects the calendly API key

calendly

### Checklist:


zricethezav:
confirmed calendly api generates jwt

zricethezav (approved):

PR #2373, opened 2024-02-03
[fixup] - correctly use the buffered file writer

ahrav:
### Description:
I forgot to actually wire up the parser in the `Git` struct to be used when calling `RepoPath`. Additionally we need to construct a new contentWriter instance for every diff, it can't be reused across diffs.

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -392,8 +392,15 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 	)
 	var latestState = Initial
 
-	writer := c.contentWriter()
-	currentDiff := NewDiff(withCustomContentWriter(writer))
+	diff := func(opts ...diffOption) *Diff {
+		return NewDiff(withCustomContentWriter(newBuffer()))
+	}
+	if c.useCustomContentWriter {
+		diff = func(opts ...diffOption) *Diff {
+			return NewDiff(withCustomContentWriter(bufferedfilewriter.New()))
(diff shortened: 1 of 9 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -100,21 +101,26 @@ func (b *buffer) String() (string, error) { return b.Buffer.String(), nil }
 // The use of contentWriter enables the management of diff data either in memory or on disk,
 // based on its size, optimizing resource usage and performance.
 type Diff struct {
-	PathB         string
-	LineStart     int
+	PathB     string
+	LineStart int
+	IsBinary  bool
+
```
ahrav, line 107: This was bugging me a bit so i just reordered the struct to group the fields a bit differently.

pkg/gitparse/gitparse.go
```diff
@@ -606,6 +612,13 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 	ctx.Logger().V(2).Info("finished parsing git log.", "total_log_size", totalLogSize)
 }
 
+func (c *Parser) contentWriter() contentWriter {
```
ahrav, line 615: There might be a cleaner way to do this, but my brain couldn't think of it. Happy to change this if anyone has a better approach.

rosecodym (approved):

pkg/gitparse/gitparse.go
ahrav: Yes, thanks you.

PR #2377, opened 2024-02-04
[not-fixup] - Reduce memory consumption for Buffered File Writer

ahrav:
### Description:
The current buffer handling allows buffers to frequently grow larger than necessary. When writing content smaller than the buffer threshold but larger than the buffer capacity, the buffer pool no longer serves its intended purpose. This unnecessary buffer growth causes a spike in memory usage.

This PR checks the write content size against the buffer's remaining capacity. If the buffer cannot accommodate the content, instead of doubling the buffer for a small amount of data, the buffer is grown to the exact size that's needed.. This approach reduces the buffered file writer's overall memory footprint.

Analysis of logging output revealed this pattern:
```
DATA SIZE: 64, BUFFER SIZE: 16371, CAPACITY: 16384
DATA SIZE: 62, BUFFER SIZE: 32726, CAPACITY: 32768
DATA SIZE: 44, BUFFER SIZE: 2026, CAPACITY: 2048
DATA SIZE: 57, BUFFER SIZE: 2038, CAPACITY: 2048
DATA SIZE: 49, BUFFER SIZE: 4063, CAPACITY: 4096
DATA SIZE: 51, BUFFER SIZE: 8154, CAPACITY: 8192
DATA SIZE: 18, BUFFER SIZE: 16376, CAPACITY: 16384
DATA SIZE: 61, BUFFER SIZE: 2038, CAPACITY: 2048
DATA SIZE: 50, BUFFER SIZE: 4069, CAPACITY: 4096
DATA SIZE: 38, BUFFER SIZE: 2037, CAPACITY: 2048
DATA SIZE: 58, BUFFER SIZE: 4048, CAPACITY: 4096
DATA SIZE: 41, BUFFER SIZE: 2030, CAPACITY: 2048
DATA SIZE: 68, BUFFER SIZE: 1983, CAPACITY: 2048
DATA SIZE: 40, BUFFER SIZE: 2035, CAPACITY: 2048
DATA SIZE: 75, BUFFER SIZE: 4071, CAPACITY: 4096
DATA SIZE: 52, BUFFER SIZE: 1998, CAPACITY: 2048
DATA SIZE: 63, BUFFER SIZE: 4036, CAPACITY: 4096
DATA SIZE: 36, BUFFER SIZE: 2015, CAPACITY: 2048
DATA SIZE: 52, BUFFER SIZE: 4063, CAPACITY: 4096
DATA SIZE: 42, BUFFER SIZE: 2038, CAPACITY: 2048
DATA SIZE: 63, BUFFER SIZE: 4084, CAPACITY: 4096
DATA SIZE: 9, BUFFER SIZE: 2041, CAPACITY: 2048
DATA SIZE: 40, BUFFER SIZE: 4093, CAPACITY: 4096
DATA SIZE: 27, BUFFER SIZE: 8182, CAPACITY: 8192
DATA SIZE: 65, BUFFER SIZE: 2034, CAPACITY: 2048
DATA SIZE: 97, BUFFER SIZE: 2029, CAPACITY: 2048
DATA SIZE: 56, BUFFER SIZE: 2006, CAPACITY: 2048
DATA SIZE: 49, BUFFER SIZE: 2019, CAPACITY: 2048
DATA SIZE: 52, BUFFER SIZE: 2014, CAPACITY: 2048
DATA SIZE: 44, BUFFER SIZE: 2023, CAPACITY: 2048
DATA SIZE: 51, BUFFER SIZE: 2043, CAPACITY: 2048
DATA SIZE: 46, BUFFER SIZE: 2032, CAPACITY: 2048
DATA SIZE: 70, BUFFER SIZE: 2024, CAPACITY: 2048
DATA SIZE: 9, BUFFER SIZE: 4096, CAPACITY: 4096
DATA SIZE: 79, BUFFER SIZE: 1996, CAPACITY: 2048
DATA SIZE: 30, BUFFER SIZE: 2032, CAPACITY: 2048
DATA SIZE: 101, BUFFER SIZE: 4093, CAPACITY: 4096
DATA SIZE: 46, BUFFER SIZE: 2007, CAPACITY: 2048
DATA SIZE: 1, BUFFER SIZE: 4096, CAPACITY: 4096
DATA SIZE: 55, BUFFER SIZE: 2005, CAPACITY: 2048
DATA SIZE: 33, BUFFER SIZE: 4085, CAPACITY: 4096
DATA SIZE: 64, BUFFER SIZE: 2025, CAPACITY: 2048
DATA SIZE: 47, BUFFER SIZE: 4077, CAPACITY: 4096
```

In all cases, the buffer was doubled despite needing only a small increase to accommodate the write content.

Profiles before and after this change:
Screenshot 2024-02-03 at 5 26 02 PM
Screenshot 2024-02-03 at 7 20 55 PM

Additionally, some performance profiles:
Screenshot 2024-02-03 at 5 48 18 PM

### Checklist:

### DIFF
--- a/pkg/engine/ahocorasick/ahocorasickcore.go
+++ b/pkg/engine/ahocorasick/ahocorasickcore.go
@@ -23,6 +23,9 @@ type DetectorKey struct {
 	customDetectorName string
 }
 
+// Type returns the detector type of the key.
+func (k DetectorKey) Type() detectorspb.DetectorType { return k.detectorType }
+
 // AhoCorasickCore encapsulates the operations and data structures used for keyword matching via the
 // Aho-Corasick algorithm. It is responsible for constructing and managing the trie for efficient
 // substring searches, as well as mapping keywords to their associated detectors for rapid lookups.
--- a/pkg/engine/ahocorasick/ahocorasickcore.go
+++ b/pkg/engine/ahocorasick/ahocorasickcore.go
@@ -63,6 +63,12 @@ func NewAhoCorasickCore(allDetectors []detectors.Detector) *AhoCorasickCore {
 	}
 }
 
+// GetDetectorByKey returns the detector associated with the given key. If no detector is found, it
+// returns nil.
+func (ac *AhoCorasickCore) GetDetectorByKey(key DetectorKey) detectors.Detector {
+	return ac.detectorsByKey[key]
+}
+
 // DetectorInfo represents a detected pattern's metadata in a data chunk.
 // It encapsulates the key identifying a specific detector and the detector instance itself.
 type DetectorInfo struct {
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -375,6 +375,9 @@ func (c *Parser) executeCommand(ctx context.Context, cmd *exec.Cmd, isStaged boo
 
 	go func() {
 		c.FromReader(ctx, stdOut, commitChan, isStaged)
+		if err := stdOut.Close(); err != nil {
+			ctx.Logger().V(2).Info("Error closing git stdout pipe.", "error", err)
+		}
 		if err := cmd.Wait(); err != nil {
 			ctx.Logger().V(2).Info("Error waiting for git command to complete.", "error", err)
 		}
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -392,8 +392,17 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 	)
 	var latestState = Initial
 
-	writer := c.contentWriter()
-	currentDiff := NewDiff(withCustomContentWriter(writer))
+	diff := func(opts ...diffOption) *Diff {
+		opts = append(opts, withCustomContentWriter(newBuffer()))
+		return NewDiff(opts...)
+	}
+	if c.useCustomContentWriter {
+		diff = func(opts ...diffOption) *Diff {
+			opts = append(opts, withCustomContentWriter(bufferedfilewriter.New()))
+			return NewDiff(opts...)
+		}
+	}
+	currentDiff := diff()
 
 	defer common.RecoverWithExit(ctx)
 	defer close(commitChan)
@@ -436,7 +445,8 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 				totalLogSize += currentCommit.Size
 			}
 			// Create a new currentDiff and currentCommit
-			currentDiff = NewDiff(withCustomContentWriter(c.contentWriter()))
+			currentCommit = &Commit{}
+			currentDiff = diff()
 			currentCommit = &Commit{Message: strings.Builder{}}
 			// Check that the commit line contains a hash and set it.
 			if len(line) >= 47 {
@@ -504,7 +514,8 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 					currentCommit.Message.WriteString(oldCommit.Message.String())
 				}
 			}
-			currentDiff = NewDiff(withCustomContentWriter(c.contentWriter()))
+			currentDiff = diff()
+			// currentDiff = NewDiff(withCustomContentWriter(c.contentWriter()))
 		case isModeLine(isStaged, latestState, line):
 			latestState = ModeLine
 			// NoOp
@@ -544,7 +555,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 				}
 				currentCommit.Diffs = append(currentCommit.Diffs, *currentDiff)
 			}
-			currentDiff = NewDiff(withCustomContentWriter(c.contentWriter()), withPathB(currentDiff.PathB))
+			currentDiff = diff(withPathB(currentDiff.PathB))
 
 			words := bytes.Split(line, []byte(" "))
 			if len(words) >= 3 {
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	bufferedfilewriter "github.com/trufflesecurity/trufflehog/v3/pkg/writers/buffered_file_writer"
 )
 
 const (
@@ -100,10 +101,11 @@ func (b *buffer) String() (string, error) { return b.Buffer.String(), nil }
 // The use of contentWriter enables the management of diff data either in memory or on disk,
 // based on its size, optimizing resource usage and performance.
 type Diff struct {
-	PathB         string
-	LineStart     int
+	PathB     string
+	LineStart int
+	IsBinary  bool
+
 	contentWriter contentWriter
-	IsBinary      bool
 }
 
 type diffOption func(*Diff)
@@ -111,10 +113,14 @@ type diffOption func(*Diff)
 // withPathB sets the PathB option.
 func withPathB(pathB string) diffOption { return func(d *Diff) { d.PathB = pathB } }
 
+// withCustomContentWriter sets the useCustomContentWriter option.
+func withCustomContentWriter(cr contentWriter) diffOption {
+	return func(d *Diff) { d.contentWriter = cr }
+}
+
 // NewDiff creates a new Diff with a threshold.
 func NewDiff(opts ...diffOption) *Diff {
 	diff := new(Diff)
-	diff.contentWriter = newBuffer()
 	for _, opt := range opts {
 		opt(diff)
 	}
@@ -200,10 +206,10 @@ func (c1 *Commit) Equal(ctx context.Context, c2 *Commit) bool {
 
 // Parser sets values used in GitParse.
 type Parser struct {
-	maxDiffSize   int
-	maxCommitSize int
-	dateFormat    string
-	contentWriter contentWriter
+	maxDiffSize            int
+	maxCommitSize          int
+	dateFormat             string
+	useCustomContentWriter bool
 }
 
 type ParseState int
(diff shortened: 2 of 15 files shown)


pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -100,33 +157,45 @@ func (w *BufferedFileWriter) Write(ctx context.Context, data []byte) (int, error
...
+		growSize := int(totalSizeNeeded) - bufferLength
+		if growSize > availableSpace {
+			ctx.Logger().V(4).Info(
+				"buffer size exceeded, growing buffer",
+				"current_size", bufferLength,
+				"new_size", totalSizeNeeded,
+				"available_space", availableSpace,
+				"grow_size", growSize,
+			)
+			w.bufPool.growBufferWithSize(w.buf, growSize)
```
rosecodym, line 196: I understand that the purpose of this change is to save memory, but it's doing so by losing the guaranteed (amortized) O(1) write _time_. This is probably fine (since it's all "constant" anyway, given our hard threshold) but I think it would be good to explicitly acknowledge it in a comment somewhere so that some future maintainer doesn't see it and get concerned.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -167,7 +235,7 @@ func (w *BufferedFileWriter) CloseForWriting() error {
 	}
 
 	if w.buf.Len() > 0 {
-		_, err := w.file.Write(w.buf.Bytes())
+		_, err := w.buf.WriteTo(w.file)
```
rosecodym, line 238: oh you sneaky guy

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Could you elaborate a bit more on this? My understanding was that without this growth the underlying buffer was going to have to be grown anyways. The only difference being we are controlling the growth instead of doubling the growth by default. If we don't grow the buffer here, the default behavior would be w.Write would allocate a new buffer with twice the cap and copy the current contents into the new buffer.

I might be missing something that feels important in your comment.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -13,12 +13,65 @@ import (
...
+		buf = bytes.NewBuffer(make([]byte, 0, bp.bufferSize))
+	}
+
+	return buf
+}
+
+func (bp *bufferPool) growBufferWithSize(buf *bytes.Buffer, size int) {
+	// Grow the buffer to accommodate the new data.
+	buf.Grow(size)
+}
```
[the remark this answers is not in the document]
mcastorina: I think that makes sense, however, from the POV of treating a PR as an individual change, I agree that it should be inlined here, and the followup PR can break it out into a helper function/method if needed.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -13,12 +13,65 @@ import (
...
+
+func (bp *bufferPool) put(buf *bytes.Buffer) {
+	// If the buffer is more than twice the default size, replace it with a new, smaller one.
+	// This prevents us from returning very large buffers to the pool.
+	const maxAllowedCapacity = 2 * defaultBufferSize
+	if buf.Cap() > maxAllowedCapacity {
+		// Replace the buffer with a new, smaller one. No need to copy data since we're resetting it.
+		buf = bytes.NewBuffer(make([]byte, 0, defaultBufferSize))
```
mcastorina, line 68: Does this mean large buffers get garbage collected?

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -100,33 +157,45 @@ func (w *BufferedFileWriter) Write(ctx context.Context, data []byte) (int, error
...
 		// If the total size is within the threshold, write to the buffer.
 		ctx.Logger().V(4).Info(
 			"writing to buffer",
 			"data_size", size,
-			"content_size", w.buf.Len(),
+			"content_size", bufferLength,
 		)
+
+		availableSpace := w.buf.Cap() - bufferLength
+		growSize := int(totalSizeNeeded) - bufferLength
```
mcastorina, line 187: `totalSizeNeeded` is `bufferLength + len(data)`, so is `growSize` equivalent to `len(data)`?

That seems wrong, or I'm misunderstanding what `growSize` means.

ahrav: No, the `growSize` is the length of the current buffer + the amount of data needed to be written to the buffer. It's the size the buffer needs to be to handle all the existing data and any new data. If you call `Grow` with a size that is less than less than the remainder of the buffer nothing happens. This is why you can't do something linke `totalSizeNeeded - w.buf.Cap()` you need to use the length for it to grow correctly.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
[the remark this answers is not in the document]
ahrav: done.

pkg/engine/ahocorasick/ahocorasickcore.go
```diff
@@ -23,6 +23,9 @@ type DetectorKey struct {
 	customDetectorName string
 }
 
+// Type returns the detector type of the key.
+func (k DetectorKey) Type() detectorspb.DetectorType { return k.detectorType }
+
```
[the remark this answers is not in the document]
ahrav: forgot to rebase

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: nvm i ripped it out.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
rosecodym: I had to read this part like four times as well. I was going to complain until I noticed that that the old code was equally confusing so I decided to just blame the buffer interface.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -78,17 +131,16 @@ func (w *BufferedFileWriter) String() (string, error) {
...
 	defer file.Close()
 
-	// Create a buffer large enough to hold file data and additional buffer data, if any.
-	fileSize := w.size
-	buf := bytes.NewBuffer(make([]byte, 0, fileSize))
-
+	var buf bytes.Buffer
 	// Read the file contents into the buffer.
-	if _, err := io.Copy(buf, file); err != nil {
+	if _, err := io.CopyBuffer(&buf, file, nil); err != nil {
```
rosecodym, line 136: What's the effect of this change?

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -143,14 +206,13 @@ func (w *BufferedFileWriter) Write(ctx context.Context, data []byte) (int, error
...
 		// This ensures all the data is in one place - either entirely in the buffer or the file.
-		if w.buf.Len() > 0 {
-			ctx.Logger().V(4).Info("writing buffer to file", "content_size", w.buf.Len())
-			if _, err := w.file.Write(w.buf.Bytes()); err != nil {
+		if bufferLength > 0 {
+			ctx.Logger().V(4).Info("writing buffer to file", "content_size", bufferLength)
+			if _, err := w.buf.WriteTo(w.file); err != nil {
 				return 0, err
 			}
 			// Reset the buffer to clear any existing data and return it to the pool.
```
rosecodym, line 214: i think this comment needs to be updated too

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -143,14 +206,13 @@ func (w *BufferedFileWriter) Write(ctx context.Context, data []byte) (int, error
 
 		// Transfer existing data in buffer to the file, then clear the buffer.
```
rosecodym, line 207: i think this comment needs to be updated

rosecodym (commented):
Is this change now just ensuring that all buffers are at a particular minimum size?

ahrav:
> Is this change now just ensuring that all buffers are at a particular minimum size?

Correct, an few optimizations around copying buffers.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Should be identical exepct we use `CopyBuffer` elsewhere in the codebase so I was trying to be consistent.
```
// CopyBuffer is identical to Copy except that it stages through the
// provided buffer (if one is required) rather than allocating a
// temporary on
```

pkg/gitparse/gitparse.go
```diff
@@ -449,7 +449,6 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 			}
 			// Create a new currentDiff and currentCommit
 			currentDiff = diff()
-			// currentDiff = NewDiff(withCustomContentWriter(c.contentWriter()))
```
ahrav, line 452: Small cleanup here.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: We could also explicitly set the buffer to nil.. maybe i'll do that so it's obvious.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Yea I am ashamed to say I spent WAY too long trying to fully grok the buffer interface... there are definitely some gotchas that took me way way too long to comprehend. 😞

rosecodym: Don't be ashamed! Intuitive API design is difficult and sometimes impossible. Some stuff is just hard to understand.

rosecodym (approved):
What are the optimizations around copying buffers? I can see the one where we drop large buffers upon return instead of reusing them, but are there others?

ahrav:
> What are the optimizations around copying buffers? I can see the one where we drop large buffers upon return instead of reusing them, but are there others?

ah, just the use of `WriteTo` in the two places. it's supposed to be a bit more performant.

Issue #2371, opened 2024-02-03
Enumerate GitHub wikis concurrently


rgmz:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description

PR #2233 added the ability to scan GitHub wikis by default, as wikis are just repos and scanning them has no rate limit. Regrettably, I overlooked how dispatching a request per repo can increase startup time. This does not have a noticeable effect when scanning repos or smaller orgst, however, it does have a noticeable impact for larger orgs.

https://github.com/trufflesecurity/trufflehog/blob/27b30e65edc98506647e099161a1e1aeb30dc9e9/pkg/sources/github/repo.go#L238-L240

Without this check, enumeration takes ~3s per page (I have terrible Internet):
```
...
2024-02-03T10:52:55-05:00       info-1  trufflehog      Enumerating with token  {"source_manager_worker_id": "cKjLf", "endpoint": "https://api.github.com"}
2024-02-03T10:52:58-05:00       info-2  trufflehog      Listed repos    {"page": 0, "last_page": 61}
2024-02-03T10:53:01-05:00       info-2  trufflehog      Listed repos    {"page": 2, "last_page": 61}
2024-02-03T10:53:04-05:00       info-2  trufflehog      Listed repos    {"page": 3, "last_page": 61}
```

With this check, enumeration takes closer to ~15s per page:
```
...
2024-02-03T10:57:03-05:00       info-1  trufflehog      Enumerating with token  {"source_manager_worker_id": "gZtkz", "endpoint": "https://api.github.com"}
2024-02-03T10:57:06-05:00       info-2  trufflehog      Listed repos    {"page": 0, "last_page": 61}
2024-02-03T10:57:20-05:00       info-2  trufflehog      Listed repos    {"page": 2, "last_page": 61}
2024-02-03T10:57:33-05:00       info-2  trufflehog      Listed repos    {"page": 3, "last_page": 61}
```

+ @rosecodym 

### Preferred Solution

Two potential solutions come to mind:

1. Run this check in a Go routine so that it does not block enumeration
2. Run this check when each repo is being scanned, not during initial enumeration.

https://github.com/trufflesecurity/trufflehog/blob/27b30e65edc98506647e099161a1e1aeb30dc9e9/pkg/sources/github/github.go#L804-L815

[...]

PR #2379, opened 2024-02-05
Refactor GitHub source

rgmz:
### Description:
This PR refactors the code for the GitHub source. The main changes are:

* Consolidate information like size and visibility into a single `repoInfo` cache.
* Remove the unnecessary API request in `visibilityOf` by storing visibility information during initial enumeration (as mentioned in https://github.com/trufflesecurity/trufflehog/pull/2308#issuecomment-1894435215).
* Move the `git.CmdCheck` to the top of the Init method, as there's no point in doing additional work if Git isn't installed correctly.
* Remove unnecessary logic under `scanComments`, such as passing in the repoUrl (string), the repoUrl (url.URL), and the repo URL in segments.
* Move `hasWiki` check to happen during repo scanning, not at startup
* Refactor `scanErrs.String()` to produce distinct error messages rather than one long string.

### Checklist:


rosecodym:
I'm guessing that this should probably wait for #2385 and possibly the resolution to #2371 in order to avoid merge conflicts.

rgmz:
> I'm guessing that this should probably wait for #2385 and possibly the resolution to #2371 in order to avoid merge conflicts.

I've moved the wiki logic from `repo.go` -> `github.go` to reduce the impact.

pkg/sources/github/github.go
```diff
@@ -793,36 +705,41 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
 			}
 
 			// Scan the repository
+			repoInfo, ok := s.repoInfoCache.get(repoURL)
+			if !ok {
+				// This should never happen.
+				err := fmt.Errorf("no repoInfo for URL: %s", repoURL)
+				s.log.Error(err, "failed to scan repository")
```
rgmz, line 712: This seems to get populated when using `--org` but not `--repo. Gotta look into that.

mcastorina: Quickly looked at this and found:

* `conn.Repositories` is set via the `--repo` flag
* `Init` takes the repos from there, normalizes them, and saves it in the `filteredRepoCache`
* `enumerate` sets `s.repos` by iterating over `filteredRepoCache`

At no point is the repo added to the `repoInfoCache`, triggering this case.

pkg/sources/github/github.go
```diff
@@ -1126,29 +1052,49 @@ func (s *Source) setProgressCompleteWithRepo(index int, offset int, repoURL stri
...
-	ctx.Logger().V(2).Info("scanning github gist comments", "repository", repoPath)
+// trimURLAndSplit removes extraneous information from the |url| and splits it into segments.
+// This is typically 3 segments: host, owner, and name/ID; however, Gists have some edge cases.
+//
+// Examples:
+// - "https://github.com/trufflesecurity/trufflehog" => ["github.com", "trufflesecurity", "trufflehog"]
+// - "https://gist.github.com/nat/5fdbb7f945d121f197fb074578e53948" => ["gist.github.com", "nat", "5fdbb7f945d121f197fb074578e53948"]
+// - "https://gist.github.com/ff0e5e8dc8ec22f7a25ddfc3492d3451.git" => ["gist.github.com", "ff0e5e8dc8ec22f7a25ddfc3492d3451"]
+func trimURLAndSplit(url string) []string {
+	trimmedURL := strings.TrimPrefix(url, "https://")
```
mcastorina, line 1084: Do you think it's worth it to use `url.Parse` here? For example if the input is `ssh://github.com/...`?

mcastorina (commented):
Thanks for taking this on, it's looking really good!

I'll take another pass once the `--repo` flag works.

pkg/sources/github/github.go
```diff
@@ -793,36 +705,41 @@ func (s *Source) scan(ctx context.Context, installationClient *github.Client, ch
...
 			if s.includeGistComments || s.includeIssueComments || s.includePRComments {
-				if err = s.scanComments(ctx, repoURL, chunksChan); err != nil {
+				if err = s.scanComments(repoCtx, repoURL, repoInfo, chunksChan); err != nil {
 					scanErrs.Add(fmt.Errorf("error scanning comments in repo %s: %w", repoURL, err))
 					return nil
 				}
 			}
 
-			ctx.Logger().V(2).Info(fmt.Sprintf("scanned %d/%d repos", scannedCount, len(s.repos)), "duration_seconds", duration)
+			repoCtx.Logger().V(1).Info(fmt.Sprintf("scanned %d/%d repos", scannedCount, len(s.repos)), "duration_seconds", duration)
```
bill-rich, line 742: I think this might be too verbose for V(1)

pkg/sources/github/github.go
```diff
@@ -853,13 +770,12 @@ func (s *Source) cloneAndScanRepo(ctx context.Context, client *github.Client, re
...
-	if !strings.HasSuffix(repoURL, ".wiki.git") {
-		repoSize := s.repoSizes.getRepo(repoURL)
-		logger = ctx.Logger().WithValues("repo_size_kb", repoSize)
+	if !strings.HasSuffix(repoURL, ".wiki.git") && repoInfo.size > 0 {
+		logger = ctx.Logger().WithValues("repo_size_kb", repoInfo.size)
 	} else {
 		logger = ctx.Logger()
 	}
-	logger.V(2).Info("scanning repo")
+	logger.V(1).Info("scanning repo")
```
bill-rich, line 778: Same about verbosity.

pkg/sources/github/github.go
bill-rich: +1

pkg/sources/github/github.go
```diff
@@ -1175,17 +1121,47 @@ func (s *Source) processGistComments(ctx context.Context, repoPath string, trimm
 	return nil
 }
 
-func extractGistID(url []string) (string, error) {
-	if len(url) < 2 || len(url) > 3 {
-		return "", fmt.Errorf("failed to parse Gist URL: length of trimmedURL should be 2 or 3")
+func extractGistID(url []string) string {
+	return url[len(url)-1]
```
bill-rich, line 1125: A check that `len(url)` is at least `1`

pkg/sources/github/repo.go
```diff
@@ -266,6 +306,7 @@ func (s *Source) hasWiki(ctx context.Context, repo *github.Repository, repoURL s
 	if err != nil {
 		return false
 	}
+	_, _ = io.Copy(io.Discard, res.Body)
```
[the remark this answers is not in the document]
bill-rich: It looks like its to clear out `res.Body`, but I thought that would happen with `res.Body.Close()`
@rgmz is that not the case?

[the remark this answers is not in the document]
rgmz: One might reasonably assume that `res.Body.Close()` would handle everything required to close a response body. That is not necessarily the case.

See: #2086

pkg/sources/github/repo.go
```diff
@@ -2,19 +2,56 @@ package github
...
+	r.cache[repoURL] = info
+}
+
+func (r *repoInfoCache) get(repoURL string) (*repoInfo, bool) {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	info, ok := r.cache[repoURL]
+	return info, ok
+}
```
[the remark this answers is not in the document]
rgmz: I'm pretty sure this is write-once read-many, so the use of pointers probably isn't necessary.

pkg/sources/github/repo.go
```diff
@@ -251,11 +295,7 @@ func (s *Source) processRepos(ctx context.Context, target string, listRepos repo
 
 // hasWiki returns true if the "has_wiki" property is true AND https://github.com/$org/$repo/wiki is not redirected.
 // Unfortunately, this isn't 100% accurate. Some repositories meet both criteria yet don't have a cloneable wiki.
-func (s *Source) hasWiki(ctx context.Context, repo *github.Repository, repoURL string) bool {
-	if !repo.GetHasWiki() {
-		return false
-	}
```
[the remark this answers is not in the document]
rgmz: Makes sense.

pkg/sources/github/github.go
rgmz: This method isn't used anywhere else, so it could be inlined in `scanComments` which already calls `url.Parse` via `git.GitURLParse`.

https://github.com/trufflesecurity/trufflehog/blob/7620906b075a7ec53926b4e5eafee4b3ff9c2c13/pkg/sources/github/github.go#L1132-L1133

https://github.com/trufflesecurity/trufflehog/blob/7620906b075a7ec53926b4e5eafee4b3ff9c2c13/pkg/sources/git/git.go#L365-L378

I have added some tests and updated the code a bit to handle SSH, though.

pkg/sources/github/github.go
rgmz: I removed the length check because it seemed a bit redundant. The length is checked in `scanComments` and this is only called in one place.

This one-liner could be inlined into `processGistComments` so it's closer.

pkg/sources/github/repo.go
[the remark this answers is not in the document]
bill-rich: Interesting. Didn't know that was the case. Thanks!

bill-rich (approved):

rosecodym:
@rgmz we've traced a number of regressions to this PR, not all of which we've fully resolved yet, and we're considering just rolling it back and re-doing it as multiple separate changes instead. Do you think there would be any unexpected negative consequences to doing this? The description doesn't read like it has anything world-saving in it, but I don't want to miss anything.

rgmz:
Apologies, the GitHub source logic is perhaps more of a complex beast than I realized. I recently tried to unify the `enumerate` logic to fix behavioural inconsistencies between auth types (e.g., [unauthenticated scans users passed to `--org`](https://github.com/trufflesecurity/trufflehog/blob/becce51335f98090b9729b00a740aed3bac16664/pkg/sources/github/github.go#L502-L509), but the other methods don't seem to.)

> @rgmz we've traced a number of regressions to this PR, not all of which we've fully resolved yet, and we're considering just rolling it back and re-doing it as multiple separate changes instead. Do you think there would be any unexpected negative consequences to doing this? The description doesn't read like it has anything world-saving in it, but I don't want to miss anything.

I'm not sure. Reverting this would likely also require reverting or fixing up a few other changes.

Most of this (mainly https://github.com/trufflesecurity/trufflehog/pull/2379/commits/063b44b769eab0af753e2042c1891cbbc069d2a8) would be difficult to break into smaller changes, as structural changes like adding `s.repoInfoCache` to reduce duplicated API calls necessitate moving a lot around. I suspect that some of the issues, such as https://github.com/trufflesecurity/trufflehog/pull/2625#issuecomment-2032075025, are not net-new bugs but rather surfacing of latent ones.

I'm happy to continue to help diagnose and push fixes, but it's your call.

rgmz:
> >  Do you think there would be any unexpected negative consequences to doing this? The description doesn't read like it has anything world-saving in it, but I don't want to miss anything.
>
> Most of this (https://github.com/trufflesecurity/trufflehog/commit/063b44b769eab0af753e2042c1891cbbc069d2a8) would be difficult to break into smaller changes, as structural changes like adding s.repoInfoCache to reduce duplicated API calls necessitates moving a lot around.

I may have buried the lede a bit. The primary motivation for this PR was reducing unnecessary API calls; most, but not all, of the subsequent refactoring was a consequence of that.

A negative consequence would quicker exhaustion of the GitHub rate limit.

> Remove the unnecessary API request in visibilityOf by storing visibility information during initial enumeration (as mentioned in https://github.com/trufflesecurity/trufflehog/pull/2308#issuecomment-1894435215).

PR #2381, opened 2024-02-05
added flyio detector

lonmarsDev:
### Description:

### Checklist:


pkg/detectors/flyio/flyio.go
```diff
@@ -0,0 +1,104 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"flyio"}) + `\b([a-zA-Z0-9_]{47})\b`)
```
zricethezav, line 26: @lonmarsDev all the tokens I generated started with `fo1_`. Does that match what you were seeing? If so we should probably add `fo1_` as part of this regex and to the keywords.

zricethezav:
We might want to rename this as `flyiopersonalaccesstoken` since fly.io also has organization tokens (see https://fly.io/docs/flyctl/tokens-create-org/ and `https://fly.io/dashboard/{user_id}/tokens`). Alternatively we could just leave this as `flyio` and use `flyioorgtoken` for the organizational token detector when we get to that.

pkg/detectors/flyio/flyio.go
btoews: Fly.io has a few legacy token formats that are no longer issued, but still might exist in the wild:

- `[a-zA-Z0-9_-]{43}`
- `fm1[ar]_[a-zA-Z0-9+\/]={0,2}`

And two formats that we currently issue

- `fo1_[a-zA-Z0-9_-]{43}`
- `fm2_[a-zA-Z0-9+\/]={0,2}`

amanfcp:
Hey @lonmarsDev, thanks for the contribution.

Can you please follow [this](https://github.com/trufflesecurity/trufflehog/blob/main/hack/docs/Adding_Detectors_external.md) document to follow conventions of adding a new detector?

pkg/detectors/flyio/flyio.go
amanfcp: @btoews Thanks. Can you please give me a resource/link where this is mentioned? Asking as I need to document it.

amanfcp: So far, the tokens (both app and org tokens) that I have created start with `FlyV1 fm2_`. Haven't seen a token with `fo1_` prefix yet.

pkg/detectors/flyio/flyio.go
```diff
@@ -0,0 +1,150 @@
...
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(FlyV1 fm\d+_[A-Za-z0-9+\/=,_-]{500,700})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"flyio", "FlyV1", "fm2_"}
```
nabeelalam: `fm2_` seems like a commonly occurring substring so I'm not sure if it would be an effective keyword. I think `FlyV1` is specific enough to be the sole required keyword here, since it's always part of the secret.

pkg/detectors/flyio/flyio.go
```diff
@@ -0,0 +1,150 @@
...
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	// Not setting org_slug intentionally, as it's not required for the token to be valid.
+	// Initially, an organization named "personal" is created by FlyIO when the user signs up for an account. We cannot rely on this as it can be deleted.
+	// 403 is returned if incorrect org_slug is sent.
+	// 401 is returned if the token is invalid.
+	// 400 is returned if the token is valid but no org_slug is sent.
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.machines.dev/v1/apps?org_slug=", nil)
```
nabeelalam: nit: Can use `http.NoBody` instead of `nil`

shahzadhaider1 (approved):

pkg/detectors/flyio/flyio.go
```diff
@@ -0,0 +1,150 @@
...
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(FlyV1 fm\d+_[A-Za-z0-9+\/=,_-]{500,700})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"flyio", "FlyV1"}
```
nabeelalam: According to the current regex, `FlyV1` is always a part of the flyio key. If that is the case, then we shouldn't need the `flyio` keyword to filter.

nabeelalam (changes requested):

pkg/detectors/flyio/flyio.go
shahzadhaider1: I noticed that as well, but then I figured that since "flyio" is a relatively unique term, it’s unlikely to introduce noise, so I decided to leave it as is.

nabeelalam (approved):
Thanks for the update @amanfcp !

kashifkhan0771 (approved):

PR #2385, opened 2024-02-05
Revert "Scan GitHub wikis #2233"

rosecodym:
### Description:
This reverts commit 8e90c4e669ff1c2e25153896156fd94eb6d05d25 (PR #2233). As pointed out in #2371, #2233 introduces a drastic performance penalty during GitHub source enumeration. There are ways around it, but we should back the original change out while we decide which one we want to do and do it.

#2386 just disables the new option by default, making it a less destruction solution to the problem. I've created them both for comparison.

### Checklist:


rosecodym:
Closed in favor of #2386

PR #2387, opened 2024-02-05
[feat] - use diff chan

ahrav:
### Description:
Replaces the commit chan with a diff chan.

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -151,8 +151,8 @@ func (d *Diff) finalize() error { return d.contentWriter.CloseForWriting() }
 type Commit struct {
 	Hash    string
 	Author  string
-	Message strings.Builder
 	Date    time.Time
+	Message strings.Builder
 	Size    int // in bytes
 }
 
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -249,6 +249,7 @@ func NewParser(options ...Option) *Parser {
 }
 
 // RepoPath parses the output of the `git log` command for the `source` path.
+// The Diff chan will returns diffs in the order they are parsed from the log.
 func (c *Parser) RepoPath(ctx context.Context, source string, head string, abbreviatedLog bool, excludedGlobs []string, isBare bool) (chan *Diff, error) {
 	args := []string{"-C", source, "log", "-p", "--full-history", "--date=format:%a %b %d %H:%M:%S %Y %z"}
 	if abbreviatedLog {
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -121,7 +121,7 @@ func withCustomContentWriter(cr contentWriter) diffOption {
 }
 
 // newDiff creates a new Diff with a threshold and an associated commit.
-// All Diff's must have an associated commit.
+// All Diffs must have an associated commit.
 // The contentWriter is used to manage the diff's content, allowing for flexible handling of diff data.
 // By default, a buffer is used as the contentWriter, but this can be overridden with a custom contentWriter.
 func newDiff(commit *Commit, opts ...diffOption) *Diff {
@@ -251,7 +251,7 @@ func NewParser(options ...Option) *Parser {
 }
 
 // RepoPath parses the output of the `git log` command for the `source` path.
-// The Diff chan will returns diffs in the order they are parsed from the log.
+// The Diff chan will return diffs in the order they are parsed from the log.
 func (c *Parser) RepoPath(ctx context.Context, source string, head string, abbreviatedLog bool, excludedGlobs []string, isBare bool) (chan *Diff, error) {
 	args := []string{"-C", source, "log", "-p", "--full-history", "--date=format:%a %b %d %H:%M:%S %Y %z"}
 	if abbreviatedLog {
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -398,7 +398,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
 			}
 			// Create a new currentDiff and currentCommit
 			currentDiff = diff()
-			currentCommit = &Commit{}
+			currentCommit = &Commit{Message: strings.Builder{}}
 			currentDiff.Commit = currentCommit
 			// Check that the commit line contains a hash and set it.
 			if len(line) >= 47 {
@@ -408,7 +408,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
 			latestState = MergeLine
 		case isAuthorLine(isStaged, latestState, line):
 			latestState = AuthorLine
-			currentCommit.Author = strings.TrimSpace(string(line[8:]))
+			currentCommit.Author = strings.TrimRight(string(line[8:]), "\n")
 
 		case isDateLine(isStaged, latestState, line):
 			latestState = DateLine
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -105,11 +105,7 @@ type Diff struct {
 	LineStart int
 	IsBinary  bool
 
-	// Commit metadata.
-	CommitHash    string
-	Author        string
-	CommitDate    time.Time
-	CommitMessage strings.Builder
+	Commit *Commit
 
 	contentWriter contentWriter
 }
@@ -153,8 +149,11 @@ func (d *Diff) finalize() error { return d.contentWriter.CloseForWriting() }
 
 // Commit contains commit header info and diffs.
 type Commit struct {
-	Hash string
-	Size int // in bytes
+	Hash    string
+	Author  string
+	Message strings.Builder
+	Date    time.Time
+	Size    int // in bytes
 }
 
 // Parser sets values used in GitParse.
@@ -360,6 +359,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
 		}
 	}
 	currentDiff := diff()
+	currentDiff.Commit = currentCommit
 
 	defer common.RecoverWithExit(ctx)
 	defer close(diffChan)
@@ -389,7 +389,6 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
 						"latest_state", latestState.String(),
 					)
 				}
-				currentDiff.CommitHash = currentCommit.Hash
 				diffChan <- currentDiff
 				currentCommit.Size += currentDiff.Len()
 			}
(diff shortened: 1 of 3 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -147,61 +153,8 @@ func (d *Diff) finalize() error { return d.contentWriter.CloseForWriting() }
...
-	Hash    string
-	Author  string
-	Date    time.Time
-	Message strings.Builder
-	Diffs   []Diff
-	Size    int // in bytes
-}
-
-// Equal compares the content of two Commits to determine if they are the same.
-func (c1 *Commit) Equal(ctx context.Context, c2 *Commit) bool {
```
ahrav, line 159: This was only used for testing so I moved it to the test file.

pkg/gitparse/gitparse_test.go
```diff
@@ -553,118 +577,18 @@ func TestLineChecks(t *testing.T) {
...
 		}
 	}
 }
 
-type messageParsingTestCase struct {
-	data     string
-	expected string
-}
-
-func TestMessageParsing(t *testing.T) {
```
ahrav, line 573: This was removed as we do not use the commit message at any point later in the pipeline. If we need it in the future we can always add it back.

rgmz: Do we scan commit messages? If not, we should.

ahrav: We do not currently, @zricethezav also mentioned there were plans to, but nothing definitive that i'm aware of.

rgmz: It's quite simple to implement with the current design, so I think it's worth considering how it could be done with this new implementation.

ahrav: Yea, I don't think it should be too much more complicated than with the prior implementation. (at least i hope not 😅 )

pkg/sources/git/git.go
```diff
@@ -542,87 +541,107 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
-			fileName := diff.PathB
-			if fileName == "" {
-				continue
+		fullHash := diff.CommitHash
+		if !strings.EqualFold(fullHash, lastCommitHash) {
+			depth++
+			lastCommitHash = fullHash
+			atomic.AddUint64(&s.metrics.commitsScanned, 1)
+		}
+		logger.V(5).Info("scanning commit", "commit", fullHash)
```
bill-rich, line 558: I think this should only be logged when moving onto a new commit.

pkg/gitparse/gitparse.go
```diff
@@ -105,6 +105,12 @@ type Diff struct {
 	LineStart int
 	IsBinary  bool
 
+	// Commit metadata.
+	CommitHash    string
```
bill-rich, line 109: What about keeping these in `Commit` and just having a pointer in `Diff`? That way we don't need to copy all this data for each `Diff`.

ahrav: I defs thought about that, and now that you bring it up, I think i'll go that route.

pkg/sources/git/git.go
ahrav: Good call, yep!

pkg/gitparse/gitparse.go
ahrav: Well actually now that i'm looking at it again, we don't even need the author or date in the commit and longer. We only need the hash to stay in the commit. Is it worth adding a ptr to the commit for just the hash?

pkg/sources/git/git.go
```diff
@@ -542,87 +541,107 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
-			if !scanOptions.Filter.Pass(diff.PathB) {
-				continue
-			}
 
-			fileName := diff.PathB
-			if fileName == "" {
-				continue
+		fullHash := diff.CommitHash
+		if !strings.EqualFold(fullHash, lastCommitHash) {
+			depth++
```
mcastorina, line 554: This relies on `diffChan` to send diffs in order.

I think that is the case currently, but if we ever add concurrency it won't hold. Maybe we can add a doc comment somewhere (`RepoPath` ?) that notes this behavior.

pkg/sources/git/git.go
```diff
@@ -542,87 +541,107 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
+			lastCommitHash = fullHash
+			atomic.AddUint64(&s.metrics.commitsScanned, 1)
```
mcastorina, line 556: Is `s.metrics.commitsScanned` shared with other goroutines?

ahrav: Yea i think it is... i'm not entirely sure so I didn't change the exiting behavior jut to be safe.

pkg/sources/git/git.go
ahrav: Yea will do.

mcastorina (changes requested):
I did a quick smoke test with your changes, and the most immediate (and hopefully only) regression I saw was the timestamps were the null time value in the found results:

```
» ./trufflehog git --no-update https://github.com/trufflesecurity/test_keys.git
```

```
...
Commit: 77b2a3e56973785a52ba4ae4b8dac61d4bac016f
File: keys
Line: 6
Repository: https://github.com/trufflesecurity/test_keys.git
Timestamp: 0001-01-01 00:00:00 +0000
           ^^^^^^^^^^^^^^^^^^^^^^^^^
```

pkg/gitparse/gitparse.go
ahrav: nvm, using the commit is a good idea after all. 🤣

ahrav:
> I did a quick smoke test with your changes, and the most immediate (and hopefully only) regression I saw was the timestamps were the null time value in the found results:
> 
> ```
> » ./trufflehog git --no-update https://github.com/trufflesecurity/test_keys.git
> ```
> 
> ```
> ...
> Commit: 77b2a3e56973785a52ba4ae4b8dac61d4bac016f
> File: keys
> Line: 6
> Repository: https://github.com/trufflesecurity/test_keys.git
> Timestamp: 0001-01-01 00:00:00 +0000
>            ^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

should be fixed now, feel free to try again.
Screenshot 2024-02-05 at 8 52 02 PM

mcastorina (approved):
Nothing really jumps out at me as incorrect, and the smoke test output matches what we have in main.

I'm approving, but I would feel better about merging if you got another approval from someone more experienced with this package.

ahrav:
> Nothing really jumps out at me as incorrect, and the smoke test output matches what we have in main.
> 
> 
> 
> I'm approving, but I would feel better about merging if you got another approval from someone more experienced with this package.

Thanks Miccah, I'm gonna let it get some more eyes tomorrow. I also have a couple more tests I'd like to work through just to be safe.

pkg/gitparse/gitparse.go
```diff
@@ -485,6 +434,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
 			// This should never be nil, but check in case the stdin stream is messed up.
 			if currentCommit == nil {
 				currentCommit = &Commit{}
+				currentDiff.Commit = currentCommit
```
zricethezav, line 437: Any reason we can't do this? (suggestion, feel free to ignore)
```
if currentCommit == nil {
   currentDiff.Commit = &Commit{}
 }
```

pkg/gitparse/gitparse.go
```diff
@@ -496,36 +446,16 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, commitChan ch
...
-						Diffs:   []Diff{},
-					}
-					// Message needs to be recreated here otherwise writing to it again will result in a panic.
-					currentCommit.Message.WriteString(oldCommit.Message.String())
-				}
+				diffChan <- currentDiff
 			}
 			currentDiff = diff()
-			// currentDiff = NewDiff(withCustomContentWriter(c.contentWriter()))
-		case isModeLine(isStaged, latestState, line):
```
zricethezav, line 522: Are there any concerns with dropping all these `isStaged`? Don't we use that for pre-commit scanning?

pkg/gitparse/gitparse_test.go
zricethezav: > It's quite simple to implement with the current design, so I think it's worth considering how it could be done with this new implementation.

Sounds like another issue and implementation PR. We should wait before this PR is merged before considering it.

pkg/gitparse/gitparse.go
ahrav: The only reason I dropped them was because my linter was yelling at me that they were unused. I couldn't find any references to them, but i'm happy to add them back in case I overlooked something. I probably should've made this change in a separate PR, but I figured I was already cleaning this file up that i'd tack it on 😓

rosecodym (approved):
This looks really nice, man! I also don't feel like enough of an expert on the package specifically to give my approval much weight but the code changes themselves look really straightforward.

pkg/gitparse/gitparse.go
zricethezav: I'd trust the linter, s'all good 👍🏻

zricethezav (approved):

bill-rich (approved):
Looks good! Excited to see these changes in action.

pkg/gitparse/gitparse.go
```diff
@@ -349,21 +305,21 @@ func (c *Parser) Staged(ctx context.Context, source string) (chan Commit, error)
 }
 
 // executeCommand runs an exec.Cmd, reads stdout and stderr, and waits for the Cmd to complete.
-func (c *Parser) executeCommand(ctx context.Context, cmd *exec.Cmd, isStaged bool) (chan Commit, error) {
-	commitChan := make(chan Commit, 64)
+func (c *Parser) executeCommand(ctx context.Context, cmd *exec.Cmd, isStaged bool) (chan *Diff, error) {
+	diffChan := make(chan *Diff, 64)
```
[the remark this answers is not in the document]
ahrav: I'm actually not sure... I think @bill-rich might have some insight into that value being chosen.

pkg/gitparse/gitparse.go
```diff
@@ -883,25 +813,26 @@ func isHunkEmptyLine(isStaged bool, latestState ParseState, line []byte) bool {
...
 }
 
-func cleanupParse(ctx context.Context, currentCommit *Commit, currentDiff *Diff, commitChan chan Commit, totalLogSize *int) {
+func cleanupParse(ctx context.Context, currentCommit *Commit, currentDiff *Diff, diffChan chan *Diff, totalLogSize *int) {
 	if err := currentDiff.finalize(); err != nil {
 		ctx.Logger().Error(err, "failed to finalize diff")
 		return
 	}
+
 	// Ignore empty or binary diffs (this condition may be redundant).
```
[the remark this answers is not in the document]
ahrav: You ask good questions.... of which I don't know the answer too. @bill-rich could i bug you again for this one. 🙇

PR #2395, opened 2024-02-07
[feat] - buffered file writer metrics

ahrav:
### Description:
Include some metrics for the buffered file writer.

Screenshot 2024-02-07 at 4 34 55 AM

### Checklist:

### DIFF
--- a/pkg/writers/buffered_file_writer/bufferedfilewriter.go
+++ b/pkg/writers/buffered_file_writer/bufferedfilewriter.go
@@ -109,11 +109,11 @@ func (bp *bufferPool) put(buf *buffer) {
 	const maxAllowedCapacity = 2 * defaultBufferSize
 	if buf.Cap() > maxAllowedCapacity {
 		bp.metrics.recordShrink(buf.Cap() - defaultBufferSize)
-		buf = nil // Release the large buffer for garbage collection.
-	} else {
-		// Reset the buffer to clear any existing data.
-		buf.Reset()
+		// Release the large buffer for garbage collection by not returning it to the pool.
+		return
 	}
+	// Reset the buffer to clear any existing data.
+	buf.Reset()
 
 	bp.Put(buf)
 }
--- a/pkg/writers/buffered_file_writer/bufferedfilewriter.go
+++ b/pkg/writers/buffered_file_writer/bufferedfilewriter.go
@@ -16,28 +16,28 @@ import (
 
 type bufferPoolMetrics struct{}
 
-func (b *bufferPoolMetrics) recordGrowth(growthAmount int) {
+func (bufferPoolMetrics) recordGrowth(growthAmount int) {
 	growCount.Inc()
 	growAmount.Add(float64(growthAmount))
 }
 
-func (b *bufferPoolMetrics) recordShrink(amount int) {
+func (bufferPoolMetrics) recordShrink(amount int) {
 	shrinkCount.Inc()
 	shrinkAmount.Add(float64(amount))
 }
 
-func (b *bufferPoolMetrics) recordCheckoutDuration(duration time.Duration) {
+func (bufferPoolMetrics) recordCheckoutDuration(duration time.Duration) {
 	checkoutDuration.Observe(float64(duration.Microseconds()))
 	checkoutCount.Inc()
 	checkoutDurationTotal.Add(float64(duration.Microseconds()))
 }
 
-func (b *bufferPoolMetrics) recordBufferRetrival() {
+func (bufferPoolMetrics) recordBufferRetrival() {
 	activeBufferCount.Inc()
 	bufferCount.Inc()
 }
 
-func (b *bufferPoolMetrics) recordBufferReturn(bufCap, bufLen int64) {
+func (bufferPoolMetrics) recordBufferReturn(bufCap, bufLen int64) {
 	activeBufferCount.Dec()
 	totalBufferSize.Add(float64(bufCap))
 	totalBufferLength.Add(float64(bufLen))
@@ -49,12 +49,12 @@ type bufferPool struct {
 	bufferSize uint32
 	*sync.Pool
 
-	metrics *bufferPoolMetrics
+	metrics bufferPoolMetrics
 }
 
-const defaultBufferSize = 2 << 12 // 4KB
+const defaultBufferSize = 1 << 12 // 4KB
 func newBufferPool(opts ...bufPoolOpt) *bufferPool {
-	pool := &bufferPool{bufferSize: defaultBufferSize, metrics: new(bufferPoolMetrics)}
+	pool := &bufferPool{bufferSize: defaultBufferSize}
 
 	for _, opt := range opts {
 		opt(pool)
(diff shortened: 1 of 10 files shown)


pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -8,29 +8,61 @@ import (
...
+	shrinkAmount.Add(float64(amount))
+}
+
+func (b *bufferPoolMetrics) recordCheckoutDuration(duration time.Duration) {
+	checkoutDuration.Observe(float64(duration.Microseconds()))
+	checkoutCount.Inc()
+	checkoutDurationTotal.Add(float64(duration.Microseconds()))
+}
+
+// recordBufferRetrival groups the metrics updates for when a buffer is fetched.
```
rosecodym, line 35: Why does this one get a comment but not the others?

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -187,6 +261,8 @@ func (w *BufferedFileWriter) Write(ctx context.Context, data []byte) (int, error
 				"available_space", availableSpace,
 				"grow_size", growSize,
 			)
+			// We are manually growing the buffer so we can track the growth via metrics.
+			w.bufPool.growBufferWithSize(w.buf, growSize)
```
rosecodym, line 265: Ah, now I get to re-make my comment about amortized write times: I would like a comment explaining why we're not worried about asymptotic write time complexity.

rosecodym (commented):
How did you pick these metrics? And did you have any Prometheus issues with all of them?

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Yea i definitely thought about that. Our manual buffer growth strategy is tailored to our application's predictable data sizes. By understanding ahead of time the exact memory needed for each operation, we can resize buffers efficiently in a single step. This diverges optimally from defaults in cases where incoming data significantly exceeds current capacity. Knowing the exact data size, we directly resize to fit it, rather than exponential growth which may require multiple allocations and copies if the size required is much larger than double the capacity.

This minimizes memory wastage and ensures we allocate just enough to hold the data. Our approach aligns with default behavior when growth sizes happen to match current capacity, retaining asymptotic efficiency benefits (at least i'm pretty sure it does). But our foresight into precise data sizes allows further optimization beyond defaults in many cases. I feel this approach strikes a balance between minimizing reallocations and avoiding under/over provisioning.

I tried to put together a small [example](https://go.dev/play/p/FqQSQFPRG71).

Let me know what you think.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: it's special. I'll add them across, i probably just got lazy 😞

ahrav:
> How did you pick these metrics? And did you have any Prometheus issues with all of them?

All these metrics started organically from print statements and curiosity as I explored the optimal buffer size. The key ones jumped out clearly at first - growths, shrinks, pool utilization. Then as subtle issues persisted, my debugging paranoia had me probing duration times, unveiling architectural culprits. My tinkering (cough cough, probably wasting more time than I should hve) snowballed metrics until a pretty clear picture emerged - when data flows in, buffer workings, writer throughput as well.

This then led me down a rabbit hole of converting our commit chan to a diff chan. But each metric tied intimately to the last, my NEED to understand wholly how data moved in the system. They all initially may have started as troubleshooting tools, but now form a pretty nice picture where once was mystery 🤣 . All from healthy (or maybe a bit unhealthy) curiosity...and a solid amount performance anxiety! 😅 

TLDR: I pondered, then tracked what puzzled me. The metrics arise intrinsically, their stories intertwined with my own problem-solving journey.

Sorry, I had to make this response a bit of an adventure.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: err actually they are pretty obvious i'll remove the comment as it seems redundant here. Happy to add them if you feel it adds value.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -8,29 +8,60 @@ import (
...
+	totalBufferLength.Add(float64(bufLen))
+}
+
 type bufPoolOpt func(pool *bufferPool)
 
 type bufferPool struct {
 	bufferSize uint32
 	*sync.Pool
+
+	metrics *bufferPoolMetrics
```
mcastorina, line 52: Why is this a pointer? It seems like `bufferPoolMetrics` is used to group together some methods to call. I think switching to a non-pointer would save some bytes `#micro-optimizations`

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -8,29 +8,60 @@ import (
...
 }
 
-const defaultBufferSize = 2 << 10 // 2KB
+const defaultBufferSize = 2 << 12 // 4KB
```
mcastorina, line 55: **Nit:** `2 << 12` is 8KB. `1 << 12` is 4KB. I'd recommend using a 1 since it's easy to get confused between shifting and `2**12`.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -87,9 +153,11 @@ type BufferedFileWriter struct {
 	state state // Current state of the writer. (writeOnly or readOnly)
 
 	bufPool  *bufferPool    // Pool for storing buffers for reuse.
-	buf      *bytes.Buffer  // Buffer for storing data under the threshold in memory.
+	buf      *buffer        // Buffer for storing data under the threshold in memory.
 	filename string         // Name of the temporary file.
 	file     io.WriteCloser // File for storing data over the threshold.
+
+	metrics *bufferedFileWriterMetrics
```
mcastorina, line 160: Same here about pointers.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: ptr's are more fun 🤣

mcastorina: If by "more fun" you mean "can cause panics" then I agree 🤣

mcastorina (approved):
doop

pkg/writers/buffered_file_writer/bufferedfilewriter.go
rosecodym: Oh jeez I'm sorry, I meant in the _code_. You've already convinced me, but I want to guard against:
* Some new grad hotshot comes in proud of paying attention in their algorithms class, sees an allocation strategy that's "wrong," and tries to "fix" it
* Something about the code changes such that we _don't_ know what we need to allocate ahead of time and nobody notices that the land mine got armed

rosecodym (approved):

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Ohhhh whoops, yea that's a good idea. I'll add a comment.

PR #2402, opened 2024-02-08
2396 since commit stopped working

### LINKED ISSUE #2396: --since-commit stopped working in v3.67.4
### TruffleHog Version
3.67.4

### Trace Output
N/A

### Expected Behavior

Only scan branch commits since base.

### Actual Behavior

Scan continues across all history.

### Steps to Reproduce

Run the following on a branch of the repo which has some pre-existing secrets in history:
```
docker run --rm -v "$PWD:/scan" ghcr.io/trufflesecurity/trufflehog:3.67.4 git file:///scan --since-commit master --branch some_branch_name --fail --no-update --github-actions --only-verified
```

## Environment
N/A

## Additional Context
N/A

### References

It seems that some change was introduced in https://github.com/trufflesecurity/trufflehog/pull/2387 where it never reaches base commit and just proceeds scanning entire history.

* #0000

ahrav:
### Description:
We should send an empty diff for commits that have no diff data. We can construct an empty diff and only set the commit.

### Checklist:


pkg/sources/git/git.go
```diff
@@ -522,7 +522,26 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 		repoCtx = context.WithValue(ctx, "repo", path)
 	}
 
-	diffChan, err := s.parser.RepoPath(repoCtx, path, scanOptions.HeadHash, scanOptions.BaseHash == "", scanOptions.ExcludeGlobs, scanOptions.Bare)
+	logger := repoCtx.Logger()
+	var (
+		logValues []any
+
+		baseHashSet bool
```
ahrav, line 529: Added these so we aren't doing string compares for every single diff.

pkg/gitparse/gitparse_test.go
```diff
@@ -652,46 +654,12 @@ func TestCommitParsing(t *testing.T) {
...
 
 		if !diff.Equal(context.Background(), expected[i]) {
-			t.Errorf("Diff does not match.\nexpected: %+v\n%s\nactual  : %+v\n%s", expected[i], expected[i].Commit.Message.String(), diff, diff.Commit.Message.String())
+			t.Errorf("Diff does not match.\nexpected: %+v\n%s\nactual  : %+v\n%s", expected[i], expected[i].Commit.Hash, diff, diff.Commit.Hash)
 		}
 		i++
 	}
 }
 
-func TestIndividualCommitParsing(t *testing.T) {
```
ahrav, line 661: Removed since this should be handled by TestCommitParsing.

pkg/sources/git/git.go
rosecodym: This looks like it only saves a comparison when the base hash isn't set - and in that case, isn't the comparison constant-time (and fast) anyway?

pkg/sources/git/git.go
```diff
@@ -522,7 +522,26 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
+		maxDepthSet bool
```
rosecodym, line 530: How much time does this save?

pkg/sources/git/git.go
```diff
@@ -532,37 +551,28 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
 		}
 
 		fullHash := diff.Commit.Hash
-		if !strings.EqualFold(fullHash, lastCommitHash) {
+		if baseHashSet && scanOptions.BaseHash == fullHash {
+			logger.V(1).Info("reached base commit", "commit", fullHash)
+			break
+		}
+
+		if fullHash != lastCommitHash {
```
rosecodym, line 570: Why aren't we worried about case sensitivity anymore?

pkg/gitparse/gitparse.go
```diff
@@ -147,8 +147,15 @@ func (d *Diff) write(ctx context.Context, p []byte) error {
...
 // finalize ensures proper closure of resources associated with the Diff.
 // handle the final flush in the finalize method, in case there's data remaining in the buffer.
+// Also set the hasDiffs flag on the commit to indicate that the commit has diffs.
+// This is needed to handle the case where a commit has no diffs, which would otherwise be missed.
 // This method should be called to release resources, especially when writing to a file.
-func (d *Diff) finalize() error { return d.contentWriter.CloseForWriting() }
+func (d *Diff) finalize() error {
+	if d.Commit != nil {
+		d.Commit.hasDiffs = true
+	}
```
rosecodym, line 156: Why are you doing this here, instead of when the diff is created? (Maybe in `newDiff`, or even better, just at the `diff` call site.)

pkg/gitparse/gitparse.go
```diff
@@ -157,6 +164,8 @@ type Commit struct {
 	Date    time.Time
 	Message strings.Builder
 	Size    int // in bytes
+
+	hasDiffs bool
```
rosecodym, line 168: It doesn't look like this _needs_ to be a struct field - it looks like it's got a lifetime that's compatible with just being a `FromReader` local variable. Did you consider tracking it there instead of here? I can see arguments both ways.

pkg/gitparse/gitparse.go
```diff
@@ -363,6 +372,12 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
 	}
 	currentDiff := diff(currentCommit)
 
+	createEmptyDiffForCommit := func(commit *Commit) *Diff {
+		// Initialize an empty Diff instance associated with the given commit.
+		// Since this diff represents "no changes", we only need to set the commit.
+		return &Diff{Commit: commit}
+	}
+
```
rosecodym, line 380: I'm usually all for functional hooliganism but this seems like overkill even to me. It's a one-line function used in one place - why not just inline it?

rosecodym (commented):
So was this bug triggered when the base commit doesn't have any diffs in it?

ahrav:
> So was this bug triggered when the base commit doesn't have any diffs in it?

Exactly and is the root cause of GHA failing for many users. We were essentially doing a full historical scan instead of just scanning the diff.

pkg/sources/git/git.go
ahrav: This was how it was initially implemented, I had changed it to a case insensitive version. I'm reverting it back to the original approach.

pkg/sources/git/git.go
ahrav: Not sure but found this a bit easier to read, also felt weird to do the same comparison (regardless of how fast) over and over again only to get the same result.

pkg/sources/git/git.go
ahrav: It's also used in the call to `RepoPath`, and similar to below, it felt wrong to do the same comparison for every diff, when we only need to do it once.

pkg/gitparse/gitparse.go
ahrav: I felt encapsulating it within the `Commit` struct made it easier to reason about. The `FromReader` method as it is has a lot going on and adding a another variable was less easier to reason about. Happy to make it a local var if you feel strongly about it 😄

mcastorina:
> > So was this bug triggered when the base commit doesn't have any diffs in it?
> 
> Exactly and is the root cause of GHA failing for many users. We were essentially doing a full historical scan instead of just scanning the diff.

Once this is fixed for users, maybe we should re-evaluate how we handle the options to git scans. It is non-obvious that we must send an empty `Diff` for a commit so that our control flow works.

rosecodym:
> > So was this bug triggered when the base commit doesn't have any diffs in it?
> 
> Exactly and is the root cause of GHA failing for many users. We were essentially doing a full historical scan instead of just scanning the diff.

"Many" users? I figured that diff-less commits would be pretty rare. Is there some cause of them I'm not thinking of?

rosecodym (approved):

pkg/gitparse/gitparse.go
```diff
@@ -393,11 +397,18 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
...
 				diffChan <- currentDiff
 				currentCommit.Size += currentDiff.Len()
+				currentCommit.hasDiffs = true
 			}
 			// If there is a currentCommit, send it to the channel.
 			if currentCommit != nil {
 				totalLogSize += currentCommit.Size
+				if !currentCommit.hasDiffs {
+					// Initialize an empty Diff instance associated with the given commit.
+					// Since this diff represents "no changes", we only need to set the commit.
```
rosecodym, line 407: This is a good spot to explain _why_ we're doing this.

PR #2407, opened 2024-02-09
Regex boundary clean and various tweaks to reduce FPs (part 1)

zricethezav:
### Description:
Adds better regex boundaries and includes various tweaks to reduce FPs

### Checklist:

### DIFF
--- a/pkg/detectors/aiven/aiven.go
+++ b/pkg/detectors/aiven/aiven.go
@@ -22,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aiven"}) + `\b([a-zA-Z0-9/+=]{372})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aiven"}) + `?\b([a-zA-Z0-9/+=]{350,380})`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/aiven/aiven.go
+++ b/pkg/detectors/aiven/aiven.go
@@ -3,9 +3,10 @@ package aiven
 import (
 	"context"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
@@ -21,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aiven"}) + `([a-zA-Z0-9/+=]{372})`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aiven"}) + `\b([a-zA-Z0-9/+=]{372})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/appointedd/appointedd.go
+++ b/pkg/detectors/appointedd/appointedd.go
@@ -2,10 +2,11 @@ package appointedd
 
 import (
 	"context"
-	regexp "github.com/wasilibs/go-re2"
 	"io"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
@@ -21,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{88})`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{88})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/appointedd/appointedd.go
+++ b/pkg/detectors/appointedd/appointedd.go
@@ -2,6 +2,7 @@ package appointedd
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
@@ -22,7 +23,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{87}=)`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{88)`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
@@ -34,6 +35,8 @@ func (s Scanner) Keywords() []string {
 // FromData will find and optionally verify appointedd secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
+	fmt.Println(keyPat.String())
+	fmt.Println("data", dataStr)
 
 	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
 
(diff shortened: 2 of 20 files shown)


pkg/detectors/appointedd/appointedd.go
```diff
@@ -21,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{88})`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{88})\b`)
```
rgmz, line 25: `\b` doesn't work unless the character range is `\w` (`[a-z0-9_`). So if this, or any other pattern, ends in `=` it won't match.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -20,8 +21,8 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	appPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
-	apiPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	appPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd_", "dd-"}) + `\b([a-zA-Z-0-9]{40})\b`)
```
zricethezav, line 24: For the reviewer: I don't feel super strongly on this change. However it does seem like `dd` alone could introduce a fair number of FPs. Open for discussion!

pkg/detectors/dovico/dovico.go
```diff
@@ -21,8 +22,8 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"dovico"}) + `\b([0-9a-z]{32}\.[0-9a-z]{1,}\b)`)
-	userPat = regexp.MustCompile(detectors.PrefixRegex([]string{"dovico"}) + `\b([0-9a-z]{32}\.[0-9a-z]{1,}\b)`)
+	keyPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"dovico"}) + `\b([0-9a-z]{32}\.[0-9a-z]{1,100})\b`)
```
zricethezav, line 25: for the reviewer: _probably_ not gonna go past 100 alpanums on the backend here but 🤷🏻 I'm open for discussion

pkg/detectors/gengo/gengo.go
```diff
@@ -29,8 +30,8 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	// Removed bounds since there are some cases where the start and end of the token is a special character
-	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"gengo"}) + `([ ]{0,1}[0-9a-zA-Z\[\]\-\(\)\{\}|_^@$=~]{64}[ \r\n]{1})`)
-	secretPat = regexp.MustCompile(detectors.PrefixRegex([]string{"gengo"}) + `([ ]{0,1}[0-9a-zA-Z\[\]\-\(\)\{\}|_^@$=~]{64}[ \r\n]{1})`)
+	keyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"gengo"}) + `([0-9a-zA-Z\[\]\-\(\)\{\}|_^@$=~]{64})\b`)
```
zricethezav, line 33: For the reviewer: this token has a real nasty pattern to match. No `\b` in front of the capture group due to the fact that some of the token's alphabet _could_ be escaped by that in-front-of-capture-group `\b`. `[ ]{0,1}` removed as it didn't provide any value according to my tests (though please double check this work if you can). 

FWIW the resulting regex is: `(?i)(?:gengo)(?:.|[\n\r]){0,40}([0-9a-zA-Z\[\]\-\(\)\{\}|_^@$=~]{64})\b`

We should probably have a discussion on the fact that the keyPat and secretPat are the same regex. I'm sure there is a way to tighten something up here.

pkg/detectors/appointedd/appointedd.go
zricethezav: There are quite a few regexes like this. I'm wondering if we should use a more strict regex like the gitleaks prefix:
```
(?i)(?:provider)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)(?:'|\"|\s|=|\x60){0,5}
```

pkg/detectors/aiven/aiven.go
```diff
@@ -21,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aiven"}) + `([a-zA-Z0-9/+=]{372})`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"aiven"}) + `\b([a-zA-Z0-9/+=]{372})\b`)
```
zricethezav, line 25: The previous regex fails detecting our test secret as does the new one. This one needs more work still

pkg/detectors/gengo/gengo.go
zricethezav: Same issue as @rgmz pointed out in https://github.com/trufflesecurity/trufflehog/pull/2407#discussion_r1484813476

pkg/detectors/appointedd/appointedd.go
rgmz: I think we'd need more data to determine if that'd be worthwhile. 

Another option would be to do something like this, which IIRC is what GitHub does.
```
[a-zA-Z0-9=+]{88}(?:[^a-zA-Z0-9=+]|\z)
```

pkg/detectors/appointedd/appointedd.go
```diff
@@ -21,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{88})`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"appointedd"}) + `\b([a-zA-Z0-9=+]{87}=)`)
```
dustin-decker, line 25: Might not always have base64 padding?

zricethezav: I think it always is base64 padded. I just created a new one and it's b64. The test secret we have on deck is also b64. Was there another case you were think of?

pkg/detectors/appointedd/appointedd.go
zricethezav: Assuming appointed is always b64 encoded, which I think it is, the boundary could just be `=` as it's proposed in the current change

https://regex101.com/r/YJeAfP/1

pkg/detectors/aiven/aiven.go
bugbaba: Just update the number of chars to 372 to 380, Also boundary won't work for this as it contains special chars within the match/key.

zricethezav: > Just update the number of chars from 372 to 380

I'm not following how this would solve the problem.

zricethezav: Generated a couple tokens just now and they all have `377` characters

zricethezav: alas, the problem still persists https://regex101.com/r/nMLU8f/1. Possible solution: https://github.com/trufflesecurity/trufflehog/pull/2407#discussion_r1484822724

rgmz: This is dumb, but try adding a `?` before the `\b`. That works in your linked example at least.

bugbaba: Super weird last month I had tried multiple times it was generating  keys of `380` length each time. Will try again.

bugbaba: Just confirmed it, Its still generating keys of `380` length for me. 
Just to be sure you are creating tokens on https://console.aiven.io/profile/tokens page right?

https://regex101.com/r/MSXILB/1

zricethezav: >  Just to be sure you are creating tokens on https://console.aiven.io/profile/tokens page right?

Yep. Try creating an account with another email and generating some tokens

bugbaba: Super Weird, Tried with new account and now its generating token of `372` length

rgmz: It's probably encoding account-specific data, which may explain the variance.

zricethezav: > This is dumb, but try adding a ? before the \b. That works in your linked example at least.

Lazy loading does the trick. Good eye @rgmz

pkg/detectors/appointedd/appointedd.go
zricethezav: reverting back to the original

zricethezav:
Going to close this PR and start fresh. After doing some testing some of these `\b`s are misplaced and have negative outcomes

PR #2409, opened 2024-02-11
[cleanup] - Extract buffer logic

ahrav:
### Description:
Ripped out most of the buffer logic and metrics from the buffered file writer into it's own package. This allows us to use the default buffer writer for diffs while still recording metrics. Most of the code should be fairly unchanged.

### Checklist:


mcastorina (commented):
It looks like `gitparse_test.go` is failing due to timeout (maybe it's hanging somewhere?).

Also, great idea putting these things in their own packages, but I'm a bit confused between the `buffer` and the `buffer_writer` packages. What do you think about adding some package level docstrings explaining their purposes?

ahrav:
> It looks like `gitparse_test.go` is failing due to timeout (maybe it's hanging somewhere?).
> 
> Also, great idea putting these things in their own packages, but I'm a bit confused between the `buffer` and the `buffer_writer` packages. What do you think about adding some package level docstrings explaining their purposes?

buffer_writer is an implementation (default) for the contentWriter interface. By default we use that unless the feature flag is enabled in which case we use the buffered file writers. Both the buffered file writer and buffer writer use the custom buffers package.

mcastorina (approved):

PR #2413, opened 2024-02-11
Ignore Kubernetes GCP test credentials

rgmz:
### Description:
The `image-pulling@authenticated-image-pulling.iam.gserviceaccount.com` service account is intentionally public. It is used by Kubernetes to test image pulling, and is found in thousands of forks or k8s-related projects.

### Checklist:

### DIFF
--- a/pkg/detectors/gcp/gcp.go
+++ b/pkg/detectors/gcp/gcp.go
@@ -1,14 +1,17 @@
 package gcp
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


ahrav (approved):

dustin-decker:
I don't love this change. It's a real, live credential that demonstrates TruffleHog's base64 decoding capabilities.

rgmz:
It might be a good test case; for end users, it's noise because there's no impact and there's nothing they can do to 'fix' it.

PR #2435, opened 2024-02-12
Update custom detector example

zricethezav:
### Description:
Updates the example used for custom detectors so you can copy/paste it as a reference. Prior to this change you would get an 'invalid escape character' warning. Also changes the len of hogID to 17 so that it matches the example secret used

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,8 +473,8 @@ detectors:
     keywords:
       - hog
     regex:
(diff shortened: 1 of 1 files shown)


mcastorina (approved):
Thanks for catching / fixing this

ahrav (approved):

PR #2450, opened 2024-02-13
Add Display method to SourceUnit and Kind member to the CommonSourceUnit

mcastorina:
### Description:
This elaborates on the `SourceUnit` interface methods to be more precise with the expectations of the API.

### Checklist:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -270,11 +270,7 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 
 // ChunkUnit implements SourceUnitChunker interface.
 func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
-	commonUnit, err := sources.IntoCommonUnit(unit)
-	if err != nil {
-		return err
-	}
-	path := commonUnit.ID
+	path, _ := unit.SourceUnitID()
 	logger := ctx.Logger().WithValues("path", path)
 
 	cleanPath := filepath.Clean(path)
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -270,7 +270,11 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 
 // ChunkUnit implements SourceUnitChunker interface.
 func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
-	path := unit.SourceUnitID()
+	commonUnit, err := sources.IntoCommonUnit(unit)
+	if err != nil {
+		return err
+	}
+	path := commonUnit.ID
 	logger := ctx.Logger().WithValues("path", path)
 
 	cleanPath := filepath.Clean(path)
--- a/pkg/sources/git/unit.go
+++ b/pkg/sources/git/unit.go
@@ -3,6 +3,9 @@ package git
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"path/filepath"
+	"strings"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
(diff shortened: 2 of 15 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -270,7 +270,11 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 
 // ChunkUnit implements SourceUnitChunker interface.
 func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
-	path := unit.SourceUnitID()
+	commonUnit, err := sources.IntoCommonUnit(unit)
+	if err != nil {
+		return err
+	}
+	path := commonUnit.ID
```
rosecodym, line 277: Having to do this whole song and dance in every source that doesn't implement its own unit type just to get an `ID` field out of the unit surprises me, given the existence of a `SourceUnitID()` method on the interface. It looks like the friction is caused by `SourceUnitID()` and `CommonSourceUnit.ID` being two different "source unit IDs" that serve two different purposes and are consumed by two different parts of the code. Is _that_ something that it makes sense to change? Perhaps we could rename one of them in order to sensibly provide an "actual" ID method on the interface, and thus avoid these type assertions in most cases...but now I'm wondering why we need a `Kind` on `CommonSourceUnit` at all. There's only ever one kind of common source unit (the, uh, common one). So what's the field for?

pkg/sources/git/unit.go
```diff
@@ -24,7 +27,33 @@ type SourceUnit struct {
...
+			// git@host:owner/repo
+			// TODO: Is this possible? We should maybe canonicalize
+			// the URL before getting here.
+			repo = path
+		}
+		return strings.TrimSuffix(repo, ".git")
+	case UnitDir:
+		return filepath.Base(u.ID)
+	default:
+		return "mysterious git unit"
```
rosecodym, line 51: This is theoretically possibly user-facing  - are you happy with the text?

pkg/sources/filesystem/filesystem.go
mcastorina: Yeah, I had a similar thought.. `ID` is no longer the `ID`, though for the common unit it probably should be?

I was intending for `Kind` to be a hint for the UI about what the unit is.

pkg/sources/git/unit.go
mcastorina: Yes, lol. Open to suggestions though :)

pkg/sources/filesystem/filesystem.go
rosecodym: Can you extend the `SourceUnit` interface with a first-class method exposing kind and stop bundling the kind into `SourceUnitID()`?

pkg/sources/filesystem/filesystem.go
```diff
@@ -270,7 +270,7 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 
 // ChunkUnit implements SourceUnitChunker interface.
 func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
-	path := unit.SourceUnitID()
+	path, _ := unit.SourceUnitID()
```
rosecodym, line 273: Does the linter complain if you leave this out?

pkg/sources/gitlab/gitlab.go
```diff
@@ -616,7 +618,7 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	// Report all repos if specified.
 	if len(repos) > 0 {
 		for _, repo := range repos {
-			unit := sources.CommonSourceUnit{ID: repo}
```
rosecodym, line 619: Oh, I never even noticed that it used to work this way. This is nice to clean up!

pkg/sources/job_progress.go
```diff
@@ -94,11 +94,11 @@ type ChunkError struct {
 }
 
 func (f ChunkError) Error() string {
-	unit := "<nil>"
+	unit, kind := "<nil>", SourceUnitKind("unit")
```
rosecodym, line 97: Would it be helpful to use a special "nil unit" kind here? Or would that be more trouble than it's worth?

rosecodym (approved):
i really like the way this ended up looking. nice work!

pkg/sources/filesystem/filesystem.go
mcastorina: Are you referring to leaving out the `_`? I think go won't compile it due to assignment mismatch.

pkg/sources/job_progress.go
mcastorina: Hm.. I don't see a strong need to add that currently, so I'm inclined to leave it as is. I'll keep that idea in the back of my head though.

PR #2465, opened 2024-02-14
lazy match prefixregex for all alphanum captures

zricethezav:
### Description:
All the `PrefixRegex` groups are greedy. Although unlikely, our greedy prefix can cause issues when two secrets with a similar regex are in close proximity to each other. 

Take this [regex101](https://regex101.com/r/vDSHxK/1/) for example. We have `glassnode` and `miro` which have the same token signature: `([0-9A-Za-z]{27})`. If a glassnode and miro secret were in close proximity to each other we would be sending the wrong secret to the wrong provider
Screenshot 2024-02-14 at 5 14 44 PM

Adding a lazy quantifier `?` to the prefixRegex fixes this problem and gives us a marginal boost in regex performance since the regex engine will be doing less unnecessary work.

Screenshot 2024-02-14 at 5 16 43 PM

Note: I've only proposed this change to regex's with alpanum captures as I know those examples work. This change should _probably_ be added to all regex's with `PrefixRegex` but I want to do a little more testing before adding that.

### Checklist:

### DIFF
--- a/pkg/detectors/abbysale/abbysale.go
+++ b/pkg/detectors/abbysale/abbysale.go
@@ -26,7 +26,7 @@ var (
 	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"abbysale"}) + `\b([a-z0-9A-Z]{40})\b`)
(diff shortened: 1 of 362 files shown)


zricethezav:
wait im an idiot, this could just be added to prefixregex

ahrav:
> wait im an idiot, this could just be added to prefixregex

I was going to suggest that lol.

PR #2470, opened 2024-02-15
Gitlab scan targets

ahrav:
### Description:
Add the ability to perform a target scan for the Gitlab source.
### Checklist:


pkg/sources/gitlab/gitlab_integration_test.go
```diff
@@ -0,0 +1,443 @@
...
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/credentialspb"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/source_metadatapb"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/sourcespb"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+func TestSource_Scan(t *testing.T) {
```
ahrav, line 21: moved from the unit test file to the integration test file.

pkg/sources/gitlab/gitlab_integration_test.go
```diff
@@ -0,0 +1,446 @@
...
+
+			assert.Equal(t, tt.wantReposScanned, len(s.repos))
+			if chunkCnt < 1 {
+				t.Errorf("0 chunks scanned.")
+			}
+		})
+	}
+}
+
+func TestSource_Validate(t *testing.T) {
```
ahrav, line 183: ditto

pkg/sources/gitlab/gitlab.go
```diff
@@ -47,6 +47,8 @@ type Source struct {
 	git                    *git.Git
 	scanOptions            *git.ScanOptions
 
+	apiClient *gitlab.Client
```
rosecodym, line 50: it looks like you've added this here so that you don't have to pass a client argument into `scanTargets` and its call tree. but now there's one path through the code that takes the client via parameter (the full-scan path - including units support) and one that uses the struct field (the targeted scan). i find this unnecessarily difficult to track mentally, and it feels like an accident waiting to happen, so i think we should settle on just doing it one way, at _least_ within a single source. (we should probably standardize _across_ sources eventually, but that's a bigger project than this pr.)

both approaches have upsides and downsides: the parameter method is easier to reason about but requires initializing the client more often; the field method means that we can theoretically initialize the client only once, but have to manage its lifetime. (which, as your other pr highlights, is a non-trivial problem!)

pkg/sources/gitlab/gitlab.go
```diff
@@ -203,6 +213,68 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	if !ok {
+		return fmt.Errorf("unable to cast metadata type for targeted scan")
+	}
+	meta := metaType.Gitlab
+	projID, sha := int(meta.GetProjectId()), meta.GetCommit()
+	if projID == 0 || sha == "" {
+		return fmt.Errorf("project ID and commit SHA must be provided for targeted scan")
+	}
+
+	logger := ctx.Logger().WithValues("project_id", projID, "commit", sha)
```
rosecodym, line 237: ❤️ this. i have a general question about the way we use this pattern though: why do we create only loggers when we do this, rather than entirely new contexts? to me, making a new context seems preferable because it's easier to propagate these logging kvps to called functions. (you didn't do that here - was it on purpose?) but i see it done so often the other way that i suspect i'm missing something.

pkg/sources/gitlab/gitlab.go
ahrav: I'll pass the client. 👍

pkg/sources/gitlab/gitlab.go
ahrav: Not sure, I just noticed we did it elsewhere so I did the same. Updated to pass a ctx instead.

pkg/sources/gitlab/gitlab.go
```diff
@@ -203,6 +210,68 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+		}
+	}
+
+	return nil
+}
+
+func (s *Source) getCommitDiffs(ctx context.Context, client *gitlab.Client, projID int, sha string) ([]*gitlab.Diff, error) {
+	const paginationLimit = 100 // Default is 20, max is 100.
+	diffOpts := &gitlab.GetCommitDiffOptions{ListOptions: gitlab.ListOptions{PerPage: paginationLimit}}
+	diffs, _, err := client.Commits.GetCommitDiff(projID, sha, diffOpts, gitlab.WithContext(ctx))
```
rosecodym, line 271: Isn't this query paginated? Do we need to look beyond the first page?

ahrav: My understanding from reading the docs suggest it is not paginated. https://docs.gitlab.com/ee/api/commits.html#get-the-diff-of-a-commit
There doesn't appear to be a way to get another page, or cursor, so I think we should be okay with this approach.
Usually the request would include the `page_token` attribute if pagination was allowed.

ahrav: probably just copied it from above. will remove

rosecodym (approved):
this looks pretty clean! what do you think of adding a logger kvp flagging this call tree as being part of a targeted scan?

PR #2471, opened 2024-02-16
add version to extra data + moving existing versioned detectors into subdirectory format

0x1:
### Description:
- we have two ways that we set up new versions of detectors. one is to create a new directory, naming it as `<detector>_v2`, other is to create subdirectories `<detector>/v1/ `. the latter follows [go convention](https://go.dev/doc/modules/version-numbers), here we update all packages that are currently not following convention
- the adding detectors docs are also being updated to reflect this
- adding version numbers to extra data which will help distinguish which detector is causing a specific detection

### Checklist:


pkg/detectors/github/v1/github_old.go
```diff
@@ -74,6 +75,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 		s1.ExtraData = map[string]string{
 			"rotation_guide": "https://howtorotate.com/docs/tutorials/github/",
+			"version":        fmt.Sprintf("%d", s.Version()),
```
mcastorina, line 78: Instead of this being a responsibility of individual detectors, what do you think about setting this generically in the engine? We could probably do the same thing for the `DetectorType` too.

hack/docs/Adding_Detectors_Internal.md
```diff
@@ -48,13 +48,18 @@ If you think that something should be included outside of these guidelines, plea
...
 In some instances, services will update their token format, requiring a new regex to properly detect secrets in addition to supporting the previous token format. Accomodating this can be done without adding a net-new detector. [We provide a `Versioner` interface](https://github.com/trufflesecurity/trufflehog/blob/e18cfd5e0af1469a9f05b8d5732bcc94c39da49c/pkg/detectors/detectors.go#L30) that can be implemented.
 
-1. Create a copy of the package and append `_v2` to the package and file names. Ex: `<packagename>/` -> `<packagename>_v2`, `<packagename>.go` -> `<packagename>_v2.go`
+1. Create two new directories `v1` and `v2`. Move the existing detector and tests into `v1`, and add new files to `v2`.
+Ex: `<packagename>/<old_files>` -> `<packagename>/v1/<old_files>`, `<packagename>/v2/<new_files>`
 
 Note: Be sure to update the tests to reference the new secret values in GSM, or the tests will fail.
 
-2. Implement the `Versioner` interface. [GitHub example implementation.](/pkg/detectors/github_old/github_old.go#L22)
+2. Implement the `Versioner` interface. [GitHub example implementation.](/pkg/detectors/github/v1/github_old.go#L22)
```
mcastorina, line 56: **Nit:** I think the link should be `#L23` now?

hack/docs/Adding_Detectors_external.md
```diff
@@ -44,13 +44,18 @@ If you think that something should be included outside of these guidelines, plea
 
 In some instances, services will update their token format, requiring a new regex to properly detect secrets in addition to supporting the previous token format. Accomodating this can be done without adding a net-new detector. [We provide a `Versioner` interface](https://github.com/trufflesecurity/trufflehog/blob/e18cfd5e0af1469a9f05b8d5732bcc94c39da49c/pkg/detectors/detectors.go#L30) that can be implemented.
 
-1. Create a copy of the package and append `_v2` to the package and file names. Ex: `<packagename>/` -> `<packagename>_v2`, `<packagename>.go` -> `<packagename>_v2.go`
+1. Create two new directories `v1` and `v2`. Move the existing detector and tests into `v1`, and add new files to `v2`.
+Ex: `<packagename>/<old_files>` -> `<packagename>/v1/<old_files>`, `<packagename>/v2/<new_files>`
 
 Note: Be sure to update the tests to reference the new secret values in GSM, or the tests will fail.
 
 2. Implement the `Versioner` interface. [GitHub example implementation.](/pkg/detectors/github_old/github_old.go#L22)
```
mcastorina, line 52: This link should to be updated like in `Adding_detectors_Internal.md`, right?

hack/docs/Adding_Detectors_external.md
```diff
@@ -78,7 +83,8 @@ Note: Be sure to update the tests to reference the new secret values in GSM, or
    2. Update the verifier code to use a non-destructive API call that can determine whether the secret is valid or not.
       * Make sure you understand [verification indeterminacy](#verification-indeterminacy).
    3. Create a [test for the detector](#testing-the-detector)
-   4. Create a pull request for review.
+   4. Add your new detector to DefaultDetectors in `/pkg/engine/defaults.go`
```
mcastorina, line 86: **Nit:** End each bullet with a period.

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -49,6 +50,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_FigmaPersonalAccessToken,
 			Raw:          []byte(resMatch),
+			ExtraData: map[string]string{
+				"version": fmt.Sprintf("%d", s.Version()),
```
mcastorina, line 54: Alternative could be [strconv.Itoa](https://pkg.go.dev/strconv#Itoa)

mcastorina (approved):
Very nice! Left a suggestion and some nits but nothing to block on.

pkg/detectors/github/v1/github_old.go
rgmz: I was going to suggest this as well. It seems much easier and less error prone to modify engine + printers.

hack/docs/Adding_Detectors_external.md
0x1: is there convention for that?

pkg/detectors/github/v1/github_old.go
```diff
@@ -1,12 +1,13 @@
-package github_old
+package github
```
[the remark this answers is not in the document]
0x1: that's a good point, i'll look into it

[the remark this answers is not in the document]
0x1: this should be doable, i'll create separate pr for it

PR #2473, opened 2024-02-16
[chore] - tighten keyword match

ahrav:
### Description:
Since the domain pattern is required to test a token, we can tighten the keyword to match the static portion of the `domainPat`.

### Checklist:


trufflesteeeve (approved):
LGTM

PR #2474, opened 2024-02-17
[chore] - use read full

ahrav:
### Description:
If the amount of data to be read is known ahead of time, it is recommended to use `io.ReadFull` as it guarantees the number of bytes to be read.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


rosecodym (approved):
i wonder if we were ever dropping any data because of this...

zricethezav (approved):

ahrav:
> i wonder if we were ever dropping any data because of this...

Very very possible. Not the biggest fan of this interface... it's not very intuitive.

PR #2491, opened 2024-02-20
Improve Gitlab default URL handling

trufflesteeeve:
### Description:
This PR does a better job handling specified gitlab URLs, particularly when they are for gitlab.com. It's currently possible wrongly specify the gitlab url, by either forgetting the protocol (https) or adding your gitlab organization to the URL. Both currently cause this source to fail, but neither should.

Additionally, if you are running against an on-prem gitlab instance, it will now fail if that instance uses http as the protocol, instead of https.

There's also a small cleanup to check for the git command earlier in the process, to prevent additional work from being done when the scan would not be successful.

Beyond those, this shouldn't change anything.

### Checklist:
      (This PR had archive test failures when running the above, looking into it)

### DIFF
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -88,6 +88,10 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
 	s.jobPool = &errgroup.Group{}
 	s.jobPool.SetLimit(concurrency)
 
+	if err := git.CmdCheck(); err != nil {
+		return err
+	}
+
 	var conn sourcespb.GitLab
 	err := anypb.UnmarshalTo(connection, &conn, proto.UnmarshalOptions{})
 	if err != nil {
@@ -115,7 +119,6 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
 		return fmt.Errorf("invalid configuration given for source %q (%s)", name, s.Type().String())
 	}
 
-	err = git.CmdCheck()
 	s.url, err = normalizeGitlabEndpoint(conn.Endpoint)
 	if err != nil {
 		return err
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -28,7 +28,9 @@ import (
 )
 
 const SourceType = sourcespb.SourceType_SOURCE_TYPE_GITLAB
-const cloudBaseURL = "https://gitlab.com/"
+
+// This is the URL for gitlab hosted at gitlab.com
+const gitlabBaseURL = "https://gitlab.com/"
 
 type Source struct {
 	name     string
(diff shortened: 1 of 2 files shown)


pkg/sources/gitlab/gitlab.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const SourceType = sourcespb.SourceType_SOURCE_TYPE_GITLAB
+const cloudBaseURL = "https://gitlab.com/"
```
zricethezav, line 31: Question: would `gitlabHostedBaseURL` be a better var name for this? `cloudBaseURL` is a bit confusing to me as `cloud` could mean a gitlab user's self-hosted gitlab instance

pkg/sources/gitlab/gitlab.go
```diff
@@ -590,6 +586,51 @@ func normalizeRepos(repos []string) ([]string, []error) {
...
+		gitlabURL, err = url.Parse("https://" + gitlabEndpoint)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return normalizeGitlabEndpointURL(gitlabURL)
+}
+
+// Do not use this, only use normalizeGitlabEndpoint.
```
rosecodym, line 613: getting your entry for the "most ignored comment twelve months from now" competition ready?

i'm guessing you just wanted the `normalizeGitlabEndpointURL` modular but not in a way that causes anyone else to consume it as such. this comment prompted a double-take in me because it looks way scarier than that - i'd vote for just inlining this function (unless i'm misunderstanding something)

pkg/sources/gitlab/gitlab.go
rosecodym: amusingly, i had the opposite reaction! but i see your point. i wonder if it's actually clearer to _not_ use a const and just inline the literal where we need it? (or maybe make the const local to the function? can you do that?)

either way, a comment is probably warranted if you can't find a magically perfectly clear identifier.

rosecodym (approved):

pkg/sources/gitlab/gitlab.go
trufflesteeeve: Yeah, inlining is definitely better than a scary comment. Also it was initially done this way to more or less allow the overall function to be recursive on the "try again but with https:// at the start" without needing to worry that we'd ever hit that case again.

rosecodym: ah ok that makes sense

pkg/sources/gitlab/gitlab.go
```diff
@@ -590,6 +588,46 @@ func normalizeRepos(repos []string) ([]string, []error) {
 	return validRepos, errs
 }
 
+// normalizeGitlabEndpoint ensures that if and endpoint is going to gitlab.com, we use https://gitlab.com/ as the endpoint.
```
rosecodym, line 591: i think there's a typo

rosecodym (approved):

zricethezav (approved):

PR #2496, opened 2024-02-21
Tell git to ignore directory ownership (fixes #2495)

### LINKED ISSUE #2495: Git complains about directory ownership in recent versions
Git [now complains](https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9) if a directory is owned by a different user. Trufflehog uses `git cat-file` for binary files, which means running inside Docker with a mounted volume typically throws an error. Changing directory ownership to root fixes it, but is not a good workaround.

### TruffleHog Version
3.67.6

### Trace Output

Here's the normal output: https://gist.github.com/marksteward/793f722133eb046e0564f74a09354676

I can include trace output if really needed, but it just obscures the error.

The git output is:

```
fatal: detected dubious ownership in repository at '/data/.git'
To add an exception for this directory, call:
        git config --global --add safe.directory /data/.git
```

### Expected Behavior

It should work.

### Actual Behavior

It throws an error.

### Steps to Reproduce

See the gist.

## Environment
WSL Ubuntu 20

marksteward:
### Description:
When running inside docker as root with a mounted volume, we need to tell git to ignore directory ownership, which will usually not match. For more details on this fix, see https://stackoverflow.com/a/73100228/37923.

There is no security risk here, because the user actually owns the directory outside docker, and a git exploit that executes code inside the container would affect a user who made the permissions match (with `chown` or by passing `-u $UID` to docker) .

I've also included a second commit that fixes `entrypoint.sh` to show usage if no arguments are passed, instead of throwing a cryptic error. Feel free to skip this commit if you think it's wrong.

Fixes #2495.

### Checklist:
* [ ] Tests passing (`make test-community`)? No, but it doesn't without my fix either.
* [x] Script passes shellcheck

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -6,4 +6,8 @@ mapfile -t extra_args < <(bash -c "for arg in ${*: -1}; do echo \$arg; done")
 # Directories might be owned by a user other than root
 git config --global --add safe.directory '*'
 
-/usr/bin/trufflehog "${@: 1: $#-1}" "${extra_args[@]}" 
+if [[ $# -eq 0 ]]; then
+  /usr/bin/trufflehog --help
+else
+  /usr/bin/trufflehog "${@: 1: $#-1}" "${extra_args[@]}"
+fi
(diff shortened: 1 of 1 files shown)


ahrav (approved):
TIL! Thanks for the addition @marksteward

PR #2516, opened 2024-02-27
JDBC test and parsing improvements

dustin-decker:
- Uses net/url for more robust URI parsing
- Supports common JDBC formats for MySQL
- Supports URI format for MSSQL
- Uses allowlist for params across all drivers
- Uses testcontainers-go for integration testing - much faster, more robust, no port collisions
- Uses gofakeit for random data (db, user, password) generation in integration tests
- Adds connection timeouts
- Use Microsoft's driver for MSSQL

### Checklist:


pkg/detectors/jdbc/postgres.go
```diff
@@ -59,41 +59,60 @@ func joinKeyValues(m map[string]string, sep string) string {
...
+		case "sslmode":
+			if isAlphabet(v[0]) {
+				params[k] = v[0]
+			}
+		case "user":
+			params[k] = v[0]
+		case "password":
+			params[k] = v[0]
+		}
 	}
```
mcastorina, line 102: Is it guaranteed that `len(v) >= 1`?

mcastorina: Also, why does `sslmode` require an alphabetic value?

mcastorina (commented):
Thanks for improving this! There's a lot of changes here so I'm going to take a closer look, but nothing big jumped out at me on the first pass.

pkg/detectors/jdbc/mysql.go
```diff
@@ -50,47 +51,54 @@ func isMySQLErrorDeterminate(err error) bool {
...
+	pass := ""
+	if u.User != nil {
+		user = u.User.Username()
+		pass, _ = u.User.Password()
 	}
-	host, database, found := strings.Cut(hostAndDB, "/")
-	if !found {
-		return nil, errors.New("expected host and database to be separated by /")
+
+	for k, v := range u.Query() {
```
bill-rich, line 87: What if user/pass aren't included in `u.User` or `q.Query`?

pkg/detectors/jdbc/postgres.go
bill-rich: +1 Should we check `sslmode` against a list of expected values?

bill-rich: Also, can't it also have a `-`? eg `verify-ca`

pkg/detectors/jdbc/mysql.go
dustin-decker: I've updated this to default to `root`, and to only provide the username if there is no password.

pkg/detectors/jdbc/postgres.go
dustin-decker: Added length check and check against expected values.

pkg/detectors/jdbc/mysql_integration_test.go
```diff
@@ -4,101 +4,107 @@
...
+	mysqlPass := gofakeit.Password(true, true, true, false, false, 10)
+	mysqlDatabase := gofakeit.Word()
+
+	ctx := context.Background()
+
+	mysqlC, err := mysql.RunContainer(ctx,
+		mysql.WithDatabase(mysqlDatabase),
+		mysql.WithUsername(mysqlUser),
+		mysql.WithPassword(mysqlPass),
+	)
```
mcastorina, line 28: Cool, I didn't know this existed!

pkg/detectors/jdbc/sqlserver.go
```diff
@@ -38,22 +37,35 @@ func parseSqlServer(subname string) (jdbc, error) {
...
+		case "server":
+			host = value
+		case "port":
+			port = value
 		}
 	}
 	return &sqlServerJDBC{
-		conn:   strings.TrimPrefix(subname, "//"),
-		params: params,
+		connStr: fmt.Sprintf("sqlserver://sa:%s@%s:%s?database=master&connection+timeout=5", password, host, port),
```
mcastorina, line 69: Does this mean we'll only ever try to connect to the `master` database for sqlserver?

mcastorina (commented):
Overall this looks good, I just had a few questions about intended behavior.

pkg/detectors/jdbc/mysql.go
```diff
@@ -50,47 +51,59 @@ func isMySQLErrorDeterminate(err error) bool {
...
+		switch k {
+		case "user":
+			user = v[0]
+		case "password":
+			pass = v[0]
+		}
```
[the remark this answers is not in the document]
dustin-decker: Thank you, I will update.

pkg/detectors/jdbc/sqlserver.go
dustin-decker: Yes, and sa for the user. This did not change. I would have improved it during this but the container did not allow them to be set. It is follow-up work.

mcastorina (approved):

PR #2518, opened 2024-02-28
Improve monogo and snowflake detectors

dustin-decker:
### Description:
- uses lighter snowflake driver
- adds timeouts to mongo detector

### Checklist:


go.mod
```diff
@@ -4,6 +4,8 @@ go 1.21
 
 replace github.com/jpillora/overseer => github.com/trufflesecurity/overseer v1.2.7
 
+replace github.com/snowflakedb/gosnowflake => github.com/trufflesecurity/gosnowflake v0.0.1
```
mcastorina, line 7: I see [trufflesecurity/gosnowflake](https://github.com/trufflesecurity/gosnowflake) is forked from `snowflakedb/gosnowflake` with a lot of code deleted. What is the affect of doing that?

Also, how much effort do you anticipate it'll be for us to maintain this fork?

pkg/detectors/mongodb/mongodb.go
```diff
@@ -85,6 +87,30 @@ func isErrDeterminate(err error) bool {
...
+			switch k {
+			case "tls":
+				if v[0] == "false" {
+					params.Set("tls", "false")
+				} else {
+					params.Set("tls", "true")
+				}
+			}
+		}
+	}
```
mcastorina, line 107: **Suggestion:**

```go
if v := parsed.Query().Get("tls"); v == "false" {
    params.Set("tls", "false")
} else if v != "" {
    params.Set("tls", "true")
}
```

pkg/detectors/mongodb/mongodb.go
```diff
@@ -85,6 +87,30 @@ func isErrDeterminate(err error) bool {
...
+	params.Set("connectTimeoutMS", "3000")
+	params.Set("timeoutMS", "3000")
```
mcastorina, line 109: Should this be using the `timeout` argument?

mcastorina (approved):

pkg/detectors/mongodb/mongodb.go
```diff
@@ -85,6 +87,30 @@ func isErrDeterminate(err error) bool {
 }
 
 func verifyUri(uri string, timeout time.Duration) error {
+	parsed, err := url.Parse(uri)
+	if err != nil {
+		return err
+	}
+
+	params := url.Values{}
+	for k, v := range parsed.Query() {
```
rgmz, line 96: The current MongoDB pattern needs to be updated to handle query parameters  (see #1550).

pkg/detectors/mongodb/mongodb.go
dustin-decker: I added this in a way to easily add more parameters in the future, even if it's not the most efficient.

pkg/detectors/mongodb/mongodb.go
dustin-decker: Oh, I'll take a look. Thank you.

PR #2528, opened 2024-02-29
[feat] - Make the client configurable

ahrav:
### Description:
Due to a bug in the underlying [retryablehttp.Clien](https://github.com/hashicorp/go-retryablehttp/blob/4165cf8897205a879a06b20d1ed0a2a76fbb6a17/client.go#L600)t the default CheckRetry logic sometimes does not correctly adhere to 429 retry after headers. 

The core issue arises from the fact that each goroutine independently tracks its own retry attempts (i) but decrements a shared retry budget (RetryMax) without coordination. This could lead to scenarios where, staggered requests might exhaust the retry limit prematurely because they do not account for the global state of retries across all requests.

This PR allows us to provide a custom `CheckRetry` fxn
### Checklist:

### DIFF
--- a/pkg/common/http.go
+++ b/pkg/common/http.go
@@ -115,6 +115,29 @@ func ConstantResponseHttpClient(statusCode int, body string) *http.Client {
 	}
 }
 
+// ClientOption configures how we set up the client.
+type ClientOption func(*retryablehttp.Client)
+
+// WithCheckRetry allows setting a custom CheckRetry policy.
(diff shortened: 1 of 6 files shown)


mcastorina (approved):
lgtm!

ahrav:
> lgtm!

Thank you. Going to hold on merging until tomorrow. Wanted to chat about the solution I was hoping making this configurable would solve. I'm not entirely sold on my approach. 😅

PR #2531, opened 2024-03-01
Canary verification

joeleonjr:
### Description:
Canaries register as valid if the secret_key meets the REGEX requirements, due to the new canary detection logic. This PR adds live verification (only for canaries) against the SNS `Publish` endpoint, which doesn't cause a Canary alert. The Canary's ARN is also extracted and added to ExtraData.

### Checklist:


dustin-decker (approved):
LGTM after updating tests and some minor adjustments. Nice job!

mcastorina:
> This PR adds live verification (only for canaries) against the SNS `Publish` endpoint

I'm not very familiar with AWS. Does the SNS `Publish` endpoint cause a side-effect?

rosecodym:
> > This PR adds live verification (only for canaries) against the SNS `Publish` endpoint
> 
> I'm not very familiar with AWS. Does the SNS `Publish` endpoint cause a side-effect?

The new verifier attempts to send the text message "foo" to the phone number `1`. If the credential actually has SNS permissions this might show up as a weird logged error message, but I think it's reasonable to not be concerned about that for canary tokens.

pkg/detectors/aws/aws_test.go
```diff
@@ -286,15 +269,44 @@ func TestAWS_FromChunk(t *testing.T) {
...
 						"account":       "171436882533",
+						"is_canary":     "true",
+						"message":       "This is an AWS canary token generated at canarytokens.org, and was not set off; learn more here: https://trufflesecurity.com/canaries",
 					},
 				},
 			},
 			wantErr:               false,
 			wantVerificationError: true,
 		},
+		//// Needs a non-canary token to test this
```
rosecodym, line 280: This test uses a fake HTTP client. It shouldn't matter what secret material gets used. Happy to chat about this synchronously to explain what it's doing and figure out a solution to whatever problem you were having.

pkg/detectors/aws/aws_test.go
```diff
@@ -333,8 +345,13 @@ func TestAWS_FromChunk(t *testing.T) {
 					t.Fatalf("wantVerificationError %v, verification error = %v", tt.wantVerificationError, got[i].VerificationError())
 				}
 			}
-			ignoreOpts := cmpopts.IgnoreFields(detectors.Result{}, "RawV2", "Raw", "verificationError")
-			if diff := cmp.Diff(got, tt.want, ignoreOpts); diff != "" {
+			ignoreOpts := []cmp.Option{
+				cmpopts.IgnoreFields(detectors.Result{}, "RawV2", "Raw", "verificationError"),
+				cmpopts.SortSlices(func(x, y detectors.Result) bool {
+					return x.Redacted < y.Redacted
+				}),
```
rosecodym, line 352: Why is this necessary now?

rosecodym (commented):
Looks like it just needs some test cleanup, which I don't expect to be difficult.

pkg/detectors/aws/aws.go
```diff
@@ -180,20 +178,36 @@ func (s scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 				s1.ExtraData["account"] = account
 			}
-			if !s.verifyCanaries && !strings.Contains(dataStr, "Redacted: "+resIDMatch) {
-				if _, ok := thinkstCanaryList[account]; ok {
-					s1.ExtraData["is_canary"] = "true"
-					s1.ExtraData["message"] = thinkstMessage
+			if _, ok := thinkstCanaryList[account]; ok {
+				s1.ExtraData["is_canary"] = "true"
+				s1.ExtraData["message"] = thinkstMessage
+				verified, arn, err := s.verifyCanary(resIDMatch, resSecretMatch)
```
rosecodym, line 184: oh @joeleonjr you raised an excellent point internally - should we be doing this if the user is running with verification disabled?

pkg/detectors/aws/aws_test.go
dustin-decker: This test case was flakey because it sometimes arrived in a different order:

```found two, returned both because the active secret for one paired with the inactive ID, despite the hash```

dustin-decker (approved):
Great job. Thanks Joe

PR #2537, opened 2024-03-04
[bug] - Fix resource leak in extractDebContent and extractRpmContent function

ahrav:
### Description:
The `extractDebContent` and `extractRpmContent` functions in the Archive type had a potential resource leak issue. The temporary file and directory created using createTempEnv were not being properly cleaned up in cases where an error occurred before reaching the defer statements responsible for their removal.

To address this issue, the defer statements for removing the temporary file and directory have been moved inside an anonymous function that is immediately invoked. This ensures that the cleanup logic is always executed, regardless of any errors that may occur during the execution of the function.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -355,8 +355,10 @@ func (a *Archive) extractDebContent(ctx logContext.Context, file io.Reader) (io.
 	if err != nil {
 		return nil, err
 	}
-	defer os.Remove(tmpEnv.tempFileName)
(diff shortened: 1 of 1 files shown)


bill-rich:
How does putting them in a `defer func` behave differently than just `defer`ing them directly?

ahrav:
> How does putting them in a `defer func` behave differently than just `defer`ing them directly?

In the even there is an error

PR #2538, opened 2024-03-05
Modified the Azure Storage proto

dkazantzas:
### Description:
I've modified the sources.proto and credentials.proto files so that the AzureStorage message and the related credentials, match the authentication types that are, currently, supported by the official Azure SDK.   

### Checklist:


dustin-decker:
Thanks for the suggestion, Dimitris. Unfortunately this conflicts with our TruffleHog Enterprise implementation of Azure Storage. We could consider adding these as additional auth methods when we open source the Azure Storage scanning integration.

PR #2539, opened 2024-03-05
Redact secret in git command output

rosecodym:
### Description:
When we fail to clone a git repository we log the command output to help with diagnosis. However, this output can include credentials in certain cases (such as certain errors associated with redirects). We don't want to log credentials when this happens.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -986,6 +986,10 @@ func resolveHash(repo *git.Repository, ref string) (string, error) {
 	return resolved.String(), nil
 }
 
+// stripPassword removes username:password contents from URLs. The first return value is the cleaned URL and the second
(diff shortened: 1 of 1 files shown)


ahrav (approved):

pkg/sources/git/git.go
```diff
@@ -979,19 +986,20 @@ func resolveHash(repo *git.Repository, ref string) (string, error) {
 	return resolved.String(), nil
 }
 
-func stripPassword(u string) (string, error) {
+func stripPassword(u string) (string, string, error) {
```
mcastorina, line 989: **Nit:** Please add a doc comment here. Now that the function is returning two strings, it would be good to note which is which, and why we would want to return the password.

mcastorina (approved):

PR #2554, opened 2024-03-08
pull out verification logic from github detectors

0x1:
### Description:
this was initially prompted by [Gomez's suggesion](https://github.com/trufflesecurity/trufflehog/pull/2471#discussion_r1493011083) that we should consolidate the verification logic for both versions of github detectors to reduce logic. but turns out according to Go's convention, each directory represents a separate package if they have different paths, even if the package declaration within the files is the same. so the detectors can't access each other's private methods. we _could_ technically make them exported but i'm not sure if that's wise.

in any case, the refactor to pull out verification was already done and should still be useful.

### Checklist:

### DIFF
--- a/pkg/detectors/github/v1/github_old.go
+++ b/pkg/detectors/github/v1/github_old.go
@@ -44,6 +44,11 @@ type userRes struct {
 	LdapDN string `json:"ldap_dn"`
 }
 
+type headerInfo struct {
+	Scopes string `json:"X-OAuth-Scopes"`
+	Expiry string `json:"github-authentication-token-expiration"`
+}
+
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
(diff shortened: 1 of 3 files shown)


dustin-decker:
I'd prefer exported functions over duplicated code. If anyone tries to import those functions elsewhere, that's their problem.

pkg/detectors/github/v1/github_old.go
```diff
@@ -72,63 +77,20 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-								if expiry != "" {
-									s1.ExtraData["expires_at"] = expiry
-								}
-							}
-						}
-					}
-				}
-			}
+
+			isVerified, userResponse, headers, err := s.VerifyGithub(ctx, client, token)
```
ahrav, line 89: question: Did you consider returning ptrs for `userResponse` and `headers`? If we return pointers you can more easily handle the nil check so you don't have to do the comparison to empty structs in both `SetUserResponse` and `SetHeaderInfo`.

You could do something like:
```
if userResponse != nil {
  SetuserResponse(*userResponse, &s1)
}
if headers != nil {
  SetHeaderInfo(*headers, &s1)
}
```

ahrav (approved):

PR #2556, opened 2024-03-08
[chore] -  Record metrics before reset

ahrav:
### Description:

### Checklist:

### DIFF
--- a/pkg/writers/buffer/buffer.go
+++ b/pkg/writers/buffer/buffer.go
@@ -24,10 +24,11 @@ func (poolMetrics) recordBufferRetrival() {
 	bufferCount.Inc()
 }
 
-func (poolMetrics) recordBufferReturn(bufCap, bufLen int64) {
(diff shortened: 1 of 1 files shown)


rosecodym:
What's the reason for this change?

ahrav:
> What's the reason for this change?

Our metrics were reporting some checkout durations that were very very large due to integer overflow for newly created buffers with the zero value for time.Time. This was then resulting in metrics that were incorrect.

rosecodym (approved):
ah, ok. this isn't necessary but you might also consider modifying `recordBufferReturn` to take the buffer itself as a parameter and then call `buf.RecordMetric()` inside of it. then all the buffer return metrics will be in one place.

ahrav:
> ah, ok. this isn't necessary but you might also consider modifying `recordBufferReturn` to take the buffer itself as a parameter and then call `buf.RecordMetric()` inside of it. then all the buffer return metrics will be in one place.

Yea, I like that idea thanks.

PR #2563, opened 2024-03-10
Create decoder for HTML entities

### LINKED ISSUE #2231: Create a decoder for HTML entites

## Description
While uncommon, data that contains secrets can be encoded as [HTML entites](https://www.freeformatter.com/html-entities.html). This likely wouldn't be detected with the current solution.

### Preferred Solution

A custom [`Decoder`](https://github.com/trufflesecurity/trufflehog/blob/b0fab16ad4b5cc42c38ee4e2ee362aa3c47dea24/pkg/decoders/decoders.go#L24-L26) that detects and decodes HTML entities. Whether it should decode _all_ entities or only basic/common ascii characters remains to be seen.

### Additional Context

N/A

#### References

N/A

rgmz:
### Description:
This creates a decoder to handle [HTML entities](https://www.freeformatter.com/html-entities.html). Tests pass, but the implementation may not be the most efficient.

This fixes #2231.

### Checklist:


pkg/decoders/html_entity.go
```diff
@@ -0,0 +1,168 @@
...
+		chunk.Data = decodeHtmlDecimal(chunk.Data)
+	}
+	if hexEntityPat.Match(chunk.Data) {
+		matched = true
+		chunk.Data = decodeHtmlHex(chunk.Data)
+	}
+
+	if matched {
+		decodableChunk := &DecodableChunk{
+			DecoderType: detectorspb.DecoderType_ESCAPED_UNICODE,
```
dustin-decker, line 41: Should this be a new decoder type?

dustin-decker:
I think we've reached the point where we should consider adding a `--enabled/disabled-decoders` flag, similar to what we have for detectors. This one seems pretty impactful on performance in its current state, and many data sources might not benefit much from it.

One potential improvement might be to implement this as a handler and do identification of the whole file before decoding and chunking it out.

rgmz:
I do worry about the impact of having too many decoders. At a minimum, having something like ahocorasick might be more efficient than checking `regexp.Match()` against each chunk. 

> One potential improvement might be to implement this as a handler and do identification of the whole file before decoding and chunking it out.

While I think identifying the mimetype of a file would be a great addition (and make way for other enhancements), I'm not sure how much it would help in this case. HTML, Markdown, and AsciiDoc files are obviously sources that would benefit, but HTML-encoded content can show up in weird places like config files, [`.txt` files](https://github.com/aciborowska/jingo/blob/c1a21f041501d5e6653f2e09046a5524af7274ea/datasets/bench4bl/spring-data-mongodb/queries/long/1216.txt#L10), or [source code](https://github.com/pulumi/pulumi-mongodbatlas/blob/b41c509550dd50f40e5368dbd1f635714ae55c3b/sdk/dotnet/Cluster.cs#L344).

This decoder was act inspired by #1550; I found several live connection strings that were not detected by TruffleHog because they contained encoded `&amp;` instead of a literal `&`.
```
mongodb://dave:password@localhost:27017/?authMechanism=DEFAULT&amp;authSource=db&amp;ssl=true&quot;
```

PR #2577, opened 2024-03-14
MaxMind detector uses the right endpoint

faktas2:
### Description:
The endpoint that the current detector uses fails in verifying the license key as some license keys do not have permissions to the geoip API. This commit is to make the detector use the right endpoint https://dev.maxmind.com/license-key-validation-api

### Checklist:

### DIFF
--- a/pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
+++ b/pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
@@ -44,7 +44,6 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			DetectorType: detectorspb.DetectorType_MaxMindLicense,
 			Redacted:     keyRes,
 			Raw:          []byte(keyRes),
-			RawV2:        []byte(keyRes),
 			ExtraData: map[string]string{
 				"rotation_guide": "https://howtorotate.com/docs/tutorials/maxmind/",
 				"version":        fmt.Sprintf("%d", s.Version()),
--- a/pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
+++ b/pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
@@ -57,9 +57,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			req, err := http.NewRequestWithContext(
 				ctx, "POST", "https://secret-scanning.maxmind.com/secrets/validate-license-key",
 				strings.NewReader(data.Encode()))
-			if err != nil {
-				continue
-			}
+			r.SetVerificationError(err, keyRes)
 			req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
 			res, err := client.Do(req)
--- a/pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
+++ b/pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"regexp"
+	"strings"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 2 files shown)


pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
```diff
@@ -37,49 +36,44 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 	for _, keyMatch := range keyMatches {
 		keyRes := strings.TrimSpace(keyMatch[1])
 
-		for _, idMatch := range idMatches {
-			if len(idMatch) != 2 {
+		r := detectors.Result{
+			DetectorType: detectorspb.DetectorType_MaxMindLicense,
+			Redacted:     keyRes,
+			Raw:          []byte(keyRes),
+			RawV2:        []byte(keyRes),
```
ugexe, line 47: I don't think `RawV2` needs to be set if it will always be the same as `Raw`

https://github.com/trufflesecurity/trufflehog/blob/e66bdfb7f911eafd465cad1d6a1bdc3462663c59/pkg/detectors/detectors.go#L53-L55

It appears other detectors that do not use an "id" also do not generally set `RawV2`

pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
```diff
@@ -37,49 +36,43 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-						}
+			req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+			req.Form.Add("license_key", keyRes)
+
+			res, err := client.Do(req)
+			if err == nil {
+				defer res.Body.Close()
+				if res.StatusCode >= 200 && res.StatusCode < 300 {
+					r.Verified = true
+				} else {
```
ugexe, line 67: I'm not sure this `IsKnownFalsePositive` branch is needed. It doesn't do anything when it does detect a false positive, and it looks like the values for `detectors.DefaultFalsePositives` wouldn't match our regex anyway

pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
```diff
@@ -37,49 +36,43 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	dataStr := string(data)
 
 	keyMatches := keyPat.FindAllStringSubmatch(dataStr, -1)
-	idMatches := idPat.FindAllStringSubmatch(dataStr, -1)
 
 	for _, keyMatch := range keyMatches {
 		keyRes := strings.TrimSpace(keyMatch[1])
```
ugexe, line 41: Is this `TrimSpace(...)` still needed?

pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
```diff
@@ -37,49 +37,41 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-		for _, idMatch := range idMatches {
-			if len(idMatch) != 2 {
+		if verify {
+			data := url.Values{}
+			data.Add("license_key", keyRes)
+			req, err := http.NewRequestWithContext(
+				ctx, "POST", "https://secret-scanning.maxmind.com/secrets/validate-license-key",
+				strings.NewReader(data.Encode()))
+			if err != nil {
 				continue
```
ugexe, line 61: It looks like other detectors use e.g `r.SetVerificationError(err, keyRes)` for the various errors

https://github.com/trufflesecurity/trufflehog/blob/1f0b53b530c50637d3a784e8a3fa1eb47097a464/pkg/detectors/accuweather/accuweather.go#L62-L67

ugexe: Although maybe it doesn't matter since we wouldn't be collecting anything with an error into `results` 🤔

faktas2: I think it should be fine AFAICT

pkg/detectors/maxmindlicense/v2/maxmindlicense_v2.go
```diff
@@ -57,9 +57,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			req, err := http.NewRequestWithContext(
 				ctx, "POST", "https://secret-scanning.maxmind.com/secrets/validate-license-key",
 				strings.NewReader(data.Encode()))
-			if err != nil {
-				continue
-			}
+			r.SetVerificationError(err, keyRes)
```
ugexe, line 60: I think if there was an `err` that we wouldn't want to try use the `req` later. So I think I mislead you about using `SetVerificationError`

faktas2: I've added it. Is that what you meant ?

ugexe (approved):

dustin-decker (approved):
Thank you

PR #2579, opened 2024-03-15
Postman Source

zricethezav:
### Description:
This PR adds a new source, [Postman](https://www.postman.com/). I encourage whoever reviews this to take a look at the Postman UI then take a crack at this. Some important things to note:
- The Postman source is unlike any of our other sources as we are _creating_ chunks from information gathered along the way as we walk Postman data. We keep track of eligible keywords from things like variable keys, collection names, workspace names, urls, request bodies, etc. We also keep track of variables that can be used for substitution if they are referenced in other parts of the Postman UI. For example, a variable `var1` defined in some environment can be referenced in a request with `{{var1}}`.
- Metadata is passed down through each subsequent `scan{thing}` and updated to keep track of the path took to generate a chunk. Once data to be scanned makes it to `scanData` we should have a link to whatever postman "object" (think request, auth, variable table, environment, etc) page, and a history of how we got there.
Example of how this is communicated in a result output:
```
Collection_id: FAKE-f91d96b5-83b6-4cca-a2f1-FAKE
Collection_name: OpenAI
Environment_id: FAKE-FAKE-40cc-4c40-9fce-FAKE
Field_type: request > request auth > authorization
Folder_id: FAKE-3471-4061-82a8-FAKE
Folder_name: embeddings
Link: https://go.postman.co/request/UUID-FAKE?tab=auth
Request_id: a32fda07-FAKE-4061-FAKE-SUPERFAKE
Request_name: embeddings
Workspace_name: Cool Org Workspace
Workspace_uuid: FAKE-FAKE-4a22-FAKE-FAKE
```
- Variable substitution is fun and uses recursion. I tried an iterative approach but wasn't able to crack it. Could be optimized later on.
- This is a hidden command for now until we get more feedback from the community.

### Checklist:


pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,678 @@
...
+			case "accessKey":
+				authData += fmt.Sprintf("accessKey:%s ", kv.Value)
+			case "secretKey":
+				authData += fmt.Sprintf("secretKey:%s ", kv.Value)
+			case "region":
+				authData += fmt.Sprintf("region:%s ", kv.Value)
+			case "service":
+				authData += fmt.Sprintf("service:%s ", kv.Value)
+			}
+		}
```
mcastorina, line 451: Why do we only take the last kv in the `"apikey"` case, but take all of them for the `"awsv4"` case? Why is that different? Same goes for the other cases but I wanted to highlight this example.

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,678 @@
...
+				username = fmt.Sprintf("%v", kv.Value)
+			case "password":
+				password = fmt.Sprintf("%v", kv.Value)
+			}
+		}
+		if u.Raw != "" {
+			parsedURL, err := url.Parse(u.Raw)
+			if err != nil {
+				ctx.Logger().V(2).Info("error parsing URL in basic auth check", "url", u.Raw)
+			}
```
mcastorina, line 475: **Suggestion:** return on error

```go
if err != nil {
    // ...
    return
}
```

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,678 @@
...
+
+			parsedURL.User = url.User(username + ":" + password)
+			decodedURL, err := url.PathUnescape(parsedURL.String())
+			if err != nil {
+				ctx.Logger().V(2).Info("error parsing URL in basic auth check", "url", u.Raw)
+			}
```
mcastorina, line 481: **Suggestion:** return on error

```go
if err != nil {
    // ...
    return
}
```

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,678 @@
...
+
+func shouldSkip(uuid string, include []string, exclude []string) bool {
+	if len(include) > 0 && !slices.Contains(include, uuid) {
+		return true
+	}
+	if slices.Contains(exclude, uuid) {
+		return true
+	}
+	return false
+}
```
mcastorina, line 678: This looks like inclusion takes precedence, which I think is in opposition of how our other sources work? I thought exclusion took precedence, though I could be wrong. I can't wait to standardize this part of sources 🤣

pkg/sources/postman/postman_client.go
```diff
@@ -0,0 +1,402 @@
...
+	}{}
+
+	r, err := c.getPostmanReq("https://api.getpostman.com/workspaces", nil)
+	if err != nil {
+		err = fmt.Errorf("could not get workspaces")
+		return workspaces, err
+	}
+	r.Body.Close()
+
+	body, err := io.ReadAll(r.Body)
```
mcastorina, line 278: I think this is supposed to be `defer r.Body.Close()`

pkg/sources/postman/postman_client.go
```diff
@@ -0,0 +1,402 @@
...
+// GetCollection returns the collection for a given collection
+func (c *Client) GetCollection(collection_uuid string) (Collection, error) {
+	obj := struct {
+		Collection Collection `json:"collection"`
+	}{}
+
+	url := fmt.Sprintf(COLLECTIONS_URL, collection_uuid)
+	r, err := c.getPostmanReq(url, nil)
+	if err != nil {
+		fmt.Println(err)
```
mcastorina, line 385: Debug print?

pkg/sources/postman/postman_client.go
```diff
@@ -0,0 +1,402 @@
...
+// GetEnvironmentVariables returns the environment variables for a given environment
+func (c *Client) GetEnvironmentVariables(environment_uuid string) (VariableData, error) {
+	obj := struct {
+		VariableData VariableData `json:"environment"`
+	}{}
+
+	url := fmt.Sprintf(ENVIRONMENTS_URL, environment_uuid)
+	r, err := c.getPostmanReq(url, nil)
+	if err != nil {
+		fmt.Println(err)
```
mcastorina, line 357: Debug print?

mcastorina (commented):
First pass, but I came across a lot of questions about the structure that I think I would benefit from a synchronous discussion.

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,678 @@
...
+	for _, keyword := range keywords {
+		s.addKeyword(keyword)
+	}
+}
+
+func (s *Source) addKeyword(keyword string) {
+	if _, ok := s.detectorKeywords[keyword]; ok {
+		s.keywords[keyword] = struct{}{}
+	}
+}
```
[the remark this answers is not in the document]
zricethezav: it's checking if the keyword is in detectorkeywords

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,678 @@
...
+				rc, err := file.Open()
+				if err != nil {
+					return err
+				}
+				defer rc.Close()
+				contents, err := io.ReadAll(rc)
+				if err != nil {
+					return err
+				}
+				if strings.Contains(file.Name, "collection") {
```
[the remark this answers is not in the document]
zricethezav: standardized

pkg/engine/postman.go
```diff
@@ -0,0 +1,65 @@
...
+		ctx.Logger().Error(errors.New("no postman workspaces, collections, environments or API token provided"), "failed to scan postman")
+		return nil
+	}
+
+	// Turn AhoCorasick keywordsToDetectors into a map of keywords
+	keywords := []string{}
+	for key := range e.ahoCorasickCore.KeywordsToDetectors {
+		keywords = append(keywords, key)
+	}
+	connection.DetectorKeywords = keywords
```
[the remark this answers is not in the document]
zricethezav: yea should probably be set in the source directly. Just need to export DetectorKeywords in the postman source

main.go
```diff
@@ -153,6 +153,20 @@ var (
...
+	postmanWorkspaces          = postmanScan.Flag("workspace", "Postman workspace to scan. You can repeat this flag.").Strings()
+	postmanCollections         = postmanScan.Flag("collection", "Postman collection to scan. You can repeat this flag.").Strings()
+	postmanEnvironments        = postmanScan.Flag("environment", "Postman environment to scan. You can repeat this flag.").Strings()
+	postmanIncludeCollections  = postmanScan.Flag("include-collections", "Collections to include in scan. Comma separated list of collections. You can repeat this flag.").Strings()
+	postmanIncludeEnvironments = postmanScan.Flag("include-environments", "Environments to include in scan. Comma separated list of environments. You can repeat this flag.").Strings()
+	postmanExcludeCollections  = postmanScan.Flag("exclude-collections", "Collections to exclude from scan. Comma separated list of collections. You can repeat this flag.").Strings()
+	postmanExcludeEnvironments = postmanScan.Flag("exclude-environments", "Environments to exclude from scan. Comma separated list of environments. You can repeat this flag.").Strings()
```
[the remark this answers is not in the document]
zricethezav: you're right

pkg/sources/postman/postman.go
zricethezav:
```
			"auth": {
				"type": "basic",
				"basic": [
					{
						"key": "username",
						"value": "parentUsername",
						"type": "string"
					},
					{
						"key": "password",
						"value": "parentPassword",
						"type": "string"
					}
				]
			}
```

We have to iterate through KV elements and build up the data to be scanned.

pkg/engine/engine.go
```diff
@@ -887,8 +887,10 @@ func (e *Engine) notifyResults(ctx context.Context) {
 		// want to include duplicate results with the same decoder type.
 		// Duplicate results with the same decoder type SHOULD have their own entry in the
 		// results list, this would happen if the same secret is found multiple times.
+		// Note: If the source type is postman, we dedupe the results regardless of decoder type.
```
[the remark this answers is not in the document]
zricethezav: There would be a bunch of duplicates representing the same information. So for example if we have built say three keywords from walking the postman response (k1, k2, k3) and we have a valid token (token1) in some auth field then the data chunk that we create would look something like:

```
k1: token1
k2: token1
k3: token1
```

We don't want to report 3 different results since `token1` is leaked in _one_ auth field.

pkg/engine/ahocorasick/ahocorasickcore.go
```diff
@@ -38,7 +38,7 @@ type AhoCorasickCore struct {
 	// type and then again from detector type to detector. We could
 	// go straight from keywords to detectors but doing it this way makes
 	// some consuming code a little cleaner.)
-	keywordsToDetectors map[string][]DetectorKey
+	KeywordsToDetectors map[string][]DetectorKey
```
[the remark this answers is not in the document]
zricethezav: sike

pkg/sources/postman/postman.go
mcastorina: I guess I'm confused why the `"apikey"` case ignores everything but the last value in the array.

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,672 @@
...
+
+func shouldSkip(uuid string, include []string, exclude []string) bool {
+	if slices.Contains(exclude, uuid) {
+		return true
+	}
+	if len(include) > 0 && !slices.Contains(include, uuid) {
+		return true
+	}
+	return false
+}
```
mcastorina, line 672: Thanks for fixing this. I created this table to help me check the logic, so I'm going to leave it here in case it helps others too.

include list provided | included | excluded | shouldSkip |
|---|---|---|---|
| no | no | no | no |
| no | no | yes | yes |
| no | yes | no | X |
| no | yes | yes | X |
| yes | no | no | yes |
| yes | no | yes | yes |
| yes | yes | no | no |
| yes | yes | yes | yes |

`X` indicates impossible state (cannot have no provided include list and be included).

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,672 @@
...
+			}
+			for _, file := range r.File {
+				rc, err := file.Open()
+				if err != nil {
+					return err
+				}
+				contents, err := io.ReadAll(rc)
+				if err != nil {
+					return err
+				}
```
mcastorina, line 170:
```suggestion
				contents, err := io.ReadAll(rc)
				rc.Close()
				if err != nil {
					return err
				}
```

pkg/sources/postman/postman.go
```diff
@@ -0,0 +1,672 @@
...
+				if strings.Contains(file.Name, "environment") {
+					e := VariableData{}
+					if err = json.Unmarshal(contents, &e); err != nil {
+						return err
+					}
+					workspace.EnvironmentsRaw = append(workspace.EnvironmentsRaw, e)
+				}
+			}
+			r.Close()
```
mcastorina, line 187: The difficult thing about this approach is that we aren't closing `r` when we encounter an error that causes us to return from the function immediately. That's what `defer` solves, but unfortunately not within a loop body.

I'm okay with this working for the happy path for now, but a followup PR should fix this leak (which I'm happy to do too).

mcastorina (approved):
Great job implementing this! I really like the idea of it being a hidden subcommand so we can continue refining it.

Just had some suggestions on the file closing aspect, but I think that can be in a sooner-than-later followup PR.

PR #2582, opened 2024-03-16
Fix --results not behaving as expected.

### LINKED ISSUE #2580: The `--only-verified` option is ignored
I think #2372 change might have broken something.

The [default](https://github.com/oxsecurity/megalinter/blob/main/megalinter/descriptors/repository.megalinter-descriptor.yml#L575) command line arguments for trufflehog in [Megalinter](https://github.com/oxsecurity/megalinter) explicitly use the `--only-verified` option to avoid false positive on the `.git/config` file (as per https://github.com/oxsecurity/megalinter/pull/2838).

Today I have stumbled upon a false positive on exactly that `.git/config` file and, as it is visible from the GHA log in the gist linked below, that is an unverified result which should have been excluded by that option.

Besides, this failure started to occur suddenly, even though the pinned version of trufflehog in Megalinter is 3.69.0. After some digging I have discovered the cause is the fact that trufflehog seems to auto-update itself when run. I have addressed that in https://github.com/oxsecurity/megalinter/pull/3430 by explicitly adding the `--no-update` option.

### TruffleHog Version

Version is 3.70.0, as it may be visible in the GitHub Actions log excerpt linked in the gist below.

### Trace Output

https://gist.github.com/wandering-tales/51311d13eccf03a3b27388069dbcd0a4

### Expected Behavior

The `--only-verified` option should have filtered out unverified results

### Actual Behavior

An unverified result was included in the final report causing the command to fail.

### Steps to Reproduce

I cannot reproduce it locally. It looks like a false positive occurring solely in GitHub Action CI environments.

## Environment
 * GHA / Megalinter action

rgmz:
### Description:

This fixes #2580.

Go switches are not Java switches. Go switches are not Java switches. Go switches are not Java switches. Go switches are not Java switches. Go switches are not Java switches. Go switches are not Java switches. 

QELQoU3

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -616,16 +616,13 @@ func parseResults(input *string) (map[string]struct{}, error) {
 		return nil, nil
 	}
 
-
 	var (
 		values  = strings.Split(strings.ToLower(*input), ",")
 		results = make(map[string]struct{}, 3)
 	)
 	for _, value := range values {
 		switch value {
-		case "verified":
-		case "unknown":
-		case "unverified":
+		case "verified", "unknown", "unverified":
 			results[value] = struct{}{}
 		default:
 			return nil, fmt.Errorf("invalid value '%s', valid values are 'verified,unknown,unverified'", value)


mcastorina (approved):
oof, that's an easy one to make. thanks for keeping an eye out for issues with the new flag.

PR #2585, opened 2024-03-16
Make Git.handleBinary work with escaped unicode characcters

### LINKED ISSUE #2418: Git does not work with escaped unicode characcters

### TruffleHog Version
3.67.5

### Trace Output

```
2024-02-11T17:22:52-05:00       error   trufflehog      error waiting for command       {"source_manager_worker_id": "sIEwQ", "repo": "https://github.com/appleboy/DeepLearning-500-questions.git", "command": "/usr/bin/git -C /tmp/trufflehog-169189-937274496/.git cat-file blob f71e35a836b7f97301a23fe6af6a8305e285467c:\\346\\267\\261\\345\\272\\246\\345\\255\\246\\344\\271\\240500\\351\\227\\256-Tan-00\\347\\233\\256\\345\\275\\225.docx", "stderr": "fatal: Not a valid object name f71e35a836b7f97301a23fe6af6a8305e285467c:\\346\\267\\261\\345\\272\\246\\345\\255\\246\\344\\271\\240500\\351\\227\\256-Tan-00\\347\\233\\256\\345\\275\\225.docx\n", "commit": "f71e35a836b7f97301a23fe6af6a8305e285467c", "error": "exit status 128"}
```

(The corresponding file is [`]文件500页-Tan-00数据统计.docx`](https://github.com/scutan90/DeepLearning-500-questions/blob/master/%E6%B7%B1%E5%BA%A6%E5%AD%A6%E4%B9%A0500%E9%97%AE-Tan-00%E7%9B%AE%E5%BD%95.docx))

### Expected Behavior

The `git cat-file` command should handle escaped filenames.

https://github.com/trufflesecurity/trufflehog/blob/f35185e2152d562a62d07031288fbd03b13ceeea/pkg/sources/git/git.go#L1168

### Actual Behavior

The command should pass in unescaped characters.

e.g.,
```
git cat-file blob e48902879384ff53812c6420440db9ac1c50bb99:"$(printf '\320\244\320\270\320\267\320\270\320\272\320\260/\320\255\320\272\320\267\320\260\320\274\320\265\320\275/\320\236\321\202\321\201\320\276\321\200\321\202\320\270\321\200\320\276\320\262\320\260\320\275\320\275\321\213\320\265 \321\200\320\265\321\210\320\265\320\275\320\275\321\213\320\265 \320\261\320\270\320\273\320\265\321\202\321\213 2017/1 \320\261\320\270\320\273\320\265\321\202.pdf')"
```

### Steps to Reproduce

Attempt to scan https://github.com/scutan90/DeepLearning-500-questions

## Environment
N/A

## Additional Context

1. This is similar to #2384. Need to figure out if the solution is to add quotes or to decode the name. (https://stackoverflow.com/a/69817595)
1. PR #2174 replaced go-git with direct calls to Git. It's worth testing the capabilities of go-git to see if it handles edge cases like this.

### References
N/A

rgmz:
### Description:
This fixes #2418. The issue appears to be that converting `[]byte` -> `string` does not decode unicode characters, whereas doing something like `fmt.Printf("Path is: '%s'\n", path)` will do it automatically.

### Checklist:


pkg/gitparse/gitparse.go
```diff
@@ -681,27 +681,33 @@ func isBinaryLine(latestState ParseState, line []byte) bool {
...
-	if ok {
 		// drop the `" differ\n`
-		return string(after[:len(after)-9]), true
+		path = string(after[:len(after)-9])
+	} else {
+		// Unknown format.
+		return "", false
+	}
+
+	unicodePath, err := strconv.Unquote(`"` + path + `"`)
```
rosecodym, line 704: What made you use this technique over sprintf? Either way, would you mind adding a comment explaining the pitfall you're now avoiding?

rosecodym (commented):
Thanks for digging into this! It looks fine to _me_, but I'm going to tag in @ahrav who's spent a lot of time mucking around with the performance of string<->unicode conversions. Ahrav, do you see any non-obvious pitfalls here?

pkg/gitparse/gitparse.go
ahrav: question: Do we need the string concat, could we just pass in `path`, because it seems that the path variable is expected to be either an unquoted string or a double-quoted string?

It's also very likely i'm just missing something obvious here 😓

rgmz: > What made you use this technique over sprintf? Either way, would you mind adding a comment explaining the pitfall you're now avoiding?

That was speculation given that `fmt.Printf` seemed to worked; I checked, and sprintf doesn't work.

> question: Do we need the string concat, could we just pass in path, because it seems that the path variable is expected to be either an unquoted string or a double-quoted string?

[The quotes are required](https://pkg.go.dev/strconv#Unquote), for some reason.

ahrav: fwiw when joining strings with few arguments using `+` is generally the fastest approach as well.
Benchmarks with a few different approaches:
[some benchmarks](https://gist.github.com/dtjm/c6ebc86abe7515c988ec)

ahrav (approved):

rgmz:
I realized this should have decoding functionality for normal non-binary paths as well. Gotta work on that + add tests.

PR #2602, opened 2024-03-20
expand keyword checks, and add collection name to keyword

zricethezav:
### Description:
This PR loosens the constraints from an exact match to "contains" for detector keywords. e.g. "datadog-token" is a variable key in postman, but "datadog" is a detector keyword. Before this change we would not be adding "datadog" to the keywords.

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman.go
+++ b/pkg/sources/postman/postman.go
@@ -60,8 +60,18 @@ func (s *Source) addKeywords(keywords []string) {
 }
 
 func (s *Source) addKeyword(keyword string) {
+	// fast check
 	if _, ok := s.DetectorKeywords[keyword]; ok {
 		s.keywords[keyword] = struct{}{}
+		return
+	}
+
+	// slow check. This is to handle the case where the keyword is a substring of a detector keyword
+	// e.g. "datadog-token" is a variable key in postman, but "datadog" is a detector keyword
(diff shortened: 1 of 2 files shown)


pkg/sources/postman/postman.go
```diff
@@ -60,8 +60,18 @@ func (s *Source) addKeywords(keywords []string) {
...
 	if _, ok := s.DetectorKeywords[keyword]; ok {
 		s.keywords[keyword] = struct{}{}
+		return
+	}
+
+	// slow check. This is to handle the case where the keyword is a substring of a detector keyword
+	// e.g. "datadog-token" is a variable key in postman, but "datadog" is a detector keyword
+	for k := range s.DetectorKeywords {
+		if strings.Contains(keyword, k) {
+			s.keywords[k] = struct{}{}
```
joeleonjr, line 73: In this example, do we still want "datadog-token" or just "datadog". I only ask b/c character distance b/t keyword and secret could be impacted.

zricethezav: We want to store just `datadog` as `datadog-token` doesn't provide additional useful information.

joeleonjr: Ah yes. And looks like that's what you originally did. Read that wrong.

ahrav (approved):

PR #2605, opened 2024-03-21
Avoid uneeded calls to strconv.Unquote

rgmz:
### Description:
This is a follow-up to #2585. I realized that unicode escapes only seem to be present if the path is quoted; hence, calling `strconv.Unquote` is unnecessary if the path isn't quoted. 

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -424,7 +424,7 @@ func (c *Parser) FromReader(ctx context.Context, stdOut io.Reader, diffChan chan
 		case isBinaryLine(latestState, line):
 			latestState = BinaryFileLine
 
-			path, ok := pathFromBinaryLine(ctx, line)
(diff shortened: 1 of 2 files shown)


rosecodym (commented):
Why did you remove the error logging?

rgmz:
Those lines were actually redundant. I decided to keep the logging with the more detailed messages.

pkg/gitparse/gitparse.go
```diff
@@ -753,29 +753,29 @@ func pathFromToFileLine(ctx context.Context, line []byte) (string, bool) {
 		return "", true
 	}
 
-	var path string
+	var (
+		path string
+		err  error
```
rosecodym, line 758: i don't think you need to declare this here? but that's extremely minor

rosecodym (approved):

dustin-decker (approved):

PR #2608, opened 2024-03-22
Fix GitHub panic and test errors

rgmz:
### Description:
This is a follow-up to #2379. It fixes a panic that could occur during enumeration, as well as some test failures.

```
2024-03-22T08:38:54-04:00       error   trufflehog      goroutine 2116 [running]:
runtime/debug.Stack()
        /home/user/sdk/go1.21.0/src/runtime/debug/stack.go:24 +0x5e
github.com/trufflesecurity/trufflehog/v3/pkg/common.Recover({0x3b3c000, 0xc0027b8030})
        /tmp/trufflehog/pkg/common/recover.go:17 +0x58
panic({0x308c1a0?, 0xc001938e10?})
        /home/user/sdk/go1.21.0/src/runtime/panic.go:914 +0x21f
github.com/trufflesecurity/trufflehog/v3/pkg/sources/github.(*Source).enumerate(0xc001dffb00, {0x3b3c000?, 0xc0027b8030?}, {0x32d2356?, 0xc0027c5c98?})
        /tmp/trufflehog/pkg/sources/github/github.go:421 +0x8b2
github.com/trufflesecurity/trufflehog/v3/pkg/sources/github.(*Source).Chunks(0xc001dffb00, {0x3b3c000, 0xc0027b8030}, 0xc002440000?, {0x0?, 0x0?, 0x0?})
        /tmp/trufflehog/pkg/sources/github/github.go:372 +0x29d
github.com/trufflesecurity/trufflehog/v3/pkg/sources.(*SourceManager).runWithoutUnits(0xc000e4a090, {0x3b3c000, 0xc0027b8030}, {0x3b3fb70?, 0xc001dffb00}, 0xc0000f9ce0, {0x0, 0x0, 0x0})
        /tmp/trufflehog/pkg/sources/source_manager.go:308 +0x204
github.com/trufflesecurity/trufflehog/v3/pkg/sources.(*SourceManager).run(0xc000e4a090, {0x3b3c000, 0xc0027b8030}, {0x3b3fb70, 0xc001dffb00}, 0xc0000f9ce0, {0x0, 0x0, 0x0})
        /tmp/trufflehog/pkg/sources/source_manager.go:283 +0x627
github.com/trufflesecurity/trufflehog/v3/pkg/sources.(*SourceManager).Run.func1()
        /tmp/trufflehog/pkg/sources/source_manager.go:159 +0x299
[...]

PR #2614, opened 2024-03-22
Fix additional GitHub test errors

rgmz:
### Description:
Follow-up to #2608. Fixes additional test errors.

### Checklist:

### DIFF
--- a/pkg/sources/errors_test.go
+++ b/pkg/sources/errors_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 )
 
+var testError = fmt.Errorf("simulated failure")
+
 func TestNewScanErrors(t *testing.T) {
 	testCases := []struct {
 		name     string
@@ -84,7 +86,7 @@ func TestScanErrorsAdd(t *testing.T) {
 				wg.Add(1)
 				go func() {
 					for j := 0; j < tc.wantErr/tc.concurrency; j++ {
-						se.Add(nil)
+						se.Add(testError)
 					}
 					wg.Done()
 				}()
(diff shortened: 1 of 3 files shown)


pkg/sources/errors_test.go
```diff
@@ -84,7 +86,7 @@ func TestScanErrorsAdd(t *testing.T) {
 				wg.Add(1)
 				go func() {
 					for j := 0; j < tc.wantErr/tc.concurrency; j++ {
-						se.Add(nil)
+						se.Add(testError)
```
rgmz, line 89: RE: https://github.com/trufflesecurity/trufflehog/pull/2608#discussion_r1535524068

My guess is that `nil` was allowed to make this test easier.

pkg/sources/github/github_test.go
```diff
@@ -742,8 +768,6 @@ func TestGetGistID(t *testing.T) {
 	}{
 		{[]string{"https://gist.github.com", "12345"}, "12345"},
 		{[]string{"https://gist.github.com", "owner", "12345"}, "12345"},
-		{[]string{"https://gist.github.com"}, ""},
```
rgmz, line 745: Technically these should never occur because there's an explicit length check earlier.

https://github.com/trufflesecurity/trufflehog/blob/9d4cf87c020a38f479daf6dab00d7ab970f13b35/pkg/sources/github/github.go#L1114-L1116

https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1496920923

rosecodym (approved):

PR #2625, opened 2024-03-26
Fix GitHub enumeration & rate-limiting logic

rgmz:
### Description:
This is a follow-up to #2379.

It fixes the following issues:
1. GitHub API calls missing rate-limit handling
2. The fix for https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788 inadvertently resulting in duplicate API calls

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -404,31 +404,50 @@ func (s *Source) enumerate(ctx context.Context, apiEndpoint string) (*github.Cli
 	}
 
 	s.repos = make([]string, 0, s.filteredRepoCache.Count())
+
+RepoLoop:
 	for _, repo := range s.filteredRepoCache.Values() {
+		repoCtx := context.WithValue(ctx, "repo", repo)
+
 		r, ok := repo.(string)
 		if !ok {
-			ctx.Logger().Error(fmt.Errorf("type assertion failed"), "unexpected value in cache", "repo", repo)
+			repoCtx.Logger().Error(fmt.Errorf("type assertion failed"), "Unexpected value in cache")
 			continue
 		}
 
-		_, urlParts, err := getRepoURLParts(r)
-		if err != nil {
-			ctx.Logger().Error(err, "failed to parse repository URL")
-			continue
-		}
+		// Ensure that |s.repoInfoCache| contains an entry for |repo|.
+		// This compensates for differences in enumeration logic between `--org` and `--repo`.
+		// See: https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788
+		if _, ok := s.repoInfoCache.get(r); !ok {
+			repoCtx.Logger().V(2).Info("Caching repository info")
 
-		// Ignore any gists in |s.filteredRepoCache|.
-		// Repos have three parts (github.com, owner, name), gists have two.
-		if len(urlParts) == 3 {
-			// Ensure that individual repos specified in --repo are cached.
-			// Gists should be cached elsewhere.
-			// https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788
-			ghRepo, _, err := s.apiClient.Repositories.Get(ctx, urlParts[1], urlParts[2])
+			_, urlParts, err := getRepoURLParts(r)
 			if err != nil {
-				ctx.Logger().Error(err, "failed to fetch repository")
+				repoCtx.Logger().Error(err, "Failed to parse repository URL")
 				continue
 			}
-			s.cacheRepoInfo(ghRepo)
+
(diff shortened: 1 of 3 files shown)


rosecodym (commented):
This change itself seems fine, but this feels squirrely enough that I feel obliged as a review to ask if there's any low-hanging test fruit we can harvest here. (The answer might be "no," but I want to check.)

rgmz:
> This change itself seems fine, but this feels squirrely enough that I feel obliged as a review to ask if there's any low-hanging test fruit we can harvest here. (The answer might be "no," but I want to check.)

I had a bit more time so I tweaked one of the tests to account for this. It could probably be better written.

rosecodym (dismissed):
@joeleonjr has identified a regression cased by #2379 that is affected by this pr, so i'm holding it up until he can write up a proper bug report so that we can figure out what's going on.

rosecodym:
According to @joeleonjr, this PR changes one of the two errors he reported in #2640. With this PR applied:
> the second one has the same error. The first one has a new error! `Unable to cache repository info	{"source_manager_worker_id": "ATqG1", "repo": "https://gist.github.com/274463.git", "error": "missing cached info for gist: https://gist.github.com/274463.git"}`

@rgmz could you take a look at whether we can fix the regression here? If not, I think it would be good to take a step back and figure out the best path forward. (Which might be merging this PR as-is and opening a new one to fix the regression.)

rgmz:
@joeleonjr try with the latest commit

joeleonjr:
This type of Gist URL works now: `https://gist.github.com/274463.git`, but the longer one with a username still errors: `failed to scan repository	{"error": "no repoInfo for URL: https://gist.github.com/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb.git"}`

rgmz:
It _seems_ like GitHub normalizes the clone URL to `https://gist.github.com/<id>.git`, regardless of whether you do `https://gist.github.com/627e15a45a596068ce8bfef3fd05ccdb.git` or `https://gist.github.com/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb.git`.

The logic caches the Gist with the URL returned by GitHub (`https://gist.github.com/627e15a45a596068ce8bfef3fd05ccdb.git`), but attempts to fetch it with the user-provided URL (`https://gist.github.com/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb.git`).
https://github.com/rgmz/trufflehog/blob/62ca9bb92648fa3f5b229a1866da6bd67a011960/pkg/sources/github/repo.go#L300-L310

rosecodym (approved):
@rgmz If I'm correct that this is a strict improvement over what's currently out there - and that you've updated the tests to include the gist case that you fixed - I'm inclined to get this in and work on the second regression as a separate work item. Any objections?

mcastorina (approved):

rgmz:
No, that makes sense to me.

joeleonjr:
One thing to note: although running `trufflehog github --repo https://gist.github.com/274463.git` works, if you exclude the .git (which I imagine lots of users might do, since it's not in the URL bar that they'd be copying from), it throws a different error:

```
2024-03-29T09:24:18-04:00	error	trufflehog	invalid repository	{"repo": "https://gist.github.com/274463", "error": "Github repo appears to be missing the repo name. Org: \"274463\" Repo url: \"https://gist.github.com/274463\"", "errorVerbose": "Github repo appears to be missing the repo name. Org: \"274463\" Repo url: \"https://gist.github.com/274463\"
```

rgmz:
> One thing to note: although running `trufflehog github --repo https://gist.github.com/274463.git` works, if you exclude the .git (which I imagine lots of users might do, since it's not in the URL bar that they'd be copying from), it throws a different error:
> 
> ```
> 2024-03-29T09:24:18-04:00	error	trufflehog	invalid repository	{"repo": "https://gist.github.com/274463", "error": "Github repo appears to be missing the repo name. Org: \"274463\" Repo url: \"https://gist.github.com/274463\"", "errorVerbose": "Github repo appears to be missing the repo name. Org: \"274463\" Repo url: \"https://gist.github.com/274463\"
> ```

@joeleonjr this appears to be an issue with `giturl.go` and, as far as I can tell, existed prior to #2379.

If the url includes `.git` it assumes that it's well-formatted and returns early.
https://github.com/trufflesecurity/trufflehog/blob/0d3023fe746163edf4a33c1cca8c9c74e6f94fd6/pkg/giturl/giturl.go#L70-L72

Without the `.git` suffix, it fails validation.
https://github.com/trufflesecurity/trufflehog/blob/0d3023fe746163edf4a33c1cca8c9c74e6f94fd6/pkg/giturl/giturl.go#L94-L96

Issue #2622, opened 2024-03-25
Proposed update to github action


jingles:
## Description
If the trufflehog action.yml had an output status, the user could programmatically determine whether the job exited due to a secret detection or some other problem.  This can be helpful to prevent false alerting when (for example) the trufflehog docker image is temporarily unavailable and we exit with some status other than 183.

### Preferred Solution
I've already done a simple version of this on my own and intend to submit a PR.  basically, just use `trap` combined with an output variable.

### Additional Context
Anyone else think this would be handy?  I'll likely submit the PR either way 🤷‍♂️...thanks for reading!

jingles:
I have a branch ready to go, but don't have access to push... how do I go about this?  I signed the CLA... thanks

jingles:
@skeweredlogic noticed you were active on the actions side of things, would appreciate your thoughts.  thanks

hassan8367:
hassan8367@outlook.sa اضافه البريد الإلكتروني للحساب وربطه للحساب hassan1666a@gmail.com

PR #2633, opened 2024-03-27
trap and set an output exit code for use by the caller

jingles:
### Description:
Adds an exit code output to the github action.

https://github.com/trufflesecurity/trufflehog/issues/2622

### Checklist:


rosecodym:
@ahrav @joeleonjr am I correct in recalling that this is either something we _intended_ to do or something that we _specifically avoided doing_ for some reason?

joeleonjr:
> @ahrav @joeleonjr am I correct in recalling that this is either something we _intended_ to do or something that we _specifically avoided doing_ for some reason?

I'm not sure why we couldn't make the error code available as part of the job `outputs`. I tested a basic workflow with this change and it doesn't seem to have any impact if you're not using the job `outputs`. Assuming it's helpful for the author and others, it seems like an easy change. Only possible suggestion or consideration is the name for the job `id`. I'm not sure if that could break anything for other users.

kashifkhan0771:
@jingles can you please resolve the conflicts

PR #2647, opened 2024-03-29
[refactor] - GCP detector

ahrav:
### Description:
This PR refactors the validation logic for GCP service account keys. Instead of directly calling the Google API to verify the credentials using `google.CredentialsFromJSON`, we now perform the validation by comparing the public keys.

Key changes:
- Renamed the validation function to `isValidGCPServiceAccountKey` to better reflect its purpose.
- Updated the function to retrieve the public key certificate from the `ClientX509CertURL` field of the `gcpKey` struct.
- Extracted the public key from the retrieved certificate and compared it with the public key derived from the `PrivateKey` field of the `gcpKey` struct.
- If the public keys match, the function returns `true`, indicating a valid service account key. Otherwise, it returns `false`.
- Updated the function comment to provide a clear explanation of the validation logic.

Benefits:
- Improved efficiency by avoiding the need to make a direct API call to verify the credentials.
- Enhanced security by validating the service account key based on the public key comparison.
- Increased clarity and readability of the code by renaming the function and updating the comment.

### Checklist:


rgmz:
Wouldn't the API call still be necessary? A genuine key does not necessarily equate to a *valid* key (e.g., relocation.)

ahrav:
> Wouldn't the API call still be necessary? A genuine key does not necessarily equate to a _valid_ key (e.g., revocation.)

I totally understand your concern about the necessity of making an API call to verify the validity of the credential, considering factors like revocation, I had the same concerns initially.

I have actually tested this scenario by disabling the service account, and I can confirm that the approach of comparing the public keys works as expected. When a service account is expired or disabled, the request to fetch the certificate from the ClientX509CertURL returns a 404 status code. This indicates that the certificate is no longer accessible, and the service account is correctly considered invalid.

By relying on the availability of the certificate at the ClientX509CertURL, we can effectively validate the status of the service account. If the certificate is not found (404 status code), it means the service account is no longer valid, regardless of the public key comparison.

This approach aligns with the recommendation provided to us by the Google team. They suggested that if we can successfully retrieve the certificate and find a matching public key corresponding to the private key in the credential, it can be considered live and verified.

Let ma update the comments to document that.

pkg/detectors/gcp/gcp.go
```diff
@@ -120,6 +115,104 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+// fetchCert downloads the certificate from the specified URL and returns its PEM-encoded string.
+func fetchCert(ctx context.Context, client *http.Client, url string) (string, bool) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return "", false
+	}
+
+	res, err := client.Do(req)
+	if err != nil || res.StatusCode != http.StatusOK {
+		return "", false
```
rosecodym, line 159: Can we support indeterminacy by checking for non-404 error codes here?

rosecodym (commented):
I think I see a way to support indeterminacy (commented inline) - is that possible?

(And out of curiosity, what prompted this change?)

ahrav:
> I think I see a way to support indeterminacy (commented inline) - is that possible?
> 
> (And out of curiosity, what prompted this change?)

The team at Google requested we use this approach instead.

pkg/detectors/gcp/gcp.go
```diff
@@ -90,41 +103,144 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		// The service account is expired or disabled.
+		return "", false, nil
+	default:
+		return "", false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
+
+}
+
+// extractPublicKeyFromCert decodes the PEM-encoded certificate to extract the RSA public key.
+func extractPublicKeyFromCert(certPEM string) (*rsa.PublicKey, bool) {
```
rgmz, line 201: Is it worth handling these errors? The input is coming from a trusted API in a known format; I'd think that any failures would be highly unusual and something worth looking at.

pkg/detectors/gcp/gcp.go
```diff
@@ -90,41 +103,144 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	if !ok {
+		return false, nil
+	}
+
+	publicKey, ok := extractPublicKeyFromCert(certPEM)
+	if !ok {
+		return false, nil
+	}
+
+	privateKey, ok := extractPrivateKey(key.PrivateKey)
```
rgmz, line 153: Would it make sense to run this prior to the `verify` block? That way, if the key isn't valid (e.g., `---- BEGIN PRIVATE KEY ----- \n<YOUR-KEY-HERE>\n---- END PRIVATE KEY -----`) it doesn't need to make a network request. It would also filter out invalid results if verification is disabled/unverified results are being shown.

pkg/detectors/gcp/gcp.go
```diff
@@ -90,41 +103,144 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 		credBytes, _ := json.Marshal(creds)
```
dylanTruffle, line 104: Do we need to actually capture the whole json blob and marshal the json? One issue with GCP keys is they're so big sometimes they get cut off in scanning windows. Maybe now that we've updated verifier logic, we could get away with just a capture group around the private key, and a capture group around the x509 URL

rgmz: Technically all that's required is the `private_key` and `client_email` (`name@project_id.iam.gserviceaccount.com`). The X509 URL is just `https://www.googleapis.com/robot/v1/metadata/x509/$client_email`

dylanTruffle: Either the x509 url or the client_email work as capture groups.

PR #2653, opened 2024-04-02
Improve handling of Gist URLs

### LINKED ISSUE #2640: Gist Scanning Errors
### TruffleHog Version
```trufflehog 3.71.1```

### Trace Output
[Trace back ](https://gist.github.com/joeleonjr/b2885eff495787216fac2d363d12b177)

### Expected Behavior
Gists should be scanned just like any other repository on GitHub. 

### Actual Behavior
Depending on the Gist URL structure, one of two different errors spits out:

URLs with the username (https://gist.github.com/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb.git)
```
Failed to fetch repository	{"source_manager_worker_id": "Xwjd7", "repo": "https://gist.github.com/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb.git", "error": "GET https://api.github.com/repos/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb: 404 Not Found []"}
```

URLs without the username (https://gist.github.com/274463.git)
```
Unable to cache repository info	{"source_manager_worker_id": "ATqG1", "repo": "https://gist.github.com/274463.git", "error": "missing cached info for gist: https://gist.github.com/274463.git"}
```

### Steps to Reproduce
Run the following commands:

`trufflehog github --repo https://gist.github.com/274463.git`
`trufflehog github --repo https://gist.github.com/raccoons-bot/627e15a45a596068ce8bfef3fd05ccdb.git`

## Environment
 * OS: OSX
 * Version 14.2.1

## Additional Context
I believe this was introduced in PR #2379 .

### PR DESCRIPTION
**Pull request by rgmz**
### Description:
This is a follow-up to #2625. It fixes #2640 (at least based on my testing) and should allow scanning of gists with GHES, which was highlighted in https://github.com/trufflesecurity/trufflehog/issues/2640#issuecomment-2025398660.

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -427,8 +427,7 @@ RepoLoop:
 				continue
 			}
 
-			if strings.EqualFold(urlParts[0], "gist.github.com") ||
-				(len(urlParts) == 4 && strings.EqualFold(urlParts[1], "gist")) {
+			if isGistUrl(urlParts) {
 				// Cache gist info.
 				for {
 					gistID := extractGistID(urlParts)
@@ -1109,7 +1108,7 @@ func (s *Source) scanComments(ctx context.Context, repoPath string, repoInfo rep
 		return err
 	}
 
-	if s.includeGistComments && urlParts[0] == "gist.github.com" {
+	if s.includeGistComments && isGistUrl(urlParts) {
 		return s.processGistComments(ctx, urlString, urlParts, repoInfo, chunksChan)
 	} else if s.includeIssueComments || s.includePRComments {
 		return s.processRepoComments(ctx, repoInfo, chunksChan)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/sources/github/github.go [hunk 1]
```diff
@@ -1128,13 +1132,29 @@ func getRepoURLParts(repoURL string) (string, []string, error) {
...
+	// Validate
+	switch len(urlParts) {
+	case 2:
+		if !strings.EqualFold(urlParts[0], "gist.github.com") {
+			err = fmt.Errorf("failed to parse repository or gist URL (%s): 2 segments are only expected if the host is 'gist.github.com'", urlString)
+		}
+	case 3:
+		// Do nothing
+	case 4:
+		if !(!strings.EqualFold(urlParts[0], "github.com") && strings.EqualFold(urlParts[1], "gist")) {
```
mcastorina, line 1146: I think there's an extra `!` in there and it should be:

```go
  !(strings.EqualFold(urlParts[0], "github.com") && strings.EqualFold(urlParts[1], "gist"))
// ^^ removed this one
```

A `gist` segment is only expected for the `github.com` host, right?

rgmz: This is difficult to grok in its current state, I'll need to change it — or at least add a comment. 😵 

> I think there's an extra `!` in there and it should be:
> 
> ```go
>   !(strings.EqualFold(urlParts[0], "github.com") && strings.EqualFold(urlParts[1], "gist"))
> // ^^ removed this one
> ```
> 
> A `gist` segment is only expected for the `github.com` host, right?

It's actually the opposite. GitHub uses the subdomain `gist.github.com` for gists, and although `github.com/gist` redirects -> `gist.github.com`, it isn't a valid gist URL (e.g., https://github.com/gist/280192b0523099d0614a95f579d99aa9 and https://github.com/gist/rgmz/280192b0523099d0614a95f579d99aa9). However, `/gist/<id>` can be valid for instances of GitHub Enterprise Server (e.g., https://github.company.org/gist/nat/5fdbb7f945d121f197fb074578e53948.git).

The logic is meant to accept a `gist` segment for any domain _except_ `github.com`.

rgmz: I'm not sure if @joeleonjr has anything to add about Gist URLs.

rgmz: > This is difficult to grok in its current state, I'll need to change it — or at least add a comment. 😵

I did both. Let me know if that looks better.

pkg/sources/github/github.go
```diff
@@ -1128,13 +1132,34 @@ func getRepoURLParts(repoURL string) (string, []string, error) {
...
+			err = fmt.Errorf("failed to parse repository or gist URL (%s): 4 path segments are only expected if the host isn't 'github.com' and the path starts with 'gist'", urlString)
+		}
+	default:
+		err = fmt.Errorf("invalid repository or gist URL (%s): length of URL segments should be between 2 and 4, not %d", urlString, len(urlParts))
 	}
 
-	return urlString, splitURL, nil
+	if err != nil {
+		return "", []string{}, err
+	} else {
```
ahrav, line 1160: suggestion: Since the if block returns we can remove the else block and reduce the indentation.

pkg/sources/github/github.go
```diff
@@ -1128,13 +1132,34 @@ func getRepoURLParts(repoURL string) (string, []string, error) {
...
+		if !strings.EqualFold(urlParts[1], "gist") || (strings.EqualFold(urlParts[0], "github.com") && strings.EqualFold(urlParts[1], "gist")) {
+			err = fmt.Errorf("failed to parse repository or gist URL (%s): 4 path segments are only expected if the host isn't 'github.com' and the path starts with 'gist'", urlString)
+		}
+	default:
+		err = fmt.Errorf("invalid repository or gist URL (%s): length of URL segments should be between 2 and 4, not %d", urlString, len(urlParts))
 	}
 
-	return urlString, splitURL, nil
+	if err != nil {
+		return "", []string{}, err
```
ahrav, line 1159: question: Could you elaborate on the choice to return `[]string{}` (an empty slice) instead of `nil` in the error case?

- Is this to indicate a specific state to the caller (e.g., no items found)?
- Could returning `nil` simplify the error handling on the caller's side?

If returning an empty slice is the intended behavior, could we consider adding a brief comment explaining this decision for future reference.

ahrav (approved):
1 non-blocking question, otherwise LGTM! Thanks for updating this, the switch make this a lot easier to grok.

pkg/sources/github/github.go
rgmz: No particular reason. I can change it to `nil`.

PR #2669, opened 2024-04-04
Revert GitHub refactor (#2379) (including followups)

rosecodym:
### Description:
The GitHub source refactor in #2379 reduced the overall number of requests to GitHub, but compressed them into a smaller period of time, resulting in more rate limit pressure that is causing noticeable degradation in usability in some cases. We've decided to back it out until we can figure out how to make improvements without this interim usability degradation. That PR had multiple followup PRs (#2608, #2614, #2625), which this PR also reverts.

Note that this doesn't revert the `go-github` version bump from 57 to 61 that has since occurred.

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -404,59 +404,31 @@ func (s *Source) enumerate(ctx context.Context, apiEndpoint string) (*github.Cli
 	}
 
 	s.repos = make([]string, 0, s.filteredRepoCache.Count())
-
-RepoLoop:
 	for _, repo := range s.filteredRepoCache.Values() {
-		repoCtx := context.WithValue(ctx, "repo", repo)
-
 		r, ok := repo.(string)
 		if !ok {
-			repoCtx.Logger().Error(fmt.Errorf("type assertion failed"), "Unexpected value in cache")
+			ctx.Logger().Error(fmt.Errorf("type assertion failed"), "unexpected value in cache", "repo", repo)
 			continue
 		}
 
-		// Ensure that |s.repoInfoCache| contains an entry for |repo|.
-		// This compensates for differences in enumeration logic between `--org` and `--repo`.
-		// See: https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788
-		if _, ok := s.repoInfoCache.get(r); !ok {
-			repoCtx.Logger().V(2).Info("Caching repository info")
+		_, urlParts, err := getRepoURLParts(r)
+		if err != nil {
+			ctx.Logger().Error(err, "failed to parse repository URL")
+			continue
+		}
 
-			_, urlParts, err := getRepoURLParts(r)
+		// Ignore any gists in |s.filteredRepoCache|.
+		// Repos have three parts (github.com, owner, name), gists have two.
+		if len(urlParts) == 3 {
+			// Ensure that individual repos specified in --repo are cached.
+			// Gists should be cached elsewhere.
+			// https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788
+			ghRepo, _, err := s.apiClient.Repositories.Get(ctx, urlParts[1], urlParts[2])
 			if err != nil {
-				repoCtx.Logger().Error(err, "Failed to parse repository URL")
+				ctx.Logger().Error(err, "failed to fetch repository")
 				continue
 			}
-
-			if strings.EqualFold(urlParts[0], "gist.github.com") {
-				// Cache gist info.
-				for {
-					gistID := extractGistID(urlParts)
-					gist, _, err := s.apiClient.Gists.Get(repoCtx, gistID)
-					if s.handleRateLimit(err) {
-						continue
-					}
-					if err != nil {
-						repoCtx.Logger().Error(err, "Failed to fetch gist")
-						continue RepoLoop
-					}
-					s.cacheGistInfo(gist)
-					break
-				}
-			} else {
-				// Cache repository info.
-				for {
-					ghRepo, _, err := s.apiClient.Repositories.Get(repoCtx, urlParts[1], urlParts[2])
-					if s.handleRateLimit(err) {
-						continue
-					}
-					if err != nil {
-						repoCtx.Logger().Error(err, "Failed to fetch repository")
-						continue RepoLoop
-					}
-					s.cacheRepoInfo(ghRepo)
-					break
-				}
-			}
+			s.cacheRepoInfo(ghRepo)
 		}
 		s.repos = append(s.repos, r)
 	}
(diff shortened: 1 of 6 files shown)


ahrav (approved):

rosecodym:
@rgmz we've decided to pull the trigger on this. I believe this is a straightforward stack of reverts but please speak up if anything looks amiss.

pkg/sources/github/github.go
```diff
@@ -1091,74 +1126,44 @@ func (s *Source) setProgressCompleteWithRepo(index int, offset int, repoURL stri
...
-
 	options := &github.ListOptions{
 		PerPage: defaultPagination,
 		Page:    initialPage,
 	}
 	for {
 		comments, _, err := s.apiClient.Gists.ListComments(ctx, gistID, options)
 		if s.handleRateLimit(err) {
-			continue
+			break
```
rgmz, line 1160: This should be `continue`.

pkg/sources/github/github.go
```diff
@@ -1259,16 +1259,16 @@ func (s *Source) processIssues(ctx context.Context, repoInfo repoInfo, chunksCha
 	}
 
 	for {
-		issues, _, err := s.apiClient.Issues.ListByRepo(ctx, repoInfo.owner, repoInfo.name, bodyTextsOpts)
+		issues, _, err := s.apiClient.Issues.ListByRepo(ctx, info.owner, info.repo, bodyTextsOpts)
 		if s.handleRateLimit(err) {
-			continue
+			break
```
rgmz, line 1264: This should be `continue`.

pkg/sources/github/github.go
```diff
@@ -1331,60 +1292,29 @@ func (s *Source) processIssueComments(ctx context.Context, repoInfo repoInfo, ch
 	}
 
 	for {
-		issueComments, _, err := s.apiClient.Issues.ListComments(ctx, repoInfo.owner, repoInfo.name, allComments, issueOpts)
+		issueComments, _, err := s.apiClient.Issues.ListComments(ctx, info.owner, info.repo, allComments, issueOpts)
 		if s.handleRateLimit(err) {
-			continue
+			break
```
rgmz, line 1297: This should be `continue`.

pkg/sources/github/github.go
rgmz, line 1264: This should be continue.

pkg/sources/github/github.go
rgmz, line 1160: This should be continue.

pkg/sources/github/github.go
```diff
@@ -1396,15 +1326,16 @@ func (s *Source) processPRs(ctx context.Context, repoInfo repoInfo, chunksChan c
 	}
 
 	for {
-		prs, _, err := s.apiClient.PullRequests.List(ctx, repoInfo.owner, repoInfo.name, prOpts)
+		prs, _, err := s.apiClient.PullRequests.List(ctx, info.owner, info.repo, prOpts)
 		if s.handleRateLimit(err) {
-			continue
+			break
```
rgmz, line 1331: This should be continue.

pkg/sources/github/github.go
```diff
@@ -1428,15 +1359,16 @@ func (s *Source) processPRComments(ctx context.Context, repoInfo repoInfo, chunk
 	}
 
 	for {
-		prComments, _, err := s.apiClient.PullRequests.ListComments(ctx, repoInfo.owner, repoInfo.name, allComments, prOpts)
+		prComments, _, err := s.apiClient.PullRequests.ListComments(ctx, info.owner, info.repo, allComments, prOpts)
 		if s.handleRateLimit(err) {
-			continue
+			break
```
rgmz, line 1364: This should be continue.

pkg/sources/github/github.go
```diff
@@ -1569,3 +1607,11 @@ func (s *Source) scanTarget(ctx context.Context, target sources.ChunkingTarget,
 
 	return common.CancellableWrite(ctx, chunksChan, chunk)
 }
+
+func removeURLAndSplit(url string) []string {
```
rgmz, line 1611: I think this would result in a regression compared to `getRepoURLParts`. e.g., it doesn't work with `http://` urls.

rgmz (changes requested):

rgmz:
> @rgmz we've decided to pull the trigger on this. I believe this is a straightforward stack of reverts but please speak up if anything looks amiss.

I did a quick scan through and left comments on any obvious issues.

> The GitHub source refactor in https://github.com/trufflesecurity/trufflehog/pull/2379 reduced the overall number of requests to GitHub, but compressed them into a smaller period of time, resulting in more rate limit pressure that is causing noticeable degradation in usability in some cases.

This is surprising to hear. I'm not aware of any interaction that would result in more traffic over a smaller period of time, but I'm also not aware of all the use cases.

pkg/sources/github/github.go
rosecodym: If I'm reading the original refactor commit correctly, this is an appropriate change for a revert. Did I miss some consideration?

rgmz: The revert was correct. However, some of these changes are fixes for things that didn't work prior. I'm just pointing out things that need to be re-fixed.

E.g., using break instead of continue means it will never retry these API requests.

rosecodym:
> > @rgmz we've decided to pull the trigger on this. I believe this is a straightforward stack of reverts but please speak up if anything looks amiss.
> 
> I did a quick scan through and left comments on any obvious issues.
> 
> > The GitHub source refactor in #2379 reduced the overall number of requests to GitHub, but compressed them into a smaller period of time, resulting in more rate limit pressure that is causing noticeable degradation in usability in some cases.
> 
> This is surprising to hear. I'm not aware of any interaction that would result in more traffic over a smaller period of time, but I'm also not aware of all the use cases.

[The refactor added a new API call](https://github.com/trufflesecurity/trufflehog/pull/2379/files#diff-6ed11f0fcaaa1c13861a19fb775375c72ba30bc9c52b9d6ee2733f08388ff8e1R420) that's invoked for each repository enumerated. When many repositories are enumerated, this results in show-stopping rate limiting.

rgmz:
> [The refactor added a new API call](https://github.com/trufflesecurity/trufflehog/pull/2379/files#diff-6ed11f0fcaaa1c13861a19fb775375c72ba30bc9c52b9d6ee2733f08388ff8e1R420) that's invoked for each repository enumerated. When many repositories are enumerated, this results in show-stopping rate limiting.

Ah, I see. That was unintentional and should have been fixed in #2625.

The change definitely could have used another pass with a fresh pair of eyes before being merged. I think part of the issue was that I hadn't touched it for quite a while prior to it being merged.

pkg/sources/github/github.go
rosecodym: Ah, so the original refactor also fixed a bug? Is that fix something we can independently deploy?

rosecodym:
Ah, got it. With that pointer we dug a little further and it appears that we were just bitten by some _very_ unfortunate release timing - the rate limit issue has, in fact, been resolved. Thanks for your patience and your eyes on this @rgmz!

PR #2677, opened 2024-04-05
Create Docker registry auth detector

rgmz:
### Description:
This implements a detector to find Docker registry credentials, inspired by [this suggestion](https://github.com/gitleaks/gitleaks/issues/1301) from @bgoareguer.

The current code works, but is a bit messy and requires more feedback + testing.

Questions/TODO
- [x] ~Should this also match and decode Kubernetes `.dockerconfigjson` results, or rely on the base64 decoder?~
- [x] Add or remove logging? IMO, there are a few potential failures (e.g., `json.Unmarshal`) that are worth noting, as it could be indicative of a bug.
- [x] Test against [a live `gcr.io` credential](https://github.com/airbnb/autoscaler/blob/f22b40eab867cbc52bdb15dc8768962e21d22837/vertical-pod-autoscaler/e2e/vendor/k8s.io/kubernetes/test/e2e/common/node/runtime.go#L283C1-L290C2), as that registry [uses base64-encoded GCP service principals as the password](https://github.com/polyaxon/polyaxon/blob/3ef89812e77819f3f07099d4cf65169f45d3c46d/site/integrations/gcr.md?plain=1#L53). It's possible that this doesn't work with the current logic (e.g., encoded newlines.) 
- [x] Add username to extradata
- [x] ~Handle username for GCR credentials? (It's a static `_json_key`, the real username is in the auth)~
- [x] ~Are there any other potential formats this data gets stored in? (e.g., YAML)~
- [x] Should this handle encoded `\n` as well as literal?
- [x] Update comments to add relevant pointers to the spec https://distribution.github.io/distribution/spec/auth/token/
- [x] Handle `docker.io` which is a special case (https://stackoverflow.com/a/68654659) 
- [x] Fix incorrect path handling in certain circumstances
  ```
  Found unverified result 🐷🔑❓
  Verification issue: unexpected HTTP response status 404 for 'https://index.docker.io/v1//v2/'
  Detector Type: Docker
  Decoder Type: PLAIN
  Link: https://github.com/lyft/docker/blob/34679e568a22b4f35ff8460f3b5b7bf7089df818/cliconfig/config_test.go#L358
  ```

Future work?
- [ ] Detect the legacy auth config format (https://github.com/openshift/machine-config-operator/blob/82011335dbdd3d4c869b959d6048a3fba7742e47/pkg/controller/build/helpers_test.go#L47, https://github.com/moby/moby/blob/145a73a36c171b34c196ad780e699b154ddf47b5/docs/api/v1.24.md?plain=1#L1767)
- [ ] Add basic registry name validation (https://github.com/moby/moby/blob/145a73a36c171b34c196ad780e699b154ddf47b5/registry/config_test.go#L358)
- [x] Better handling of potentially false positive 200 responses (https://github.com/HubSpot/kubernetes-client/blob/e57f1ccdd9066343cb5547044e76f355025fe9e6/kubernetes-client-api/src/test/java/io/fabric8/kubernetes/client/internal/KubernetesResourceUtilTest.java#L239), especially if the response isn't json 
- [ ] Handle registries indiscriminately returning 200 even for invalid credentials? (see https://github.com/containers/podman issue 22400)
- [x] Use `json.NewDecoder` instead of `json.Unmarshal` (https://github.com/Azure/ARO-RP/blob/25b8569698b61abd7af6cdefe680fdc1306b001f/pkg/operator/admission/validation/pullsecret/request.json#L1)
- [ ] Check if `base64-encoded auth does not match source` is a legitimate issue (probably not, seems to be caused by base64 decoder)

### Checklist:

### DIFF
--- a/pkg/detectors/docker/docker_auth_config.go
+++ b/pkg/detectors/docker/docker_auth_config.go
@@ -1,0 +1,316 @@
+package docker
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/go-logr/logr"
+	regexp "github.com/wasilibs/go-re2"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	logContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct {
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ interface {
+	detectors.Detector
+	detectors.MaxSecretSizeProvider
+} = (*Scanner)(nil)
+
+func (s Scanner) Type() detectorspb.DetectorType {
+	return detectorspb.DetectorType_Docker
+}
+
+func (s Scanner) Description() string {
+	return "Docker credentials can be used to pull images from private registries."
+}
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{`"auths"`, `\"auths\`}
+}
+
+func (s Scanner) MaxSecretSize() int64 {
+	return 4096
+}
+
+var (
+	keyPat          = regexp.MustCompile(`{(?:\s|\\+[nrt])*\\*"auths\\*"(?:\s|\\+t)*:(?:\s|\\+t)*{(?:\s|\\+[nrt])*\\*"(?i:https?:\/\/)?[a-z0-9\-.:\/]+\\*"(?:\s|\\+t)*:(?:\s|\\+t)*{(?:(?:\s|\\+[nrt])*\\*"(?i:auth|email|username|password)\\*"\s*:\s*\\*".*\\*"\s*,?)+?(?:\s|\\+[nrt])*}(?:\s|\\+[nrt])*}(?:\s|\\+[nrt])*}`)
+	escapedReplacer = strings.NewReplacer(
+		`\n`, "",
+		`\r`, "",
+		`\t`, "",
+		`\\`, ``,
+		`\"`, `"`,
+	)
+
+	// Common false-positives used in examples.
+	exampleRegistries = map[string]struct{}{
+		"https://index.docker.io/v1/":       {}, // https://github.com/moby/moby/blob/34679e568a22b4f35ff8460f3b5b7bf7089df818/cliconfig/config_test.go#L259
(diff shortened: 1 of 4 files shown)


pkg/detectors/docker/docker_auth_config.go
```diff
@@ -0,0 +1,319 @@
...
+type dockerAuth struct {
+	Auth     string `json:"auth"`
+	Username string `json:"username"`
+	Password string `json:"password"`
+	Email    string `json:"email"`
+}
+
+// parseBasicAuth handles cases where configs can have `username` and `password` but no `auth`,
+// or vice-versa.
+func parseBasicAuth(logger logr.Logger, auth dockerAuth) (string, string, string) {
```
ahrav, line 254: question: Do you think we could achieve the same result by using a few more conditionals instead of relying on `goto`?

Maybe something like the following? (I think these should be identical)
```
// parseBasicAuth handles configs that may supply either separate
// username/password fields, an 'auth' field, or both.
//
func parseBasicAuth(logger logr.Logger, auth dockerAuth) (string, string, string) {
	// ── Start with the explicit fields (if any) ──────────────────────────────
	user := auth.Username
	pass := auth.Password

	// ── If an encoded "auth" string is present, try to decode & overwrite ───
	if auth.Auth != "" {
		if data, err := base64.StdEncoding.DecodeString(auth.Auth); err == nil {
			if parts := strings.SplitN(string(data), ":", 2); len(parts) == 2 {
				if (user != "" && parts[0] != user) || (pass != "" && parts[1] != pass) {
					logger.V(2).Info("WARNING: creds contain multiple, conflicting usernames or passwords")
				}
				user, pass = parts[0], parts[1]
			} else {
				logger.V(2).Info("Skipping invalid parts", "length", len(parts), "parts", parts)
			}
		}
	}

	// ── Nothing usable? Bail early ──────────────────────────────────────────
	if user == "" && pass == "" {
		return "", "", ""
	}

	// ── Canonical basic‑auth string & consistency check ─────────────────────
	basic := base64.StdEncoding.EncodeToString([]byte(user + ":" + pass))
	if auth.Auth != "" && basic != auth.Auth {
		logger.Error(fmt.Errorf("base64‑encoded auth does not match source"), "failed to parse auths JSON")
	}

	return user, pass, basic
}
```

ahrav (approved):
Small question/suggestion, otherwise lgtm! Thanks @rgmz

pkg/detectors/docker/docker_auth_config.go
rgmz: I was musing about this locally, though I see you've merged this (thanks!).

If I may opine for a moment: I feel that `goto` is a misunderstood and unfairly maligned construct. I'd argue the classic "Go To Statement Considered Harmful" is less relevant in modern languages,  which lack the ability to jump _anywhere_, and what's left is a niche but useful feature for control flow (e.g., the Go stdlib makes extensive use of `goto`).

Personally, I think `goto` makes the code in that section straightforward and easy to follow, compared to heavy use of nesting and inline declarations. What do you think?

PR #2701, opened 2024-04-14
[refactor] - Refactor Archive Handling Logic - Part 1: Remove `SpecializedHandler` and Simplify File Handling Interface

ahrav:
### Description:
This PR is the first in a series of four PRs aimed at refactoring our archive handling logic. The main changes in this PR include:
1. Removal of the `SpecializedHandler`:
   - The `SpecializedHandler` was previously used to extract .deb and .rpm archives by shelling out to handle archive types not supported by the archiver library.
   - The extraction logic for .deb and .rpm archives will be replaced with native library implementations in the upcoming PRs.
2. Simplification of the file handling interface:
   - The file handling interface has been streamlined to a single method: `HandleFile(ctx logContext.Context, input io.Reader) (chan []byte, error)`.
   - This change allows for separate implementations to process different file types.
   - The `DefaultHandler` implementation, which leverages the archiver package, should handle most archive file types.
   - File types not supported by the archiver package will require their own custom implementations for file handling, such as `RPMHandler`, `ARHandler`, etc.
Subsequent PRs will address the currently unimplemented code for:
- [rpm handling](https://github.com/trufflesecurity/trufflehog/pull/2702)
- [deb/ar handling](https://github.com/trufflesecurity/trufflehog/pull/2703)
This PR lays the groundwork for the upcoming changes by removing the `SpecializedHandler` and simplifying the file handling interface. The goal is to improve the maintainability and extensibility of our archive handling logic by using native libraries for specific archive types and providing a more flexible file handling interface.
Please note that this PR is part of a larger refactoring effort, and the subsequent PRs will build upon the changes introduced here to complete the implementation of the new archive handling logic.

### Checklist:


pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// extractorHandler creates a closure that handles individual files extracted by an archiver.
+// It logs the extraction, checks for cancellation, and decides whether to skip the file based on its name or type,
+// particularly for binary files if configured to skip. If the file is not skipped, it recursively calls openArchive
+// to handle nested archives or to continue processing based on the file's content and depth in the archive structure.
+func (h *DefaultHandler) extractorHandler(archiveChan chan []byte) func(context.Context, archiver.File) error {
+	return func(ctx context.Context, f archiver.File) error {
+		lCtx := logContext.AddLogger(ctx)
+		lCtx.Logger().V(5).Info("Handling extracted file.", "filename", f.Name())
```
rgmz, line 154: Why use `context` and immediately convert it to `trufflehog/v3/pkg/context` instead of accepting `trufflehog/v3/pkg/context`? It seems like the only thing that calls this is `openArchive`, which already uses logContext.

rgmz, line 154: Since `filename` is repeated, what about adding it as a value to the context?

(Pseudo-code, may not compile)
```suggestion
		lCtx := logContext.WithValues(logContext.AddLogger(ctx), "filename", f.Name())
		lCtx.Logger().V(5).Info("Handling extracted file.")
```

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,226 @@
...
+
+		if common.SkipFile(f.Name()) {
+			lCtx.Logger().V(5).Info("skipping file", "filename", f.Name())
+			return nil
+		}
+
+		if h.skipBinaries && common.IsBinary(f.Name()) {
+			lCtx.Logger().V(5).Info("skipping binary file", "filename", f.Name())
+			return nil
+		}
```
rgmz, line 179: Does it make a difference if this is before `fReader, err := f.Open()`?

pkg/handlers/defaulthandler.go
ahrav: I believe it's because this method is used as an argument to `archive.Extract` (which is a third party lib) and that doesn't not accept our custom context. [this method](https://github.com/mholt/archiver/blob/de08bfa4c5588130c63dfcede2da42f26abccbb4/interfaces.go#L93)
Which we use [here](https://github.com/trufflesecurity/trufflehog/blob/41030c5b0737f3cf4ca150190cbc761a86a351d9/pkg/handlers/defaulthandler.go#L141)

pkg/handlers/defaulthandler.go
ahrav: nope, i think moving it before likely makes more sense. Thanks for the suggestion.

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,234 @@
...
+		return h.openArchive(lCtx, depth, fReader, archiveChan)
+	}
+}
+
+// handleNonArchiveContent processes files that do not contain nested archives, serving as the final stage in the
+// extraction/decompression process. It reads the content to detect its MIME type and decides whether to skip based
+// on the type, particularly for binary files. It manages reading file chunks and writing them to the archive channel,
+// effectively collecting the final bytes for further processing. This function is a key component in ensuring that all
+// file content, regardless of being an archive or not, is handled appropriately.
+func (h *DefaultHandler) handleNonArchiveContent(ctx logContext.Context, reader io.Reader, archiveChan chan []byte) error {
```
ahrav, line 198: This code was just moved from archive.go

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,234 @@
...
+	default:
+		return fmt.Errorf("unknown archive type: %s", format.Name())
+	}
+}
+
+// extractorHandler creates a closure that handles individual files extracted by an archiver.
+// It logs the extraction, checks for cancellation, and decides whether to skip the file based on its name or type,
+// particularly for binary files if configured to skip. If the file is not skipped, it recursively calls openArchive
+// to handle nested archives or to continue processing based on the file's content and depth in the archive structure.
+func (h *DefaultHandler) extractorHandler(archiveChan chan []byte) func(context.Context, archiver.File) error {
```
ahrav, line 151: This code was just moved from archive.go

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,234 @@
...
+	return archiveChan, nil
+}
+
+var ErrMaxDepthReached = errors.New("max archive depth reached")
+
+// openArchive recursively extracts content from an archive up to a maximum depth, handling nested archives if necessary.
+// It takes a reader from which it attempts to identify and process the archive format. Depending on the archive type,
+// it either decompresses or extracts the contents directly, sending data to the provided channel.
+// Returns an error if the archive cannot be processed due to issues like exceeding maximum depth or unsupported formats.
+func (h *DefaultHandler) openArchive(ctx logContext.Context, depth int, reader io.Reader, archiveChan chan []byte) error {
```
ahrav, line 111: This code was just moved from archive.go

PR #2703, opened 2024-04-14
[refactor] - Refactor Archive Handling Logic

ahrav:
### Description:
This PR refactors our archive handling logic to improve maintainability and performance by replacing shell commands with native library implementations. Here are the key changes:

1. **General Refactoring:**
   - Removed the `SpecializedHandler` used for extracting `.deb` and `.rpm` archives.
   - Streamlined the file handling interface to a single method: `HandleFile(ctx logContext.Context, input io.Reader) (chan []byte, error)`.
   - Introduced the `DefaultHandler` leveraging the `archiver` package to handle most archive types. Custom implementations like `RPMHandler` and `ARHandler` are created for unsupported file types.

2. **RPM Archive Handling:**
   - Implemented `RPMHandler` which uses the `github.com/sassoftware/go-rpmutils` library to handle `.rpm` archives efficiently by extracting the payload natively.

3. **DEB and AR Archive Handling:**
   - Introduced `ARHandler` that utilizes the `pault.ag/go/debian/deb` package for `.deb` and `.ar` archives.
   - Modified the `FileHandler` interface to support `*diskbufferreader.DiskBufferReader` to accommodate the io.ReaderAt interface required by the `pault.ag/go/debian/deb` package.

Note: This PR is part of a larger refactoring effort, and the full benefits of the changes will be realized once all the related PRs are completed.
### Checklist:


rosecodym:
Are you going to close #2701 and #2702?

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,235 @@
...
+	if common.IsDone(ctx) {
+		return ctx.Err()
+	}
+
+	if depth >= maxDepth {
+		return ErrMaxDepthReached
+	}
+
+	format, arReader, err := archiver.Identify("", reader)
+	if errors.Is(err, archiver.ErrNoMatch) && depth > 0 {
```
rosecodym, line 122: What is this depth check for?

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,235 @@
...
+		compReader, err := archive.OpenReader(arReader)
+		if err != nil {
+			return err
+		}
+
+		defer compReader.Close()
+
+		return h.openArchive(ctx, depth+1, compReader, archiveChan)
+	case archiver.Extractor:
+		return archive.Extract(logContext.WithValue(ctx, depthKey, depth+1), arReader, nil, h.extractorHandler(archiveChan))
```
rosecodym, line 142: Isn't `depthKey` an int here? What does that mean as a log key?

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,235 @@
...
+
+// handleNonArchiveContent processes files that do not contain nested archives, serving as the final stage in the
+// extraction/decompression process. It reads the content to detect its MIME type and decides whether to skip based
+// on the type, particularly for binary files. It manages reading file chunks and writing them to the archive channel,
+// effectively collecting the final bytes for further processing. This function is a key component in ensuring that all
+// file content, regardless of being an archive or not, is handled appropriately.
+func (h *DefaultHandler) handleNonArchiveContent(ctx logContext.Context, reader io.Reader, archiveChan chan []byte) error {
+	bufReader := bufio.NewReaderSize(reader, defaultBufferSize)
+	// A buffer of 512 bytes is used since many file formats store their magic numbers within the first 512 bytes.
+	// If fewer bytes are read, MIME type detection may still succeed.
```
rosecodym, line 202: Can you use `determineMimeType` here?

rosecodym (commented):
Am I correct in inferring that this PR still can't handle convoluted nesting patterns, e.g. RPM->standard?

pkg/handlers/defaulthandler.go
ahrav: I believe this configuration controls the maximum depth of nested archives that we support. It gets set [here](https://github.com/trufflesecurity/trufflehog/blob/4a5fbf84178018da2d92aa4d0a236add497da27b/main.go#L66).

I’m fairly confident that I haven’t altered the existing functionality. However, let me know if there are any unintended changes.

pkg/handlers/defaulthandler.go
ahrav: Correct, `depthKey` is an integer, specifically defined as `ctxKey` to facilitate structured access to values in the context.Context. This definition allows us to use depthKey as a log key to track the current depth of archive processing.

For example within `extractorHandler` we do the following:
```
depth := 0
if ctxDepth, ok := ctx.Value(depthKey).(int); ok {
    depth = ctxDepth
}
```

pkg/handlers/defaulthandler.go
ahrav: I considered it, but given this was a lift and shift job, I opted to leave it as is.

pkg/handlers/defaulthandler.go
rosecodym: Does this also mean that the log will include weird numeric kvps though?

pkg/handlers/defaulthandler.go
ahrav: Oh, this occurs because when the depth is greater than 0 and the `archiver.Identify` function fails to identify a format, it indicates that the `openArchive` function was invoked recursively. Under these conditions, the new reader has been successfully extracted/decompressed.

pkg/handlers/defaulthandler.go
ahrav: I'm not clear how that works, @mcastorina could you provide some insights?

mcastorina: non-string keys are not allowed and will be ignored. This message will print along with the invalid keys

```
non-string key argument passed to logging, ignoring all later arguments
```

ahrav: Just to clarify, when you say _ignored_, you are referring to the key getting displayed in the log message?

mcastorina: Yes - the kvp will get added to the context, and it will get added to the logger iff the key is a string.

pkg/handlers/defaulthandler.go
rosecodym: At the very least I think this deserves a comment, because I think it's very non-obvious. But I also think it's worth taking a step back and seeing if there's a more straightforward way to write this. In its current state, `openArchive` expects a reader that represents either archived content or non-archived content that's been retrieved from inside an archive. That's a pretty idiosyncratic precondition, and I find it difficult to reason about. At the very least, would it be possible to lift this check up into the caller so that `openArchive` treats all passed-in readers the same, irrespective of the depth they were extracted at?

ahrav: I agree with your points 100%. My hesitation stems from trying to avoid unnecessary changes in this refactor PR. I’m willing to add a comment to clarify this aspect. However, I suggest handling the check relocation in a separate PR to keep the current one focused and free from new logic. Thoughts?

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,235 @@
...
+// openArchive recursively extracts content from an archive up to a maximum depth, handling nested archives if necessary.
+// It takes a reader from which it attempts to identify and process the archive format. Depending on the archive type,
+// it either decompresses or extracts the contents directly, sending data to the provided channel.
+// Returns an error if the archive cannot be processed due to issues like exceeding maximum depth or unsupported formats.
+func (h *DefaultHandler) openArchive(ctx logContext.Context, depth int, reader io.Reader, archiveChan chan []byte) error {
+	if common.IsDone(ctx) {
+		return ctx.Err()
+	}
+
+	if depth >= maxDepth {
```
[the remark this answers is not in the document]
ahrav: Let me find out what happens if I change this 😅

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,245 @@
+package handlers
```
bill-rich, line 1: Long-term I think this should be split into `DefaultHandler` and `ArchiveHandler`. I know the name change was made to minimize changes, but leaving it as `archive.go` for this PR would make it a lot easier to see and comment what was changed. Also name should be "Default" since it's part of the handlers package already.

pkg/handlers/arhandler.go
```diff
@@ -0,0 +1,63 @@
+package handlers
```
bill-rich, line 1: Think the naming should be `AR` since it is already part of the `handlers` package.

pkg/handlers/defaulthandler.go
```diff
@@ -0,0 +1,245 @@
...
+
+		if common.IsDone(ctx) {
+			return ctx.Err()
+		}
+
+		depth := 0
+		if ctxDepth, ok := ctx.Value(depthKey).(int); ok {
+			depth = ctxDepth
+		}
+		if int(file.Size()) > maxSize {
```
bill-rich, line 179: I might have just overlooked where this is handled, but I don't see where the total bytes being extracted is being counted. This was in the original archive handler to prevent zip bombs.

ahrav: Ah, so I think in the next PR this check will actually get removed since we'll use our bufferedFileReader to read the contents. That would prevent us from running into any memory issues. This was just a temporary change since we no longer use `ReadToMax`.

pkg/handlers/defaulthandler.go
ahrav: I have a follow up PR to address this issue.

pkg/handlers/default.go
```diff
@@ -0,0 +1,245 @@
...
+			lCtx.Logger().V(3).Info("skipping file due to size")
+			return nil
+		}
+
+		if common.SkipFile(file.Name()) {
+			lCtx.Logger().V(5).Info("skipping file")
+			return nil
+		}
+
+		if h.skipBinaries && common.IsBinary(file.Name()) {
```
rosecodym, line 189: there are two "skip binaries" checks in this pr - do they get unified? later in your change stack?

rosecodym (approved):

pkg/handlers/default.go
ahrav: Yea it should eventually get consolidated into a single check within the defaultHandler.

PR #2710, opened 2024-04-17
[bug] - Enhanced Archive Handling to Address Interface Constraints

ahrav:
### Description:
This PR addresses the critical issue encountered when processing archive files, specifically the error “input type must be an io.ReaderAt and io.Seeker because of zip format constraints.” This problem stems from interface limitations in our existing `openArchive` function when using the archiver library.

#### Issue Details:
In our existing implementation, the `openArchive` function is designed to handle archive files recursively. It does this by passing the file stream obtained from the `Open()` method of `archiver.File` type to subsequent calls to itself. However, the `archive.Extract` method, which is a crucial part of `openArchive`, requires an input that implements the `archiver.seekReaderAt` interface. Unfortunately, the stream type returned by the `Open()` method does not meet this interface requirement, resulting in the aforementioned error.

#### Implemented Solution:
To resolve this issue and enhance our system’s ability to process various archive types, we have introduced a significant update within the `openArchive` function:

- **Introduction of `DiskBufferReader`:** We now wrap the file stream returned by `Open()` with `DiskBufferReader`. This wrapper implements both `io.ReaderAt` and `io.Seeker` interfaces, which are essential for meeting the requirements of the `archive.Extract` method.

- **Handling in `archiver.Decompressor` Case:**
  - Within the switch case handling `archiver.Decompressor` in `openArchive`, we further refined our approach. When dealing with decompression formats, the original non-seekable reader provided by `archive.OpenReader()` was inadequate for recursive archive extraction. We now ensure that this reader is also wrapped in `DiskBufferReader`. This modification allows decompressed data to be reprocessed effectively for potential nested archives, maintaining consistency across different archive formats.

This enhancement ensures that all archive files, whether they require decompression or direct extraction, are processed without errors. The `DiskBufferReader` provides the necessary functionality to overcome the interface constraints previously imposed by the zip format and the archiver library.
### Checklist:


ahrav:
There might be a less heavyweight approach than using the DiskbufferReader here, it was just the simplest in order to get this to work.

Still doing some testing to make sure this covers most of the issues we've seen so far.

rosecodym:
> There might be a less heavyweight approach than using the DiskbufferReader here, it was just the simplest in order to get this to work.

This was going to be my question. Are we going to start spilling a ton more data to the disk by doing this? Are we worried about that?

ahrav:
> > There might be a less heavyweight approach than using the DiskbufferReader here, it was just the simplest in order to get this to work.
> 
> This was going to be my question. Are we going to start spilling a ton more data to the disk by doing this? Are we worried about that?

Yea, I have another [PR](https://github.com/trufflesecurity/trufflehog/pull/2718) in the works to replace this with a more performant solution. Happy to forego this PR and just use that one if you feel that's better. I went with this approach only because it fixed the issue without introducing too many new changes. I might've made one too many PRs in the process 😓

rosecodym (approved):
This looks fine for what it is, and I think you're the one best positioned to judge the risk that we start blowing up temp storage usage. I'm approving, but I ask that you only merge this if you feel confident enough to affirmatively declare that you don't expect problems from all the new temp file writing this might cause, and have some evidence to support that confidence.

PR #2719, opened 2024-04-18
[bug] -  Fix bug and simplify git cat-file command execution and output handling

ahrav:
### Description:
This PR addresses a bug and simplifies the code that executes the `git cat-file` command and handles its output. The changes include:

1. Fixed a bug where, for large files, the code started processing the reader before the entire output was available. This could lead to issues downstream when reading the mime type, seeking, or performing other operations on the partially available data.
2. Replaced the usage of `cmd.StdoutPipe()` and manual pipe handling with `cmd.Output()` to capture the command’s output directly. This ensures that the entire output is available before processing begins.
3. Removed the need for explicit waiting for the command to complete by using `cmd.Output()`, which waits for the command to finish before returning the output.
4. Passed the command’s output as a `bytes.Reader` to `handlers.HandleFile` for further processing.

These changes not only make the code more concise and readable but also fix the bug where processing started before the entire output was available. By using `cmd.Output()`, we ensure that the command’s output is fully captured before being processed by `handlers.HandleFile`, eliminating potential issues related to partially available data.

However, it’s important to note that using `cmd.Output()` stores the entire output in memory. If the `git cat-file` command produces a large amount of output, it may consume a significant amount of memory. This could potentially impact the performance and scalability of the code, especially if dealing with large repositories or frequent invocations.

To address this concern, we can consider the following approach in a future PR:

- Investigate the typical size of the output produced by `git cat-file` in our specific use case.
- If the output is consistently large, we can consider reverting back to using pipes and processing the output incrementally, as in the original code. This would help reduce memory consumption by reading and processing the output in smaller chunks.
- Alternatively, we can explore other optimizations or techniques to handle large output efficiently, such as using disk-based buffers or our BufferedFileWriter.

For now, this PR focuses on fixing the bug, simplifying the code, and ensuring the correctness of the output handling. We can address the potential memory concern in a separate PR based on further analysis and requirements.

Snuck in a few test updates to directly use the file vs creating a a DiskbufferReader.

### Checklist:


rosecodym (approved):
This does look way cleaner, and I'm glad it fixes a bug, but slurping entire files seems _quite_ worrisome to me. I'm approving because you've been closest to this for a while, but I'm curious about what makes you confident that this won't be a destructive change.

ahrav:
> Is there any effective difference between this method and what go-git did pre #2174?

The method discussed here and the approach used by go-git before #2174 are quite similar in terms of functionality. However, this method has a slight advantage in terms of reduced overhead compared to using go-git directly.

Moreover, by adopting this approach, we gain the ability to leverage our [bufferedfilewriter](https://github.com/trufflesecurity/trufflehog/blob/3fa86a1008d96e377f7f7f944b357b66be620c6f/pkg/writers/buffered_file_writer/bufferedfilewriter.go#L52) more effectively, particularly when dealing with very large files. This allows us to write these large files to disk more easily and efficiently, streamlining the process and improving overall performance.

In summary, while the differences may be subtle, this method offers some benefits in terms of overhead reduction and simplified handling of large files through the use of our custom bufferedfilewriter.

ahrav:
> This does look way cleaner, and I'm glad it fixes a bug, but slurping entire files seems _quite_ worrisome to me. I'm approving because you've been closest to this for a while, but I'm curious about what makes you confident that this won't be a destructive change.

I completely understand your concern about slurping entire files, and I agree that it is a worrisome approach. However, this is not intended to be the final solution. The current changes were implemented as an initial step to resolve the specific error we were facing.

To gain some confidence in the viability of this approach, I conducted testing last week on some of the largest open-source organizations. During these tests, I did not encounter any issues related to this specific change, which gives me some optimism that it serves as a good starting point.

Moving forward, my plan is to utilize our [bufferedfilewriter](https://github.com/trufflesecurity/trufflehog/blob/3fa86a1008d96e377f7f7f944b357b66be620c6f/pkg/writers/buffered_file_writer/bufferedfilewriter.go#L52) to handle the writing of larger files to disk. This will help mitigate the concerns associated with loading entire files into memory. I decided to separate these changes into multiple pull requests to keep the modifications more focused and manageable.

rosecodym:
> > Is there any effective difference between this method and what go-git did pre #2174?
> 
> The method discussed here and the approach used by go-git before #2174 are quite similar in terms of functionality. However, this method has a slight advantage in terms of reduced overhead compared to using go-git directly.
> 
> Moreover, by adopting this approach, we gain the ability to leverage our [bufferedfilewriter](https://github.com/trufflesecurity/trufflehog/blob/3fa86a1008d96e377f7f7f944b357b66be620c6f/pkg/writers/buffered_file_writer/bufferedfilewriter.go#L52) more effectively, particularly when dealing with very large files. This allows us to write these large files to disk more easily and efficiently, streamlining the process and improving overall performance.
> 
> In summary, while the differences may be subtle, this method offers some benefits in terms of overhead reduction and simplified handling of large files through the use of our custom bufferedfilewriter.

I think what @rgmz was getting at (or at the observation his comment clued me onto) is that #2174 was done specifically to _stop_ loading entire files into memory because doing that caused a lot of memory pressure. Is this PR undoing all that work? If so, why is it ok to do now what you specifically made a change to avoid then?

rosecodym:
@rgmz I talked synchronously with @ahrav about this, and he said that pre-#2174, go-git did a lot of things to inflate memory usage _besides_ simply slurping data. Re-adding the slurping but not the other stuff apparently does not trigger re-occurrence of the previous memory issues. (@ahrav do I have that right?)

dustin-decker:
>However, it’s important to note that using cmd.Output() stores the entire output in memory. If the git cat-file command produces a large amount of output, it may consume a significant amount of memory. This could potentially impact the performance and scalability of the code, especially if dealing with large repositories or frequent invocations.

GitHub blocks files larger than 100MB and makes you use LFS at that point. This limit depends on the VCS server. As long as we are not scanning LFS, I think we'll see files up to 100MB.

ahrav:
> > However, it’s important to note that using cmd.Output() stores the entire output in memory. If the git cat-file command produces a large amount of output, it may consume a significant amount of memory. This could potentially impact the performance and scalability of the code, especially if dealing with large repositories or frequent invocations.
> 
> GitHub blocks files larger than 100MB and makes you use LFS at that point. This limit depends on the VCS server. As long as we are not scanning LFS, I think we'll see files up to 100MB.

I personally still feel like 100MB is a bit high. Nonetheless, this issue will be remediated once I leverage the changes in [this PR](https://github.com/trufflesecurity/trufflehog/pull/2731#discussion_r1576318358)

mcastorina:
> 1. Fixed a bug where, for large files, the code started processing the reader before the entire output was available. This could lead to issues downstream when reading the mime type, seeking, or performing other operations on the partially available data.

Does this signal that the downstream code isn't properly operating on a stream of data?

ahrav:
> > 1. Fixed a bug where, for large files, the code started processing the reader before the entire output was available. This could lead to issues downstream when reading the mime type, seeking, or performing other operations on the partially available data.
> 
> Does this signal that the downstream code isn't properly operating on a stream of data?

I should add a test.

zricethezav (approved):

rgmz:
> Namely loading the _entire_ repo into memory

The entire **repo**!? I figured that it just loaded the entire file into memory, hence why I asked.

```go
stdout, err := cmd.Output()

return handlers.HandleFile(fileCtx, **bytes.NewReader(stdout)**, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
```

zricethezav:
> The entire repo!? I figured that it just loaded the entire file into memory, hence why I asked.

Err actually upon further review, it uses their billy fs which operates from disk. You _can_ load a repo entirely into memory using `storage/memory` https://github.com/go-git/go-git/blob/master/storage/memory/storage.go#L31. I think billy fs does some in-mem caching though which might have some drawbacks.

PR #2727, opened 2024-04-22
Elastic adapter

camgunz:
### Description:
This PR is to add an elasticsearch source.

### Checklist:
* [*] Tests passing (`make test-community`)?
  - Looks like some issues w/ the `tar` source?
* [*] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?
  - unrelated issues in LDAP seems like


pkg/sources/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,270 @@
...
+		scanCoverageRate := 1.0
+		if previousDocumentCount > 0 && indices.documentCount > 0 {
+			scanCoverageRate =
+				float64(previousDocumentCount) / float64(indices.documentCount)
+		}
+
+		unitsOfWork := distributeDocumentScans(&indices, s.concurrency, scanCoverageRate)
+
+		for uowIndex, outerUOW := range unitsOfWork {
+			uow := outerUOW
```
zricethezav, line 155: @camgunz 
**Suggestion**: Even though `uowIndex` is only used in logs, it still needs to have a copy similar to `uow` inside the loop otherwise we have the potential for incorrect log output

pkg/sources/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,270 @@
...
+				return nil
+			})
+		}
+
+		err = workerPool.Wait()
+		if err != nil {
+			s.log.V(2).Info(fmt.Sprintf("Error waiting on worker pool: %s\n", err))
+		}
+
+		if !s.bestEffortScan {
```
zricethezav, line 264: **Question**: In the case where we are doing continuous scans will this burn up the ES API? Will we run into rate limits? Without having run an ES scan, my intuition suggests we put some sort of sleep after each iteration.

pkg/sources/elasticsearch/unit_of_work.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (ds *DocumentSearch) String() string {
+	if ds.offset > 0 {
+		return fmt.Sprintf("%s [%d:]", ds.index.name, ds.offset)
+	} else {
+		return ds.index.name
+	}
+}
+
+func (uow *UnitOfWork) AddSearch(
```
zricethezav, line 34: **Question**: any reason we need to export this function? Looks like the only place it's being used is `distributeDocumentScans` and the tests.

pkg/sources/elasticsearch/api.go
```diff
@@ -0,0 +1,471 @@
...
+	filterParams            *FilterParams
+	lock                    sync.RWMutex
+}
+
+type elasticSearchRequest interface {
+	Do(providedCtx context.Context, transport esapi.Transport) (*esapi.Response, error)
+}
+
+func (fp *FilterParams) Query(latestTimestamp time.Time) (map[string]any, error) {
+	range_ := make(map[string]any)
```
zricethezav, line 69: **Suggestion**:  Could we rename this variable to be more descriptive or give it a comment explaining what it does?

pkg/sources/elasticsearch/api.go
```diff
@@ -0,0 +1,471 @@
...
+	return query, nil
+}
+
+func NewIndex() *Index {
+	return &Index{}
+}
+
+func (i *Index) DocumentAlreadySeen(document *Document) bool {
+	i.lock.Lock()
+	defer i.lock.Unlock()
```
zricethezav, line 108: **Suggestion**: rather than locking at the top and deferring the unlock we should just surround contentious access with straight up lock/unlocks rather than defer. Easier to figure out what needs to be protected that way.

pkg/sources/elasticsearch/api.go
```diff
@@ -0,0 +1,471 @@
...
+		return true
+	}
+
+	i.latestDocumentIDs = append(i.latestDocumentIDs, document.id)
+	return false
+}
+
+func (i *Index) UpdateLatestTimestampLastRun() {
+	i.lock.Lock()
+	defer i.lock.Unlock()
```
zricethezav, line 132: **Suggestion**: Same thing here, defer doesn't really provide anything for us imo

pkg/sources/elasticsearch/api.go
```diff
@@ -0,0 +1,471 @@
...
+		Index:     []string{docSearch.index.name},
+		KeepAlive: "1m",
+	}
+
+	data, err := makeElasticSearchRequest(ctx, client, req)
+	if err != nil {
+		return "", err
+	}
+
+	pitID, found := data["id"].(string)
```
zricethezav, line 248: **nit**: use `ok` instead of `found` for consistency

pkg/sources/elasticsearch/api.go
```diff
@@ -0,0 +1,471 @@
...
+
+		for _, jsonHit := range hits {
+			documentsFetched++
+
+			hit, ok := jsonHit.(map[string]any)
+			if !ok {
+				continue
+			}
+
+			jsonSort, found := hit["sort"].([]any)
```
zricethezav, line 344: **nit**: use `ok` for consistency

pkg/sources/elasticsearch/elasticsearch.go
camgunz: Oof, nice catch; I think this started life as `_` but then I used it for logging and missed pulling it in 👍

pkg/sources/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,270 @@
...
+							chunk.Data = []byte(document.message)
+
+							return common.CancellableWrite(ctx, chunksChan, &chunk)
+						},
+					)
+					if err != nil {
+						return err
+					}
+
+					fmt.Printf(
```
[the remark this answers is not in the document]
camgunz: Whew 🤦

pkg/sources/elasticsearch/elasticsearch.go
camgunz: Great point; yeah will add a `Sleep`

pkg/sources/elasticsearch/unit_of_work.go
camgunz: Yeah you're right, no need to export here 👍

pkg/sources/elasticsearch/api.go
camgunz: What you don't like mysterious variables that shadow keywords? I'll change it yeah, thanks 👍

pkg/sources/elasticsearch/api.go
camgunz: 👍

pkg/sources/elasticsearch/api.go
```diff
@@ -0,0 +1,471 @@
...
+	indices.lock.RLock()
+	defer indices.lock.RUnlock()
+
+	return indices.processedDocumentsCount
+}
+
+// Adds documents processed to the count, used for progress
+func (indices *Indices) UpdateProcessedDocumentCount(additionalDocumentsProcessed int) {
+	indices.lock.Lock()
+	defer indices.lock.Unlock()
```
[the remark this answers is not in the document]
camgunz: Oh no I think you have a point here, no problem changing 👍

pkg/sources/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		return err
+	}
+
+	s.client = client
+
+	return nil
+}
+
+func (s *Source) buildElasticClient() (*es.TypedClient, error) {
```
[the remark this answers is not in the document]
camgunz: Ah yeah, nice 👍

zricethezav (approved):

PR #2729, opened 2024-04-22
Add false positive info to protobuf

rosecodym:
### Description:
This PR adds false positive information to the `Result` protobuf message in anticipation of us tracking it as first-class secret metadata. We're not doing that yet (it's blocked behind #2643) but setting up the messages now means we'll be able to do it later with less of a code delta.

### Checklist:


proto/detectors.proto
```diff
@@ -1011,6 +1011,13 @@ message Result {
   // This field should only be populated if the verification process itself failed in a way that provides no information
   // about the verification status of the candidate secret, such as if the verification request timed out.
   string verification_error_message = 10;
+
+  FalsePositiveInfo false_positive_info = 11;
+}
+
+message FalsePositiveInfo {
```
ahrav, line 1018: question: Do you foresee a situation where multiple fields in this message will be set to true? I'm wondering if an enum would also do the trick.

ahrav (approved):

dustin-decker (approved):

proto/detectors.proto
rosecodym: potentially! i can also foresee us adding additional information to it down the road (e.g. what word from the list was actually matched). i think an enum would also work but it feels more constraining.

PR #2731, opened 2024-04-22
[feat] - buffered file reader

ahrav:
### Description:
This PR introduces a new `readers` package with a `bufferedFileReader` implementation for efficient random access reading, seeking, and closing operations. The `bufferedFileReader` combines the functionality of `BufferedFileWriter` for buffered writing and an `io.ReadSeekCloser` for random access reading and seeking. It also provides a `Close` method to release the buffer pool.

#### **BUT Why??**

The main reasons for introducing this new reader implementation are two-fold:

1. **Memory Management**: This will provide a way to create a custom reader that can accept an `io.Reader` with any amount of data without worrying about holding all the data in memory. This addresses the current issue with how we handle the result of “git cat-file” where we hold the entire contents in memory. It will also alleviate memory concerns in other scenarios where the reader we pass to `handlers.HandleFile` holds a lot of data.

2. **Archiver Library Compatibility**: The archiver library we use depends on the underlying reader implementing the `seekReaderAt` (io.ReaderAt & io.Seeker) interface. This new `bufferedFileReader` struct allows us to use it in all those places while leveraging our existing `BufferedFileWriter` when writing data.

#### **Ok... but, why not use the existing `DiskbufferReader`??**

While we have an existing `diskbufferReader` implementation which is a valid solution, we decided to introduce the new `bufferedFileReader` for the following reasons:

1. **Efficiency for Small Files**: The majority of files we need to process are under 10MB in size. Creating a temporary file for every single file, as done by the `diskbufferReader`, can be inefficient for small files. The `bufferedFileReader` avoids this overhead by storing small files in memory using the `BufferedFileWriter`.

2. **Observability and Metrics**: The `BufferedFileWriter`, which the `bufferedFileReader` leverages, provides observability through existing instrumented metrics. This allows us to monitor and track the performance of the new reader implementation more effectively.

3. **Buffer Pool Optimization**: The `BufferedFileWriter` is optimized to use an existing buffer pool to reduce memory allocation. This optimization can improve performance.

Key features:

- Utilizes `BufferedFileWriter` for buffered writing and data storage.
- Supports random access reading and seeking using an `io.ReadSeekCloser`.
- Implements `io.Reader`, `io.Seeker`, and `io.ReaderAt` interfaces.
- `Close` method releases the buffer back to the pool.
- Designed for efficient handling of large data with minimal memory usage.

### Checklist:


pkg/readers/bufferedfilereader.go
```diff
@@ -0,0 +1,88 @@
...
+// It implements the io.Seeker interface.
+func (b *bufferedFileReader) Seek(offset int64, whence int) (int64, error) {
+	return b.reader.Seek(offset, whence)
+}
+
+// ReadAt reads len(p) bytes from the underlying io.ReadSeekCloser starting at byte offset off.
+// It returns the number of bytes read and any error encountered.
+// If the io.ReadSeekCloser reaches the end of the available data before len(p) bytes are read,
+// ReadAt returns the number of bytes read and io.EOF.
+// It implements the io.ReaderAt interface.
```
rosecodym, line 81: Should this be asserted in the file?

pkg/readers/bufferedfilereader.go
```diff
@@ -0,0 +1,88 @@
...
+	if err := writer.CloseForWriting(); err != nil {
+		return nil, err
+	}
+
+	reader, err := writer.ReadCloser()
+	if err != nil {
+		return nil, err
+	}
+
+	rdr, ok := reader.(io.ReadSeekCloser)
```
rosecodym, line 40: Since we control all the relevant implementations, is it correct to say that we always expect this to succeed? And if so, should `ReadCloser()` be changed to return a `*io.ReadSeekCloser` so that we don't have to do this at runtime?

ahrav: I considered raising this question in the pr: If we modify the return type to io.ReadSeekCloser, we will also need to revise the ReadCloser method in the contentReader interface within gitparse.go. This adjustment left me somewhat uncertain, but I'm happy to make the change.

I also tend to get a little squimish around interface assertions. 😅

pkg/readers/bufferedfilereader.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating bufferedFileReader: %w", err)
+	}
+
+	// Ensure that the BufferedFileWriter is in read-only mode.
+	if err := writer.CloseForWriting(); err != nil {
+		return nil, err
+	}
+
+	reader, err := writer.ReadCloser()
```
mcastorina, line 38: This is using the `BufferedFileWriter` [ReadCloser](https://github.com/trufflesecurity/trufflehog/blob/a5f04e65f78191ce388f93c9cb2cec29feea7421/pkg/writers/buffer_writer/bufferwriter.go#L72-L81) implementation, correct? Doesn't that load the entire thing into memory, contradicting the first point? I feel like I'm missing something.

> 1. This will provide a way to create a custom reader that can accept an `io.Reader` with any amount of data without worrying about holding all the data in memory.

ahrav: It should not. It would only do that _if_ the writer is under the threshold for requiring a file. If it is not, we return an _autoDeletingFile_.

mcastorina: Oops, I was looking at the `BufferedWriter` implementation, not the [BufferedFileWriter](https://github.com/trufflesecurity/trufflehog/blob/a5f04e65f78191ce388f93c9cb2cec29feea7421/pkg/writers/buffered_file_writer/bufferedfilewriter.go#L201) implementation.

ahrav: Yea... 😢 I could have done a better job distinguishing between the two. I'm open to renaming them to avoid confusion moving forward. Do you have any suggestions?

mcastorina: I think the names are fine, I just made a mistake 🙂

rosecodym (approved):

rosecodym:
the test failure should be fixed on `main`

PR #2733, opened 2024-04-23
[bug] - Fix disk write metric and update BufferedFileWriter file field

ahrav:
### Description:
This PR fixes the disk write metric recording in the `BufferedFileWriter` and updates the `file` field type from `io.Writer` to `os.File`. I don't _feel_ like the `io.Writer` abstraction is necessary, but I'm open to differing opinions.

### Checklist:

### DIFF
--- a/pkg/writers/buffered_file_writer/bufferedfilewriter.go
+++ b/pkg/writers/buffered_file_writer/bufferedfilewriter.go
@@ -167,7 +167,6 @@ func (w *BufferedFileWriter) Write(data []byte) (int, error) {
 		return n, err
 	}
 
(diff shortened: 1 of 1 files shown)


dustin-decker (approved):

mcastorina:
> I don't _feel_ like the `io.Writer` abstraction is necessary, but I'm open to differing opinions.

Removing layers of indirection is generally a good thing. If we only ever use a `*os.File` then we should use that.

I agree with your decision to switch it, especially considering it's an implementation detail of the `BufferedFileWriter` struct. Interfaces are generally a better option to allow users to supply their own implementation.

mcastorina (approved):

PR #2740, opened 2024-04-24
[chore] - update buffered file writer metric

ahrav:
### Description:
Missed one update in my previous PR.

NOTE: We need the initial bucket to exceed the `defaultThreshold` since data of that exact size is never written to disk

### Checklist:


rosecodym (approved):
Are you tuning the buckets configuration based on what you're seeing?

ahrav:
> Are you tuning the buckets configuration based on what you're seeing?

Yes, I have revised the bucket configuration after reviewing our current metrics. I realized that the initial setup did not effectively capture the variations in file sizes. To address this, I expanded the range and increased the interval between buckets, aiming for a more precise depiction of the data distribution.

PR #2742, opened 2024-04-24
[bug] - Refactor newDiff constructor to avoid double initialization of contentWriter

ahrav:
### Description:
This PR addresses an issue in the `newDiff` constructor where the `contentWriter` could potentially be double initialized, leading to duplicate resource creation and potential resource leaks.

Screenshot 2024-04-24 at 2 01 27 PM

**Suggested improvement for bufferwriter package**:
While investigating the issue with the newDiff constructor, it became apparent that the current design of the bufferwriter package could be improved to make it more ergonomic and less prone to misuse.
Instead of allocating resources (i.e., the underlying buffer) in the `New` constructor, we should consider modifying the implementation to allocate resources lazily, only when the `Write` method is called for the first time. This lazy allocation approach would prevent potential resource leaks and make it easier to use the package correctly.

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse_test.go
+++ b/pkg/gitparse/gitparse_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 	bufferwriter "github.com/trufflesecurity/trufflehog/v3/pkg/writers/buffer_writer"
 	bufferedfilewriter "github.com/trufflesecurity/trufflehog/v3/pkg/writers/buffered_file_writer"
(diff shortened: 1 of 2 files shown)


pkg/gitparse/gitparse_test.go
```diff
@@ -2346,3 +2348,48 @@ index 2ee133b..12b4843 100644
...
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctx := context.Background()
+			commit := new(Commit)
+
+			mockWriter := newMockContentWriter()
+			assert.NotNil(t, mockWriter, "Failed to create mockWriter")
```
mcastorina, line 2386: Is `mockWriter` doing anything useful in this test? It seems to be just a test fixture.

mcastorina (approved):

pkg/gitparse/gitparse_test.go
ahrav: 🤦 Thank you for catching this. I updated the test to ensure `mockContentWriter` implements `contentWriter` as well as using it as part of the `opts` for the second test case.

mcastorina: I'm still not sure I understand how the test catches double initialization. It looks like counter is set to `1` in `newMockContentWriter()` and never modified, so it'll always be `1` regardless of what `newDiff` does.

PR #2743, opened 2024-04-24
Expose detector-specific false positive logic

rosecodym:
### Description:
This PR:
* Creates an optional interface that detectors can use to customize their false positive detection
* Implements this interface on detectors that have custom logic
 * In most cases this "custom logic" is simply a no-op because the detector does not participate in false positive detection
* Eliminates inline (old-style) false positive exclusion in a few detectors that #2643 missed

### Checklist:


pkg/detectors/wiz/wiz.go
```diff
@@ -69,14 +69,6 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				s1.SetVerificationError(verificationErr, idMatch, secretMatch)
 			}
 
-			// This function will check false positives for common test words, but also it will make sure the key appears 'random' enough to be a real key.
-			if !s1.Verified {
-				if detectors.IsKnownFalsePositive(idMatch, detectors.DefaultFalsePositives, true) ||
-					detectors.IsKnownFalsePositive(secretMatch, detectors.DefaultFalsePositives, true) {
```
rosecodym, line 75: This custom logic wasn't moved into a `CustomFalsePositiveChecker` implementation because `secretMatch` is no longer available by the time that runs. This type of change (which "narrows" false positive detection) was made for multiple other detectors in #2643 so I'm not worried about it.

ahrav (approved):

pkg/custom_detectors/custom_detectors.go
```diff
@@ -109,6 +110,10 @@ func (c *CustomRegexWebhook) FromData(ctx context.Context, verify bool, data []b
 	return results, nil
 }
 
+func (c *CustomRegexWebhook) IsFalsePositive(_ detectors.Result) bool {
```
mcastorina, line 113: **Nit:** If all parameters are ignored, you can omit `_`. You can even omit `c` in the pointer receiver.

```go
func (*CustomRegexWebhook) IsFalsePositive(detectors.Result) bool
```

pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
```diff
@@ -94,6 +95,10 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func (s Scanner) IsFalsePositive(_ detectors.Result) bool {
+	return false
+}
```
mcastorina, line 100: Did you consider creating a struct that implements this so it can be embedded into the `Scanner` struct?

```go
// in detectors package
type SkipFalsePositiveCheck struct{}

func (SkipFalsePositiveCheck) IsFalsePositive(Result) bool {
    return false
}

// elsewhere

type Scanner struct {
     detectors.SkipFalsePositiveCheck
}
```

rosecodym: no, that's a much better idea!

mcastorina: Marginally? imo it's a "nice-to-have" for a common scenario, especially for areas of the code that get a lot of outside contribution.

PR #2713, opened 2024-04-18
Scan commit metadata

rgmz:
### Description:
This fixes #2683. It scans the commit author, committer (which is typically `GitHub <noreply@github.com>` for GitHub, but can be different), and message.

```
commit 8104611d6e2e3993c104974ea4c507faced7e847
Author:     Richard Gomez <32133502+rgmz@users.noreply.github.com>
AuthorDate: Tue Feb 6 10:13:53 2024 -0500
Commit:     GitHub <noreply@github.com>
CommitDate: Tue Feb 6 10:13:53 2024 -0500

    fix: case-insensitive ext check (#2383)
```

Admittedly, this implementation is a bit of guess-work and may not be the best way.

### Checklist:
* [x] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #2754, opened 2024-04-26
Scan commit metadata

rgmz:
### Description:
This is a follow-up to #2713 that fixes the strange test error.

As suspected, the failure was caused by additional diffs not being included in the test's expected data.

### Checklist:


rosecodym (approved):
Ok I pulled this down and ran the whole test suite this time. I don't see any git-related failures! (There are some GCS failures, but they also occur on `main` :( )

rgmz:
Bah, it seems like this test also needs to be updated to factor in the new diffs:
```
 === Failed
=== FAIL: pkg/sources/git TestChunkUnit (0.17s)
    git_test.go:588: 
        	Error Trace:	/home/runner/work/trufflehog/trufflehog/pkg/sources/git/git_test.go:588
        	Error:      	Not equal: 
        	            	expected: 11
        	            	actual  : 22
        	Test:       	TestChunkUnit

DONE 734 tests, 1 failure in 136.019s
```

https://github.com/trufflesecurity/trufflehog/actions/runs/8885322690/job/24396435333

Issue #2753, opened 2024-04-26
Postman not working?


marcelo321:
I  am not sure if it isn't working or I am doing something wrong.

```
trufflehog postman --workspace "linode" --token mytoken

2024-04-26T05:44:22Z    error   trufflehog      engine failed to finish execution       {"error": "fatal: error getting workspace linode: could not get workspace: linode"}
```

It says error getting workspace for any term I use, including paypal or any

marcelo321:
I just created the key here:

![image](https://github.com/trufflesecurity/trufflehog/assets/29487296/9c20a953-a0e6-4bc3-a42e-f1b42671055c)

dustin-decker:
I think you need to use the workspace ID instead. The flag could have been named more clearly.

marcelo321:
@dustin-decker giving a exact workspace id also fails to me, can you give me an example of a workspace and what id would it use so I can try it?? would be appreciated

marcelo321:
Ooh I see. Do you have any info on how the team was able to scrap ids when searching on the public workspaces..? I read the blog post which is great, but I am finding pretty difficult to understand the postman api. Not sure if you were part of that research or not

dustin-decker:
I do not know the details of that part.

zricethezav:
> Do you have any info on how the team was able to scrap ids when searching on the public workspaces

That's something we did internally and don't plan on releasing. Closing this issue as I believe the original issue has been answered.

PR #2755, opened 2024-04-26
Update postman flags to be less confusing

dustin-decker:
### Description:
Addresses #2753 

A number of flags require IDs rather than names and are not clearly specified. This change makes those flags hidden (they still work) and replaces them with more clearly defined flags.

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -158,11 +158,11 @@ var (
 	postmanScan  = cli.Command("postman", "Scan Postman")
 	postmanToken = postmanScan.Flag("token", "Postman token. Can also be provided with environment variable").Envar("POSTMAN_TOKEN").String()
 
-	postmanWorkspaces    = postmanScan.Flag("workspace", "Postman workspace to scan. You can repeat this flag. Deprecated flag.").Hidden().Strings()
-	postmanCollectionIDs = postmanScan.Flag("collection-id", "Postman collection ID to scan. You can repeat this flag.").Strings()
-
-	postmanCollections  = postmanScan.Flag("collection", "Postman collection to scan. You can repeat this flag. Deprecated flag.").Hidden().Strings()
+	postmanWorkspaces   = postmanScan.Flag("workspace", "Postman workspace to scan. You can repeat this flag. Deprecated flag.").Hidden().Strings()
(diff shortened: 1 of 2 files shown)


zricethezav:
@dustin-decker good call. This is much clearer. Is this ready for a review?

main.go
```diff
@@ -155,14 +155,25 @@ var (
...
-	postmanScan                = cli.Command("postman", "Scan Postman")
-	postmanToken               = postmanScan.Flag("token", "Postman token. Can also be provided with environment variable").Envar("POSTMAN_TOKEN").String()
-	postmanWorkspaces          = postmanScan.Flag("workspace", "Postman workspace to scan. You can repeat this flag.").Strings()
-	postmanCollections         = postmanScan.Flag("collection", "Postman collection to scan. You can repeat this flag.").Strings()
-	postmanEnvironments        = postmanScan.Flag("environment", "Postman environment to scan. You can repeat this flag.").Strings()
-	postmanIncludeCollections  = postmanScan.Flag("include-collections", "Collections to include in scan. You can repeat this flag.").Strings()
+	postmanScan  = cli.Command("postman", "Scan Postman")
+	postmanToken = postmanScan.Flag("token", "Postman token. Can also be provided with environment variable").Envar("POSTMAN_TOKEN").String()
+
+	postmanWorkspaces    = postmanScan.Flag("workspace", "Postman workspace to scan. You can repeat this flag. Deprecated flag.").Hidden().Strings()
```
zricethezav, line 161: @dustin-decker **Nit**: would it be better to group the workspaces/workspaceIDs and collection/collectionID variables together instead of the way it is now?

dustin-decker: I wanted it to be clear that there are two flags that do the same thing but I don't think it makes a big difference either way.

zricethezav: > I wanted it to be clear that there are two flags that do the same thing but I don't think it makes a big difference either way.

These two flags do different things.
```
postmanWorkspaces    = postmanScan.Flag("workspace", "Postman workspace to scan. You can repeat this flag. Deprecated flag.").Hidden().Strings()
postmanCollectionIDs = postmanScan.Flag("collection-id", "Postman collection ID to scan. You can repeat this flag.").Strings()
```

It would be better if the grouping was logically consistent.

dustin-decker: updated

zricethezav (approved):

PR #2767, opened 2024-04-30
Fix Git source test

rgmz:
### Description:
This fixes one missed test in https://github.com/trufflesecurity/trufflehog/pull/2754#issuecomment-2083679022.

The number of chunks doubled because each commit now has metadata + data.

### Checklist:


rosecodym (approved):
I'm going to investigate why my local test run didn't catch this :/

rgmz:
I suspect that the GCS errors either buried this one, or cause the tests to exit early.

> (There are some GCS failures, but they also occur on main :( )

rosecodym:
Ah, yep, they were there, I just missed them :(

PR #2774, opened 2024-05-01
Fix some GitHub source test issues

rgmz:
### Description:
The [`test-detectors` workflow is failing due to numerous issues](https://github.com/trufflesecurity/trufflehog/actions/runs/8906992237/job/24460068279). This fixes a few of them.

### Checklist:

### DIFF
--- a/pkg/sources/github/github_test.go
+++ b/pkg/sources/github/github_test.go
@@ -85,6 +85,7 @@ func TestAddReposByOrg(t *testing.T) {
 	assert.Equal(t, 1, s.filteredRepoCache.Count())
 	ok := s.filteredRepoCache.Exists("super-secret-repo")
 	assert.True(t, ok)
+	assert.False(t, gock.HasUnmatchedRequest())
 	assert.True(t, gock.IsDone())
 }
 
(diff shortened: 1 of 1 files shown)


pkg/sources/github/github.go
```diff
@@ -501,11 +501,11 @@ func (s *Source) enumerateUnauthenticated(ctx context.Context, apiEndpoint strin
...
 			s.log.Error(err, "error fetching repos for org")
-		}
 
-		// We probably don't need to do this, since getting repos by org makes more sense?
-		if err := s.getReposByUser(ctx, org); err != nil {
-			s.log.Error(err, "error fetching repos for user")
+			// We probably don't need to do this, since getting repos by org makes more sense?
+			if err := s.getReposByUser(ctx, org); err != nil {
+				s.log.Error(err, "error fetching repos for user")
+			}
```
rgmz, line 508: It only makes sense to try `/users/{name}` if `/orgs/{name}` fails.

pkg/sources/github/github_test.go
```diff
@@ -94,27 +95,28 @@ func TestAddReposByOrg_IncludeRepos(t *testing.T) {
...
 	err := s.getReposByOrg(context.Background(), "super-secret-org")
 	assert.Nil(t, err)
 	assert.Equal(t, 2, s.filteredRepoCache.Count())
-	ok := s.filteredRepoCache.Exists("secret/super-secret-repo")
+	ok := s.filteredRepoCache.Exists("super-secret-org/super-secret-repo")
 	assert.True(t, ok)
-	ok = s.filteredRepoCache.Exists("secret/super-secret-repo2")
+	ok = s.filteredRepoCache.Exists("super-secret-org/super-secret-repo2")
 	assert.True(t, ok)
+	assert.False(t, gock.HasUnmatchedRequest())
```
rgmz, line 119: Add explicit failures if gock has unmatched requests. Otherwise, this will be logged but not fail the test which IMO isn't correct.

rgmz:
Fixed the merge conflict.

ahrav (approved):
Thanks for fixing these.

pkg/sources/github/github_test.go
rgmz: Already paying off. It's already caught an (inconsequential) issue introduced by #2814.

(This gets called twice.)
https://github.com/rgmz/trufflehog/blob/5102e3ae1178e320e749ca14ac94b4073165c975/pkg/sources/github/github.go#L620-L622

PR #2779, opened 2024-05-02
Update result's extra data for Slack

abmussani:
### Description:
This PR updates the result's extra data for slack with name of user and team to whom token belongs to. Also, Since Slack have multiple type of tokens, It will be good to have token_type in extra data of result.

### Checklist:


zricethezav (approved):
Seems like a good addition. Thanks for the contribution!

PR #2780, opened 2024-05-02
fix for infinite recursion in Postman var sub

zricethezav:
### Description:
There was a bug in the variable substitution code for Postman that caused infinite recursion which happened when a variable value referenced itself. This is a stopgap solution. More investigation needs to be done but this at least relieves the infinite recursion bug. 

### Checklist:

### DIFF
--- a/pkg/sources/postman/substitution.go
+++ b/pkg/sources/postman/substitution.go
@@ -74,8 +74,7 @@ func (s *Source) buildSubstitution(data string, metadata Metadata, combos *map[s
 			// to ensure we don't infinitely recurse, we will trim all `{{}}` from the values before replacement.
 			// this is to prevent the case where a variable is replaced with a value that contains the same variable causing
 			// an infinite loop
-			removedBrackets := strings.ReplaceAll(slice.value, "{{", "")
-			removedBrackets = strings.ReplaceAll(removedBrackets, "}}", "")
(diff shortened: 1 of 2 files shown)


joeleonjr:
LGTM. This fixes the infinite recursion issue and maintains the original capability to sub key/value pairs.

pkg/sources/postman/substitution.go
```diff
@@ -70,8 +70,14 @@ func (s *Source) buildSubstitution(data string, metadata Metadata, combos *map[s
 			if slice.Metadata.CollectionInfo.PostmanID != "" && slice.Metadata.CollectionInfo.PostmanID != metadata.CollectionInfo.PostmanID {
 				continue
 			}
-			// to ensure we don't infinitely recurse, we will trim all `{{}}` from the values before replacement
-			d := strings.ReplaceAll(data, match, strings.Trim(slice.value, "{}"))
+
+			// to ensure we don't infinitely recurse, we will trim all `{{}}` from the values before replacement.
+			// this is to prevent the case where a variable is replaced with a value that contains the same variable causing
+			// an infinite loop
+			removedBrackets := strings.ReplaceAll(slice.value, "{{", "")
```
ahrav, line 77: Using the same variable on subsequent lines always has me doing doing a double take. We could consolidate this into a single expression, though it might be a bit fugly: 😅 

`removedBrackets := strings.ReplaceAll(strings.ReplaceAll(slice.value, “{{“, “”), “}}“, “”)`

ahrav (approved):

PR #2789, opened 2024-05-06
Add webhook source protos

dustin-decker:
### Description:
Add protos for webhook source

### Checklist:


proto/source_metadata.proto
```diff
@@ -298,6 +298,20 @@ message Postman {
   string variable_type = 15;
 }
 
+
+
+message Vector {
+  string timestamp = 1;
```
ahrav, line 304: suggestion: Could we use a [Timestamp](https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp) type here instead? I see the following drawbacks of using a string here.

- Inefficient Parsing: Converting strings to timestamps and vice-versa requires extra code and processing. It's also likely more error prone.
- Ambiguous Formatting: The ambiguity in the different formats complicates parsing and can lead to timezone-related errors. (e.g., "2023-12-25 10:30:00", "1671960600")
- Inflexible Operations: Manipulating timestamps—like calculating durations or filtering by time ranges—is much harder with strings than with purpose-built timestamp types.

ahrav (approved):

PR #2807, opened 2024-05-08
Update results's extra data for Twilio

abmussani:
### Description:
Twilio returns services information in response of verification endpoint. From that, Friendly name and account_sid is added in extra data.

### Checklist:

### DIFF
--- a/pkg/detectors/twilio/twilio.go
+++ b/pkg/detectors/twilio/twilio.go
@@ -88,12 +88,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 					defer res.Body.Close()
 
 					if res.StatusCode >= 200 && res.StatusCode < 300 {
+						s1.Verified = true
 						var serviceResponse serviceResponse
-						if err := json.NewDecoder(res.Body).Decode(&serviceResponse); err != nil { // unexpected response
-							err = fmt.Errorf("failed to decode service response: %w", err)
-							s1.SetVerificationError(err, key)
-						} else {
-							s1.Verified = true
+						if err := json.NewDecoder(res.Body).Decode(&serviceResponse); err == nil && len(serviceResponse.Services) > 0 { // no error in parsing and have atleast one service
 							service := serviceResponse.Services[0]
 							s1.ExtraData["friendly_name"] = service.FriendlyName
 							s1.ExtraData["account_sid"] = service.AccountSID
--- a/pkg/detectors/twilio/twilio.go
+++ b/pkg/detectors/twilio/twilio.go
@@ -2,9 +2,11 @@ package twilio
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
 	"net/http"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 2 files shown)


pkg/detectors/twilio/twilio.go
```diff
@@ -73,10 +85,18 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				req.SetBasicAuth(sid, key)
 				res, err := client.Do(req)
 				if err == nil {
-					res.Body.Close() // The request body is unused.
+					defer res.Body.Close()
+					var serviceResponse serviceResponse
+					if err := json.NewDecoder(res.Body).Decode(&serviceResponse); err != nil {
+						err = fmt.Errorf("failed to decode service response: %w", err)
+						s1.SetVerificationError(err, key)
```
rosecodym, line 92: "Verification errors" are errors that interfered with TruffleHog's ability to determine whether a secret is verified, so they should only be set when TruffleHog was unable to do that. In particular, `SetVerificationError` shouldn't be called on results with `Verified == true`, which would be a possibility after your change.

This does mean that there's not a good way to store non-fatal errors within a result structure, which is admittedly unfortunate, but it's the reality right now. In this particular case I believe your two choices if this decoding fails are:
1. Assume that verification failed (because the endpoint is probably behaving unexpectedly), even if the HTTP response is 2xx. In that case you would just `continue` here to skip the actual HTTP status check.
1. Just ignore it - the important part is whether the HTTP response indicates a verified credential, not this additional metadata.

I think that either position is defensible. Personally I lean towards (2) but I don't speak with authority on that.

rosecodym (changes requested):
Thanks for this! This metadata will be useful. Unfortunately, your implementation violates a (non-obvious) invariant that TruffleHog assumes - I've left a specific comment inline.

pkg/detectors/twilio/twilio.go
rosecodym: @rgmz do you have any good examples you can provide? The invariant exists because it simplifies the underlying logical model and we've lacked concrete cases that would justify adding more complexity. But if you've got some that calculus might change.

abmussani: In case of Twilio, I would also lean towards (2) as well because originally the verification is dependent on Status.OK not on the response. Additional Metadata is optional, I believe we can skip it if there is error in parsing response or reading the stream.

pkg/detectors/twilio/twilio.go
```diff
@@ -73,10 +85,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 
 					if res.StatusCode >= 200 && res.StatusCode < 300 {
-						s1.Verified = true
+						var serviceResponse serviceResponse
+						if err := json.NewDecoder(res.Body).Decode(&serviceResponse); err != nil { // unexpected response
+							err = fmt.Errorf("failed to decode service response: %w", err)
+							s1.SetVerificationError(err, key)
+						} else {
+							s1.Verified = true
+							service := serviceResponse.Services[0]
```
rosecodym, line 97: We should be extra safe and ensure that the slice is non-empty before accessing its first element. I wouldn't want to panic here.

rosecodym (commented):
~I noticed that you answered the question of how to handle an error when decoding the body in the opposite way you did in #2808 - was there a particular reason for that?~ Are you going to update this PR to ignore the body decoding error, like you did in #2808?

pkg/detectors/twilio/twilio.go
abmussani: I am going to update the PR.

abmussani: I have pushed the changes

pkg/detectors/twilio/twilio.go
abmussani: Btw, I found the similar issue in Slack detector. I will create a separate PR for that.

rosecodym (approved):
Thanks for this!

PR #2808, opened 2024-05-08
Update azure storage extra data

abmussani:
### Description:
Blob service of Azure storage returns containers name in response. From that, containers name is added in extra data.

### Checklist:


pkg/detectors/azurestorage/azurestorage.go
```diff
@@ -98,6 +111,19 @@ func verifyAzureStorageKey(ctx context.Context, client *http.Client, accountName
...
+	if err == nil {
+		response := storageResponse{}
+		if err := xml.Unmarshal(body, &response); err == nil {
+			var containerNames []string
+			for _, c := range response.Containers.Container {
+				containerNames = append(containerNames, c.Name)
+			}
+			extraData["container_names"] = strings.Join(containerNames, ", ")
+		}
+	}
```
rgmz, line 125: This should probably be inside the `http.StatusOk` block, as I doubt it would return the info for non-success responses.

pkg/detectors/azurestorage/azurestorage.go
```diff
@@ -98,6 +111,19 @@ func verifyAzureStorageKey(ctx context.Context, client *http.Client, accountName
 	}
 	defer res.Body.Close()
 
+	// parse container names and append in extra data
+	body, err := io.ReadAll(res.Body)
+	if err == nil {
+		response := storageResponse{}
+		if err := xml.Unmarshal(body, &response); err == nil {
```
rgmz, line 118: I would personally invert the conditions so `if err != nil` returns the error, rather than discarding it.

pkg/detectors/azurestorage/azurestorage.go
abmussani: make sense. I have moved the parsing and unmarshaling code in StatusOk block.

pkg/detectors/azurestorage/azurestorage.go
rgmz: The decoding failing is notable because it's reasonable to expect that a `200 OK` response means the key is valid. Personally, I might return the error AND set `verified = true`, but per @rosecodym's [comment](https://github.com/trufflesecurity/trufflehog/pull/2807#discussion_r1594184573) you should return the error and mark verification as failed.

abmussani: @rgmz I believe @rosecodym shared 2 choices. the second one seems to be like ignoring the additional metadata if error occures in parsing and reading the stream as verification is based on Status.OK. Please correct me if my understanding is wrong @rosecodym.

rosecodym: Yes, that's correct. There are clearly some general questions to resolve:
* If we can't decode the response body, is that a fatal error?
* How should we handle non-fatal errors?

We need to decide on general answers to these general questions, but until we do, I think it is acceptable to answer them with "no" and "ignore them" respectively, because that represents a minimal change from the current behavior.

(Another option we haven't discussed yet is to redact the errors and put them themselves into `ExtraData`.)

PR #2811, opened 2024-05-08
[bug] - Fix case-sensitivity issue in PrefixRegex function

ahrav:
### Description:
The `PrefixRegex` function is used to ensure that at least one of the given keywords is within 40 characters of the capturing group that follows. This helps prevent false positives in regex matching.

However, there was an issue with the case-sensitivity handling in the function. The `(?i)` flag was used to make the keyword matching case-insensitive, but it inadvertently made the entire regex, including the capturing group, case-insensitive as well.

To fix this issue, the `pre` variable in the `PrefixRegex` function has been updated to `(?i:`. This limits the case-insensitivity to only the keyword matching portion of the regex. The capture group that follows will be matched case-sensitively.

### Checklist:

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -156,9 +156,9 @@ func CleanResults(results []Result) []Result {
 // 40 characters of the capturing group that follows.
 // This can help prevent false positives.
 func PrefixRegex(keywords []string) string {
-	pre := `(?i)(?:`
+	pre := `(?i:`
 	middle := strings.Join(keywords, "|")
-	post := `)(?:.|[\n\r]){0,40}?(?-i)`
+	post := `)(?:.|[\n\r]){0,40}?`
 	return pre + middle + post
 }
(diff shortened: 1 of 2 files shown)


pkg/detectors/detectors.go
```diff
@@ -154,12 +153,12 @@ func CleanResults(results []Result) []Result {
 }
 
 // PrefixRegex ensures that at least one of the given keywords is within
-// 20 characters of the capturing group that follows.
+// 40 characters of the capturing group that follows.
 // This can help prevent false positives.
 func PrefixRegex(keywords []string) string {
 	pre := `(?i)(?:`
```
rgmz, line 159: This would also work:
```diff
-(?i)(?:
+(?i:
```

ahrav: Thanks, I like that better.

zricethezav (approved):

Issue #2739, opened 2024-04-24
re2 error: `re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35`


rgmz:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

### Trace Output

```
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
re2/re2.cc:772: DFA out of memory: pattern length 102, program size 928, list count 352, bytemap range 35
```

### Expected Behavior

The chunk data should be scanned.

### Actual Behavior

TruffleHog outputs the aforementioned error from re2, making it unclear what the cause is and whether certain chunks were skipped.

### Steps to Reproduce

The error seems semi-random so it's difficult to reproduce. Additionally, the log comes directly from `re2.cc`, meaning there is no context associated with it.

## Environment
 * OS: [e.g. iOS]
 * Version [e.g. 22]

## Additional Context

https://github.com/google/re2/issues/186

### References

* #0000

zricethezav:
Maybe providing an option for users to pick which regex engine they want, re2 or default, would be worthwhile since re2 is a drop-in replacement of `regex`

dustin-decker:
The whole diff is never scanned, we use a sliding-window-with-overlap chunker to break up data into more manageable chunks: 

https://github.com/trufflesecurity/trufflehog/blob/333c4f52961bf1d06d04a82fbdea35a796d102db/pkg/sources/chunker.go#L13-L18

Looks like the default max_mem is 8MB, so i'm guessing we have an expensive regex on some data?

[...]

Issue #2262, opened 2023-12-25
Avoid verifying the same key multiple times in a session


bugbaba:
Hello Team,

## Description
Verify a key only once even if it is found multiple times in the same or different files.

For example, in the below screenshot, we can see that the same gitlab key(revoked) is getting verified twice as it's mentioned twice in the file.

![image](https://github.com/trufflesecurity/trufflehog/assets/29960776/11ac43f2-65f4-450f-879a-f1f24ecc5134)

### Preferred Solution
To avoid wasting resources on re-verification and hitting detectors with the same keys multiple times, it is ideal to check if the given key has already been verified.

Maybe a check before the `if verify` block? to confirm if the key has not been verified previously.

--
Best Regards,
@bugbaba

rgmz:
This would benefit both trufflehog and any endpoints it calls.

I was actually playing around with adding caches to detectors so that known verifications only happen once. Work smarter, not harder. :)

rgmz:
@ahrav has an experimental implementation of this in #2276.

ahrav:
I agree, using a cache reduces duplicate external API calls for credential verification, improving performance and API stewardship. However, storing plaintext credentials presents a potential security risk if exposed, even though the chance of in-memory hash compromise is low. An option to mitigate exposure while retaining detection speed/efficiency could be hashing credentials before caching them with a high-speed algorithm like [XXHash](https://github.com/cespare/xxhash). This safeguards credentials while still allowing cache hits on matched hashes. Overall this balances security, performance, and responsible API usage - preventing duplicate verification calls for the same credentials.

bugbaba:
Also, I think we shouldn't add `ExtraData` as metadata in the cache as it mostly contains sensitive info extracted from valid responses.


We just need to check if the given match exists in the cache or not. So maybe we can ignore all metadata if there are no plans to use metadata from cache.

```
if match in cachelist{
  ignore
}
verify block {
  verification logic
  add match to cachelist
}
```

kashifkhan0771:
We want to retain duplicate secrets but avoid re-verifying them and instead, we should reuse the verification results from earlier scans. Also, this behavior should apply not just to a single chunk, but across all data.

Is that correct?

bugbaba:
Yes,  We want to first capture all the matches for a detector and deduplicate those so that each match is only checked once if its valid or not.



[...]

PR #2812, opened 2024-05-09
[feat] - Optimize detector performance by reducing data passed to regex

ahrav:
### Description:
This PR introduces optimizations to improve the performance of detectors by reducing the amount of data passed to the regex within the `FromData` method. The changes leverage the knowledge of keyword positions to extract relevant portions of the chunk data, where the secret is likely to reside.

Key changes:

1. Introduced `DetectorMatch` struct to represent a detected pattern’s metadata, including the detector key, detector instance, and a slice of `matchSpan` structs representing the start and end offsets of matched keywords within the chunk.

2. Added `matchSpan` struct to represent a single occurrence of a matched keyword, containing the start and end byte offsets within the chunk.

3. Implemented `Matches` method for `DetectorMatch` to extract the relevant portions of the chunk data based on the start and end positions of each match. The end position is determined by taking the minimum of the keyword position + `maxMatchLength` (set to 300) and the length of the chunk data.

4. Introduced `FindDetectorMatches` function (previously `PopulateMatchingDetectors`) to return a slice of `DetectorMatch` instances, each containing the detector key, detector, and a slice of matches. Adjacent or overlapping matches are merged using the `mergeMatches` function to avoid duplicating or overlapping the matched portions of the chunk data.

5. Updated the detection logic to use the `Matches` method of `DetectorMatch` to extract the relevant portions of the chunk data before passing them to the `FromData` method of the detector.

6. Introduced `MaxSecretSizeProvider` interface that detectors can optionally implement to provide a custom maximum size for the secrets they detect. The interface includes a single method `ProvideMaxSecretSize() int64` that returns the maximum size of the secret the detector expects to find.

7. As part of the `FindDetectorMatches` function, it checks if a detector implements the `MaxSecretSizeProvider` interface. If implemented, the `ProvideMaxSecretSize` method is called to obtain the detector-specific maximum secret size, which is used to determine the end position of the match span. If the interface is not implemented, the default `maxMatchLength` constant is used.

8. Implemented the `MaxSecretSizeProvider` interface in the relevant detectors (`PrivateKeyDetector`, `GCPDetector`, and `GCPApplicationDefaultCredentialsDetector`) and provided appropriate values for the maximum secret size based on the expected size of the secrets they detect. I might be missing some detectors that should implement this interface... i just can't think of them right now 😞 

The optimization is based on the assumption that most secrets shouldn’t exceed a certain length from the keyword’s position. By default, the `maxMatchLength` constant is set to 300 characters. However, detectors that require a larger or smaller max size can implement the `MaxSecretSizeProvider` interface and provide their own value through the `ProvideMaxSecretSize` method.

These changes significantly reduce the amount of data the regex within `FromData` has to process, leading to improved detector performance while still ensuring accurate secret detection. The introduction of the `MaxSecretSizeProvider` interface allows for flexibility in handling different secret sizes based on the specific requirements of each detector.

### Sequence Diagram
optimized engine regex-2024-05-09-003033

### Benchmarks
Benchmark assessing the performance of `FromData` with verification disabled across various chunk sizes.

orange: old chunk size (10kB)
green: new chunk size (512B overestimate for most detectors)
Screenshot 2024-05-08 at 8 50 56 AM

### Checklist:


rgmz:
> The end position is determined by taking the minimum of the keyword position + maxMatchLength (set to 300) and the length of the chunk data.

While I think this is a good idea, it would be a mistake to set a default max length. I can think of several detectors that can easily exceed this — JWT, private key, GCP, and Docker (#2677) to name a few. Because there are hundreds of detectors, the safer approach would be to make this opt-in.

This would also interfere with detectors that require multiple parts (e.g., client ID & secret, username & password, secret & URL).

pkg/engine/engine.go
```diff
@@ -814,38 +800,48 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
...
-	}
+	// To reduce the overhead of regex calls in the detector, we limit the amount of data passed to each detector.
+	// The Matches method of the DetectorMatch struct is used to extract the relevant portions of the chunk data
+	// based on the start and end positions of each match. The end position is determined by taking the minimum
+	// of the keyword position + maxMatchLength and the length of the chunk data.
+	// This optimization is based on the assumption that most secrets shouldn't exceed maxMatchLength (300)
+	// characters in length from the keyword's position.
+	// The value of maxMatchLength is a constant and can be adjusted based on further analysis and requirements.
+	matchedBytes := data.detector.Matches(data.chunk.Data)
+	for _, match := range matchedBytes {
```
rgmz, line 811: Does `mergeMatches` aide in de-duplication at all? Some of [my recent changes](https://github.com/trufflesecurity/trufflehog/blob/c7b72b98677722afac24a2bafe84bd98074129b4/pkg/detectors/alchemy/alchemy.go#L39-L42) have been around de-duplicating results in chunks to prevent making the same network calls 2/10/20 times. Admittedly, a better solution would be #2262 rather than caching matches in a given chunk.

pkg/engine/engine_test.go
```diff
@@ -622,3 +623,65 @@ func TestLikelyDuplicate(t *testing.T) {
...
+	largeChunk := 1 << 14  // 16 KB
+	xlChunk := 1 << 15     // 32 KB
+	xxlChunk := 1 << 16    // 64 KB
+	xxxlChunk := 1 << 18   // 256 KB
+	chunkSizes := []int{smallChunk, mediumChunk, largeChunk, xlChunk, xxlChunk, xxxlChunk}
+
+	for _, chunkSize := range chunkSizes {
+		b.Run(fmt.Sprintf("ChunkSize_%d", chunkSize), func(b *testing.B) {
+			b.ReportAllocs()
+			b.SetBytes(int64(dataSize))
```
dustin-decker, line 646: Update this to chunkSize to update the throughput on the benchmarks

dustin-decker:
> > The end position is determined by taking the minimum of the keyword position + maxMatchLength (set to 300) and the length of the chunk data.
> > ...
> > 8. Implemented the MaxSecretSizeProvider interface in the relevant detectors (PrivateKeyDetector, GCPDetector, and GCPApplicationDefaultCredentialsDetector) and provided appropriate values for the maximum secret size based on the expected size of the secrets they detect. I might be missing some detectors that should implement this interface... i just can't think of them right now 😞
> 
> While I think this is a good idea, it would be a mistake to set a default max length. I can think of several detectors that can easily exceed this — JWT, private key, GCP, and Docker (#2677) to name a few. Because there are hundreds of detectors, the safer approach would be to make this opt-in.
> 
> This would also interfere with detectors that require multiple parts (e.g., client ID & secret, username & password, secret & URL).
> 
> Edit: Incidentally, this would likely solve #2739.

I think we can increase the default to 1024 bytes for the multi-part credential case. Should be adequate in most cases. I'd like to see this optimization on by default, but we can provide an opt-out flag.

pkg/engine/engine_test.go
ahrav: I made a mistake; the screenshot above was from a detector benchmark in `aws_test.go`. This benchmark should compare the performance of `FindDetectorMatch` across different chunk sizes. I intended to show that using a larger chunk size (64kB) can increase throughput without significantly affecting the operation's duration. This was my basis for suggesting an increase in the default chunk size from 10kB to 64kB.

Here is the correct benchmark:
Screenshot 2024-05-16 at 2 49 06 PM
:

rgmz:
> I think we can increase the default to 1024 bytes for the multi-part credential case. Should be adequate in most cases. I'd like to see this optimization on by default, but we can provide an opt-out flag.

1024 bytes would be safer, but could definitely still miss valid secrets.

I think there'd be tremendous value in at least having a (hidden?) flag that runs both and checks for missed results (kind of like https://github.com/github/scientist), rather than binary on/off.

pkg/engine/engine_test.go
ahrav: I made a mistake; the screenshot above was from a detector benchmark in `aws_test.go`. This benchmark should compare the performance of `FindDetectorMatch` across different chunk sizes. 

Here is the correct benchmark:
Screenshot 2024-05-16 at 3 40 02 PM

dustin-decker:
> > I think we can increase the default to 1024 bytes for the multi-part credential case. Should be adequate in most cases. I'd like to see this optimization on by default, but we can provide an opt-out flag.
> 
> 1024 bytes would be safer, but could definitely still miss valid secrets.
> 
> I think there'd be tremendous value in at least having a (hidden?) flag that runs both and checks for missed results (kind of like https://github.com/github/scientist), rather than binary on/off. Otherwise it can be tricky to identify affected detectors, which has been an issue with the verification overlap change.

@rgmz, this idea is implemented in https://github.com/trufflesecurity/trufflehog/pull/2918
Great suggestion!

dustin-decker (approved):
Great improvement!

PR #2825, opened 2024-05-10
[refactor] - Create separate handler for non-archive data

ahrav:
### Description:
This PR splits the original default handler into two separate handlers: default and archive. The default handler now processes non-archive files, and the archive handler manages all archive file types supported by the `archiver` library.

Additionally, this PR restores the `git cat file` functionality for binary diffs to a streaming method. This change is due to `HandleFile` now properly consuming the entire reader before beginning file processing.

Note: Most of the _important_ changes are in `handlers.go`.

### Checklist:


pkg/handlers/archive.go
```diff
@@ -0,0 +1,192 @@
+package handlers
```
ahrav, line 1: Most of this code was moved from the old default handler.

pkg/handlers/handlers.go
```diff
@@ -1,44 +1,99 @@
...
+	defer func() {
+		if err != nil && rdr != nil {
+			_ = rdr.Close()
+		}
+	}()
+
+	format, arReader, err := archiver.Identify("", rdr)
+	switch {
+	case err == nil: // Archive detected
+		reader.isArchive = true
```
rosecodym, line 59: I find this field name confusing because it's only set to true for certain types of archives. I'm not sure what a better one would be, though - maybe `isGenericArchive` or `isNonSpecializedArchive` or something like that?

rosecodym (commented):
I have a non-blocking thought about a possible different way to structure the code (that you may have already considered) and a question about a suspicious comment.

pkg/handlers/handlers.go
```diff
@@ -1,44 +1,99 @@
...
+// used by FileHandler implementations to make decisions on how to process the file.
+//
+// By encapsulating the file type detection logic, fileReader simplifies the implementation of FileHandler and
+// promotes a more cohesive and maintainable codebase. It also embeds a BufferedFileReader to provide efficient
+// random access to the file content.
+type fileReader struct {
+	format   archiver.Format
+	mimeType mimeType
+	*readers.BufferedFileReader
+	isArchive bool
```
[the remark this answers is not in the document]
ahrav: It should also be used in `archive.go` `L88-91` [here](https://github.com/trufflesecurity/trufflehog/pull/2825/files#diff-ad1adabfbf61332dbddae0d14b27a18a8f23b9f372874e1e340ebb0b2cbc54daR88-R91), sorry I should've explicitly called it out since the rest of the code was a lift and shift.

pkg/handlers/rpm.go
```diff
@@ -11,18 +11,18 @@ import (
 	logContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
-// rpmHandler specializes defaultHandler to manage RPM package files. It leverages shared behaviors
-// from defaultHandler and introduces additional logic specific to RPM packages.
+// rpmHandler specializes archiveHandler to manage RPM package files. It leverages shared behaviors
+// from archiveHandler and introduces additional logic specific to RPM packages.
```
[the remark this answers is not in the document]
ahrav: My mistake, thanks for catching.

pkg/handlers/handlers.go
ahrav: That's a valid point. I will rename it to `isGenericArchive`.

ahrav:
> I have a non-blocking thought about a possible different way to structure the code (that you may have already considered) and a question about a suspicious comment.

Replied inline.

rosecodym (approved):
🎉

PR #2827, opened 2024-05-10
Docker scan errors should be surfaced to the user beyond just logs

dustin-decker:
### Description:
Currently if a Docker image scan fails due to invalid auth in TruffleHog Enterprise, the source integration is still showing as healthy because the error is not surfaced. This PR returns the errors.

### Checklist:

### DIFF
--- a/pkg/sources/docker/docker.go
+++ b/pkg/sources/docker/docker.go
@@ -132,6 +132,7 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 	_ = workers.Wait()
 	if scanErrs.Count() > 0 {
 		ctx.Logger().V(2).Info("scan errors", "errors", scanErrs.String())
+		return errors.New(scanErrs.String())
 	}
 
 	return nil


pkg/sources/docker/docker.go
```diff
@@ -132,6 +132,7 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 	_ = workers.Wait()
 	if scanErrs.Count() > 0 {
 		ctx.Logger().V(2).Info("scan errors", "errors", scanErrs.String())
+		return errors.New(scanErrs.String())
```
ahrav, line 135: Question: Are we certain this approach is what we want? Adopting this logic means any single error during the scan will lead to its failure. We've encountered this issue with other sources and found it detrimental to the user experience. However, the situation with the Docker source might be different, and perhaps we are willing to accept that trade-off

rosecodym: As an alternative, implementing `SourceUnitEnumChunker` would provide a way to differentiate between fatal and non-fatal errors. It would be a little more work than adding this single line, but it doesn't look like it would be too much more, and it's work we want to do anyway.l

PR #2829, opened 2024-05-10
[feat] - additional buffer pool

ahrav:
### Description:
This PR introduces a second buffer pool alongside the existing one to optimize the handling of different content types when dealing with archives. The original buffer pool, with a default size of 4kB, will be used for diff content, while the newly added buffer pool will have a larger buffer size of 64kB and will be dedicated to file content.

This change is based on tests conducted across ~120k repositories, which showed that most files handled during archive operations exceeded 4kB. Consequently, there was a significant increase in operations needed to expand the underlying buffer. By maintaining separate buffer pools for diff content and file content, we can reduce the overhead associated with constructing BufferedFileWriters and improve overall performance.

### Checklist:


pkg/buffers/buffer/buffer.go
```diff
@@ -29,7 +29,7 @@ func (b *Buffer) RecordMetric() {
 	dur := time.Since(b.checkedOutAt)
 	checkoutDuration.Observe(float64(dur.Microseconds()))
 	checkoutDurationTotal.Add(float64(dur.Microseconds()))
-	totalBufferSize.Add(float64(b.Len()))
+	totalBufferSize.Add(float64(b.Cap()))
```
rosecodym, line 32: you love sneaking in these lil' bug fixes

pkg/buffers/pool/pool.go
```diff
@@ -25,25 +25,19 @@ func (poolMetrics) recordBufferReturn(buf *buffer.Buffer) {
...
 }
 
-// Opts is a function that configures a BufferPool.
-type Opts func(pool *Pool)
-
 // Pool of buffers.
 type Pool struct {
 	*sync.Pool
-	bufferSize uint32
+	bufferSize int
```
rosecodym, line 31: Why did you change this?

pkg/writers/buffer_writer/bufferwriter.go
```diff
@@ -1,4 +1,4 @@
-// Package bufferwriter provides a contentWriter implementation using a shared buffer pool for memory management.
+// Package bufferwritter provides a contentWriter implementation using a shared buffer pool for memory management.
```
rosecodym, line 1: typo

pkg/writers/buffer_writer/bufferwriter.go
```diff
@@ -44,13 +45,21 @@ type BufferWriter struct {
...
 func (b *BufferWriter) Write(data []byte) (int, error) {
 	if b.state != writeOnly {
 		return 0, fmt.Errorf("buffer must be in write-only mode to write data; current state: %d", b.state)
 	}
+	if b.buf == nil {
+		b.buf = b.bufPool.Get()
+		if b.buf == nil {
+			b.buf = buffer.NewBuffer()
+		}
+	}
```
rosecodym, line 62: a block so nice you wrote it twice

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -285,6 +329,6 @@ func newAutoDeletingFileReader(file *os.File) *autoDeletingFileReader {
 
 // Close implements the io.Closer interface, deletes the file after closing.
 func (r *autoDeletingFileReader) Close() error {
-	defer os.Remove(r.Name()) // Delete the file after closing
+	defer os.RemoveAll(r.Name()) // Delete the file after closing
```
rosecodym, line 332: Why `RemoveAll`? I thought these were individual files.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
```diff
@@ -67,28 +85,51 @@ func WithThreshold(threshold uint64) Option {
...
+	if w.bufPool == nil {
+		w.bufPool = defaultBufferPool
+	}
+
 	return w
 }
 
 // NewFromReader creates a new instance of BufferedFileWriter and writes the content from the provided reader to the writer.
 func NewFromReader(r io.Reader, opts ...Option) (*BufferedFileWriter, error) {
+	opts = append(opts, WithBufferSize(Large))
```
rosecodym, line 123: If I'm reading this right, you only ever use the explicit option to use the large buffer pool, and get the default one via default method calls. I think this is a little awkward in a way that could be remediated by making the option a _boolean_ called something like `WithUseLargeBuffer`. Did you consider something like that?

pkg/buffers/buffer/buffer.go
ahrav: Ahh it's because the base PR actually reverted this change. I'll update the base and this should get fixed. Thanks for catching it.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Yes, I did consider that approach. I thought making it configurable might be appropriate, anticipating additional sizes for future use cases. However, I may have jumped the gun and overcomplicated the solution.

pkg/buffers/pool/pool.go
ahrav: The main reason is convenience. Using an `int` instead of a `uint32` eliminates the need to cast in all the places where it's used.

pkg/writers/buffered_file_writer/bufferedfilewriter.go
ahrav: Didn't realize I changed this. Thanks.

pkg/writers/buffer_writer/bufferwriter.go
ahrav: Whoops. Removed. Thank you.

dustin-decker (approved):

PR #2833, opened 2024-05-11
Update SendGrid detector

rgmz:
### Description:
This PR:
1. Makes the detector use tri-state verification
2. Removes the `sendgrid` prefix regex, as sendgrid tokens already have a distinct pattern
3. Changes the test endpoint to `/v3/scopes` and adds the scopes as extra data

### Checklist:

### DIFF
--- a/pkg/detectors/sendgrid/sendgrid.go
+++ b/pkg/detectors/sendgrid/sendgrid.go
@@ -2,10 +2,13 @@ package sendgrid
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
+	"io"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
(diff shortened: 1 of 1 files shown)


ahrav (approved):
Thank you for getting this updated ❤️

pkg/detectors/sendgrid/sendgrid.go
```diff
@@ -36,53 +39,27 @@ func (s Scanner) Keywords() []string {
...
-
-	for _, match := range matches {
-		if len(match) != 2 {
-			continue
-		}
-		resMatch := strings.TrimSpace(match[1])
+	uniqueMatches := make(map[string]struct{})
+	for _, match := range keyPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueMatches[match[0]] = struct{}{}
+	}
```
bugbaba, line 45: This is to avoid verifying the same matches found in same chunk multiple times right? 
@ahrav if this is helpful shouldn't we have this in all detectors then? Maybe another `[chore]`  task for you :smiling_face_with_tear:

rgmz: It's a bit of a bandaid solution to #2262. I mentioned this as well in https://github.com/trufflesecurity/trufflehog/pull/2812#discussion_r1594877538.

ahrav: Yea, we're aware of the task and are considering moving the caching logic into the engine to streamline processes by centralizing the logic, which would eliminate the need to replicate it across all detectors. For now, implementing the caching within each individual detector is an acceptable temporary solution.

rgmz: IMO it's probably another use case for SQLite, similar to #2696.

PR #2835, opened 2024-05-13
Add "Intra42" detector

alexandregv:
### Description:
This PR adds a new detector: `Intra42` (987)
Intra42 is the intranet and API of "42 School", a Computer-Science school.  
The API is open to all students in 57 campus around the world. It is thus often subject to credentials leaks on GitHub.  
Sensitive data is available, like administrative information, scolarship, exam grades, etc. Actions are also possible on the account of the leaked secret. If the app has elevated rights and is leaked (which happened), it can read AND modify all students data, administrative or pedagogical. The Intellectual Property (pedagogical content) may also be retrieved from the API.

OAuth2 verification is implemented, and tests with simple pattern test and some verification tests, all passing.

Test repo (with revoked secret): https://github.com/zoirj0n/42_AbuDhabi
Screenshot from 2024-05-13 01-49-35
I noticed that on this screen, "Detector Type" appears as "987" instead of the text value "Intra42", is it normal? It seems that other detectors get their text name correctly.

Public website: https://42.fr
API doc: https://api.intra.42.fr/ (OAuth2)

### Checklist:

### DIFF
--- a/pkg/detectors/intra42/intra42.go
+++ b/pkg/detectors/intra42/intra42.go
@@ -35,7 +35,7 @@ var (
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"intra", "intra42", "secret", "s-s4t2ud-", "s-s4t2af-"}
+	return []string{"s-s4t2ud-", "s-s4t2af-"}
 }
 
 // FromData will find and optionally verify Intra42 secrets in a given set of bytes.
--- a/pkg/detectors/intra42/intra42.go
+++ b/pkg/detectors/intra42/intra42.go
@@ -28,8 +28,8 @@ var (
 	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"intra", "intra42", "client", "secret"}) + `\b(s-s4t2(?:ud|af)-[abcdef0123456789]{64})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"intra", "intra42", "client", "id"}) + `\b(u-s4t2(?:ud|af)-[abcdef0123456789]{64})\b`)
+	keyPat = regexp.MustCompile(`\b(s-s4t2(?:ud|af)-[abcdef0123456789]{64})\b`)
+	idPat  = regexp.MustCompile(`\b(u-s4t2(?:ud|af)-[abcdef0123456789]{64})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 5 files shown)


alexandregv:
A verified secret test (blurred and now revoked):  
Screenshot from 2024-05-13 02-12-10

pkg/detectors/intra42/intra42.go
```diff
@@ -0,0 +1,117 @@
...
+// Ensure the Scanner satisfies the interface at compile time
+var _ detectors.Detector = (*Scanner)(nil)
+
+const verifyURL = "https://api.intra.42.fr/oauth/token"
+
+var (
+	defaultClient = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"intra", "intra42", "client", "secret"}) + `\b(s-s4t2(?:ud|af)-[abcdef0123456789]{64})\b`)
```
ahrav, line 31: suggestion: We could simplify the alphanumeric portion of the regex from `abcdef0123456789` -> `a-f0-9`. We could also make the same update below.

We can also likely remove the `detectors.PrefixRegex` here since the keywords are already a prefix that is part of the key.

Ex:
```
keyPat = regexp.MustCompile(`\b(s-s4t2(?:ud|af)-[a-f0-9]{64})\b`)
idPat  = regexp.MustCompile(`\b(u-s4t2(?:ud|af)-[a-f0-9]{64})\b`)
```

pkg/detectors/intra42/intra42.go
```diff
@@ -0,0 +1,117 @@
...
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"intra", "intra42", "client", "id"}) + `\b(u-s4t2(?:ud|af)-[abcdef0123456789]{64})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"intra", "intra42", "secret", "s-s4t2ud-", "s-s4t2af-"}
```
ahrav, line 38: suggestion: I suggest using only `s-s4t2ud-` and `s-s4t2af-` as keywords since they are unique and already serve as prefixes. Including `secret` and `intra` could lead to unnecessary matches that will require filtering out through a costly regex operation later.

ahrav (approved):
Thanks a lot @alexandregv we really appreciate the contribution. 😄

pkg/detectors/intra42/intra42.go
alexandregv: Totally agree for both! I went with PrefixRegex because the example used it, without thinking of removing it. Will update.

pkg/detectors/intra42/intra42.go
alexandregv: I wasn't sure if the keywords should be words outside of the regex or just any word in the files. Seems its the later, so I will keep only the prefixes as suggested. Thanks!

PR #2844, opened 2024-05-13
Update the FetchRSS detector

rgmz:
### Description:
The previous logic resulted in false positives if an error other than "Not authorised" was returned.

```
Found verified result 🐷🔑
Detector Type: Fetchrss
Decoder Type: PLAIN
Raw result: .impl.uri.rules.ResourceClassRule.accept
Commit: 0b3c625b155c993fda14f97759fcea7278ff2a8e
...
```

### Checklist:

### DIFF
--- a/pkg/detectors/fetchrss/fetchrss.go
+++ b/pkg/detectors/fetchrss/fetchrss.go
@@ -2,26 +2,30 @@ package fetchrss
 
 import (
 	"context"
-	regexp "github.com/wasilibs/go-re2"
+	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
-	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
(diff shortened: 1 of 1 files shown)


pkg/detectors/fetchrss/fetchrss.go
```diff
@@ -73,6 +66,51 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	res, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	// The API seems to always return a 200 status code.
```
ahrav, line 84: Sigh, it makes me so sad when APIs don't use status codes correctly. 😢

ahrav (approved):
Thanks for getting this updated.

PR #2854, opened 2024-05-16
[bug] - Handle empty reader case in newFileReader

### LINKED ISSUE #2853: Failing to scan archived files
### TruffleHog Version
3.76.0

### Expected Behavior

It should unarchive the files and scan.

### Actual Behavior

Scan is failing when trying to unarchive the .zip, .gz, .tat.gz etc files.

### Steps to Reproduce

1. Try to scan a repo with lot of different types of archive file.
2. Error:

2024-05-15T13:12:35-04:00	error	trufflehog	error unarchiving chunk.	{"source_manager_worker_id": "W7Rg1", "unit": "/var/folders/7h/bpqg847x6xx5zt42f9sbj1040000gq/T/trufflehog-30084-2945869955", "unit_kind": "dir", "commit": "5736d30", "path": "aspnetcore-runtime-8.0.1-linux-x64.tar.gz", "timeout": 30, "error": "error extracting archive with format: .tar: archive/tar: invalid tar header"}

## Additional Context
This is an intermittent issue but happens very often.

ahrav:
### Description:
This PR introduces changes to handle the scenario where `newFileReader` receives an empty reader. Previously, when an empty reader was passed to `newFileReader`, it would result in an error and halt processing of that archive.

Changes:
- Introduce a new sentinel error `ErrEmptyReader` to represent the case when the reader is empty.
- Modify the `newFileReader` function to check if the size of the `BufferedFileReader` is zero, indicating an empty reader.
  - If the reader is empty, return a zero value reader and the `ErrEmptyReader` error.
- Update the usage of `newFileReader` to handle the `ErrEmptyReader` error explicitly.
  - If `ErrEmptyReader` is returned, log a message indicating that the reader is empty and skip the file by returning `nil`.

### Checklist:


pkg/handlers/archive.go
```diff
@@ -21,7 +21,7 @@ const (
 
 var (
 	maxDepth   = 5
-	maxSize    = 250 * 1024 * 1024 // 250 MB
+	maxSize    = 2 << 30 // 2 GB
```
ahrav, line 24: Reverted this to the old value. This should no longer be an issue since we **do not** read the entire file into memory.

ahrav: I considered removing it, but to preserve backward compatibility, I left it in. This value can be configured via a command-line flag, so removing it isn't straightforward. 😞

dustin-decker (approved):

PR #2855, opened 2024-05-16
[refactor] - remove redundant chunking

ahrav:
### Description:
This PR removes redundant chunking from the engine. All chunks supplied to the ChunksChan are already sized appropriately.

### Checklist:


rosecodym (approved):
Why were we doing this before?

dustin-decker (approved):

ahrav:
> Why were we doing this before?



> Why were we doing this before?

Originally, sources would place their data into a chunk directly. Later, we changed this to use our `chunkReader`, which chunks the data before creating a chunk. However, we ended up with chunking in both places, rendering the chunking in the engine ineffective. This redundancy wasted cycles, created and destroyed goroutines, and added an unnecessary level of indentation 🤣.

PR #2863, opened 2024-05-17
Update OpenAI detector

rgmz:
### Description:
User-level API keys have been deprecated, now project-level API keys (`sk-proj-`) are the default. They have also introduced service account API keys (`sk-$name-`).

https://help.openai.com/en/articles/9186755-managing-your-work-in-the-api-platform-with-projects#h_20805964da

### Checklist:

### DIFF
--- a/pkg/detectors/openai/openai.go
+++ b/pkg/detectors/openai/openai.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
+	"io"
 	"net/http"
 	"strconv"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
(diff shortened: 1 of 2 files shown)


pkg/detectors/openai/openai.go
```diff
@@ -103,6 +70,67 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	if err != nil {
+		return false, nil, err
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case 200:
```
abmussani, line 93: @rgmz previously verification was based on range of 2xx Http status. Is there any particular reason to change it ?

pkg/detectors/openai/openai.go
```diff
@@ -103,6 +70,67 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+			"id":          org.ID,
+			"title":       org.Title,
+			"user":        org.User,
+			"description": org.Description,
+			"role":        org.Role,
+			"is_personal": strconv.FormatBool(org.Personal),
+			"is_default":  strconv.FormatBool(org.Default),
+			"total_orgs":  fmt.Sprintf("%d", len(orgs.Data)),
+		}
+		return true, extraData, nil
```
abmussani, line 110: @rgmz Metadata here is an additional information and the credentials are already been verified based on HTTP Status code. Base on the conversation we had on [#2807 ](https://github.com/trufflesecurity/trufflehog/pull/2807#discussion_r1594184573) and [#2808 ](https://github.com/trufflesecurity/trufflehog/pull/2808#discussion_r1596810738), these are non-fatal errors and can be ignored without affecting the verification result.

abmussani (commented):
Thanks @rgmz for updating the open-ai detector. I've left some comments.

pkg/detectors/openai/openai.go
rgmz: Many of the detectors are based on a template with generic conditions. I think it's prudent to only consider something verified if it matches known/documented status codes.

Unexpected API/site changes, and thus status code changes, will propagate as an error rather than creating false-positives.

abmussani: The API reference of OpenAI has well defined the [error codes](https://platform.openai.com/docs/guides/error-codes) (also In there [python library](https://github.com/openai/openai-python/blob/main/src/openai/_client.py)). In my opinion, Its good to parse the response but It wont make any effect on verification.

pkg/detectors/openai/openai.go
rgmz: > ..these are non-fatal errors and can be ignored without affecting the verification result.

The question I posed was actually whether it makes sense to return "verified" alongside an error to indicate a potential problem with the response. Generally speaking, I think it's much safer to handle errors rather than ignore them `if err == nil { ... }`.

There's a larger problem of surfacing errors and giving them precedence, but that's a different topic. _e.g.,_ the GitHub detector failing due to a DNS timeout and the JDBC detector failing because of an invalid hosts shouldn't be treated equally.

pkg/detectors/openai/openai.go
rgmz: > In my opinion, Its good to parse the response but It wont make any effect on verification.

I agree that parsing error codes doesn't make sense for OpenAI (although there are some APIs where 401 vs 403 is an importance difference).

I don't think it's a good practice to be liberal with what status codes are considered "valid" when only a specific one is expected. It's a recipe for false positives, imo.

pkg/detectors/openai/openai.go
abmussani: You are right that these errors should not be ignored. Unfortunately, right now, Detector result struct does not supports to hold these kind of errors (other than verification). In future, we can have separate error field(s) like @dustin-decker proposed to name it as EnrichmentError.

rgmz: > Unfortunately, right now, Detector result struct does not supports to hold these kind of errors (other than verification).

I would still consider them verification errors. e.g., invalid JSON can be indicative of false-positives or the API not behaving as expected (#2099, #2218).

abasit-folio3: Thanks @rgmz for sharing this case study. It is valid use case to convince me. One more case came in my mind is if an API started to redirected to valid page (with status 200) with HTML in response (may be home page), It will also be indicative of false-positives.

abmussani: Thanks @rgmz for sharing this case study. It is valid use case to convince me. One more case came in my mind is if an API started to redirected to valid page (with status 200) with HTML in response (may be home page), It will also be indicative of false-positives.

abmussani (approved):
Thank you. LGTM

dustin-decker (approved):

PR #2866, opened 2024-05-20
Update regex for Organization in Azure Devops detector

abmussani:
### Description:
As per [AzureDevOps documentation](https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/rename-organization?view=azure-devops) about Organization renaming, The old regex is not accommodating hyphens and length conditions. This PR allows to detect correct naming of organization.

Here are some key-points from documentation about organization renaming:

      Adhere to the following guidelines when you create an organization name.

      - Use only letters from the English alphabet
      - Start your organization name with a letter or number
      - Use letters, numbers, or hyphens after the initial character
      - Ensure that your organization doesn't exceed 50 Unicode characters
      - End with a letter or number

      If you use any of the disallowed characters, you get the following error message: VS850015: The specified name is not allowed to be used: {Organization name}.

### Checklist:


zricethezav (approved):
Very nice. Thanks for linking to the documentation that outlines this credential's requirements!

PR #2868, opened 2024-05-21
Update regex for OpenAI's API key 

abmussani:
### Description:
Recently for testing I generated two kind of OpenAI’s api keys, user and service based. Both of them were not detected with the current regex.

I could not find any documentation but This is what I observed:

> 1) for user based api keys, sk-proj- is being used as prefix in keys.
> 2) for service based keys, sk-{service_name} is being used as prefix. The service name can have letters, numbers & hyphen. I manually hit and try for the length. on more than 42 characters, I got invalid service ID error. under 42, portal generated the service api key. To be on safe side, Regex allows at max 50 characters. 

To support old key format, minimum 0 is set {0,50}.

### Checklist:

### DIFF
--- a/pkg/detectors/openai/openai.go
+++ b/pkg/detectors/openai/openai.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
 	"net/http"
 	"strconv"
+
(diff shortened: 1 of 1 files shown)


rgmz:
BTW this is covered by #2863, which adds unit tests and a stricter regex taken from their JavaScript code.

dustin-decker:
Closed in favor of #2863

PR #2873, opened 2024-05-22
adding Groq detector

0x1:
### Description:
Groq has become a very popular AI tool in the last few months, especially among startups and enterprise. They provide cloud inference with their own custom chips. https://console.groq.com/docs/quickstart

image

### Checklist:


pkg/detectors/groq/groq.go
```diff
@@ -0,0 +1,97 @@
...
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	keyPat        = regexp.MustCompile(`\b(gsk_[a-zA-Z0-9]{52})\b`)
+)
+
+func (s Scanner) Keywords() []string {
+	return []string{"groq"}
```
ahrav, line 29: question: Do you think we could just use `gsk_` as the keyword?

ahrav (approved):
Nice 🥳

pkg/detectors/groq/groq.go
0x1: ya makes sense

PR #2879, opened 2024-05-24
Update metadata for DataDog for API + APPKey

abmussani:
### Description:
In case of API + APP keys, the [API](https://docs.datadoghq.com/api/latest/users/#list-all-users) returns list of user belongs to organization with their emails along with optional information of organization.

This PR includes following information in metadata:
- Email addresses of user which are verified, not disabled and not service account. If there is no user exists of this kind then includes the email of first user.
- If there is organization information then include name and url in metadata.

### Checklist:


dustin-decker (approved):
Nice, thank you!

PR #2889, opened 2024-05-27
[feat] - Introduce `channelmetrics` Package for Channel Metrics Collection

ahrav:
### Description:
This PR introduces a new `channelmetrics` package that provides functionality to collect and record various metrics related to channel operations. This enhancement will help in monitoring the performance and utilization of Go channels, allowing for better optimization and troubleshooting.

### Additions

1. **MetricsCollector:**
   - Interface for recording channel metrics:
     - `produceDuration`: Histogram for production time.
     - `consumeDuration`: Histogram for consumption time.
     - `channelLen`: Gauge for current buffer size.
     - `channelCap`: Gauge for buffer capacity.

2. **NewMetricsCollector:**
   - Creates a `MetricsCollector` with metrics organized by `namespace`, `subsystem`, and `chanName`.
   - Ensures a `MetricsCollector` is initialized only once per unique combination of `chanName`, `namespace`, and `subsystem` using a map to track instances and `sync.Once` for thread-safe initialization.

3. **ObservableChan:**
   - Wraps a Go channel to collect metrics:
     - `Close()`: Records buffer size on close.
     - `Send(ctx, item)`: Records send duration and buffer size.
     - `Recv(ctx)`: Records receive duration and buffer size.

### Example Metrics
Screenshot 2024-05-27 at 12 16 28 PM

### Checklist:


pkg/channelmetrics/observablechan.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Close closes the channel and records the current size of the channel buffer.
+func (oc *ObservableChan[T]) Close() {
+	close(oc.ch)
+	oc.RecordChannelLen()
+}
+
+// Send sends an item into the channel and records the duration taken to do so.
+// It also updates the current size of the channel buffer.
+func (oc *ObservableChan[T]) Send(_ context.Context, item T) {
```
ahrav, line 53: ctx is unused for now, same for `Recv`

dustin-decker:
This is cool!

pkg/channelmetrics/observablechan.go
dustin-decker: Combine with CancelableWrite? https://github.com/trufflesecurity/trufflehog/blob/024b219dfbb18442d71da5c1bdfcdc93920e41d0/pkg/common/context.go#L17

dustin-decker (approved):
Good stuff, I think we should include cancelable write and receive though.

ahrav:
> Good stuff, I think we should include cancelable write and receive though.

Yea, I like that idea. Thanks.

pkg/channelmetrics/observablechan.go
```diff
@@ -0,0 +1,84 @@
...
+	RecordChannelCap(capacity int)
+}
+
+// ObservableChan wraps a Go channel and collects metrics about its usage.
+// It supports any type of channel and records metrics using a provided
+// MetricsCollector implementation.
+type ObservableChan[T any] struct {
+	ch        chan T
+	metrics   MetricsCollector
+	bufferCap int
```
mcastorina, line 28: Why is this its own struct member? Should it always be equal to `cap(ch)`?

pkg/channelmetrics/observablechan.go
```diff
@@ -0,0 +1,84 @@
...
+// using the provided MetricsCollector and channel name. The chanName is used to
+// distinguish between metrics for different channels by incorporating it into
+// the metric names.
+func NewObservableChan[T any](ch chan T, metrics MetricsCollector) *ObservableChan[T] {
+	oChan := &ObservableChan[T]{
+		ch:        ch,
+		metrics:   metrics,
+		bufferCap: cap(ch),
+	}
+	oChan.RecordChannelCapacity() // Record capacity immediately
```
mcastorina, line 42: Should we also record the length immediately? In most cases it'll be 0, but theoretically any channel can be converted into an `ObservableChan`.

pkg/channelmetrics/observablechan.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Recv receives an item from the channel and records the duration taken to do so.
+// It also updates the current size of the channel buffer.
+func (oc *ObservableChan[T]) Recv(_ context.Context) T {
+	startTime := time.Now()
+	defer func() {
+		oc.metrics.RecordConsumeDuration(time.Since(startTime))
+		oc.RecordChannelLen()
+	}()
```
mcastorina, line 72: **Nit:**

```go
defer func(start time.Time) {
    // ...
}(time.Now())
```

pkg/channelmetrics/observablechan.go
```diff
@@ -0,0 +1,84 @@
...
+	return <-oc.ch
```
mcastorina, line 73: Cancellable recv?

pkg/channelmetrics/metrics_collector/prometheus/collector.go
```diff
@@ -0,0 +1,111 @@
...
+// you can identify bottlenecks, optimize performance, and ensure that the ObservableChan is operating efficiently.
+func NewMetricsCollector(chanName, namespace, subsystem string) *MetricsCollector {
+	key := fmt.Sprintf("%s_%s_%s", namespace, subsystem, chanName)
+
+	collectorsMu.Lock()
+	defer collectorsMu.Unlock()
+
+	if collector, exists := collectors[key]; exists {
+		return collector
+	}
```
mcastorina, line 59: Is there ever a case where we want `NewMetricsCollector` to return an existing collector? Or would this indicate we have a name collision?

ahrav: That would result in a collision, correct.

pkg/channelmetrics/observablechan.go
ahrav: I didn't account for that use case, thanks.

pkg/channelmetrics/observablechan.go
ahrav: Do we have an existing CancellableRecv, or are you suggesting I implement it here?

mcastorina: No existing one that I know of.

pkg/channelmetrics/metrics_collector/prometheus/collector.go
mcastorina: Is a collision something that should never happen?

PR #2892, opened 2024-05-28
Add Jenkins scanning

dustin-decker:
### Description:
Adds jenkins build log scanning

### Checklist:


main.go
```diff
@@ -661,6 +667,16 @@ func run(state overseer.State) {
 		if err := e.ScanElasticsearch(ctx, cfg); err != nil {
 			logFatal(err, "Failed to scan Elasticsearch.")
 		}
+	case jenkinsScan.FullCommand():
+		cfg := engine.JenkinsConfig{
+			Endpoint:              *jenkinsEndpoint,
```
joeleonjr, line 672: I think `jenkinsEndpoint` should be renamed to `jenkinsURL` to match the CLI flag arg name.

pkg/engine/jenkins.go
```diff
@@ -0,0 +1,77 @@
...
+	err := anypb.MarshalFrom(&conn, connection, proto.MarshalOptions{})
+	if err != nil {
+		ctx.Logger().Error(err, "failed to marshal Jenkins connection")
+		return err
+	}
+
+	sourceName := "trufflehog - Jenkins"
+	sourceID, jobID, _ := e.sourceManager.GetIDs(ctx, sourceName, circleci.SourceType)
+
+	circleSource := &circleci.Source{}
```
joeleonjr, line 71: Looks like this is referencing circleCI instead of jenkins. I think this should fix it:

```
jenkinsSource := &jenkins.Source{}
	if err := jenkinsSource.Init(ctx, "trufflehog - Jenkins", jobID, sourceID, true, &conn, runtime.NumCPU()); err != nil {
		return err
	}
	_, err = e.sourceManager.Run(ctx, sourceName, jenkinsSource)
```

dustin-decker: Oops, thanks

pkg/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,273 @@
+package roundtripper
```
ahrav, line 1: question: Would it be worth either moving this code to the common pkg with `http.go` or move that code here? I find it a little confusing to have such similar code in two different locations.

ahrav (approved):

zricethezav (approved):

juandediosflores114-hash (approved):

PR #2897, opened 2024-05-29
Update Zulip detector

rgmz:
### Description:
This makes the Zulip detector tri-state and adds some additional logic to prevent false positives.

Previously, [this file](https://github.com/zulip/zulip/blob/5df0e7b857b37beb2999cf94606c02c30310f08c/zulip/templates/zerver/integrations.html#L182) was triggered a false "verified" result.
```
✅ Found verified result 🐷🔑
Detector Type: ZulipChat
Decoder Type: PLAIN
Raw result: 0123456789abcdef0123456789abcdef
Domain: www.zulip.org
Id: zulip-devel@googlegroups.com
File: /tmp/integrations.html
Line: 181
```

### Checklist:

### DIFF
--- a/pkg/detectors/zulipchat/zulipchat.go
+++ b/pkg/detectors/zulipchat/zulipchat.go
@@ -2,10 +2,12 @@ package zulipchat
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


dustin-decker (approved):
Thank you

PR #2905, opened 2024-06-02
Add a flag to disable ssl verify

rgmz:
### Description:
This creates an experimental flag to disable SSL verification for detectors. This is useful in corporate environments where [MITMing SSL traffic is common](https://security.stackexchange.com/questions/107542/is-it-common-practice-for-companies-to-mitm-https-traffic) and certificates are replaced with self-signed ones.

This implementation is an incomplete POC.

- [ ] There may not be total coverage for http clients
- [ ] Detectors that rely on third-party libraries likely need to be configured separately. 

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -81,6 +81,7 @@ var (
 	includeDetectors     = cli.Flag("include-detectors", "Comma separated list of detector types to include. Protobuf name or IDs may be used, as well as ranges.").Default("all").String()
 	excludeDetectors     = cli.Flag("exclude-detectors", "Comma separated list of detector types to exclude. Protobuf name or IDs may be used, as well as ranges. IDs defined here take precedence over the include list.").String()
 	jobReportFile        = cli.Flag("output-report", "Write a scan report to the provided path.").Hidden().OpenFile(os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
+	SslVerify            = cli.Flag("ssl-verify", "Whether to verify the SSL certificates when making requests.").Default("true").Bool()
 
 	noVerificationCache = cli.Flag("no-verification-cache", "Disable verification caching").Bool()
 
@@ -300,7 +301,7 @@ func init() {
 
 	cmd = kingpin.MustParse(cli.Parse(os.Args[1:]))
 
-	// Configure logging.
+	// Configure log level.
 	switch {
 	case *trace:
 		log.SetLevel(5)
@@ -324,6 +325,10 @@ func init() {
 
 	if *noColor || *noColour {
 		color.NoColor = true // disables colorized output
+	}
+	// Disable certificate validation, if specified.
+	if !*SslVerify {
+		feature.NoVerifySsl.Store(true)
 	}
 }
 
--- a/pkg/common/http.go
+++ b/pkg/common/http.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/feature"
 )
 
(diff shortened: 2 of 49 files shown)


zricethezav:
Slapping a draft label on this since it's an incomplete POC

rgmz:
It could use a review to determine whether it's worth putting any more effort into.

ahrav:
> It could use a review to determine whether it's worth putting any more effort into.

I like the idea, but I'm a bit nervous about the global variable. Is there a way to use the flag to construct the transport instead? I'm concerned that it's too easy for anyone to reach into the transport and change the TLSConfig. That being said, if others are okay with this approach, I won't object. 😅

rgmz:
> I like the idea, but I'm a bit nervous about the global variable. 

It's a bit of a smell. I couldn't think of an easier way to propagate such a value to multiple packages before the dependent code initializes.

rgmz:
Unfortunately, it seems this isn't possible. The hundreds of detectors with logic like `defaultClient = common.SaneHttpClient()` means that [`SaneHttpClient()`](https://github.com/trufflesecurity/trufflehog/blob/e6cc78b95ee7101193f7db828a32f8135306c28f/pkg/common/http.go#L229) gets invoked well before `init` in main.go.


As far as I'm aware, there's no equivalent to something like [Kotlin's `lazy { }` initializer](https://kotlinlang.org/docs/delegated-properties.html#lazy-properties). Would it be feasible to initialize the client on each `FromData` call instead?

rgmz:
> Unfortunately, it seems this isn't possible. The hundreds of detectors with logic like `defaultClient = common.SaneHttpClient()` means that [`SaneHttpClient()`](https://github.com/trufflesecurity/trufflehog/blob/e6cc78b95ee7101193f7db828a32f8135306c28f/pkg/common/http.go#L229) gets invoked well before `init` in main.go.
> 
> As far as I'm aware, there's no equivalent to something like [Kotlin's `lazy { }` initializer](https://kotlinlang.org/docs/delegated-properties.html#lazy-properties). Would it be feasible to initialize the client on each `FromData` call instead?

For this to work, the top-level var would need to be replaced with a lazy initialization.

https://github.com/trufflesecurity/trufflehog/blob/316316e60ef94093c481d6b70b402a253dc55ad5/pkg/detectors/scrapingbee/scrapingbee.go#L61-L63

rosecodym (changes requested):
_In theory_, if you could toggle `ssl-verify` while the scan was running, changes to it wouldn't affect detectors that had already been initialized, right?

If so - and I really hate to be the bearer of bad news on this - that means the new feature flag operates differently from the other defined ones in a way that actually matters not to _you_ (or any other OSS user) but to TruffleHog Enterprise, which also uses this code. TruffleHog Enterprise runs as a persistent daemon, so we need the ability to dynamically toggle feature flags of running processes and have the changes be reflected. Having some flags require scanner restarts and other flags not require them is an accident waiting to happen.

I know this kind of sends you back to the drawing board, but I really want to avoid some future person having a really irritating surprise.

(Or I'm misreading this and imagined a problem that doesn't exist!)

rgmz:
> _In theory_, if you could toggle `ssl-verify` while the scan was running, changes to it wouldn't affect detectors that had already been initialized, right?

That is correct for HTTP-based detectors. Non-HTTP detectors like MongoDB would work dynamically.

Detectors that use [`common.SaneClient()`](https://github.com/trufflesecurity/trufflehog/commit/4218c39d99b5f30153f62471c1be1c1596f0a4d4#diff-6cdf2ccf7c7d59da10b55c79b78ba31babff84f598ebd9e5720d57dc2d1c64a1R20) or [`detectors.DetectorHttpClientWithNoLocalAddresses`](https://github.com/trufflesecurity/trufflehog/pull/3221/files) create a singleton at start-up, either globally or per-detector. 'Fixing' this would require changing 800+ detectors to dynamically create HTTP clients, and that may have other implications. 

Since this pattern is [literally from the first commit of v3](https://github.com/trufflesecurity/trufflehog/commit/4218c39d99b5f30153f62471c1be1c1596f0a4d4), I think a change of that scale is probably outside of the scope of this. :/

PR #2906, opened 2024-06-03
[refactor] - multi part detectors

ahrav:
### Description:
Detectors are split up by commits.

### Checklist:


dustin-decker (approved):
review-robot says all good

>Any file with multiple different regexp patterns required for credential detection and verification should have 'detectors.DefaultMultiPartCredentialProvider' in the Scanner struct. If there is only one regexp pattern required and 'detectors.DefaultMultiPartCredentialProvider' is not in the Scanner struct, reply "No changes needed".

PR #2907, opened 2024-06-03
integration testing for mongodb.

abmussani:
### Description:
Implemented integration testing for MongoDB using testcontainers-go. Test cases includes:

- Verified connection string
- Unverified connection string (incorrect password)
- Verified connection string but time out is too short (1 Millisecond)
- Bad host (by changing the host)

### Checklist:


pkg/detectors/mongodb/mongodb_integration_test.go
```diff
@@ -0,0 +1,182 @@
+//go:build detectors && integration
```
ahrav, line 1: oooh TIL! I did not realize you could do this, very cool.

ahrav (approved):
Love it! Thanks for adding these.

PR #2908, opened 2024-06-03
Make the github action work with a path as input

benbridts:
### Description:
This change executes all shell commands relative to `path`, and makes sure that `git` is always working by running `git status` first. If the `path` is not a git repository, `git status` should give a clear error message about this.

When passing `path` as input before this change, trufflehog would:

- scan nothing (when not passing `head` or `base`)
- fail to run (when passing either or both)

I'm happy to close this PR and raise this as an issue of that is the preferred way to handle this.

### Checklist:

Note: `make lint` fails for me, but I have not touched any go related files. I have not investigated why this fails on my machine yet

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -30,8 +30,8 @@ runs:
   using: "composite"
   steps:
   - shell: bash
+    working-directory: ${{ inputs.path }}
     env:
-      REPO_PATH: ${{ inputs.path }}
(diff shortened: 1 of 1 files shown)


ahrav (approved):
This looks great! Thanks for the enhancement, we really appreciate it.

PR #2909, opened 2024-06-03
[chore] Always log git repositories being scanned

mcastorina:
### Description:
Always logs the git repositories that are being scanned for users to see.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -550,7 +550,7 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 
 	gitDir := filepath.Join(path, gitDirName)
 
-	logger.V(1).Info("scanning repo", logValues...)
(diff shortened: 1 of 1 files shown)


pkg/sources/git/git.go
```diff
@@ -550,7 +550,7 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 
 	gitDir := filepath.Join(path, gitDirName)
 
-	logger.V(1).Info("scanning repo", logValues...)
+	logger.Info("scanning repo", logValues...)
```
ahrav, line 553: question: Would it be worth distinguishing this log from the one below, or do we not really care?

ahrav (approved):

pkg/sources/git/git.go
mcastorina: I'm not really sure what the difference is so I'm going to leave it as is.

PR #2917, opened 2024-06-05
Improve GitHub wiki scan errs

rgmz:
### Description:
This handles a not-uncommon edge case where GitHub's API says a repo has a wiki, and the wiki page resolves, but the repository is empty ([example](https://github.com/watson-developer-cloud/text-to-speech-nodejs/wiki)). When scanning a large org with `--include-wikis` enabled this could produce a bunch of 'false' scan errors.

It also removes redundant wrapping of clone errors to make the errors more clear.

### Checklist:


dustin-decker (approved):
Thank you!

PR #2918, opened 2024-06-05
[feat] - Add Support for `compareDetectionStrategies` Mode

ahrav:
### Description:
This PR introduces a new feature, `compareDetectionStrategies` mode, which allows for the comparison of two span calculation strategies used in our detection process: the custom span calculator and the entire chunk span calculator.

When the `--compare-detection-strategies` flag is enabled, we run the scan multiple times using different engine configurations. We opted not to run the scans concurrently as this mode is intended for regression checking to identify any missed secrets, rather than for performance.

### Summary

These changes enable users to compare the effectiveness of different span calculation strategies.

Note: I'm aware there might be a more elegant solution, but this seemed like the best balance between time and effort.

Screenshot 2024-06-04 at 6 47 33 PM
Screenshot 2024-06-04 at 6 48 33 PM

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,11 +50,12 @@ var (
 	onlyVerified        = cli.Flag("only-verified", "Only output verified results.").Bool()
 	results             = cli.Flag("results", "Specifies which type(s) of results to output: verified, unknown, unverified. Defaults to all types.").Hidden().String()
 
-	allowVerificationOverlap = cli.Flag("allow-verification-overlap", "Allow verification of similar credentials across detectors").Bool()
-	filterUnverified         = cli.Flag("filter-unverified", "Only output first unverified result per chunk per detector if there are more than one results.").Bool()
-	filterEntropy            = cli.Flag("filter-entropy", "Filter unverified results with Shannon entropy. Start with 3.0.").Float64()
-	scanEntireChunk          = cli.Flag("scan-entire-chunk", "Scan the entire chunk for secrets.").Hidden().Default("true").Bool()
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -443,31 +444,138 @@ func run(state overseer.State) {
...
+	return nil
+}
+
+func logComparisonMetrics(customMetrics, entireMetrics engine.Metrics) {
+	fmt.Printf("Comparison of scan results: \n")
+	fmt.Printf("Custom span - Chunks: %d, Bytes: %d, Verified Secrets: %d, Unverified Secrets: %d, Duration: %s\n",
+		customMetrics.ChunksScanned, customMetrics.BytesScanned, customMetrics.VerifiedSecretsFound, customMetrics.UnverifiedSecretsFound, customMetrics.ScanDuration.String())
+	fmt.Printf("Entire chunk - Chunks: %d, Bytes: %d, Verified Secrets: %d, Unverified Secrets: %d, Duration: %s",
+		entireMetrics.ChunksScanned, entireMetrics.BytesScanned, entireMetrics.VerifiedSecretsFound, entireMetrics.UnverifiedSecretsFound, entireMetrics.ScanDuration.String())
+}
```
dustin-decker, line 539: I think it would be good to emit an error if the secret counts don't match so we can surface and exit code. This would make a good future smoke test case.

dustin-decker (approved):
Looks great! I think we should surface an error and exit code for mismatched secret counts for use in a smoke test.

main.go
ahrav: Done.

PR #2920, opened 2024-06-05
[chore] - address comments

ahrav:
### Description:
Address comment from https://github.com/trufflesecurity/trufflehog/pull/2889
### Checklist:

### DIFF
--- a/pkg/channelmetrics/observablechan.go
+++ b/pkg/channelmetrics/observablechan.go
@@ -83,12 +83,7 @@ func (oc *ObservableChan[T]) RecvCtx(ctx context.Context) (T, error) {
 		oc.RecordChannelLen()
 	}(time.Now())
 
-	item, err := common.CancellableRecv(ctx, oc.ch)
-	if err != nil {
-		ctx.Logger().Error(err, "failed to read item from observable channel")
-	}
-
-	return item, nil
+	return common.CancellableRecv(ctx, oc.ch)
 }
 
 // RecordChannelCapacity records the capacity of the channel buffer.
--- a/pkg/channelmetrics/observablechan.go
+++ b/pkg/channelmetrics/observablechan.go
@@ -52,6 +52,7 @@ func (oc *ObservableChan[T]) Close() {
 
 // Send sends an item into the channel and records the duration taken to do so.
 // It also updates the current size of the channel buffer.
+// This method blocks until the item is sent.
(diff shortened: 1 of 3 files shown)


pkg/channelmetrics/observablechan.go
```diff
@@ -64,21 +65,22 @@ func (oc *ObservableChan[T]) Send(ctx context.Context, item T) {
 
 // Recv receives an item from the channel and records the duration taken to do so.
 // It also updates the current size of the channel buffer.
-func (oc *ObservableChan[T]) Recv(_ context.Context) T {
-	startTime := time.Now()
-	defer func() {
-		oc.metrics.RecordConsumeDuration(time.Since(startTime))
+func (oc *ObservableChan[T]) Recv(ctx context.Context) T {
```
mcastorina, line 68: **Suggestion (optional):**

I'm thinking two methods will cover potential use cases:

```go
// also returns an error in case of cancellation
func (oc *ObeservableChan[T]) RecvCtx(ctx context.Context) (T, error) {
    // ...
}

func (oc *ObservableChan[T]) Recv() T {
    // blocks forever
    v, _ := oc.RecvCtx(context.Background())
    return v
}
```

mcastorina: Same goes for `Send`.

mcastorina (approved):

pkg/channelmetrics/observablechan.go
ahrav: Just to confirm, are you proposing adding two additional methods so it can be used in both blocking and non-blocking fashions?

mcastorina: Yeah, well both blocking. The `RecvCtx` and `SendCtx` would also return an `error` so the caller knows whether it returned because the context was cancelled.

pkg/channelmetrics/observablechan.go
```diff
@@ -76,7 +88,7 @@ func (oc *ObservableChan[T]) Recv(ctx context.Context) T {
 		ctx.Logger().Error(err, "failed to read item from observable channel")
 	}
 
-	return item
+	return item, nil
```
mcastorina, line 91: This could be

```go
return common.CancellableRecv(ctx, oc.ch)
```

ahrav: sigh... I think my brain stopped working last night. Thanks for being patient 🙇

PR #2922, opened 2024-06-05
Compare detection strategies per chunk/detector

rgmz:
### Description:
This is an alternative implementation to #2918, with two main advantages:

1. It only requires the data to be downloaded and parsed/chunked once
2. It provides specific context about the affected detector(s) and chunk(s), rather than vague high-level metrics which are hard to action.

**Note:** Anything I deleted wasn't necessarily because I didn't think it was useful, it was just the easiest way to implement this.

### Example

**Compare per chunk/detector**

When testing #2894 against a known true positive, it provides immediate and unambiguous feedback that there's a discrepancy. 
```sh
$ ./trufflehog github --repo="..." --token="..." --compare-detection-strategies  --allow-verification-overlap
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2024-06-05T19:11:32-04:00       info-0  trufflehog      running source  {"source_manager_worker_id": "XlWGy", "with_units": false, "target_count": 0, "source_manager_units_configurable": true}
...
2024-06-05T19:11:34-04:00       error   trufflehog      Scan results do not match       {"detector_worker_id": "vMNq4", "chunk": {"Github":{"link":"...","repository":"...","commit":"...","email":"...","file":"...","timestamp":"2019-05-06 16:03:25 +0000","line":709}}, "detector": "KubeConfig", "error": "mismatch between custom span and entire chunk: 0 vs 1"}
```

**Compare per scan**

Information is only logged at the end, and doesn't provide any insight to _which_ chunk(s) or detector(s) were affected.
```
$ ./trufflehog github --repo="..." --token="..." --compare-detection-strategies  --allow-verification-overlap
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2024-06-05T19:22:41-04:00       info-0  trufflehog      running source  {"source_manager_worker_id": "2xLZv", "with_units": false, "target_count": 0, "source_manager_units_configurable": true}
2024-06-05T19:22:41-04:00       info-0  trufflehog      running source  {"source_manager_worker_id": "lzBGQ", "with_units": false, "target_count": 0, "source_manager_units_configurable": true}
2024-06-05T19:22:41-04:00       info-0  trufflehog      Completed enumeration   {"num_repos": 1, "num_orgs": 0, "num_members": 0}
2024-06-05T19:22:41-04:00       info-0  trufflehog      Completed enumeration   {"num_repos": 1, "num_orgs": 0, "num_members": 0}
[...]

PR #2929, opened 2024-06-06
Add explicit warning when an overlap is detected

rgmz:
### Description:
Similar to #2922, the goal of this change is to provide clear and actionable information if a detection is impacted.

The problem is currently that "verification overlap" doesn't provide useful feedback. Even if you use `--results=unknown` to see the error, it doesn't tell you which detectors overlapped.

```
2024-06-06T08:27:05-04:00       info-0  trufflehog      WARNING: A result will not be verified because more than one detector matches. You can override this behavior by using the --allow-verification-overlap flag    {"verification_overlap_worker_id": "2hlPu", "detectors": ["AzureSearchAdminKey", "AzureDevopsPersonalAccessToken"]}
```

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -884,12 +884,22 @@ type chunkSecretKey struct {
 	detectorKey ahocorasick.DetectorKey
 }
 
-func likelyDuplicate(ctx context.Context, val chunkSecretKey, dupes map[chunkSecretKey]struct{}) bool {
-	const similarityThreshold = 0.9
+const similarityThreshold = 0.9
 
-	valStr := val.secret
+func likelyDuplicate(ctx context.Context, val chunkSecretKey, dupes map[chunkSecretKey]struct{}) (bool, detectorspb.DetectorType) {
+	var (
+		detectorType detectorspb.DetectorType
+		valStr       = val.secret
+		valRedacted  string
+	)
+	if len(valStr) < 3 {
+		return false, detectorType
+	} else {
+		valRedacted = valStr[:3] + "..."
+	}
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -752,7 +779,20 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 						continue
 					}
 
-					if likelyDuplicate(ctx, key, chunkSecrets) {
+					if ok, t := likelyDuplicate(ctx, key, chunkSecrets); ok {
```
rgmz, line 782: Personally, I don't understand the threat model for this feature. It silently skips verification if a result matches more than one detector, which happens frequently during normal operation.

The notion of a "malicious" detector is confusing — is Truffle adding malicious detectors to TruffleHog? Either this check needs a whitelist of common overlaps (e.g., Azure-related, URL and FTP), or it should only trigger between built-in and custom detectors (IMO).

pkg/engine/engine_test.go
```diff
@@ -633,7 +633,7 @@ func TestLikelyDuplicate(t *testing.T) {
 			name:     "empty strings",
 			val:      chunkSecretKey{"", detectorA.Key},
 			dupes:    map[chunkSecretKey]struct{}{{"", detectorB.Key}: {}},
-			expected: true,
+			expected: false,
```
rgmz, line 636: Should empty strings be considered duplicates? I guess it makes sense, but I'd initially tweaked the logic for `likelyDuplicate` to ignore anything smaller than 3 characters.

PR #2952, opened 2024-06-10
Update twitter regex

abmussani:
### Description:
Twitter has changed the structure of its bearer token. Now, token has minimum 2 parts to maximum 5 parts, separated by % sign. Also, they have started to supported multiple type of authentication, on which I am looking into.

### Checklist:


dustin-decker:
Could theres still be the previous bearer token format in the wild or do they expire?

abasit-folio3:
> Could theres still be the previous bearer token format in the wild or do they expire?

App level bearer token does not have expiry. They can be revoke and regenerated either via Portal or API.

abmussani:
App level bearer token does not have expiry. They can be revoke and regenerated either via Portal or API.

dustin-decker:
> App level bearer token does not have expiry. They can be revoke and regenerated either via Portal or API.

Okay, in that case, I think we have to version this detector. Please see the `github` detector package for an example.

abmussani:
> Okay, in that case, I think we have to version this detector. Please see the `github` detector package for an example.

@dustin-decker The change in token format is not that drastic to be considered as a new version. That's what I believe.  

As per the old regex, The bearer token would be consist of 6 smaller parts (fixed) separated by '%' character. 
Here is the old regex :

`\b([A-Z]{22}%[a-zA-Z-0-9]{23}%[a-zA-Z-0-9]{6}%[a-zA-Z-0-9]{3}%[a-zA-Z-0-9]{9}%[a-zA-Z-0-9]{52})\b`

The newer format is almost similar but the number of parts has changed. It varies from 2 to 4 sub parts, separated by '%' character.  Documentation has refused to specify the format. I verified it by regenerating the tokens via portal (10+ times).

A single regex can accommodate the old and newer format of bearer token in current detector. Thoughts ?

abmussani:
Closing in favor #3016

PR #2953, opened 2024-06-10
[fix] - implement MaxSecretSizeProvider for `auth0managementapitoken` detector

ahrav:
### Description:
This PR introduces the MaxSecretSizeProvider for the `auth0managementapitoken` detector. Additionally, it updates the keyword list to include `ey`, the prefix of the API token.

### Checklist:

### DIFF
--- a/pkg/detectors/auth0managementapitoken/auth0managementapitoken.go
+++ b/pkg/detectors/auth0managementapitoken/auth0managementapitoken.go
@@ -3,9 +3,10 @@ package auth0managementapitoken
 import (
 	"context"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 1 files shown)


pkg/detectors/auth0managementapitoken/auth0managementapitoken.go
```diff
@@ -18,35 +19,38 @@ type Scanner struct {
...
+
+const maxSecretSize = 5000
+
+func (Scanner) MaxSecretSize() int64 { return maxSecretSize }
 
 // FromData will find and optionally verify Auth0ManagementApiToken secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
 
-	managementApiTokenMatches := managementApiTokenPat.FindAllStringSubmatch(dataStr, -1)
```
ahrav, line 41: Renamed to more closely align with the go style guide.

ahrav:
Removed `ey` as a keyword to avoid a gazillion matches. Thanks @rosecodym

dustin-decker (approved):

pkg/detectors/auth0managementapitoken/auth0managementapitoken.go
```diff
@@ -18,35 +19,38 @@ type Scanner struct {
...
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
-func (s Scanner) Keywords() []string {
-	return []string{"auth0"}
-}
+func (s Scanner) Keywords() []string { return []string{"auth0"} }
+
+const maxSecretSize = 5000
+
+func (Scanner) MaxSecretSize() int64 { return maxSecretSize }
```
bugbaba, line 39: `MaxSecretSize()` is not used I believe it was intended to be used in `if len(managementAPITokenRes) < 2000 || len(managementAPITokenRes) > 5000 {`

ahrav: We should use it [here](https://github.com/trufflesecurity/trufflehog/blob/a150103e2bc7917baa00bc27af2c15721b649ff0/pkg/engine/ahocorasick/ahocorasickcore.go#L82) to determine the endOffset for the chunk. This ensures we extract only the section of the chunk containing the credential before passing the data to the detector. Consequently, the regex runs on a smaller dataset.

PR #2955, opened 2024-06-10
[chore] - Update`databrickstoken` detector keyword

ahrav:
### Description:
This PR updates the keyword for the `databrickstoken` detector to `cloud.databricks.com`. This more specific keyword should reduce the number of regex matches the detector needs to process.

### Checklist:

### DIFF
--- a/pkg/detectors/databrickstoken/databrickstoken.go
+++ b/pkg/detectors/databrickstoken/databrickstoken.go
@@ -3,9 +3,10 @@ package databrickstoken
 import (
 	"context"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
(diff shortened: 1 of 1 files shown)


pkg/detectors/databrickstoken/databrickstoken.go
```diff
@@ -30,9 +32,7 @@ var (
 
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
-func (s Scanner) Keywords() []string {
-	return []string{"databricks", "dapi"}
-}
+func (s Scanner) Keywords() []string { return []string{"cloud.databricks.com", "dapi"} }
```
rgmz, line 35: Databricks has a number of related domains, including: `cloud.databricks.com`, `.azuredatabricks.net`, `gcp.databricks.com`, `dev.databricks.com`, and possibly more.

ahrav: Whoops I completely read the regex incorrectly. Thanks.

PR #2961, opened 2024-06-11
adding eraser ai detector

0x1:
AI Copilot for creating technical design: https://www.eraser.io/ai

### Description:
Add EraserAI detector

### Checklist:


pkg/detectors/eraser/eraser.go
```diff
@@ -0,0 +1,108 @@
...
+			isVerified, extraData, verificationErr := verifyMatch(ctx, client, match)
+			s1.Verified = isVerified
+			s1.ExtraData = extraData
+			s1.SetVerificationError(verificationErr, match)
+		}
+
+		results = append(results, s1)
+	}
+
+	return
```
abmussani, line 66: I guess you have missed to return the results with error ( probably nil).

abmussani (changes requested):
Thank you @0x1 for adding this. With little change, You have also not added new detector in [pkg/engine/defaults.go](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/defaults.go).

pkg/detectors/eraser/eraser.go
rosecodym: This is actually a [Go language feature](https://go.dev/tour/basics/7)! It's kind of confusing sometimes, but it has its uses, so we allow it.

ahrav (approved):

pkg/detectors/eraser/eraser.go
abmussani: ahan, I did not know about 'naked returns'. Thank you @rosecodym for increasing my Golang knowledge.

PR #2964, opened 2024-06-13
Fix test compilation errors

rgmz:
### Description:
PRs #2812 and #2930 inadvertently broke a few tests. 

This wasn't noticed because Go doesn't seem to have a way to compile all tests/tags. The closest you can get is manually specifying all the tags and running a pattern that won't be matched.🤷‍♂️ 
```sh
$ go test -run=XXX_SHOULD_NEVER_MATCH_XXX -tags="detectors,integration,localInfra" ./...
```

### Checklist:


ahrav (approved):
Thank you ❤️

PR #2966, opened 2024-06-13
update Asana PAT detector to match new token format

slashnick:
### Description:
Asana updated their Personal Access Token format to include an extra ID before the secret part.

Newly issued Asana PATs look something like `2/1234567890123456/1234567890123457:abcdef0123456789abcdef0123456789`

### Checklist:


zricethezav (commented):
@slashnick do you know if the old format for the api keys are still being accepted by asana? If so this should probably be a versioned detector

kashifkhan0771:
The old format is still working. @slashnick Can you add it as a new version?

kashifkhan0771:
Closing against: https://github.com/trufflesecurity/trufflehog/pull/4273

PR #2969, opened 2024-06-13
Return match/reason from `detectors.IsKnownFalsePositive`

rgmz:
### Description:
Recently, there have been several confusing bugs related to false-positive filtering. In each case, determining the cause of this has required manually adding print statements in order to see why.

#2643 added a log statement when a result is filtered, however, it doesn't include the reason why. 

This updates the method to include the specific match/reason behind it returning `true`.

### Checklist:

### DIFF
--- a/pkg/custom_detectors/custom_detectors.go
+++ b/pkg/custom_detectors/custom_detectors.go
@@ -8,11 +8,12 @@ import (
 	"regexp"
 	"strings"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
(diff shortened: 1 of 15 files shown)


rosecodym (approved):
thanks for this! it looks good to me but i want to give @ahrav an opportunity to hold this merge off until he's done fiddling with all the engine code he was planning to fiddle with

ahrav:
> thanks for this! it looks good to me but i want to give @ahrav an opportunity to hold this merge off until he's done fiddling with all the engine code he was planning to fiddle with

I think we can go ahead and merge this. I'll adjust my PR to deal with the merge conflicts, it shouldn't be too bad. Thanks @rgmz

ahrav (approved):

PR #2976, opened 2024-06-16
Add UUIDs to false positive checker

### LINKED ISSUE #1953: `IsKnownFalsePositive` incorrectly marks value as false-positive

### TruffleHog Version
3.60.1

### Trace Output
N/A

### Expected Behavior

The following (valid) secret is not marked as a false-positive
```
9ed918b9-7ea6-4880-9734-e3822649e20b
```

### Actual Behavior

The value is marked as a false-positive in [`IsKnownFalsePositive`](https://github.com/trufflesecurity/trufflehog/blob/855aba2407ba2f195eb47c62a63a512b39e46be3/pkg/detectors/falsepositives.go#L38) because it contains `0-9` from [`badlist.txt`](https://github.com/trufflesecurity/trufflehog/blob/855aba2407ba2f195eb47c62a63a512b39e46be3/pkg/detectors/badlist.txt#L91).

## Environment
N/A

## Additional Context

For things like [UUIDs](https://github.com/trufflesecurity/trufflehog/blob/855aba2407ba2f195eb47c62a63a512b39e46be3/pkg/detectors/azure/azure.go#L78) it doesn't make sense to use the standard IsKnownFalsePositivr function because the pattern is so specific. Perhaps it's worth adding a UUID-specific one to catch stuff like `00000-0000-...`?

### References
N/A

rgmz:
### Description:
This fixes #1953 by adding a list of UUID false positives. It also renames the lists so they appear together in the file explorer (nitpick). 

> For things like [UUIDs](https://github.com/trufflesecurity/trufflehog/blob/855aba2407ba2f195eb47c62a63a512b39e46be3/pkg/detectors/azure/azure.go#L78) it doesn't make sense to use the standard IsKnownFalsePositivr function because the pattern is so specific. Perhaps it's worth adding a UUID-specific one to catch stuff like 00000-0000-...?

It also changes a few things:
- Remove the `detectors` label so false positive unit tests can catch issues more frequently
- Changes the `DefaultFalsePositives` from a slice to a map to allow for exact matching

### Checklist:


pkg/detectors/falsepositives.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
-var DefaultFalsePositives = []FalsePositive{"example", "xxxxxx", "aaaaaa", "abcde", "00000", "sample", "*****"}
+var (
+	DefaultFalsePositives = []FalsePositive{"example", "xxxxxx", "aaaaaa", "abcde", "00000", "sample", "*****"}
+	UuidFalsePositives    []FalsePositive
```
rgmz, line 17: Having a list for UUIDs is useful if you have a value that's supposed to be a UUID.

pkg/detectors/falsepositives_test.go
```diff
@@ -1,6 +1,3 @@
-//go:build detectors
```
rgmz, line 1: As far as I can tell, nothing in this test relies on 'real' values. It would be beneficial for this test to be included in `make test` and `make test-community`.

pkg/detectors/falsepositives.go
```diff
@@ -12,7 +12,12 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
 
-var DefaultFalsePositives = []FalsePositive{"example", "xxxxxx", "aaaaaa", "abcde", "00000", "sample", "*****"}
+var (
+	DefaultFalsePositives = map[FalsePositive]struct{}{
+		"example": {}, "xxxxxx": {}, "aaaaaa": {}, "abcde": {}, "00000": {}, "sample": {}, "*****": {},
+	}
+	UuidFalsePositives map[FalsePositive]struct{}
```
rosecodym, line 19: Is this actually used anywhere?

pkg/detectors/falsepositives.go
```diff
@@ -64,16 +81,19 @@ func GetFalsePositiveCheck(detector Detector) func(Result) (bool, string) {
...
+func IsKnownFalsePositive(match string, falsePositives map[FalsePositive]struct{}, wordCheck bool) (bool, string) {
 	if !utf8.ValidString(match) {
 		return true, "invalid utf8"
 	}
 	lower := strings.ToLower(match)
-	for _, fp := range falsePositives {
+	for fp := range falsePositives {
 		fps := string(fp)
-		if strings.Contains(lower, fps) {
+		if lower == fps {
```
rosecodym, line 91: You still do exact matching by iterating over `falsePositives` - do you actually use your new map as a map anywhere?

pkg/detectors/falsepositives_test.go
rosecodym: @zricethezav are you aware of any non-obvious reason that this test is behind `detectors`? (Also tagging @abmussani for visibility)

pkg/detectors/falsepositives.go
rgmz: #2978 and eventually #1979.

pkg/detectors/falsepositives_test.go
zricethezav: > @zricethezav are you aware of any non-obvious reason that this test is behind detectors? (Also tagging @abmussani for visibility)

negative, seems like this test should be part of the community PR tests

abmussani (approved):
LGTM. just a little comment about utilizing the Map.

pkg/detectors/falsepositives.go
rgmz: That makes sense to me.

ahrav (approved):
Nice! Wow, sorry this has been open so long.

PR #2978, opened 2024-06-16
Create detector for Azure refresh tokens

rgmz:
### Description:

(This should only be merged after #2985.)

Azure refresh tokens are long-lived opaque tokens returned alongside access tokens[^1]. They can remain valid for an indefinite period, and can be used to generate a new valid access token on behalf of the subject.

This detector can be tested by obtaining a live refresh token. I like using [ROADtools](https://github.com/dirkjanm/ROADtools):
```sh
# Obtain a token for Microsoft Office
# https://learn.microsoft.com/en-us/troubleshoot/azure/entra/entra-id/governance/verify-first-party-apps-sign-in
$ roadrecon auth --device-code -c d3590ed6-52b3-4102-aeff-aad2292ab01c -r msgraph
To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code ABCDEFG123 to authenticate. 
Tokens were written to .roadtools_auth
```

This detector only works with refresh tokens created for [public clients](https://learn.microsoft.com/en-us/entra/identity-platform/msal-client-applications). If a client is confidential, the required tenant, id, and secret should trigger the service principal detector.

[^1]: https://learn.microsoft.com/en-us/azure/energy-data-services/how-to-generate-auth-token#get-an-auth-token-and-a-refresh-token

### Checklist:
* [x] Dependent PR merged? #2976 


pkg/detectors/azure_entra/refreshtoken/refreshtoken.go
```diff
@@ -0,0 +1,312 @@
...
+	return detectorspb.DetectorType_AzureRefreshToken
+}
+
+func (s Scanner) Description() string {
+	return "Azure Entra ID refresh tokens provide long-lasting access to an account."
+}
+
+func (Scanner) MaxSecretSize() int64 { return 2048 }
+
+func (Scanner) StartOffset() int64 { return 4096 }
```
rgmz, line 58: @ahrav In this case the token can be quite long (1000+) and the other fields can be either before or after it. What's the correct way to look +/- ~1024 on either side?

ahrav (approved):

PR #2982, opened 2024-06-17
Fix giturl handling of paths containing `%`

rgmz:
### Description:
This fixes an issue where paths that contain `%` cause `UpdateLinkLineNumber` to fail.

```
parse \"https://github.com/coinbase/cbpay-js/blob/abcdefg/folder/%/name\": invalid URL escape \"%/n\"
```

### Checklist:

### DIFF
--- a/pkg/giturl/giturl.go
+++ b/pkg/giturl/giturl.go
@@ -174,6 +174,7 @@ var linePattern = regexp.MustCompile(`L\d+`)
 // UpdateLinkLineNumber updates the line number in a repository link.
 // Used post-link generation to refine reported issue locations within large scanned blocks.
 func UpdateLinkLineNumber(ctx context.Context, link string, newLine int64) string {
+	link = strings.Replace(link, "%", "%25", -1)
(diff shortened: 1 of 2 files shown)


dustin-decker:
Should we consider url encoding path elements? It seems like this could have just as easily been a different special character. I'm surprised it wasn't already encoded.

rgmz:
That's probably a better solution. This is an old change I had sitting around, I didn't look to deeply into other solutions.

zricethezav:
> Should we consider url encoding path elements? 

@rgmz did you want to follow up on this? Seems like a worthwhile change.

ahrav (approved):

ahrav:
Ah, looks like my approval is worthless here. 😞 🤣

rgmz:
> > Should we consider url encoding path elements?
> 
> @rgmz did you want to follow up on this? Seems like a worthwhile change.

Could be a useful follow-up, but it'd need a lot more testing.

This is a relatively straightforward and safe change.

PR #2995, opened 2024-06-20
Return targeted scan errors

rosecodym:
### Description:
Targeted scans should return their errors so that consumers can process them. By creating a type that combines an error with a targeted secret ID, we can return these errors without having to modify the `Source` interface.

### Checklist:

### DIFF
--- a/pkg/sources/errors.go
+++ b/pkg/sources/errors.go
@@ -70,3 +70,7 @@ var _ error = (*TargetedScanError)(nil)
 func (t TargetedScanError) Error() string {
 	return t.Err.Error()
 }
+
+func (t TargetedScanError) Unwrap() error {
+	return t.Err
+}
--- a/pkg/sources/errors.go
+++ b/pkg/sources/errors.go
@@ -58,16 +58,15 @@ func (s *ScanErrors) Errors() error {
 	return errors.Join(s.errors...)
 }
 
-// TargetedScanErrorGroup maps secret IDs to errors. It can be used by targeted scans to return error information for
-// multiple targets together.
-type TargetedScanErrorGroup map[int64]error
+// TargetedScanError is an error with a secret ID attached. Collections of them can be returned by targeted scans that
+// scan multiple targets in order to associate individual errors with individual scan targets.
+type TargetedScanError struct {
+	Err      error
+	SecretID int64
+}
 
-var _ error = (*TargetedScanErrorGroup)(nil)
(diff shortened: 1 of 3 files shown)


pkg/sources/github/github.go
```diff
@@ -1514,14 +1519,16 @@ func (s *Source) chunkPullRequestComments(ctx context.Context, repoInfo repoInfo
 	return nil
 }
 
-func (s *Source) scanTargets(ctx context.Context, targets []sources.ChunkingTarget, chunksChan chan *sources.Chunk) error {
+func (s *Source) scanTargets(ctx context.Context, targets []sources.ChunkingTarget, chunksChan chan *sources.Chunk) []*sources.TargetedScanError {
+	var errs []*sources.TargetedScanError
```
mcastorina, line 1523: **Suggestion:** You could replace `[]*sources.TargetedScanError` with `[]error` on both of these lines and avoid the conversion in the parent function.

pkg/sources/errors.go
```diff
@@ -57,3 +57,16 @@ func (s *ScanErrors) Errors() error {
...
+type TargetedScanError struct {
+	Err      error
+	SecretID int64
+}
+
+var _ error = (*TargetedScanError)(nil)
+
+func (t TargetedScanError) Error() string {
+	return t.Err.Error()
+}
```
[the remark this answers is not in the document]
rosecodym: The reason I didn't was that these error messages might end up in front of users, who wouldn't be able to use that information, but I could be talked out of it if you feel strongly the other way.

pkg/sources/errors.go
```diff
@@ -57,3 +57,16 @@ func (s *ScanErrors) Errors() error {
 	defer s.mu.RUnlock()
 	return errors.Join(s.errors...)
 }
+
+// TargetedScanError is an error with a secret ID attached. Collections of them can be returned by targeted scans that
+// scan multiple targets in order to associate individual errors with individual scan targets.
+type TargetedScanError struct {
+	Err      error
+	SecretID int64
+}
```
mcastorina, line 66: It might be good to provide an `Unwrap()` method for this type too:

```go
func (t TargetedScanError) Unwrap() error {
    return t.Err
}
```

mcastorina (approved):

rosecodym:
Targeted scans should return their errors so that consumers can process them. By creating a type that combines an error with a targeted secret ID, we can return these errors without having to modify the Source interface.

PR #3000, opened 2024-06-21
New Source: HuggingFace

joeleonjr:
### Description:
This PR adds HuggingFace as a new source. Users will have the ability to scan a HF model, dataset or space. A HF token is required for all scans except basic `git` scans of a public model, dataset or space. That means all org/user enumeration, discussions/PR enumeration,  private scanning, etc. requires a token. Tokens are free and rate limiting doesn't seem to come into play when using the API.

I added a couple test files: one to address the client functionality (since there is no golang HF package) and one to address the scanning logic. Coverage is pretty high on both, but not perfect.

### Checklist:


pkg/sources/huggingface/client.go
```diff
@@ -0,0 +1,214 @@
...
+		HTTPClient: &http.Client{
+			Timeout: timeout,
+		},
+	}
+}
+
+// get makes a GET request to the Hugging Face API
+// Note: not addressing rate limit, since it seems very permissive. (ex: "If \
+// your account suddenly sends 10k requests then you’re likely to receive 503")
+func (c *HFClient) get(ctx context.Context, url string, target interface{}) error {
```
rgmz, line 136: `any` is preferred to `interface{}` now. It could also be a generic `T`.

pkg/sources/huggingface/client.go
```diff
@@ -0,0 +1,214 @@
...
+
+	req.Header.Set("Authorization", "Bearer "+c.APIKey)
+
+	resp, err := c.HTTPClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to make request to HuggingFace API: %w", err)
+	}
+	defer resp.Body.Close()
+
+	return json.NewDecoder(resp.Body).Decode(target)
```
rgmz, line 153: HuggingFace's API doesn't have the best documentation, but I believe there's a rate limit indicated by a 429 response.

https://discuss.huggingface.co/t/hugging-face-api-rate-limits/16746/6

pkg/sources/huggingface/client.go
zricethezav: >  any is preferred to interface{} now. It could also be a generic T.

Embrace tradition. `interface{}` for life

joeleonjr: soooo... leave it? change it? lol

zricethezav: I'm saying leave it but if one of the other engineers has a strong opinion the other way then we can reconsider.

pkg/sources/huggingface/huggingface.go
```diff
@@ -0,0 +1,757 @@
...
+
+	filteredModelsCache   *filteredRepoCache
+	filteredSpacesCache   *filteredRepoCache
+	filteredDatasetsCache *filteredRepoCache
+
+	repoInfoCache repoInfoCache
+
+	git *git.Git
+
+	// scanOptMu   sync.Mutex // protects the scanOptions
```
zricethezav, line 72: same here, delete if not needed

rgmz:
Some public resources require sharing your information before you can view them. Idk if this is something that can be handled programmatically.
image

```
2024-06-26T19:08:35-04:00       error   trufflehog      failed to fetch model   {"organization": "ibm", "model": "https://huggingface.co/ibm/testing-patchtst_etth1_forecast.git", "error": "access is restricted."}
2024-06-26T19:08:40-04:00       error   trufflehog      failed to fetch model   {"model": "https://huggingface.co/ibm/testing-patchtst_etth1_forecast.git", "error": "access is restricted."}
```

joeleonjr:
> Some public resources require sharing your information before you can view them. Idk if this is something that can be handled programmatically, or at least logged/documented. image
> 
> ```
> 2024-06-26T19:08:35-04:00       error   trufflehog      failed to fetch model   {"organization": "ibm", "model": "https://huggingface.co/ibm/testing-patchtst_etth1_forecast.git", "error": "access is restricted."}
> 2024-06-26T19:08:40-04:00       error   trufflehog      failed to fetch model   {"model": "https://huggingface.co/ibm/testing-patchtst_etth1_forecast.git", "error": "access is restricted."}
> ```
> 
> Addendum: Git LFS support (in the future) would be a tremendous benefit for Huggingface. The nature of the platform means there's lots of large files hosted external to Git.
> 
> e.g., https://huggingface.co/dmis-lab/biobert-v1.1/blob/main/flax_model.msgpack

I attempted to handle that with the “access is restricted” message. What did you envision?

Also, completely agree on LFS. But that’s a much bigger endeavor.

rgmz:
> I attempted to handle that with the “access is restricted” message. What did you envision?

To me, "access is restricted" implies that you _can't_ access those models. You _can_, you just need to click an "I agree" button. A call to action would make this clearer; the language in HuggingFace's prompt is perfect, actually.

Does the API have a specific error code or message for "you must share your contact information"?

joeleonjr:
> https://huggingface.co/ibm/testing-patchtst_etth1_forecast

I hear what you're saying

> > I attempted to handle that with the “access is restricted” message. What did you envision?
> 
> To me, "access is restricted" implies that you _can't_ access those models. You _can_, you just need to click an "I agree" button. A call to action would make this clearer; the language in HuggingFace's prompt is perfect, actually.
> 
> Does the API have a specific error code or message for "you must share your contact information"?

I hear what you're saying. The challenge is there are situations where you have to wait for an org to approve your request. The error message is the same for both types: `{"error":"Access to model ibm/testing-patchtst_etth1_forecast is restricted. You must be authenticated to access it."}` So afaik there's no easy/reliable way to differentiate between models that require a simple click vs. waiting for admin approval. Also, not sure if we'd want to click the agree button on behalf of user's accounts.

rgmz:
> Does the API have a specific error code or message for "you must share your contact information"?

It turns out that the API contains a property for gated models (`"gated": "auto" | true | false,`), however, you can't see that until you have access. 😩

> The error message is the same for both types: {"error":"Access to model ibm/testing-patchtst_etth1_forecast is restricted. You must be authenticated to access it."}

That's only for unauthenticated requests. There seem to be three different types of errors.

1. It's private and you don't have access
```
HTTP/2 404
{"error":"Repository not found"}
```

2. It's gated and your request isn't authenticated or auth is invalid
```sh
# For some reason, excluding the 
$ curl -i "https://huggingface.co/api/models/meta-llama/Meta-Llama-3-8B" -H "Authorization: Bearer hf_fake"
HTTP/2 401

{"error":"Access to model meta-llama/Meta-Llama-3-8B is restricted. You must be authenticated to access it."}
```

3. It's gated and your request is authenticated.
```sh
$ curl -i "https://huggingface.co/api/models/meta-llama/Meta-Llama-3-8B" -H "Authorization: Bearer $TOKEN"
HTTP/2 403

{"error":"Access to model meta-llama/Meta-Llama-3-8B is restricted and you are not in the authorized list. Visit https://huggingface.co/meta-llama/Meta-Llama-3-8B to ask for access."}
```

joeleonjr:
> ```shell
> and you are not in the authorized list
> ```

Just pushed a change so 403 error msgs will provide more clear information. 401s will result in an "API Key is Invalid" type error message. Lmk if you think that language is sufficient.

pkg/sources/huggingface/huggingface.go
```diff
@@ -0,0 +1,745 @@
...
+
+	return repoInfo.visibility
+}
+
+// Chunks emits chunks of bytes over a channel.
+func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, targets ...sources.ChunkingTarget) error {
+	// If targets are provided, we're only scanning the data in those targets.
+	// Otherwise, we're scanning all data.
+	// This allows us to only scan the commit where a vulnerability was found.
+	// if len(targets) > 0 {
```
zricethezav, line 344: @joeleonjr remove commented code

zricethezav:
@joeleonjr PR looking good, one thing that did stand out is if you give the `./trufflehog huggingface` command with no arguments the program will continue and give users a false sense of scanning something. We need to check to make sure if at least one of `model, space, dataset, org, user` is set

joeleonjr:
> @joeleonjr PR looking good, one thing that did stand out is if you give the `./trufflehog huggingface` command with no arguments the program will continue and give users a false sense of scanning something. We need to check to make sure if at least one of `model, space, dataset, org, user` is set

Done. I followed the same logic used for GitHub.

zricethezav (approved):
Fantastic work! @joeleonjr

PR #3001, opened 2024-06-21
Clarify "no decoder found for chunk" log message

rgmz:
### Description:
When looking at trace logs you will see countless lines with the message: "no decoder found for chunk".

```
2024-05-01T15:10:28Z    info-4  trufflehog      finished scanning chunks        {"secret_worker_id": "MMpLT"}
2024-05-01T15:10:28Z    info-4  trufflehog      no decoder found for chunk      {"secret_worker_id": "w1TPA", "chunk": ...}
2024-05-01T15:10:28Z    info-4  trufflehog      no decoder found for chunk      {"secret_worker_id": "w1TPA", "chunk": {...}
2024-05-01T15:10:28Z    info-4  trufflehog      no decoder found for chunk      {"secret_worker_id": "w1TPA", "chunk": ...}
```

This message is unhelpful and somewhat confusing. It makes it seem like there's an issue, when in reality it's that a specific detector didn't find any content in a given chunk.

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -771,7 +771,7 @@ func (e *Engine) scannerWorker(ctx context.Context) {
 			decodeLatency.WithLabelValues(decoder.Type().String(), chunk.SourceName).Observe(float64(decodeTime))
 
 			if decoded == nil {
-				ctx.Logger().V(4).Info("no decoder found for chunk", "chunk", chunk)
+				ctx.Logger().V(4).Info("decoder not applicable for chunk", "decoder", decoder.Type().String(), "chunk", chunk)
 				continue
 			}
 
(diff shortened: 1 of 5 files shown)


zricethezav:
Seems like a worthwhile change even though we are adding another method to an interface. `Type()` may be useful down the line if we want to do use conditionals based on decoder type in the engine. Approved

zricethezav (approved):

rgmz:
> Seems like a worthwhile change even though we are adding another method to an interface.

Apparently it was already added in #3209. I'll clean up the duplicate methods.

pkg/engine/engine.go
```diff
@@ -797,7 +797,6 @@ func (e *Engine) scannerWorker(ctx context.Context) {
 					wgDoneFn: wgDetect.Done,
 				}
 			}
-			continue
```
zricethezav, line 800: Just double checking but this `continue` looks like a NOP right? @rgmz

rgmz: Correct. It's at the very end of the `for` block and thus has no actual effect.

PR #3015, opened 2024-06-27
Add endorlabs detector

shreyas-sriram:
### Description:
This PR adds a detector for [EndorLabs](https://docs.endorlabs.com/). Find the REST API documentation for authentication [here](https://docs.endorlabs.com/rest-api/authentication/).

### Checklist:


pkg/detectors/endorlabs/endorlabs.go
```diff
@@ -0,0 +1,114 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyAndSecretPat = regexp.MustCompile(detectors.PrefixRegex([]string{"endor"}) + `\b(endr\+[a-zA-Z0-9]+)\b`)
```
ahrav, line 27: question/suggestion: While creating test keys and secrets, I noticed that all were 16 characters long. I'm not sure if this is consistent across all subscription models, as I used the free trial.

If you also observe the 16-character limit after the prefix, we can refine the regex. For example: `\b(endr\+[a-zA-Z0-9-]{16})\b`

Additionally, we should include `-` in the character class since some of the keys I created had hyphens.

pkg/detectors/endorlabs/endorlabs_test.go
```diff
@@ -0,0 +1,247 @@
...
+			s:    Scanner{},
+			args: args{
+				ctx:    context.Background(),
+				data:   []byte(fmt.Sprintf("You can find a endorlabs key %s and endorlabs secret %s within", key, secret)),
+				verify: true,
+			},
+			want: []detectors.Result{
+				{
+					DetectorType: detectorspb.DetectorType_EndorLabs,
+					Verified:     false,
```
ahrav, line 121: suggestion: I think this should be `true` and L125 should be `false`. The first Cartesian product is key + secret (`true`), and the second is secret + key (`false`).

ahrav (approved):
Couple tiny suggestions and a question, but otherwise this looks great! Thanks @shreyas-sriram appreciate the contribution. 😄

pkg/detectors/endorlabs/endorlabs.go
shreyas-sriram: Yes, it is. I wanted to keep it generic to begin with; maybe we can look at it in the future if things don't seem right.

PR #3016, opened 2024-06-27
Added Twitter v2 Detector

abmussani:
### Description:
Twitter has modified their format of bearer token which is no longer detected by old detector and have not mentioned that old format is still working or not. As per the guide, this PR includes new version of twitter detector, also includes the test for it.

### Checklist:


0x1 (commented):
the v1 tests need version info added to the extra data

rgmz:
> the v1 tests need version info added to the extra data

Is there a reason this isn't done in the engine? Doing it by hand for every detector seems sloppy.

0x1:
> > the v1 tests need version info added to the extra data
> 
> Is there a reason this isn't done in the engine? Doing it by hand for every detector seems sloppy.

i don't think the engine can modify the want cases

rgmz:
> i don't think the engine can modify the want cases

Ah, I meant adding version info to the result automatically so you don't have to do this for every detector:
```go
  ExtraData: map[string]string{
    "version": fmt.Sprintf("%d", s.Version()),
  },
```

But you're correct that you'd still need to do:
```diff
 {
   DetectorType: detectorspb.DetectorType_Twitter,
+  DetectorVersion: 2,
   Verified:     true,
 },
```

0x1:
> > i don't think the engine can modify the want cases
> 
> Ah, I meant adding version info to the result automatically so you don't have to do this for every detector:
> 
> ```go
>   ExtraData: map[string]string{
>     "version": fmt.Sprintf("%d", s.Version()),
>   },
> ```
> 
> But you're correct that you'd still need to do:
> 
> ```diff
>  {
>    DetectorType: detectorspb.DetectorType_Twitter,
> +  DetectorVersion: 2,
>    Verified:     true,
>  },
> ```

fair point.

abmussani:
> > the v1 tests need version info added to the extra data
> 
> Is there a reason this isn't done in the engine? Doing it by hand for every detector seems sloppy.

I believe its not mandatory to implement Version for all the detectors. that's where engine cannot do that.

rosecodym:
> > > the v1 tests need version info added to the extra data
> > 
> > 
> > Is there a reason this isn't done in the engine? Doing it by hand for every detector seems sloppy.
> 
> I believe its not mandatory to implement Version for all the detectors. that's where engine cannot do that.

I believe that @rgmz is proposing new engine logic that checks to see if a detector has a version, and adding it to `ExtraData` if it does, but doing nothing otherwise. Is there any reason we couldn't do that?

abmussani:
> > > > the v1 tests need version info added to the extra data
> > > 
> > > 
> > > Is there a reason this isn't done in the engine? Doing it by hand for every detector seems sloppy.
> > 
> > 
> > I believe its not mandatory to implement Version for all the detectors. that's where engine cannot do that.
> 
> I believe that @rgmz is proposing new engine logic that checks to see if a detector has a version, and adding it to `ExtraData` if it does, but doing nothing otherwise. Is there any reason we couldn't do that?

@rosecodym It is (doable)[https://go.dev/play/p/ojkLkTvnI7Z] to detect detectors are implementing Version. I am not sure what will be impact of it, and also engine is suppose to modify the results.

zricethezav:
> I believe that @rgmz is proposing new engine logic that checks to see if a detector has a version, and adding it to ExtraData if it does, but doing nothing otherwise. Is there any reason we couldn't do that?

This is good, we should do this

0x1 (approved):

Issue #3017, opened 2024-06-27
Improved integration with LaunchDarkly


pkaeding:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description

Hi, I work on the LaunchDarkly product security team, and I'm working on adding a `caller-identity` API endpoint, which could be used to validate a LaunchDarkly credential (API token, SDK key, mobile key, etc). It will give some information about which account/environment that token belongs to, to help give the caller some context.

The idea is that this will work for any LD token, of any type, with any permission level (so you don't have to risk trying an API that might not be allowed for the credential you have, but it is still valid for other types of access).

My intended use-cases include tools like trufflehog, but also LD admins/users who find a token in a config management tool somewhere, and need to determine if it is still valid, and what it is used for. 

So, my question: would the Trufflehog project be interested in a contribution from me to update the [current](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/launchdarkly/launchdarkly.go) LaunchDarkly detector to use this new purpose-built API?

### Preferred Solution
I am proposing simplifying https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/launchdarkly/launchdarkly.go to call a single API that is made for this purpose, rather than using the API token list endpoint, and instantiating an LD SDK, to try to validate the different types of credentials.

#### References

* API Docs: https://apidocs.launchdarkly.com/tag/Other#operation/getCallerIdentity

PR #3018, opened 2024-06-27
update LaunchDarkly detector to use the caller-identity API

pkaeding:
### Description:
Updates the LaunchDarkly detector to use the new `caller-identity` API, built by the LaunchDarkly security team to support this kind of use case.

This way, secret scanners don't need to depend on other APIs, which may result in false-negative results if the credential found simply doesn't have permission to access that resource. Also, this means there is no need to instantiate an LD SDK to test SDK keys, which is a rather heavyweight operation (compared to what is needed here).

Also, improved the tests so they pass (unless I misunderstood what the tests were trying to do... they seemed to fail on main when I ran them, because the ExtraData was not present in the expected objects, and the `Scanner` object that was defined in the test cases was not being used by the tests, so the test case for the unexpected response code failed.

Addresses https://github.com/trufflesecurity/trufflehog/issues/3017

### Checklist:


pkg/detectors/launchdarkly/launchdarkly_test.go
```diff
@@ -99,30 +100,26 @@ func TestLaunchDarkly_FromChunk(t *testing.T) {
...
-					ctx:    context.Background(),
-					data:   []byte(fmt.Sprintf("You can find a launchdarkly sdk secret %s within", sdkSecret)),
-					verify: true,
-				},
-				want: []detectors.Result{
-					{
-						DetectorType: detectorspb.DetectorType_LaunchDarkly,
-						Verified:     true,
-					},
+		{
```
ahrav, line 103: ❤️

pkg/detectors/launchdarkly/launchdarkly_test.go
```diff
@@ -132,6 +129,14 @@ func TestLaunchDarkly_FromChunk(t *testing.T) {
 					t.Fatalf("no raw secret present: \n %+v", got[i])
 				}
 				got[i].Raw = nil
+
+				// Do we expect to be comparing the ExtraData?
```
ahrav, line 133: The template detector doesn't include the comparison feature because not all detectors support it. Implementors can choose to test it, but it's not required.

ahrav (approved):
Thanks so much @pkaeding for taking the time to get this updated, we really appreciate the improvements.

PR #3041, opened 2024-07-04
[feat] - Streamlined File Handling with BufferedReaderSeeker

ahrav:
### Description:
This PR introduces an enhancement to our file handling mechanism by integrating a `BufferedReaderSeeker`. The primary goal is to allow streaming files without writing them to disk before scanning for secrets.

#### Key Changes:

1. **BufferedReaderSeeker Implementation:**
   - A new type `BufferedReaderSeeker` is introduced, which provides a buffered reading interface with seeking capabilities.
   - It wraps an `io.Reader` and optionally an `io.Seeker`, enabling efficient reading and seeking operations, even on non-seekable underlying readers.

   **Main Features:**
   - **Buffered Reading:** Handles non-seekable readers by buffering the data, allowing for subsequent reads without re-reading from the source.
   - **Seeking Capability:** Supports seeking operations for both seekable and non-seekable readers, enabling random access reading.
   - **Buffer Management:** Includes the ability to disable buffering after initial reads, optimizing performance for subsequent operations. (Ex: after mime-type detection)

### Benchmarks and testing:
Screenshot 2024-07-04 at 1 50 50 PM

Filesytem scan over ~20GB:

#### BEFORE:

Write to disk:
Screenshot 2024-07-04 at 1 58 01 PM

**Scan Time: 2m 37s**
```

2024-07-04T13:56:04-07:00	info-0	trufflehog	running source	{"source_manager_worker_id": "08obj", "with_units": true}
2024-07-04T13:58:41-07:00	info-0	trufflehog	finished scanning	{"chunks": 1577988, "bytes": 21006119463, "verified_secrets": 0, "unverified_secrets": 0, "scan_duration": "2m37.427194291s", "trufflehog_version": "dev"}
./trufflehog filesystem ../json-files --no-update --no-verification   920.18s user 43.53s system 601% cpu 2:40.21 total
```

#### AFTER:

No write to disk:
Screenshot 2024-07-04 at 2 01 24 PM

**Scan Time: 2m 13s**
```

2024-07-04T13:59:27-07:00	info-0	trufflehog	running source	{"source_manager_worker_id": "PKwjp", "with_units": true}
2024-07-04T14:01:40-07:00	info-0	trufflehog	finished scanning	{"chunks": 1577988, "bytes": 21006119463, "verified_secrets": 0, "unverified_secrets": 0, "scan_duration": "2m13.164004083s", "trufflehog_version": "dev"}
./trufflehog filesystem ../json-files --no-update --no-verification   829.88s user 25.79s system 629% cpu 2:15.86 total
```

### Checklist:


jsvd:
The improvements seems "good" but not "amazing" only due to the fact that the typical dev stations nowadays have pretty good ssds with amazing write performance.

On many real situations where a machine in a cloud provider has a standard volume that is not backed by ssds, the improvement of this PR is **tremendous**:

```
$ time ./trufflehog filesystem /tmp/out_production.json -j --concurrency=24 --only-verified
{"level":"info-0","ts":"2024-07-05T08:49:22Z","logger":"trufflehog","msg":"running source","source_manager_worker_id":"kmnI3","with_units":true}
{"level":"info-0","ts":"2024-07-05T08:54:34Z","logger":"trufflehog","msg":"finished scanning","chunks":360149,"bytes":4793666710,"verified_secrets":0,"unverified_secrets":0,"scan_duration":"5m11.821583106s","trufflehog_version":"3.79.0"}

real	5m21.270s
user	13m47.248s
sys	0m22.904s

$ time ./trufflehog_3041 filesystem /tmp/out_production.json -j --concurrency=24 --only-verified
{"level":"info-0","ts":"2024-07-05T08:54:46Z","logger":"trufflehog","msg":"running source","source_manager_worker_id":"Gp1B7","with_units":true}
{"level":"info-0","ts":"2024-07-05T08:55:51Z","logger":"trufflehog","msg":"finished scanning","chunks":360149,"bytes":4793666710,"verified_secrets":0,"unverified_secrets":0,"scan_duration":"1m4.47256054s","trufflehog_version":"dev"}

real	1m7.586s
user	13m46.814s
sys	0m16.923s
```

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -0,0 +1,201 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+)
+
+// BufferedReaderSeeker provides a buffered reading interface with seeking capabilities.
+// It wraps an io.Reader and optionally an io.Seeker, allowing for efficient
+// reading and seeking operations, even on non-seekable underlying readers.
+type BufferedReaderSeeker struct {
```
mcastorina, line 13: **Nit:** `BufferedReadSeeker` ?

mcastorina (commented):
Are the git changes relevant to this PR?

pkg/sources/git/git.go
```diff
@@ -1238,29 +1238,51 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
+	// Read from stdout to prevent the pipe buffer from filling up and causing the command to hang.
+	// This allows us to stream the file contents to the handler.
+	go func() {
+		defer close(done)
+		done <- handlers.HandleFile(ctx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
+	}()
 
+	// Wait for the command to finish and the handler to complete.
+	// Capture any error from the file handling process.
+	processErr := <-done
```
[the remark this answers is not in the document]
ahrav: When I tried that it just hung indefinitely.

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -0,0 +1,201 @@
...
+				br.bytesRead += int64(n)
+			}
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			if err != nil {
+				return 0, err
+			}
+		}
+		newIndex = min(br.bytesRead+offset, br.bytesRead)
```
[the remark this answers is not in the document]
ahrav: Yea, but I can't think of a different way to handle this. The Archiver library mandates a Seeker, so implementing seek functionality for a non-seekable reader seemed like the only solution based on my research.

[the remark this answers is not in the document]
ahrav: We encounter this issue with certain archive types but not all, particularly when they are initially provided via a non-seekable reader, such as the output of `git cat-file` or any response body.

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -0,0 +1,198 @@
...
+		newIndex, err := br.seeker.Seek(offset, whence)
+		if err != nil {
+			return 0, fmt.Errorf("error seeking in reader: %w", err)
+		}
+		br.index = newIndex
+		if newIndex > br.bytesRead {
+			br.bytesRead = newIndex
+		}
+		return newIndex, nil
+	}
```
mcastorina, line 136: Why do we need to track the index internally if the underlying reader is `io.Seeker`?

ahrav: We probably don't need it. I likely added it, thinking we needed it for a different reason, but I can't remember what that was.

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -0,0 +1,198 @@
...
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+)
+
+// BufferedReadSeeker provides a buffered reading interface with seeking capabilities.
+// It wraps an io.Reader and optionally an io.Seeker, allowing for efficient
+// reading and seeking operations, even on non-seekable underlying readers.
```
mcastorina, line 12: I like that you can give `BufferedReadSeeker` any `io.Reader` and it'll figure out if it needs to buffer internally or not, but it seems to be doing two things. If it's an `io.Seeker`, it uses that. If not, it buffers. Am I understanding the functionality correctly?

If that's the case, I'd suggest refactoring the buffering functionality into an internal struct that does only that and utilizing it in the `BufferedReadSeeker`. I think it will make it easier to test / maintain / debug. I'm happy to take that on in a follow-up refactor if you agree.

ahrav: I understand the suggestion to refactor the buffering functionality into a separate internal struct. However, I believe keeping the buffering logic within the `BufferedReadSeeker` struct makes it easier to reason about for a few reasons:

1. **Cohesion**: The primary responsibility of `BufferedReadSeeker` is to provide a unified interface for reading and seeking, regardless of the underlying reader’s capabilities. By integrating the buffering logic directly within this struct, we maintain a cohesive design where all the necessary logic for handling read and seek operations is centralized.

2. **Simplicity**: Splitting the buffering logic into a separate struct would add additional complexity without significant benefits. The current implementation already effectively handles both scenarios (seekable and non-seekable readers) in a straightforward manner. Introducing another struct might complicate the design, leading to additional indirection.

3. **Test Coverage**: The existing tests _should_ cover the core functionalities of `BufferedReadSeeker`, ensuring that both buffered and direct seeking operations work as expected. Since the primary functionality is already well-tested, refactoring purely for the sake of separation doesn’t necessarily enhance testability or maintainability.

While I understand the potential benefits of separating concerns for clarity and modularity, in this case, I believe the current design strikes a good balance between simplicity and functionality.

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -0,0 +1,198 @@
...
+	buffer *bytes.Buffer // Internal buffer to store read bytes for non-seekable readers
+
+	bytesRead int64 // Total number of bytes read from the underlying reader
+	index     int64 // Current position in the virtual stream
+
+	// Flag to control buffering. This flag is used to indicate whether buffering is active.
+	// Buffering is enabled during initial reads (e.g., for MIME type detection and format identification).
+	// Once these operations are done, buffering can be disabled to prevent further writes to the buffer
+	// and to optimize subsequent reads directly from the underlying reader.
+	activeBuffering bool
```
mcastorina, line 26: What happens if we try to seek on a non-seeker reader and active buffering is disabled?

mcastorina (approved):

pkg/handlers/default_test.go
```diff
@@ -20,7 +20,6 @@ func TestHandleNonArchiveFile(t *testing.T) {
 
 	rdr, err := newFileReader(file)
 	assert.NoError(t, err)
-	defer rdr.Close()
```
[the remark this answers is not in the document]
ahrav: The newFileReader function used to handle closing the passed-in io.ReadCloser. However, it now accepts an io.Reader instead, eliminating the need for newFileReader to manage closure. If a calling function passes an io.ReadCloser (e.g., a response body), the caller must now handle closing it. This change simplifies the logic, in my opinion.

pkg/iobuf/bufferedreaderseeker.go
ahrav: We might receive invalid data. We can add a check to prevent this.

ahrav: Actually, i'll have to include that logic in the next [PR](https://github.com/trufflesecurity/trufflehog/pull/3052).

PR #3065, opened 2024-07-13
Atlassian Token Detector

ankushgoel27:
### Description:
This PR adds the Atlassian Admin api token detector.

I don't have a working V1 token to test but before 2023, there was an older format.

References:
https://support.atlassian.com/organization-administration/docs/manage-an-organization-with-the-admin-apis/
https://developer.atlassian.com/cloud/admin/organization/rest/api-group-orgs/#api-v1-orgs-get
atlassian

### Checklist:

### DIFF
--- a/pkg/detectors/atlassian/v2/atlassian.go
+++ b/pkg/detectors/atlassian/v2/atlassian.go
@@ -1,0 +1,120 @@
+package atlassian
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"encoding/json"
+
+	regexp "github.com/wasilibs/go-re2"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct {
+	client *http.Client
+}
+
+func (s Scanner) Version() int { return 2 }
+
+type OrgRes struct {
+	Data []struct {
+		Attributes struct {
+			Name string `json:"name"`
+		} `json:"attributes"`
+	} `json:"data"`
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.Versioner = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"atlassian"}) + `\b([A-Za-z0-9+/=_-]+=[A-Za-z0-9]{8})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"atlassian"}
+}
+
+// FromData will find and optionally verify Atlassian secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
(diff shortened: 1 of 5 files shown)


ankushgoel27 (commented):
Fixed version

pkg/detectors/atlassian/v2/atlassian.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.Versioner = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"atlassian"}) + `\b([A-Za-z0-9+/=_-]+=[A-Za-z0-9]{8})\b`)
```
ahrav, line 38: suggestion: I don’t believe the regex is quite correct. The `+` after the first character set `([A-Za-z0-9+/=_-]+)` allows for additional characters. While this is unlikely, we could avoid it by using exact character counts.

For example, here’s a revoked token:
```
ATCTT3xFfGN0lrL3T46HpEVWoOPlVEdOOIjhHfJizEXhYButp76pIT6y0YmHBfCq4-HVJ_PP0GMOkywZj42ns74TvCjtsXZE0SgqpvM-mc2gzwrATbVLgupnFHS5DpUgBA7CM5F0T-hJRd3QSch5aM33ih68tc92D00SnFCNbdR9Nw5Z2UtOwII=8A96C53E
```
But if I include a few more characters, it still matches:
```
ATCTT3xFfGN0lrL3T46HpEVWoOPlVEoooodOOIjhHfJizEXhYButp76pIT6y0YmHBfCq4-HVJ_PP0GMOkywZj42ns74TvCjtsXZE0SgqpvM-mc2gzwrATbVLgupnFHS5DpUgBA7CM5F0T-hJRd3QSch5aM33ih68tc92D00SnFCNbdR9Nw5Z2UtOwII=8A96C53E
```
I’ve also noticed that tokens created through the admin section often start with `ATCTT`, while others start with `ATATT`. They also share a "common identifier" after the prefix, `3xFfG`.

What do you think about tightening the regex to something like: `\b((?:ATCTT|ATATT)3xFfG[A-Za-z0-9-_=]{173}=[A-Za-z0-9]{8})\b`? This way, we can adjust the keywords to `ATATT` and `ATCTT` and remove the `detectors.PrefixRegex`.

If this detector is only for tokens prefixed with `ATCTT`, and we’ll have a separate detector for `ATATT`, we can use the single prefix.

ankushgoel27: Hello @ahrav 

```
On 18 Jan 2023, we extended the length of API tokens for Atlassian accounts. Using a varied length, rather than a fixed length, ensures tokens are more secure and reliable. Tokens created before 18 Jan 2023 were not affected.
```

according to atlassian documentation, the api keys generated are varied length . that's why i have the `+` sign. It is the same exact regex used in the jira detector.

I also tried to create multiple keys and this first part was the same for admin keys - `ATCTT3xFfGN0` and all keys were exact 192 char length so atlassian is still not using varied key length even though their documentation says otherwise.

i am happy to use your regex if it get rid of the prefix regex. Let's not use ATATT since it could clash with the jira detector.

let me know how to proceed.

ahrav: Thanks for the additional context. I as unaware of the variable length aspect. I think in that case, let's just update the prefix and keyword but use the original regex which included the _+_. By using the more specific keyword we'll be able to filter out a lot more potential matches earlier in the detection pipeline. 

Thanks for the explanation.

ankushgoel27: i have fixed the regex in my new commit. please validate

ahrav: I've added the test token and the tests look good. Thanks again Ankush, appreciate your patience on getting this in.

ahrav (approved):

PR #3066, opened 2024-07-15
[refactor] - Free Chunk Data

ahrav:
### Description:
This PR introduces optimizations to reduce memory usage and improve efficiency in the scanner worker process.

#### Changes

1. **Memory Optimization**: 
   - Setting `chunk.Data` to `nil` to allow the GC to reclaim the memory sooner, thereby reducing overall memory usage during scans.

2. **Efficiency Improvement**:
   - Updating the `FragmentLineOffset` function to process only a subset of the chunk data, rather than the entire chunk, improving both memory overhead and processing time.

NOTE: Easier to review without whitespace.

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -729,12 +729,15 @@ func (e *Engine) scannerWorker(ctx context.Context) {
 	var wgVerificationOverlap sync.WaitGroup
 
 	for chunk := range e.ChunksChan() {
-		startTime := time.Now()
-
-		// Free the chunk's data.
 		func() {
+			// Set chunk.Data to nil to allow the GC to reclaim the memory earlier.
+			// This field can contain up to 64kB of data, so keeping it around
+			// for the entire duration of the scan can lead to higher memory usage.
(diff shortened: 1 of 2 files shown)


mcastorina:
> Updating the FragmentLineOffset function to process only a subset of the chunk data

Does that affect the line number calculation? Theoretically it seems like it would, but I can't tell from the code.

ahrav:
> > Updating the FragmentLineOffset function to process only a subset of the chunk data
> 
> Does that affect the line number calculation? Theoretically it seems like it would, but I can't tell from the code.

oof, I _think_ we do need the whole chunk. That's a shame 😢

mcastorina:
> > > Updating the FragmentLineOffset function to process only a subset of the chunk data
> > 
> > 
> > Does that affect the line number calculation? Theoretically it seems like it would, but I can't tell from the code.
> 
> oof, I _think_ we do need the whole chunk. That's a shame 😢

darn 😞 

Add this to the list of reasons we need to refactor line number calculations..

PR #3068, opened 2024-07-15
implemented a netsuite detector

abmussani:
### Description:
Implemented a netsuite detector. The test are not working and under discussion.  The netsuite credentials will be consist of:

- Consumer key (64 characters)
- Consumer Secret (64 characters)
- Token Key (64 characters)
- Token Secret (64 characters)
- Account ID 

Due to similar nature of token, It was difficult to write test. Based on @ahrav's suggestion, I wrote test based on count of results. Along with that, The credentials I have used to test, are propriety of someone else and cannot be added in GCP detectors buckets. But I have put random strings in KEYs. If someone wants to run the netsuite's test, then first need to replace them with correct one.

### Checklist:


pkg/detectors/netsuite/netsuite.go
```diff
@@ -0,0 +1,227 @@
...
+		"oauth_nonce=\"" + params["nonce"] + "\"," +
+		"oauth_version=\"" + params["version"] + "\"," +
+		"oauth_signature=\"" + url.QueryEscape(signature) + "\""
+}
+
+// NetsuiteNoncer reads 11 bytes from crypto/rand and
+// returns those bytes as string
+type NetsuiteNoncer struct{}
+
+func (n NetsuiteNoncer) Nonce() string {
```
ahrav, line 217: question: How come you opted for a method on a struct instead of just a function? For example:
```
func nonce() (string, error) {
	const nonceSize = 11
	b := make([]byte, nonceSize)
	if _, err := rand.Read(b); err != nil {
		return "", err
	}

	return base64.RawURLEncoding.EncodeToString(b)[:nonceSize], nil
}
```

pkg/detectors/netsuite/netsuite.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// Nonce provides a random nonce string.
+	var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
+
+	b := make([]rune, 11)
+	for i := range b {
+		b[i] = letterRunes[rand.Intn(len(letterRunes))]
```
ahrav, line 224: suggestion: Using the `crypto/rand` package will handle all the heavy lifting and provide us with cryptographically secure randomness.

pkg/detectors/netsuite/netsuite.go
abmussani: @ahrav thanks, its a good suggestion but I have not worked on code cleaning yet. Authentication on netsuite was a hassle due to signature, So I just copied the code from library to generate the nonce.

pkg/detectors/netsuite/netsuite.go
abmussani: @ahrav I copied the code from some library. Let me use `crypto/rand` for random string generation.

pkg/detectors/netsuite/netsuite.go
```diff
@@ -0,0 +1,253 @@
...
+	h.Write([]byte(message))
+
+	// Get the resulting HMAC as a byte slice
+	hash := h.Sum(nil)
+
+	// Encode the byte slice to a hexadecimal string
+	return base64.StdEncoding.EncodeToString(hash)
+}
+
+func makeSignature(method, baseUrl, path string, params map[string]string) string {
```
ahrav, line 190: question/suggestion: Do you think we should use a strings.Builder here to avoid all these concatenations? It's a little more verbose but it would be more performant. I'm not sure if the tradeoff is worth it?

Something like:
```
func addOAuthParam(builder *strings.Builder, key, value string) {
	builder.WriteString(key)
	builder.WriteString("=\"")
	builder.WriteString(value)
	builder.WriteString("\",")
}

func makeSignature(method, baseUrl, path string, params map[string]string) string {
	var baseStringBuilder strings.Builder

	baseStringBuilder.WriteString(method)
	baseStringBuilder.WriteString("&")
	baseStringBuilder.WriteString(url.QueryEscape(baseUrl + path))
	baseStringBuilder.WriteString("&")
	baseStringBuilder.WriteString(url.QueryEscape(
		"oauth_consumer_key=" + url.QueryEscape(params["consumer_key"]) +
			"&oauth_nonce=" + url.QueryEscape(params["nonce"]) +
			"&oauth_signature_method=" + url.QueryEscape(params["signature_method"]) +
			"&oauth_timestamp=" + url.QueryEscape(params["timestamp"]) +
			"&oauth_token=" + url.QueryEscape(params["token_id"]) +
			"&oauth_version=" + url.QueryEscape(params["version"]),
	))

	baseString := baseStringBuilder.String()
	key := url.QueryEscape(params["consumer_secret"]) + "&" + url.QueryEscape(params["token_secret"])

	signature := generateHMACSHA256(baseString, key)

	var authHeaderBuilder strings.Builder

	authHeaderBuilder.WriteString("OAuth realm=\"")
	authHeaderBuilder.WriteString(params["realm"])
	authHeaderBuilder.WriteString("\",")

	addOAuthParam(&authHeaderBuilder, "oauth_consumer_key", params["consumer_key"])
	addOAuthParam(&authHeaderBuilder, "oauth_token", params["token_id"])
	addOAuthParam(&authHeaderBuilder, "oauth_signature_method", params["signature_method"])
	addOAuthParam(&authHeaderBuilder, "oauth_timestamp", params["timestamp"])
	addOAuthParam(&authHeaderBuilder, "oauth_nonce", params["nonce"])
	addOAuthParam(&authHeaderBuilder, "oauth_version", params["version"])
	addOAuthParam(&authHeaderBuilder, "oauth_signature", url.QueryEscape(signature))

	// Remove the trailing comma
	authHeader := authHeaderBuilder.String()
	authHeader = authHeader[:len(authHeader)-1]

	return authHeader
}
```

ahrav (approved):
Just a couple of small suggestions, otherwise, it looks good to me! Thanks for tackling this one; it definitely seems like it wasn't straightforward.

pkg/detectors/netsuite/netsuite.go
```diff
@@ -0,0 +1,253 @@
...
+		if _, exists := seen[cred]; exists {
+			return false
+		}
+		seen[cred] = struct{}{}
+	}
+	return true
+}
+
+// return a random nonce of 'n' character long
+func netsuiteNonce(n int) (string, error) {
```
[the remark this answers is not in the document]
abmussani: Umm, I did not think in that direction. Thank you for the suggestion. I will in corporate this change in my PR and test it.

pkg/detectors/netsuite/netsuite.go
abmussani: Using string builder is making code readability difficult. I am also not that sure about the performance.

PR #3075, opened 2024-07-18
fixed cloudflare global api key detector

ankushgoel27:
### Description:
This detector is failing due to the check on email match. when i initially submitted a PR #2101 , i fixed it there but this was not committed into the main then. Not sure why. it should have.

Also, not sure why the detector fails with the email check in place. ideally, it should work but it doesn't.

@zricethezav

### Checklist:


ankushgoel27:
readded the check in the second commit, if we compare to 3 instead of 2, the detector works fine

PR #3084, opened 2024-07-21
[refactor] - Mime Type Detection

ahrav:
### Description:
This PR refactors the handling of MIME type detection during file processing. Instead of performing this check at the final stage of the pipeline, it is moved to an earlier stage to avoid unnecessary operations.

### Checklist:


pkg/handlers/archive.go
```diff
@@ -89,7 +89,7 @@ func (h *archiveHandler) openArchive(ctx logContext.Context, depth int, reader f
 
 	if reader.format == nil {
 		if depth > 0 {
-			mtr, err := newSizedMimetypeReaderFromFileReader(reader)
+			mtr, err := newSizedReaderFromFileReader(reader)
 			if err != nil {
 				return fmt.Errorf("error reading MIME type: %w", err)
```
rosecodym, line 94: should this error message be updated?

pkg/handlers/default.go
```diff
@@ -44,7 +44,7 @@ func (h *defaultHandler) HandleFile(ctx logContext.Context, input fileReader) (c
 			h.metrics.incFilesProcessed()
 		}()
 
-		mtr, err := newSizedMimetypeReaderFromFileReader(input)
+		mtr, err := newSizedReaderFromFileReader(input)
 		if err != nil {
 			ctx.Logger().Error(err, "error reading MIME type")
```
rosecodym, line 49: should this error message be updated?

pkg/handlers/handlers.go
```diff
@@ -64,36 +64,67 @@ func newSizedMimetypeReaderFromFileReader(r fileReader) (sizedMimeTypeReader, er
...
-func newSizedMimeTypeReader(r io.Reader, size int64) (sizedMimeTypeReader, error) {
+func newSizedReader(r io.Reader, size int64) (sizedReader, error) {
+	if r == nil {
+		return sizedReader{}, errors.New("reader is nil")
+	}
+	if size == 0 {
+		return sizedReader{}, errEmptyReader
+	}
+
+	bufReader, _, err := determineMIMEType(r)
```
rosecodym, line 84: Now you determine the MIME type but immediately throw it away? This seems to reverse all the optimizations you've been adding :/

pkg/handlers/handlers.go
```diff
@@ -64,36 +64,67 @@ func newSizedMimetypeReaderFromFileReader(r fileReader) (sizedMimeTypeReader, er
...
 
-// newSizedMimeTypeReader creates a new sizedMimeTypeReader from an io.Reader.
-// It uses a bufio.Reader to perform MIME type detection on the input reader
-// without consuming it, by peeking into the first 512 bytes of the input.
-// This encapsulates both the original reader and the detected MIME type information.
+// newSizedReader creates a new sizedReader from an io.Reader.
 // This function is particularly useful for specialized archive handlers
 // that need to pass extracted content to the default handler without modifying the original reader.
-func newSizedMimeTypeReader(r io.Reader, size int64) (sizedMimeTypeReader, error) {
+func newSizedReader(r io.Reader, size int64) (sizedReader, error) {
```
rosecodym, line 76: It would never occur to me that this function could fail because of an unsupported file type - that behavior isn't hinted at in the name or the doc comment.

pkg/handlers/handlers.go
```diff
@@ -295,12 +326,12 @@ func HandleFile(
...
-			ctx.Logger().V(5).Info("empty reader, skipping file")
-			return nil
-		}
-		return fmt.Errorf("error creating custom reader: %w", err)
+	if errors.Is(err, errEmptyReader) {
+		ctx.Logger().V(5).Info("empty reader, skipping file")
+		return nil
+	} else if errors.Is(err, errUnsupportedMIME) {
+		ctx.Logger().V(5).Info("skipping file")
+		return nil
```
rosecodym, line 334: Why not use `handleReaderError` here?

rosecodym (commented):
I'm pretty confused by adding undocumented file type detection way down in the call stack (specific comments inline), as well as the thrown-away MIME type detection.

PR #3113, opened 2024-07-27
[chore] Use string instead of time.Time for expiration metadata

mcastorina:
### Description:
`time.Time` should not be used due to unsupported types in protobufs.

### Checklist:


dustin-decker:
I think we should consider using json instead of structpb so we can support timestamps. Structpb types are pretty barebones

PR #2887, opened 2024-05-27
Modularize scanning engine

ahrav:
### Description:
This PR refactors the Engine to be more modular.

Note: This is easier to review with "Hide whitespace" enabled.

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #3115, opened 2024-07-29
Include default detectors when using a config that contains detectors

harmonherring-pro:
### Description:
After #2887, Trufflehog no longer included the default detectors when using a config file to pass custom detectors to the engine. This PR ensures that default detectors are still included even when using a config file with custom detectors in it. 

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -308,9 +308,7 @@ func (e *Engine) setDefaults(ctx context.Context) {
 		e.decoders = decoders.DefaultDecoders()
 	}
 
-	if len(e.detectors) == 0 {
-		e.detectors = DefaultDetectors()
-	}
+	e.detectors = append(e.detectors, DefaultDetectors()...)
 
 	if e.dispatcher == nil {
 		e.dispatcher = NewPrinterDispatcher(new(output.PlainPrinter))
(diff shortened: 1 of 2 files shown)


ahrav (approved):
Thank you @harmonherring-pro for fixing this. Really appreciate it 😄

berenteb:
Is there any option to revert back to this behavior? We have several false positive results and use a custom regex detector only which suits our needs. Also it is strange that I see this new behavior with version 3.80.1 which worked before.

Don't get me wrong, having this behavior is also a great addition for those who want to extend the base detector set.

rosecodym:
@berenteb does using the `--include-detectors` flag and providing your custom detector's name allow you to get the previous behavior?

berenteb:
> you

It might be, but as I tried it did not recognise the detector I defined in the config.

rosecodym:
reverted by #3147

mcastorina:
Re-fixed in https://github.com/trufflesecurity/trufflehog/pull/3152 (sorry I didn't see this thread and caused the accidental reversion).

@berenteb once the above is released, you should be able to use `--include-detectors customregex` to filter the detectors to just the custom ones.

There currently isn't a way to filter by the custom detector's name, so the best way to do that would probably be to edit the config file to comment out the custom detectors you don't want.

berenteb:
Okay, sounds okay for our use-case, thank you very much!

rosecodym:
Thanks for closing the loop on this @mcastorina!

PR #3131, opened 2024-07-31
Auth GitHub in Init

rosecodym:
### Description:
The GitHub source currently applies its authentication configuration as the first step of enumeration. This is incompatible with both targeted scans and scan job reports, and also means that authentication logic has to be duplicated into the validation flow. This PR moves it into `Init` so that it's available to targeted scans and, eventually, unit-specific scans. This also allows us to remove the copy of the old logic that was in `Validate`.

As part of the work I've also cleaned up the integration test suite. (Several of them were apparently disabled back when they ran on every push, but now that we're not doing that, we can re-enable them.)

### Checklist:


pkg/sources/github/github_integration_test.go
```diff
@@ -51,32 +51,24 @@ func TestSource_Token(t *testing.T) {
...
-	// user provided
-	_, _, err = git.CloneRepoUsingToken(ctx, token, "https://github.com/truffle-test-integration-org/another-test-repo.git", user)
+	err = s.enumerateWithApp(ctx)
 	assert.NoError(t, err)
 
-	// no user provided
-	_, _, err = git.CloneRepoUsingToken(ctx, token, "https://github.com/truffle-test-integration-org/another-test-repo.git", "")
-	assert.Error(t, err)
-
-	_, _, err = s.cloneRepo(ctx, "https://github.com/truffle-test-integration-org/another-test-repo.git", installationClient)
```
rosecodym, line 79: This removed code was testing implementation details that no longer exist.

pkg/sources/github/github.go
```diff
@@ -198,16 +188,19 @@ func (s *Source) Init(aCtx context.Context, name string, jobID sources.JobID, so
...
 	if err != nil {
 		return fmt.Errorf("error unmarshalling connection: %w", err)
 	}
 	s.conn = &conn
 
+	connector, err := newConnector(s)
+	if err != nil {
+		return fmt.Errorf("could not create connector: %w", err)
+	}
+	s.connector = connector
```
mcastorina, line 202: Love this abstraction!

pkg/sources/github/github.go
```diff
@@ -369,38 +305,32 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, tar
...
-		err                error
-	)
-
-	switch cred := s.conn.GetCredential().(type) {
-	case *sourcespb.GitHub_BasicAuth:
-		if err = s.enumerateBasicAuth(ctx, apiEndpoint, cred.BasicAuth); err != nil {
-			return nil, err
+func (s *Source) enumerate(ctx context.Context) error {
+	// I'm not wild about switching on the connector type here (as opposed to dispatching to the connector itself) but
+	// this felt like a compromise that allowed me to isolate connection logic without rewriting the entire source.
```
mcastorina, line 318: I agree with this comment. Currently enumeration is different based on how we access GitHub, but I wonder if in the future we could always enumerate the same way and ignore any errors. Maybe that's a bit wasteful with rate-limits though.. 🤔

mcastorina (commented):
This is a really great step towards cleaning up the GitHub source!

mcastorina (approved):

PR #3132, opened 2024-07-31
Add Analyze interface to Stripe

abmussani:
### Description:
Implements Analyzer interface for Stripe. Also linked to detector

### Checklist:

### DIFF
--- a/pkg/analyzer/analyzers/stripe/stripe.go
+++ b/pkg/analyzer/analyzers/stripe/stripe.go
@@ -61,20 +61,18 @@ func secretInfoToAnalyzerResult(info *SecretInfo) *analyzers.AnalyzerResult {
 			Metadata:           nil,
 			Parent:             nil,
 		}
-		result.UnboundedResources = append(result.UnboundedResources, *parentResource)
-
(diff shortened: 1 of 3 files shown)


pkg/analyzer/analyzers/stripe/stripe.go
```diff
@@ -16,9 +16,75 @@ import (
...
+func secretInfoToAnalyzerResult(info *SecretInfo) *analyzers.AnalyzerResult {
+	if info == nil {
+		return nil
+	}
+	result := &analyzers.AnalyzerResult{
+		AnalyzerType: analyzerpb.AnalyzerType_Stripe,
+		Metadata: map[string]any{
+			"key_type": info.KeyType,
+			"key_env":  info.KeyEnv,
+			"valid":    info.Valid,
```
dustin-decker, line 54: I think you can drop this field, we already know it's valid at this point.

dustin-decker (approved):

pkg/analyzer/analyzers/stripe/stripe.go
abmussani: Fair enough. Already returning error if it is not valid.

PR #3139, opened 2024-08-01
[analyze] Add Analyzer for Asana

abmussani:
### Description:
Implements Analyzer interface for Asana. Also linked to detector

### Checklist:


pkg/analyzer/analyzers/asana/asana.go
```diff
@@ -102,3 +148,50 @@ func printMetadata(me *MeJSON) {
...
+		TeamWrite,
+		UserRead,
+		UserWrite,
+		UserTaskRead,
+		UserTaskWrite,
+		MembershipRead,
+		MembershipWrite,
+		RuleRead,
+		RuleWrite,
+	}
```
dustin-decker, line 186: nit: you can just iterate one of the generated Permission maps.

dustin-decker: Does this access depend on the permissions of the user that generated the token?

abmussani: > Does this access depend on the permissions of the user that generated the token?

In Asana, Personal Access Tokens (PATs) do not have specific, customizable permissions. The permissions of a PAT are inherently tied to the permissions of the Asana user account that creates the token

pkg/analyzer/analyzers/asana/asana.go
```diff
@@ -25,14 +26,59 @@ type Analyzer struct {
...
+
+	// resoures/permission setup
+	permissions := allPermissions()
+	userResource := analyzers.Resource{
+		Name: info.Data.Name,
+		Type: "user",
+		Metadata: map[string]any{
+			"email": info.Data.Email,
+			"type":  info.Data.Type,
+		},
```
mcastorina, line 57: I think we've added the `FullyQualifiedResourceName` field since the time you originally made these changes, so this'll need to be updated.

pkg/analyzer/analyzers/asana/asana.go
```diff
@@ -102,3 +148,11 @@ func printMetadata(me *MeJSON) {
 	}
 	t.Render()
 }
+
+func allPermissions() []analyzers.Permission {
+	permissions := make([]analyzers.Permission, 1)
+	permissions = append(permissions, analyzers.Permission{
+		Value: "Full Access",
```
mcastorina, line 155: Two options:
1. Use `analyzers.FullAccess` constant for the value here
2. Generate permissions / data types for just this one type of permission

I'm leaning towards 2 for consistency, even if it may be overkill.

abmussani: I would also lean toward option # 2.

pkg/analyzer/analyzers/asana/asana.go
abmussani: Updated the resource ID as fullyqualifiedName.

pkg/analyzer/analyzers/asana/asana.go
```diff
@@ -25,19 +27,68 @@ type Analyzer struct {
...
+		},
+	}
+
+	for _, perm := range permissions {
+		binding := analyzers.Binding{
+			Resource:   userResource,
+			Permission: perm,
+		}
+		result.Bindings = append(result.Bindings, binding)
+	}
```
mcastorina, line 68: **Nit:** I think we could use `analyzers.BindAllPermissions` here

pkg/analyzer/analyzers/asana/asana.go
```diff
@@ -102,3 +153,13 @@ func printMetadata(me *MeJSON) {
...
 	t.Render()
 }
+
+func allPermissions() []analyzers.Permission {
+	permissions := make([]analyzers.Permission, 0, len(PermissionStrings))
+	for permission := range PermissionStrings {
+		permissions = append(permissions, analyzers.Permission{
+			Value: PermissionStrings[permission],
+		})
+	}
```
mcastorina, line 163: **Nit:** I think it's more idiomatic to iterate with the second values here:

```go
for _, permission := range PermissionStrings {
    permissions = append(permissions, analyzers.Permission{
        Value: permission,
    }
}
```

mcastorina (approved):
lgtm!

pkg/analyzer/analyzers/asana/asana.go
abmussani: Thanks for pointing it our. Code updated

pkg/analyzer/analyzers/asana/asana.go
```diff
@@ -59,13 +59,8 @@ func secretInfoToAnalyzerResult(info *SecretInfo) *analyzers.AnalyzerResult {
...
 
-	for _, perm := range permissions {
-		binding := analyzers.Binding{
-			Resource:   userResource,
-			Permission: perm,
-		}
-		result.Bindings = append(result.Bindings, binding)
-	}
+	// bindings to all permissions to resources
+	analyzers.BindAllPermissions(userResource, permissions...)
```
mcastorina, line 63: `BindAllPermissions` returns a `[]Binding`, so we'll need to save that in the result.

```go
result.Bindings = append(result.Bindings, analyzers.BindAllPermissions(userResource, permissions...)...)
```

... feel free to format or break it up in whichever way you see fit. I admit it's not the most readable as a one-liner 😅

abmussani: ** Feeling dumb ** 
Thank you for point it out. Fixed.

mcastorina (approved):

PR #3143, opened 2024-08-01
[chore] - move automaxprocs to init

ahrav:
### Description:
This needs to run before we parse the CLI flags.
### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -96,7 +96,7 @@ var (
 	jsonOut             = cli.Flag("json", "Output in JSON format.").Short('j').Bool()
 	jsonLegacy          = cli.Flag("json-legacy", "Use the pre-v3.0 JSON format. Only works with git, gitlab, and github sources.").Bool()
 	gitHubActionsFormat = cli.Flag("github-actions", "Output in GitHub Actions format.").Bool()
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -96,7 +96,7 @@ var (
 	jsonOut             = cli.Flag("json", "Output in JSON format.").Short('j').Bool()
 	jsonLegacy          = cli.Flag("json-legacy", "Use the pre-v3.0 JSON format. Only works with git, gitlab, and github sources.").Bool()
 	gitHubActionsFormat = cli.Flag("github-actions", "Output in GitHub Actions format.").Bool()
-	concurrency         = cli.Flag("concurrency", "Number of concurrent workers.").Default(strconv.Itoa(runtime.NumCPU())).Int()
+	concurrency         = cli.Flag("concurrency", "Number of concurrent workers.").Default(strconv.Itoa(runtime.GOMAXPROCS(0))).Int()
```
mcastorina, line 99: What's the difference between `NumCPU()` and `GOMAXPROCS(0)` ?

mcastorina (approved):

main.go
ahrav: Whoops.. I forgot to change that back after testing. Using GOMAXPROCS(0) is equivalent to GOMAXPROCS(runtime.NumCPU()) because both use the number of CPUs if 0 is passed. I just need to move the automaxprocs call to before the CLI flag parsing.

PR #3145, opened 2024-08-01
CFOR Commit Scanner

joeleonjr:
### Description:
GitHub's repository network architecture creates lots of Cross Fork Object References (see post: https://trufflesecurity.com/blog/anyone-can-access-deleted-and-private-repo-data-github). These aren't currently being scanned. This PR adds support to scan single GitHub repos for these commits. 

The code is housed in a file named `object_discovery.go` inside the `github_experimental` source. It eventually hooks into the `git.go` source file and isolated from the normal GitHub source. It includes two new CLI flags:

The general approach here is to query GitHub's GraphQL in a way that returns information for valid Short SHA-1 commit hashes. Since this is a try-and-guess approach, we have the `--collision-threshold` value to allow users to set their tolerance for the number of estimated hash collisions. The higher the tolerance, the faster the enumeration will go.

You can expect this process to take 20-30 minutes for a small-medium sized repo. Of course, it could be made to run faster with multiple GH tokens, parallel requests, or using another approach separate from GraphQL, but our goal was to be respectful of GitHub's servers. 

If GitHub changes their behavior regarding CFOR, then we might remove this in the near future or adapt it as required.

This should be viewed as an alpha version.

Also, this caches results as they're being processed into a new folder on disk: $HOME/.trufflehog/<user>/<repo>.

To run this:
```
./trufflehog github-experimental --object-discovery  --repo https://github.com/<user>/<repo>.git --debug
```

### Checklist:


dustin-decker (approved):

dustin-decker (approved):

pkg/sources/github_experimental/object_discovery.go
```diff
@@ -0,0 +1,654 @@
...
+		}
+		gitArgs = append(gitArgs, batch...)
+		_, err := runGitCommand(gitArgs)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Checkout each commit
+	// Note: path and worktree are needed or else git will do something funny with the actual cwd
```
rgmz, line 509: @joeleonjr What is the purpose of checking out each commit after fetching it?

(Also, what "funny" thing does Git do? :o)

joeleonjr: Not an expert, but this is my best understanding: When we call `git fetch origin <commit hash>`, it's downloading those commit objects to your local. However, it's not attached to any branch or working state until you check it out. As a result, if you run `trufflehog` against your local without first checking out that commit, it won't get scanned. It's possible a config update to the way we run git scanning would address this and we could remove the need to check out. If you want to test it out:

```
git clone https://github.com/joeleonjr/cfor_test
cd cfor_test 
trufflehog git file://.
git fetch origin 92398a91fde2f43813723a0ed52e64e6967f7e8e
trufflehog git file://.         
git checkout 92398a91fde2f43813723a0ed52e64e6967f7e8e
trufflehog git file://.           
```

Only the last time you run `trufflehog` above, will you see a canary token.

PR #3163, opened 2024-08-02
[bug] - Create a new context with timeout per request

ahrav:
### Description:
This PR creates a new context with a timeout for each call to FromData. This ensures a cancelled context only affects the request in flight.
### Checklist:

### DIFF
--- a/pkg/detectors/privatekey/ssh.go
+++ b/pkg/detectors/privatekey/ssh.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"time"
 
 	"golang.org/x/crypto/ssh"
 )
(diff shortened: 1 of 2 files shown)


mcastorina (approved):
Did you consider drying up these changes with a helper function? Or maybe a `const` so the timeouts are the same? Not sure if that's important to keep consistent.

ahrav:
> Did you consider drying up these changes with a helper function? Or maybe a `const` so the timeouts are the same? Not sure if that's important to keep consistent.

I might add a helper, but I'm unsure if they should be the same. The FromData in the verificationOverlapWorker should probably have a shorter timeout since verification is disabled. In fact, we might not even need it for that call.

PR #3164, opened 2024-08-02
[refactor] - return privatekey verification errors

ahrav:
### Description:
This PR logs any errors encountered during private key verification, ensuring the errors are both available in the result and recorded in the engine.

Screenshot 2024-08-02 at 2 44 52 PM

### Checklist:


rgmz:
Why both `verificationErrors` and `verificationErrs`? Doesn't the existing code do this, minus returning it with FromData?

ahrav:
> Why both `verificationErrors` and `verificationErrs`? Doesn't the existing code do this, minus returning it with FromData?


I didn't want to change the existing behavior where the error appears in the result. However, we still need the error returned from FromData. Otherwise, the only place to identify these errors would be in the results. This issue should be visible during the scan as early as possible. Additionally, if errors occur only on non-verified results and the scan is run with `--only-verified`, these errors will be completely missed.

rgmz:
I see. It seems like a big code smell to have two different ways of aggregating errors, and the intention behind the code isn't clear.

It would be clearer to use the same method throughout. e.g., having `allErrs` outside the loop and `resultErrs` inside the verify block.

```go
			for err := range errChan {
				allErrs = errors.Join(allErrs, err)
				resultErrs = errors.Join(resultErrs, err)
			}
```

PR #3172, opened 2024-08-05
Analyze TUI

mcastorina:
### Description:
TUI interface for the analyze sub-command.

### Checklist:

### DIFF
--- a/pkg/analyzer/cli.go
+++ b/pkg/analyzer/cli.go
@@ -26,6 +26,7 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/analyzers/square"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/analyzers/stripe"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/analyzers/twilio"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/config"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/tui"
 )
 
@@ -51,6 +52,9 @@ func Run(cmd string) {
 	if err != nil {
 		// TODO: Log error.
 		return
+	}
+	if secretInfo.Cfg == nil {
+		secretInfo.Cfg = &config.Config{}
 	}
 	switch keyType {
 	case "github":
--- a/pkg/analyzer/cli.go
+++ b/pkg/analyzer/cli.go
@@ -1,6 +1,9 @@
 package analyzer
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/analyzers"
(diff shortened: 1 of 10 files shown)


pkg/analyzer/tui/form.go
```diff
@@ -0,0 +1,89 @@
...
+		inputs = append(inputs, textinputs.InputConfig{
+			Label:    "Shopify URL",
+			Key:      "url",
+			Required: true,
+		})
+	}
+
+	form := textinputs.New(inputs).
+		SetHeader(titleStyle.Render(fmt.Sprintf("Configuring %s analyzer", keyType))).
+		SetFooter("⚠️  Running TruffleHog Analyze will send a lot of requests ⚠️").
```
dustin-decker, line 38: Can we make this dependent on an attribute on the analyzers? Not all do this.

dustin-decker (approved):
LGTM barring a couple of non-blocking points. 
Can we support filtering by typing?

pkg/analyzer/tui/form.go
mcastorina: Is there an existing attribute I can reference? If not, do you have a list of analyzers that do brute-force?

Also open to phrasing suggestions. We could change it to something like "Running TruffleHog Analyze may send a lot of requests"

pkg/analyzer/cli.go
```diff
@@ -24,227 +27,75 @@ import (
...
-	case sourcegraphScan.FullCommand():
-		cfg.LogFile = analyzers.CreateLogFileName("sourcegraph")
-		sourcegraph.AnalyzeAndPrintPermissions(cfg, *sourcegraphKey)
-	case shopifyScan.FullCommand():
-		cfg.LogFile = analyzers.CreateLogFileName("shopify")
-		shopify.AnalyzeAndPrintPermissions(cfg, *shopifyKey, *shopifyStoreURL)
-	case opsgenieScan.FullCommand():
-		cfg.LogFile = analyzers.CreateLogFileName("opsgenie")
-		opsgenie.AnalyzeAndPrintPermissions(cfg, *opsgenieKey)
+	switch keyType {
```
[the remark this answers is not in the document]
mcastorina: When the TUI exits it returns the keytype and a secretinfo struct in this function, so this is the part that's responsible for actually running the analysis.

mcastorina:
> LGTM barring a couple of non-blocking points. Can we support filtering by typing?

Filtering the list is supported but you have to hit `/` first. There's a help menu at the bottom of the screen (in gray, may be hard to spot).

You can also start the TUI via `trufflehog analyze keytype`.

dustin-decker:
> > LGTM barring a couple of non-blocking points. Can we support filtering by typing?
> 
> Filtering the list is supported but you have to hit `/` first. There's a help menu at the bottom of the screen (in gray, may be hard to spot).
> 
> You can also start the TUI via `trufflehog analyze keytype`.

Ah okay. I misread it as being part of the previous one, like `j = down / filter`. Is the formatting there customizable? Maybe a tab would make it a bit more visually distinct

pkg/analyzer/tui/form.go
dustin-decker: We'd have to look at the analyzers and define that attribute.

PR #3174, opened 2024-08-05
[analyze] Add Analyzer for Sendgrid

abmussani:
### Description:
Implemented analyzer interface for Sengrid.

### Checklist:


pkg/analyzer/analyzers/sendgrid/sendgrid.go
```diff
@@ -13,8 +13,95 @@ import (
...
+		Bindings:           []analyzers.Binding{},
+		UnboundedResources: []analyzers.Resource{},
+	}
+
+	for _, scope := range SCOPES {
+		resource := getCategoryResource(scope)
+
+		if len(scope.Permissions) == 0 {
+			result.UnboundedResources = append(result.UnboundedResources, *resource)
+		} else {
```
mcastorina, line 69: **Nit:**

```go
if len(scope.Permissions) == 0 {
    result.UnboundedResources = append(result.UnboundedResources, *resource)
    continue
}

for _, permission := range scope.Permissions {
// ...
```

pkg/analyzer/analyzers/sendgrid/sendgrid.go
```diff
@@ -13,8 +13,95 @@ import (
...
+		Name:               scope.Category,
+		FullyQualifiedName: scope.Category,
+		Type:               "category",
+		Metadata:           nil,
+	}
+
+	if scope.SubCategory != "" {
+		return &analyzers.Resource{
+			Name:               scope.SubCategory,
+			FullyQualifiedName: scope.SubCategory,
```
mcastorina, line 95: I think the `FullyQualifiedName` should be

```go
fmt.Sprintf("%s/%s", scope.Category, scope.SubCategory)
```

pkg/analyzer/analyzers/sendgrid/sendgrid.go
```diff
@@ -13,8 +15,96 @@ import (
...
+		AnalyzerType: analyzerpb.AnalyzerType_Sendgrid,
+		Metadata: map[string]any{
+			"key_type":     keyType,
+			"2fa_required": slices.Contains(info.RawScopes, "2fa_required"),
+		},
+		Bindings:           []analyzers.Binding{},
+		UnboundedResources: []analyzers.Resource{},
+	}
+
+	for _, scope := range SCOPES {
```
mcastorina, line 66: I would feel better about the permissions generated from this `SCOPES` object if the data structure used the generated permission type instead of strings. Otherwise, I'm worried we could get a permission that doesn't exist in that set.

abmussani: I need to iterate through that logic. The `AnalyzePermission` func was mutating this `SCOPES` object based on the categories and subcategories. that's why Keep it as it is.

abmussani: @mcastorina Changes are made to exclude permission which are not part of generated list.

mcastorina (approved):

PR #3177, opened 2024-08-05
[analyze] Separate SID from token in twilio analyzer

mcastorina:
### Description:
Instead of submitting a single key of `sid:token` it now has two inputs.

### Checklist:


dustin-decker (approved):
You'll need to change AnalyzeInfo in the detector as well

PR #3180, opened 2024-08-06
[analyze] Add analyzer for Postman

abmussani:
### Description:
Implemented analyzer for postman. 
- Test implemented
- Linked with detectors

### Checklist:


pkg/analyzer/analyzers/postman/postman.go
```diff
@@ -1,17 +1,116 @@
...
+
+	result := analyzers.AnalyzerResult{
+		AnalyzerType:       analyzerpb.AnalyzerType_Postman,
+		Metadata:           nil,
+		UnboundedResources: []analyzers.Resource{},
+	}
+
+	resource := analyzers.Resource{
+		Name:               info.User.User.FullName,
+		FullyQualifiedName: info.User.User.FullName,
```
mcastorina, line 52: Is there a more unique value we can use for the `FullyQualifiedName`?

pkg/analyzer/analyzers/postman/postman.go
```diff
@@ -1,17 +1,116 @@
...
+
+	permissions := bakePermissions(info.User.User.Roles)
+
+	// bind all permissions with resources
+	result.Bindings = analyzers.BindAllPermissions(resource, permissions...)
+
+	for _, workspace := range info.Workspace.Workspaces {
+		result.UnboundedResources = append(result.UnboundedResources, analyzers.Resource{
+			Name:               workspace.Name,
+			FullyQualifiedName: workspace.Name,
```
mcastorina, line 71: Same question here about uniqueness

pkg/analyzer/analyzers/postman/postman.go
```diff
@@ -1,17 +1,116 @@
...
+		permStr, err := perm.ToString()
+		if err != nil {
+			continue
+		}
+		permissions[count] = analyzers.Permission{
+			Value:  permStr,
+			Parent: nil,
+		}
+		count++
+	}
```
mcastorina, line 109: **Nit:** Instead of introducing a `count` variable, it might be better to allocate capacity then `append`, especially if we're not always assigning to the slice:

```go
permissions := make([]analyzers.Permission, 0, len(permissionMap))
for perm := range permissionMap {
    permStr, err := perm.ToString()
    if err != nil {
        continue
    }
    permissions = append(permissions, analyzers.Permission{
        Value: permStr,
    })
}
```

mcastorina: Although in this case, `err` *should* always be `nil`. But this pattern is generally less error prone when the loop body is non-trivial imo.

pkg/analyzer/analyzers/postman/postman.go
abmussani: I do have User's Email here which is unique.

pkg/analyzer/analyzers/postman/postman.go
abmussani: I do have workspace Id here which is unique.

pkg/analyzer/analyzers/postman/postman.go
```diff
@@ -1,17 +1,114 @@
...
+
+	result := analyzers.AnalyzerResult{
+		AnalyzerType:       analyzerpb.AnalyzerType_Postman,
+		Metadata:           nil,
+		UnboundedResources: []analyzers.Resource{},
+	}
+
+	resource := analyzers.Resource{
+		Name:               info.User.User.FullName,
+		FullyQualifiedName: info.User.User.Email,
```
mcastorina, line 52: I'm not sure if there can be postman tokens across different teams or domains, but do you think it makes sense to add that information to the fully qualified name?

```
<TeamDomain> / <TeamName> / <Email>
```

mcastorina (approved):

pkg/analyzer/analyzers/postman/postman.go
abmussani: I do not have that knowledge about sharing token across a team. The token I used had empty team domain and name. I can only see the email as unique string.

mcastorina (approved):

PR #3181, opened 2024-08-06
[analyze] Add Analyzer for Opsgenie

abmussani:
### Description:
Implemented analyzer interface for opsgenie.

### Checklist:


pkg/analyzer/analyzers/opsgenie/opsgenie.go
```diff
@@ -14,8 +14,79 @@ import (
...
+		result.UnboundedResources = append(result.UnboundedResources, analyzers.Resource{
+			Name:               user.FullName,
+			FullyQualifiedName: user.FullName,
+			Type:               "user",
+			Metadata: map[string]any{
+				"username": user.Username,
+				"role":     user.Role.Name,
+			},
+		})
+	}
```
mcastorina, line 85: This is appending to `result.UnboundedResources`, which was allocated as a slice of non-zero length above: `make([]analyzers.Resource, len(info.Users))`.

Either change the assignment here to use the index:
```go
for i, user := range info.Users {
    result.UnboundedResources[i] = analyzers.Resource{
        // ...
    }
}
```

or change the initial assignment to allocate capacity:
```go
make([]analyzers.Resource, 0, len(info.Users))
```

mcastorina (changes requested):
This is also missing the enumerated permissions / generated types.

pkg/analyzer/analyzers/opsgenie/opsgenie.go
abmussani: Nice catch 🙇

abmussani:
> This is also missing the enumerated permissions / generated types.

Add the enumerated permissions in the updates.

mcastorina (approved):

mcastorina (approved):

PR #3189, opened 2024-08-06
[bug] - Improve seekability check for stdout pipes in BufferedReadSeeker

ahrav:
### Description:
When handling files over stdout.Pipe, our current method of checking if a reader is seekable using a simple interface assertion is insufficient. The underlying type of the response from stdout.Pipe is an os.File, which typically implements the io.Seeker interface. However, in this case, it cannot be treated like a regular file and doesn't support seeking operations.

### Checklist:


pkg/handlers/handlers_test.go
```diff
@@ -167,6 +168,7 @@ func TestSkipArchive(t *testing.T) {
 
 func TestHandleNestedArchives(t *testing.T) {
 	file, err := os.Open("testdata/nested-dirs.zip")
+	// file, err := os.Open("/Users/ahrav.dutta/Thog/json-files/md_random_data.json.zip")
```
rosecodym, line 171: i doubt this is supposed to be here

rosecodym (commented):
This has a few changes that don't appear related to the described fix - are they actually related?

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -87,6 +129,23 @@ func (br *BufferedReadSeeker) Read(out []byte) (int, error) {
...
+			return 0, io.EOF
+		}
+		if _, err := br.tempFile.Seek(br.index, io.SeekStart); err != nil {
+			return 0, err
+		}
+		n, err := br.tempFile.Read(out)
+		br.index += int64(n)
+		return n, err
+	}
+
```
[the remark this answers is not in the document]
ahrav: After adjusting the seek logic there was a failing test that was masked. This was added to fix the test. I can delete the test and this code and add it separately if that's better?

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -53,32 +54,73 @@ type BufferedReadSeeker struct {
...
+	// Check if it's an *os.File.
+	file, isFile := r.(*os.File)
+	if isFile {
+		// Check if it's a pipe or character device.
+		fi, err := file.Stat()
+		if err == nil && (fi.Mode()&os.ModeNamedPipe != 0 || fi.Mode()&os.ModeCharDevice != 0) {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
ahrav: Nope, but that's less complex, so i'll update. Thanks.

ahrav:
> This has a few changes that don't appear related to the described fix - are they actually related?

Removed. Will add to a separate PR.

ahrav:
> > This has a few changes that don't appear related to the described fix - are they actually related?
> 
> Removed. Will add to a separate PR.

Added [here](https://github.com/trufflesecurity/trufflehog/pull/3198)

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -33,8 +33,9 @@ func init() { defaultBufferPool = pool.NewBufferPool(defaultBufferSize) }
 // on it. For non-seekable readers, seeking is emulated using the buffer or
 // temporary file.
 type BufferedReadSeeker struct {
-	reader io.Reader
-	seeker io.Seeker // If the reader supports seeking, it's stored here for direct access
+	reader     io.Reader
+	seeker     io.Seeker // If the reader supports seeking, it's stored here for direct access
+	isSeekable bool      // Indicates if the reader supports reliable seeking operations
```
mcastorina, line 38: Is it possible to remove `isSeekable` in favor of checking `seeker != nil`?

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -53,32 +54,52 @@ type BufferedReadSeeker struct {
...
+		buf      *buffer.Buffer
+		seeker   io.Seeker
+		seekable bool
+	)
 
-	var buf *buffer.Buffer
-	if seeker == nil {
+	if isSeekable(r) {
+		seeker, _ = r.(io.Seeker)
+		seekable = true
```
mcastorina, line 84: Since it's only used here, I'd suggest changing `isSeekable` to:

```go
func asSeeker(r io.Reader) io.Seeker { /* ... */ }
```

and using it like:

```go
if s := asSeeker(r); s != nil {
    seeker = s
} else {
    buf = defaultBufferPool.Get()
}
```

or maybe

```go
seeker = asSeeker(r)
if seeker == nil {
    buf = defaultBufferPool.Get()
}
```

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -74,22 +77,21 @@ func NewBufferedReaderSeeker(r io.Reader) *BufferedReadSeeker {
...
+		isSeekable bool
 	)
 
-	if isSeekable(r) {
-		seeker, _ = r.(io.Seeker)
-		seekable = true
-	} else {
+	seeker = asSeeker(r)
+	if seeker == nil {
+		isSeekable = true
```
mcastorina, line 87: `isSeekable` should be set to `false`, correct? But what do you think about removing `isSeekable` altogether in favor of `seeker != nil` checks?

ahrav: I initially had it like that but found the boolean a bit easier to reason about. I'll change it back.

mcastorina (approved):

PR #3192, opened 2024-08-07
[analyze] Add Analyzer for Postgres

abmussani:
### Description:
Implemented analyzer interface for postgres

### Checklist:


pkg/analyzer/analyzers/postgres/postgres.go
```diff
@@ -14,8 +14,168 @@ import (
...
+		}
+
+		// populate map to reference later for tables
+		dbNameToResourceMap[db.DatabaseName] = &dbResource
+
+		DB_PRIVILIGES := map[string]bool{
+			"connect": db.Connect,
+			"create":  db.Create,
+			"temp":    db.CreateTemp,
+		}
```
mcastorina, line 115: **Nit:** Naming conventions should be `dbPrivileges`

mcastorina (commented):
This looks good, but is missing the enumerated permissions and generated data types.

abmussani:
> This looks good, but is missing the enumerated permissions and generated data types.

Added enumerated permissions.

pkg/analyzer/analyzers/postgres/postgres.go
```diff
@@ -14,8 +16,168 @@ import (
...
+		FullyQualifiedName: info.User,
+		Type:               "user",
+		Metadata: map[string]any{
+			"role": info.Role,
+		},
+	}
+
+	var bindings []analyzers.Binding
+
+	for role_priv, exists := range info.RolePrivs {
```
mcastorina, line 81: **Nit:** `rolePriv`

mcastorina (approved):

mcastorina (approved):

pkg/analyzer/analyzers/postgres/postgres.go
```diff
@@ -14,8 +16,168 @@ import (
...
+	for dbName, tableMap := range info.TablePrivs {
+		dbResource, ok := dbNameToResourceMap[dbName]
+		if !ok {
+			continue
+		}
+
+		for tableName, tableData := range tableMap {
+			tableResource := analyzers.Resource{
+				Name:               tableName,
+				FullyQualifiedName: dbResource.Name + "." + tableName,
```
mcastorina, line 146: Is there a reason you use `.` here instead of `/`? I thought `/` was the standard for separating fields in the fully qualified name.

Also, after reviewing the MySQL analyzer I _think_ we may need to make this more unique to include the URI or hostname?

abmussani: @mcastorina Usually, in queries, tables are reference with database name appended with '.' To have similarity with that, I made fully qualified name. We can go for '/' as well.

Hostname seems to be a good idea to make fully qualified name unique.

mcastorina: Ah, if it makes more sense to use `.` then go for it, I just wasn't aware of that idiom!

mcastorina (approved):

pkg/analyzer/analyzers/postgres/postgres.go
mcastorina: I'm going to go ahead and merge now, but feel free to open another PR if you'd like to make that change.

PR #3194, opened 2024-08-07
Fixes for a few finegrained token issues

dustin-decker:
### Description:
Fixes a few issues encountered with finegrained tokens.

### Checklist:

### DIFF
--- a/pkg/analyzer/analyzers/github/finegrained/finegrained.go
+++ b/pkg/analyzer/analyzers/github/finegrained/finegrained.go
@@ -428,11 +428,6 @@ func getDependabotAlertsPermission(client *gh.Client, repo *gh.Repository, curre
 	// Risk: Extremely Low
 	// GET /repos/{owner}/{repo}/dependabot/alerts
 	_, resp, err := client.Dependabot.ListRepoAlerts(context.Background(), *repo.Owner.Login, *repo.Name, &gh.ListAlertsOptions{})
-	if err != nil {
-		if !strings.Contains(err.Error(), "disabled") {
-			return NoAccess, err
-		}
-	}
 	defer resp.Body.Close()
 
 	switch resp.StatusCode {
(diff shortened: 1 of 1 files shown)


pkg/analyzer/analyzers/github/finegrained/finegrained.go
```diff
@@ -1336,31 +1331,32 @@ func PrintFineGrainedToken(cfg *config.Config, info *common.SecretInfo) {
...
+		keys = append(keys, val)
 	}
 	// Sort the slice
 	sort.Strings(keys)
 
-	for _, key := range keys {
-		value := accessMap[key]
-		if value == Invalid {
+	for idx, permStr := range keys {
+		perm := accessMap[idx]
```
mcastorina, line 1350: If we're relying on the index on line 1350, I think we also need to remove the `sort` on 1347.

pkg/analyzer/analyzers/github/finegrained/finegrained.go
```diff
@@ -428,11 +428,6 @@ func getDependabotAlertsPermission(client *gh.Client, repo *gh.Repository, curre
 	// Risk: Extremely Low
 	// GET /repos/{owner}/{repo}/dependabot/alerts
 	_, resp, err := client.Dependabot.ListRepoAlerts(context.Background(), *repo.Owner.Login, *repo.Name, &gh.ListAlertsOptions{})
-	if err != nil {
-		if !strings.Contains(err.Error(), "disabled") {
-			return NoAccess, err
-		}
-	}
```
mcastorina, line 435: It looks like this `err` is completely unchecked now, which seems wrong.

pkg/analyzer/analyzers/github/finegrained/finegrained.go
dustin-decker: Fixed, just accessing directly now

pkg/analyzer/analyzers/github/finegrained/finegrained.go
dustin-decker: It's returned in the default case

mcastorina (approved):

PR #3201, opened 2024-08-08
Log when a detector ignores the timeout

rosecodym:
### Description:
If a detector ignores the configured timeout it is _probably_ because of I/O blocking, which degrades the efficiency of the detector worker pool when it happens a lot. In the worst case, a detector that fully hangs will zombify its worker, causing really bad performance problems. When this happens, we don't really have a good way to notice other than seeing scan throughput drop suspiciously. This PR adds explicit logging when detection takes longer than it should so we have a better chance of catching this.

(This problem theoretically can spring up anywhere, in any worker, but the detector fleet is vast, uses network I/O, and is implemented by a much larger group of people, so this sort of problem is much more likely to slip into detector implementations than anywhere else in the codebase. We _could_ generalize this mechanism, but I don't want to make that investment before seeing if this smaller change captures the information we need.)

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -29,6 +29,8 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
 
+const detectionTimeout = 10 * time.Second
+
 var errOverlap = errors.New(
 	"More than one detector has found this result. For your safety, verification has been disabled." +
 		"You can override this behavior by using the --allow-verification-overlap flag.",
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -1000,13 +1002,14 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
 		detectBytesPerMatch.Observe(float64(len(matchBytes)))
 
 		ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+		t := time.AfterFunc(time.Second*11, func() {
```
trufflesteeeve, line 1005: __optional nitpick__: It might be worth setting the `time.Second * 10` into a variable and use the reference for `AfterFunc` so that an adjustment to the context timeout doesn't cause the `time.Second * 11` to no longer be accurate.

trufflesteeeve (approved):
One suggestion, but LGTM.

ahrav (approved):

PR #3203, opened 2024-08-08
[bug] - Correctly Handle Large Files in BufferedReadSeeker

ahrav:
### Description:
This PR addresses a bug in the BufferedReadSeeker implementation where large files were not handled correctly. The fix ensures that the transition from in-memory buffering to disk-based buffering is seamless and efficient, allowing for proper management of large files.

### Checklist:

### DIFF
--- a/pkg/handlers/handlers_test.go
+++ b/pkg/handlers/handlers_test.go
@@ -390,3 +390,27 @@ func BenchmarkHandleTar(b *testing.B) {
 		assert.NoError(b, err)
 	}
 }
+
+func TestHandleLargeHTTPJson(t *testing.T) {
+	resp, err := http.Get("https://raw.githubusercontent.com/ahrav/nothing-to-see-here/main/md_random_data.json.zip")
+	assert.NoError(t, err)
+	defer func() {
+		if resp != nil && resp.Body != nil {
+			resp.Body.Close()
+		}
+	}()
+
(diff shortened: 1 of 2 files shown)


pkg/handlers/handlers_test.go
```diff
@@ -390,3 +390,27 @@ func BenchmarkHandleTar(b *testing.B) {
 		assert.NoError(b, err)
 	}
 }
+
+func TestHandleLargeHTTPJson(t *testing.T) {
+	resp, err := http.Get("https://raw.githubusercontent.com/ahrav/nothing-to-see-here/main/md_random_data.json.zip")
+	assert.NoError(t, err)
```
rosecodym, line 396: i learned recently that `NoError` returns a value that you can use to bail on the rest of the test if appropriate

rosecodym (approved):
thanks!

pkg/handlers/handlers_test.go
ahrav: Do you prefer that over an explicit err != nil + t.Fatal?

ahrav: I ask because we also use the if err != nil + t.Fatal approach, and I want to ensure consistency if possible. I don't have a strong opinion either way, so i'll go with whatever people feel is more appropriate.

PR #3206, opened 2024-08-09
[analyze] Add Analyzer for Mailgun

abmussani:
### Description:
Implemented Analyzer interface for Mailgun.

### Checklist:


mcastorina (approved):

pkg/analyzer/analyzers/mailgun/mailgun.go
```diff
@@ -24,11 +26,64 @@ type Analyzer struct {
...
+	}
+
+	apiResource := analyzers.Resource{
+		Name:               "Mailgun API Key",
+		FullyQualifiedName: "Mailgun API Key",
+		Type:               "api key",
+		Metadata: map[string]any{
+			"domains_count": info.TotalCount,
+		},
+	}
```
mcastorina, line 58: Do you think we need to differentiate this resource somehow? Or is it okay for all keys (across all accounts) to have access to this `Mailgun API Key` resource?

Since we're adding the `domains_count` here it seems like it's specific to *this* key.

mcastorina (approved):

PR #3213, opened 2024-08-13
[th-563] new metrics for scan job reports

0x1:
### Description:

Seven new metrics for monitoring Scan Job Reports:

`backPressureOccurrences`: 
- frequency of back pressure occurrences (when the finishedMetrics channel is full)
- will tell us if metric generation is outpacing consumption

`reportGenerationTotal`, `reportGenerationErrors`:
- number of successful report generations vs. total attempts
- will tell us if reports are being generated as expected

`unitMetricsTotal`:
- number of UnitMetrics generated per job
- will tell us if all expected units are being processed

`unitMetricsWithErrors`:
- proportion of UnitMetrics containing errors
- will help us identify problematic scan units

`reportFileSize`:
- size of generated report files
- will help us detect unexpectedly large or small reports

`reportGenerationDuration`:
- time taken to generate each report
- will help us monitor performance and detect slowdowns

### Checklist:

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+)
+
+var (
+	reportGenerationTotal = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "report_generation_total",
+		Help:      "Total number of report generation attempts",
+	})
+
+	reportGenerationErrors = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "report_generation_errors",
+		Help:      "Total number of errors encountered during report generation, by error type",
+	},
+		[]string{"error_type"},
+	)
+
+	unitMetricsTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "unit_metrics_total",
+		Help:      "Total number of UnitMetrics generated, by unit type",
+	},
+		[]string{"unit_type"},
+	)
+
+	unitMetricsWithErrors = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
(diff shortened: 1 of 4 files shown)


metrics.go
```diff
@@ -0,0 +1,59 @@
...
+		Help:      "Size of the generated report file in bytes",
+	})
+
+	reportGenerationDuration = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "report_generation_duration_seconds",
+		Help:      "Time taken to generate a report",
+		Buckets:   prometheus.ExponentialBuckets(0.1, 2, 10),
+	})
```
mcastorina, line 58: This will have some correlation with the length of a scan, since the data is being handled as it is found, but before any detectors or decoders are used.

pkg/sources/job_progress_hook.go
```diff
@@ -39,6 +39,7 @@ func NewUnitHook(ctx context.Context, opts ...UnitHookOpt) (*UnitHook, <-chan Un
 		logBackPressure: func() {
 			once.Do(func() {
 				ctx.Logger().Info("back pressure detected in unit hook")
+				backPressureOccurrences.Inc()
 			})
 		},
```
mcastorina, line 44: The log message is in a `once.Do` to not flood the logs if it happens a lot. It would be better to move the metric outside of that:

```go
logBackPressure: func() {
    once.Do(func() {
        ctx.Logger().Info("back pressure detected in unit hook")
    })
    backPressureOccurrences.Inc()
},
```

pkg/sources/metrics.go
```diff
@@ -21,4 +21,11 @@ var (
 		Name:      "hooks_channel_size",
 		Help:      "Total number of metrics waiting in the finished channel.",
 	}, nil)
+
+	backPressureOccurrences = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "back_pressure_occurrences_total",
+		Help:      "Total number of times back pressure was detected",
```
mcastorina, line 29: It may be worth qualifying these values with the area of the codebase we're measuring (like the above uses a prefix `hooks_`).

metrics.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+)
+
+var (
+	reportGenerationTotal = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "report_generation_total",
+		Help:      "Total number of report generation attempts",
+	})
```
mcastorina, line 16: I'm not sure I understand what this metric is trying to measure, but looking at the code it seems covered by `unitMetricsTotal`.

PR #3218, opened 2024-08-14
[chore] Fix lint errors

mcastorina:
### Description:
Fixes some lint errors that were failing in the CI. I think `govet` was updated recently to start complaining about non-constant printf format strings.

### Checklist:

### DIFF
--- a/pkg/sources/source_manager_test.go
+++ b/pkg/sources/source_manager_test.go
@@ -207,7 +207,7 @@ func (c *unitChunker) ChunkUnit(ctx context.Context, unit SourceUnit, rep ChunkR
 			continue
 		}
 		if step.err != "" {
-			if err := rep.ChunkErr(ctx, fmt.Errorf(step.err)); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/sources/source_manager_test.go
```diff
@@ -207,7 +207,7 @@ func (c *unitChunker) ChunkUnit(ctx context.Context, unit SourceUnit, rep ChunkR
 			continue
 		}
 		if step.err != "" {
-			if err := rep.ChunkErr(ctx, fmt.Errorf(step.err)); err != nil {
+			if err := rep.ChunkErr(ctx, fmt.Errorf("%s", step.err)); err != nil {
```
dustin-decker, line 210: %w for error wrapping?

dustin-decker (approved):

pkg/sources/source_manager_test.go
mcastorina: This is in a test and `step.err` is a `string` so it's just converting the `string` into an `error` for testing.

PR #3221, opened 2024-08-15
Improve domain / url handling in detectors

dustin-decker:
### Description:
In some scenarios, input data could be crafted to have the scanner make a blind request to a target controlled by the input data. Typically this would effectively be a blind, unauthenticated GET request to the endpoint.

Mitigations include,
- stripping any paths or params, and setting those explicitly in the detector as needed
- disabling following redirects
- a http client that does not allow local IPs

Only a few detectors were affected by this issue, but the new http clients were also introduced for some detectors that were only detecting subdomains to a trusted domain, or domains to be used in a query parameter on the upstream provider.

### Checklist:


pkg/detectors/ringcentral/ringcentral.go
```diff
@@ -20,10 +20,11 @@ type Scanner struct {
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	client = detectors.DetectorHttpClientWithNoLocalAddresses
```
bill-rich, line 23: Only odd thing I'm seeing is `ringcentral` failing with the changes because verification is failing. I'm guessing that is because of the `NoLocalAddress` change.

bill-rich: The verification isn't against a local address though, so I'm not sure.
https://github.com/trufflesecurity/trufflehog/blob/d125e64751edc2e999bb0c32e4230debe42f0af2/pkg/detectors/ringcentral/ringcentral.go#L61

dustin-decker: Reverted this one. It might need to follow redirects.

bill-rich (approved):

PR #3226, opened 2024-08-16
[analyze] Add analyzer interface for Shopify

abmussani:
### Description:
This PR implements the analyzer interface for shopify.

### Checklist:


mcastorina (approved):

pkg/analyzer/analyzers/shopify/shopify.go
```diff
@@ -12,8 +15,80 @@ import (
...
+	result := analyzers.AnalyzerResult{
+		AnalyzerType: analyzerpb.AnalyzerType_Shopify,
+		Metadata: map[string]any{
+			"status_code": info.StatusCode,
+		},
+	}
+
+	resource := &analyzers.Resource{
+		Name:               info.ShopInfo.Shop.Name,
+		FullyQualifiedName: info.ShopInfo.Shop.Email,
```
mcastorina, line 66: It seems plausible that the email can be reused across shops. Maybe we can use the domain instead? There's a lot of information in the [shop.json API](https://shopify.dev/docs/api/admin-rest/2024-01/resources/shop) to choose from.

pkg/analyzer/analyzers/shopify/expected_output.json
```diff
@@ -0,0 +1,219 @@
...
+             "Metadata":{
+                "created_at":"2024-08-16T17:16:17+05:00"
+             },
+             "Parent":null
+          },
+          "Permission":{
+             "Value":"Read",
+             "Parent":null
+          }
+       },
```
mcastorina, line 33: These bindings look exactly the same. Are we setting the permission correctly?

mcastorina: It's also just "Read" which makes me think we're not using the generated permission values.

abmussani: thank you for pointing it out. I noted that permissions where being assigned to shop resource instead of categories. Fixed this issue.

mcastorina (approved):

PR #3235, opened 2024-08-20
Customize results cleaning (using smuggled interface)

rosecodym:
### Description:
We have identified some cases in which the results "cleaning" logic (the logic that eliminates superfluous results) should not run. In order to allow this, we need to expose the cleaning logic to the engine. This PR does so by doing these things:
* Create a `CustomResultsCleaner` interface that can be implemented by detectors that want to use custom cleaning logic
* Implement this interface for the `aws` and `awssessionkey` detectors (and remove their previous invocation of their custom cleaning logic)
* Modify the engine to invoke this logic (conditionally)

This PR also removes the "custom" cleaning logic for the `opsgenie`, `razorpay`, and `twilio` detectors, because it was added erroneously.

This is an alternative implementation of #3233.

### Checklist:


pkg/detectors/detectors.go
```diff
@@ -26,6 +26,22 @@ type Detector interface {
...
+// be implemented outside results generation because there are circumstances under which the engine should not execute
+// it.)
+type CustomResultsCleaner interface {
+	// CleanResults removes "superfluous" results from a result set (where the definition of "superfluous" is detector-
+	// specific).
+	CleanResults(results []Result) []Result
+	// ShouldCleanResultsIrrespectiveOfConfiguration allows a custom cleaner to instruct the engine to ignore
+	// user-provided configuration that controls whether results are cleaned. (User-provided configuration is not the
+	// only factor that determines whether the engine runs cleaning logic.)
+	ShouldCleanResultsIrrespectiveOfConfiguration() bool
```
mcastorina, line 42: Do you think this is something that would happen often?

mcastorina (approved):

pkg/detectors/detectors.go
rosecodym: It's needed for `aws` and `awssessionkey` today.

mcastorina: Hm.. what are your thoughts on special-casing `aws` and `awssessionkey` detectors instead of making this a required method for the interface?

rosecodym: I am extremely opposed to special-casing any detectors in the code. My position is that if we've created a bad abstraction, we should own that, rather than splitting it into multiple parts with invisible coupling just so that we don't have to look at the mess we've made.

PR #3243, opened 2024-08-21
Skip filtration for targeted scans

rosecodym:
### Description:
There is a scenario in which results filtration is known to cause problems, and this PR disables it in that scenario. (It _should_ cause problems more generally, but lacking any concrete cases of that, I want to tread lightly.)

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -905,7 +905,15 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 					detectorKeysWithResults[detector.Key] = detector
 				}
 
-				results = e.filterResults(ctx, detector, results)
+				// If results filtration eliminates a rotated secret, then that rotation will never be reported. This
+				// problem can theoretically occur for any scan, but we've only actually seen it in practice during
+				// targeted scans. (The reason for this discrepancy is unclear.) The simplest fix is therefore to
+				// disable filtration for targeted scans, but if you're here because this problem surfaced for a
+				// non-targeted scan then we'll have to solve it correctly.
+				if chunk.chunk.SecretID == 0 {
+					results = e.filterResults(ctx, detector, results)
+				}
+
 				for _, res := range results {
 					var val []byte
 					if res.RawV2 != nil {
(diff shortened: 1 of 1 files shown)


pkg/engine/engine.go
```diff
@@ -905,7 +905,15 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 					detectorKeysWithResults[detector.Key] = detector
 				}
 
-				results = e.filterResults(ctx, detector, results)
+				// If results filtration eliminates a rotated secret, then that rotation will never be reported. This
+				// problem can theoretically occur for any scan, but we've only actually seen it in practice during
+				// targeted scans. (The reason for this discrepancy is unclear.) The simplest fix is therefore to
+				// disable filtration for targeted scans, but if you're here because this problem surfaced for a
+				// non-targeted scan then we'll have to solve it correctly.
+				if chunk.chunk.SecretID == 0 {
```
mcastorina, line 913: I know there are comments saying `SecretID != 0` for targeted scans, but what do you think about adding a little helper function to whatever type `chunk.chunk` is.

```go
func (c Chunk) TargetedScan() bool { return c.SecretID != 0 }
```

pkg/engine/engine.go
```diff
@@ -1043,7 +1051,14 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
...
 		}
 
-		results = e.filterResults(ctx, data.detector, results)
+		// If results filtration eliminates a rotated secret, then that rotation will never be reported. This problem
+		// can theoretically occur for any scan, but we've only actually seen it in practice during targeted scans. (The
+		// reason for this discrepancy is unclear.) The simplest fix is therefore to disable filtration for targeted
+		// scans, but if you're here because this problem surfaced for a non-targeted scan then we'll have to solve it
+		// correctly.
+		if data.chunk.SecretID == 0 {
+			results = e.filterResults(ctx, data.detector, results)
```
mcastorina, line 1060: We filter in two places? 😕

mcastorina (approved):

pkg/engine/engine.go
rosecodym: yeah the verification overlap prevention basically just duplicated a stage of the pipeline :(

pkg/engine/engine.go
rosecodym: I like the sentiment, but I also hate that whether a scan is targeted is stored in _chunk_ data and I think I'd prefer to not double down on that.

Issue #3200, opened 2024-08-08
Postman module panic


domwhewell-sage:
### TruffleHog Version
3.81.7

### Trace Output
https://gist.github.com/domwhewell-sage/e3165ffcc78638e9d66cf74e1c17c7c8

### Expected Behavior

Trufflehog should scan the postman workspace

### Actual Behavior

Postman module panics when being run on a workspace ID. I have tried this on a few public workspaces and collections and get the same panic result.

### Steps to Reproduce

1. Identify a workspace in postman I used the public workspace "[Test examples in Postman](https://www.postman.com/postman/workspace/test-examples-in-postman/overview)"
2. Find the ID by selecting the 3 dots and "Workspace Info"
3. Copy the ID and paste it into the `--workspace-id=` option on the CLI
4. Run trufflehog. I used the command `trufflehog postman --workspace-id=70c7199b-6aee-49a6-a90f-025ad614f294`

## Environment
```
> lsb_release -a
No LSB modules are available.
Distributor ID: Kali
Description:    Kali GNU/Linux Rolling
Release:        2024.2
Codename:       kali-rolling
```

domwhewell-sage:
Thank you @LaraCroftDev this has resolved the issue!

PR #3245, opened 2024-08-23
Th 899 postman panic issue

LaraCroftDev:
Postman data can be accessed in two ways for scanning:
- either via an API call, for which the user needs to provide a valid token, or
- if the user has already exported the data, via WorkspacePaths, CollectionPaths, or EnvironmentPaths

If neither of these pieces of information is provided, the scanning should not be completed; the scanner should throw an error, and the error should be handled gracefully.

[The panic error](https://github.com/trufflesecurity/trufflehog/issues/3200) occurred because, in the absence of a token, the file paths were not checked correctly and the error was not returned. This PR should resolve the issue.
A new test file has been added to ensure that the checks are performed correctly.


joeleonjr:
LGTM!

pkg/engine/postman.go
```diff
@@ -26,17 +26,17 @@ func (e *Engine) ScanPostman(ctx context.Context, c sources.PostmanConfig) error
...
-	} else {
+	} else if len(c.WorkspacePaths) > 0 || len(c.CollectionPaths) > 0 || len(c.EnvironmentPaths) > 0 {
 		connection.Credential = &sourcespb.Postman_Unauthenticated{}
-	}
-
-	if len(c.Workspaces) == 0 && len(c.Collections) == 0 && len(c.Environments) == 0 && len(c.Token) == 0 && len(c.WorkspacePaths) == 0 && len(c.CollectionPaths) == 0 && len(c.EnvironmentPaths) == 0 {
-		ctx.Logger().Error(errors.New("no postman workspaces, collections, environments or API token provided"), "failed to scan postman")
-		return nil
+	} else {
+		return errors.New("no path to locally exported data or API token provided")
```
rosecodym, line 39: You've changed this function to return an error when the configuration is invalid, rather than `nil` - what are the consequences of that?

pkg/engine/postman_test.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		{
+			name: "scanned Postman with environmentPath",
+			postmanConfig: sources.PostmanConfig{
+				EnvironmentPaths: []string{"Downloads/Mobile - Points Unlock Redeemables.postman_environment.json"},
+			},
+		},
+		{
+			name: "no token or file path provided",
+			// postmanConfig: sources.PostmanConfig{},
```
rosecodym, line 40: Why is this commented out?

pkg/engine/postman_test.go
```diff
@@ -0,0 +1,89 @@
...
+			}()
+
+			err = e.ScanPostman(ctx, test.postmanConfig)
+			if err != nil && !test.wantErr && err.Error() != test.errMsg {
+				t.Errorf("ScanPostman() got: %v, want: %v", err, nil)
+				return
+			}
+			if err == nil && test.wantErr {
+				t.Errorf("ScanPostman() got: %v, want: %v", err, "error")
+			}
```
rosecodym, line 86: It looks like you've created two separate checks here: One that checks for a particular error message, and a second that checks simply for the presence of any error at all. I think that there's enough overlap that you don't need both, so I think it would be best to get rid of one. My vote would be to stop checking for a specific error message, because that test is more brittle than simply checking for any error at all.

rosecodym (commented):
This is great, thanks for finding it! I have one question about the changed implementation and a suggestion about your tests - feel free to reach out synchronously if you want to discuss that way.

pkg/engine/postman_test.go
```diff
@@ -0,0 +1,86 @@
...
+				Dispatcher:    NewPrinterDispatcher(new(discardPrinter)),
+			}
+
+			e, err := NewEngine(ctx, &conf)
+			assert.NoError(t, err)
+
+			e.Start(ctx)
+
+			go func() {
+				resultCount := 0
```
rosecodym, line 70: Do you actually read from `resultCount` anywhere? If not, I think you can remove it.

pkg/engine/postman.go
rosecodym: Oh, that's great, thanks for fixing it!

rosecodym (approved):

PR #3250, opened 2024-08-27
fixed gitlab regex

ankushgoel27:
### Description:
 The current regex contains a `=` which hinders detection. if the token is say - `gitlab_token=ttfitFp-eHzVZ3F478re`, the current regex will select `=ttfitFp-eHzVZ3F478rE` instead of `ttfitFp-eHzVZ3F478rE` for validation and the validation will fail. Also, I haven't seen `=` in a gitlab token before so I think we are safe to remove this. Let me know otherwise.

### Checklist:


pkg/detectors/gitlab/v1/gitlab.go
```diff
@@ -32,7 +32,7 @@ func (Scanner) DefaultEndpoint() string { return "https://gitlab.com" }
 
 var (
 	defaultClient = common.SaneHttpClient()
-	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"gitlab"}) + `\b([a-zA-Z0-9\-=_]{20,22})\b`)
+	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"gitlab"}) + `\b([a-zA-Z0-9\-_]{20,22})\b`)
```
kashifkhan0771, line 35: Similar changes are required for Gitlab v2 detector if we are sure that Gitlab token does not have `=` character in them.

kashifkhan0771:
I will also recommend adding Pattern test cases [like this one](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/eraser/eraser_test.go#L21). This will ensure the regex pattern detect the tokens whenever we make a change in the regex.

pkg/detectors/gitlab/v1/gitlab.go
kashifkhan0771: The reason I said is that we should be consistent as they key pattern is same. As mentioned in the description that `=` is not seen as part of the token. If we are updating regex for v1 I believe we should do it for v2 as well.

PR #3253, opened 2024-08-29
Added Flexport Detector

0x1:
This detector is for https://www.flexport.com/.
Flexport is a digital freight forwarding and logistics platform that helps businesses manage and optimize their global supply chains by providing end-to-end visibility and control over shipments. 

This detector is specifically for their logistics-api. There are other APIs that we should also add.

### Description:

### Checklist:


pkg/detectors/flexport/flexport.go
```diff
@@ -0,0 +1,100 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"flexport"}) + `\b(shltm_[0-9a-zA-Z-_]{40})\b`)
```
rgmz, line 26: Two thoughts:

1. The `shltm_` prefix seems unique enough that the flexport prefix isn't warranted.
2. The pattern `[0-9a-zA-Z-_]{40})\b` wouldn't match things like `abcdef-` (dash at end). Is it possible for tokens like that to be generated?

0x1: (1) agreed.
(2) that's a really good point. i'm not sure but will check

0x1: (2) we have not been able to reproduce a token with a dash at the end, but we did find tokens with multiple dashes, and as such i think we should ere on the side of this being possible. as such i think we can remove the ending boundary condition. thanks for pointing it out. does that seem okay?

pkg/detectors/flexport/flexport.go
```diff
@@ -0,0 +1,103 @@
...
+			s1.SetVerificationError(verificationErr, match)
+		}
+
+		results = append(results, s1)
+	}
+
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, map[string]string, error) {
```
kashifkhan0771, line 71: We are always returning `nil` for extraData.

abmussani (commented):
Thank you @0x1 for the new addition. To implement the required functionality, please ensure that the new detector is added within [defaults.go](https://github.com/abmussani/trufflehog/blob/5af94e252d7148417e3551d5fa41d60f31607e8c/pkg/engine/defaults.go#L1634)

0x1:
too many merge conflicts, closing in favor of: https://github.com/trufflesecurity/trufflehog/pull/3633

pkg/detectors/flexport/flexport.go
0x1: what do u mean?

PR #3263, opened 2024-09-02
Add Sentry protobufs

camgunz:
### Description:

### Checklist:

### DIFF
--- a/proto/source_metadata.proto
+++ b/proto/source_metadata.proto
@@ -331,6 +331,18 @@ message Elasticsearch {
   string timestamp = 3;
 }
 
+message Sentry {
+  string id = 1;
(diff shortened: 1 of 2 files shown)


proto/source_metadata.proto
```diff
@@ -331,6 +331,18 @@ message Elasticsearch {
   string timestamp = 3;
 }
 
+message Sentry {
+  string id = 1;
```
dustin-decker, line 335: I would prefer this to be named event_id so we know what it represents.

camgunz: Oof, dunno why I did that, will change 👍

dustin-decker (approved):

PR #3269, opened 2024-09-06
GitHub source logger clean up

LaraCroftDev:
This PR is for [TH-924](https://trufflesecurity.atlassian.net/browse/TH-924), it removes logger from GitHub source and uses context logger instead for logging. 
Related tests have been updated and they all pass.

[TH-924]: https://trufflesecurity.atlassian.net/browse/TH-924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rosecodym (commented):
great first step! as discussed, the next step is to pass `logContext.Context` objects instead of loggers

pkg/sources/github/github.go
```diff
@@ -705,12 +702,12 @@ func (s *Source) handleRateLimit(errIn error) bool {
 		if retryAfter > 0 {
 			retryAfter = retryAfter + jitter
 			rateLimitResumeTime = now.Add(retryAfter)
-			s.log.V(0).Info(fmt.Sprintf("exceeded %s rate limit", limitType), "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
+			ctx.Logger().V(0).Info(fmt.Sprintf("exceeded %s rate limit", limitType), "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
```
mcastorina, line 705: **Nit:** While we're here I think we can remove the `V(0)` as it's not doing anything (the default is 0).

pkg/sources/github/github.go
```diff
@@ -705,12 +702,12 @@ func (s *Source) handleRateLimit(errIn error) bool {
...
 		} else {
 			retryAfter = (5 * time.Minute) + jitter
 			rateLimitResumeTime = now.Add(retryAfter)
 			// TODO: Use exponential backoff instead of static retry time.
-			s.log.V(0).Error(errIn, "unexpected rate limit error", "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
+			ctx.Logger().V(0).Error(errIn, "unexpected rate limit error", "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
```
mcastorina, line 710: **Nit:** Same here, except the reason is slightly different in that `Error` logs do not have a verbosity level, so calling `V` does nothing here.

mcastorina (approved):
lgtm, thanks!

PR #2985, opened 2024-06-18
Update Azure application credentials detector

rgmz:
### Description:
This closes #1979, closes #2394, closes #3039, and closes #3362, 

Depends on #2976.

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #3271, opened 2024-09-07
Azure Detector Regex fix

ankushgoel27:
### Description:

Azure now uses a 40 char string as a secret for the service principal instead of 34 chars. Also, even both 34 and 40 char versions used [A-Z] which was missing from the current regex. I don't think there was a need for a v2 detector unless the format for the secret key was changed. This new regex will detect both 40 char and 34 char versions.

See this video for the 34 char version - https://www.youtube.com/watch?v=Kf1Tai_BkWU. At 3:19, it shows that the 34 char version also had [A-Z] in it.

### Checklist:

### DIFF
--- a/pkg/detectors/azure/azure.go
+++ b/pkg/detectors/azure/azure.go
@@ -33,7 +33,7 @@ var (
 	tenantIDPat = mustFmtPat("tenant_id", idPatFmt)
 
 	// TODO: support old patterns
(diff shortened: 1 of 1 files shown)


rgmz:
> I don't think there was a need for a v2 detector unless the format for the secret key was changed.

The new secrets are substantially different from the old ones. Length, allowed characters, fixed values, etc.

See #2985.

ankushgoel27:
closing this then

PR #3274, opened 2024-09-08
Separate detector tests into unit/integration

rgmz:
### Description:
Unlike integration tests, the detector's unit tests should always run. This helps catch bugs early; there is no upside and tremendous downside to keeping them together.

### Checklist:


zricethezav:
@rgmz I think this is a worthwhile change. Is there a reason only a subset of the detectors are included in this PR?

rgmz:
Those are the only ones that have unit tests AFAIK.

rgmz:
I'm confused by the lint error: https://github.com/trufflesecurity/trufflehog/actions/runs/10848445504/job/30590845421?pr=3274

That variable appears to have _never_ been used. I guess it didn't get flagged due to the flag?

kashifkhan0771:
Thanks @rgmz for this.
LGTM! It will be really cool once we have pattern test cases for all the existing detectors. I will also be working on this whenever I get a chance.

zricethezav:
Looks good to me, we should update the contributing docs too. That can be in a followup PR

PR #3283, opened 2024-09-11
Refactor GitHub source to allow asynchronous enumeration

mcastorina:
### Description:
There's quite a bit of spaghetti that I tried to leave untouched. This works by adding a `UnitReporter`, passing those values into a channel, and draining all the values out of the channel for the synchronous case.

This reorganizes the code but does not change the existing behavior of enumerating then scanning.

### Checklist:

### DIFF
--- a/pkg/sources/github/filter_set.go
+++ b/pkg/sources/github/filter_set.go
@@ -1,0 +1,77 @@
+package github
+
+import (
+	"github.com/gobwas/glob"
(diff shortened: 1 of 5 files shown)


rgmz:
Drive-by comment, but I like how they implemented async enumeration in https://github.com/boostsecurityio/poutine

mcastorina:
Closing in favor of #3292

PR #3297, opened 2024-09-13
Add user agent suffix feature flag

dustin-decker:
### Description:
Allows specifying a user agent suffix

### Checklist:

### DIFF
--- a/pkg/common/http.go
+++ b/pkg/common/http.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/feature"
 )
 
 var caCerts = []string{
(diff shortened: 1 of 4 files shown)


pkg/common/http.go
```diff
@@ -88,8 +89,15 @@ type CustomTransport struct {
 	T http.RoundTripper
 }
 
+func userAgent() string {
+	if len(feature.UserAgentSuffix.Load()) > 0 {
+		return "TruffleHog" + " " + feature.UserAgentSuffix.Load()
```
mcastorina, line 94: Is there a reason for the explicit concatenation of the space `"TruffleHog" + " "`? Instead of a single string `"TruffleHog "`.

mcastorina (approved):
Cool! I like the `AtomicString` struct to maintain consistency with our other feature flags.

pkg/common/http.go
dustin-decker: Overlooked that, thank you

PR #3296, opened 2024-09-13
Instrument GitHub source with a ChunkReporter

mcastorina:
### Description:
Replaces all `chan *Chunk` parameters with a `ChunkReporter`.

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

Issue #3005, opened 2024-06-23
Git source doesn't check if directory exists


rgmz:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

### TruffleHog Version
HEAD

### Description

When using TruffleHog with a `file://` uri, the scan will complete with a 0 exit code even if the directory doesn't exist.

```sh
$ ls /tmp/fake-dir/
ls: cannot access '/tmp/fake-dir/': No such file or directory
$ ./trufflehog git file:///tmp/fake-dir/
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2024-06-23T18:12:55-04:00       info-0  trufflehog      running source  {"source_manager_worker_id": "XIh5z", "with_units": true}
2024-06-23T18:12:55-04:00       info-0  trufflehog      finished scanning       {"chunks": 0, "bytes": 0, "verified_secrets": 0, "unverified_secrets": 0, "scan_duration": "4.326315ms", "trufflehog_version": "dev"}
$ echo $?
0
```

rgmz:
It seems that `git.scanDir` returns an error (`repository does not exist`), however, the error seems to vanish into a void. 

https://github.com/trufflesecurity/trufflehog/blob/de19a39f2cc6a103a4e207b780716d46e77de732/pkg/sources/git/git.go#L324-L330

https://github.com/trufflesecurity/trufflehog/blob/de19a39f2cc6a103a4e207b780716d46e77de732/pkg/sources/git/git.go#L1272

https://github.com/trufflesecurity/trufflehog/blob/de19a39f2cc6a103a4e207b780716d46e77de732/pkg/sources/source_manager.go#L369-L372


Confusingly, `scanDirs` and `scanRepos` — which would log such an error — don't seem to run in any configuration I tested. They only get called `runWithoutUnits`.

https://github.com/trufflesecurity/trufflehog/blob/de19a39f2cc6a103a4e207b780716d46e77de732/pkg/sources/source_manager.go#L286

sahil9001:
I have raised a PR for this @rgmz , please check

PR #3298, opened 2024-09-14
Implement SourceUnitEnumChunker for GitHub

mcastorina:
### Description:
Implements `SourceUnitEnumChunker` for GitHub source. When merged, `github` scans will use the `Enumerate` and `ChunkUnit` methods instead of `Chunks` to scan.

Builds off of #3296 and #3292

### Checklist:

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -642,6 +642,10 @@ func (s *Source) scan(ctx context.Context, reporter sources.ChunkReporter) error
 	return nil
 }
 
+// scanRepo attempts to scan the provided URL and any associated wiki and
+// comments if configured. An error is returned if we could not find necessary
+// repository metadata or clone the repo, otherwise all errors are reported to
+// the ChunkReporter.
 func (s *Source) scanRepo(ctx context.Context, repoURL string, reporter sources.ChunkReporter) error {
 	if !strings.HasSuffix(repoURL, ".git") {
 		return fmt.Errorf("repo does not end in .git")
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -84,20 +84,20 @@ var _ sources.SourceUnit = (*RepoUnit)(nil)
 var _ sources.SourceUnit = (*GistUnit)(nil)
 
 type RepoUnit struct {
-	name string
-	url  string
+	Name string `json:"name"`
+	URL  string `json:"url"`
 }
 
-func (r RepoUnit) SourceUnitID() (string, sources.SourceUnitKind) { return r.url, "repo" }
-func (r RepoUnit) Display() string                                { return r.name }
+func (r RepoUnit) SourceUnitID() (string, sources.SourceUnitKind) { return r.URL, "repo" }
+func (r RepoUnit) Display() string                                { return r.Name }
 
 type GistUnit struct {
-	name string
-	url  string
+	Name string `json:"name"`
+	URL  string `json:"url"`
 }
 
-func (g GistUnit) SourceUnitID() (string, sources.SourceUnitKind) { return g.url, "gist" }
-func (g GistUnit) Display() string                                { return g.name }
+func (g GistUnit) SourceUnitID() (string, sources.SourceUnitKind) { return g.URL, "gist" }
+func (g GistUnit) Display() string                                { return g.Name }
 
 // --------------------------------------------------------------------------------
 
@@ -373,7 +373,7 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	// Report any values that were already configured.
 	for _, name := range s.filteredRepoCache.Keys() {
 		url, _ := s.filteredRepoCache.Get(name)
-		_ = dedupeReporter.UnitOk(ctx, RepoUnit{name: name, url: url})
+		_ = dedupeReporter.UnitOk(ctx, RepoUnit{Name: name, URL: url})
 	}
 
 	// I'm not wild about switching on the connector type here (as opposed to dispatching to the connector itself) but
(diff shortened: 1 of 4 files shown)


pkg/sources/github/github.go
```diff
@@ -372,7 +373,11 @@ func (s *Source) enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	// Report any values that were already configured.
 	for _, name := range s.filteredRepoCache.Keys() {
 		url, _ := s.filteredRepoCache.Get(name)
-		_ = dedupeReporter.UnitOk(ctx, RepoUnit{name: name, url: url})
+		url, err := s.ensureRepoInfoCache(ctx, url)
+		if err != nil {
+			_ = dedupeReporter.UnitErr(ctx, err)
```
rosecodym, line 378: Why are we discarding these errors?

pkg/sources/github/github.go
```diff
@@ -372,7 +373,11 @@ func (s *Source) enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	// Report any values that were already configured.
 	for _, name := range s.filteredRepoCache.Keys() {
 		url, _ := s.filteredRepoCache.Get(name)
-		_ = dedupeReporter.UnitOk(ctx, RepoUnit{name: name, url: url})
+		url, err := s.ensureRepoInfoCache(ctx, url)
```
rosecodym, line 376: The day we get rid of the fourteen different caches in this source I will whistle a jaunty tune

pkg/sources/github/github.go
```diff
@@ -454,6 +416,58 @@ RepoLoop:
...
+// provided repository URL. If not, it fetches and stores the metadata for the
+// repository. In some cases, the gist URL needs to be normalized, which is
+// returned by this function.
+func (s *Source) ensureRepoInfoCache(ctx context.Context, repo string) (string, error) {
+	if _, ok := s.repoInfoCache.get(repo); ok {
+		return repo, nil
+	}
+	// Ensure that |s.repoInfoCache| contains an entry for |repo|.
+	// This compensates for differences in enumeration logic between `--org` and `--repo`.
+	// See: https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788
```
rosecodym, line 429: Am I understanding correctly that with this function extracted, this comment is somewhat out of date? It's not exactly _wrong_, but it doesn't tell the whole story anymore, and the information it contains seems like it belongs in the doc comment.

pkg/sources/github/github.go
```diff
@@ -603,7 +617,6 @@ func (s *Source) scan(ctx context.Context, reporter sources.ChunkReporter) error
 	reposToScan, progressIndexOffset := sources.FilterReposToResume(s.repos, s.GetProgress().EncodedResumeInfo)
 	s.repos = reposToScan
 
-	scanErrs := sources.NewScanErrors()
```
rosecodym, line 606: ❤️

pkg/sources/github/github.go
```diff
@@ -1478,3 +1500,9 @@ func (s *Source) scanTarget(ctx context.Context, target sources.ChunkingTarget,
 		Verify: s.verify}
 	return handlers.HandleFile(ctx, readCloser, &chunkSkel, reporter)
 }
+
+func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
+	repoURL, _ := unit.SourceUnitID()
+	ctx = context.WithValue(ctx, "repo", repoURL)
+	return s.scanRepo(ctx, repoURL, reporter)
+}
```
rosecodym, line 1508: I am worried about this, because this implementation only works when called on a source that previously had `Enumerate` called on it, but that precondition is not in any part of the formal or informal `ChunkUnit` contract, so this feels like an accident waiting to happen. I think the rest of this PR is great - is it possible to do this part as a separate, atomic unit of work?

rosecodym (commented):
The title _terrified_ me, but this ended up looking super clean, nice work! The one thing I'm worried about is `ChunkUnit`, which I commented on inline.

pkg/sources/github/github.go
mcastorina: I was being lazy :(

pkg/sources/github/github.go
mcastorina: Great callout! I think the only change necessary is a call to `ensureRepoInfoCache` before scanning. I've added an integration test that calls `Init()` followed by `ChunkUnit()` to ensure this works.

rosecodym (approved):
This looks great! But I have thought of another question - is running it in "actual separate jobs" mode (which we won't be doing yet), where the source scanning a repo is a distinct object from the source that enumerated, going to result in twice as many API calls to get repo information? We have users actively hitting rate limits _now_, so I think we're going to need to think up a solution for that. (I realize that I don't even know why we cache repo info in the first place.)

mcastorina:
> is running it in "actual separate jobs" mode going to result in twice as many API calls to get repo information?

Unfortunately, yes. It seems that we obtain repo metadata during enumeration, so we can opportunistically save it as chunk metadata. If we enumerate and scan in two separate steps, the current solution throws away the metadata during enumeration and fetches it again during the scan.

One idea is to formalize metadata rehydration, where we only fetch if we find a secret in the chunk. We're still duplicating some calls, but it is theoretically better than duplicating *all* calls.

rgmz:
@mcastorina Am I missing something or is [`scan`](https://github.com/trufflesecurity/trufflehog/blame/5130a9dbb0692d74009ea71f0bc819250afadc1a/pkg/sources/github/github.go#L622) now effectively dead and/or duplicate code? In what circumstances would it be called over `ChunkUnit`?

I have similar confusion with the Git source (https://github.com/trufflesecurity/trufflehog/issues/3005#issuecomment-2185344148), which seems to swallow errors and not log them (at least for the OSS CLI).

mcastorina:
We're sort of in a limbo for sources currently. I'd like to move all sources to implement `SourceUnitEnumChunker` instead of just a single `Chunks` method that does both enumeration and scanning, but right now we need both.

The main (unfortunate) reason, is that enterprise is still using the `Chunks` method, but we're working on transitioning to the new methods.

OSS does not need the `Chunks` method anymore, which is why it always uses the new methods.

rgmz:
> The main (unfortunate) reason, is that enterprise is still using the Chunks method, but we're working on transitioning to the new methods.

How do you feel about marking `scan` (and equivalent methods) to clarity that `Chunks` is the newer/preferred approach?

PR #3308, opened 2024-09-19
Endpoint customizer refresh

kashifkhan0771:
### Description:
Updated Endpoint Customizer and Detectors using it.

This change separates configuration for using found URLs in the data and using known cloud URLs, while centralizing the logic for detectors to use in the `EndpointSetter` struct. The driving force for this change is that we have some URLs known ahead of time and some URLs known only at detection time, and we would like to configure both.

The embeddable `EndpointSetter` struct has a simplified interface: A detector calls `s.Endpoints(foundURLs...)` to get a slice of the endpoints to perform verification against.

A `CloudProvider` interface is also introduced in this PR for a detector to signal there is a cloud endpoint associated with it.

### Checklist:

### DIFF
--- a/pkg/detectors/artifactory/artifactory.go
+++ b/pkg/detectors/artifactory/artifactory.go
@@ -15,11 +15,13 @@ import (
 type Scanner struct {
 	client *http.Client
 	detectors.DefaultMultiPartCredentialProvider
+	detectors.EndpointSetter
 }
 
 var (
 	// Ensure the Scanner satisfies the interface at compile time.
-	_ detectors.Detector = (*Scanner)(nil)
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
 
 	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
 
@@ -52,6 +54,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		if len(URLmatch) != 2 {
 			continue
 		}
+
 		resURLMatch = strings.TrimSpace(URLmatch[1])
 	}
 
@@ -61,20 +64,24 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 		resMatch := strings.TrimSpace(match[1])
 
-		s1 := detectors.Result{
-			DetectorType: detectorspb.DetectorType_ArtifactoryAccessToken,
-			Raw:          []byte(resMatch),
-			RawV2:        []byte(resMatch + resURLMatch),
+		client := s.getClient()
+
+		for _, URL := range s.Endpoints(resURLMatch) {
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_ArtifactoryAccessToken,
+				Raw:          []byte(resMatch),
+				RawV2:        []byte(resMatch + URL),
+			}
+
+			if verify {
+				isVerified, verificationErr := verifyArtifactory(ctx, client, URL, resMatch)
+				s1.Verified = isVerified
+				s1.SetVerificationError(verificationErr, resMatch)
+			}
+
+			results = append(results, s1)
 		}
 
-		if verify {
-			client := s.getClient()
-			isVerified, verificationErr := verifyArtifactory(ctx, client, resURLMatch, resMatch)
-			s1.Verified = isVerified
-			s1.SetVerificationError(verificationErr, resMatch)
-		}
-
-		results = append(results, s1)
 	}
 
 	return results, nil
--- a/pkg/detectors/endpoint_customizer.go
+++ b/pkg/detectors/endpoint_customizer.go
@@ -42,7 +42,7 @@ func (e *EndpointSetter) UseFoundEndpoints(enabled bool) {
 
 func (e *EndpointSetter) Endpoints(foundEndpoints ...string) []string {
 	endpoints := e.configuredEndpoints
-	if e.useCloudEndpoint {
+	if e.useCloudEndpoint && e.cloudEndpoint != "" {
 		endpoints = append(endpoints, e.cloudEndpoint)
 	}
 	if e.useFoundEndpoints {
--- a/pkg/detectors/endpoint_customizer.go
+++ b/pkg/detectors/endpoint_customizer.go
@@ -28,12 +28,16 @@ func (e *EndpointSetter) SetConfiguredEndpoints(userConfiguredEndpoints ...strin
 	return nil
 }
 
+func (e *EndpointSetter) SetCloudEndpoint(url string) {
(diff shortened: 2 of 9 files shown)


pkg/detectors/artifactory/artifactory.go
```diff
@@ -61,20 +64,24 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				s1.Verified = isVerified
+				s1.SetVerificationError(verificationErr, resMatch)
+			}
 
-		if verify {
-			client := s.getClient()
-			isVerified, verificationErr := verifyArtifactory(ctx, client, resURLMatch, resMatch)
-			s1.Verified = isVerified
-			s1.SetVerificationError(verificationErr, resMatch)
+			results = append(results, s1)
```
rgmz, line 82: Shouldn't a verified token break the loop? A token can't be valid on more than one instance, AFAIK.

kashifkhan0771: The idea is to capture the result of all URLs we verify against even if the verification passes or not.

abmussani: @rgmz I believe these are access token. If yes, then Form jfrog artifcatory [documentation](https://jfrog.com/help/r/jfrog-platform-administration-documentation/access-tokens), 

>  
> Cross-instance authentication
> 
> Access tokens can be used for authentication, not only by the instance or cluster where they were created but also for other instances and clusters that are all part of the same circle of trust.

rgmz: Interesting. Just be careful: Artifactory cloud has aggressive rate limiting. 

More than a few failed auth in a minute and you won't be able to hit it anymore, in my experience.

kashifkhan0771: @rgmz If Artifactory has any rate limiting, I believe that should be on the same endpoint. Here the endpoint will be changing for each request unless user pass same endpoint multiple times.

rosecodym: @kashifkhan0771 if an endpoint is found multiple times in a chunk, won't we query it once per occurrence? That could cause rate-limiting, right? It's unlikely but theoretically possible. (I don't know whether it's worth worrying about it.)

pkg/detectors/github/v1/github_old.go
```diff
@@ -19,9 +19,10 @@ type Scanner struct{ detectors.EndpointSetter }
 var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.Versioner = (*Scanner)(nil)
 var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+var _ detectors.CloudProvider = (*Scanner)(nil)
 
-func (Scanner) Version() int            { return 1 }
-func (Scanner) DefaultEndpoint() string { return "https://api.github.com" }
+func (Scanner) Version() int          { return 1 }
+func (Scanner) CloudEndpoint() string { return "https://api.github.com" }
```
zricethezav, line 25: **Question**: Is there a reason we're changing `DefaultEndpoint` to `CloudEndpoint`? I feel like `DefaultEndpoint` is much easier to wrap my head around.

rosecodym (approved):

pkg/detectors/endpoint_customizer.go
```diff
@@ -10,27 +10,43 @@ import (
...
+
+func (e *EndpointSetter) UseFoundEndpoints(enabled bool) {
+	e.useFoundEndpoints = enabled
+}
+
+func (e *EndpointSetter) Endpoints(foundEndpoints ...string) []string {
+	endpoints := e.configuredEndpoints
+	if e.useCloudEndpoint {
+		endpoints = append(endpoints, e.cloudEndpoint)
+	}
```
mcastorina, line 47: Just thought of something: we should probably change this to only append if `e.cloudEndpoint != ""`:

```go
if e.useCloudEndpoint && e.cloudEndpoint != "" {
    endpoints = append(endpoints, e.cloudEndpoint)
}
```

pkg/detectors/artifactory/artifactory.go
kashifkhan0771: @rosecodym good point! In that case I think it's good to add a check to verify that the endpoints should be unique only.

pkg/detectors/github/v1/github_old.go
kashifkhan0771: As per @mcastorina `EndpointCustomizer` was created for detectors that have cloud and on-prem endpoints. So we renamed it accordingly.

pkg/detectors/artifactory/artifactory.go
kashifkhan0771: @rosecodym As per our logic for match URLs:
```
resURLMatch := ""
	for _, URLmatch := range URLmatches {
		if len(URLmatch) != 2 {
			continue
		}

		resURLMatch = strings.TrimSpace(URLmatch[1])
	}
```
We only pick one URL from the matches. I think a duplicate endpoint can only occur if user provide same endpoint multiple times. 
Shall we add a unique check for that or we can ignore it as that should be handled from user end?

pkg/detectors/endpoint_customizer.go
kashifkhan0771: Done @mcastorina

pkg/detectors/github/v1/github_old.go
mcastorina: `DefaultEndpoint` felt ambiguous and was really only defining a default *static* endpoint. Would `DefaultCloudEndpoint()` be a better name?

zricethezav: `CloudEndpoint` is fine, thanks for the explanation

pkg/detectors/artifactory/artifactory.go
```diff
@@ -52,6 +54,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		if len(URLmatch) != 2 {
 			continue
 		}
+
 		resURLMatch = strings.TrimSpace(URLmatch[1])
```
zricethezav, line 58: On a separate note, this detector appears to ignore all other URL matches other than the last one matched. https://github.com/trufflesecurity/trufflehog/blob/b2311b4ad2169f5d6c24e3ad136ce698ecfab042/pkg/detectors/mite/mite.go#L52 is how we should be handling this; nested loop w/ matches and urlMatches. Please correct me if I'm off basis here @mcastorina

mcastorina: That's correct - we noticed that oddity when we were updating this detector but decided it was out of scope for this PR.

zricethezav: Gotcha, in that case I think this is okay to merge

PR #3333, opened 2024-09-25
[fix] Correctly initialize detectors with cloud endpoint customization

mcastorina:
### Description:
We were only initializing if the detector was configured with a custom endpoint, but not in the default case.

### Checklist:


trufflesteeeve (approved):
LGTM

kashifkhan0771:
Thanks @mcastorina!

PR #3337, opened 2024-09-26
[feat] - Add Generic Hasher Interface with Blake2b Implementation

ahrav:
### Description:
This PR includes the following:
- Introduced a `Hasher` interface to abstract hashing functionality.
- Implemented `FNVHasher` using the FNV-64a algorithm.
- Implemented `SHA256Hasher` using the SHA-256 algorithm.

### Bechmarks
Screenshot 2024-09-26 at 8 16 35 AM

### Checklist:


dustin-decker:
May I suggest using blake2b? I've added it to this PR. It's a fast cryptographic hash function that is only ~4x slower than FNV

ahrav:
> May I suggest using blake2b? I've added it to this PR. It's a fast cryptographic hash function that is only ~4x slower than FNV when using the default 256 bit settings.

Oh whoops, I might've forced pushed over your change 🤦

dustin-decker:
> > May I suggest using blake2b? I've added it to this PR. It's a fast cryptographic hash function that is only ~4x slower than FNV when using the default 256 bit settings.
> 
> Oh whoops, I might've forced pushed over your change 🤦

re-added

pkg/hasher/hasher.go
```diff
@@ -0,0 +1,54 @@
...
+// that previous hashing operations do not affect the result.
+func (b *baseHasher) Hash(data []byte) ([]byte, error) {
+	if len(data) > maxInputSize {
+		return nil, &InputTooLargeError{inputSize: len(data), maxSize: maxInputSize}
+	}
+	b.hash.Reset()
+	// nolint:errcheck
+	// The hash.Hash interface does not return errors on Write.
+	// (https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/hash/hash.go;l=27-28)
+	_, _ = b.hash.Write(data)
```
mcastorina, line 52: Should we check the error anyway? 🙃

pkg/hasher/hasher.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (e *InputTooLargeError) Error() string {
+	return fmt.Sprintf("input data exceeds the maximum allowed size: %d > %d", e.inputSize, e.maxSize)
+}
+
+const maxInputSize = 1 << 14 // 16KB
+
+// Hash computes the hash of the given data.
+// It returns an ErrEmptyData if the input is empty or if writing to the hash fails.
```
mcastorina, line 41: Did you end up removing `ErrEmptyData` from the implementation?

pkg/hasher/mutex.go
```diff
@@ -0,0 +1,24 @@
+package hasher
+
+import (
+	"sync"
+)
+
+// MutexHasher wraps a Hasher with a sync.Mutex to ensure thread-safe access.
```
mcastorina, line 7: Are `Hasher` objects not thread-safe? That seems unexpected.

Is there a situation where we don't want them to be thread-safe?

mcastorina (commented):
overall this lgtm; love the extensive testing and benchmarks! just had one question about thread safety.

pkg/hasher/mutex.go
ahrav: Correct, Hasher objects are generally not thread-safe.

There are situations where we don't want Hasher objects to be thread-safe, particularly In our case, each worker would have its own instance of a Hasher, ensuring no race conditions, and allowing us to bypass mutexes entirely. By doing this, we avoid the overhead of locking mechanisms, resulting in a ~20x speedup. (There should be benchmarks attached to the description)

I added a test (`TestMutexHasherConcurrentHash`) that verifies the thread safety of `MutexHasher`. If the mutex logic in `MutexHasher` were commented out or removed, the test would fail due to concurrency issues, and running the test with the `-race` flag would also detect data races.

pkg/hasher/hasher.go
ahrav: 🤣 
Never a bad thing to be extra sure.

pkg/hasher/hasher.go
ahrav: I agree with your approach that this package shouldn't be responsible for it. Now, I'm unsure if I should also remove the max size check. 🤣

pkg/hasher/mutex.go
mcastorina: Followup question: are we actually using the `MutexHasher` anywhere (or planning to)?

Separate idea: I imagine it's probably slower, but have you tested a hash implementation that is inherently thread safe by always generating a new underlying hash object? For example:

```go
// Blaker2bHasher implements the Hasher interface using Blake2b algorithm.
type Blaker2bHasher struct{}

// NewBlaker2bHasher creates a new Blaker2bHasher.
func NewBlaker2bHasher() Blaker2bHasher {
	return Blaker2bHasher{}
}

func (Blaker2bHasher) Hash(data []byte) ([]byte, error) {
        h, _ := blake2b.New256(nil)
        _, _ = h.Write(data)
        return h.Sum(nil), nil
}
```

ahrav: Yea, I had started with that, but after I ran benchmarks comparing the two approaches:
The thread-safe version (creating a new hash object each time) is indeed about 5x slower and causes an allocation on every call, versus a single allocation for the reusable version.

mcastorina (approved):
Just an out of date comment and possibly useless `MutexHasher` struct, but lgtm!

dustin-decker (approved):
LGTM. Should we have anything other than Blake2b for what we'll be using this for though?

ahrav:
> LGTM. Should we have anything other than Blake2b for what we'll be using this for though?

I was thinking we'd use the SHA256 hasher for enterprise, but I supposed we could implement the interface outside of OSS.

dustin-decker:
> > LGTM. Should we have anything other than Blake2b for what we'll be using this for though?
> 
> I was thinking we'd use the SHA256 hasher for enterprise, but I supposed we could implement the interface outside of OSS.

No, it's a different alg that we'd implement in enterprise.

dustin-decker (approved):

PR #3339, opened 2024-09-26
Improve process cleanup

dustin-decker:
### Description:
Makes process cleanup more robust for gitparse, and adds missing process cleanup to the git package.
The git package was leaving zombie processes and the gitparse one could under some failure scenarios.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -1241,10 +1241,12 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 	}
 
 	cmd := exec.Command("git", "-C", gitDir, "cat-file", "blob", commitHash.String()+":"+path)
-	stdout, err := s.executeCatFileCmd(cmd)
+	stdout, catCmd, err := s.executeCatFileCmd(cmd)
 	if err != nil {
 		return err
 	}
+	// Wait must be called after closing the pipe (defer is a stack, so first defer is executed last)
+	defer catCmd.Wait()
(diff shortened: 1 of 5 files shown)


ahrav (approved):

pkg/sources/git/git.go
```diff
@@ -1241,32 +1241,44 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
+	defer func() {
+		_ = catCmd.Wait()
+	}()
+	defer stdout.Close()
+
+	err = handlers.HandleFile(ctx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
 
-	if err = handlers.HandleFile(ctx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives)); err != nil {
+	// Always call Wait() to ensure the process is properly cleaned up
+	waitErr := cmd.Wait()
```
rgmz, line 1257: This seems like a bug. `cmd` and `catCmd` are the same thing. It seems that `cmd.Wait()` won't exit until `stdout.Close()` is called, but that won't run into `cmd.Wait()` is done.

See: #3374

dustin-decker: Oops! Thank you for digging into this.

dustin-decker: Doesn't really seem like there is benefit in executeCatFileCmd being a separate function IMO, it's only used once anyway.

PR #3341, opened 2024-09-26
[fix] Move detector initialization to DefaultDetectors function

mcastorina:
### Description:
Moves initialization out on `NewEngine` and into `DefaultDetectors` so all calls to it will be given a list of detectors that are ready to use.

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -267,21 +267,6 @@ func NewEngine(ctx context.Context, cfg *Config) (*Engine, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	// Abuse filters to initialize endpoint customizer detectors.
-	filters = append(filters, func(d detectors.Detector) bool {
-		customizer, ok := d.(detectors.EndpointCustomizer)
-		if !ok {
-			return true
-		}
-		customizer.UseFoundEndpoints(true)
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -293,20 +278,15 @@ func NewEngine(ctx context.Context, cfg *Config) (*Engine, error) {
 				return false
 			}
 
-			if !cfg.CustomVerifiersOnly || len(urls) == 0 {
-				customizer.UseFoundEndpoints(true)
-				customizer.UseCloudEndpoint(true)
+			if cfg.CustomVerifiersOnly && len(urls) > 0 {
+				customizer.UseCloudEndpoint(false)
+				customizer.UseFoundEndpoints(false)
```
dustin-decker, line 283: Are we sure we want found endpoints off here?

dustin-decker (approved):

pkg/engine/engine.go
mcastorina: That's a good question. We currently don't have controls to separate the configured cloud endpoint vs found endpoints though, so I defaulted to tying them together.

If a user specifies `--custom-verifiers-only` imo that means "only use the endpoints I specifically configured"

PR #3344, opened 2024-09-27
[feat] - Add SizedLRU Cache

ahrav:
### Description:
This PR introduces a size-limited, generic LRU cache with optional metrics tracking, which wraps the golang-lru caching library. Key features include customizable metrics collection (e.g., hits, misses, evictions) and configurable behavior through functional options like setting the maximum cache size.

### Checklist:

### DIFF
--- a/pkg/cache/metrics.go
+++ b/pkg/cache/metrics.go
@@ -1,0 +1,107 @@
+package cache
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// BaseMetricsCollector defines the interface for recording cache metrics.
+// Each method corresponds to a specific cache-related operation.
+type BaseMetricsCollector interface {
+	RecordHit(cacheName string)
+	RecordMiss(cacheName string)
+	RecordSet(cacheName string)
+	RecordDelete(cacheName string)
+	RecordClear(cacheName string)
+}
+
+// MetricsCollector encapsulates all Prometheus metrics with labels.
+// It holds Prometheus counters for cache operations, which help track
+// the performance and usage of the cache.
+type MetricsCollector struct {
+	// Base metrics.
+	hits    *prometheus.CounterVec
+	misses  *prometheus.CounterVec
+	sets    *prometheus.CounterVec
+	deletes *prometheus.CounterVec
+	clears  *prometheus.CounterVec
+}
+
+var (
+	collectorOnce sync.Once // Ensures that the collector is initialized only once.
+	collector     *MetricsCollector
+)
+
+// InitializeMetrics initializes the singleton MetricsCollector.
+// It sets up Prometheus counters for cache operations (hits, misses, sets, deletes, clears).
+// Should be called once at the start of the application.
+func InitializeMetrics(namespace, subsystem string) {
+	collectorOnce.Do(func() {
+		collector = &MetricsCollector{
+			hits: promauto.NewCounterVec(prometheus.CounterOpts{
+				Namespace: namespace,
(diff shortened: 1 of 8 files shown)


rgmz:
Is this meant to replace the existing memory cache? If not, what are the use cases of this vs that?

ahrav:
> Is this meant to replace the existing memory cache? If not, what are the use cases of this vs that?

I don't believe so. The plan is to use this for detection caching.

rgmz:
> I don't believe so. The plan is to use this for detection caching.

It might be worth tweaking the naming + comments in that case. "Memory" and "LRU" aren't exactly mutually exclusive and don't have clear purposes.

pkg/cache/sizedlru/sizedlru.go
```diff
@@ -0,0 +1,120 @@
...
+func WithCapacity[T any](capacity int) Option[T] {
+	return func(lc *Cache[T]) { lc.capacity = capacity }
+}
+
+// NewCache creates a new Cache with optional configuration parameters.
+// It takes a cache name and a variadic list of options.
+func NewCache[T any](cacheName string, opts ...Option[T]) (*Cache[T], error) {
+	// Default values for cache configuration.
+	const defaultSize = 1 << 29 // 512MB
+
```
dustin-decker, line 56: I think a smaller default (like 32MB) would be sensible here.

pkg/cache/sizedlru/sizedlru.go
```diff
@@ -0,0 +1,120 @@
...
+// It sets the metrics field of the Cache.
+func WithMetricsCollector[T any](collector collector) Option[T] {
+	return func(lc *Cache[T]) { lc.metrics = collector }
+}
+
+// WithCapacity is a functional option to set the maximum capacity of the cache.
+// If the capacity is not set, the default value (512MB) is used.
+func WithCapacity[T any](capacity int) Option[T] {
+	return func(lc *Cache[T]) { lc.capacity = capacity }
+}
```
dustin-decker, line 49: Can you document the units expected here?

pkg/cache/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	collectorOnce sync.Once // Ensures that the collector is initialized only once.
+	collector     *MetricsCollector
+)
+
+// InitializeMetrics initializes the singleton MetricsCollector.
+// It sets up Prometheus counters for cache operations (hits, misses, sets, deletes, clears).
+// Should be called once at the start of the application.
+func InitializeMetrics(namespace, subsystem string) {
+	collectorOnce.Do(func() {
+		collector = &MetricsCollector{
```
dustin-decker, line 42: Consider having this called automatically when the package is imported using init()?

dustin-decker (approved):

dustin-decker (approved):
LGTM. Love the metric instrumentation!

pkg/cache/sizedlru/sizedlru.go
ahrav: Yep, great idea.

pkg/cache/metrics.go
ahrav: Yea, that makes sense. Good call.

ahrav: Also cleans up some of the test code.

dustin-decker: Another option is to just declar them as package vars like we do with most others

pkg/cache/sizedlru/sizedlru.go
ahrav: Whoops, I got the cache implementations mixed up. This capacity refers to the number of items, not total size.

PR #3354, opened 2024-10-01
Added Pattern test cases for detectors

kashifkhan0771:
### Description:
This PR adds pattern test cases for some detectors. It also fix some regex of detectors.

To Test all Pattern Test Cases, run following command:
```
go test ./pkg/detectors/... -run "^*Pattern$"
```
**Output:**
pattern_tests

### Checklist:

### DIFF
--- a/pkg/detectors/agora/agora_test.go
+++ b/pkg/detectors/agora/agora_test.go
@@ -1,21 +1,13 @@
-//go:build detectors
-// +build detectors
-
 package agora
 
 import (
 	"context"
-	"fmt"
 	"testing"
-	"time"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 
-	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/engine/ahocorasick"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
 func TestAgora_Pattern(t *testing.T) {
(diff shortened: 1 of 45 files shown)


pkg/detectors/agora/agora_test.go
```diff
@@ -11,12 +11,92 @@ import (
...
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/engine/ahocorasick"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
+func TestAgora_Pattern(t *testing.T) {
```
kashifkhan0771, line 21: **Question:** Agora detector key and secret have same pattern and because of that it actually picks same string twice. Is it expected or we need to add some logic to handle this. You can see in test cases that the want is actually a duplicate string.

kashifkhan0771:
`adafruitio` detector is not letting me add a Pattern test cases. It says cannot push secrets 🥲. Any ideas?

rgmz:
@kashifkhan0771 I'd recommend using #3274 as a base. Right now these tests don't run unless the -detectors flag is specified.

kashifkhan0771:
> @kashifkhan0771 I'd recommend using #3274 as a base. Right now these tests don't run unless the -detectors flag is specified.

Thanks @rgmz - That is a good approach. I'll rebase and structure these unit test cases accordingly.

kashifkhan0771:
@rgmz I made the changes according to your PR but I kept the base branch same.

rgmz:
Yes, sorry I meant "based on" not "as the PR merge base".

pkg/detectors/alegra/alegra.go
```diff
@@ -24,7 +24,7 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-z0-9-]{20})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9.-@]{25,30})\b`)
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9\.\-\@]{25,30})\b`)
```
kashifkhan0771, line 27: The old regex was not accepting special character `-` because we were not escaping special characters. Fixed it and added a test case for it as well

pkg/detectors/agora/agora_test.go
zricethezav: @kashifkhan0771 great catch. Because having a `secret` _and_ `key` as the same string _should_ be impossible, we should include some logic to detect if `secret==key` and skip that case.

ahrav (approved):
lgtm 👍 Looks great!

pkg/detectors/alegra/alegra.go
rgmz: It's typically safest to put `-` at the end for this reason. It's an insidious bug.
https://web.archive.org/web/20240525200717/https://pboyd.io/posts/comma-dash-dot/

PR #3355, opened 2024-10-01
[refactor] - Decouple Metrics From Cache Implementation

ahrav:
### Description:
This PR attempts to enhance the clarity and separation of concerns in metrics collection by segregating base cache metrics from eviction-specific metrics and externalizing metrics configuration.
### Checklist:


pkg/cache/metrics.go
```diff
@@ -32,10 +37,10 @@ type MetricsCollector struct {
 }
 
 func init() {
-	// Initialize the singleton MetricsCollector.
+	// Initialize the singleton BaseCollector.
 	// Set up Prometheus counters for cache operations (hits, misses, sets, deletes, clears).
-	collectorOnce.Do(func() {
-		collector = &MetricsCollector{
+	baseCollectorOnce.Do(func() {
```
mcastorina, line 42: Why does this need to be in a `once.Do` if it's getting called from `init()`? Are there other places calling this to initialize it?

pkg/cache/metrics.go
```diff
@@ -72,37 +77,69 @@ func init() {
...
-	return collector
+	return baseCollector
+}
+
+// GetEvictionMetricsCollector returns the singleton EvictionMetrics instance.
+func GetEvictionMetricsCollector() *EvictionMetrics {
+	if evictionCollector == nil {
+		panic("EvictionMetrics not initialized. Call InitializeMetrics first.")
+	}
+	return evictionCollector
```
mcastorina, line 116: What's the advantage of having `evictionCollector` be private if there's a public getter function for it?

pkg/cache/metrics.go
ahrav: ooo, I think I forgot to get rid of the sync.Once after Dustin recommended using init().

mcastorina (approved):

PR #3358, opened 2024-10-02
[feat] - Cache Credentials to Avoid Reverification

ahrav:
### Description:
This PR enhances the engine by introducing credential caching to avoid re-verifying the same credential multiple times. We hash the credential data to generate cache keys, optimizing performance by skipping redundant verification steps. This reduces scan times, minimizes external API calls, and mitigates issues like account lockouts due to excessive verification attempts.

NOTE: I'm still figuring out the verificationErr bit.

**Key Features:**
- Configurable detection cache with an option to inject a custom cache and hasher.
- Caches verification results to avoid duplicate checks.

### Some Metrics
Screenshot 2024-09-29 at 1 12 32 PM

### Checklist:


pkg/engine/engine.go
```diff
@@ -468,14 +560,20 @@ func filterDetectors(filterFunc func(detectors.Detector) bool, input []detectors
...
-// initialize prepares the engine's internal structures. The LRU cache optimizes
+// initialize prepares the engine's internal structures. The LRU detectionCache optimizes
 // deduplication efforts, allowing the engine to quickly check if a chunk has
 // been processed before, thereby saving computational overhead.
 func (e *Engine) initialize(ctx context.Context) error {
 	// TODO (ahrav): Determine the optimal cache size.
-	const cacheSize = 512 // number of entries in the LRU cache
+	const (
+		cacheName = "dedupe_cache"
+		cacheSize = 512 // number of entries in the LRU cache
```
rgmz, line 570: I suspect the optimal size is at least 4x this.

pkg/engine/engine.go
```diff
@@ -955,6 +1042,53 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 						val = res.Raw
 					}
 
+					// Only check the cache for non-targeted scans.
+					// Targeted scans should always be verified.
+					if e.enableDetectionCache && !isTargetedScan {
+						// Check the cache to see if we've already seen this secret.
+						keyBytes := make([]byte, len(res.Raw)+len(res.RawV2))
+						copy(keyBytes, res.Raw)
+						copy(keyBytes[len(res.Raw):], res.RawV2)
```
rgmz, line 1051: Is it necessary to use both `Raw` and `RawV2`? Isn't the former redundant of V2 is defined?

pkg/engine/engine.go
```diff
@@ -1094,6 +1228,30 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
 		}
 
 		for _, res := range results {
+			if e.enableDetectionCache {
+				// Cache the result to avoid re-verifying the same secret.
+				hashKeyBytes := make([]byte, len(res.Raw)+len(res.RawV2))
+				copy(hashKeyBytes, res.Raw)
+				copy(hashKeyBytes[len(res.Raw):], res.RawV2)
```
rgmz, line 1235: Same question as above in `verificationOverlapWorker` (which should be removed IMO...)

pkg/engine/engine.go
```diff
@@ -1197,7 +1355,7 @@ func (e *Engine) notifierWorker(ctx context.Context) {
 			result.SourceType == sourcespb.SourceType_SOURCE_TYPE_POSTMAN) {
 			continue
 		}
-		e.dedupeCache.Add(key, result.DecoderType)
+		e.dedupeCache.Set(key, result.DecoderType)
```
rgmz, line 1358: How does the cache handle indeterminate errors? 
- status code 429
- i/o timeout
- etc.

rgmz (commented):
This is a good improvement, though I think there's still the issue of deduplicating results within a single chunk.

pkg/engine/engine.go
ahrav: I agree, but we can address this in a separate PR to keep the focus of this one on the detection cache.

pkg/engine/engine.go
ahrav: Thanks for pointing this out. The key should actually be `hash(Raw) + hash(RawV2)`. This is due to an unfortunate decision we made with the enterprise solution, where we store both `Hash` and `HashV2` (with `HashV2` being the hash of an empty string if `RawV2` doesn't exist). 😞  For that reason, I believe we need to use both `Raw` and `RawV2` in the key.

I'll have a few others weigh in to double-check if my understanding is correct.

rgmz:
Also, thoughts on leveraging [`golang.org/x/sync/singleflight`](https://www.codingexplorations.com/blog/understanding-singleflight-in-golang-a-solution-for-eliminating-redundant-work) to prevent concurrent requests for the same secret?

ahrav:
> Also, thoughts on leveraging [`golang.org/x/sync/singleflight`](https://www.codingexplorations.com/blog/understanding-singleflight-in-golang-a-solution-for-eliminating-redundant-work) to prevent concurrent requests for the same secret?

Good point. I actually have a draft from a while back that uses that package. Funny enough, I think I just read about it in the Go blog this week or last. 😅

pkg/engine/engine.go
rgmz: After thinking about this for a bit, I worry the implementation isn't intuitive and/or is hard to follow. I can't quite put my finger on it... the change itself isn't bad by any means.

I expected most of the logic to be in `detectChunk` but a lot of the heavy lifting is done in `verificationOverlapWorker` (in general). Those functions share a decent amount of duplicate code (e.g., why do we need to call `processResult` in both?), yet the relevant bits are close to 200 lines a part.

This is probably an issue greater than the scope of this PR.

pkg/engine/engine.go
rgmz: I'm reading this on my phone so it's possible I missed it, but shouldn't the hash key contain the detector type ID?

pkg/engine/engine.go
```diff
@@ -955,6 +1042,53 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
...
+						cacheVal, exists := e.detectionCache.Get(string(keyHash))
+						if exists {
+							// Update the "res" object with the cached result.
+							res.Verified = cacheVal.Verified
+							res.DecoderType = cacheVal.DecoderType
+							res.DetectorType = cacheVal.DetectorType
+							res.ExtraData = cacheVal.ExtraData
+							res.Redacted = cacheVal.Redacted
+							res.StructuredData = cacheVal.StructuredData
+							res.AnalysisInfo = cacheVal.AnalysisInfo
```
mcastorina, line 1067: This seems error prone. What about storing the `detector.Result` directly instead of a new type?

mcastorina (commented):
This looks like a good first step, but my main concern is the maintainability of this feature (noted in-line concerning the duplicate type).

pkg/engine/engine.go
```diff
@@ -88,6 +90,32 @@ func (p *PrinterDispatcher) Dispatch(ctx context.Context, result detectors.Resul
...
+
+	// This field should only be populated if the verification process itself failed in a way that provides no
+	// information about the verification status of the candidate secret, such as if the verification request timed out.
+	verificationError error
+
+	// AnalysisInfo should be set with information required for credential
+	// analysis to run. The keys of the map are analyzer specific and
+	// should match what is expected in the corresponding analyzer.
+	AnalysisInfo map[string]string
+}
```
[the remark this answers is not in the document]
ahrav: It is excluding the Raw and RawV2 fields. I thought that by manually using a custom type, we could ensure we don't accidentally store those fields.

pkg/engine/engine.go
ahrav: We could, we'd just have to ensure we remove the Raw and RawV2 fields.

pkg/engine/engine.go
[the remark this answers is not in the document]
mcastorina: I like that the type makes it impossible to store the values, but I don't like that there are two (or three) places that would need to get updated if we change the `Result`. I can't even imagine the weird behaviors we would get if we only cache a subset of the metadata by accident.

pkg/engine/engine.go
```diff
@@ -355,7 +423,31 @@ func (e *Engine) setDefaults(ctx context.Context) {
 	e.notifyUnverifiedResults = true
 	e.notifyUnknownResults = true
 
+	// Set enableCache to true by default if not explicitly set.
+	if !e.enableDetectionCache {
+		e.enableDetectionCache = true
+	}
```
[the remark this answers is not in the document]
ahrav: Whoops. Yep, that's wrong.

pkg/engine/engine.go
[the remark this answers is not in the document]
ahrav: I'll use the Result in that case.

pkg/engine/engine.go
```diff
@@ -1057,15 +1026,9 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context, hasher hasher.Ha
...
-							res.DecoderType = cacheVal.DecoderType
-							res.DetectorType = cacheVal.DetectorType
-							res.ExtraData = cacheVal.ExtraData
-							res.Redacted = cacheVal.Redacted
-							res.StructuredData = cacheVal.StructuredData
-							res.AnalysisInfo = cacheVal.AnalysisInfo
-							// TODO: figure out how to set verification error.
+							// Update the cached result with the Raw and RawV2 values.
+							cacheVal.Raw = res.Raw
+							cacheVal.RawV2 = res.RawV2
```
mcastorina, line 1031: `detectionCache` is storing pointers, right? If that's the case, then I think that means setting values on `cacheVal` will result in them being available from the cache.

Do you think that storing `detectors.Result` in the cache instead of `*detectors.Result` would be a viable solution?

ahrav: I initially considered that, but opted for the pointer to reduce the size from 136B to 8B per value. Do you think the increased memory footprint is okay?

mcastorina: Hmm.. I don't know! Do larger values affect the performance? In terms of overall memory usage I would expect it to be roughly the same, since the pointers need to point to actual data.

ahrav: I thought that using a value instead of a pointer would double memory usage by creating a copy of the struct, and that the extra memory would persist until the gc cleans up the unused struct. You are thinking the increased memory in that case is negligible because the original Result will only stay in memory briefly before being gc'd?

mcastorina: Ah, I see. Yeah that makes sense that a copy of the data would be stored in the cache. I hadn't thought about the original being heap allocated.

pkg/engine/engine.go
rgmz: @ahrav Moreover, inserting `verificationOverlapWorker` has completely broken the `--allow-verification-overlap` flag:
```sh
$ ./trufflehog filesystem /tmp/overlap --allow-verification-overlap

Found unverified result 🐷🔑❓
Verification issue: More than one detector has found this result. For your safety, verification has been disabled.You can override this behavior by using the --allow-verification-overlap flag.
...
```

Because of: 
```diff
-		if len(matchingDetectors) > 1 && !e.verificationOverlap {
				wgVerificationOverlap.Add(1)
				e.verificationOverlapChunksChan <- verificationOverlapChunk{
					chunk:                       *decoded.Chunk,
					detectors:                   matchingDetectors,
					decoder:                     decoded.DecoderType,
					verificationOverlapWgDoneFn: wgVerificationOverlap.Done,
				}
```

pkg/engine/engine.go
rgmz: The answer is yes: if a secret overlaps between detectors, it can incorrectly report the wrong result.

**Input**
```
snyk = 6b3ae999-3924-486a-9d3b-32315bf0f76b
coda = 6b3ae999-3924-486a-9d3b-32315bf0f76b
```

**Output**
(Note: I have modified the code to always mark the Coda result as verified and delay handling of the Snyk result so that it uses the cached value.) 
```
✅ Found verified result 🐷🔑
Detector Type: Coda
Decoder Type: PLAIN
Raw result: 6b3ae999-3924-486a-9d3b-32315bf0f76b
File: /tmp/overlap

✅ Found verified result 🐷🔑
Detector Type: Coda
Decoder Type: PLAIN
Raw result: 6b3ae999-3924-486a-9d3b-32315bf0f76b
File: /tmp/overlap
```

ahrav:
> > Configurable detection cache with an option to inject a custom cache and hasher.
> 
> Why is this necessary?

When initializing the engine for enterprise, we plan to allow the use of a different hasher, like SHA-256, for hashing raw results since that's what we currently do on the enterprise side prior to reporting the results to the API. Additionally, we'd support a pre-loaded cache for secrets that we may not want to verify.

PR #3360, opened 2024-10-03
improved the agora detector

kashifkhan0771:
### Description:
This PR improve the agora detector logic to avoid processing when secret and key matches as both have same regex.
**Pattern Test Cases Output:**
agora-detector-tests

### Checklist:


kashifkhan0771:
**Changes:**

- Updated the logic to avoid processing same secret and key
- Renamed few variables for better readability and understanding
- Updated Pattern test cases accordingly. You can see in the output of test cases that it now process only different key and secret.

kashifkhan0771:
**Question:** Currently, both the key and secret regex patterns include the `agora` prefix, which expects the keyword 'agora' to appear near both the key and secret in the data. This helps reduce false positives by narrowing the search. However, it's more likely that the keyword will be present near either the key or the secret, but not both. Removing the prefix would allow scanning the data more broadly for either the key or secret, but it could also increase false positives, as any 32-character alphanumeric string could match.

I need some suggestions here 😃

rgmz:
My advice: use GitHub search to find real-world examples and use them as inspiration for unit test cases.

bugbaba:
@kashifkhan0771 My advice would be to remove those and run against a good number of projects in the wild. Then analyze the false positives and try to observe if there is any pattern in it.

Personally I spent most of time analyzing the false positives and false negatives :)

This is what I am using personally currently
image

rgmz:
> This is what I am using personally currently

I like `key/token/secret` prefixes. I find that not using the `keyword` as a prefix tends to find more results. The keyword already guarantees the 

e.g., this is a common pattern that doesn't get detected if you only use `detectors.PrefixRegex([]string{"agora"})`. 
```
KEY=abdcefg...
SECRET=abdcefg...
curl "https://api.agora.com/" ...
```

kashifkhan0771:
Thank you @bugbaba @rgmz for suggestions. I updated the keywords and now I can see that it also does not pick duplicate strings because of different keywords. I still kept the check for same key and secret just in case.

zricethezav (approved):

PR #3362, opened 2024-10-03
[Fix] Update Azure Client, Tenant Ids and Client Secret Detection Regexs.

abmussani:
### Description:
This PR includes the correction of few things in Azure Detector. Here are the details:
- Support for multiple formats of Client Secret based on [microsoft Documentation](https://learn.microsoft.com/en-us/purview/sit-defn-client-secret-api-key).
- Tenant Id can be part of oauth token URL and the autogenerated url has microsoftonline.com as domain. Which can be used as keyword for detector (apart from 'azure').
- Simplified the way to pass keywords in client id, tenant id & client secret expressions. For the consistency and readability, I have used `PrefixRegex` for this purpose.

Thanks to @justanothernate & @rgmz, for having a fruitful discussion over [PR](https://github.com/trufflesecurity/trufflehog/pull/3039), Helped me to fixed these issue in azure detectors.

### Checklist:

### DIFF
--- a/pkg/detectors/azure/azure.go
+++ b/pkg/detectors/azure/azure.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"strings"
 
+	regexp "github.com/wasilibs/go-re2"
+
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	regexp "github.com/wasilibs/go-re2"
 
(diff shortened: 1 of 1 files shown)


rgmz:
The link cited in the other discussion is not correct and does not represent Azure Client Secrets AFAIK.

Per my comment, I am only aware of two formats.
https://github.com/trufflesecurity/trufflehog/pull/3039#issuecomment-2376869690

abmussani:
> The link cited in the other discussion is not correct and does not represent Azure Client Secrets AFAIK. It is a generic set of patterns Microsoft uses for detecting client secrets (not Azure-specific ones.)

Correct. The link depict the general client secret format used by Microsoft. But as per my understanding, it will also includes the Azure one. The service I used to test is Microsoft Entra.

> This change does not reliably detect the new format.

My changes successfully detect the new format of the secret. Are you referring to the old format by any chance?

rgmz:
> Correct. The link depict the general client secret format used by Microsoft. But as per my understanding, it will also includes the Azure one. The service I used to test is Microsoft Entra.

It might be a coincidence that one of those patterns also matches the Entra ID secrets. Personally, I don't think it makes sense to introduce several broad patterns (and thus increase false matches + network traffic).

>> This change does not reliably detect the new format.
>
> My changes successfully detect the new format of the secret. Are you referring to the old format by any chance?

Emphasis on *reliably*. The new format has a distinct pattern which can be matched with high confidence; a large number of Entra client IDs and secrets I've found do not contain "azure" or "microsoft" anywhere in the file. 

#2985 is a rather extensive update that fixes most issues with the Azure detector, including test cases.

PR #3367, opened 2024-10-04
Added Cisco Meraki API Key detector

kashifkhan0771:
### Description:
This PR adds a new detector for Cisco Meraki API Keys.
Meraki: https://meraki.cisco.com/
API Documentation: https://developer.cisco.com/meraki/api-v1/introduction/

## Test Cases:
Screenshot from 2024-10-04 15-49-29
## Meraki API Key Detection:
Screenshot from 2024-10-04 15-54-50

### Checklist:


pkg/detectors/meraki/meraki.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	// set the required auth header
+	req.Header.Set("X-Cisco-Meraki-API-Key", match)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer resp.Body.Close()
```
rgmz, line 105: It's ostensibly 'best' to manually drain the response body. https://github.com/trufflesecurity/trufflehog/pull/2086#issue-1974701062

```suggestion
        defer func() {
	  _, _ = io.Copy(io.Discard, res.Body)
	  _ = res.Body.Close()
        }()
```

pkg/detectors/meraki/meraki.go
```diff
@@ -0,0 +1,115 @@
...
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return true, nil
```
rgmz, line 109: Does `api/v1/organizations` return any useful info that could be put into extra data?
https://developer.cisco.com/meraki/api-v1/get-organizations/

kashifkhan0771: It returns the organizations user has access to.
Sample Response:
```
[
    {
        "id": "2930418",
        "name": "My organization",
        "url": "https://dashboard.meraki.com/o/VjjsAd/manage/organization/overview",
        "api": { "enabled": true },
        "licensing": { "model": "co-term" },
        "cloud": {
            "region": {
                "name": "North America",
                "host": {
                    "name": "United States"
                }
            }
        },
        "management": {
            "details": [
                {
                    "name": "MSP ID",
                    "value": "123456"
                }
            ]
        }
    }
]
```

pkg/detectors/meraki/meraki.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// uniqueMatches will hold unique match values and ensure we only process unique matches found in the data string
+	var uniqueMatches = make(map[string]bool)
+
+	for _, match := range apiKey.FindAllStringSubmatch(dataStr, -1) {
+		if len(match) != 2 {
+			continue
+		}
+
+		uniqueMatches[strings.TrimSpace(match[1])] = true
```
rgmz, line 60: This is a strange practice established in earlier detector code. At best, it is completely pointless; at worst, it silently breaks the code if the pattern changes.
https://github.com/trufflesecurity/trufflehog/pull/2264#discussion_r1443382434

TrimSpace also isn't required unless the capture group includes whitespace.
```suggestion
	uniqueMatches[match[1]] = true
```

pkg/detectors/meraki/meraki.go
```diff
@@ -0,0 +1,115 @@
...
+	return "Cisco Meraki is a cloud-managed IT solution that provides networking, security, and device management through an easy-to-use interface." +
+		"Meraki APIs make it possible to rapidly deploy and manage networks at scale, build on a platform of intelligent, cloud-connected IT products, and engage with users in powerful new ways."
+}
+
+// FromData will find and optionally verify Meraki API Key secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	// uniqueMatches will hold unique match values and ensure we only process unique matches found in the data string
+	var uniqueMatches = make(map[string]bool)
```
rgmz, line 53: Not really a big deal, but for sets the convention is to use `map[string]struct{}` instead of `map[string]bool`. This is both because the value of `bool` isn't used and empty structs save allocations.

e.g., https://www.reddit.com/r/golang/comments/15vaxlu/the_set_data_structure_in_go/
```suggestion
	var uniqueMatches = make(map[string]struct{})
```

pkg/detectors/meraki/meraki.go
rgmz: I think it's worthwhile to include the org names (e.g., #2222). Up to you and the team.

kashifkhan0771: That's a good suggestion. @zricethezav @abasit-folio3 thoughts?

pkg/detectors/meraki/meraki.go
kashifkhan0771: TrimSpace can be removed. About the pattern, are you referring to index `1` we use from match?

rgmz: > About the pattern, are you referring to index 1 we use from match?

The pattern and `match[1]` are fine. I'm referring to the `if len(match) != 2` check.

kashifkhan0771: Okay that is now removed 😎

pkg/detectors/meraki/meraki.go
zricethezav: Worthwhile to include org name and id

kashifkhan0771: @rgmz @zricethezav Updated the code. Now we capture the Api response and save orgs id and name in the extra data.

zricethezav (approved):

PR #3351, opened 2024-10-01
[fix] - remove size check

ahrav:
### Description:
This PR removes the sick check.
### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/usage/install/#local-installation))?

PR #3375, opened 2024-10-06
Fix deadlock in `Git.handleBinary`

rgmz:
### Description:
This closes #3339.

`cmd` and `catCmd` are the same reference. It seems that explicitly calling `cmd.Wait()` could cause a deadlock where the deferred `stdout.Close()` was never run. Albeit, this could be merely fixing the 'symptom' and not actually curing the issue.

Edit: hmm, this seems produce the beloved `signal: broken pipe` error when calling `cmd.Wait()`. 
```
2024-10-05T21:16:08-04:00	error	trufflehog	error while waiting.	{"source_manager_worker_id": "VksFa", "unit": "https://github.com/cohere-ai/tokenizer.git", "unit_kind": "repo", "repo": "https://github.com/cohere-ai/tokenizer.git", "commit": "6ead2eb", "path": "tokenizer/tokenizer_go.so", "error": "signal: broken pipe"}
2024-10-05T21:16:08-04:00	error	trufflehog	error handling binary file	{"source_manager_worker_id": "VksFa", "unit": "https://github.com/cohere-ai/tokenizer.git", "unit_kind": "repo", "repo": "https://github.com/cohere-ai/tokenizer.git", "filename": "tokenizer/tokenizer_go.so", "commit": "6ead2eb2b9c67ae9dea31cc859edad88e75ae6dc", "file": "tokenizer/tokenizer_go.so", "error": "signal: broken pipe"}
2024-10-05T21:16:08-04:00	info-0	trufflehog	finished scanning	{"chunks": 2381, "bytes": 11359546, "verified_secrets": 0, "unverified_secrets": 0, "scan_duration": "5.686829476s", "trufflehog_version": "dev"}
```

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -1241,19 +1241,18 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 	}
 
 	cmd := exec.Command("git", "-C", gitDir, "cat-file", "blob", commitHash.String()+":"+path)
-	stdout, catCmd, err := s.executeCatFileCmd(cmd)
+	stdout, err := s.executeCatFileCmd(cmd)
 	if err != nil {
 		return err
 	}
-	// Wait must be called after closing the pipe (defer is a stack, so first defer is executed last)
-	defer func() {
-		_ = catCmd.Wait()
-	}()
-	defer stdout.Close()
 
 	err = handlers.HandleFile(ctx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
 
 	// Always call Wait() to ensure the process is properly cleaned up
+	// Wait must be called after closing the pipe (https://stackoverflow.com/a/40706762).
+	if closeErr := stdout.Close(); closeErr != nil {
+		fileCtx.Logger().Error(closeErr, "error closing stdout for cat-file command")
+	}
 	waitErr := cmd.Wait()
 
 	// If there was an error in HandleFile, return that error
@@ -1265,20 +1264,20 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
 	return waitErr
 }
 
-func (s *Git) executeCatFileCmd(cmd *exec.Cmd) (io.ReadCloser, *exec.Cmd, error) {
+func (s *Git) executeCatFileCmd(cmd *exec.Cmd) (io.ReadCloser, error) {
 	var stderr bytes.Buffer
 	cmd.Stderr = &stderr
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		return nil, nil, fmt.Errorf("error running git cat-file: %w\n%s", err, stderr.Bytes())
+		return nil, fmt.Errorf("error running git cat-file: %w\n%s", err, stderr.Bytes())
 	}
 
 	if err := cmd.Start(); err != nil {
-		return nil, nil, fmt.Errorf("error starting git cat-file: %w\n%s", err, stderr.Bytes())
+		return nil, fmt.Errorf("error starting git cat-file: %w\n%s", err, stderr.Bytes())
 	}
 
-	return stdout, cmd, nil
+	return stdout, nil
 }
 
 func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) error {


ahrav:
I initially chose not to call `stdout.Close()` because I saw the following comment in the documentation for `StdoutPipe`, or at least I think that's why. 🤔 

I'm guessing maybe the command wasn't exiting which is what lead to the rogue processes Dustin might've been trying to address with his PR.
```
// StdoutPipe returns a pipe that will be connected to the command's
// standard output when the command starts.
//
// [Cmd.Wait] will close the pipe after seeing the command exit, so most callers
// need not close the pipe themselves. It is thus incorrect to call Wait
// before all reads from the pipe have completed.
// For the same reason, it is incorrect to call [Cmd.Run] when using StdoutPipe.
// See the example for idiomatic usage.
```

rgmz:
It seems like the process still doesn't exit after several hours, yet running the command manually it completes in seconds.

FWIW, it doesn't show up as Z in `ps` either.

rgmz:
I can reliably reproduce this hang by making `HandleFile` return immediately.
https://github.com/trufflesecurity/trufflehog/blob/40fdf440734931798e79c21e6dd06315aa43f8c4/pkg/handlers/handlers.go#L274-L276

This leads me to suspect that the hang is caused by the Read/Write of `io.ReadCloser`.

dustin-decker (approved):

rgmz:
Pulling this back into draft because I'm not comfortable that this actually fixes the problem.

dustin-decker:
Oh, ignore my approval, looks like we have more to figure out. 

I introduced this regression with https://github.com/trufflesecurity/trufflehog/pull/3339.

rgmz:
> I introduced this regression with #3339.

It looks like #3339 introduced the deadlock, based on a latent issue. I don't think the code was working 100% prior to that.

rgmz:
The deadlock doesn't occur if I revert #3351. I have no idea why.

Did that change introduce a regression or was it covering up a latent issue?

ahrav:
> > I introduced this regression with #3339.
> 
> It looks like #3339 introduced the deadlock, based on a latent issue. I don't think the code was working 100% prior to that.

That's what I thought too 😅 . I’ll let Dustin respond. I vaguely remember mentions of processes not being killed, but I could be wrong.

dustin-decker:
Hmm, it does look like Wait should close all pipes, so I'm not sure #3339 actually introduced an issue, but it's not correct.

I've fixed that up in #3379 and added smoke tests for git and zombie processes remaining after a scan. It worked locally, let's see what happens in CI 🤞

rgmz:
> The issue doesn't occur if I revert #3351. I have no idea why.

I still don't understand why adding the call to `.Size()` in HandleFile prevents the issue from occurring.

It seems like the issue is within the handler or bufferedreader code and that Git.handleBinary is just incidental.

ahrav:
> > The issue doesn't occur if I revert #3351. I have no idea why.
> 
> I still don't understand why adding the call to `.Size()` in HandleFile prevents the issue from occurring.
> 
> It seems like the issue is within the handler or bufferedreader code and that Git.handleBinary is just incidental.

I suspect there's an issue with reading from the stdout pipe and buffering the result. I noticed that reading the entire stdout content at once prevents it from hanging.

Ex:
```
// newFileReader creates a fileReader from an io.Reader, optionally using BufferedFileWriter for certain formats.
func newFileReader(r io.Reader) (fileReader, error) {
	var fReader fileReader

	data, err := io.ReadAll(r)
	if err != nil {
		return fReader, fmt.Errorf("unable to read file: %w", err)
	}

	fReader.BufferedReadSeeker = iobuf.NewBufferedReaderSeeker(bytes.NewReader(data))
..
}
```

I created some tests to see if I could reproduce it but the tests pass event without consuming the entire reader with the snippet above..
```
func TestHandleGitCatFileLargeBlob(t *testing.T) {
	fileName := "largefile.bin"
	fileSize := 50 * 1024 * 1024 // 50 MB

	// Set up a temporary git repository with a large file
	gitDir := setupTempGitRepo(t, fileName, fileSize)
	defer os.RemoveAll(gitDir)

	cmd := exec.Command("git", "-C", gitDir, "rev-parse", "HEAD")
	hashBytes, err := cmd.Output()
	assert.NoError(t, err, "Failed to get commit hash")
	commitHash := strings.TrimSpace(string(hashBytes))

	// Create a pipe to simulate the git cat-file stdout
	cmd = exec.Command("git", "-C", gitDir, "cat-file", "blob", fmt.Sprintf("%s:%s", commitHash, fileName))

	var stderr bytes.Buffer
	cmd.Stderr = &stderr

	stdout, err := cmd.StdoutPipe()
	assert.NoError(t, err, "Failed to create stdout pipe")

	err = cmd.Start()
	assert.NoError(t, err, "Failed to start git cat-file command")

	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
	defer cancel()

	chunkCh := make(chan *sources.Chunk, 1000) // Adjust buffer size as needed

	go func() {
		defer close(chunkCh)
		err := HandleFile(ctx, stdout, &sources.Chunk{}, sources.ChanReporter{Ch: chunkCh}, WithSkipArchives(false))
		assert.NoError(t, err, "HandleFile should not return an error")
	}()

	err = cmd.Wait()
	assert.NoError(t, err, "git cat-file command should complete without error")

	count := 0
	for range chunkCh {
		count++
	}

	expectedChunks := 100 // This needs to be updated
	assert.Equal(t, expectedChunks, count, "Number of chunks should match the expected value")
}

func setupTempGitRepo(t *testing.T, archiveName string, fileSize int) string {
	tempDir := t.TempDir()

	// Initialize the Git repository
	cmd := exec.Command("git", "init", tempDir)
	var initStderr bytes.Buffer
	cmd.Stderr = &initStderr
	err := cmd.Run()
	if err != nil {
		t.Fatalf("Failed to initialize git repository: %v, stderr: %s", err, initStderr.String())
	}

	archivePath := filepath.Join(tempDir, archiveName)

	zipFile, err := os.Create(archivePath)
	if err != nil {
		t.Fatalf("Failed to create archive file: %v", err)
	}

	zipWriter := zip.NewWriter(zipFile)

	innerFileName := "largefile.txt"

	// Create a new file entry in the ZIP archive with no compression
	header := &zip.FileHeader{
		Name:   innerFileName,
		Method: zip.Store, // No compression
	}
	zipFileWriter, err := zipWriter.CreateHeader(header)
	if err != nil {
		t.Fatalf("Failed to create file in ZIP archive: %v", err)
	}

	dataChunk := bytes.Repeat([]byte("A"), 1024) // 1KB chunk
	totalWritten := 0
	for totalWritten < fileSize {
		remaining := fileSize - totalWritten
		if remaining < len(dataChunk) {
			_, err = zipFileWriter.Write(dataChunk[:remaining])
			if err != nil {
				t.Fatalf("Failed to write to inner file in ZIP archive: %v", err)
			}
			totalWritten += remaining
		} else {
			_, err = zipFileWriter.Write(dataChunk)
			if err != nil {
				t.Fatalf("Failed to write to inner file in ZIP archive: %v", err)
			}
			totalWritten += len(dataChunk)
		}
	}

	if err := zipWriter.Close(); err != nil {
		zipFile.Close()
		t.Fatalf("Failed to close ZIP writer: %v", err)
	}

	if err := zipFile.Close(); err != nil {
		t.Fatalf("Failed to close ZIP file: %v", err)
	}

	// Verify the ZIP archive's integrity
	verifyCmd := exec.Command("unzip", "-t", archivePath)
	verifyOutput, err := verifyCmd.CombinedOutput()
	if err != nil {
		t.Fatalf("Failed to verify ZIP archive: %v, output: %s", err, string(verifyOutput))
	}

	// Add the archive to the Git repository
	cmd = exec.Command("git", "-C", tempDir, "add", archiveName)
	var addStderr bytes.Buffer
	cmd.Stderr = &addStderr
	err = cmd.Run()
	if err != nil {
		t.Fatalf("Failed to add archive to git: %v, stderr: %s", err, addStderr.String())
	}

	cmd = exec.Command("git", "-C", tempDir, "commit", "-m", "Add archive file")
	var commitStderr bytes.Buffer
	cmd.Stderr = &commitStderr
	err = cmd.Run()
	if err != nil {
		t.Fatalf("Failed to commit archive to git: %v, stderr: %s", err, commitStderr.String())
	}

	return tempDir
}
```

ahrav:
> > The issue doesn't occur if I revert #3351. I have no idea why.
> 
> I still don't understand why adding the call to `.Size()` in HandleFile prevents the issue from occurring.
> 
> It seems like the issue is within the handler or bufferedreader code and that Git.handleBinary is just incidental.

I believe Size() consumes the entire reader if it's not seekable. It would hold it in memory if it's small enough otherwise, I _think_ it writes it to disk.

rgmz:
> I created some tests to see if I could reproduce it but the tests pass event without consuming the entire reader with the snippet above..

I can get it to reproduce [specifically with this file](https://github.com/cohere-ai/tokenizer/blob/8d3111ddb786e6f1d870f18fd2c408e2227e60ac/tokenizer/_tokenizer.cpython-38-x86_64-linux-gnu.so).

This issue isn't going anywhere, so probably something best looked at after a good night's rest.

```
=== RUN   TestHandleGitCatFileLargeBlob
    handlers_test.go:518: cat-file command still alive after 15 seconds
    handlers_test.go:522: 
        	Error Trace:	/home/gomez/dev/trufflehog/pkg/handlers/handlers_test.go:522
        	Error:      	Received unexpected error:
        	            	signal: broken pipe
        	Test:       	TestHandleGitCatFileLargeBlob
        	Messages:   	git cat-file command should complete without error
    handlers_test.go:530: 
        	Error Trace:	/home/gomez/dev/trufflehog/pkg/handlers/handlers_test.go:530
        	Error:      	Not equal: 
        	            	expected: 100
        	            	actual  : 0
        	Test:       	TestHandleGitCatFileLargeBlob
        	Messages:   	Number of chunks should match the expected value
--- FAIL: TestHandleGitCatFileLargeBlob (15.17s)

Expected :100
Actual   :0
<Click to see difference>
```

<details>
<summary>Updated test code (click to expand)</summary>

```go
func TestHandleGitCatFileLargeBlob(t *testing.T) {
	fileName := "_tokenizer.cpython-38-x86_64-linux-gnu.so"

	// Set up a temporary git repository with a large file
	gitDir := setupTempGitRepo(t, fileName)
	defer os.RemoveAll(gitDir)

	cmd := exec.Command("git", "-C", gitDir, "rev-parse", "HEAD")
	hashBytes, err := cmd.Output()
	assert.NoError(t, err, "Failed to get commit hash")
	commitHash := strings.TrimSpace(string(hashBytes))

	// Create a pipe to simulate the git cat-file stdout
	cmd = exec.Command("git", "-C", gitDir, "cat-file", "blob", fmt.Sprintf("%s:%s", commitHash, fileName))

	var stderr bytes.Buffer
	cmd.Stderr = &stderr

	stdout, err := cmd.StdoutPipe()
	assert.NoError(t, err, "Failed to create stdout pipe")

	err = cmd.Start()
	assert.NoError(t, err, "Failed to start git cat-file command")

	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
	defer cancel()

	chunkCh := make(chan *sources.Chunk, 1000) // Adjust buffer size as needed

	go func() {
		defer close(chunkCh)
		err := HandleFile(ctx, stdout, &sources.Chunk{}, sources.ChanReporter{Ch: chunkCh}, WithSkipArchives(false))
		assert.NoError(t, err, "HandleFile should not return an error")
	}()

	time.AfterFunc(15*time.Second, func() {
		t.Errorf("cat-file command still alive after 15 seconds")
		stdout.Close()
	})
	err = cmd.Wait()
	assert.NoError(t, err, "git cat-file command should complete without error")

	count := 0
	for range chunkCh {
		count++
	}

	expectedChunks := 100 // This needs to be updated
	assert.Equal(t, expectedChunks, count, "Number of chunks should match the expected value")
}

func setupTempGitRepo(t *testing.T, archiveName string) string {
	tempDir := t.TempDir()

	// Initialize the Git repository
	cmd := exec.Command("git", "init", tempDir)
	var initStderr bytes.Buffer
	cmd.Stderr = &initStderr
	err := cmd.Run()
	if err != nil {
		t.Fatalf("Failed to initialize git repository: %v, stderr: %s", err, initStderr.String())
	}

	archivePath := filepath.Join(tempDir, archiveName)
	cmd = exec.Command("cp", "/tmp/_tokenizer.cpython-38-x86_64-linux-gnu.so", archivePath)
	var cpStderr bytes.Buffer
	cmd.Stderr = &cpStderr
	err = cmd.Run()
	if err != nil {
		t.Fatalf("Failed to copy file: %v, stderr: %s", err, cpStderr.String())
	}

	// Add the archive to the Git repository
	cmd = exec.Command("git", "-C", tempDir, "add", archiveName)
	var addStderr bytes.Buffer
	cmd.Stderr = &addStderr
	err = cmd.Run()
	if err != nil {
		t.Fatalf("Failed to add archive to git: %v, stderr: %s", err, addStderr.String())
	}

	cmd = exec.Command("git", "-C", tempDir, "commit", "-m", "Add archive file")
	var commitStderr bytes.Buffer
	cmd.Stderr = &commitStderr
	err = cmd.Run()
	if err != nil {
		t.Fatalf("Failed to commit archive to git: %v, stderr: %s", err, commitStderr.String())
	}

	return tempDir
}
```
</details>

rgmz:
>>> Edit: furthermore, changing executeCatFileCmd to simply write the files to disk causes it to run buttery smooth, as expected. No hanging.
>> 
> > It seems like the issue is within the handler or bufferedreader code and that Git.handleBinary is just incidental.
> 
> I believe Size() consumes the entire reader if it's not seekable. It would hold it in memory if it's small enough otherwise, I _think_ it writes it to disk.

I think this affirms that the issue is with `NewBufferedReaderSeeker` not consuming the entire output, hence why `cmd.Wait()` never completes.

> Wait waits for the command to exit and **waits for any copying ... from stdout or stderr to complete**. 
> https://pkg.go.dev/os/exec#Cmd.Wait

rgmz:
>  I think this affirms that the issue is with NewBufferedReaderSeeker not consuming the entire output, hence why cmd.Wait() never completes.

I think I've figured it out: the file is skipped due to its extension/mimetype, therefore the reader is never read/drained.

https://github.com/trufflesecurity/trufflehog/blob/40fdf440734931798e79c21e6dd06315aa43f8c4/pkg/handlers/default.go#L77-L81

I had to manually comment out the noisy `decoder not applicable for chunk` to notice this line in the `--trace` output. Perhaps it should be moved up?

```
[BufferedReadSeeker#Read] Read '3072' bytes, err = <nil>
[BufferedReadSeeker#Read] read position inside memory buffer matches out = '2'
[BufferedReadSeeker#Read] read position inside memory buffer matches out = '8'
[BufferedReadSeeker#Read] read position inside memory buffer matches out = '502'
>>>>>>>>>>>>>>>>>> 2024-10-07T09:04:53-04:00	info-5	trufflehog	skipping file	{"source_manager_worker_id": "Om7lm", "unit": "https://github.com/cohere-ai/tokenizer.git", "unit_kind": "repo", "repo": "https://github.com/cohere-ai/tokenizer.git", "mime": "application/x-sharedlib", "timeout": 60, "ext": ".so"}
2024-10-07T09:04:53-04:00	info-5	trufflehog	handler channel closed, all chunks processed	{"source_manager_worker_id": "Om7lm", "unit": "https://github.com/cohere-ai/tokenizer.git", "unit_kind": "repo", "repo": "https://github.com/cohere-ai/tokenizer.git", "mime": "application/x-sharedlib", "timeout": 60}
```

rgmz:
@ahrav what's the most efficient way to ensure the reader is flushed?

I would have thought it was handled by .Close():
https://github.com/trufflesecurity/trufflehog/blob/57802abf5276961426482ace808057016c714958/pkg/handlers/handlers.go#L286

ahrav:
> specifically with this file

I was initially thinking something like:
```
	handler := selectHandler(mimeT, rdr.isGenericArchive)
	archiveChan, err := handler.HandleFile(processingCtx, rdr)
	if err != nil {
		// Ensure all data is read to prevent broken pipe
		_, _ = io.Copy(io.Discard, reader)
		if closeErr := rdr.Close(); closeErr != nil {
			ctx.Logger().Error(closeErr, "error closing reader after handler error")
		}
		return fmt.Errorf("error handling file: %w", err)
	}

	handleErr := handleChunks(processingCtx, archiveChan, chunkSkel, reporter)
	if handleErr != nil {
		// Ensure all data is read to prevent broken pipe
		_, _ = io.Copy(io.Discard, reader)
		if closeErr := rdr.Close(); closeErr != nil {
			ctx.Logger().Error(closeErr, "error closing reader after handleChunks error")
		}
		return fmt.Errorf("error processing chunks: %w", handleErr)
	}
```

But i'm not sure that handles all cases given `handleNonArchiveContent` happens in a goroutine. We might need to add the `io.Copy(io.Discard, reader)` in there as well.

```
if common.SkipFile(mimeExt) || common.IsBinary(mimeExt) {
    ctx.Logger().V(5).Info("skipping file", "ext", mimeExt)
    h.metrics.incFilesSkipped()
    // Make sure we consume the reader to avoid potentially blocking indefinitely.
    _, _ = io.Copy(io.Discard, reader)
    return nil
}
```

I'm working on a mechanism to propagate errors from the goroutines back to the caller, but I think using io.Copy _should_ work.

rgmz:
That looks worthwhile, though it wouldn't fix this specific issue as `handleNonArchiveContent` returns `nil` when a file is skipped. Unless your intention is to return a sentinel error for that?

Either way, I think this can be closed in favour of #3379 + your future change.

ahrav:
> That looks worthwhile, though it wouldn't fix this specific issue as `handleNonArchiveContent` returns `nil` when a file is skipped. Unless your intention is to return a sentinel error for that?
> 
> Either way, I think this can be closed in favour of #3379 + your future change.
> 
> Edit: I misread the second example. That could work -- however, any future changes or new paths could cause a regression. I think an ideal solution would be to handle it at the top level (`handlers.HandleFile`).

I agree. We either need to drain the reader or propagate the error. What do you think?

Ideally, I'd like a single place to handle error management, but that might not be practical.

`handleNonArchiveContent` _seems_ central since all handlers eventually pass data to it. However, putting the logic in a specific implementation still doesn't feel ideal.

rgmz:
> handleNonArchiveContent seems central since all handlers eventually pass data to it. However, putting the logic in a specific implementation still doesn't feel ideal.

After looking through the code, it seems like the best place — for now. Adding test cases for top-level and ignored nested files for each handler should help mitigate anything being missed.

dustin-decker:
I've updated the smoke test in #3379 to include the problematic file in a test fixture repo.

@ahrav do you want to update that PR with the fix required?

PR #3384, opened 2024-10-08
Enhanced the easyinsight detector

kashifkhan0771:
### Description:
This PR enhance the following things in `easyinsight` detector.

- Fixed Regex
- Updated logic to skip processing when key and id is same
- Skip processing remaining id's for a key when a key is successfully validated with one id.
- Added detail pattern test cases.
- Added comments
- Renamed variables for better readability

### Pattern Test Cases:
easyinsight

### Checklist:

### DIFF
--- a/pkg/detectors/easyinsight/easyinsight.go
+++ b/pkg/detectors/easyinsight/easyinsight.go
@@ -4,16 +4,18 @@ import (
 	"context"
 	b64 "encoding/base64"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
+	"io"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{
+type Scanner struct {
 	detectors.DefaultMultiPartCredentialProvider
 }
 
@@ -24,8 +26,8 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight"}) + `\b([0-9Aa-zA-Z]{20})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight"}) + `\b([a-zA-Z0-9]{20})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight", "key"}) + `\b([0-9a-zA-Z]{20})\b`)
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight", "id"}) + `\b([a-zA-Z0-9]{20})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
@@ -38,20 +40,22 @@ func (s Scanner) Keywords() []string {
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
 
-	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+	keyMatches := keyPat.FindAllStringSubmatch(dataStr, -1)
 
-	idmatches := idPat.FindAllStringSubmatch(dataStr, -1)
+	idMatches := idPat.FindAllStringSubmatch(dataStr, -1)
 
-	for _, match := range matches {
-		if len(match) != 2 {
-			continue
-		}
-		resMatch := strings.TrimSpace(match[1])
-		for _, idmatch := range idmatches {
-			if len(idmatch) != 2 {
+	for _, keyMatch := range keyMatches {
+		resMatch := strings.TrimSpace(keyMatch[1])
+
+		for _, idMatch := range idMatches {
+			resIdMatch := strings.TrimSpace(idMatch[1])
+			/*
+				as key and id regex are same, the strings captured by both regex will be same.
+				avoid processing when key is same as id. This will allow detector to process only different combinations
+			*/
+			if resMatch == resIdMatch {
 				continue
 			}
-			resIdMatch := strings.TrimSpace(idmatch[1])
 
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_EasyInsight,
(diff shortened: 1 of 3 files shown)


pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -24,8 +26,8 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight"}) + `\b([0-9Aa-zA-Z]{20})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight"}) + `\b([a-zA-Z0-9]{20})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight", "key"}) + `\b([0-9a-zA-Z]{20})\b`)
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"easyinsight", "easy-insight", "id"}) + `\b([a-zA-Z0-9]{20})\b`)
```
abmussani, line 30: @kashifkhan0771 'key' and 'id' are very generic keywords. Using them will increase the chances of false positive cases

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -38,20 +40,22 @@ func (s Scanner) Keywords() []string {
...
 
-	for _, match := range matches {
-		if len(match) != 2 {
-			continue
-		}
-		resMatch := strings.TrimSpace(match[1])
-		for _, idmatch := range idmatches {
-			if len(idmatch) != 2 {
+	for _, keyMatch := range keyMatches {
+		resMatch := strings.TrimSpace(keyMatch[1])
```
rgmz, line 48: TrimSpace is redundant since the pattern doesn't capture space.

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -38,20 +40,22 @@ func (s Scanner) Keywords() []string {
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
 
-	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+	keyMatches := keyPat.FindAllStringSubmatch(dataStr, -1)
```
rgmz, line 43: I think it's worthwhile to filter unique values for both key and ID prior to looping through both.

@zricethezav do you think this would benefit from creating a cartesian product of possible values?

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -60,27 +64,37 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+
+				// add required headers to the request
 				req.Header.Add("Content-Type", "application/json")
 				req.Header.Add("Accept", "application/json")
 				req.Header.Add("Authorization", fmt.Sprintf("Basic %s", sEnc))
+
 				res, err := client.Do(req)
 				if err == nil {
-					defer res.Body.Close()
+					// discard the body content and close it at the end of each iteration.
```
rgmz, line 82: I think this should handle the error case. Only handling `err == nil` means the site could go offline and stop resolving, and we'd never know.

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -60,27 +64,37 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					_, _ = io.Copy(io.Discard, res.Body)
+					_ = res.Body.Close()
+
 					if res.StatusCode >= 200 && res.StatusCode < 300 {
```
rgmz, line 86: We should pick specific status codes that are considered valid. Permissive ranges like this have been a source of many false positives.

rgmz (commented):
Some suggestions on improvement, which is applicable for any future detector changes.

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -60,27 +64,37 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 
 			if verify {
-				data := fmt.Sprintf("%s:%s", resIdMatch, resMatch)
-				sEnc := b64.StdEncoding.EncodeToString([]byte(data))
+				auth := fmt.Sprintf("%s:%s", resIdMatch, resMatch)
+				sEnc := b64.StdEncoding.EncodeToString([]byte(auth))
+
```
abmussani, line 69: **Nit**: Moving the verification code will improve the resources management (memory clean-up etc) and code readability.

abmussani (commented):
Thanks for taking this up. Here are few comments regarding the selection of keywords and improvements of resources utilization.

pkg/detectors/easyinsight/easyinsight.go
kashifkhan0771: I kept them because the data chunk contains the keyword `easyinsight` or `easy-insight`, and this will allow us to search for a key or ID within that chunk. This approach should increase the likelihood of detecting the key and ID in the data which has the keywords for this detector.

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -38,20 +40,22 @@ func (s Scanner) Keywords() []string {
...
-		}
-		resMatch := strings.TrimSpace(match[1])
-		for _, idmatch := range idmatches {
-			if len(idmatch) != 2 {
+	for _, keyMatch := range keyMatches {
+		resMatch := strings.TrimSpace(keyMatch[1])
+
+		for _, idMatch := range idMatches {
+			resIdMatch := strings.TrimSpace(idMatch[1])
+			/*
```
zricethezav, line 54: **nit**: Let's keep the commenting style consistent -- `//` instead of `/*` blocks

pkg/detectors/easyinsight/easyinsight.go
rgmz: > @rgmz Aren't we doing this already with the nested loops?

I suppose so, once the unique filtering is added. I was thinking the manual checks for `key == id` are a bit awkward.

pkg/detectors/easyinsight/easyinsight.go
```diff
@@ -38,49 +38,48 @@ func (s Scanner) Keywords() []string {
...
-				req.Header.Add("Accept", "application/json")
-				req.Header.Add("Authorization", fmt.Sprintf("Basic %s", sEnc))
-				res, err := client.Do(req)
-				if err == nil {
-					defer res.Body.Close()
-					if res.StatusCode >= 200 && res.StatusCode < 300 {
-						s1.Verified = true
-					}
+				auth := fmt.Sprintf("%s:%s", idMatch, keyMatch)
+				sEnc := b64.StdEncoding.EncodeToString([]byte(auth))
```
rgmz, line 68: Can be simplified with https://pkg.go.dev/net/http#Request.SetBasicAuth

kashifkhan0771: `id` as username and `key` as password in Basic Auth?

rgmz: Yeah, rather than manually base64 encoding.

pkg/detectors/easyinsight/easyinsight.go
kashifkhan0771: Documentation actually does not provide any information about which status code the API will return for different scenario.

abmussani: @kashifkhan0771 Here is the [documentation](https://www.easy-insight.com/api/responses.html) about supported HTTP Status codes.

kashifkhan0771: Thanks @abmussani!

rgmz (approved):
Looks good to me.

zricethezav (approved):

PR #3386, opened 2024-10-09
Updated Fastly Personal Token Detector

kashifkhan0771:
### Description:

- Updated the verification API to `/tokens/self`. 
- Restructured the code
- Added pattern test cases

### Output format for valid Token:
```
✅ Found verified result 🐷🔑
Detector Type: FastlyPersonalToken
Decoder Type: PLAIN
Raw result: <found_token>
Token_id: <token_id>
User_id: <user_id>
Token_expires_at: <expires_at> (Format: 2024-10-09T19:00:00Z)
Token_scope: global global:read
File: <file>
Line: <line_no>
```
### Pattern Test Cases:
Screenshot from 2024-10-09 13-07-59

### Checklist:

### DIFF
--- a/pkg/detectors/fastlypersonaltoken/fastlypersonaltoken.go
+++ b/pkg/detectors/fastlypersonaltoken/fastlypersonaltoken.go
@@ -35,32 +35,30 @@ type token struct {
 	TokenID   string `json:"id"`
 	UserID    string `json:"user_id"`
 	ExpiresAt string `json:"expires_at"`
+	Scope     string `json:"scope"`
 }
 
 // FromData will find and optionally verify FastlyPersonalToken secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
 
-	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+	var uniqueMatches = make(map[string]struct{})
 
-	for _, match := range matches {
-		resMatch := match[1]
+	for _, matches := range keyPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueMatches[matches[1]] = struct{}{}
+	}
 
+	for match := range uniqueMatches {
 		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_FastlyPersonalToken,
-			Raw:          []byte(resMatch),
+			Raw:          []byte(match),
 		}
 
 		if verify {
-			extraData, verified, verificationErr := verifyFastlyApiToken(ctx, resMatch)
+			extraData, verified, verificationErr := verifyFastlyApiToken(ctx, match)
 			s1.Verified = verified
-			if extraData != nil {
-				s1.ExtraData = extraData
-			}
-
-			if verificationErr != nil {
-				s1.SetVerificationError(verificationErr)
-			}
+			s1.ExtraData = extraData
+			s1.SetVerificationError(verificationErr)
 		}
 
 		results = append(results, s1)
@@ -108,8 +106,15 @@ func verifyFastlyApiToken(ctx context.Context, apiToken string) (map[string]stri
 			"token_id": self.TokenID,
 			// user id is the alphanumeric string uniquely identifying the user
 			"user_id": self.UserID,
-			// expires at is time-stamp (UTC) of when the token will expire.
+			// expires at is time-stamp (UTC) of when the token will expire
 			"token_expires_at": self.ExpiresAt,
+			// token scope is space-delimited list of authorization scope of the token
+			"token_scope": self.Scope,
+		}
+
+		// if expires at is empty which mean token is set to never expire, add 'Never' as the value
+		if extraData["token_expires_at"] == "" {
+			extraData["token_expires_at"] = "never"
 		}
 
 		return extraData, true, nil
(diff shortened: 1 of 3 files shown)


pkg/detectors/fastlypersonaltoken/fastlypersonaltoken.go
```diff
@@ -31,12 +31,10 @@ func (s Scanner) Keywords() []string {
...
-type fastlyUserRes struct {
-	Login                string `json:"login"`
-	Name                 string `json:"name"`
-	Role                 string `json:"role"`
-	TwoFactorAuthEnabled bool   `json:"two_factor_auth_enabled"`
-	Locked               bool   `json:"locked"`
+type token struct {
+	TokenID   string `json:"id"`
+	UserID    string `json:"user_id"`
+	ExpiresAt string `json:"expires_at"`
```
abmussani, line 37: @kashifkhan0771 Response also contains scope of that token. My suggestion would be to include it in the extraData.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken_integration_test.go
```diff
@@ -0,0 +1,120 @@
...
+				t.Errorf("FastlyPersonalToken.FromData() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			for i := range got {
+				if len(got[i].Raw) == 0 {
+					t.Fatalf("no raw secret present: \n %+v", got[i])
+				}
+				got[i].Raw = nil
+			}
+			if diff := pretty.Compare(got, tt.want); diff != "" {
```
abmussani, line 99: I have doubt that this test will fail due to extraData diff (nil vs Map) in valid & verified cases.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken_test.go
```diff
@@ -1,120 +1,92 @@
...
-			want: []detectors.Result{
-				{
-					DetectorType: detectorspb.DetectorType_FastlyPersonalToken,
-					Verified:     true,
-				},
-			},
-			wantErr: false,
+			name:  "valid pattern",
+			input: validPattern,
+			want:  []string{"TVAWji0p7uDI6OP9DyWvmV-vgoUoXIuf"},
```
abmussani, line 42: Can we pass on validPattern variable in `want`? It will consolidate valid token in one variable, update will become easy.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken.go
```diff
@@ -100,3 +76,48 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+		}
+
+		// capture token details in the map
+		extraData := map[string]string{
+			// token id is the alphanumeric string uniquely identifying a token
+			"token_id": self.TokenID,
+			// user id is the alphanumeric string uniquely identifying the user
+			"user_id": self.UserID,
+			// expires at is time-stamp (UTC) of when the token will expire.
+			"token_expires_at": self.ExpiresAt,
```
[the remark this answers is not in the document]
kashifkhan0771: I have actually created a token which will expire tomorrow. I will try to call the API with that token and will see how API response. Either 200OK with expired at or 401. But as per their documentation it should return 401 when a token get expired.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken_integration_test.go
kashifkhan0771: Yeah, good catch! I'll update these test cases.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken_test.go
kashifkhan0771: ValidPattern is actually a long testing string, I can create a separate variable with output.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken.go
```diff
@@ -31,60 +31,34 @@ func (s Scanner) Keywords() []string {
...
-						"role":                    userRes.Role,
-						"locked":                  fmt.Sprintf("%t", userRes.Locked),
-						"two_factor_auth_enabled": fmt.Sprintf("%t", userRes.TwoFactorAuthEnabled),
-					}
-				}
-			}
+			extraData, verified, verificationErr := verifyFastlyApiToken(ctx, match)
+			s1.Verified = verified
+			s1.ExtraData = extraData
+			s1.SetVerificationError(verificationErr)
```
rgmz, line 61: I think we're supposed to pass the secret(s) as varargs to SetVerificationErr so they get redacted? Been a while and I'm on my phone so Idk.

I think @ahrav would know.

kashifkhan0771: Yeah, it does take secrets as a optional second parameter.

pkg/detectors/fastlypersonaltoken/fastlypersonaltoken.go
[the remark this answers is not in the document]
zricethezav: @kashifkhan0771 keep us posted. Let's hold off on merging until the expiry behavior has been confirmed.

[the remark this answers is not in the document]
kashifkhan0771: It returns 401 😃 - Once token expires, we can no longer see it on the fastly and the API returns 401.

zricethezav (approved):
Great work!

PR #3395, opened 2024-10-10
Add SliceContainsString common util

bill-rich:


sysread (approved):
Personally, I would include the matched index in the return value, but that's a nitpick for sure.

pkg/common/utils.go
```diff
@@ -81,3 +81,20 @@ func GetAccountNumFromAWSID(AWSID string) (string, error) {
...
 }
+
+// SliceContainsString searches a slice to determine if it contains a specified string.
+func SliceContainsString(origTargetString string, stringSlice []string, ignoreCase bool) (bool, string) {
+	targetString := origTargetString
+	if ignoreCase {
+		targetString = strings.ToLower(origTargetString)
+	}
+	for _, stringFromSlice := range stringSlice {
+		if ignoreCase {
```
ahrav, line 92: suggestion: If `ignoreCase` is true, we can avoid the overhead of calling `strings.ToLower` on each slice item by performing the conversion once.

Ex:
```
if ignoreCase {
		targetString := strings.ToLower(origTargetString)
		for _, stringFromSlice := range stringSlice {
			if targetString == strings.ToLower(stringFromSlice) {
				return true, origTargetString
			}
		}
	} else {
		for _, stringFromSlice := range stringSlice {
			if origTargetString == stringFromSlice {
				return true, origTargetString
			}
		}
	}
	return false, ""
```

pkg/common/utils.go
```diff
@@ -81,3 +81,20 @@ func GetAccountNumFromAWSID(AWSID string) (string, error) {
 	accountNum := (z & mask) >> 7
 	return fmt.Sprintf("%012d", accountNum), nil
 }
+
+// SliceContainsString searches a slice to determine if it contains a specified string.
+func SliceContainsString(origTargetString string, stringSlice []string, ignoreCase bool) (bool, string) {
```
ahrav, line 86: question: How large are you anticipating `stringSlice` being?

pkg/common/utils.go
bill-rich: It currently only does the `ToLower` conversion if `ignoreCase` is true. The original way does the check on every iteration, uses the same loop.

bill-rich:
> Personally, I would include the matched index in the return value, but that's a nitpick for sure.

Good call. Added

pkg/common/utils.go
bill-rich: I imagine most cases won't be huge, but didn't really have a limit in mind.

pkg/common/utils.go
ahrav: Oh right duh!

pkg/common/utils.go
```diff
@@ -81,3 +81,22 @@ func GetAccountNumFromAWSID(AWSID string) (string, error) {
 	accountNum := (z & mask) >> 7
 	return fmt.Sprintf("%012d", accountNum), nil
 }
+
+// SliceContainsString searches a slice to determine if it contains a specified string.
+// Returns the index of the first match in the slice.
+func SliceContainsString(origTargetString string, stringSlice []string, ignoreCase bool) (bool, string, int) {
+	targetString := origTargetString
```
ahrav, line 88: question: How come you created `targetString` and `stringFromSlice` vars as opposed to updating the existing values? Was it solely for clarity?
```
if ignoreCase {
    origTargetString = strings.ToLower(origTargetString)
}
```

similarly for:
```
if ignoreCase {
    origStringFromSlice = strings.ToLower(origStringFromSlice)
}
```

PR #3401, opened 2024-10-11
[bug] - fix resource leaks

ahrav:
### Description:
This PR fixes an issue that can cause resource exhaustion when scanning binaries in Git. It ensures the `stdoutPipe` reader passed to `HandleFile` is always drained, even in case of an error or context cancellation. If the reader isn't fully consumed, the `cmd.Wait()` method may hang, failing to release file descriptors. Additionally, this PR adds a context timeout to the `git cat-file` command, with a brief wait to allow the process time to complete.

Error msg: ` scanner error: error running git cat-file: pipe2: too many open files`
Screenshot 2024-10-11 at 3 10 20 PM

### Checklist:

### DIFF
--- a/pkg/handlers/ar.go
+++ b/pkg/handlers/ar.go
@@ -31,8 +31,6 @@ func (h *arHandler) HandleFile(ctx logContext.Context, input fileReader) (chan [
 	}
 
 	go func() {
-		ctx, cancel := logContext.WithTimeout(ctx, maxTimeout)
-		defer cancel()
 		defer close(archiveChan)
 
 		// Update the metrics for the file processing.
--- a/pkg/handlers/handlers.go
+++ b/pkg/handlers/handlers.go
@@ -83,16 +83,32 @@ func newMimeTypeReader(r io.Reader) (mimeTypeReader, error) {
 func newFileReader(r io.Reader) (fileReader, error) {
 	var fReader fileReader
 
-	fReader.BufferedReadSeeker = iobuf.NewBufferedReaderSeeker(r)
+	// To detect the MIME type of the input data, we need a reader that supports seeking.
+	// This allows us to read the data multiple times if necessary without losing the original position.
+	// We use a BufferedReaderSeeker to wrap the original reader, enabling this functionality.
+	// If an error occurs during MIME type detection, it is important we close the BufferedReaderSeeker
+	// to release any resources it holds and prevent potential memory leaks.
+	bufferedReader := iobuf.NewBufferedReaderSeeker(r)
+	fReader.BufferedReadSeeker = bufferedReader
 
-	mime, err := mimetype.DetectReader(fReader)
+	var err error
+	defer func() {
+		if err != nil {
+			if closeErr := bufferedReader.Close(); closeErr != nil {
+				err = fmt.Errorf("%w; error closing reader: %w", err, closeErr)
+			}
+		}
+	}()
+
+	var mime *mimetype.MIME
+	mime, err = mimetype.DetectReader(fReader)
 	if err != nil {
 		return fReader, fmt.Errorf("unable to detect MIME type: %w", err)
 	}
 	fReader.mime = mime
 
 	// Reset the reader to the beginning because DetectReader consumes the reader.
-	if _, err := fReader.Seek(0, io.SeekStart); err != nil {
+	if _, err = fReader.Seek(0, io.SeekStart); err != nil {
 		return fReader, fmt.Errorf("error resetting reader after MIME detection: %w", err)
 	}
 
@@ -102,7 +118,8 @@ func newFileReader(r io.Reader) (fileReader, error) {
 		return fReader, nil
 	}
 
-	format, _, err := archiver.Identify("", fReader)
+	var format archiver.Format
+	format, _, err = archiver.Identify("", fReader)
 	switch {
 	case err == nil:
 		fReader.isGenericArchive = true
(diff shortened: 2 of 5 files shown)


pkg/handlers/rpm.go
```diff
@@ -31,8 +31,6 @@ func (h *rpmHandler) HandleFile(ctx logContext.Context, input fileReader) (chan
 	}
 
 	go func() {
-		ctx, cancel := logContext.WithTimeout(ctx, maxTimeout)
```
ahrav, line 34: Removed because the timeout is set by the caller.

pkg/handlers/ar.go
```diff
@@ -31,8 +31,6 @@ func (h *arHandler) HandleFile(ctx logContext.Context, input fileReader) (chan [
 	}
 
 	go func() {
-		ctx, cancel := logContext.WithTimeout(ctx, maxTimeout)
```
ahrav, line 34: Removed because the timeout is set by the caller.

pkg/handlers/handlers_test.go
```diff
@@ -684,3 +682,144 @@ func setupTempGitRepoCommon(t *testing.T, fileName string, fileSize int, isUnsup
 
 	return tempDir
 }
+
+// errorOnceReader is a custom io.Reader that simulates a read error only once.
+// On the first read, it returns an error. Subsequent reads return the remaining data normally.
+// This is used to simulate an error during mime type detection in newFileReader.
+type errorOnceReader struct {
```
ahrav, line 689: @mcastorina I feel like I did something like this before and you mentioned there was a pkg to do this, or maybe i'm thinking of something else?

pkg/sources/git/git.go
```diff
@@ -1253,9 +1270,17 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
-	defer func() { _ = cmd.Wait() }()
+	var waitErr error
+	defer func() {
+		waitErr = cmd.Wait()
+		if err != nil && waitErr != nil {
+			err = fmt.Errorf("handle error: %v; wait error: %v", err, waitErr)
+			return
+		}
+		err = waitErr
+	}()
```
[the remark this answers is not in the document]
mcastorina: Also, it feels weird to not wrap one of the errors even though I know we don't unwrap anywhere 🤣 

What about wrapping the original error?

```go
err = fmt.Errorf("handle error: %w; wait error: %v", err, waitErr)
```

pkg/handlers/handlers.go
```diff
@@ -83,16 +83,32 @@ func newMimeTypeReader(r io.Reader) (mimeTypeReader, error) {
...
 	var fReader fileReader
 
-	fReader.BufferedReadSeeker = iobuf.NewBufferedReaderSeeker(r)
+	// To detect the MIME type of the input data, we need a reader that supports seeking.
+	// This allows us to read the data multiple times if necessary without losing the original position.
+	// We use a BufferedReaderSeeker to wrap the original reader, enabling this functionality.
+	// If an error occurs during MIME type detection, it is important we close the BufferedReaderSeeker
+	// to release any resources it holds and prevent potential memory leaks.
+	bufferedReader := iobuf.NewBufferedReaderSeeker(r)
+	fReader.BufferedReadSeeker = bufferedReader
```
mcastorina, line 92: Why was this split into two lines?

pkg/handlers/handlers.go
```diff
@@ -83,16 +83,32 @@ func newMimeTypeReader(r io.Reader) (mimeTypeReader, error) {
...
+
+	var err error
+	defer func() {
+		if err != nil {
+			if closeErr := bufferedReader.Close(); closeErr != nil {
+				err = fmt.Errorf("%w; error closing reader: %w", err, closeErr)
+			}
+		}
```
mcastorina, line 100: Why do we only close in the error case?

pkg/handlers/handlers_test.go
mcastorina: I'll usually look in [testing/iotest](https://pkg.go.dev/testing/iotest) for stuff like that, but I don't see any similar to this functionality.

pkg/handlers/handlers.go
ahrav: In the non-error case, the caller is responsible for closing the reader, so I should clarify that the caller generally handles this after use. In this case, `Close` only returns the buffer to the pool if MIME type detection fails.

To reduce confusion, I could change the return type to a pointer and return `nil`. Also, since we lazily fetch the buffer in `Read`, I might avoid checking out the buffer from the pool in `NewBufferedReaderSeeker`, which leaves the responsibility of closing on the caller.

pkg/sources/git/git.go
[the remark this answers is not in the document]
ahrav: Nope, not sure why I thought i'd need it scoped at that level. Moved it inside the defer and used the short assignment approach.

pkg/sources/git/git.go
```diff
@@ -1240,9 +1247,19 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
+	const (
+		cmdTimeout = 60 * time.Second
+		waitDelay  = 5 * time.Second
+	)
+	// Create a timeout context for the 'git cat-file' command to ensure it does not run indefinitely.
+	// This prevents potential resource exhaustion by terminating the command if it exceeds the specified duration.
+	catFileCtx, cancel := context.WithTimeoutCause(fileCtx, cmdTimeout, errors.New("git cat-file timeout"))
+	defer cancel()
+
+	cmd := exec.CommandContext(catFileCtx, "git", "-C", gitDir, "cat-file", "blob", commitHash.String()+":"+path)
```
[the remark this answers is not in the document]
ahrav: Sigh, this would've saved us so much headache 😭 .

ahrav:
@mcastorina Gonna close this PR and split it into multiple PRs. I addressed your comments in the new PR, but replied here. Sorry i'm making you double review. 😅 
[new PR](https://github.com/trufflesecurity/trufflehog/pull/3402)

PR #3402, opened 2024-10-12
[fix] - resource leak

ahrav:
### Description:
This PR fixes an issue that can cause resource exhaustion when scanning binaries in Git. It ensures the `stdoutPipe` reader passed to `HandleFile` is always drained. If the reader isn't fully consumed, the `cmd.Wait()` method may hang, failing to release file descriptors.

This PR adds a context timeout to the `git cat-file` command, with a brief wait to allow the process time to complete before the process is killed.

Error msg: ` scanner error: error running git cat-file: pipe2: too many open files`
Screenshot 2024-10-11 at 3 10 20 PM

### Checklist:


pkg/handlers/handlers.go
```diff
@@ -285,14 +285,9 @@ func HandleFile(
 	}
 	defer func() {
 		// Ensure all data is read to prevent broken pipe.
-		_, copyErr := io.Copy(io.Discard, rdr)
```
ahrav, line 288: This is only necessary when the underlying reader is a pipe. If the reader isn't fully consumed, `cmd.Wait()` hangs indefinitely. Since `HandleFile` is used by other parts of the codebase with different readers, the logic to consume the reader (if it's a pipe) has been moved to the caller.

pkg/sources/git/git.go
```diff
@@ -1253,9 +1283,44 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
+
+	return handlers.HandleFile(catFileCtx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
+}
 
-	return handlers.HandleFile(ctx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
+// combineErrors combines two errors into one, preserving both error messages if they exist.
+func combineErrors(err1, err2 error) error {
+	switch {
+	case err1 != nil && err2 != nil:
+		return fmt.Errorf("%w; %w", err1, err2)
```
mcastorina, line 1318: Wait, you can do this? Which error is returned when you call `Unwrap()`?

pkg/sources/git/git.go
```diff
@@ -1253,9 +1283,44 @@ func (s *Git) handleBinary(ctx context.Context, gitDir string, reporter sources.
...
+			"bytes_discarded", n)
+
+		err = combineErrors(err, fmt.Errorf("error discarding excess stdout data: %w", copyErr))
+
+		// Wait for the command to finish and handle any errors.
+		waitErr := cmd.Wait()
+		err = combineErrors(err, fmt.Errorf("command execution error: %w", waitErr))
+	}()
+
+	return handlers.HandleFile(catFileCtx, stdout, chunkSkel, reporter, handlers.WithSkipArchives(s.skipArchives))
```
mcastorina, line 1311: Does `HandleFile` call `rdr.Close()` (and in this case `rdr` is `stdout`)? If yes, is it an issue to read from `stdout` after `Close()` has been called on it (in the `defer` block above)?

pkg/sources/git/git.go
ahrav: I thought you could do this—I was wrong. I can just go with `errors.Join`.

pkg/sources/git/git.go
ahrav: It does, but the `HandleFile` function does not directly close `stdout`. Instead, it wraps `stdout` in a `BufferedReadSeeker`, and when `HandleFile` calls `rdr.Close()`, it only closes the `BufferedReadSeeker`, not `stdout`. The `Close()` method of `BufferedReadSeeker` manages internal resources like returning buffers to a pool or closing temporary files without affecting the underlying `stdout`. [here](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/iobuf/bufferedreaderseeker.go#L371-L381)

So `stdout` remains open until `cmd.Wait()` completes in `handleBinary`. The deferred function in `handleBinary` ensures proper cleanup by reading any remaining data from `stdout` and waiting for the command to finish, which effectively closes `stdout`.

I feel like this approach effectively separates ownership: `HandleFile` manages the custom reader, while `handleBinary` is responsible for handling and closing `stdout` via `cmd.Wait()`.

pkg/sources/git/git.go
ahrav: Oh, nvm 🤣 Just read `In particular Unwrap does not unwrap errors returned by [Join](https://pkg.go.dev/errors#Join).`

pkg/sources/git/git.go
mcastorina: Okay, so `HandleFile` takes in an `io.Reader`, not `io.ReadCloser`. Closing is up to the caller.

Yeah, this is a great approach!

ahrav: It used to be an `io.ReadCloser`, but I confused myself a few times by mistakenly closing the passed-in reader and then trying to close it again at the call site. That's when I decided to change it to an `io.Reader` and have the callers handle closing. I think that's maybe why you also thought `rdr.Close()` closed the underlying reader.

dustin-decker (approved):
Great! CommandContext is useful, we should update all usage to that.

pkg/handlers/handlers.go
```diff
@@ -285,14 +285,9 @@ func HandleFile(
...
-		if copyErr != nil {
-			err = fmt.Errorf("error discarding remaining data: %w", copyErr)
-		}
-		closeErr := rdr.Close()
-		if closeErr != nil {
+		if closeErr := rdr.Close(); closeErr != nil {
 			if err != nil {
-				err = fmt.Errorf("%v; error closing reader: %w", err, closeErr)
+				err = fmt.Errorf("%w; error closing reader: %w", err, closeErr)
```
mcastorina, line 290: Should this be updated to `errors.Join` too? Or keep only one wrapped? imo the double `%w` is misleading/confusing.

mcastorina (approved):
I am liking the clearer API boundaries, this is really nice!

pkg/handlers/handlers.go
ahrav: Yea definitely. My stupid linter kept telling me to use %w, but that isn't a thing. 🤦

PR #3403, opened 2024-10-13
[fix] - Propagate Async File Handling Errors

ahrav:
### Description:
This PR updates error reporting during file processing. Previously, each Handler spawned a goroutine to handle file processing and returned a channel for the caller to collect results. However, this method only returned processed data and failed to propagate errors due to the file processing happening in separate goroutines. As a result, errors were simply logged, and critical errors were returned within the goroutine, leaving the calling function unaware of any issues. This became especially problematic when `HandleFile` was called via `handleBinary`, as the reader passed to `HandleFile` is a pipe. If the caller isn’t informed of an error, it can't properly consume the reader, potentially leaving it open and exhausting resources.

To resolve this, the PR introduces the `DataOrErr` struct, allowing handlers to return both critical and non-critical errors to the caller. This ensures better resource management and prevents resource leaks.
### Checklist:


pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -70,10 +70,10 @@ func asSeeker(r io.Reader) io.Seeker {
 	return seeker
 }
 
-// NewBufferedReaderSeeker creates and initializes a BufferedReadSeeker.
+// NewBufferedReadSeeker creates and initializes a BufferedReadSeeker.
 // It takes an io.Reader and checks if it supports seeking.
 // If the reader supports seeking, it is stored in the seeker field.
-func NewBufferedReaderSeeker(r io.Reader) *BufferedReadSeeker {
```
ahrav, line 76: Rename.

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -82,9 +82,6 @@ func NewBufferedReaderSeeker(r io.Reader) *BufferedReadSeeker {
 	)
 
 	seeker = asSeeker(r)
-	if seeker == nil {
```
ahrav, line 85: This was removed to avoid checking out a buffer eagerly. We already [lazily](https://github.com/trufflesecurity/trufflehog/blob/cf54b71a9410df1960bad536b227958e776ace71/pkg/iobuf/bufferedreaderseeker.go#L123) check out a buffer when needed.

rgmz:
🐇 🕳️

pkg/handlers/ar.go
```diff
@@ -22,56 +22,56 @@ func newARHandler() *arHandler {
...
 
 	if feature.ForceSkipArchives.Load() {
-		close(archiveChan)
-		return archiveChan, nil
+		close(dataOrErrChan)
+		return dataOrErrChan
 	}
 
 	go func() {
-		ctx, cancel := logContext.WithTimeout(ctx, maxTimeout)
```
rosecodym, line 34: Why did you remove this?

pkg/handlers/archive.go
```diff
@@ -44,46 +44,50 @@ func newArchiveHandler() *archiveHandler {
...
-				// Return the panic as an error.
+				var panicErr error
 				if e, ok := r.(error); ok {
-					err = e
+					panicErr = e
 				} else {
-					err = fmt.Errorf("panic occurred: %v", r)
+					panicErr = fmt.Errorf("panic occurred: %v", r)
+				}
+				ctx.Logger().Error(panicErr, "Panic occurred when attempting to open archive")
```
rosecodym, line 70: Why is this error in particular loggable?

pkg/handlers/handlers.go
```diff
@@ -38,7 +39,15 @@ type fileReader struct {
 	*iobuf.BufferedReadSeeker
 }
 
-var ErrEmptyReader = errors.New("reader is empty")
+var (
+	ErrEmptyReader = errors.New("reader is empty")
+
+	// ErrCriticalProcessing indicates a critical error that should halt processing.
```
rosecodym, line 45: "Critical" is one of those words that unfortunately doesn't really convey any information to someone who doesn't already know the domain. Should these errors halt processing of the source? The file? Something else?

pkg/handlers/handlers.go
```diff
@@ -80,10 +89,25 @@ func newMimeTypeReader(r io.Reader) (mimeTypeReader, error) {
...
 
-	fReader.BufferedReadSeeker = iobuf.NewBufferedReaderSeeker(r)
+	// To detect the MIME type of the input data, we need a reader that supports seeking.
+	// This allows us to read the data multiple times if necessary without losing the original position.
+	// We use a BufferedReaderSeeker to wrap the original reader, enabling this functionality.
+	fReader.BufferedReadSeeker = iobuf.NewBufferedReadSeeker(r)
+
+	// If an error occurs during MIME type detection, it is important we close the BufferedReaderSeeker
+	// to release any resources it holds (checked out buffers or temp file).
+	var err error
```
rosecodym, line 103: `err` gets shadowed later in this function. If that's on purpose, I find it very confusing, and would understand much better if you used a different variable name.

pkg/handlers/handlers.go
```diff
@@ -256,14 +291,25 @@ var maxTimeout = time.Duration(60) * time.Second
...
-// extraction or processing. Errors at any stage result in an error return value.
-// Successful handling passes the file content through a channel to be chunked and reported.
-// The function will close the reader when it has consumed all the data.
+// It determines the MIME type of the file,
+// selects the appropriate handler based on this type, and processes the file.
+// This function initializes the handling process and delegates to the specific
+// handler to manage file extraction or processing.
+//
+// The function will return nil (success) in the following cases:
+// - If the reader is empty (ErrEmptyReader)
```
rosecodym, line 300: Will it return `nil`? Or `ErrEmptyReader`?

pkg/handlers/rpm.go
```diff
@@ -22,63 +22,66 @@ func newRPMHandler() *rpmHandler {
...
 
 	if feature.ForceSkipArchives.Load() {
-		close(archiveChan)
-		return archiveChan, nil
+		close(dataOrErrChan)
+		return dataOrErrChan
 	}
 
 	go func() {
-		ctx, cancel := logContext.WithTimeout(ctx, maxTimeout)
```
rosecodym, line 34: same question as above: why the removal?

pkg/handlers/rpm.go
```diff
@@ -22,63 +22,66 @@ func newRPMHandler() *rpmHandler {
...
+					Err:  fmt.Errorf("%w: panic error: %w", ErrCriticalProcessing, panicErr),
 				}
-				ctx.Logger().Error(err, "Panic occurred when reading rpm archive")
 			}
 		}()
 
 		var rpm *rpmutils.Rpm
 		rpm, err = rpmutils.ReadRpm(input)
 		if err != nil {
-			ctx.Logger().Error(err, "error reading RPM")
```
rosecodym, line 62: Are these not real errors?

pkg/handlers/ar.go
ahrav: I thought I left a comment, but it was probably on the closed PR. This was removed because the context timeout is set at the call site, so setting it here has no effect since it inherits the context from `HandleFile`.
[here](https://github.com/trufflesecurity/trufflehog/blob/88b8c862a6ee7cb884257038a06e5f8aaa9d2db8/pkg/handlers/handlers.go#L306)

pkg/handlers/archive.go
ahrav: It shouldn't be. Logging should be handled while consuming from the `dataOrErrChan` at the call-site. Will remove thanks.

pkg/handlers/handlers.go
ahrav: Yea, I agree. I'll update it. 👍

pkg/handlers/handlers.go
ahrav: It should return `nil` if the error from `newFileReader` is `ErrEmptyReader`. Do think think it should return `ErrEmptyReader` instead?

pkg/handlers/rpm.go
ahrav: I understand why you thought I removed them. 😢 In the `defer func()`, we call `.measureLatencyAndHandleErrors`, which uses the shadowed `err`. This eventually gets logged when we consume from `dataOrErrChan`. I can avoid shadowing by explicitly sending `err` to the channel.

pkg/handlers/handlers.go
rosecodym: No, that makes sense, I was just confused because within the context the comment it wasn't clear that an empty reader was signaled (elsewhere) by `ErrEmptyReader`

rosecodym (commented):
I see what looks like three separable PRs here:
* Moving around some metric observations
* Switching to `DataOrErr`
* Surfacing a new archive extraction error

_Could_ they be done separately?

ahrav:
> I see what looks like three separable PRs here:
> 
> * Moving around some metric observations
> * Switching to `DataOrErr`
> * Surfacing a new archive extraction error
> 
> _Could_ they be done separately?

Yep yep, will do. 👍

pkg/handlers/handlers.go
```diff
@@ -117,18 +142,29 @@ func newFileReader(r io.Reader) (fileReader, error) {
...
+// DataOrErr represents a result that can either contain data or an error.
+// The Data field holds the byte slice of data, and the Err field holds any error that occurred.
+// This structure is used to handle asynchronous file processing where each chunk of data
+// or potential error needs to be communicated back to the caller. It allows for
+// efficient streaming of file contents while also providing a way to propagate errors
+// that may occur during the file handling process.
+type DataOrErr struct {
+	Data []byte
+	Err  error
+}
```
mcastorina, line 161: Just noticing that `DataOrErr` pretty much describes Rust's [Result](https://doc.rust-lang.org/std/result/) enum. There are some useful [methods](https://doc.rust-lang.org/std/result/enum.Result.html#implementations) if you wanted to follow their pattern.

ahrav: 🦀 🦀 🦀

Issue #3385, opened 2024-10-09
Include ExtraData from customDetector http response


CameronLonsdale:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description
When validating a customDetector, include the response body as part of the ExtraData
https://github.com/trufflesecurity/trufflehog/blob/23afcd77ee6d368027a3c756a5b6c691ed6a81e0/pkg/custom_detectors/custom_detectors.go#L173

### Preferred Solution
Read the response body and include the body as ExtraData

PR #3411, opened 2024-10-15
Handle custom detector response and include in extra data

kashifkhan0771:
### Description:
This PR handle custom detector response and include it in the ExtraData

### Issue:
https://github.com/trufflesecurity/trufflehog/issues/3385
### Checklist:


zricethezav:
@kashifkhan0771 could you take a look at the test failure:
```
panic: assignment to entry in nil map

goroutine 191 [running]:
github.com/trufflesecurity/trufflehog/v3/pkg/custom_detectors.(*CustomRegexWebhook).createResults(0xc0007f4770, {0x7fd2f8a84330, 0xc0014fc240}, 0xc0014fc330, 0x1, 0xc0014255e0)
	/home/runner/work/trufflehog/trufflehog/pkg/custom_detectors/custom_detectors.go:208 +0xb05
github.com/trufflesecurity/trufflehog/v3/pkg/custom_detectors.(*CustomRegexWebhook).FromData.func1()
	/home/runner/work/trufflehog/trufflehog/pkg/custom_detectors/custom_detectors.go:97 +0x2b
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:78 +0x50
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 172
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:75 +0x96
FAIL	github.com/trufflesecurity/trufflehog/v3/pkg/engine	1.[302](https://github.com/trufflesecurity/trufflehog/actions/runs/11340538845/job/31537134772#step:4:303)s
```

kashifkhan0771:
> @kashifkhan0771 could you take a look at the test failure:
@zricethezav It is fixed now.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -166,14 +169,45 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
+		if resp.StatusCode == http.StatusOK {
+			// mark the result as verified
 			result.Verified = true
+
+			// read the Content-Type header and response body
+			respContentType := resp.Header.Get("Content-Type")
+			body, err := io.ReadAll(resp.Body)
+			if err != nil {
+				return fmt.Errorf("failed to read response body: %v", err)
+			}
```
mcastorina, line 190: This looks like it's returning an error for the `FromData` call, is that right? I think that may not be what we want to do for two reasons:
1. it will break the loop so other configs aren't tested
2. it seems like it would be a non-determinant error (if those exist in custom detectors), and should maybe set the verification error

pkg/custom_detectors/custom_detectors.go
```diff
@@ -263,3 +297,20 @@ func (c *CustomRegexWebhook) Description() string {
...
+	}
+
+	// convert JSON map to a formatted string
+	jsonString, err := json.MarshalIndent(respBody, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal JSON: %v", err)
+	}
+
+	return strings.TrimSpace(string(jsonString)), nil
+}
```
mcastorina, line 316: Why does this function unmarshal then marshal?

pkg/custom_detectors/custom_detectors.go
mcastorina: There other parts that also return an error, this is just one example.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -166,14 +169,45 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
+			_ = resp.Body.Close()
+		}()
+
+		if resp.StatusCode == http.StatusOK {
+			// mark the result as verified
 			result.Verified = true
+
+			// read the Content-Type header and response body
+			respContentType := resp.Header.Get("Content-Type")
+			body, err := io.ReadAll(resp.Body)
```
abmussani, line 187: @mcastorina I could not recall where we had a discussion about PII. I remember, we decided not to include PII in extraData. Will this needed to be care here as well ?

pkg/custom_detectors/custom_detectors.go
kashifkhan0771: Ah, I think I might have over-engineered this, even though @ahrav warned me not to! 😅 I was trying to get the response into a nicely formatted JSON string, but I realize now that it's not really necessary. We can just stick to simple string conversion instead.

pkg/custom_detectors/custom_detectors.go
kashifkhan0771: I think I did that out of habit. I’ve fixed it now, and we’re continuing to the next step instead of returning an error.

pkg/custom_detectors/custom_detectors.go
mcastorina: That's a great point. Custom detector servers are user supplied (or at least user configured), so I don't *think* PII would be an issue here.. @zricethezav what are your thoughts?

pkg/custom_detectors/custom_detectors.go
```diff
@@ -166,14 +168,30 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
+			if err != nil {
+				continue
+			}
+
+			// TODO: handle different content-type responses seperatly when implement custom detector configurations
+			responseStr := string(body)
+
+			// store the processed response in ExtraData
+			result.ExtraData["response"] = responseStr
```
zricethezav, line 193: @kashifkhan0771 **Thought**: storing the entire response could lead to some large outputs. This could be annoying for users who already are using custom detectors. 

One option could be limit the response to 200 characters. I don't feel comfortable merging this PR as is. 

@kashifkhan0771 additionally, I tried spinning up a verification server and testing this myself but found that the only thing that could printed in the output was the name of the customer verifier: `"ExtraData":{"name":"HogTokenDetector"}` when using the `--json` flag. A mock test might help here, although I have a suspicion there might be some other issues considering I'm not able to surface additional `ExtraData` past the name of the custom verifier in the output.

@CameronLonsdale what is your use case for `ExtraData`? How are you using TruffleHog to surface this data (what command and flags)?

CameronLonsdale: For me, I always run trufflehog with --json --only-verified. I wanted to surface data from my custom patterns that are the responses from verifying the credential against the API. So this would typically be a JSON blob which tells me info about the user which corresponds to the token I just verified.

I wrap trufflehog around custom code to sync results to Jira etc so hence why JSON. But I would expect the extraData to show up in CLI output aswell, so truncating it to be readable at the terminal makes sense.

kashifkhan0771: @zricethezav @CameronLonsdale I added the limit of 200 chars for response. I am not sure how to test the custom detector. If anyone of you can guide me how to do it. I can test locally to understand why it is not printing the extra data in output.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -101,10 +102,6 @@ func (c *CustomRegexWebhook) FromData(ctx context.Context, verify bool, data []b
 	close(resultsCh)
 
 	for result := range resultsCh {
-		// NOTE: I don't believe this is being set anywhere else, hence the map assignment.
```
kashifkhan0771, line 104: This was the reason for response result not being printed 😛 @zricethezav

pkg/custom_detectors/custom_detectors.go
kashifkhan0771: Thank you, @mcastorina, for your guidance on running the custom detector. I successfully tested it and identified the issue that was preventing the results from printing. I've fixed it now. Please review my changes.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -101,10 +102,6 @@ func (c *CustomRegexWebhook) FromData(ctx context.Context, verify bool, data []b
...
-		result.ExtraData = map[string]string{
-			"name": c.GetName(),
-		}
```
kashifkhan0771, line 107: This was the reason the response result wasn't being printed! 😛 @zricethezav

kashifkhan0771: Let me know if the name is something we want to print. I can make the change to append it.

zricethezav (approved):
lgtm

pkg/custom_detectors/custom_detectors.go
ahrav: @kashifkhan0771 not sure if you saw the issue that came in which @rgmz mentioned, but do you think you could include the name back. Thanks.
https://github.com/trufflesecurity/trufflehog/issues/3567

kashifkhan0771: Yeah sure

kashifkhan0771: Added back: https://github.com/trufflesecurity/trufflehog/pull/3569

Issue #3394, opened 2024-10-10
git --since-commit does not notify on invalid inputs


JonZeolla:
### TruffleHog Version

trufflehog 3.82.7

### Trace Output

https://gist.github.com/JonZeolla/bb270c47471284baa5a9572f0c473234

### Expected Behavior

I expect to be notified of an invalid `--since-commit` input

### Actual Behavior

It finishes successfully with no chunks or bytes.

### Steps to Reproduce

 1. `cd` to any git repository
 2. run `trufflehog git --since-commit=THISISPROBABLYINVALID file://.`

## Environment

 * OS: macOS
 * Version 15.0.1 (24A348)

sahil9001:
@JonZeolla I have raised a PR, please check.

PR #3413, opened 2024-10-15
feat: added check for valid git commit and warning message

sahil9001:
### Description:
This PR addresses this issue #3394 , wherein we check the validity of git commit and throw the invalid commit message if commit doesn't match the GitHub SHA standard.

Screenshot 2024-10-15 at 8 56 34 PM

### Checklist:


main.go
```diff
@@ -312,6 +312,18 @@ func main() {
 	}
 }
 
+func isValidCommit(commit string) bool {
+	if len(commit) != 40 {
+		return false
+	}
```
rgmz, line 318: I don't believe it's necessary to provide the full SHA. 

> Git is smart enough to figure out what commit you’re referring to if you provide the first few characters of the SHA-1 hash, as long as that partial hash is at least four characters long and unambiguous; that is, no other object in the object database can have a hash that begins with the same prefix.

https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_short_sha_1

sahil9001: Can you check now, I have fixed the issue, also how to link the issue with the PR? I can't see the option to do so

sahil9001: @rgmz

rgmz: Doing basic validation is helpful.

However, I think the only way to properly satisfy #3394 is to dispatch a command to Git. (e.g., https://stackoverflow.com/questions/18515488/how-to-check-if-the-commit-exists-in-a-git-repository-by-its-sha-1)

sahil9001: Yeah, you are right, I have fixed by dispatching this command from the function.

ahrav:
Thanks @sahil9001, sorry for the delayed review here.

PR #3421, opened 2024-10-15
Fix RabbitMQ detector not respecting timeout

rgmz:
### Description:

This adds a timeout to the RabbitMQ detector.
```
2024-10-15T18:28:56-04:00	error	trufflehog	a detector ignored the context timeout	{"detector_worker_id": "4ZclX", "detector": {"type":"RabbitMQ"}, "timeout": 10}
```

### Checklist:

### DIFF
--- a/pkg/detectors/rabbitmq/rabbitmq.go
+++ b/pkg/detectors/rabbitmq/rabbitmq.go
@@ -2,12 +2,13 @@ package rabbitmq
 
 import (
 	"context"
+	"net"
 	"net/url"
 	"strings"
-
-	regexp "github.com/wasilibs/go-re2"
+	"time"
 
 	amqp "github.com/rabbitmq/amqp091-go"
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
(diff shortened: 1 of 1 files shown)


pkg/detectors/rabbitmq/rabbitmq.go
```diff
@@ -51,37 +59,49 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		Dial: func(network, addr string) (net.Conn, error) {
+			return net.DialTimeout(network, addr, 10*time.Second)
+		},
+	})
+	defer func() {
+		if conn != nil {
+			_ = conn.Close()
+		}
+	}()
+	return err == nil, err
```
rgmz, line 102: Future work would be to categorize the errors provided by this, differentiating between valid and invalid.

Such as:
`Verification issue: lookup xcws0001: no such host`
`Verification issue: i/o timeout`

abmussani (commented):
@rgmz thanks for the fix. Instead of 10s constant timeout, Will it be a good idea to calculate the remaining time using context.Deadline() ? and use default 10s if there is no deadline define ?

ahrav (approved):

PR #3422, opened 2024-10-16
increase timeout to 30s

ahrav:
### Description:
We're seeing frequent context timeout errors during `GetObject` API calls. This PR increases the default timeout from 5 to 30 seconds.

Error:
```
RequestCanceled: request context canceled
caused by: context deadline exceeded
```
### Checklist:


mcastorina (approved):

rosecodym:
Are we sure that the previous timeout is wrong? My concern is that these timeouts were caused by something other than a bunch of large objects, so if we do this then we won't scan any more data but each scan will take six times as long.

ahrav:
> Are we sure that the previous timeout is wrong? My concern is that these timeouts were caused by something other than a bunch of large objects, so if we do this then we won't scan any more data but each scan will take six times as long.

According to the documentation for the library we're using:
```
// GetObject API operation for Amazon Simple Storage Service.
//
// Retrieves an object from Amazon S3.
```

I _believe_ the timeout was initially set under the assumption that we were only retrieving metadata, but the docs confirm it retrieves the entire object. I'm not certain what else might be causing the context deadline.

ahrav:
Adjusted the logger to use our ctx. I also included the object size in the logger kvp.

pkg/sources/s3/s3.go
```diff
@@ -131,6 +131,7 @@ func (s *Source) setMaxObjectSize(maxObjectSize int64) {
 func (s *Source) newClient(region, roleArn string) (*s3.S3, error) {
 	cfg := aws.NewConfig()
 	cfg.CredentialsChainVerboseErrors = aws.Bool(true)
+	cfg.LogLevel = aws.LogLevel(aws.LogDebugWithRequestErrors)
```
mcastorina, line 134: Was this just for testing or are we keeping it in?

pkg/sources/s3/s3.go
```diff
@@ -271,46 +272,63 @@ func (s *Source) getRegionalClientForBucket(ctx context.Context, defaultRegionCl
...
+	page *s3.ListObjectsV2Output,
+	errorCount *sync.Map,
+	pageNumber int,
+	objectCount *uint64,
+) {
+	for _, o := range page.Contents {
+		if o == nil {
+			continue
 		}
+		obj := *o
```
mcastorina, line 289: Why do we need to make a shallow copy here?

mcastorina (approved):
This seems good, just had a few non-blocking questions.

One other thought: can `log` be removed from the struct now, or is it still referenced in other parts? Not saying it should be removed in this PR, but I saw you corrected many of the log calls to use the passed `ctx`.

ahrav:
> This seems good, just had a few non-blocking questions.
> 
> One other thought: can `log` be removed from the struct now, or is it still referenced in other parts? Not saying it should be removed in this PR, but I saw you corrected many of the log calls to use the passed `ctx`.

Unfortunately, it's used across the rest of the source 😢 I can clean it up in a separate PR.

pkg/sources/s3/s3.go
ahrav: Oh whoops, I forgot to remove this. I was testing something. 😅 Thanks.

pkg/sources/s3/s3.go
ahrav: I thought it would be useful even for non-debugging purposes since it would only log during errors. What do you think?

mcastorina: Yeah, makes sense to me.

PR #3425, opened 2024-10-16
Ignore `https:` as false positive for slackwebhook detector

kashifkhan0771:
### Description:
This PR resolves the issue where unverified Slack webhook secrets were incorrectly ignored due to the presence of `https:` being flagged as a false positive. The updated logic ensures that https: is no longer considered a false positive, allowing Slack webhook secrets to be properly detected and processed.

### Output of unverified slackwebhook:
slackwebhook-falsepostive

### Ticket:
OSS-36

### Checklist:

### DIFF
--- a/pkg/detectors/slackwebhook/slackwebhook.go
+++ b/pkg/detectors/slackwebhook/slackwebhook.go
@@ -20,6 +20,7 @@ type Scanner struct {
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
 
 var (
 	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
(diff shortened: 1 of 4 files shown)


dustin-decker (approved):

rosecodym:
Thanks for this!

pkg/detectors/slackwebhook/slackwebhook.go
```diff
@@ -115,3 +116,14 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
 }
+
+func (s Scanner) IsFalsePositive(result detectors.Result) (bool, string) {
+	// ignore "https:" as a false positive for slack webhook detector
+	if strings.Contains(string(result.Raw), "https:") {
+		return false, ""
+	}
+
+	// back to the default false positive checks
+	return detectors.IsKnownFalsePositive(string(result.Raw), detectors.DefaultFalsePositives, true)
```
rgmz, line 127: When would this trigger? Won't the result always contain `https:`?

Personally, I think `https:` should be removed from the wordlist.

kashifkhan0771: Hmm, that can be done as well. For now it was for only one detector so I went with this approach.

rgmz: Sorry, the second sentence is a bit of a tangent.

I don't think the second part of the function will ever be reached.

kashifkhan0771: You’re absolutely right. I’ll simplify the function in next enhancement and remove the second part as it's unnecessary in this case. Appreciate the catch!

Issue #1458, opened 2023-07-06
JDBC regex captures invalid url data


rgmz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### TruffleHog Version

3.43.0

### Trace Output

```sh
Found unverified result 🐷🔑❓
Detector Type: JDBC
Decoder Type: PLAIN
Raw result: jdbc:postgresql://localhost:5432/postgres</jdbc-url>
Commit: fbc14303ffbf8fb1c2c1914e8dda7d0121633aca
Email: John Smith <john.smith@example.com>
File: .idea/dataSources.xml
Line: 8
Link: https://github.example.com/org/repo/blob/fbc14303ffbf8fb1c2c1914e8dda7d0121633aca/.idea/dataSources.xml
Repository: https://github.example.com/org/repo.git
Timestamp: 2021-09-04 15:29:17 +0000
```

### Expected Behavior

The JDBC detector should only fetch valid parts of a JDBC url:

```
jdbc:postgresql://localhost:5432/postgres
```

### Actual Behavior

The JDBC detector includes the xml end tag, resulting in an invalid URL that will fail to verify.
```
<jdbc-url>**jdbc:postgresql://localhost:5432/postgres</jdbc-url>**
```

It _might_ be possible to write a simple regular expression that only finds compliant URLs, however, I'd have to look into that further.

### Steps to Reproduce

1. Create a file like below (`.idea/dataSources.xml`):
  ```xml
  <?xml version="1.0" encoding="UTF-8"?>
  <project version="4">
    <component name="DataSourceManagerImpl" format="xml" multifile-model="true">
      <data-source source="LOCAL" name="PostgreSQL - postgres@localhost" uuid="18f0f64d-b804-471d-9351-e98a67c8389f">
        <driver-ref>postgresql</driver-ref>
        <synchronize>true</synchronize>
        <jdbc-driver>org.postgresql.Driver</jdbc-driver>
        <jdbc-url>jdbc:postgresql://localhost:5432/postgres</jdbc-url>
        <working-dir>$ProjectFileDir$</working-dir>
      </data-source>
[...]

PR #3429, opened 2024-10-16
Ignore invalid SQL Server URLs

rgmz:
### Description:
Reduce the number of false positives from invalid URLs. Related to https://github.com/trufflesecurity/trufflehog/issues/1458#issuecomment-2416783812.

e.g.,
```
Found unverified result 🐷🔑❓
Verification issue: parse "sqlserver://sa:@{sql:1433?database=master&connection+timeout=5": invalid character "{" in host name
Detector Type: JDBC
Decoder Type: PLAIN
Raw result: jdbc:sqlserver://{sql
File: /tmp/clusters_client_example_test.go
Line: 203
```
https://github.com/Azure/azure-sdk-for-go/blob/7ea57e99feeadc51cb635d0c8f05be18cd9b2809/sdk/resourcemanager/hdinsight/armhdinsight/clusters_client_example_test.go#L1558

### Checklist:


pkg/detectors/jdbc/jdbc.go
```diff
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	logContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
```
ahrav, line 8: nit: Could we move this import below the std library pkgs.

ahrav (approved):
Thanks, lgtm! ❤️

pkg/detectors/jdbc/jdbc.go
rgmz: Done.

PR #3437, opened 2024-10-16
fix: added correct verification endpoint & validation logic for alegra

### LINKED ISSUE #3436: Issue with correct endpoint and secret validation for alegra

### Expected Behavior

Use the correct validation logic for secret — `\b([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-z]+)\b`
Screenshot 2024-10-21 at 2 16 36 PM

The endpoint should point to `/api/v1/users/self` to get the details for the user in current session.
Reference : https://developer.alegra.com/reference/get_users-self

### Actual Behavior

Currently, the key regex for email validation is wrong : `\b([a-zA-Z0-9\.\-\@]{25,30})\b`, 
Screenshot 2024-10-21 at 2 15 50 PM

Currently the endpoint points to `https://api.alegra.com/api/v1/users`, which doesn't seem to be found in the documentation.

## Environment
 * OS: [e.g. iOS]
 * Version [e.g. 22]

### References

* #0000

sahil9001:
### Description:
Fixes #3436 

Screenshot 2024-10-21 at 3 22 10 PM

### Checklist:

### DIFF
--- a/pkg/detectors/alegra/alegra.go
+++ b/pkg/detectors/alegra/alegra.go
@@ -60,7 +60,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 
 			if verify {
-				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.alegra.com/api/v1/users", nil)
+				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.alegra.com/api/v1/users/self", nil)
 				if err != nil {
 					continue
 				}
--- a/pkg/detectors/alegra/alegra.go
+++ b/pkg/detectors/alegra/alegra.go
@@ -24,7 +24,7 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-z0-9-]{20})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9\.\-\@]{25,30})\b`)
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-z]+)\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/alegra/alegra.go
+++ b/pkg/detectors/alegra/alegra.go
@@ -2,6 +2,8 @@ package alegra
 
 import (
 	"context"
+	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
@@ -13,18 +15,17 @@ import (
 )
 
 type Scanner struct {
-	detectors.DefaultMultiPartCredentialProvider
+	client *http.Client
 }
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
-
+	defaultClient = common.SaneHttpClient()
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-z0-9-]{20})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-z]+)\b`)
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + common.EmailPattern)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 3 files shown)


sahil9001:
@rgmz @zricethezav

pkg/detectors/alegra/alegra.go
```diff
@@ -60,7 +60,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 
 			if verify {
-				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.alegra.com/api/v1/users", nil)
+				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.alegra.com/api/v1/users/self", nil)
```
kashifkhan0771, line 63: I created an account on [Alegra](https://mi.alegra.com) and generated a token. Initially, I tested the `/users` API endpoint using an API client, and it worked as expected, returning the same response as `/users/self`. However, the Alegra documentation only mentions the `/users/self` endpoint, and there’s no reference to `/users`, so I'm unsure why it’s functioning similarly.

Additionally, I noticed an issue with the detector failing to identify my user ID, which I believe is always the email used during signup. The regex used for detecting IDs has a length constraint of 25 to 30 characters. However, I was able to create an account with an email shorter than this limit, and despite using that email (as my username) along with my token to successfully call the API via the API client, the detector did not pick it up.

I suggest testing and adjusting the regex for detecting user IDs to ensure it covers valid email lengths.

As for the API behavior, I’ll leave that to @zricethezav and @abmussani. Both `/users` and `/users/self` worked for me, though only the `/users/self` endpoint is mentioned in the documentation.

kashifkhan0771 (changes requested):

pkg/detectors/alegra/alegra.go
sahil9001: Thanks for the detailed review, @kashifkhan0771 !

I have expanded the email regex to include all case for any possible email.

kashifkhan0771: I would feel more confident approving this if you could also test the current email regex to confirm that it isn't working correctly(I tested it but a second round of testing from the PR owner will be good I believe. You can create an account and test this email regex change). A screenshot or any form of documentation from your testing would be sufficient.

sahil9001: I have added the testing screenshot, please check

kashifkhan0771: New Changes are picking up the key and ID correctly now and verifying. Wait for @zricethezav and @abmussani to respond about the API change.

sahil9001: Can you apply the label for Hacktoberfest @kashifkhan0771?

kashifkhan0771: Once the PR is approved. We can add it 😃 - Once again thanks for the all the fixes and contributions!

pkg/detectors/alegra/alegra.go
```diff
@@ -24,7 +24,7 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-z0-9-]{20})\b`)
-	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9\.\-\@]{25,30})\b`)
+	idPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"alegra"}) + `\b([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-z]+)\b`)
```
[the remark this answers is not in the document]
zricethezav: https://github.com/trufflesecurity/trufflehog/blob/86d2c6ded3069d9527b83d7068ac756a91514866/pkg/common/patterns.go#L10 Looks like we already have one, although I'd be curious on your thoughts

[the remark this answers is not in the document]
sahil9001: Should I use this, I actually picked this pattern from one of the existing detectors only.

https://github.com/trufflesecurity/trufflehog/blob/3ab60865c84ee384a099f4e9b6001f6a50a7d1a2/pkg/detectors/zipbooks/zipbooks.go#L26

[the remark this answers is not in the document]
zricethezav: Yes, please use `const EmailPattern` for consistency. We are planning to standardize across all the detectors.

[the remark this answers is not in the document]
sahil9001: @zricethezav I have removed the pattern tests, as that was not the ideal pattern for new detectors that we followed, moved the integration tests to main tests file now.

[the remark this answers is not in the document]
sahil9001: Screenshot 2024-10-21 at 3 22 10 PM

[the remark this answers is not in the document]
sahil9001: @rgmz updated this to alchemy pattern too.

sahil9001:
> The changes are working fine now. I tested locally. One thing I noticed is if we add an email like `alegra_email=abc@example.com` it actually picks the whole string including `alegra_email=` 😄 and results in unverified result. If we can somehow avoid that it will be more better. Anyway approving the PR as it is working as expected now.

Thanks @kashifkhan0771 , can you please re-run the workflow again, for some reason it is failing on GitHub, locally works fine.

Also I am thinking of a case where we actually have an email like `alegra_email=abc@example.com`, as the current email regex is too accurate for this.

sahil9001:
Can we merge this @zricethezav ?

Issue #3446, opened 2024-10-17
Issue with correct verification endpoint for autoklose


sahil9001:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

### Expected Behavior
Should be using `/me` api endpoint for verification.

Reference : https://api.aklab.xyz/#09ae29ca-c105-4a08-bcea-45bf11cedd24
<img width="1437" alt="Screenshot 2024-10-17 at 11 08 39 AM" src="https://github.com/user-attachments/assets/9ab260c6-6d7e-4251-b660-644e4b69c231">

### Actual Behavior
The current endpoint works, but isn't recommended for verification.

## Environment
 * OS: [e.g. iOS]
 * Version [e.g. 22]

### References

* #0000

PR #3447, opened 2024-10-17
fix: fixed autoklose verification endpoint

sahil9001:
### Description:
Relates to #3446 

### Checklist:

### DIFF
--- a/pkg/detectors/autoklose/autoklose.go
+++ b/pkg/detectors/autoklose/autoklose.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
 	"io"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
@@ -50,7 +51,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		}
 
 		if verify {
-			req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://api.autoklose.com/api/campaigns/?api_token=%s", resMatch), nil)
+			req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://api.autoklose.com/api/me/?api_token=%s", resMatch), nil)
 			if err != nil {
 				continue
 			}


sahil9001:
@rgmz @zricethezav

kashifkhan0771 (changes requested):

pkg/detectors/autoklose/autoklose.go
```diff
@@ -63,7 +66,15 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				defer res.Body.Close()
 				if res.StatusCode >= 200 && res.StatusCode < 300 {
```
kashifkhan0771, line 67: The API returns a 200 status code for successful calls, which we can explicitly handle in this case.

pkg/detectors/autoklose/autoklose.go
```diff
@@ -63,7 +66,15 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 					if json.Valid(bodyBytes) {
-						s1.Verified = true
+						var responseBody map[string]interface{}
+						if err := json.Unmarshal(bodyBytes, &responseBody); err == nil {
+							s1.Verified = true
```
kashifkhan0771, line 71: Result should be marked as verified as soon as we get expected status code. Extra Data related things should be handled afterwards. In this case if for some reason un-marshal fails the result will be marked as unverified even though we have received a 200.

kashifkhan0771 (changes requested):
Minor Feedback! Rest looks ok to me.

pkg/detectors/autoklose/autoklose.go
sahil9001: Hmm, that seems right. I observed that API returns 200 but HTML response if the token is expired, so I have set that condition that if the code is not able to parse the response as JSON, it is probably invalid.

pkg/detectors/autoklose/autoklose.go
sahil9001: done

pkg/detectors/autoklose/autoklose.go
kashifkhan0771: Since we’ve added the content-type header as application/json, the response will be in JSON format. The logic should be: if we receive a 200 status code, we should unmarshal the response and capture the email. If we receive a 401, it indicates the token is expired or invalid, and we should mark it as unverified. Any status code other than 200 or 401 should be handled separately.

sahil9001: nope, even after setting `Content-Type: application/json` we are getting HTML response, the server doesn't respect client headers.

Screenshot 2024-10-17 at 4 32 05 PM

kashifkhan0771: We need to set `Accept` header as `application/json` in order to tell server we are expecting a JSON response.

sahil9001: Can you check now if it is correct @kashifkhan0771 ?

pkg/detectors/autoklose/autoklose.go
```diff
@@ -50,23 +51,38 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					var responseBody map[string]interface{}
+					if err := json.Unmarshal(bodyBytes, &responseBody); err == nil {
+						if email, ok := responseBody["email"].(string); ok {
+							s1.ExtraData = map[string]string{
+								"email": email,
+							}
+						}
 					}
+				} else {
+					s1.Verified = false
```
kashifkhan0771, line 85: Verified by default is unverified so not required to mark it explicitly. Following code should work fine. You can cross check one time. 
```
res, err := client.Do(req)
if err == nil {
    defer func() {
	_ = res.Body.Close()
    }()

if res.StatusCode == http.StatusOK {
    s1.Verified = true
    bodyBytes, err := io.ReadAll(res.Body)
    if err != nil {
        continue
    }

var responseBody map[string]interface{}
if err := json.Unmarshal(bodyBytes, &responseBody); err == nil {
    if email, ok := responseBody["email"].(string); ok {
	    "email": email,
	    }
	}
    }
}
} else {
    s1.SetVerificationError(err, resMatch)
}
```

kashifkhan0771 (changes requested):

pkg/detectors/autoklose/autoklose.go
sahil9001: Thanks made the change

kashifkhan0771 (approved):
Thanks for the contribution @sahil9001 - Changes LGTM!

sahil9001:
This counts eligible for Hacktoberfest, right? Can you add label is possible?

sahil9001:
@kashifkhan0771 Thanks for this, can you please review other PRs as well, and add the label, actually I don't know if I am doing this right or not. Your overview will give me an assurance if I am on the right path (eligibility)

zricethezav:
Nice work @sahil9001. Thanks for the review @kashifkhan0771. Merging

PR #3455, opened 2024-10-17
Revert "Compress release with UPX"

dustin-decker:
Build failed, reverting so we can re-release

Reverts trufflesecurity/trufflehog#3445


dustin-decker:
I think something’s different between goreleaser-pro and what we have in OSS that cause the build failure

ahrav (approved):

PR #3460, opened 2024-10-17
[fix] - Context timeout

ahrav:
### Description:
This PR isolates the context used for object retrieval, ensuring the timeout is applied only to the `GetObjectWithContext` call. Previously, the context was propagated to `HandleFile`, causing operations to fail if a file wasn't processed within 5 seconds.

### Checklist:


rosecodym (approved):
🤞

PR #3471, opened 2024-10-18
Addeded 403 account block status code handling for gitlab

kashifkhan0771:
### Description:
This PR handle 403 status code for Gitlab detector and verify if the user account is blocked or not and add that information in the Extra Data

### JIRA Ticket:
CSM-637

### Checklist:

### DIFF
--- a/pkg/detectors/gitlab/v1/gitlab.go
+++ b/pkg/detectors/gitlab/v1/gitlab.go
@@ -20,6 +20,10 @@ type Scanner struct {
 	detectors.EndpointSetter
 }
 
+type gitlabMessage struct {
+	Message string `json:"message"`
+}
+
 // Ensure the Scanner satisfies the interfaces at compile time.
 var (
 	_ detectors.Detector           = (*Scanner)(nil)
@@ -34,6 +38,8 @@ func (Scanner) CloudEndpoint() string { return "https://gitlab.com" }
 var (
 	defaultClient = common.SaneHttpClient()
 	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"gitlab"}) + `\b([a-zA-Z0-9\-=_]{20,22})\b`)
+
+	blockedUserMessage = "403 Forbidden - Your account has been blocked"
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
@@ -60,6 +66,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_Gitlab,
 			Raw:          []byte(resMatch),
+			ExtraData:    map[string]string{},
 		}
 		s1.ExtraData = map[string]string{
 			"rotation_guide": "https://howtorotate.com/docs/tutorials/gitlab/",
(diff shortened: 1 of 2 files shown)


pkg/detectors/gitlab/v1/gitlab.go
```diff
@@ -98,33 +110,42 @@ func (s Scanner) verifyGitlab(ctx context.Context, resMatch string) (bool, error
...
+			return json.Valid(body), false, nil
 		case http.StatusForbidden:
+			// check if the user account is blocked or not
+			var apiResp gitlabMessage
+			if err := json.Unmarshal(body, &apiResp); err == nil {
+				if apiResp.Message == blockedUserMessage {
+					return true, true, nil
+				}
+			}
+
```
abmussani, line 136: What if we just find that string in the bodyString ? I would love to see @ahrav's input in this matter.

pkg/detectors/gitlab/v1/gitlab.go
```diff
@@ -60,15 +66,21 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+			ExtraData:    map[string]string{},
 		}
 		s1.ExtraData = map[string]string{
 			"rotation_guide": "https://howtorotate.com/docs/tutorials/gitlab/",
 			"version":        fmt.Sprintf("%d", s.Version()),
 		}
 
 		if verify {
-			isVerified, verificationErr := s.verifyGitlab(ctx, resMatch)
+			isVerified, isUserBlocked, verificationErr := s.verifyGitlab(ctx, resMatch)
```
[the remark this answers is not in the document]
ahrav: I think it makes sense to prioritize consistency across the detectors, especially since we have many of them. One other aspect to consider is that returning multiple values with the same type can cause ambiguity. It's easy to mix up with bool represents what. While named return types can reduce ambiguity, they can also introduce inconsistency if not applied uniformly across the codebase.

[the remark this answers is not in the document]
kashifkhan0771: I actually thought of that first but dropped the idea thinking that extraData will be returned for just one case. Your comments make sense. I will make that change now.

pkg/detectors/gitlab/v1/gitlab.go
ahrav: String matching on errors in general makes me uneasy, but if we’re confident in the response structure and error location, it should work. Alternatively, we could use bytes.Contains to check the entire response after consuming the body. I’m not sure about the performance difference, but I suspect bytes.Contains might be slightly faster. JSON decoding in Go is typically slow due to the standard library’s use of reflection. That said, I’d go with whichever option is clearer to everyone.

pkg/detectors/gitlab/v1/gitlab.go
```diff
@@ -60,15 +66,21 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 			s1.Verified = isVerified
+			// if user account is blocked, add this information in the extra data
+			if isUserBlocked {
+				s1.ExtraData["User Account Blocked"] = "Yes"
```
rgmz, line 81: This is a bit wordy. I think just "blocked = true" would be fine. 
https://docs.gitlab.com/ee/api/users.html#as-a-regular-user

e.g., https://github.com/trufflesecurity/trufflehog/blob/871a2b0f06f602899f8c928ce19d2737a3758e6c/pkg/detectors/github/v1/github_old.go#L152-L154

rgmz (commented):
What about for [V2 tokens](https://github.com/trufflesecurity/trufflehog/blob/871a2b0f06f602899f8c928ce19d2737a3758e6c/pkg/detectors/gitlab/v2/gitlab_v2.go)? Maybe it's worth consolidating the logic like GitHub (linked below)?

ahrav (approved):
Nice!

pkg/detectors/gitlab/v1/gitlab.go
kashifkhan0771: Removed unmarshalling logic

PR #3472, opened 2024-10-18
[Fix] - buffer leak

ahrav:
### Description:
The original `NewBufferedReadSeeker` constructor checked out a buffer from the pool, which could result in an orphaned buffer if MIME type detection failed during `newFileReader`. This PR addresses the issue by lazily initializing the buffer for `BufferedReadSeeker` only when needed.

Screenshot 2024-10-18 at 12 15 49 PM

### Checklist:

### DIFF
--- a/pkg/iobuf/bufferedreaderseeker.go
+++ b/pkg/iobuf/bufferedreaderseeker.go
@@ -70,10 +70,10 @@ func asSeeker(r io.Reader) io.Seeker {
 	return seeker
 }
 
-// NewBufferedReaderSeeker creates and initializes a BufferedReadSeeker.
+// NewBufferedReadSeeker creates and initializes a BufferedReadSeeker.
 // It takes an io.Reader and checks if it supports seeking.
 // If the reader supports seeking, it is stored in the seeker field.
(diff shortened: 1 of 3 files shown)


pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -70,10 +70,10 @@ func asSeeker(r io.Reader) io.Seeker {
 	return seeker
 }
 
-// NewBufferedReaderSeeker creates and initializes a BufferedReadSeeker.
+// NewBufferedReadSeeker creates and initializes a BufferedReadSeeker.
 // It takes an io.Reader and checks if it supports seeking.
 // If the reader supports seeking, it is stored in the seeker field.
-func NewBufferedReaderSeeker(r io.Reader) *BufferedReadSeeker {
```
ahrav, line 76: Renamed to match the struct name.

mcastorina (commented):
The actual change seems innocuous, but I'm not understanding something because in my head a leak isn't fixed by lazy allocation. That just slows it down.

ahrav:
superseded by [3403](https://github.com/trufflesecurity/trufflehog/pull/3403)

PR #3478, opened 2024-10-19
fix: fixed verifcation pattern logic for `bulksms`

### LINKED ISSUE #3477: Issue with verification logic & tests for `bulksms`

### Expected Behavior

The pattern validation should detect other characters as well, correct regex should be : `[a-zA-Z0-9!@#$%^&*()]{29}`

Screenshot 2024-10-19 at 2 04 03 PM

### Actual Behavior
Currently, the pattern detection is invalid which fails to detect the correct API pattern. 

Tests failing because `got[I].RawV2` is not set to `nil`.

## Environment
 * OS: [e.g. iOS]
 * Version [e.g. 22]

### References

* #0000

sahil9001:
### Description:
Fixes #3477 

Screenshot 2024-10-19 at 1 51 50 PM

### Checklist:


pkg/detectors/bulksms/bulksms.go
```diff
@@ -24,7 +25,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"bulksms"}) + `\b([a-fA-Z0-9*]{29})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"bulksms"}) + `\b([a-zA-Z0-9!@#$%^&*()]{29})\b`)
```
kashifkhan0771, line 28: I generated over 10 tokens, and it appears the new regex is correct. So far, I’ve only received tokens containing the special characters `!_*#`, though I believe other special characters are possible.

**Potential Improvements for the Detector:**

- Instead of manually encoding the key and ID for basic authentication, we can leverage the SetBasicAuth function to simplify the process.
- Currently, we consider the result verified if the API returns a status code between 200 and 300. We could improve this by being more specific. Let's add the comment to reference the API docs as well.
- Based on my observations, both the ID and key are unique for each token generation. If a key is verified successfully with any ID, we can skip the remaining verification attempts for that key and proceed to the next one.

sahil9001: Thanks for the review @kashifkhan0771 , I have addressed the changes

pkg/detectors/bulksms/bulksms.go
```diff
@@ -59,25 +59,30 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			}
 
 			if verify {
-				data := fmt.Sprintf("%s:%s", resIdMatch, resMatch)
-				sEnc := b64.StdEncoding.EncodeToString([]byte(data))
+				// data := fmt.Sprintf("%s:%s", resIdMatch, resMatch)
```
kashifkhan0771, line 62: Please remove these commented lines. Also I guess you missed the third point from my last review.

kashifkhan0771 (changes requested):

pkg/detectors/bulksms/bulksms.go
sahil9001: Sure, I am unclear about the last point, are you talking about the tests? Can you please elaborate more?

kashifkhan0771: The secret in this context (Basic Auth token) is a combination of two components: an ID and a Key. Both the ID and Key are unique for each secret. Currently, we loop through each ID and attempt to verify it against all the available keys. 

For example, if we find 3 IDs and 3 Keys, we start by verifying the first ID with all the keys. If, on the first attempt, we successfully verify that the first ID and the first Key form a valid Basic Auth token, there is no need to check the remaining keys for that ID since it has already been verified. We can then move on to the next ID.

To further optimize this process, it would be beneficial to avoid comparing the subsequent IDs with any keys that have already been verified with previous IDs. This would reduce the number of unnecessary comparisons and improve efficiency by reducing the total number of combinations checked.

kashifkhan0771: The secret (Basic Auth token) is a combination of two components here, an ID and a Key. Both the ID and Key are unique for each secret as I observed after generating multiple tokens. Currently, we loop through each ID and attempt to verify it against all the available keys. [O(nxm)]

If we find 3 IDs and 3 Keys, we start by verifying the first ID with all the keys. If, on the first attempt, we successfully verify that the first ID and the first Key form a valid Basic Auth token, there is no need to check the remaining keys for that ID since it has already been verified. We can then move on to the next ID.

To further optimize this process, it would be more better to avoid comparing the remaining IDs with any keys that have already been verified with previous IDs. This would reduce the number of unnecessary comparisons and improve efficiency.

sahil9001: Thanks, I have created a map of string -> boolean to mark the keys if a combination matches.

pkg/detectors/bulksms/bulksms.go
```diff
@@ -38,46 +37,66 @@ func (s Scanner) Keywords() []string {
...
 			}
-			resIdMatch := strings.TrimSpace(idmatch[1])
 
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_Bulksms,
-				Raw:          []byte(resMatch),
-				RawV2:        []byte(resMatch + resIdMatch),
+				Raw:          []byte(resKeyMatch),
+				RawV2:        []byte(resKeyMatch + resIDMatch),
 			}
```
kashifkhan0771, line 70: Few suggestions:

- Can remove trim space funcs as there is no space in regex
- Instead of creating `map[string]bool`, we can define `map[string]struct{}` and filter unique values before processing the matches. This will also keep it same as other detectors.

Suggested Code: (This is a suggestion code, you can run it locally and verify and make changes if required)

```
	var uniqueIds = make(map[string]struct{})
	var uniqueKeys = make(map[string]struct{})

	for _, match := range idPat.FindAllStringSubmatch(dataStr, -1) {
		uniqueIds[match[1]] = struct{}{}
	}

	for _, match := range keyPat.FindAllStringSubmatch(dataStr, -1) {
		uniqueKeys[match[1]] = struct{}{}
	}

	for id := range uniqueIds {
		for key := range uniqueKeys {
			s1 := detectors.Result{
				DetectorType: detectorspb.DetectorType_Bulksms,
				Raw:          []byte(key),
				RawV2:        []byte(key + id),
			}

			if verify {
				req, err := http.NewRequestWithContext(ctx, "GET", "https://api.bulksms.com/v1/messages", nil)
				if err != nil {
					continue
				}
				req.SetBasicAuth(id, key)
				res, err := client.Do(req)
				if err == nil {
					defer func() {
						_ = res.Body.Close()
					}()

					if res.StatusCode == http.StatusOK {
						s1.Verified = true
						results = append(results, s1)
						// move to next key, by skipping remaining id's
						break
					}
				} else {
					s1.SetVerificationError(err, key)
				}
			}

			results = append(results, s1)
		}
	}
```

kashifkhan0771 (changes requested):

pkg/detectors/bulksms/bulksms.go
sahil9001: Thanks @kashifkhan0771 , this method is better, I agree. I have made the changes and tested it too, please check.

kashifkhan0771 (approved):
LGTM!

PR #3489, opened 2024-10-21
Remove analyzer protos

bill-rich:
Only used for tracking analyzer types


mcastorina (approved):
lgtm! A few thoughts:

* The analyzer types / mappings could probably be generated with a similar generation script that the permissions uses.
* I think the protobuf was imported into enterprise, so there will need to be some coordination when vendoring.

dustin-decker:
What was the motivation for this? We had the types generated for us before already.

zricethezav:
> What was the motivation for this? We had the types generated for us before already.

Speaking from my own personal experience here. I find it burdensome having to update a proto file and run `make protos` to generate code for types that could be defined in a regular non-generated go file. It adds an extra step in the development process that can easily be missed.

rgmz:
> Speaking from my own personal experience here. I find it burdensome having to update a proto file and run `make protos` to generate code for types that could be defined in a regular non-generated go file. It adds an extra step in the development process that can easily be missed.

+1 to this. In my case, I have to modify the script locally to fix an issue with SELinux.

Wrangling protos is a regular source of headaches for me, and the 10,000+ lines of generated code is difficult to comprehend. While I understand that it's used by the enterprise code, to me it feels like a lot of added complexity with little benefit.

dustin-decker:
> > What was the motivation for this? We had the types generated for us before already.
> 
> Speaking from my own personal experience here. I find it burdensome having to update a proto file and run `make protos` to generate code for types that could be defined in a regular non-generated go file. It adds an extra step in the development process that can easily be missed.

There are two parts on the non-generated file that need updating now, and it was suggested that we replace it with a custom generator. Which is fine, we could do that and have it run by `go generate`, but it hasn't been done yet.

PR #3494, opened 2024-10-22
Move DecoderType into ResultWithMetadata

rosecodym:
### Description:
`detectors.Result.DecoderType` is only ever used as part of an embedding into `ResultWithMetadata`, so we might as well move the field there to simplify the relationship between the structures. (I had the idea for this because it makes the sketch in #3457 simpler.)

(Draft until we double-check that enterprise consumption won't be affected.)

### Checklist:


rosecodym:
superseded by #3502

PR #3500, opened 2024-10-24
[feat] Gitlab inclusion globbing

abmussani:
### Description:
As on the customer demands, similar to github source, This PR implements inclusion of globs for Gitlab sources. Two new flags are introduce similar to Github, `--include-repos` and `--exclude-repos` respectively. Since, glob patterns are validated and ignore if not compiled, therefore, test to check those errors is removed.

`Gitlab` Soruce test results:

Screenshot 2024-10-24 at 4 26 06 PM

`--help` results for `Gitlab` source:

Screenshot 2024-10-24 at 4 19 03 PM

### Checklist:


camgunz (approved):
This looks like a pretty faithful reproduction of the behavior of globbing in the GitHub source, which is smart: it's great for consistency and it reduces the chance of weird errors creeping in. I have some suggestions to improve, but they're just suggestions, especially considering the GitHub source behavior here:
- Defer to @rosecodym here, but what do you think about glob compilation errors becoming error conditions? My thinking here is scanning something you don't want to scan, or not scanning something you definitely wanted to scan is probably something you want to address upfront, so canceling the scan might be more desirable.
- You could note that `exclude` takes precedence over `include` in the CLI flags usage info
- Not 100% sure about this, but you could also make the condition where the configuration of both `include` and `exclude` is such that no repos are included an error condition. The thing this would avoid is having a user typo a glob, get a ✅ from trufflehog, and think everything is peachy when they inadvertently didn't scan anything.

If you do decide to do any of these things (and to be clear IMO it's totally up to you), there's the [Validator interface](https://github.com/trufflesecurity/trufflehog/blob/0608b22e4fea3b63e75125e3926a76758fa5be21/pkg/sources/sources.go#L371) where you could do checking of the glob itself and any effects of the globbing.

pkg/sources/gitlab/gitlab.go
```diff
@@ -81,6 +84,57 @@ func (s *Source) JobID() sources.JobID {
 	return s.jobID
 }
 
+// filteredRepoCache is a wrapper around cache.Cache that filters out repos
+// based on include and exclude globs.
+type filteredRepoCache struct {
+	cache.Cache[string]
```
rosecodym, line 90: It doesn't look like the cache part of `filteredRepoCache` is ever used, so I don't think this is really a cache! That's a good thing - the repo caching in the GitHub source is pretty complicated, and it's good that we don't have to do it here too. You can remove the embedded `cache.Cache` and rename this structure - or even eliminate it entirely and pass the include/exclude globs directly to `buildIgnorer` (which would mean making `buildIgnorer` a little more complicated, but would also mean that you don't have to make anything _else_ more complicated).

pkg/sources/gitlab/gitlab.go
```diff
@@ -81,6 +84,57 @@ func (s *Source) JobID() sources.JobID {
...
+	include, exclude []glob.Glob
+}
+
+func (s *Source) newFilteredRepoCache(ctx context.Context, c cache.Cache[string], include, exclude []string) *filteredRepoCache {
+	includeGlobs := make([]glob.Glob, 0, len(include))
+	excludeGlobs := make([]glob.Glob, 0, len(exclude))
+	for _, ig := range include {
+		g, err := glob.Compile(ig)
+		if err != nil {
+			ctx.Logger().V(1).Info("invalid include glob", "include_value", ig, "err", err)
```
rosecodym, line 100: If possible, I definitely think that glob compilation errors should be a fatal error. At the very least we should log it with `Error` instead of `Info`.

rosecodym (changes requested):
Thanks for taking a look at this! Copying the GitHub implementation is a great idea, but that implementation is more complicated than what GitLab needs - in particular, GitLab doesn't need repo caching, as far as I can tell. I've left some comments inline with more details, and feel free to reach out to me directly with questions.

pkg/sources/gitlab/gitlab.go
abmussani: @rosecodym the idea is to compile it very first time and just reuse it, similar to Github. Moving compilation in buildignorer will compile globs for all kind of Projects (user projects, group projects etc).

abmussani: btw I do agree that `cache.Cache` is not needed in Gitlab source. I will take a look on that.

abmussani: @rosecodym  I have removed the concept of caching and renamed the entity 'globRepoFilter'. I did not eliminate the whole structure as I believe having that is just making the code more clear. Let me know what are your thoughts on this.

pkg/sources/gitlab/gitlab.go
abmussani: Reverted the code back to support compilation code. infact, now compilation will be done at `validation`, `enumeration` or `chunk` func like it was done before.

pkg/sources/gitlab/gitlab.go
```diff
@@ -81,6 +82,56 @@ func (s *Source) JobID() sources.JobID {
 	return s.jobID
 }
 
+// globRepoFilter is a wrapper around cache.Cache that filters out repos
+// based on include and exclude globs.
+type globRepoFilter struct {
+	include, exclude []glob.Glob
+}
+
+func newGlobRepoFilter(ctx context.Context, include, exclude []string, onCompileErr func(err error, pattern string)) *globRepoFilter {
```
rosecodym, line 91: it doesn't look like `ctx` is used here

pkg/sources/sources.go
```diff
@@ -266,6 +266,10 @@ type GitlabConfig struct {
 	Filter *common.Filter
 	// SkipBinaries allows skipping binary files from the scan.
 	SkipBinaries bool
+	// IncludeRepos is a list of repositories to include in the scan.
```
rosecodym, line 269: Is this a list of _repositories_ or _repository glob patterns_?

rosecodym (approved):
looks good, thanks! i added a note about an unused parameter and a question about some comments, but nothing blocking. can you also confirm my reading that this pr exposes the repo exclusion globbing that was always present in the engine but was not previously exposed on the command line?

pkg/sources/gitlab/gitlab.go
abmussani: Thanks for highlighting. Right, Its not being used. Let me clean it.

pkg/sources/sources.go
abmussani: List of repository glob patterns.

abmussani:
> this pr exposes the repo exclusion globbing that was always present in the engine but was not previously exposed on the command line?

Correct. I could not see a way to set `IgnoreRepo` in Gitlab source.

pkg/sources/gitlab/gitlab.go
abmussani: Removed.

PR #3513, opened 2024-10-26
[refactor] - base64 decoder

ahrav:
### Description:
Reduce the chance of potentially wasting work when decoding. Avoid extra allocations with strings.Trim

### Benchmarks
Screenshot 2024-10-25 at 10 48 55 PM

### Checklist:

### DIFF
--- a/pkg/decoders/base64.go
+++ b/pkg/decoders/base64.go
@@ -3,6 +3,7 @@ package decoders
 import (
 	"bytes"
 	"encoding/base64"
+	"sort"
 	"unicode"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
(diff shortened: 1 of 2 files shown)


pkg/decoders/base64.go
```diff
@@ -129,12 +146,38 @@ func getSubstringsOfCharacterSet(data []byte, threshold int, charsetMapping [128
...
+		right--
+	}
+
+	// Find equals sign in the middle if present.
+	for i := left; i < right; i++ {
+		if data[i] == '=' {
+			left = i + 1
+			break
+		}
+	}
```
mcastorina, line 175: This finds the first `=` between `left` and `right` and sets `left` to be one past it. This would be equivalent, correct?

```go
if equalPos := bytes.IndexByte(data[left:right], '='); equalPos != -1 {
    left = equalPos + 1
}
```

pkg/decoders/base64.go
```diff
@@ -129,12 +146,38 @@ func getSubstringsOfCharacterSet(data []byte, threshold int, charsetMapping [128
...
+
+	// Only append if we have valid data after trimming.
+	if right >= left {
+		substrings = append(substrings, string(data[left:right+1]))
 	}
```
mcastorina, line 180: Does it make sense to append if `right == left`? Or maybe a better question: should we be checking the `threshold` here too?

PR #3516, opened 2024-10-28
Feat: Added ZohoCRM detector

nabeelalam:
### Description:
This PR introduces a new detector for Zoho CRM API access tokens.

For the integration test, the active `ZOHOCRM` access token only has a 60 minute validity before becoming inactive, so an active token would have to be generated and added to the Secrets Manager for these tests to pass.

Access tokens can be generated by following the following steps in the Zoho documentation:
https://www.zoho.com/accounts/protocol/oauth/self-client/authorization-code-flow.html

### Checklist:


pkg/detectors/zohocrm/zohocrm_test.go
```diff
@@ -0,0 +1,230 @@
...
+			b.ResetTimer()
+			for n := 0; n < b.N; n++ {
+				_, err := s.FromData(ctx, false, data)
+				if err != nil {
+					b.Fatal(err)
+				}
+			}
+		})
+	}
+}
```
rgmz, line 230: I'd move this + the build tags into `zohocrm_integration_test.go` so that the pattern check always runs.

nabeelalam: Thanks for pointing that out @rgmz . Appreciate the insight. I've updated it now.

abmussani (commented):
@nabeelalam Thank you for new addition in Detectors. Along with some comments, You need to also add new detector in [Defaults.go](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/defaults.go#L1646) otherwise engine won't load it.

pkg/detectors/zohocrm/zohocrm.go
```diff
@@ -0,0 +1,127 @@
...
+
+type Scanner struct {
+	client *http.Client
+}
+
+type UnauthorizedResponseBody struct {
+	Code    string                 `json:"code"`
+	Details map[string]interface{} `json:"details"`
+	Message string                 `json:"message"`
+	Status  string                 `json:"status"`
```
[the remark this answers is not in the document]
nabeelalam: Yeah I had mapped the output, in order to add extra data in the response but it felt unnecessary. I've removed this fields.

pkg/detectors/zohocrm/zohocrm.go
```diff
@@ -0,0 +1,127 @@
...
+
+		results = append(results, result)
+	}
+
+	return
+}
+
+// Verifies the Zoho CRM API key by making an HTTP request to the Zoho CRM API.
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	endpoint := "https://www.zohoapis.com/crm/v2/Leads"
```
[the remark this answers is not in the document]
nabeelalam: No particular reason, I've updated it to v7, difference being that it requires a `fields` query parameter

pkg/detectors/zohocrm/zohocrm.go
```diff
@@ -0,0 +1,124 @@
...
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(1000\.[a-f0-9]{32}\.[a-f0-9]{32})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"1000."}
```
kashifkhan0771, line 37: I think it will be good to add `zoho` as a Keyword too.

kashifkhan0771 (changes requested):
Overall LGTM! Rebasing is required.

abmussani (approved):

pkg/detectors/zohocrm/zohocrm.go
abmussani: I believe token consists of unique `1000.` pattern. should be enough like [LaunchDarkly](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/launchdarkly/launchdarkly.go)

kashifkhan0771 (approved):

pkg/detectors/zohocrm/zohocrm.go
```diff
@@ -0,0 +1,124 @@
...
+		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
+}
+
+func (s Scanner) Type() detectorspb.DetectorType {
+	return detectorspb.DetectorType_ZohoCRM
+}
+
+func (s Scanner) Description() string {
+	return "Zohocrm is a blockchain development platform that provides a suite of tools and services for building and scaling decentralized applications. Zohocrm API keys can be used to access these services."
```
rgmz, line 123: I don't think this is correct?

https://www.zoho.com/crm/

PR #3519, opened 2024-10-28
[refactor] - Adjust File Handling Errors

ahrav:
### Description:
This PR address the first point in [this](https://github.com/trufflesecurity/trufflehog/pull/3403#issuecomment-2430039719) comment.

Avoid shadowing `err` by handling it immediately upon encountering an error.
### Checklist:

### DIFF
--- a/pkg/handlers/ar.go
+++ b/pkg/handlers/ar.go
@@ -31,41 +31,35 @@ func (h *arHandler) HandleFile(ctx logContext.Context, input fileReader) (chan [
 	}
 
 	go func() {
-		ctx, cancel := logContext.WithTimeout(ctx, maxTimeout)
-		defer cancel()
 		defer close(archiveChan)
-
-		// Update the metrics for the file processing.
-		start := time.Now()
-		var err error
-		defer func() {
-			h.measureLatencyAndHandleErrors(start, err)
-			h.metrics.incFilesProcessed()
-		}()
 
 		// Defer a panic recovery to handle any panics that occur during the AR processing.
 		defer func() {
 			if r := recover(); r != nil {
-				// Return the panic as an error.
+				var panicErr error
 				if e, ok := r.(error); ok {
-					err = e
+					panicErr = e
 				} else {
-					err = fmt.Errorf("panic occurred: %v", r)
+					panicErr = fmt.Errorf("panic occurred: %v", r)
 				}
-				ctx.Logger().Error(err, "Panic occurred when reading ar archive")
+				ctx.Logger().Error(panicErr, "Panic occurred when attempting to open ar archive")
 			}
 		}()
 
-		var arReader *deb.Ar
-		arReader, err = deb.LoadAr(input)
+		start := time.Now()
+		arReader, err := deb.LoadAr(input)
 		if err != nil {
-			ctx.Logger().Error(err, "error reading AR")
+			ctx.Logger().Error(err, "Error loading AR file")
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/handlers/ar.go
```diff
@@ -31,41 +31,35 @@ func (h *arHandler) HandleFile(ctx logContext.Context, input fileReader) (chan [
...
-		defer cancel()
 		defer close(archiveChan)
 
-		// Update the metrics for the file processing.
-		start := time.Now()
-		var err error
-		defer func() {
-			h.measureLatencyAndHandleErrors(start, err)
-			h.metrics.incFilesProcessed()
-		}()
```
rosecodym, line 44: Does this mean that archives that generate errors during processing are no longer counted by the metric and no longer have their latency measured?

ahrav: This change increments `incFilesProcessed` only for Non-Fatal errors, but latency is recorded regardless of success or failure. For Fatal errors, we increment the `incErrors` metric within `measureLatencyAndHandleErrors` instead.

I was concerned when I saw that we’re incrementing the file processed metric even in cases of errors. The metric name may be unclear, as I assumed "processed" meant successfully processed, not just acknowledged. With the current approach, even files we can’t open are counted as processed, which feels somewhat misleading. I’m open to changing it back, though.

pkg/handlers/rpm.go
```diff
@@ -31,48 +31,41 @@ func (h *rpmHandler) HandleFile(ctx logContext.Context, input fileReader) (chan
...
 			}
 		}()
 
-		var rpm *rpmutils.Rpm
-		rpm, err = rpmutils.ReadRpm(input)
+		start := time.Now()
+		rpm, err := rpmutils.ReadRpm(input)
 		if err != nil {
-			ctx.Logger().Error(err, "error reading RPM")
+			ctx.Logger().Error(err, "Error reading rpm file")
```
mcastorina, line 52: Are we capitalizing log messages now? Is that a best practice style?

mcastorina (approved):

pkg/handlers/ar.go
dustin-decker: Food for thought, the status could be represented with a metric tag and then we could use the same metric visualizations 'by status', which could be "success", "failure", "skipped" etc.

pkg/handlers/rpm.go
```diff
@@ -31,48 +31,41 @@ func (h *rpmHandler) HandleFile(ctx logContext.Context, input fileReader) (chan
...
 			return
 		}
 
-		var reader rpmutils.PayloadReader
-		reader, err = rpm.PayloadReaderExtended()
+		reader, err := rpm.PayloadReaderExtended()
 		if err != nil {
-			ctx.Logger().Error(err, "error getting RPM payload reader")
+			ctx.Logger().Error(err, "Error reading rpm file")
```
casey-tran, line 58: Would you want to have the error log string on line 58 to be different than the one located in line 52 in order to make debugging a little easier to find the line that caused the error?

ahrav: Great idea! I’ve been caught reusing the same error message too often. Thanks for pointing that out. 🙇

pkg/handlers/rpm.go
ahrav: Oof, nope.

pkg/handlers/ar.go
ahrav: That's an interesting idea. I'll put together a separate PR with that. Thanks for the suggestion.

PR #3520, opened 2024-10-28
[refactor] - Add DataOrErr

ahrav:
### Description:
This PR addresses the second point in [this](https://github.com/trufflesecurity/trufflehog/pull/3403#issuecomment-2430039719) comment.
### Checklist:


pkg/handlers/default.go
```diff
@@ -30,37 +31,53 @@ func newDefaultHandler(handlerType handlerType) *defaultHandler {
...
 
 // measureLatencyAndHandleErrors measures the latency of the file processing and updates the metrics accordingly.
 // It also records errors and timeouts in the metrics.
-func (h *defaultHandler) measureLatencyAndHandleErrors(start time.Time, err error) {
+func (h *defaultHandler) measureLatencyAndHandleErrors(
+	ctx logContext.Context,
+	start time.Time,
+	err error,
+	dataErrChan chan DataOrErr,
+) {
```
mcastorina, line 61: I'm not super familiar with this area of the codebase, but it's starting to smell like this method is doing a little too much.

Am I reading this right, that we're now writing to the `dataErrChan` instead of just recording metrics? (side nit: make `dataErrChan` a write-only parameter).

pkg/handlers/handlers.go
```diff
@@ -80,19 +80,32 @@ func newMimeTypeReader(r io.Reader) (mimeTypeReader, error) {
...
+	var mime *mimetype.MIME
+	mime, err = mimetype.DetectReader(fReader)
 	if err != nil {
 		return fReader, fmt.Errorf("unable to detect MIME type: %w", err)
 	}
 	fReader.mime = mime
 
 	// Reset the reader to the beginning because DetectReader consumes the reader.
-	if _, err := fReader.Seek(0, io.SeekStart); err != nil {
+	if _, err = fReader.Seek(0, io.SeekStart); err != nil {
```
mcastorina, line 108: These changes look unrelated to `DataOrErr`, were they accidentally added to this PR?

pkg/iobuf/bufferedreaderseeker.go
```diff
@@ -266,6 +266,10 @@ func (br *BufferedReadSeeker) readToEnd() error {
 }
 
 func (br *BufferedReadSeeker) writeData(data []byte) error {
+	if br.buf == nil {
+		br.buf = br.bufPool.Get()
+	}
+
```
mcastorina, line 272: Same question here about this being unrelated to `DataOrErr`.

mcastorina (commented):
I'm still grumpy we can't call it `Result`, but I'll live I guess

dustin-decker (approved):
Nice improvement

pkg/handlers/default.go
ahrav: I agree. I can create a follow-up PR to break this function into smaller parts. You're correct—we need to pass all errors to dataErrChan instead of just logging them and moving on, as we did before.

pkg/handlers/handlers.go
ahrav: Ah, yes—these changes were part of an earlier [PR](https://github.com/trufflesecurity/trufflehog/pull/3403/files#diff-3d5d44f7e881ea6d878fc0163dc6c9199eafb0d6eda163d23ba70e3952b3dde9R95) that I split into three separate PRs as requested. This felt like the most appropriate PR to include the change originally in the initial PR."

PR #3521, opened 2024-10-28
[feat] - Introduce Fatal/Non-Fatal File Handling Errors

ahrav:
### Description:
This PR addresses the third point in [this](https://github.com/trufflesecurity/trufflehog/pull/3403#issuecomment-2430039719) comment.
### Checklist:


pkg/handlers/handlers.go
```diff
@@ -351,7 +361,7 @@ func HandleFile(
 // handleChunksWithError processes data and errors received from the dataErrChan channel.
 // For each DataOrErr received:
 // - If it contains data, the function creates a chunk based on chunkSkel and reports it through the reporter.
-// - If it contains an error, the function logs the error.
+// - If it contains an error, the function returns the error immediately.
```
rosecodym, line 364: This is for _fatal_ errors, right? This comment seems like a good place to explain non-fatal errors as well. I would also like in some comment _somewhere_ a description of what types of errors are fatal versus non-fatal - maybe in the doc comments of functions that use `DataOrErr`?

rosecodym (approved):
I have a request for some additional comments, but this looks pretty clean! Thanks for breaking it out.

pkg/handlers/ar.go
```diff
@@ -42,14 +42,18 @@ func (h *arHandler) HandleFile(ctx logContext.Context, input fileReader) chan Da
...
+				}
 			}
 		}()
 
 		start := time.Now()
 		arReader, err := deb.LoadAr(input)
 		if err != nil {
-			ctx.Logger().Error(err, "Error loading AR file")
+			dataOrErrChan <- DataOrErr{
+				Err: fmt.Errorf("%w: loading AR error: %v", ErrProcessingFatal, err),
```
mcastorina, line 55: Do we care that we "lose" the original `err` since we're wrapping the sentinel error?

mcastorina (approved):

pkg/handlers/ar.go
ahrav: We would still maintain the error string though correct? Do you think we should use `errors.Join` here instead?

mcastorina: > We would still maintain the error string though correct?

Yeah

> Do you think we should use errors.Join here instead?

I think it would have the same effect. It's probably better to wrap the sentinel, right?

ahrav: I believe so. We need to be able to use `errors.Is` to determine if it's a fatal error.

PR #3522, opened 2024-10-28
Create global log redaction capability

rosecodym:
### Description:
Some source use client libraries that can emit errors that contain sensitive information - in particular, git-facing libraries that embed tokens into repository URLs. This PR introduces a way of redacting them - starting with GitLab (where we've seen this most recently), but in theory extensible to other sources as needed.

This implementation uses a custom zap core; this might also be possible with a custom zap encoder, but I didn't test it out.

(The deleted core.go file was entirely unused.)

### Checklist:


dustin-decker (approved):
This is great. Don't forget to update the redactor for other sources though (can be in a follow-up PR)

zricethezav (approved):

PR #3550, opened 2024-11-01
[bug] - Correct Line Number Calculation

ahrav:
### Description:
This PR updates the `fragmentStart` value to use 1-based indexing, aligning it with file numbering. This change resolves the off-by-one line number issue that occurred when the `Line` value was not specified in the `source_metadata` proto.
### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -1285,8 +1285,14 @@ func FragmentFirstLineAndLink(chunk *sources.Chunk) (int64, *int64, string) {
 		fragmentStart = &metadata.AzureRepos.Line
 		link = metadata.AzureRepos.Link
 	default:
-		return 0, nil, ""
+		return 1, nil, ""
 	}
+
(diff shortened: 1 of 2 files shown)


rosecodym (commented):
Thanks for this (and for the test!) But the last time we fixed an "obvious" off-by-one line number error, a bunch of people complained about a breaking change, and we reverted it. Will those complaints happen again here? And if it does, do we have a plan for it? (This is largely a @zricethezav question I think)

zricethezav:
> But the last time we fixed an "obvious" off-by-one line number error, a bunch of people complained about a breaking change, and we reverted it. Will those complaints happen again here?

Who can say. Ahrav wrote a test here that looks solid to me. If it breaks something, we'll find out about it quickly and revert it.

zricethezav (approved):

dustin-decker (approved):

PR #3554, opened 2024-11-03
[chore] - minor cleanup S3 source

ahrav:
### Description:
This PR resolves linter issues and removes the logger from the S3 Source struct, switching to using the context for logging instead.
### Checklist:

### DIFF
--- a/pkg/sources/s3/s3.go
+++ b/pkg/sources/s3/s3.go
@@ -15,8 +15,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/go-errors/errors"
-	"github.com/go-logr/logr"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/log"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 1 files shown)


pkg/sources/s3/s3.go
```diff
@@ -198,53 +199,62 @@ func (s *Source) getBucketsToScan(client *s3.S3) ([]string, error) {
...
-		logger := logger.WithValues("bucket", bucket)
+		ctx := context.WithValue(ctx, "bucket", bucket)
 
 		if common.IsDone(ctx) {
 			return
 		}
 
-		s.SetProgressComplete(i, len(bucketsToScan), fmt.Sprintf("Bucket: %s", bucket), "")
-		logger.V(3).Info("Scanning bucket")
+		s.SetProgressComplete(i, len(bucketsToScan), "Bucket: "+bucket, "")
```
rosecodym, line 222: I have no opinion about your other stylistic changes, but this one took me a second to understand. Do you have a strong objection to retaining `fmt.Sprintf` here?

rosecodym (approved):
thanks so much for doing this!

pkg/sources/s3/s3.go
ahrav: Happy to change it back. I'll also consider disabling perfsprintf from my linter.

PR #3568, opened 2024-11-07
[feat] - S3 Progress Tracker

ahrav:
### Description:
This PR adds a mechanism to track S3 scan progress, enabling resumption.

### Checklist:


pkg/sources/s3/progress_tracker.go
```diff
@@ -0,0 +1,208 @@
...
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+// ProgressTracker maintains scan progress state for S3 bucket scanning,
+// enabling resumable scans by tracking which objects have been successfully processed.
+// It provides checkpoints that can be used to resume interrupted scans without missing objects.
+type ProgressTracker struct {
+	enabled bool
```
mcastorina, line 19: If `enabled` does not change at runtime, have you considered a noop implementation?

pkg/sources/s3/progress_tracker.go
```diff
@@ -0,0 +1,208 @@
...
+	}, nil
+}
+
+// Complete marks the entire scanning operation as finished and clears the resume state.
+// This should only be called once all scanning operations are complete.
+func (p *ProgressTracker) Complete(_ context.Context, message string) error {
+	// Preserve existing progress counters while clearing resume state.
+	p.progress.SetProgressComplete(
+		int(p.progress.SectionsCompleted),
+		int(p.progress.SectionsRemaining),
```
mcastorina, line 114: Why preserve the counters?

pkg/sources/s3/progress_tracker.go
```diff
@@ -0,0 +1,208 @@
...
+const defaultMaxObjectsPerPage = 1000
+
+// NewProgressTracker creates a new progress tracker for S3 scanning operations.
+// The enabled parameter determines if progress tracking is active, and progress
+// provides the underlying mechanism for persisting scan state.
+func NewProgressTracker(ctx context.Context, enabled bool, progress *sources.Progress) *ProgressTracker {
+	if progress == nil {
+		ctx.Logger().Info("Nil progress provided. Progress initialized.")
+		progress = new(sources.Progress)
+	}
```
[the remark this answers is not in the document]
ahrav: Sigh, I thought I removed this. 😢 Yea this doesn't work, we should just return an error here. If a source doesn't have Progress, we have bigger problems.

pkg/sources/s3/progress_tracker.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+// ProgressTracker maintains scan progress state for S3 bucket scanning,
+// enabling resumable scans by tracking which objects have been successfully processed.
+// It provides checkpoints that can be used to resume interrupted scans without missing objects.
+type ProgressTracker struct {
+	enabled bool
+
+	isResuming bool // Indicates if the tracker is resuming a previous scan
```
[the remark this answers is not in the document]
ahrav: The name is likely misleading—it was intended as a flag to indicate **some** progress, primarily used in the update flow for calculating SectionRemaining. However, I don’t think we need to manage progress at all, so I’ll likely remove this and any related logic.

ex:
```
pageSize := int32(len(pageContents))
    if p.currentPageSize == 0 {
        if !p.isResuming {
            // Only add to the total if this is a fresh scan.
            p.progress.SectionsRemaining += pageSize
        }
	p.currentPageSize = pageSize
}
```

pkg/sources/s3/progress_tracker.go
ahrav: I was being lazy, assuming this would only be temporary. Since progress tracking will eventually always be enabled, I just went with this approach. 😓

pkg/sources/s3/progress_tracker.go
ahrav: This was because `Complete` is only called at the end of the scan. Since progress was tracked at the object level, there’s no need to set these values. I’ll be removing all progress logic. Progress tracking will still happen at the bucket level, but resuming will remain at the object level.

mcastorina (approved):
I can't say I fully understand what `UpdateObjectProgress` is doing, but it would be nice if there were some package-level docs on how `ProgressTracker` should be used for people like me who don't know the structure of s3 buckets.

Btw, this isn't being used yet, right?

pkg/sources/s3/progress_tracker.go
```diff
@@ -0,0 +1,185 @@
...
+	// progress holds the scan's overall progress state and enables persistence.
+	progress *sources.Progress // Reference to source's Progress
+}
+
+const defaultMaxObjectsPerPage = 1000
+
+// NewProgressTracker creates a new progress tracker for S3 scanning operations.
+// The enabled parameter determines if progress tracking is active, and progress
+// provides the underlying mechanism for persisting scan state.
+func NewProgressTracker(_ context.Context, enabled bool, progress *sources.Progress) (*ProgressTracker, error) {
```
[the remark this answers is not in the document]
ahrav: Ah, I believe it's because initially there was some logging in here. Let me add the log I had removed. I can remove it from Reset though.

PR #3572, opened 2024-11-07
Add GitLab shared exclusion flag

rosecodym:
### Description:
We've been seeing some suspicious behavior from the GitLab API when we're scanning a very large organization that has lots of group sharing. Since reproducing the behavior in a test environment isn't really tractable, this PR adds an optional configuration flag that can be used to investigate the phenomenon in the actual environment we're seeing it in.

### Checklist:


pkg/sources/gitlab/gitlab.go
```diff
@@ -577,6 +582,7 @@ func (s *Source) getAllProjectRepos(
 			ListOptions:      listOpts,
 			OrderBy:          gitlab.Ptr(orderBy),
 			IncludeSubGroups: gitlab.Ptr(true),
+			WithShared:       gitlab.Ptr(s.enumerateSharedProjects),
```
mcastorina, line 585: The [docs](https://docs.gitlab.com/ee/api/groups.html#list-a-groups-projects) indicate `with_shared` defaults to `true`. Is this config option here to disable `with_shared` for debug testing?

zricethezav (approved):

pkg/sources/gitlab/gitlab.go
rosecodym: yes exactly

mcastorina (approved):

casey-tran (approved):

PR #3577, opened 2024-11-08
[feat] - S3 metrics

ahrav:
### Description:
This PR adds metrics for tracking a S3 scan.
### Checklist:


pkg/sources/s3/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+var metricsInstance metricsCollector
+
+func init() { metricsInstance = newS3MetricsCollector() }
+
+// getMetricsCollector returns the singleton metrics collector instance..
+func getMetricsCollector() metricsCollector { return metricsInstance }
+
+func newS3MetricsCollector() metricsCollector {
```
mcastorina, line 39: Could the contents of this function be directly in `init()`? Why have it separate? It suggests to me that it's getting called more than once.

pkg/sources/s3/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+	rolesScanned   *prometheus.GaugeVec
+	bucketsPerRole *prometheus.GaugeVec
+}
+
+var metricsInstance metricsCollector
+
+func init() { metricsInstance = newS3MetricsCollector() }
+
+// getMetricsCollector returns the singleton metrics collector instance..
+func getMetricsCollector() metricsCollector { return metricsInstance }
```
mcastorina, line 37: Why have a `get` function if `metricsInstance` is global?

mcastorina (approved):
The changes seem fine, I'm just a little confused why the metrics collector object is a field of the s3 source instead of a global like our other metrics.

ahrav:
> The changes seem fine, I'm just a little confused why the metrics collector object is a field of the s3 source instead of a global like our other metrics.

Mainly for testability. I haven’t added the metrics tests yet, but I want to make them testable in a similar way to how I tested the metrics for our cache package [here](https://github.com/trufflesecurity/trufflehog/blob/8f2ebc924e8dc4ae882a322ea61f3b984b485cb8/pkg/cache/lru/lru_test.go#L138)

dustin-decker (approved):

pkg/sources/s3/metrics.go
ahrav: I was following the approach we use to access metrics in our cache package, but it doesn’t apply well here since the metrics are private. I'll adjust.

PR #3587, opened 2024-11-11
Sort defaults.go

rgmz:
### Description:
`defaults.go` is a common source of merge conflicts despite no actual functional changes.

Sorting the order of `detectorList` should significantly reduce these occurrences.

```
		weightsandbiases.Scanner{},
<<<<<<< HEAD
		hubspotapikeyv2.Scanner{},
=======
		&hubspot_apikey_v1.Scanner{},
		&hubspot_apikey_v2.Scanner{},
>>>>>>> e06685e2 (feat(hubspot): unify v1 and v2 implementation)
	}
```

### Checklist:


ahrav:
I agree that this solution helps avoid merge conflicts, but I’m concerned new contributors might not maintain the ordering of the detector list in `defaults.go`.

rgmz:
> ... but I’m concerned new contributors might not maintain the ordering of the detector list in defaults.go.

Yeah, it would have to be caught in PR review. I believe many people will notice the existing convention and stick to it.

ahrav (approved):

ahrav:
> @ahrav the test failure seems related to a problem that #3333 was trying to prevent — but I don't know what it is.
> 
> @mcastorina

I _think_ it's because this PR changed `github_oauth2.Scanner{}` -> `&github_oauth2.Scanner{}`.

ahrav:
> > @ahrav the test failure seems related to a problem that #3333 was trying to prevent — but I don't know what it is.
> > @mcastorina
> 
> I _think_ it's because this PR changed `github_oauth2.Scanner{}` -> `&github_oauth2.Scanner{}`.

I guess when using a ptr the embedded `EndpointSetter`'s methods don't get promoted correctly?

rgmz:
> > @ahrav the test failure seems related to a problem that #3333 was trying to prevent — but I don't know what it is.
> > @mcastorina
> 
> I _think_ it's because this PR changed `github_oauth2.Scanner{}` -> `&github_oauth2.Scanner{}`.

Ah, I was trying to make the list consistent. The old ones are all pointers — I don't know why.

ahrav:
> > > @ahrav the test failure seems related to a problem that #3333 was trying to prevent — but I don't know what it is.
> > > @mcastorina
> > 
> > 
> > I _think_ it's because this PR changed `github_oauth2.Scanner{}` -> `&github_oauth2.Scanner{}`.
> 
> Ah, I was trying to make the list consistent. The old ones are all pointers — I don't know why.

Actually, I `believe` this is a bug. I had the logic reversed. When using a value type for `github_oauth2.Scanner{}`, the following interface assertion fails:  
`if endpoints, ok := det.(interface{ Endpoints(...string) []string }); ok {`  

This suggests the test wasn’t working correctly. When switching to pointer semantics, the assertion passes, but the endpoint list is empty. This indicates a potential issue either with this test or with the `EndpointCustomer` implementation.

TBH, i'm not really sure how any of that stuff works. 😅 I'll defer to Miccah here.

mcastorina:
The goal of the test is to make sure `Endpoints()` returns at least one value, otherwise we'll silently "verify" against 0 endpoints.

It looks like `github_oauth2` embedded `EndpointSetter` but isn't calling the `Endpoints()` method anywhere, so imo the bug is that it is embedding it and not using it. The fix would be to remove that struct embedding.

ahrav:
> The goal of the test is to make sure `Endpoints()` returns at least one value, otherwise we'll silently "verify" against 0 endpoints.
> 
> It looks like `github_oauth2` embedded `EndpointSetter` but isn't calling the `Endpoints()` method anywhere, so imo the bug is that it is embedding it and not using it. The fix would be to remove that struct embedding.

Ah thanks. I can remove it.

ahrav:
https://github.com/trufflesecurity/trufflehog/pull/3592 should fix the test.

PR #3383, opened 2024-10-07
Log skipped files on debug level

rgmz:
### Description:
This was inspired by my comment [here](https://github.com/trufflesecurity/trufflehog/pull/3375#issuecomment-2396885980). Trace is extremely noisy[^1], such information should be logged on the debug level IMO.

```sh
$ ./trufflehog github ... --debug
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2024-10-07T19:50:42-04:00	info-2	trufflehog	starting scanner workers	{"count": 4}
2024-10-07T19:50:42-04:00	info-2	trufflehog	starting detector workers	{"count": 32}
2024-10-07T19:50:42-04:00	info-2	trufflehog	starting verificationOverlap workers	{"count": 4}
2024-10-07T19:50:42-04:00	info-2	trufflehog	starting notifier workers	{"count": 4}
2024-10-07T19:50:42-04:00	info-0	trufflehog	running source	{"source_manager_worker_id": "skxXF", "with_units": true}
2024-10-07T19:50:42-04:00	info-2	trufflehog	enumerating source	{"source_manager_worker_id": "skxXF"}
2024-10-07T19:50:42-04:00	info-2	trufflehog	Caching repository info	{"source_manager_worker_id": "skxXF"}
2024-10-07T19:50:42-04:00	info-1	trufflehog	Enumerating with token	{"source_manager_worker_id": "skxXF"}
2024-10-07T19:50:42-04:00	info-2	trufflehog	attempting to clone repo	{"source_manager_worker_id": "skxXF", "unit": "https://github.com/cohere-ai/tokenizer.git", "unit_kind": "repo", "repo": "https://github.com/cohere-ai/tokenizer.git"}
2024-10-07T19:50:42-04:00	info-0	trufflehog	Completed enumeration	{"source_manager_worker_id": "skxXF", "num_repos": 1, "num_orgs": 0, "num_members": 0}
2024-10-07T19:50:46-04:00	info-1	trufflehog	successfully cloned repo	{"source_manager_worker_id": "skxXF", "unit": "https://github.com/cohere-ai/tokenizer.git", "unit_kind": "repo", "repo": "https://github.com/cohere-ai/tokenizer.git", "subcommand": "git clone", "repo": "https://github.com/cohere-ai/tokenizer.git", "path": "/tmp/trufflehog-41733-1261391252", "args": []}
[...]

PR #3589, opened 2024-11-12
Increase level for verbose log

rgmz:
### Description:
This bumps `skipping file: extension is ignored` from `--debug` to `--trace` as it can be quite noisy.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -182,7 +182,7 @@ func (h *archiveHandler) extractorHandler(archiveChan chan []byte) func(context.
 		}
 
 		if common.SkipFile(file.Name()) || common.IsBinary(file.Name()) {
-			lCtx.Logger().V(2).Info("skipping file: extension is ignored")
+			lCtx.Logger().V(3).Info("skipping file: extension is ignored")
 			h.metrics.incFilesSkipped()
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


ahrav:
> ### Description:
> This bumps `skipping file: extension is ignored` from `--debug` to `--trace` as it can be quite noisy.
> 
> ### Checklist:
> * [ ]  Tests passing (`make test-community`)?
> * [ ]  Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

I see you had a change of heart 😉 
https://github.com/trufflesecurity/trufflehog/pull/3383#issue-2571741034

rgmz:
Lol, I am torn on it. Maybe it's fine, idk.

I think part of the problem is that you can't specify a level so it's all or nothing. There's a lot of stuff in 3 & 4 that I'd like to see without 5.

ahrav:
>  I think part of the problem is that you can't specify a level 

+1000 to this. I’ve always found the way we map log levels to CLI flags a bit puzzling. I don’t have strong opinions on what level this should be, but if we stick with level 3, we should update the contributing.md [here](https://github.com/trufflesecurity/trufflehog/blob/35d5baf6d2f43760ab863a7f0f23ad7329b6a376/CONTRIBUTING.md?plain=1#L32-L32) 🤣

ahrav (approved):
Approving even though it doesn't actually have any effect. 😅

dustin-decker:
> > I think part of the problem is that you can't specify a level
> 
> +1000 to this. I’ve always found the way we map log levels to CLI flags a bit puzzling. I don’t have strong opinions on what level this should be, but if we stick with level 3, we should update the contributing.md [here](https://github.com/trufflesecurity/trufflehog/blob/35d5baf6d2f43760ab863a7f0f23ad7329b6a376/CONTRIBUTING.md?plain=1#L32-L32) 🤣

Those CLI flags predated our leveled logger

dustin-decker (approved):

PR #3612, opened 2024-11-15
[chore] Log non-fatal errors encountered during a scan

mcastorina:
### Description:
Non-fatal errors produced by a source during enumeration or chunking are recorded in the report, but we don't do anything with them. This PR logs them after the scan completes.

### Checklist:


main.go
```diff
@@ -881,6 +882,15 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
 		return scanMetrics, fmt.Errorf("engine failed to finish execution: %v", err)
 	}
 
+	// Print any errors reported during the scan.
+	if errs := ref.Snapshot().Errors; len(errs) > 0 {
+		errMsgs := make([]string, len(errs))
+		for i := 0; i < len(errs); i++ {
+			errMsgs[i] = errs[i].Error()
+		}
```
rosecodym, line 890: Why did you stringify these yourself in a new slice instead of just logging them directly?

rosecodym (approved):

main.go
mcastorina: I thought it looked better stringified:

```
{"errors": ["error chunking dir \"/tmp/fake-dir/\": repository does not exist"]}
```

vs

```
{"errors": [{"error": "error chunking dir \"/tmp/fake-dir/\": repository does not exist"}]}
```

PR #3615, opened 2024-11-18
added godaddy detector

kashifkhan0771:
### Description:
Added a new detector for GoDaddy.
API Documentation link: https://developer.godaddy.com/doc

### Test Cases:
Screenshot from 2024-11-18 13-19-26

### Checklist:


pkg/detectors/godaddy/v1/godaddy.go
```diff
@@ -0,0 +1,145 @@
...
+	defaultClient = common.SaneHttpClient()
+
+	// the key for the GoDaddy OTE environment is a 37-character alphanumeric string that may include underscores.
+	keyPattern = regexp.MustCompile(detectors.PrefixRegex([]string{"godaddy", "ote"}) + common.BuildRegex("a-zA-Z0-9", "_", 37))
+	// the secret for the GoDaddy OTE environment is a 22-character alphanumeric string.
+	secretPattern = regexp.MustCompile(detectors.PrefixRegex([]string{"godaddy", "ote"}) + common.BuildRegex("a-zA-Z0-9", "", 22))
+)
+
+// supported environments for godaddy
+var (
```
ahrav, line 36: questions: Is there a specific reason these can’t be `const`s? Also, I believe `OTE` is only used in the v1 detector, while `Prod` is used in the v2 detector. Could we move `Prod` to the v2 file? Lastly, do you think we could make them unexported? It doesn’t seem like they need to be visible outside this package, though I might be mistaken.

pkg/detectors/godaddy/v1/godaddy.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return defaultClient
+}
+
+func (s *Scanner) Version() int { return 1 }
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"godaddy"}
```
ahrav, line 54: question: I noticed that `keyPattern` and `secretPattern` include `ote` as a keyword. Should we also include it here?

pkg/detectors/godaddy/v1/godaddy.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	return results, nil
+
+}
+
+// VerifyGoDaddySecret make a call to godaddy api with given secret to check if secret is valid or not.
+func VerifyGoDaddySecret(ctx context.Context, client *http.Client, environment GoDaddyEnv, secret string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://%s/v1/domains/available?domain=example.com", environment), http.NoBody)
```
ahrav, line 112: TIL about `http.NoBody`, nice. 🎉

pkg/detectors/godaddy/v1/godaddy.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+type Scanner struct {
+	client *http.Client
+}
+
+type GoDaddyEnv string
```
ahrav, line 20: nit: Since the package is already named `godaddy`, we could rename this to `env` to avoid stuttering. It also seems safe to unexport it.

[ref](https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name)

ahrav (approved):
Few non-blocking comments. Otherwise, LGTM, thanks for adding this.

pkg/detectors/godaddy/v1/godaddy.go
kashifkhan0771: Your suggestion is absolutely spot on. I guess I just did it for the style 😆

pkg/detectors/godaddy/v1/godaddy.go
kashifkhan0771: Yes, we can! Initially, I thought that if we find the `godaddy` keyword in the chunk, we could then look for the `ote` or `godaddy` prefix. However, including `ote` as a keyword directly won’t cause any harm.

ahrav (approved):

zricethezav (approved):

PR #3625, opened 2024-11-19
Recover general chunker panics

rosecodym:
### Description:
We have recently seen panics when underlying readers panic (which can happen due to third-party library bugs). I'm not sure why the panics are _new_ - it could be a case of bad luck, but there have also been recent changes to error handling. This is the smallest change that seems to fix a real problem but I'm not sure if it's the best global solution.

### Checklist:

### DIFF
--- a/pkg/sources/chunker_test.go
+++ b/pkg/sources/chunker_test.go
@@ -2,6 +2,7 @@ package sources
 
 import (
 	"bytes"
+	"io"
 	"math/rand"
 	"runtime"
 	"strings"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
 )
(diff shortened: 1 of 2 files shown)


zricethezav (approved):

pkg/sources/chunker_test.go
```diff
@@ -121,6 +123,21 @@ func TestNewChunkedReader(t *testing.T) {
...
 
+type panicReader struct{}
+
+var _ io.Reader = (*panicReader)(nil)
+
+func (_ panicReader) Read([]byte) (int, error) {
+	panic("panic for testing")
+}
+
+func TestChunkReader_UnderlyingReaderPanics_DoesNotPanic(t *testing.T) {
```
0x1, line 134: should the test verify that the error is returned through the channel?

0x1 (approved):
one question. otherwise looks good

pkg/sources/chunker_test.go
rosecodym: good question! my vote is "no," because imo the specific correct behavior is not obvious, beyond "shouldn't panic" - which is what the test checks right now.

ahrav (approved):
Thank you ❤️

PR #3626, opened 2024-11-19
[scan-9] Update enumeration logic

0x1:
Added functionality to enumerate Source with Unit information. Renamed older functions to reflect new functionality.

### Checklist:


pkg/sources/sources.go
```diff
@@ -103,6 +103,30 @@ type SourceUnitEnumerator interface {
 	Enumerate(ctx context.Context, reporter UnitReporter) error
 }
 
+// BaseUnitReporter is a helper struct that implements the UnitReporter interface
+// and includes a JobProgress reference.
+type baseUnitReporter struct {
```
rosecodym, line 108: What makes this a "base" unit reporter?

rosecodym (approved):
Looks good - just one question about naming.

pkg/sources/sources.go
0x1: the passed-in reporter gets wrapped into this, and so this reporter becomes the parent. 'base' made sense in that respect.

PR #3632, opened 2024-11-19
Update Azure Container Registry detector

rgmz:
### Description:
This updates the Azure Container registry detector, which is a frequent source of noise.

1. Cache registries so we don't make multiple attempts to things that don't exist
2. The registry name cannot contain uppercase letters AFAIK.
3. The registry name probably shouldn't be censored.
4. The password seems to follow some sort of prefedined format that ends with `+ACR......`.

### Checklist:

### DIFF
--- a/pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
+++ b/pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
@@ -2,13 +2,17 @@ package azurecontainerregistry
 
 import (
 	"context"
-	"encoding/base64"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
+	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/cache/simple"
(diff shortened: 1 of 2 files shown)


pkg/detectors/azurecontainerregistry/azurecontainerregistry.go
```diff
@@ -35,59 +41,75 @@ func (s Scanner) Keywords() []string {
...
+				Raw:          []byte(password),
+				RawV2:        []byte(`{"username":"` + username + `","password":"` + password + `"}`),
+				Redacted:     username,
 			}
 
 			if verify {
+				if invalidHosts.Exists(username) {
+					logger.V(3).Info("Skipping invalid registry", "username", username)
+					continue EndpointLoop
+				}
```
rgmz, line 90: I don't think there's an established 'best practice' for skipping values like this.

Should it be returned as invalid? Should it be skipped altogether? etc.

ahrav: I'd lean towards skipping it. Returning it as an invalid result might not add much value, but I don't have a strong preference

ahrav (approved):

PR #3633, opened 2024-11-20
Add Flexport detector

0x1:
### Description:
This detector is for https://www.flexport.com/.
Flexport is a digital freight forwarding and logistics platform that helps businesses manage and optimize their global supply chains by providing end-to-end visibility and control over shipments.

This detector is specifically for their logistics-api. There are other APIs that we should also add.
### Checklist:


abmussani (approved):
Thanks @0x1 for the new addition in the detectors ❤️ I have a small question regarding the regex. Any reason to not put the closing word boundary ? isn't it useful to put as we already new that length of token will be 40 characters (excluding magic string).

0x1:
> Thanks @0x1 for the new addition in the detectors ❤️ I have a small question regarding the regex. Any reason to not put the closing word boundary ? isn't it useful to put as we already new that length of token will be 40 characters (excluding magic string).

good catch. i had to omit the closing boundary bc unfortunately `_` is a valid character for the token, but is also considered a boundary character by `\b`. so if we add a closing boundary it will not catch a lot of valid tokens, including the test tokens.

idk of a way to exclude `_` from the boundary set, but if we can do so we should totally add it back.

Issue #2128, opened 2023-11-27
Improve multipart secret data output


khai-tran:
## Description
For multipart secrets (containing `key_id/secret_key` or `username/password` pairs`), currently Trufflehog performs the following steps:
1.  Pre-flight check if it's worth scanning. Logic defined in func `func (s Scanner) Keywords() []string`
2.  Report `key_id` as `Raw`
3. Report `key_id + secret` as `RawV2`
4. Report line where it detects `secret` as `line`.

Example: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/twilio/twilio.go. There  are roughly > 100 multipart secret detections.

We've been relying on `Raw` and `RawV2` to re-construct the raw multipart secret to re-validate. It's useful when secret is deleted, but still active.

However, the reconstructed multipart-secret still miss the `Keywords` element, so the rescan did not trigger the same result.
Consider this Twilio case:
```
sid = key_id
secret = secret
```
Rescanning on `key_id + secret`, or `key_id + `\n` + secret` yielded no result as it's missing `sid` prefix.

### Preferred Solution
I would be nice to either:
- collect and report line number where `key_id` is detected, or
- report the Keyword that triggered the scan step in the result model, or
- expose detector's Keywords list in the protobuf model, so it can be inferred outside of the scan result

PR #3634, opened 2024-11-20
Update AzureSearchAdminKey detector

rgmz:
### Description:
There are a number of problems with this detector. This Pr attempts to fix some of the glaring issues.

e.g.,
>  Verification issue: unexpected HTTP response status 400 | {"error":{"code":"","message":"Service name must only contain lowercase letters, digits or dashes, cannot start or end with or contain consecutive dashes and is limited to 60 characters."}}

### Checklist:

### DIFF
--- a/pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
+++ b/pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
@@ -3,8 +3,8 @@ package azuresearchadminkey
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
-	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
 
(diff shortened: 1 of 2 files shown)


pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
```diff
@@ -3,10 +3,9 @@ package azuresearchadminkey
 import (
 	"context"
 	"fmt"
-	"net/http"
-	"strings"
-
 	regexp "github.com/wasilibs/go-re2"
+	"io"
```
ahrav, line 7: nit: Could we please get these import sorted.

```
import (
	"context"
	"fmt"
	"io"
	"net/http"

	regexp "github.com/wasilibs/go-re2"

)
```

pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
```diff
@@ -23,79 +22,95 @@ var _ detectors.Detector = (*Scanner)(nil)
...
 
-			s1 := detectors.Result{
+	for key := range keyMatches {
+		for service := range serviceMatches {
+			r := detectors.Result{
 				DetectorType: detectorspb.DetectorType_AzureSearchAdminKey,
-				Raw:          []byte(resMatch),
-				RawV2:        []byte(resMatch + resServiceMatch),
+				Raw:          []byte(key),
+				RawV2:        []byte(`{"service":"` + service + `","key":"` + key + `"}`),
```
ahrav, line 67: Changing the `RawV2` value will break some enterprise customers, as we currently have secrets verified using the old `key+service` format. This change would orphan previously verified credentials and create duplicates. Is there a reason we can't continue supporting the old format?

ahrav (changes requested):
Single blocking change as it pertains to reformatting `RawV2`, otherwise LGTM!

pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
rgmz: > Is there a reason we can't continue supporting the old format?

Obviously you get the final say here, but #2128 has been a PITA for me and others using OSS.

TL;DR Most of the existing `RawV2` values are suboptional because they smash together text (sometimes without a delimiter) and are devoid of context/keywords. This makes it impossible to programmatically re-construct and re-verify results.

ahrav: My guess is that when `RawV2` was proposed, we didn't account for its visibility in OSS reporting. Wonder if we can find a solution that preserves readability for OSS while avoiding duplicates on the enterprise side?

This likely requires a broader discussion, and I think @zricethezav  will have the most input, but I agree that combining multiple credentials makes them difficult to interpret and largely ineffective

rgmz:
Seems like a new/different error introduced by the azure storage change. I'll take a look when I'm back on the computer.

pkg/detectors/azuresearchadminkey/azuresearchadminkey.go
rgmz: > My guess is that when RawV2 was proposed, we didn't account for its visibility in OSS reporting. Wonder if we can find a solution that preserves readability for OSS while avoiding duplicates on the enterprise side?

I can't speak to the impact on enterprise. I can revert the change to `RawV2`, though I'd prefer to improve it (eventually).

PR #3638, opened 2024-11-20
adding archive entry paths

joeleonjr:
### Description:
Related to and inspired by #1551. This PR adds a new ExtraData field named `Archive Entry Path` that contains the relative path of the file containing a secret within an archive file. 

Note: When decompressing files that only contain one file (ex: `.gz`, `.xz`, `.lz4`, etc), it won't list the decompressed file's name due to a limitation in the archiver library we use. However, since it only results in one decompressed file, it should be self-explanatory to the user.

For example:
```
/archive.tar
-- secret.zip.gz
--/--/secret.zip
--/--/--/secret.txt
```
If there is a file in secret.txt, we'll see:

`Archive Entry Path: secret.zip.gz/secret.txt`
`File: /path/to/archive.tar`

### Checklist:


pkg/handlers/archive.go
```diff
@@ -240,6 +245,6 @@ func (h *archiveHandler) extractorHandler(dataOrErrChan chan DataOrErr) func(con
 		h.metrics.observeFileSize(fileSize)
 
 		lCtx.Logger().V(4).Info("Processed file successfully", "filename", file.Name(), "size", file.Size())
-		return h.openArchive(lCtx, depth, rdr, dataOrErrChan)
+		return h.openArchive(lCtx, depth, (archiveEntryPath + "/" + file.Name()), rdr, dataOrErrChan)
```
mcastorina, line 248: Do we care about OS-specific paths here? If so, using [filepath.Join](https://pkg.go.dev/path/filepath#Join) would be a better option.

pkg/sources/sources.go
```diff
@@ -40,6 +40,8 @@ type Chunk struct {
 	SourceMetadata *source_metadatapb.MetaData
 	// SourceType is the type of Source that produced the chunk.
 	SourceType sourcespb.SourceType
+	// HandleMetadata holds the metadata from a handler if one was used.
+	HandleMetadata map[string]string
```
mcastorina, line 44: I like this being part of the chunk.

Out of curiosity, do we have a formal definition of what a handler is? Does it take a chunk and produce multiple chunks, or is it 1:1?

pkg/engine/engine.go
```diff
@@ -1155,6 +1155,17 @@ func (e *Engine) processResult(
...
 	}
 
+	// Add in handler metadata
+	if res.ExtraData == nil && data.chunk.HandleMetadata != nil {
+		res.ExtraData = data.chunk.HandleMetadata
+	} else {
+		for k, v := range data.chunk.HandleMetadata {
+			if _, ok := res.ExtraData[k]; !ok {
+				res.ExtraData[k] = v
+			}
```
mcastorina, line 1165: A comment stating that `ExtraData` entries takes precedence would be nice.

pkg/handlers/archive.go
```diff
@@ -101,12 +101,13 @@ func (h *archiveHandler) HandleFile(ctx logContext.Context, input fileReader) ch
...
 // openArchive recursively extracts content from an archive up to a maximum depth, handling nested archives if necessary.
-// It takes a reader from which it attempts to identify and process the archive format. Depending on the archive type,
-// it either decompresses or extracts the contents directly, sending data to the provided channel.
+// It takes a string representing the path to the archive and a reader from which it attempts to identify and process the archive format.
+// Depending on the archive type, it either decompresses or extracts the contents directly, sending data to the provided channel.
 // Returns an error if the archive cannot be processed due to issues like exceeding maximum depth or unsupported formats.
 func (h *archiveHandler) openArchive(
 	ctx logContext.Context,
 	depth int,
+	archiveEntryPath string,
```
mcastorina, line 110: I think these could be combined into a single variable that's a slice of the paths. For example: `["foo", "bar"]` has `depth = 2` and `archiveEntryPath = foo/bar`. Is there a reason I'm not seeing to keep them as two distinct variables?

mcastorina (commented):
This looks pretty clean, nice job!

pkg/handlers/archive.go
joeleonjr: That's probably a cleaner way to implement it. And for the case of decompression, just add in `["foo", ".gz decompression", "bar"]`?

pkg/handlers/archive.go
joeleonjr: Definitely a better choice. Will update.

pkg/sources/sources.go
joeleonjr: From what I understand, the implementation is: input `reader` -> create as many chunks as needed and then scan for secrets.

pkg/handlers/archive.go
```diff
@@ -152,9 +153,13 @@ func (h *archiveHandler) openArchive(
...
 
-		return h.openArchive(ctx, depth+1, rdr, dataOrErrChan)
+		// Note: We're limited in our ability to add file names to the archiveEntryPath here, as the decompressor doesn't have access to a fileName value. Instead, we're adding a generic string to indicate that the file is decompressed. This could be improved.
+		if depth > 0 {
+			archiveEntryPath = archiveEntryPath + "(decompressed " + reader.format.Name() + " file)"
+		}
+		return h.openArchive(ctx, depth+1, archiveEntryPath, rdr, dataOrErrChan)
 	case archiver.Extractor:
-		err := archive.Extract(logContext.WithValue(ctx, depthKey, depth+1), reader, nil, h.extractorHandler(dataOrErrChan))
+		err := archive.Extract(logContext.WithValue(ctx, depthKey, depth+1), reader, nil, h.extractorHandler(archiveEntryPath, dataOrErrChan))
```
joeleonjr, line 162: @mcastorina do you think we need `logContext.WithValue(ctx, depthKey, depth+1)` if we're keeping track of depth with a `[]string`?

pkg/handlers/archive.go
```diff
@@ -152,9 +153,13 @@ func (h *archiveHandler) openArchive(
 		}
 		defer rdr.Close()
 
-		return h.openArchive(ctx, depth+1, rdr, dataOrErrChan)
+		// Note: We're limited in our ability to add file names to the archiveEntryPath here, as the decompressor doesn't have access to a fileName value. Instead, we're adding a generic string to indicate that the file is decompressed. This could be improved.
```
joeleonjr, line 156: @mcastorina any thoughts on how to make this a little cleaner? I imagine an edge case of a zip file with 10 `.gz` files nested and then we just say `decompressed .gz file` and it's super unclear.

PR #3642, opened 2024-11-21
Remove `AzureAD/microsoft-authentication-library-for-go`

rgmz:
### Description:
This library is no longer used for the Azure sp detector. The only remaining place was for an `errors.Is` function call, which is more than likely a mistake by automatic import.

### Checklist:


ahrav (approved):
Always fan of removing dependencies 🎉 Thanks.

PR #3643, opened 2024-11-21
Replace --only-verified with --results

rgmz:
This PR hides `--only-verified` and unhides `--results`. The latter has been in use for the better part of a year and seems to be working well behind the scenes.

Switching the documentation to an explicit `--results` flag will give people more control over what they see. 

### Checklist:


README.md
```diff
@@ -185,7 +185,7 @@ This required Cosign binary to be installed prior to running installation script
 Command:
 
 ```bash
-trufflehog git https://github.com/trufflesecurity/test_keys --only-verified
+trufflehog git https://github.com/trufflesecurity/test_keys --results=verified,unknown
```
rgmz, line 188: Including both `verified` and `unknown` introduces people to the concept beyond simply `--only-verified`. Surfacing errors is an important part of using TruffleHog effectively.

If they don't like what they see, they can easily remove `unknown`.

main.go
```diff
@@ -917,10 +917,12 @@ func parseResults(input *string) (map[string]struct{}, error) {
 	)
 	for _, value := range values {
 		switch value {
-		case "verified", "unknown", "unverified", "filtered_unverified":
+		case "verified", "unknown", "unverified", "filtered":
 			results[value] = struct{}{}
+		case "filtered_unverified":
```
rgmz, line 922: `filtered_unverified` is a bit wordy. I think `filtered` is better and fits in with the existing options.

pkg/engine/engine.go
```diff
@@ -354,6 +354,7 @@ func (e *Engine) setDefaults(ctx context.Context) {
 	}
 	e.notifyVerifiedResults = true
 	e.notifyUnverifiedResults = true
+	e.retainFalsePositives = true
```
rgmz, line 357: "Why didn't TruffleHog find my secret" is a commonly asked question (e.g., #3579). I think it makes sense to log _everything_ by default. Requiring people to know about and opt-in to a special flag can be confusing.

What do you think?

rgmz:
@dustin-decker @rosecodym Thoughts?

rosecodym (commented):
Thanks for taking a look at this! I see three changes here:
* Changing the CLI and docs to use `--results` instead of `--only-verified`
  * And including `unknown` results in the documentation
* Renaming `filtered_unverified` to `filtered`
* Disabling false positive elimination by default

I support all three of these changes, but I don't think that they should be combined into a single PR like this. I know each individual part is pretty simple, but this stuff has historically been a minefield of unexpected consequences, so I'd like to tread as lightly as possible. Would you mind splitting them up?

rosecodym:
Sorry for the delay on reviewing this - I came down with pneumonia and this kind of got lost in the shuffle :( It mostly looks good to me, but I'm wondering how `unknown` results will interact with CI, given that build agents frequently don't have as much network access as local runs. (But maybe that's an argument _for_ including `unknown` results in CI?) I've asked @zricethezav for his opinion as well.

rosecodym:
Zach says:
> Include it and if it's too noisy then we just change the documentation

rosecodym (approved):

PR #3650, opened 2024-11-21
Add Scan method to SourceManager to scan a single SourceUnit

mcastorina:
### Description:
Allows for scanning a single source unit as a job, with proper progress tracking.

### Checklist:


pkg/sources/source_manager.go
```diff
@@ -369,6 +413,42 @@ func (s *SourceManager) enumerate(ctx context.Context, source Source, report *Jo
...
+		ctx = context.WithValue(ctx, "source_id", report.SourceID)
+	}
+	if ctx.Value("source_name") == "" {
+		ctx = context.WithValue(ctx, "source_name", report.SourceName)
+	}
+	if ctx.Value("source_type") == "" {
+		ctx = context.WithValue(ctx, "source_type", source.Type().String())
+	}
+
+	// Check for the preferred method of tracking source units.
```
rosecodym, line 442: This comment confused me, because there aren't multiple methods of tracking units in play here. There's only "supported" and "not supported".

pkg/sources/source_manager.go
```diff
@@ -511,6 +591,44 @@ func (s *SourceManager) runWithUnits(ctx context.Context, source SourceUnitEnumC
 	}
 }
 
+// scanWithUnits produces chunks from a single SourceUnit.
+func (s *SourceManager) scanWithUnits(ctx context.Context, source SourceUnitChunker, report *JobProgress, unit SourceUnit) error {
```
rosecodym, line 595: This method name confuses me - it refers to unit**s** (plural), but only scans one unit, right?

pkg/sources/source_manager.go
```diff
@@ -511,6 +591,44 @@ func (s *SourceManager) runWithUnits(ctx context.Context, source SourceUnitEnumC
...
+	chunkReporter := &mgrChunkReporter{
+		unit:    unit,
+		chunkCh: make(chan *Chunk, defaultChannelSize),
+		report:  report,
+	}
+	// Produce chunks from the given unit.
+	var chunkErr error
+	go func() {
+		report.StartUnitChunking(unit, time.Now())
+		// TODO: Catch panics and add to report.
```
rosecodym, line 607: Can we ticket this? It seems important.

pkg/sources/source_manager.go
```diff
@@ -369,6 +413,42 @@ func (s *SourceManager) enumerate(ctx context.Context, source Source, report *Jo
...
+	}
+
+	// Check for the preferred method of tracking source units.
+	canUseSourceUnits := s.useSourceUnitsFunc != nil
+	if unitChunker, ok := source.(SourceUnitChunker); ok && canUseSourceUnits && s.useSourceUnitsFunc() {
+		ctx.Logger().Info("running source",
+			"with_units", true)
+		return s.scanWithUnits(ctx, unitChunker, report, unit)
+	}
+	return fmt.Errorf("source units not supported or configured for source: %s", source.Type().String())
```
rosecodym, line 449: Can we add the source name to this error message? I think it could be really helpful.

pkg/sources/source_manager.go
```diff
@@ -369,6 +413,42 @@ func (s *SourceManager) enumerate(ctx context.Context, source Source, report *Jo
 	return fmt.Errorf("Enumeration not supported or configured for source: %s", source.Type().String())
 }
 
+// scan runs a scan against a single SourceUnit as its only job. This method
+// manages the lifecycle of the provided report.
+func (s *SourceManager) scan(ctx context.Context, source Source, report *JobProgress, unit SourceUnit) error {
```
rosecodym, line 418: This is a very generic method name for what it actually does.

pkg/sources/source_manager.go
```diff
@@ -212,6 +212,50 @@ func (s *SourceManager) Enumerate(ctx context.Context, sourceName string, source
...
+			progress.ReportError(Fatal{err})
+		}
+	}()
+	return progress.Ref(), nil
+}
+
+// Scan blocks until a resource is available to run the source against a single
+// SourceUnit, then asynchronously runs it. Error information is stored and
+// accessible via the JobProgressRef as it becomes available.
+func (s *SourceManager) Scan(ctx context.Context, sourceName string, source Source, unit SourceUnit) (JobProgressRef, error) {
```
rosecodym, line 224: This is a very generic method name for what it does.

rosecodym (approved):
The bones of this look good - my only commentary is about naming and organization. It looks like you've added a single new code path but broken it up into three functions with pretty generic names. Did you consider writing one Big Function? I know it would be Quite Big but if we've got a ball of mud anyway I think it would be easier to refactor later if it's all in one method.

0x1:
> The bones of this look good - my only commentary is about naming and organization. It looks like you've added a single new code path but broken it up into three functions with pretty generic names. Did you consider writing one Big Function? I know it would be Quite Big but if we've got a ball of mud anyway I think it would be easier to refactor later if it's all in one method.

we _did_ consider one giant function. but it felt more manageable and readable to have it be decomposed.

pkg/sources/source_manager.go
0x1: correct. updated

mcastorina:
> Did you consider writing one Big Function? I know it would be Quite Big but if we've got a ball of mud anyway I think it would be easier to refactor later if it's all in one method.

In addition to what Ahmed said, this also seemed like the less error-prone way to duplicate functionality by copying the existing methods for `EnumerateAndScan` and removing functionality.

Also, I tried following the same naming pattern, of `EnumerateAndScan`, `Enumerate`, `Scan` but totally open to renaming them. These codepaths should only be temporary, but we all know how that can go sometimes.

rosecodym:
Ah, I see why they're called `scan` (and variants) now.

pkg/sources/source_manager.go
```diff
@@ -369,6 +413,42 @@ func (s *SourceManager) enumerate(ctx context.Context, source Source, report *Jo
 	return fmt.Errorf("Enumeration not supported or configured for source: %s", source.Type().String())
 }
 
+// scan runs a scan against a single SourceUnit as its only job. This method
+// manages the lifecycle of the provided report.
+func (s *SourceManager) scan(ctx context.Context, source Source, report *JobProgress, unit SourceUnit) error {
+	report.Start(time.Now())
+	defer func() { report.End(time.Now()) }()
```
ahrav, line 420: question: Is there a specific reason to use a separate function here instead of including it in the defer statement below?

PR #3654, opened 2024-11-21
Separate org listing error from finding 0 members error cases

trufflesteeeve:
### Description:
Currently, if we find 0 members when listing members in an org, _or_ if we have an error when listing members in an org, we report an error message that says we can't list the org members. Now that message is more specific to the error case vs. the zero members case, and also includes the org name in the error message.

### Checklist:


rosecodym (approved):
I personally think the mocked test is overkill, but I suppose it's pretty effective as a manager enrichment activity. Thanks for doing this!

PR #3660, opened 2024-11-23
feat(typeform): add v2 detector for new key formats

ggmolly:
### Description:
Typeform has apparently introduced a new format for API keys, according to the [API documentation](https://www.typeform.com/developers/get-started/personal-access-token/) and a key I have generated, I wrote a second version of the Typeform.

### Checklist:


pkg/detectors/typeform/v2/typeform.go
```diff
@@ -0,0 +1,96 @@
...
+type Scanner struct{}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\btfp_[a-zA-Z0-9_\.=\-]{40,59}\b`)
```
rgmz, line 25: The comment's suggestion to use a boundary character is actually bad advice, it won't match if the last character is `.=-`.

One possible alternative is to say "anything but":
```suggestion
	keyPat = regexp.MustCompile(`\btfp_[a-zA-Z0-9_\.=\-]{40,59}(?:[^a-zA-Z0-9_\.=\-]|\z)`)
```

pkg/detectors/typeform/v2/typeform.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+func (s Scanner) Version() int { return 2 }
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"typeform", "tfp_"}
```
rgmz, line 33:
```suggestion
	return []string{"tfp_"}
```

pkg/detectors/typeform/v2/typeform.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, match := range matches {
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Typeform,
+			Raw:          []byte(match),
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://api.typeform.com/me", nil)
```
rgmz, line 60: The older convention for detector logic has a bunch of issues. I'd recommend moving this into a separate function (for the sake of `defer`) and having explicit error handling + status code checks.

e.g., https://github.com/trufflesecurity/trufflehog/blob/abaacd7e8af3019d622848641a095e471b6daf74/pkg/detectors/alchemy/alchemy.go#L68

ggmolly:
Upon further testing I wasn't able to neither generate nor find a key with any of these characters: `.=-`. Therefore, I removed them from the regex entirely. However, I refactored the verification part of the code, removed the extra keyword, and squashed the changes.

pkg/detectors/typeform/v2/typeform.go
```diff
@@ -0,0 +1,107 @@
...
+	Language string `json:"language,omitempty"`
+}
+
+// FromData will find and optionally verify Typeform secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	matches := keyPat.FindAllString(dataStr, -1)
+
+	if len(matches) == 0 {
```
kashifkhan0771, line 48: Not required. If length is zero the loop will be skipped.

pkg/detectors/typeform/v2/typeform.go
```diff
@@ -0,0 +1,107 @@
...
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", secret))
+	res, err := client.Do(req)
+	if err != nil {
+		return false, response, err
+	}
+	defer res.Body.Close()
+	if err = json.NewDecoder(res.Body).Decode(&response); err != nil {
+		return false, response, err
+	}
+	if res.StatusCode >= 200 && res.StatusCode < 300 {
```
kashifkhan0771, line 95: We should be more specific for success validation status codes.

pkg/detectors/typeform/v2/typeform.go
```diff
@@ -0,0 +1,107 @@
...
+		return true, response, nil
```
kashifkhan0771, line 96: Not handling the fail validation status like 401 or 403 etc? Error should be returned in case of unexpected status code.

pkg/detectors/typeform/v2/typeform_test.go
```diff
@@ -0,0 +1,120 @@
+//go:build detectors
```
kashifkhan0771, line 1: According to new structure the `test.go` file should contain pattern test cases and the `_integration_test.go` file should have the integration test cases.

kashifkhan0771 (changes requested):

ggmolly:
1. Updated the code to consider only HTTP 200 as valid response, other codes return an error
2. Removed the redundant check
3. Changed tests to match the new structure
(squashed again)

kashifkhan0771 (approved):

rosecodym:
Thank you for this!

PR #3666, opened 2024-11-25
fixed api flash detector

kashifkhan0771:
### Description:
Fixed the API Flash detector and refactored the approach for verification.

### Checklist:

### DIFF
--- a/pkg/detectors/apiflash/apiflash.go
+++ b/pkg/detectors/apiflash/apiflash.go
@@ -24,7 +24,8 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"apiflash"}) + `\b([a-z0-9]{32})\b`)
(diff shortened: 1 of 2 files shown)


pkg/detectors/apiflash/apiflash.go
```diff
@@ -24,7 +24,8 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"apiflash"}) + `\b([a-z0-9]{32})\b`)
-	urlPat = regexp.MustCompile(detectors.PrefixRegex([]string{"apiflash"}) + `\b([a-zA-Z0-9\S]{21,30})\b`)
```
kashifkhan0771, line 27: I wasn't sure of the original intention behind detecting a URL specifically for verification, as we can use any fixed URL with an access key for this purpose. To improve flexibility, I removed the URL pattern and introduced a fixed Google URL.

zricethezav: that makes sense to me

zricethezav (approved):

PR #3687, opened 2024-11-27
[refactor] - `detectorKeywordMatcher` initialization

ahrav:
### Description:
This PR relocates the initialization of the `detectorKeywordMatcher` to the `apk` handler file. This change offers several benefits:  

1. **Efficiency**: Constructing the matcher is relatively resource-intensive and requires a decent amount of memory (~45MB). By initializing it as a singleton in the `apk` handler, we incur this cost only once, enabling reuse across all `apk` handlers throughout the program's lifecycle.  
2. **Relevance**: The matcher’s functionality is closely tied to the `apk` handler rather than the detectors, so keeping it within the `apk` codebase improves logical cohesion (imo).

Sorry @joeleonjr I totally missed the location of this logic in your earlier PR. 😭 
### Checklist:

### DIFF
--- a/pkg/engine/defaults/defaults.go
+++ b/pkg/engine/defaults/defaults.go
@@ -1,12 +1,6 @@
 package defaults
 
 import (
-	"bytes"
-	"strings"
-	"sync"
-
-	ahocorasick "github.com/BobuSumisu/aho-corasick"
-
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors/abbysale"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors/abuseipdb"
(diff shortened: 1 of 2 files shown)


pkg/engine/defaults/defaults.go
```diff
@@ -1697,60 +1691,3 @@ func DefaultDetectorTypesImplementing[T any]() map[detectorspb.DetectorType]stru
...
-				keywords = append(keywords, kwLower)
-			}
-		}
-	}
-	return keywords
-}
-
-// DefaultDetectorKeywordMatcher encapsulates the Aho-Corasick trie for keyword matching.
-type DefaultDetectorKeywordMatcher struct {
-	mu   sync.RWMutex
```
ahrav, line 1729: Removed the mutex as the `trie` is immutable after construction in our case.

ahrav:
> I agree that there's no need to incur the cost when we're not scanning APKs, but I envision needing access to this for other file types + sources in the future. Knowing that, would you still suggest locating this inside the `apk.go` file or would it make sense to put it somewhere else that is then imported into `apk.go`?

I'd argue, that until we have a clear understanding of its future use, it seems more practical to keep it confined to the APK handling logic. If we later need it elsewhere, we can relocate it to a better place where it's only initialized when necessary.

If broader use is anticipated, we could also consider redesigning the interface to make it more ergonomic.

Should the need arise to use this logic in multiple places, we could inject the same underlying trie into each location, leveraging its compatibility with our Aho-Corasick core package.

rosecodym (approved):
Do we have any written literature on why apk scanning has its own keyword matching step?

joeleonjr:
There are a few comments in the `apk.go` file, but the basic idea is this:

We're using a DEX file parser that takes the dex bytecode and makes it a bit more intelligible, ex:
```
invoke-interface method=android/os/Parcelable$Creator:createFromParcel:([Landroid/os/Parcel;]):Ljava/lang/Object; args=[2 1]
   move-result-object dst=0
   return-object value=0
   const/4 dst=0 value='0'
   return-object value=0
``` 

All const-string instructions could be potential secrets; however, due to the output of this tool, we won't necessarily have the keyword within 45 characters. So instead, we get a huge list of all potential keywords, search for the keywords in the class and method names that are close to those strings. If we find a match, we'll append that matched keyword to each string. Without this additional step, TruffleHog misses secrets in DEX code that would be obvious if we did a full APK decompiling. 

It's a very similar process to what we had to do for Postman and will likely need to do for additional data sources. It is imperfect (we're scanning a lot more chunks of data than we probably need to), but without that keyword context, TruffleHog misses secrets.

PR #3688, opened 2024-11-27
Improved cloudflarecakey detector

kashifkhan0771:
### Description:
This PR improves the following points in cloudflare ca key detector:
- Improve regex pattern
- Unique keys
- Response Handling
- Handle verification errors

### Checklist:

### DIFF
--- a/pkg/detectors/cloudflarecakey/cloudflarecakey.go
+++ b/pkg/detectors/cloudflarecakey/cloudflarecakey.go
@@ -2,8 +2,8 @@ package cloudflarecakey
 
 import (
 	"context"
+	"io"
 	"net/http"
-	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
 
@@ -20,7 +20,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"cloudflare"}) + `\b(v[A-Za-z0-9._-]{173,})\b`)
+	// origin ca keys documentation: https://developers.cloudflare.com/fundamentals/api/get-started/ca-keys/
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"cloudflare"}) + `\b(v1\.0-[A-Za-z0-9-]{171})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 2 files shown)


pkg/detectors/cloudflarecakey/cloudflarecakey.go
```diff
@@ -20,7 +20,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"cloudflare"}) + `\b(v[A-Za-z0-9._-]{173,})\b`)
+	// origin ca keys documentation: https://developers.cloudflare.com/fundamentals/api/get-started/ca-keys/
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"cloudflare"}) + `\b(v1\.0-[A-Za-z0-9-]{171})\b`)
```
kashifkhan0771, line 24: As per the [documentation](https://developers.cloudflare.com/fundamentals/api/get-started/ca-keys/) origin ca keys always start with `v1.0-`

kashifkhan0771: I generated around 5 keys and all of them had same length size of 171 excluding `v1.0-`

kashifkhan0771: I also thought to remove `cloudflare` as prefix because we have a fix `v1.0-` as prefix for each key. I would like input from others on this, if we do this or not.

pkg/detectors/cloudflarecakey/cloudflarecakey.go
```diff
@@ -54,13 +54,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+			resp, err := client.Do(req)
 			if err == nil {
-				defer res.Body.Close()
-				if res.StatusCode >= 200 && res.StatusCode < 300 {
+				defer func() {
+					_, _ = io.Copy(io.Discard, resp.Body)
+					_ = resp.Body.Close()
+				}()
+
+				if resp.StatusCode == 200 {
```
rgmz, line 65: This will silently ignore any non-200 responses.

pkg/detectors/cloudflarecakey/cloudflarecakey.go
```diff
@@ -54,13 +54,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 			req.Header.Add("user-agent", "curl/7.68.0") // pretend to be from curl so we do not wait 100+ seconds -> nice try did not work
 
-			req.Header.Add("X-Auth-User-Service-Key", resMatch)
-			res, err := client.Do(req)
+			req.Header.Add("X-Auth-User-Service-Key", caKey)
+			resp, err := client.Do(req)
 			if err == nil {
-				defer res.Body.Close()
-				if res.StatusCode >= 200 && res.StatusCode < 300 {
+				defer func() {
```
rgmz, line 60: The whole verify block should be moved into its own function. It's both cleaner and Defer in for loops isn't ideal.

rgmz (changes requested):

pkg/detectors/cloudflarecakey/cloudflarecakey.go
kashifkhan0771: Done ✅

pkg/detectors/cloudflarecakey/cloudflarecakey.go
```diff
@@ -20,7 +21,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"cloudflare"}) + `\b(v[A-Za-z0-9._-]{173,})\b`)
+	// origin ca keys documentation: https://developers.cloudflare.com/fundamentals/api/get-started/ca-keys/
+	keyPat = regexp.MustCompile(`\b(v1\.0-[A-Za-z0-9-]{171})\b`)
```
zricethezav, line 25: This is probably restrictive and unique enough to not surface many FPs but something to keep an eye on once we merge.

zricethezav (approved):

PR #3696, opened 2024-11-29
Fix/saucelabs detector

kashifkhan0771:
### Description:
This PR fixes the SauceLabs detector. It fixes the username pattern and overall logic.

### Checklist:

### DIFF
--- a/pkg/detectors/saucelabs/saucelabs.go
+++ b/pkg/detectors/saucelabs/saucelabs.go
@@ -2,18 +2,18 @@ package saucelabs
 
 import (
 	"context"
-	b64 "encoding/base64"
 	"fmt"
-	regexp "github.com/wasilibs/go-re2"
+	"io"
 	"net/http"
-	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct{
+type Scanner struct {
 	detectors.DefaultMultiPartCredentialProvider
 }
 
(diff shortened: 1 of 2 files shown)


pkg/detectors/saucelabs/saucelabs.go
```diff
@@ -24,8 +24,9 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	idPat  = regexp.MustCompile(`\b(oauth\-[a-z0-9]{8,}\-[a-z0-9]{5})\b`)
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"saucelabs"}) + `\b([a-z0-9]{8}\-[a-z0-9]{4}\-[a-z0-9]{4}\-[a-z0-9]{4}\-[a-z0-9]{12})\b`)
+	// as per signup page username can be between 2 to 70 characters and must only contain letters, numbers, or characters (_-.)
+	usernamePat = regexp.MustCompile(detectors.PrefixRegex([]string{"saucelabs", "username"}) + `\b([a-zA-Z0-9_\.-]{2,70})\b`)
```
rgmz, line 28: `[.-]\b` isn't valid

pkg/detectors/saucelabs/saucelabs.go
```diff
@@ -91,3 +78,30 @@ func (s Scanner) Type() detectorspb.DetectorType {
 func (s Scanner) Description() string {
 	return "A service for cross browser testing, API keys can create and access tests from potentially sensitive internal websites"
 }
+
+func verifySauceLabKey(ctx context.Context, client *http.Client, userName, key string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.eu-central-1.saucelabs.com/team-management/v1/teams", nil)
```
rgmz, line 83: What about us-west and us-east?

kashifkhan0771: I thought about that and the solution I have on my mind is to search for domain in chunk, if found use that else used a fixed one. What do you think? Earlier we were only using one.
Also during testing I noticed that if we hit wrong domain with valid token it gives us `403` and if token is incorrect we get `401`.

kashifkhan0771: @rgmz can you check now? If current approach looks ok to you.

rgmz: That's probably fine. I did something similar [here](https://github.com/trufflesecurity/trufflehog/blob/31b4dc2fb7633124245f30b3eb8e1b35e031ee39/pkg/detectors/sumologickey/sumologickey.go#L67).

In theory, I think you're supposed to use [`CloudProvider`](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/detectors.go#L83) / [`EndpointCustomizer`](https://github.com/trufflesecurity/trufflehog/blob/e5d138d824cac29d3e99d41ffcda9e2b7397d12d/pkg/detectors/detectors.go#L76) / [`EndpointSetter`](https://github.com/trufflesecurity/trufflehog/blob/e5d138d824cac29d3e99d41ffcda9e2b7397d12d/pkg/detectors/endpoint_customizer.go#L12); I have no clue how tf to actually use them.

kashifkhan0771: I think that is to provide users the functionality to pass their own custom endpoints for the application/software vs using the default cloud endpoint. Here we have only three fix endpoints to choose from. Correct me @mcastorina If I am wrong about the usage of EndpointCustomizer.

zricethezav (approved):

PR #3703, opened 2024-12-01
Replace `--debug` and `--trace` with a fine-grained `--log-level` flag

### LINKED ISSUE #3668: How to suppress info log output?
- [x] Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description

I use the following command to scan a single Git commit for secrets:

```sh
trufflehog git --bare --branch 0bd1cf13755a31da56ad0d4ddfaa41a67858f436 --concurrency=1 --fail --max-depth 1 --no-update file://.
```

The output looks like this:

```

2024-11-25T19:37:59Z    info-0  trufflehog      running source  {"source_manager_worker_id": "KSlmc", "with_units": true}
2024-11-25T19:37:59Z    info-0  trufflehog      scanning repo   {"source_manager_worker_id": "KSlmc", "unit_kind": "dir", "unit": ".", "repo": ".", "head": "1d85abb40defe9669b2f963f20f801e776e53ae0", "max_depth": 1}
Found unverified result 🐷🔑❓
Detector Type: Gitlab
Decoder Type: PLAIN
Raw result: REDACTED
Rotation_guide: https://howtorotate.com/docs/tutorials/gitlab/
Version: 2
Commit: 1d85abb40defe9669b2f963f20f801e776e53ae0
Email: REDACTED
File: README.md
Line: 498
Timestamp: 2024-11-25 19:37:57 +0000

2024-11-25T19:37:59Z    info-0  trufflehog      finished scanning       {"chunks": 2, "bytes": 174, "verified_secrets": 0, "unverified_secrets": 1, "scan_duration": "302.260364ms", "trufflehog_version": "3.84.1"}
```

### Preferred Solution

I would like to get rid of the `info-0  trufflehog` log output lines but I was not able to find out how. I am not sure if this is only a question about the usage or if it is a feature request. :sweat_smile: 

### Additional Context

TruffleHog version: 3.84.1

rgmz:
### Description:
This was motivated by the discussion in https://github.com/trufflesecurity/trufflehog/pull/3589.

It fixes #3668.

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -40,10 +40,12 @@ import (
 )
 
 var (
-	cli                 = kingpin.New("TruffleHog", "TruffleHog is a tool for finding credentials.")
-	cmd                 string
-	debug               = cli.Flag("debug", "Run in debug mode.").Bool()
-	trace               = cli.Flag("trace", "Run in trace mode.").Bool()
+	cli = kingpin.New("TruffleHog", "TruffleHog is a tool for finding credentials.")
+	cmd string
+	// https://github.com/trufflesecurity/trufflehog/blob/main/CONTRIBUTING.md#logging-in-trufflehog
+	logLevel            = cli.Flag("log-level", `Logging verbosity on a scale of 0 (info) to 5 (trace). Can be disabled with "-1".`).Default("0").Int()
+	debug               = cli.Flag("debug", "Run in debug mode.").Hidden().Bool()
+	trace               = cli.Flag("trace", "Run in trace mode.").Hidden().Bool()
 	profile             = cli.Flag("profile", "Enables profiling and sets a pprof and fgprof server on :18066.").Bool()
 	localDev            = cli.Flag("local-dev", "Hidden feature to disable overseer for local dev.").Hidden().Bool()
 	jsonOut             = cli.Flag("json", "Output in JSON format.").Short('j').Bool()
@@ -275,11 +277,26 @@ func init() {
 
 	cmd = kingpin.MustParse(cli.Parse(os.Args[1:]))
 
+	// Configure logging.
 	switch {
 	case *trace:
 		log.SetLevel(5)
 	case *debug:
 		log.SetLevel(2)
+	default:
+		l := int8(*logLevel)
+		if l < -1 || l > 5 {
+			fmt.Fprintf(os.Stderr, "invalid log level: %d\n", *logLevel)
+			os.Exit(1)
+		}
+
+		if l == -1 {
+			// Zap uses "5" as the value for fatal.
+			// We need to pass in "-5" because `SetLevel` passes the negation.
+			log.SetLevel(-5)
+		} else {
+			log.SetLevel(l)
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -275,11 +277,26 @@ func init() {
...
+		l := int8(*logLevel)
+		if l < -1 || l > 5 {
+			fmt.Fprintf(os.Stderr, "invalid log level: %d\n", *logLevel)
+			os.Exit(1)
+		}
+
+		if l == -1 {
+			// Zap uses "5" as the value for fatal.
+			// We need to pass in "-5" because `SetLevel` passes the negation.
+			log.SetLevel(-5)
```
rgmz, line 296: This technically works because there's no logs with this level. Perhaps there's a better way to disable logging.

main.go
```diff
@@ -275,11 +277,26 @@ func init() {
...
+	default:
+		l := int8(*logLevel)
+		if l < -1 || l > 5 {
+			fmt.Fprintf(os.Stderr, "invalid log level: %d\n", *logLevel)
+			os.Exit(1)
+		}
+
+		if l == -1 {
+			// Zap uses "5" as the value for fatal.
+			// We need to pass in "-5" because `SetLevel` passes the negation.
```
rosecodym, line 295: I have some dumb questions:
* What does "uses 5 as the value for fatal" mean here? Are fatal errors are logged at level 5?
* Why does `SetLevel` pass the negation? That seems really strange.

main.go
rosecodym: Yeah, this is unfortunate if it's the best we can do. If it is, I think it's worth leaving an apologetic, explanatory comment.

rosecodym (approved):
This is neat, thanks for doing it! I would like to also get the opinions of @zricethezav and @mcastorina.

main.go
rgmz: > What does "uses 5 as the value for fatal" mean here? Are fatal errors are logged at level 5?

[Zap's logging levels](https://github.com/uber-go/zap/blob/0ba452dbe15478739ad4bab1067706018a3062c6/zapcore/level.go#L34) use `0` for info, with negatives increasing verbosity and 1-5 decreasing verbosity.

TruffleHog only seems to use levels <= 0, so setting the level to `5` effectively prevents anything from being logged.

> Why does SetLevel pass the negation? That seems really strange.

My guess is that someone preferred the look of (+)1->5 for verbosity and not (-)1->5. 🤷‍♂️

main.go
rgmz: I'm not familiar enough with the logging setup. How would you implement `logr.Discard`?

zricethezav: I think this PR provides a cleaner interface to logging and is in a mergable state. I'm unfamiliar with `logr.Discard` as well. We can always follow-up if it turns out this change needs more.

main.go
mcastorina: It originates from the combination of `zap` as the log implementation for `logr`.

* `logr` defines higher numbers as being more verbose
* `zap` defines lower numbers as being more verbose

I left a comment about it in `SetLevel`: https://github.com/trufflesecurity/trufflehog/blob/2a010914749d701e8921f3291fb808960cae989a/pkg/log/level.go#L27-L29

rgmz: Ah, I didn't realize it was two different implementations.

Issue #3658, opened 2024-11-22
GitHub detector silently ignores rate-limit, reports valid tokens as unverified


rgmz:
The GitHub detector does not have any status code checks and minimal error handling. Consequently, important responses like [rate limit errors](https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#rate-limit-errors) are silently skipped.

https://github.com/trufflesecurity/trufflehog/blob/726a1b71efe9947d6bb775620f6fe102614bd6e7/pkg/detectors/github/v1/github_old.go#L135

This is exacerbated by the fact that:

1. the V1 detector seems to incorrectly match uppercase characters
2. The detectors do not implement entropy checks to filter obvious false-positives
3. matches are not cached and instead indiscriminately re-tested, causing an excessive amount of outgoing requests.

IMO, [any detector that contains `if err == nil`](https://github.com/search?q=repo%3Atrufflesecurity%2Ftrufflehog+%2Ferr+%3D%3D+nil%2F+path%3Apkg%2Fdetectors%2F&type=code) is defective.

**Example**

This required custom code in order to surface.
```
Found unverified result 🐷🔑❓
Verification issue: unexpected response: 403, '{
  "documentation_url": "https://docs.github.com/free-pro-team@latest/rest/overview/rate-limits-for-the-rest-api#about-secondary-rate-limits",
  "message": "You have exceeded a secondary rate limit. Please wait a few minutes before you try again. If you reach out to GitHub Support for help, please include the request ID FB58:28A6E9:8CC7F:110648:673FC6EF."
}'
Detector Type: Github
Decoder Type: BASE64
Raw result: 6DA9C48BA71FBD34DBD8E460B7CE81EE238CA410
Rotation_guide: https://howtorotate.com/docs/tutorials/github/
Version: 1
...
```

---

This is likely related to the issue @ankushgoel27 noted in Slack. 
![image](https://github.com/user-attachments/assets/f565c424-2909-4c58-8620-398c80db2676)
https://trufflehog-community.slack.com/archives/CK3UYAQF2/p1731731489015699

kashifkhan0771:
Looks like [this pull request](https://github.com/rgmz/trufflehog/pull/42) fixes this github issue.

rgmz:
> Looks like [this pull request](https://github.com/rgmz/trufflehog/pull/42) fixes this github issue.

That's another repo, no change has been made in TruffleHog.

PR #3709, opened 2024-12-02
[UPDATE] Updated plaidkey detector results, and added uniqueness check

nabeelalam:
### Description:
Updated the `plaidkey` detector.
Makes it so that only unique keys and IDs will be processed.
Also adds the `s1` object into the `results` when the `verify` flag is `false` in order to get required output in pattern test.

### Checklist:


pkg/detectors/plaidkey/plaidkey.go
```diff
@@ -37,20 +38,22 @@ func (s Scanner) Keywords() []string {
...
+		uniqueKeys[foundKey[1]] = struct{}{}
+	}
 
-		for _, idMatch := range idMatches {
-			if len(idMatch) != 2 {
-				continue
-			}
-			idresMatch := strings.TrimSpace(idMatch[1])
+	for _, foundId := range idPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueIds[foundId[1]] = struct{}{}
```
rgmz, line 49: Eliminates stuff like `PLAID_CLIENT_ID=XXXXXXXXXXXXXXXXXXXXXXXX`.
```suggestion
		id := foundId[1]
		if detectors.StringShannonEntropy(id) < 3 {
			continue
		}

		uniqueIds[id] = struct{}{}
```

pkg/detectors/plaidkey/plaidkey.go
```diff
@@ -37,20 +38,22 @@ func (s Scanner) Keywords() []string {
...
+	// find all the matching keys and ids in the data and make a unique maps for both.
+	uniqueKeys, uniqueIds := make(map[string]struct{}), make(map[string]struct{})
 
-	for _, match := range matches {
-		if len(match) != 2 {
-			continue
-		}
-		resMatch := strings.TrimSpace(match[1])
+	for _, foundKey := range keyPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueKeys[foundKey[1]] = struct{}{}
```
rgmz, line 45: Eliminates stuff like `PLAID_SECRET=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX`.
```suggestion
		key := foundKey[1]
		if detectors.StringShannonEntropy(key) < 3 {
			continue
		}

		uniqueKeys[key] = struct{}{}
```

pkg/detectors/plaidkey/plaidkey.go
rgmz, line 1: The detector should be updated to handle failures and check for a specific status code. Anything that uses `if err == nil` or `if res.StatusCode >= 200 && res.StatusCode < 300` is defective IMO (#3658).

See the implementation [here](https://github.com/trufflesecurity/trufflehog/blob/31b4dc2fb7633124245f30b3eb8e1b35e031ee39/pkg/detectors/alchemy/alchemy.go#L51-L59).

pkg/detectors/plaidkey/plaidkey.go
nabeelalam: Thanks @rgmz! I didn't think of this, but any detectors I may work on later I'll definitely add this!

kashifkhan0771 (approved):

PR #3727, opened 2024-12-03
Add analysis info for GCP creds

bill-rich:
### Description:

### Checklist:

### DIFF
--- a/pkg/detectors/gcp/gcp.go
+++ b/pkg/detectors/gcp/gcp.go
@@ -113,7 +113,13 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				"rotation_guide": "https://howtorotate.com/docs/tutorials/gcp/",
 				"project":        creds.ProjectID,
 			},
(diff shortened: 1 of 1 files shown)


pkg/detectors/gcp/gcp.go
```diff
@@ -113,6 +113,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				"rotation_guide": "https://howtorotate.com/docs/tutorials/gcp/",
 				"project":        creds.ProjectID,
 			},
+			AnalysisInfo: map[string]string{"user": creds.ClientEmail},
```
dustin-decker, line 116: Usually this would be a service account. Maybe we should stick with gcp terminology and call it `principal`?

https://cloud.google.com/iam/docs/principal-identifiers

Also, what is this field for in the detector result?

bill-rich: Good call. There is a type usually in the full key file. I'll add that if it exists. Otherwise the plan was to check both.

dustin-decker (approved):

PR #3734, opened 2024-12-05
updated twilio detector

kashifkhan0771:
### Description:
This PR fix the CSM#742
Removed restrictive identifier pattern which is not necessary for verification
Added twilio as keyword
Improved overall logic and verification process 

### Checklist:

### DIFF
--- a/pkg/detectors/twilio/twilio.go
+++ b/pkg/detectors/twilio/twilio.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 
 	regexp "github.com/wasilibs/go-re2"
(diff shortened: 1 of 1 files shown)


pkg/detectors/twilio/twilio.go
```diff
@@ -71,46 +73,20 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-					}
-				} else {
-					s1.SetVerificationError(err, key)
+
+				if s1.Verified {
+					s1.AnalysisInfo = map[string]string{"key": key, "sid": sid}
 				}
 			}
 
-			if len(keyMatches) > 0 {
```
kashifkhan0771, line 111: I am not sure why this check was added initially as if we are inside `keyMatches` loop it will always be true

zricethezav (approved):
great work!

PR #3769, opened 2024-12-12
updated and fixed typeform detectors

kashifkhan0771:
### Description:
This PR improve overall logic of TypeForm detectors and fixes it's integration test cases.
Screenshot from 2024-12-12 16-21-36

### Checklist:

### DIFF
--- a/pkg/detectors/typeform/v2/typeform_integration_test.go
+++ b/pkg/detectors/typeform/v2/typeform_integration_test.go
@@ -1,3 +1,6 @@
+//go:build detectors
+// +build detectors
+
 package typeform
 
 import (
--- a/pkg/detectors/typeform/v2/typeform_integration_test.go
+++ b/pkg/detectors/typeform/v2/typeform_integration_test.go
@@ -1,6 +1,3 @@
-//go:build detectors
-// +build detectors
-
 package typeform
 
 import (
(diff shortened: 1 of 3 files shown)


pkg/detectors/typeform/v2/typeform_integration_test.go
```diff
@@ -86,40 +89,6 @@ func TestTypeform_FromChunk(t *testing.T) {
 			wantErr:             false,
 			wantVerificationErr: false,
 		},
-		{
-			name: "found, would be verified if not for timeout",
```
kashifkhan0771, line 90: Even with 1 Microsecond timeout this test case was failing with no error 😄 (🥲) So I removed them.

pkg/detectors/typeform/v2/typeform_integration_test.go
```diff
@@ -1,6 +1,3 @@
-//go:build detectors
```
ahrav, line 1: question: Have we started removing the build tag from detector integration tests, or is it only being removed from non-integration tests? Sorry if i'm late to the part on this. 😅

ahrav (approved):
Single non-blocking comment, but otherwise LGTM! Thanks for getting this cleaned up. ❤️

pkg/detectors/typeform/v2/typeform_integration_test.go
kashifkhan0771: Ah good catch 😃 I removed it accidentally

ahrav (approved):

PR #3784, opened 2024-12-16
Fix Azure Devops detector

### LINKED ISSUE #3680: AzureDevopsPersonalAccessToken detector: `unexpected HTTP response status 404`
Example:
```
Found unverified result 🐷🔑❓
Verification issue: unexpected HTTP response status 404
Detector Type: AzureDevopsPersonalAccessToken
Decoder Type: PLAIN
Raw result: mi75vq7fet2kokgq5uyk252zlpbmq32ewhqtymkma5tvfzl4iykqDefaultEndpointsProtocol
Commit: abcbf10e5fd621c69d27f0bd8a460ac0e5433974
Email: Jack G Kafaty <jackgkafaty@github.com>
File: Example03_Csharp/.env
Line: 8
Link: https://github.com/advanced-security/codeql_container_example/blob/abcbf10e5fd621c69d27f0bd8a460ac0e5433974/Example03_Csharp/.env#L8
Repository: https://github.com/advanced-security/codeql_container_example.git
Timestamp: 2023-06-30 17:04:31 +0000
```

rgmz:
### Description:
This fixes #3680.

### Checklist:

### DIFF
--- a/pkg/detectors/azuredevopspersonalaccesstoken/azuredevopspersonalaccesstoken.go
+++ b/pkg/detectors/azuredevopspersonalaccesstoken/azuredevopspersonalaccesstoken.go
@@ -1,13 +1,17 @@
-package azuredevopspersonalaccesstoken
+package azure_devops
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/cache/simple"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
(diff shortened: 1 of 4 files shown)


kashifkhan0771 (approved):

rosecodym:
@rgmz thanks for this! Would you mind providing a brief summary of the changes you made to the logic? The simultaneous code reorganization has made me unable to find a useful diff. (Or maybe you could just tell the right buttons to push to get one :/ )

rgmz:
It seems that GitHub isn't smart enough to track the rename; oddly, Git seems to track it just fine.

TLDR

- Fix the keywords
- Fix the pattern (both key and organization)
- Filter obviously invalid matches
- Fix unhandled status codes
- Add extradata

rosecodym:
> It seems that GitHub isn't smart enough to track the rename; oddly, Git seems to track it just fine.

Huh, weird. It looks like pushed a new version that renamed the package but not the files - was that intentional?

Regardless, thanks for looking into this. It looks like there's some good cleanup in here. I have two specific concerns, though:
* The invalid org cache introduces state to the detector, which in turn suggests that someday it might need lifecycle management that nobody's going to realize. To be clear, I don't think this necessarily going to cause an acute problem _now_, but I'm worried about a hidden maintenance cost. How much does this cache speed detection up? My experience is that verification in general is rarely/never the bottleneck on a scan, so I'm wondering if it's possible to remove the cache in order to simplify the implementation without a performance cost.
* Changing `Raw` or `RawV2` "orphans" secrets that are tracked externally. Do you have any guesses about _why_ the original implementation was the way it was? Specifically, can we get away with just making a V2 of this detector without feeling ashamed of ourselves?

rosecodym (commented):
Actually, I think that we can retain the existing `Raw`/`RawV2` behavior by just:
* Making `dev\.azure\.com/` non-capturing in its regex
* Reverting to the existing `Raw`/`RawV2` concatenation logic

rgmz:
> * Specifically, can we get away with just making a V2 of this detector without feeling ashamed of ourselves?

This might be the ideal change. The existing RawV2 is a hiderance for OSS users because of reasons mentioned in https://github.com/trufflesecurity/trufflehog/pull/3634#discussion_r1851024698.

rgmz:
> The invalid org cache introduces state to the detector, which in turn suggests that someday it might need lifecycle management that nobody's going to realize. To be clear, I don't think this necessarily going to cause an acute problem now, but I'm worried about a hidden maintenance cost. How much does this cache speed detection up? My experience is that verification in general is rarely/never the bottleneck on a scan, so I'm wondering if it's possible to remove the cache in order to simplify the implementation without a performance cost.

In my view, we should not be flooding services with requests for resources we _know_ are invalid. Doing so can result in [providers blocking TruffleHog](https://github.com/trufflesecurity/trufflehog/blob/2b43d341d370aacf772d055d855f06b9af6ab6ef/pkg/engine/defaults/defaults.go#L842) or [legitimate requests failing due to ratelimiting](https://github.com/trufflesecurity/trufflehog/issues/3658) / captcha (e.g., BrowserStack). It also hampers throughput; my home Internet is DSL and TruffleHog is largely unusable because of how noisy it is.

Perhaps it'll be pointless once granular result caching is introduced. For now, the pros heavily outweigh potential cons IMO.

rgmz:
> Actually, I think that we can retain the existing `Raw`/`RawV2` behavior by just:
> 
>     * Making `dev\.azure\.com/` non-capturing in its regex
> 
>     * Reverting to the existing `Raw`/`RawV2` concatenation logic

@rosecodym @kashifkhan0771 Is the only change required for this to be merged no. 2? I'm not sure what no. 1 refers to.

PR #3792, opened 2024-12-18
[fix] - use typed const

### LINKED ISSUE #3791: Build fails on 32-bit systems: cannot use defaultMaxDiffSize (untyped int constant 2147483648) as int value in struct literal (overflows)
Build fails.

Errors:
```
pkg/gitparse/gitparse.go:213:18: cannot use defaultMaxDiffSize (untyped int constant 2147483648) as int value in struct literal (overflows)
pkg/gitparse/gitparse.go:214:18: cannot use defaultMaxCommitSize (untyped int constant 2147483648) as int value in struct literal (overflows)
```

[log](https://pkg-status.freebsd.org/beefy17/data/main-i386-default/p545eb5d3ac42_sda2c88dfcf4/logs/trufflehog-3.86.1.log)

Version: 3.86.1
go1.21

ahrav:
### Description:
This PR uses typed int64 constants to prevent overflow on 32-bit platforms for `defaultMaxSize` and `defaultCommitSize`. This can happen because integer constants in Go are initially untyped and can hold arbitrarily large values. The overflow issue arises when the compiler tries to assign that constant to a typed entity that cannot represent its value.
### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -25,10 +25,10 @@ const (
 	defaultDateFormat = "Mon Jan 2 15:04:05 2006 -0700"
 
 	// defaultMaxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
-	defaultMaxDiffSize = 2 * 1024 * 1024 * 1024 // 2GB
+	defaultMaxDiffSize int64 = 2 * 1024 * 1024 * 1024 // 2GB
 
 	// defaultMaxCommitSize is the maximum size for a commit. Larger commits will be cut off.
-	defaultMaxCommitSize = 2 * 1024 * 1024 * 1024 // 2GB
+	defaultMaxCommitSize int64 = 2 * 1024 * 1024 * 1024 // 2GB
 )
 
 // contentWriter defines a common interface for writing, reading, and managing diff content.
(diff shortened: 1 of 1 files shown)


abmussani (approved):
Nice ❤️

pkg/gitparse/gitparse.go
```diff
@@ -210,8 +210,8 @@ type Option func(*Parser)
 func NewParser(options ...Option) *Parser {
 	parser := &Parser{
 		dateFormat:    defaultDateFormat,
-		maxDiffSize:   defaultMaxDiffSize,
-		maxCommitSize: defaultMaxCommitSize,
+		maxDiffSize:   int(defaultMaxDiffSize),
+		maxCommitSize: int(defaultMaxCommitSize),
```
abmussani, line 214: why narrow casting here ? Gopher won't throw error but might loss data silently.

ahrav: Ah, yep. Good call, this will actually break on 32-bit machines. I'll create a new PR to address this. Thanks for keeping me honest.

abmussani: I should have catch it before approving the PR. Sorry about that.

ahrav: My eyes just glossed over the fact that I was undoing the earlier type change 🤣 
follow-up [PR](https://github.com/trufflesecurity/trufflehog/pull/3793)

PR #3801, opened 2024-12-19
Implement verification cache

rosecodym:
### Description:
This PR introduces a cache that allows the scanner to avoid emitting multiple requests to verify the same credential. In practice, it doesn't seem to reduce scan time at all, but it does seem to reduce the number of calls to `FromData` rather drastically.

The cache is implemented as an opt-out feature that can be disabled with a new CLI flag. If we don't like this, we can change it.

The metrics collection hopefully isn't _too_ architecture-astronauty; I wanted to create something useful here that could also accommodate future Prometheus configuration without making the implementation all stupid.

Here's the result of scanning [rails](https://github.com/rails/rails) a few times, with and without caching:
```
with caching
{"chunks": 497160, "bytes": 198670727, "verified_secrets": 0, "unverified_secrets": 100, "scan_duration": "29.301196125s", "trufflehog_version": "dev", "verification_caching": {"Hits":50,"Misses":60,"HitsWasted":0,"AttemptsSaved":50,"VerificationTimeSpentMS":48020}}
{"chunks": 497160, "bytes": 198670727, "verified_secrets": 0, "unverified_secrets": 100, "scan_duration": "27.765202709s", "trufflehog_version": "dev", "verification_caching": {"Hits":50,"Misses":60,"HitsWasted":0,"AttemptsSaved":50,"VerificationTimeSpentMS":41405}}

without caching
{"chunks": 497160, "bytes": 198670727, "verified_secrets": 0, "unverified_secrets": 100, "scan_duration": "28.843573s", "trufflehog_version": "dev", "verification_caching": {"Hits":0,"Misses":0,"HitsWasted":0,"AttemptsSaved":0,"VerificationTimeSpentMS":56733}}
{"chunks": 497160, "bytes": 198670727, "verified_secrets": 0, "unverified_secrets": 100, "scan_duration": "27.674744458s", "trufflehog_version": "dev", "verification_caching": {"Hits":0,"Misses":0,"HitsWasted":0,"AttemptsSaved":0,"VerificationTimeSpentMS":53908}}
```
The caching doesn't speed the scans up, which is expected, because scans aren't generally bottlenecked by verification. However, it does reduce the number of verification requests, which is good for avoiding account lockouts and stuff like that.

Cached results get this cool new text:
image

### Checklist:
* [x] Double-check names and comments and suchlike
* [x] Put some metrics in this PR description


rosecodym:
> i understand it doesnt verify the same credential but I hope it does report it as verified(if verified) in the output.

Yep, that's the idea! If the credential is cached as verified, then the output will report it as verified.

pkg/verificationcache/metrics_reporter.go
```diff
@@ -0,0 +1,28 @@
+package verificationcache
+
+import "time"
+
+// MetricsReporter is an interface used by a verification cache to report various metrics related to its operation.
+// Implementations must be thread-safe.
+type MetricsReporter interface {
```
zricethezav, line 7: **Question**: @rosecodym any reason we need this to be an interface? I'm seeing `InMemoryMetrics` as the only struct that implements this interface.

zricethezav (approved):
lgtm, just one non-blocking question

pkg/verificationcache/metrics_reporter.go
rosecodym: Oh, good question. The plan is to use this seam to plug in a Prometheus-based reporter in enterprise. Rather than drape those metrics all over the OSS codebase I opted to create this interface, which looks silly on its own (as you've pointed out) but I hope will save some headaches in the long run.

mcastorina:
This looks great!

pkg/verificationcache/metrics_reporter.go
dustin-decker: We have prometheus metrics in the OSS codebase elsewhere, it doesn't cause any issues.

bugbaba:
This verification cache is not working on all instances.

The core reason being since the multiple files are processed concurrently by different workers that by the time one of them completes verification and adding the cache others workers have already started the verification process as well.

As a result if same key is inside two separate files then key is still being verified twice instead of only once.

The solution is letting only worker access the cache at a time by locking it and make other workers wait. But I am assuming this will increase scan timing considering the workers waiting for each other.

PR #3813, opened 2024-12-23
Fake test cases sample

kashifkhan0771:
### Description:
This PR serves as a sample for new fake test cases. It helps us verify how our detector behaves when receiving specific responses from the API. While the integration test cases work similarly, they rely on real credentials pulled from GCP, which prevents non-Truffle members from running them locally.

Please review and feel free to share any ideas or suggestions in the comments! 🚀 

Do not Merge ❌ unless approved by @zricethezav @ahrav @abmussani and other key members.

### Checklist:


rgmz:
Seems like it would be easier to use [`ConstantResponseHttpClient`](https://github.com/trufflesecurity/trufflehog/blob/ddc015e5ed99942b2253d8ea16a0586a01ef2ab1/pkg/detectors/zulipchat/zulipchat_integration_test.go#L111-L113).

rgmz (commented):
Frankly, I don't see value in this beyond testing rare/complicated edge cases.

> While the integration test cases work similarly, they rely on real credentials pulled from GCP, which prevents non-Truffle members from running them locally.

There is no substitution for integration tests. [GitHub Actions should run the integration tests](https://github.com/trufflesecurity/trufflehog/actions/workflows/detector-tests.yml) but this has been broken for a while.

kashifkhan0771:
> Frankly, I don't see value in this beyond testing rare/complicated edge cases.
This will become helpful when we will implement the detectors as config feature I guess.

PR #3815, opened 2024-12-23
fixed uri regex issue

### LINKED ISSUE #3686: URI detector: regex matches invalid characters, greedily overlaps with other URIs
**Example**
```
<p><a href="http://username:password@127.0.0.1">http://username:password@127.0.0.1</a></p>
```

This matches `http://username:password@127.0.0.1">http://username:password@127.0.0.1` and not `http://username:password@127.0.0.1` twice.

```
Found unverified result 🐷🔑❓
Verification issue: lookup 127.0.0.1">http: no such host
Detector Type: URI
Decoder Type: PLAIN
Raw result: http://username:password@127.0.0.1">http://username:password@127.0.0.1
Commit: 95a76587eea5863480ff23e34bd415d8ec0d6b2a
File: test/mattermost.txt
Line: 151
Link: https://github.com/mattermost/commonmark.js/blob/95a76587eea5863480ff23e34bd415d8ec0d6b2a/test/mattermost.txt#L151
```

kashifkhan0771:
### Description:
This PR fixes github issue #3686
Screenshot from 2024-12-23 19-05-15

### Checklist:


pkg/detectors/uri/uri.go
```diff
@@ -23,7 +23,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\w-\.]{3,50}:([\w-\.]{3,50})@[-.%\w\/:]+\b`)
```
kashifkhan0771, line 26: `\S` matches any non-whitespace character, which is very broad. Instead, we are now using `\w`, which matches `[A-Za-z0-9_]`, and extending it by adding a few special characters to suit our needs.

abmussani, line 26: In my opinion, passwords in URIs can consist of any non-whitespace characters. Using \w limits the match to a specific set of characters, which might exclude valid ones. Are you noticing any detection issues when using \S instead?

abmussani: e.g. [http://username:p%40ssword@127.0.0.1"](http://username:p%40ssword@127.0.0.1) Will not be detected when using `\w`.

kashifkhan0771: Yes, actually as the scenario mentioned in the linked issue, it matches some special chars like `"` which it should not I guess. I can add more special characters like `%$^...` in the set to expand the functionality.

kashifkhan0771: I add some more possible special characters for password

pkg/detectors/uri/uri.go
```diff
@@ -23,7 +23,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\w-\.]{3,50}:([\w-\.%$^&#@]{3,50})@[-.%\w\/:]+\b`)
```
rgmz, line 26: A good source of truth is [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1). An unencoded `@` isn't a valid password character because it's the delimiter between `userinfo` and `host`; likewise, `:` isn't a valid username character.

The right-hand side of `@` is also problematic. e.g., [`%` and `/` aren't valid characters for a host](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2).

pkg/detectors/uri/uri_test.go
```diff
@@ -30,6 +31,11 @@ func TestURI_Pattern(t *testing.T) {
 			input: fmt.Sprintf("%s token = '%s'", keyword, validPattern),
 			want:  []string{validPattern},
 		},
+		{
+			name:  "valid pattern - capture two outputs",
+			input: fmt.Sprintf("%s token = '%s'", keyword, validPattern2),
+			want:  []string{"http://username:password@127.0.0.1", "http://username:password@127.0.0.1"},
```
rgmz, line 37: While not related to #3815, the results should be deduplicated.

```suggestion
			want:  []string{"http://username:password@127.0.0.1"},
```

kashifkhan0771: Done

pkg/detectors/uri/uri.go
kashifkhan0771: I improved the regex and removed the invalid chars as you mentioned.

ahrav (approved):

pkg/detectors/uri/uri.go
```diff
@@ -23,7 +23,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\b(?:https?:\/\/)?[\w-\.$~!]{3,50}:([\w-\.%$^&#]{3,50})@[-.\w]+\b`)
```
rgmz, line 26: This is missing a large number of valid characters for usernames and passwords. The host pattern is also still fairly permissive and would match things that could never be valid, e.g. `@----__-2as-2`.

rgmz (changes requested):

pkg/detectors/uri/uri.go
kashifkhan0771: How about something like:
```regex
\b(?:https?:\/\/)?[\w-\.$~!&'()*+,;=:%-]{3,50}:([\w-\.%$^#&'()*+,;=:%-]{3,50})@[a-zA-Z0-9.-]+(?:\.[a-zA-Z]{2,})?\b
```
Added additional valid characters and fixed the host pattern too.

kashifkhan0771: Example: https://regex101.com/r/qBF4nS/1

rgmz: That's looking a bit better. Some notes:

1. The scheme prefix shouldn't be optional
2. `:` isn't a valid username character
3. Username isn't always required ([example](https://stackoverflow.com/a/68783135))
4. The username and password patterns are both missing special characters. It would be pragmatic to add [all applicable special characters from `[[:graph:]]`](https://pkg.go.dev/regexp/syntax), and remove them later if they're causing issues. e.g.,
  ```regexp
  \bhttps?:\/\/[\w!#$%&()*+,\-./;<=>?@[\\\]^_{|}~]{0,50}:([\w!#$%&()*+,\-./:;<=>?[\\\]^_{|}~]{3,50})@[a-zA-Z0-9.-]+(?:\.[a-zA-Z]{2,})?\b
  ```
5. The pattern needs to be able to detect port as well as path

kashifkhan0771: Updated with port detection

pkg/detectors/uri/uri.go
```diff
@@ -30,7 +30,7 @@ var _ interface {
 } = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\bhttps?:\/\/[\w!#$%&()*+,\-./;<=>?@[\\\]^_{|}~]{0,50}:([\w!#$%&()*+,\-./:;<=>?[\\\]^_{|}~]{3,50})@[a-zA-Z0-9.-]+(?:\.[a-zA-Z]{2,})?(?:[\d:]{1,5})?\b`)
```
rgmz, line 33:
```suggestion
	keyPat = regexp.MustCompile(`\bhttps?:\/\/[\w!#$%&()*+,\-./;<=>?@[\\\]^_{|}~]{0,50}:([\w!#$%&()*+,\-./:;<=>?[\\\]^_{|}~]{3,50})@[a-zA-Z0-9.-]+(?:\.[a-zA-Z]{2,})?(?::\d{1,5})?\b`)
```

pkg/detectors/uri/uri.go
```diff
@@ -30,7 +30,7 @@ var _ interface {
 } = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`\b(?:https?:)?\/\/[\S]{3,50}:([\S]{3,50})@[-.%\w\/:]+\b`)
+	keyPat = regexp.MustCompile(`\bhttps?:\/\/[\w!#$%&()*+,\-./;<=>?@[\\\]^_{|}~]{0,50}:([\w!#$%&()*+,\-./:;<=>?[\\\]^_{|}~]{3,50})@[a-zA-Z0-9.-]+(?:\.[a-zA-Z]{2,})?(?::\d{1,5})?\b`)
```
rgmz, line 33: Hmm, this no longer matches paths after host/port. That's probably something worth keeping.

ahrav:
@kashifkhan0771 Is this ready to be merged? Wanted to double check before merging. Thanks.

kashifkhan0771:
> @kashifkhan0771 Is this ready to be merged? Wanted to double check before merging. Thanks.

It's done from myside. @rgmz need to do a final review of the regex.

PR #3816, opened 2024-12-24
fixed github issue 3774 for custom detector secret size

### LINKED ISSUE #3774: Last matched custom regex incorrect
### TruffleHog Version

trufflehog 3.86.1

### Trace Output

https://gist.github.com/cfkirby/9116ab886ed2dc1c743a7164d3084744

### Expected Behavior

The match on line 2 should have caught the entire string, i.e.:
```
secret2abcdefghijmagic-bin64-stringabcdefghij1abcdefghij2abcdefghij3abcdefghij4abcdefghij5abcdefghij6abcdefghij7abcdefghij8abcdefghij9abcdefghij10abcdefghij11abcdefghij12abcdefghij13abcdefghij14abcdefghij15abcdefghij16abcdefghij17abcdefghij18abcdefghij19abcdefghij20abcdefghij21abcdefghij22abcdefghij23abcdefghij24abcdefghij25abcdefghij26abcdefghij27abcdefghij28abcdefghij29abcdefghij1abcdefghij2abcdefghij3abcdefghij4abcdefghij5abcdefghij6abcdefghij7abcdefghij8abcdefghij9abcdefghij10abcdefghij11abcdefghij12abcdefghij13abcdefghij14abcdefghij15abcdefghij16abcdefghij17abcdefghij18
```

### Actual Behavior

The match on line 2 got cut off early:
```
secret2abcdefghijmagic-bin64-stringabcdefghij1abcdefghij2abcdefghij3abcdefghij4abcdefghij5abcdefghij6abcdefghij7abcdefghij8abcdefghij9abcdefghij10abcdefghij11abcdefghij12abcdefghij13abcdefghij14abcdefghij15abcdefghij16abcdefghij17abcdefghij18abcdefghij19abcdefghij20abcdefghij21abcdefghij22abcdefghij23abcdefghij24abcdefghij25abcdefghij26abcdefghij27abcdefghij28abcdefghij29abcdefghij1abcdefghij2abcdefghij3abcdefghij4abcdefghij5abcdefghij6abcdefghij7abcdefghij8abcdefghij9abcdefghij10abcdefghij11abcdefghij12abcdefghij13abcdefgh
```
The full string length is 581 characters, but the matched string was only 529. 

### Steps to Reproduce

1. Create the files shown in the gist and run the command demonstrated in the gist

## Environment
 * OS: MacOS
 * Sequoia 15.1.1

## Additional Context

This seems to only happen with the last match in the file and only if the string is rather large (approximately more than 520 characters, not sure the exact amount). Any prior tokens match correctly. It doesn't seem to be important that the last match is on the last line of the file either, there can be junk text on the last few lines but the last match will still have a problem.

My custom regex is more complicated than this, but I was able to reproduce the bug I was seeing with this over-simplified example. This is a pretty annoying bug because I know my custom regex is going to report many false negatives currently.

kashifkhan0771:
### Description:
This PR adds a custom secret max size to `1000` for custom detector, so it can detector large secrets.
This is related to issue #3774

Issue Example Fix:
Screenshot from 2024-12-24 15-03-09

### Checklist:

### DIFF
--- a/pkg/custom_detectors/custom_detectors.go
+++ b/pkg/custom_detectors/custom_detectors.go
@@ -32,6 +32,7 @@ type CustomRegexWebhook struct {
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*CustomRegexWebhook)(nil)
 var _ detectors.CustomFalsePositiveChecker = (*CustomRegexWebhook)(nil)
+var _ detectors.MaxSecretSizeProvider = (*CustomRegexWebhook)(nil)
 
 // NewWebhookCustomRegex initializes and validates a CustomRegexWebhook. An
 // unexported type is intentionally returned here to ensure the values have
@@ -113,6 +114,11 @@ func (c *CustomRegexWebhook) FromData(ctx context.Context, verify bool, data []b
 
 func (c *CustomRegexWebhook) IsFalsePositive(_ detectors.Result) (bool, string) {
 	return false, ""
+}
+
+// custom max size for custom detector
+func (c *CustomRegexWebhook) MaxSecretSize() int64 {
+	return 1000
 }
 
 func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string][]string, verify bool, results chan<- detectors.Result) error {


ahrav:
I think this solution works for now. It would be helpful if implementers of custom detectors could identify that they are using this interface and also have a way to set the secret size in the config. However, that's likely a separate discussion, which is why I believe this approach is appropriate. Well done!

ahrav (approved):

PR #3823, opened 2024-12-30
fixed github issue 3819 for endpoint customizer tests

### LINKED ISSUE #3819: EmbeddedEndpointSetter test failing
> @mcastorina This test is failing; not sure why.
>
> Edit: it looks like #3308 changed the logic in a way that subtly broke the test.
>
> ```
> --- FAIL: TestEmbeddedEndpointSetter (0.00s)
>     endpoint_customizer_test.go:12: 
>                 Error Trace:    /tmp/trufflehog/pkg/detectors/endpoint_customizer_test.go:12
>                 Error:          Not equal: 
>                                 expected: []string{"baz"}
>                                 actual  : []string(nil)
>                                 
>                                 Diff:
>                                 --- Expected
>                                 +++ Actual
>                                 @@ -1,4 +1,2 @@
>                                 -([]string) (len=1) {
>                                 - (string) (len=3) "baz"
>                                 -}
>                                 +([]string) <nil>
>                                  
>                 Test:           TestEmbeddedEndpointSetter
> FAIL
> ```
> https://github.com/trufflesecurity/trufflehog/blob/bf6f3136971c9d21c6ef5a6d9bd1d11520102c30/pkg/detectors/endpoint_customizer_test.go#L12

_Originally posted by @rgmz in https://github.com/trufflesecurity/trufflehog/issues/3773#issuecomment-2541588054_

----

Unclear whether the feature is broken or the test is broken.

kashifkhan0771:
### Description:
This PR fixes the github issue #3819 
Screenshot from 2024-12-30 13-16-48

### Checklist:

### DIFF
--- a/pkg/detectors/endpoint_customizer_test.go
+++ b/pkg/detectors/endpoint_customizer_test.go
@@ -9,8 +9,19 @@ import (
 func TestEmbeddedEndpointSetter(t *testing.T) {
 	type Scanner struct{ EndpointSetter }
 	var s Scanner
+	// use foundendpoint to set the "baz" in endpoints
+	s.useFoundEndpoints = true
 	assert.Equal(t, []string{"baz"}, s.Endpoints("baz"))
+	// setting "foo" and "bar" as configured endpoint
 	assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))
+	// return error as no endpoints are passed
 	assert.Error(t, s.SetConfiguredEndpoints())
-	assert.Equal(t, []string{"foo", "bar"}, s.Endpoints("baz"))
+	// as use foundendpoint is true, "baz" will be added in the endpoints list
+	assert.Equal(t, []string{"foo", "bar", "baz"}, s.Endpoints("baz"))
+	// setting cloudendpoints
+	s.useCloudEndpoint = true
+	// cloudendpoint must be set in order to be added in the list along with useCloudEndpoint as true
+	s.cloudEndpoint = "test"
+	// reason for "foo" and "bar" here is that they are already configured
+	assert.Equal(t, []string{"foo", "bar", "test"}, s.Endpoints())
 }
--- a/pkg/detectors/endpoint_customizer_test.go
+++ b/pkg/detectors/endpoint_customizer_test.go
@@ -8,29 +8,53 @@ import (
 
 func TestEmbeddedEndpointSetter(t *testing.T) {
 	type Scanner struct{ EndpointSetter }
+
 	var s Scanner
-	// set useFoundEndpoints to true to add the "baz" in endpoints
-	s.useFoundEndpoints = true
-	assert.Equal(t, []string{"baz"}, s.Endpoints("baz"))
-	// setting "foo" and "bar" as configured endpoint
-	assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))
-	// return error as no endpoints are passed
-	assert.Error(t, s.SetConfiguredEndpoints())
-	// as useFoundEndpoints is true, "baz" will be added in the endpoints list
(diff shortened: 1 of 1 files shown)


abmussani (approved):

pkg/detectors/endpoint_customizer_test.go
```diff
@@ -9,8 +9,28 @@ import (
 func TestEmbeddedEndpointSetter(t *testing.T) {
 	type Scanner struct{ EndpointSetter }
 	var s Scanner
+	// set useFoundEndpoints to true to add the "baz" in endpoints
```
ahrav, line 12: suggestion/question: Would it make sense to split this into subtests or separate tests? The EndpointSetter has always been a bit confusing for me, so clearer labels on the tests might make them easier to grok.

Ex:
```
func TestEmbeddedEndpointSetter(t *testing.T) {
    t.Run("useFoundEndpoints is true", func(t *testing.T) {
        var s Scanner
        s.useFoundEndpoints = true

        // "baz" is passed to Endpoints, should appear in the result
        assert.Equal(t, []string{"baz"}, s.Endpoints("baz"))
    })

    t.Run("setting configured endpoints", func(t *testing.T) {
        var s Scanner

        // Setting "foo" and "bar"
        assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))

        // Returning error because no endpoints are passed
        assert.Error(t, s.SetConfiguredEndpoints())
    })

    t.Run("useFoundEndpoints adds new endpoints", func(t *testing.T) {
        var s Scanner
        s.useFoundEndpoints = true

        // "foo" and "bar" are configured
        assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))
        // "baz" is added because useFoundEndpoints is true
        assert.Equal(t, []string{"foo", "bar", "baz"}, s.Endpoints("baz"))
    })

    t.Run("useCloudEndpoint is true", func(t *testing.T) {
        var s Scanner
        s.useCloudEndpoint = true
        s.cloudEndpoint = "test"

        // "foo" and "bar" must still be configured
        assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))
        // "test" is added because useCloudEndpoint is true and cloudEndpoint is set
        assert.Equal(t, []string{"foo", "bar", "test"}, s.Endpoints())
    })

    t.Run("disable both foundEndpoints and cloudEndpoint", func(t *testing.T) {
        var s Scanner
        s.useFoundEndpoints = true
        s.useCloudEndpoint = true
        s.cloudEndpoint = "test"
        assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))

        // Confirm "test" is in the list
        assert.Equal(t, []string{"foo", "bar", "test"}, s.Endpoints())

        // Now disable both
        s.useFoundEndpoints = false
        s.useCloudEndpoint = false

        // "test" won't be added
        assert.Equal(t, []string{"foo", "bar"}, s.Endpoints("test"))
    })

    t.Run("cloudEndpoint not added when useCloudEndpoint is false", func(t *testing.T) {
        var s Scanner
        s.useFoundEndpoints = false
        s.useCloudEndpoint = false
        s.cloudEndpoint = "new"

        // "foo" and "bar" configured
        assert.NoError(t, s.SetConfiguredEndpoints("foo", "bar"))
        // "new" is not added because useCloudEndpoint is false
        assert.Equal(t, []string{"foo", "bar"}, s.Endpoints())
    })
}
```

ahrav (approved):
Small non-blocking question/suggestion, otherwise lgtm!

pkg/detectors/endpoint_customizer_test.go
kashifkhan0771: Your approach is much cleaner, so I adopted it. However, I combined it with a single scanner object to further reduce the number of lines.

PR #3826, opened 2024-12-31
fixed jirav1 detector email pattern

### LINKED ISSUE #3818: JiraToken v1 detector is subtly broken
The [JiraToken v1 detector](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/jiratoken/v1/jiratoken.go) detector has a defect that prevents it from working in many cases.

If the email match contains any whitespace (e.g., `email = foo.bar@baz.com`), the `strings.Split` can silently fail `len != 2` (#2746)
https://github.com/trufflesecurity/trufflehog/blob/def734a783b2d4542f2b353d326130d618f3c421/pkg/detectors/jiratoken/v1/jiratoken.go#L56-L59

## Related

It seems to have been broken over a year ago by #1288 without anyone noticing, even after unit tests were supposedly added in #3614.

See also: https://github.com/trufflesecurity/trufflehog/pull/2746#issuecomment-2079406298, #3773, #3817

kashifkhan0771:
### Description:
This PR fixes the github issue #3818

### Checklist:


pkg/detectors/jiratoken/v1/jiratoken_test.go
```diff
@@ -16,7 +16,7 @@ var (
 	invalidTokenPattern  = "Z7VoI?J0K4rF#LBfkhO&LAWX"
 	validDomainPattern   = "hereisavalidsubdomain.heresalongdomain.com"
 	invalidDomainPattern = "?y4r3fs1ewqec12v1e3tl.5Hcsrcehic89saXd.ro@"
-	validEmailPattern    = "xfKF_BZq7@grum.com"
```
abmussani, line 19: @kashifkhan0771 Not sure but the old input is also valid email and should pass the user case.

kashifkhan0771: Our common email pattern is not configured to capture capital letters. I guess we need a small enhancement in the common email pattern. Including `(?i)` in start will work.
Thoughts @abmussani @zricethezav @ahrav ?

ahrav: Oh wow, I didn't realize the common email pattern regex was case sensitive. I think making it insensitive is the right approach. 👍

zricethezav (approved):

PR #3835, opened 2025-01-09
Update log for large s3 file

0x1:
### Description:
Updating s3 large file log. Adding the file size, max file size, and file name (for further investigations if needed)

### Checklist:

### DIFF
--- a/pkg/sources/s3/s3.go
+++ b/pkg/sources/s3/s3.go
@@ -457,7 +457,7 @@ func (s *Source) pageChunker(
 
 		// Ignore large files.
 		if *obj.Size > s.maxObjectSize {
-			ctx.Logger().V(5).Info("Skipping large file", "key", *obj.Key, "size", *obj.Size, "max_object_size", s.maxObjectSize)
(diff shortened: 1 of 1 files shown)


pkg/sources/s3/s3.go
```diff
@@ -457,7 +457,7 @@ func (s *Source) pageChunker(
 
 		// Ignore large files.
 		if *obj.Size > s.maxObjectSize {
-			ctx.Logger().V(5).Info("Skipping %d byte file (over maxObjectSize limit)")
+			ctx.Logger().V(5).Info("Skipping large file", "key", *obj.Key, "size", *obj.Size, "max_object_size", s.maxObjectSize)
```
ahrav, line 460: note: I believe the key, and size should already be in the [context](https://github.com/trufflesecurity/trufflehog/pull/3835/files#diff-dd074dd61c256804b0485d58de413da6ba4589eb889e5da850cadf3040f0d8fdR442)

0x1: aah. good catch Ahrav!

rosecodym (approved):
thanks!

Issue #1906, opened 2023-10-18
Fetch GitHub comment history using the GraphQL API


rgmz:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description

The current issue and pr scanning uses GitHub's REST API (e.g., [`processIssueComments`](https://github.com/trufflesecurity/trufflehog/blob/b9f49933b8831e59b98537e96bdddb1f8544ed15/pkg/sources/github/github.go#L1207) uses the SDK to call [List Issue Comments](https://docs.github.com/en/free-pro-team@latest/rest/issues/comments?apiVersion=2022-11-28#list-issue-comments)). Based on my research, switching to the GraphQL API should allow us to not only retrieve *current* content, but historical revisions as well.

### Preferred Solution

This discussion provides insight into how we could fetch the desired information using GitHub's GraphQL API. More work is required to understand how difficult it would be to implement these changes.

https://github.com/orgs/community/discussions/33551

### Additional Context

N/A

#### References

N/A

kashifkhan0771:
This is something we are planning to implement once the #3837 PR is merged

PR #3837, opened 2025-01-11
[GitHub] Add a GraphQL client to the connector 

rgmz:
### Description:
This introduces a latent [GraphQL](https://docs.github.com/en/graphql/guides/introduction-to-graphql) client without making any functional changes. The GraphQL API is required for certain features (e.g., #1906), and it is already being manually called in at least once place:

https://github.com/trufflesecurity/trufflehog/blob/6d1c59f018be0ea3b37aec7fc4b12815c91ab69a/pkg/sources/github_experimental/object_discovery.go#L429-L432

[`shurcooL/githubv4`](https://github.com/shurcooL/githubv4) is the [library recommended by the authors of `google/go-github`](https://github.com/google/go-github/?tab=readme-ov-file#development).

### Checklist:


pkg/sources/github/connector.go
```diff
@@ -2,43 +2,90 @@ package github
...
 	case *sourcespb.GitHub_Unauthenticated:
-		return newUnauthenticatedConnector(apiEndpoint)
+		return newUnauthenticatedConnector(ctx, apiEndpoint)
 	default:
 		return nil, fmt.Errorf("unknown connection type")
 	}
 }
+
+func createAPIClient(ctx context.Context, httpClient *http.Client, apiEndpoint string) (*github.Client, error) {
+	ctx.Logger().WithName("github").V(2).
```
rosecodym, line 57: Shouldn't this logger name be added way higher than the call stack, at entry to the overall `github` code? I presume you didn't do that because `github.com/trufflesecurity/trufflehog/v3/pkg/context` doesn't support `WithName`, but I don't think kludging around that by adding contextual information in the "wrong" spot like this is a good solution overall because we've seen it contribute to bifurcation of our loggers (where a chunk of code is using two loggers at once, each with different contextual information), which in turn has caused concrete debugging problems.

To be clear, I'm not worried that this specific code change will cause an issue - I'm worried that in nine months somebody else is going to copy the pattern somewhere else without thinking too hard about it and create a maintainability problem. If you want to add this piece of context, I think it's safer in the long term to add it at the entry to the `github` code either using `WithValue` or with a new `WithName` that you add to our logging package.

rosecodym (commented):
Pretty cool! It looks like these clients are all unauthenticated - out of curiosity, do you foresee them ever being authenticated?

rgmz:
> Pretty cool! It looks like these clients are all unauthenticated - out of curiosity, do you foresee them ever being authenticated?

It should [inherit authentication from the HTTP client's transport](https://github.com/shurcooL/githubv4/?tab=readme-ov-file#authentication):

https://github.com/trufflesecurity/trufflehog/blob/c0675048c491f1f4f013f5b5432c0ffed9ac8dc7/pkg/sources/github/connector_token.go#L34-L39

pkg/sources/github/connector.go
rgmz: Yeah, the current implementation doesn't make it easy to have a named logger for a package. You either have to call `.WithName()` repeatedly or pass a `logger` separately -- neither are ideal. 

> I think it's safer in the long term to add ... a new WithName that you add to our logging package.

I can try to tackle that in a different PR.

rosecodym:
🤦 no idea how i missed that

pkg/sources/github/connector.go
rgmz: `WithName` functionality can't be added to the logger `Context`. Calls to `WithName` are appended to the logger, meaning that a context passed between packages could become `trufflehog.github.git.detector.etc`.

The best approach is probably a simple helper function in the package.

rgmz: > I'm not sure if this is what you're going for, but a logger in a context can be modified via:
> 
> ```go
> ctx = context.WithLogger(ctx, ctx.Logger().WithName("github"))
> ```
> 
> Not the most ergonomic, but imo not completely foreign (might be just me since I wrote `trufflehog/pkg/context` and `trufflehog/pkg/log` though).

I think this would still have the issue of 'unbounded' logger names:

> > Calls to `WithName` are appended to the logger, meaning that a context passed between packages could become `trufflehog.github.git.detector.etc`.

There's no way to unset or overwrite the logger name that I'm aware of.

rgmz:
Hey @rosecodym, do you have any other concerns with this PR? If not, can it be merged?

pkg/sources/github/connector_basicauth.go
```diff
@@ -5,37 +5,46 @@ import (
...
 
-	apiClient, err := createGitHubClient(httpClient, apiEndpoint)
+	apiClient, err := createAPIClient(ctx, httpClient, apiEndpoint)
 	if err != nil {
 		return nil, fmt.Errorf("could not create API client: %w", err)
 	}
 
+	graphqlClient, err := createGraphqlClient(ctx, httpClient, apiEndpoint)
+	if err != nil {
+		return nil, err
```
rosecodym, line 40: We wrap the API client creation error; we should either wrap this one as well, or wrap neither.

pkg/sources/github/connector_token.go
```diff
@@ -33,15 +37,21 @@ func NewTokenConnector(apiEndpoint string, token string, handleRateLimit func(co
...
 
-	apiClient, err := createGitHubClient(httpClient, apiEndpoint)
+	apiClient, err := createAPIClient(ctx, httpClient, apiEndpoint)
 	if err != nil {
 		return nil, fmt.Errorf("could not create API client: %w", err)
 	}
 
+	graphqlClient, err := createGraphqlClient(ctx, httpClient, apiEndpoint)
+	if err != nil {
+		return nil, err
```
rosecodym, line 47: same error-wrapping comment as above

pkg/sources/github/connector_unauthenticated.go
```diff
@@ -5,27 +5,36 @@ import (
...
 	httpClient := common.RetryableHTTPClientTimeout(int64(httpTimeoutSeconds))
-	apiClient, err := createGitHubClient(httpClient, apiEndpoint)
+	apiClient, err := createAPIClient(ctx, httpClient, apiEndpoint)
 	if err != nil {
 		return nil, fmt.Errorf("could not create API client: %w", err)
 	}
+
+	graphqlClient, err := createGraphqlClient(ctx, httpClient, apiEndpoint)
+	if err != nil {
+		return nil, err
```
rosecodym, line 32: same error-wrapping comment as above

pkg/sources/github/connector.go
```diff
@@ -1,17 +1,91 @@
...
+			return nil, fmt.Errorf("could not create GraphQL client: %w", err)
+		}
+
+		// GitHub Enterprise uses `/api/v3` for the base. (https://github.com/google/go-github/issues/958)
+		// Swap it, and anything before `/api`, with GraphQL.
+		before, _ := strings.CutSuffix(parsedURL.Path, "/api/v3")
+		parsedURL.Path = before + "/api/graphql"
+		graphqlEndpoint = parsedURL.String()
+	}
+	getLogger(ctx).V(2).Info("Creating GraphQL client", "url", graphqlEndpoint)
```
rosecodym, line 88: I don't remember the original reason you used a named logger, but I don't see anything in the current iteration of the PR that requires it. Am I missing something or can you replace this with normal `ctx.Logger()` stuff?

rosecodym: The only other places we use logger names in the codebase is in a handful of detectors and at the very top of the program where we set the logger name to `trufflehog`. I don't think it's necessarily a bad pattern, but if we're going to introduce it I'd like to do so with a plan - having our sources log things using an inconsistent set of techniques has historically caused problems when people expecting things to work one way work a different way. (The new usage here also suggests a caller should persist the generated logger - even though this particular PR immediately throws it away - which has also historically caused logging information to be dropped.)

The fact that this source's logging is coming from a GitHub source should already be indicated by keys that are attached to the injected context higher up the call stack (in the pipeline).

pkg/sources/github/connector.go
```diff
@@ -1,17 +1,92 @@
...
+func createGraphqlClient(ctx context.Context, client *http.Client, apiEndpoint string) (*githubv4.Client, error) {
+	var graphqlEndpoint string
+	if apiEndpoint == cloudV3Endpoint {
+		graphqlEndpoint = cloudGraphqlEndpoint
+	} else {
+		// Use the root endpoint for the host.
+		// https://docs.github.com/en/enterprise-server@3.11/graphql/guides/introduction-to-graphql
+		parsedURL, err := url.Parse(apiEndpoint)
+		if err != nil {
+			return nil, fmt.Errorf("could not create GraphQL client: %w", err)
```
camgunz: I can't decide if this is very nitpicky or a strong commitment to consistency, buuuuut I think we could improve this little thing by doing:

```go
return nil, fmt.Errorf("error parsing URL: %w", err)
```

...and then anything calling `createGraphqlClient` can then wrap _that_ with "error creating GraphQL client: %w". It *is definitely* double wrapping, but for someone looking into what might have gone wrong, it's more explicit about what happened and that's super helpful.

Option 2 IMO is, it seems like most of what this function does is validate a REST API url and convert it to a GraphQL endpoint URL. That _could_ be its own function, which would be very testable, and then an "error parsing URL" error makes even more sense. Or, it might not need its own function--there's a very natural spot for it up in `newConnector` (line 38-ish), and then you can pass a known-valid `graphqlEndpoint` URL to the different connector constructors, which then don't have to check for errors when creating their GraphQL clients as it (apparently) can't error.

---

I appreciate that's more work though? I think just the little fix from option 1 is 👌🏻

camgunz (approved):
Thanks!

kashifkhan0771 (approved):
LGTM!

kashifkhan0771:
me

PR #3843, opened 2025-01-14
Remove global variable scanning implementation for Postman

casey-tran:
### Description:
Remove global variable scanning from the Postman source since the Postman API is unreliable when it comes to global variables.

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman.go
+++ b/pkg/sources/postman/postman.go
@@ -251,21 +251,6 @@ func (s *Source) scanWorkspace(ctx context.Context, chunksChan chan *sources.Chu
 		Type:          "workspace",
 	}
 
-	// scan global variables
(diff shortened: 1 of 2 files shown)


casey-tran:
@zricethezav Yes I was just double checking some stuff. Going to mark it for review.

ahrav:
Are we concerned with breaking any existing users, or is this feature so unreliable that it’s not a concern?

rosecodym (approved):
lgtm, but I would feel safest with either @zricethezav or @joeleonjr's approval as well.

rosecodym:
> Are we concerned with breaking any existing users, or is this feature so unreliable that it’s not a concern?

My understanding is that we're thinking mostly the latter.

zricethezav (approved):

PR #3847, opened 2025-01-21
[Fix] use unrestricted http client only for non-safe requests

abmussani:
### Description:
PR https://github.com/trufflesecurity/trufflehog/pull/3841 removes all the safety majors on OpsGenie Analyzer. This PR use unrestricted http client only for non-safe operations. Other than that restricted one will be used as usual.

### Checklist:

### DIFF
--- a/pkg/analyzer/analyzers/client.go
+++ b/pkg/analyzer/analyzers/client.go
@@ -114,7 +114,7 @@ type AnalyzerRoundTripper struct {
 
 func (r AnalyzerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp, err := r.parent.RoundTrip(req)
-	if err != nil || MethodIsSafe(req.Method) {
+	if err != nil || IsMethodSafe(req.Method) {
 		return resp, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/analyzer/analyzers/client.go
```diff
@@ -126,7 +126,7 @@ func (r AnalyzerRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 
 // methodIsSafe is a helper method to check whether the HTTP method is safe according to MDN Web Docs.
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods#safe_idempotent_and_cacheable_request_methods
-func methodIsSafe(method string) bool {
+func MethodIsSafe(method string) bool {
```
ahrav, line 129: nit: If we rename this to `IsMethodSafe`, it might read more clearly at the call site as a bool check. Otherwise, it could give the impression that we’re trying to convert the method into a safe one

ahrav (approved):

RENSAAS:
https://github.com/ahrav

ahrav (approved):

PR #3855, opened 2025-01-27
Improve Jenkins error message

rosecodym:
### Description:
This PR updates an error message to be more helpful.

### Checklist:


0x1 (approved):
neat. tho, do logs need not start with capitalized letter?

rosecodym:
> neat. tho, do logs need not start with capitalized letter?

It doesn't look like we do that consistently. I think log messages actually get linted to start with a _lowercase_ letter, but this is an error message rather than a direct logging call so I don't think the linter has an opinion.

PR #3857, opened 2025-01-28
Enable Auth0

zricethezav:
### Description:
Previously, I think, we disabled auth0 because it was too noisy since the verify loop is buried in a 3 deep nested loop. This means a single domain could be matched multiple times in the same chunk.

### Checklist:

### DIFF
--- a/pkg/detectors/auth0oauth/auth0oauth.go
+++ b/pkg/detectors/auth0oauth/auth0oauth.go
@@ -37,23 +37,21 @@ func (s Scanner) Keywords() []string {
 // FromData will find and optionally verify Auth0oauth secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
-	clientIdMatches := clientIdPat.FindAllStringSubmatch(dataStr, -1)
-	clientSecretMatches := clientSecretPat.FindAllStringSubmatch(dataStr, -1)
-	domainMatches := domainPat.FindAllStringSubmatch(dataStr, -1)
 	uniqueDomainMatches := make(map[string]struct{})
-	for _, m := range domainMatches {
-		if len(m) > 1 {
-			uniqueDomainMatches[strings.TrimSpace(m[1])] = struct{}{}
-		}
-
+	uniqueClientIDs := make(map[string]struct{})
+	uniqueSecrets := make(map[string]struct{})
+	for _, m := range domainPat.FindAllStringSubmatch(dataStr, -1) {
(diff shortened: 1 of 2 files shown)


ahrav (approved):

pkg/detectors/auth0oauth/auth0oauth.go
```diff
@@ -37,20 +37,24 @@ func (s Scanner) Keywords() []string {
...
 	dataStr := string(data)
-
 	clientIdMatches := clientIdPat.FindAllStringSubmatch(dataStr, -1)
 	clientSecretMatches := clientSecretPat.FindAllStringSubmatch(dataStr, -1)
 	domainMatches := domainPat.FindAllStringSubmatch(dataStr, -1)
+	uniqueDomainMatches := make(map[string]struct{})
+	for _, m := range domainMatches {
+		if len(m) > 1 {
+			uniqueDomainMatches[strings.TrimSpace(m[1])] = struct{}{}
+		}
```
rgmz, line 47: This is superfluous:
```suggestion
	uniqueDomainMatches[m[1]] = struct{}{}
```

pkg/detectors/auth0oauth/auth0oauth.go
```diff
@@ -37,20 +37,24 @@ func (s Scanner) Keywords() []string {
 // FromData will find and optionally verify Auth0oauth secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
-
 	clientIdMatches := clientIdPat.FindAllStringSubmatch(dataStr, -1)
 	clientSecretMatches := clientSecretPat.FindAllStringSubmatch(dataStr, -1)
```
abmussani, line 41: Why isn’t the uniqueness logic applied to both Id and Secret?

pkg/detectors/auth0oauth/auth0oauth.go
zricethezav: is it always?

rgmz: I'd say so.

zricethezav: https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatch guess you're right

pkg/detectors/auth0oauth/auth0oauth.go
zricethezav: Good point, each pattern should be unique

kashifkhan0771 (approved):

PR #3862, opened 2025-01-31
[SCAN-165] Use Err Reporting

0x1:
### Description:
Currently we were not using the error reporting functionality in the GitHub Source. This PR updates that functionality.

### Checklist:


pkg/sources/github/connector_token.go
```diff
@@ -17,14 +17,14 @@ type tokenConnector struct {
 	apiClient          *github.Client
 	token              string
 	isGitHubEnterprise bool
-	handleRateLimit    func(context.Context, error) bool
+	handleRateLimit    func(context.Context, error, ...errorReporter) bool
```
mcastorina, line 20: Is `errorReporter` a vararg to allow for omission at the callsite? Are there places that we don't want to report the rate limit error?

pkg/sources/github/github.go
```diff
@@ -417,9 +417,12 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	for _, repo := range s.filteredRepoCache.Values() {
 		ctx := context.WithValue(ctx, "repo", repo)
 
-		repo, err := s.ensureRepoInfoCache(ctx, repo)
+		repo, err := s.ensureRepoInfoCache(ctx, repo, &unitErrorReporter{reporter})
 		if err != nil {
 			ctx.Logger().Error(err, "error caching repo info")
+			if err := dedupeReporter.UnitErr(ctx, fmt.Errorf("error caching repo info: %w", err)); err != nil {
+				ctx.Logger().Error(err, "failed to report unit error")
+			}
```
mcastorina, line 425: An error returned by `UnitErr` does not mean it failed to record the error. Instead, it's a signal to stop processing and return. That was the intention, anyway, and in practice I think it really would only happen for a context cancellation, which sources should be context-aware anyway.

I'm thinking of removing the `error` return from those interfaces as it's kind of confusing and not really providing much value.

pkg/sources/github/github.go
```diff
@@ -450,23 +453,23 @@ func (s *Source) ensureRepoInfoCache(ctx context.Context, repo string) (string,
...
-			if s.handleRateLimit(ctx, err) {
+			if s.handleRateLimit(ctx, err, reporter) {
 				continue
 			}
 			if err != nil {
 				return repo, fmt.Errorf("failed to fetch gist")
 			}
+			// Normalize the URL to the Gist's pull URL.
+			// See https://github.com/trufflesecurity/trufflehog/pull/2625#issuecomment-2025507937
+			repo = gist.GetGitPullURL()
```
mcastorina, line 464: Why was this added here? It looks unrelated to error reporting.

pkg/sources/github/github.go
```diff
@@ -785,6 +811,12 @@ func (s *Source) handleRateLimit(ctx context.Context, errIn error) bool {
 			retryAfter = retryAfter + jitter
 			rateLimitResumeTime = now.Add(retryAfter)
 			ctx.Logger().Info(fmt.Sprintf("exceeded %s rate limit", limitType), "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
+			// Only report the error if a reporter was provided
+			if len(reporter) > 0 {
+				if err := reporter[0].Err(ctx, fmt.Errorf("exceeded %s rate limit", limitType)); err != nil {
+					ctx.Logger().Error(err, "failed to report rate limit error")
+				}
+			}
```
mcastorina, line 819: If you want to go with the vararg approach, I'd suggest iterating over all the reporters so at least valid calls with multiple reporters aren't silently ignored. It has the added bonus of not having to check the length of the input too.

```go
for _, reporter := range reporters {
    if err := reporter.Err(ctx, ...)
}
```

mcastorina (commented):
Overall this seems like a good approach, just had a few comments.

It looks like the linter is flagging some compilation errors in the test file too.

pkg/sources/github/connector_token.go
0x1: yes, for example getUser doesn't have a reporter: https://github.com/trufflesecurity/trufflehog/blob/e37fa8768ae05685c7b4e17250d0bc9d47c17607/pkg/sources/github/connector_token.go#L71


i _suppose_ we could support it, but we'd need to modify it's entire call chain to pass-in the reporter.

pkg/sources/github/github.go
0x1: it was just moved down. we don't need that URL until after we handled the rate limit. although on a second look, it should probably be before the err check, because we return the repo

pkg/sources/github/connector_token.go
0x1: ya just looked into it. it's way more trouble than it's worth

pkg/sources/github/github.go
```diff
@@ -812,8 +813,8 @@ func (s *Source) handleRateLimit(ctx context.Context, errIn error, reporter ...e
 			rateLimitResumeTime = now.Add(retryAfter)
 			ctx.Logger().Info(fmt.Sprintf("exceeded %s rate limit", limitType), "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
 			// Only report the error if a reporter was provided
-			if len(reporter) > 0 {
-				if err := reporter[0].Err(ctx, fmt.Errorf("exceeded %s rate limit", limitType)); err != nil {
+			for _, reporter := range reporter {
```
mcastorina, line 816: **Nit:** This technically works, but the slice being named `reporter` (in the singular form) reads weird to me.

pkg/sources/github/github.go
```diff
@@ -785,6 +812,12 @@ func (s *Source) handleRateLimit(ctx context.Context, errIn error) bool {
 			retryAfter = retryAfter + jitter
 			rateLimitResumeTime = now.Add(retryAfter)
 			ctx.Logger().Info(fmt.Sprintf("exceeded %s rate limit", limitType), "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
+			// Only report the error if a reporter was provided
+			for _, reporter := range reporters {
+				if err := reporter.Err(ctx, fmt.Errorf("exceeded %s rate limit", limitType)); err != nil {
+					ctx.Logger().Error(err, "failed to report rate limit error")
+				}
```
mcastorina, line 819: Like I mentioned last week, an error returned here doesn't mean it failed to add the error to the report.

mcastorina (commented):
lgtm! just need to update the tests

pkg/sources/github/connector_token.go
rosecodym: another option is to create a no-op reporter to use at all the call sites you don't want to touch. that avoids having to modify entire call trees - you only have to modify the call sites themselves.

pkg/sources/github/github.go
```diff
@@ -785,6 +812,10 @@ func (s *Source) handleRateLimit(ctx context.Context, errIn error) bool {
 			retryAfter = retryAfter + jitter
 			rateLimitResumeTime = now.Add(retryAfter)
 			ctx.Logger().Info(fmt.Sprintf("exceeded %s rate limit", limitType), "retry_after", retryAfter.String(), "resume_time", rateLimitResumeTime.Format(time.RFC3339))
+			// Only report the error if a reporter was provided
+			for _, reporter := range reporters {
+				_ = reporter.Err(ctx, fmt.Errorf("exceeded %s rate limit", limitType))
```
rosecodym, line 817: @mcastorina are we abandoning the idea that `UnitErr` and `ChunkErr` themselves returning an error should be fatal? i can see that there's not really a way to _do_ that in this function anyway, but i wanted to ask about the overall system.

pkg/sources/github/github.go
```diff
@@ -803,6 +834,16 @@ func (s *Source) handleRateLimit(ctx context.Context, errIn error) bool {
 	return true
 }
 
+// handleRateLimitWithUnitReporter is a wrapper around handleRateLimit that includes unit reporting
+func (s *Source) handleRateLimitWithUnitReporter(ctx context.Context, reporter sources.UnitReporter, errIn error) bool {
```
rosecodym, line 838: i did a double-take because these functions are unlike every other one that this pr changes in that they add `reporter` as an argument in a position other than the last. it's an incredibly minor thing but i think it affects readability!

rosecodym (approved):
i left a couple of little comments you can respond to if you want but no pressure either way. thanks for this!

pkg/sources/github/github.go
0x1: none of the other functions take-in an err param, my understanding was that err should always be the last parameter.

mcastorina (approved):
Thanks for improving this!

PR #3864, opened 2025-02-01
Use smaller array to count commits in the GHA action

sryabkov:
### Description:
It is hard but possible to exceed the max size of the `COMMITS` environment variable because of a very large JSON array produced by `${{ toJson(github.event.commits) }}`, for example if commit messages are very long and the number of commits is high. We actually ran into this situation. The value of the variable is only used to get the _count_ of commits. It is possible to get the same count from a much smaller array of _commit IDs_ instead of from a potentially very large array of _full commit objects_. 

### Notes:
- If there are plans to use `COMMITS` for something else other than getting the count of commits, please disregard this PR and close it with the appropriate comment.
- Despite the nudge, I didn't create an issue because this is a small fix that most people should never notice, and it would take double the time to create an issue and a fix. If issues are required for fixes as well as feature additions, I am happy to create the issue.
- I renamed the env variable `COMMITS` to `COMMIT_IDS` to better reflect what the variable contains, but that change is in a separate commit and can be reverted.
- Testing was done in our private repo. 
- I did sign the contributor agreement.
- Credit to @LiptonB for collaborating on this work.

### Checklist:


action.yml
```diff
@@ -35,7 +35,7 @@ runs:
       BASE: ${{ inputs.base }}
       HEAD: ${{ inputs.head }}
       ARGS: ${{ inputs.extra_args }}
-      COMMITS: ${{ toJson(github.event.commits) }}
```
zricethezav, line 38: @sryabkov removing this might break the action for some users

sryabkov: @ zricethezav Can you elaborate and give an example?

sryabkov:
Here is an example of the action failing

image

action.yml
zricethezav: > Maybe I am missing something, but the way I am looking at things, this variable is only set for the one step and it is not available outside of it, so it is not usable by anything other than the code in that step... If there is something I am missing, I am very curious...

You're not missing anything, I was lacking sleep. With slightly better rested eyes I now see this won't break anything for the reasons you explained. 

Changes LGTM 👍🏻

zricethezav (approved):

action.yml
zricethezav: Appreciate the thorough descriptions and clarifying comments. Top notch PR 🤌

PR #3878, opened 2025-02-06
anthropic api key analyzer

kashifkhan0771:
### Description:
This PR add a new analyzer for anthropic detector

### Test Cases:
Screenshot from 2025-02-06 18-18-04
Screenshot from 2025-02-06 18-19-12

### Checklist:


pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, err
+	}
+
+	if err := listMessageBatches(client, key, secretInfo); err != nil {
+		return nil, err
+	}
+
+	// anthropic key has full access only
+	secretInfo.Permissions = PermissionStrings[FullAccess]
+	secretInfo.Valid = true
```
ahrav, line 99: question: Is there ever a scenario where `secretInfo.Valid = false`?

pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+
+// secretInfoToAnalyzerResult translate secret info to Analyzer Result
+func secretInfoToAnalyzerResult(info *SecretInfo) *analyzers.AnalyzerResult {
+	if info == nil {
+		return nil
+	}
+
+	result := analyzers.AnalyzerResult{
+		AnalyzerType: analyzers.AnalyzerAnthropic,
+		Metadata:     map[string]any{},
```
ahrav, line 112: question: It looks like the Metadata is just `Valid_Key`. Could we just set it here directly?
`Metadata: map[string]any{"Valid_Key": info.Valid}`

pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+		Bindings:     make([]analyzers.Binding, 0),
```
ahrav, line 113: suggestion: Looks like we can pre-allocate this slice.
`make([]analyzers.Binding, 0, len(info.AnthropicResources))`

pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// anthropic key has full access only
+	secretInfo.Permissions = PermissionStrings[FullAccess]
+	secretInfo.Valid = true
+
+	return secretInfo, nil
+}
+
+// secretInfoToAnalyzerResult translate secret info to Analyzer Result
+func secretInfoToAnalyzerResult(info *SecretInfo) *analyzers.AnalyzerResult {
```
ahrav, line 105: question: It looks like we collect `AnthropicResources` from `listModels` and `listMessageBatches` in `*SecretInfo`, then iterate over them again to create `Bindings`. Could we skip the intermediary `*SecretItem` and build `Bindings` directly by passing it to `listModels` and `listMessageBatches`? That seems like it would eliminate the extra iteration. I might be missing something, so feel free to correct me if I'm off base.

ahrav (approved):
I don't have a ton of experience reviewing Analyzers, so if any of my comments feel strange, just ignore them 😅

pkg/analyzer/analyzers/anthropic/anthropic.go
ahrav: EDIT: After reviewing another analyzer, I don't think we should buck the trend of how these currently work. Please ignore the above comment.

pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+var _ analyzers.Analyzer = (*Analyzer)(nil)
+
+type Analyzer struct {
+	Cfg *config.Config
+}
+
+// SecretInfo hold the information about the anthropic key
+type SecretInfo struct {
+	Valid              bool
+	Type               string // key type - TODO: Handle Anthropic Admin Keys
```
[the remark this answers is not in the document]
kashifkhan0771: It's a TODO. Right now our anthropic detector does not detect Admin Keys. Once we update the detector we can update the analyzer as well to work with both type of keys.

pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+	Reference          string
+	AnthropicResources []AnthropicResource
+	Permissions        string // always full_access
+	Misc               map[string]string
+}
+
+// AnthropicResource is any resource that can be accessed with anthropic key
+type AnthropicResource struct {
```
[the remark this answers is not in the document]
kashifkhan0771: Makes sense, the reason I did this is to avoid the confusion between AnalyzerResult Binding Resource and Anthropic Resource.
@abmussani @zricethezav your thoughts? Shall we go with this name or rename it to just Resource?

pkg/analyzer/analyzers/anthropic/anthropic.go
```diff
@@ -0,0 +1,160 @@
...
+
+		color.Yellow("\n[i] Expires: Never")
+	}
+}
+
+func AnalyzePermissions(cfg *config.Config, key string) (*SecretInfo, error) {
+	// create a HTTP client
+	client := analyzers.NewAnalyzeClient(cfg)
+
+	var secretInfo = &SecretInfo{}
```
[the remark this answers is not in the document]
kashifkhan0771: My bad! 😶‍🌫️

pkg/analyzer/analyzers/anthropic/anthropic.go
kashifkhan0771: Yes, it happens when the key passed is not valid. We print this on terminal. Once we are able to get resources that mean the key is valid.

pkg/detectors/anthropic/anthropic_integration_test.go
```diff
@@ -104,23 +104,6 @@ func TestAnthropic_FromChunk(t *testing.T) {
 			wantErr:             false,
 			wantVerificationErr: true,
 		},
-		{
```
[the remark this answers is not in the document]
kashifkhan0771: Honestly, I don't know. This test case was failing with valid secret so I removed it 😆

zricethezav (approved):

PR #3888, opened 2025-02-07
Make log command extensible internally

rosecodym:
### Description:
I'm experimenting with ways to manipulate the git log that's being used to generate diffs (specifically, limiting its depth in come cases). Tweaking `RepoPath` for this locally isn't a huge deal but making the log command extensible seems like it could generally be useful, so I figured I'd do it.

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -247,10 +247,10 @@ func (c *Parser) RepoPath(
 	} else {
 		args = append(args, "--all")
 	}
+	args = append(args, additionalArgs...) // These need to come before --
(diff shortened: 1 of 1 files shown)


zricethezav (approved):

zricethezav:
@rgmz caught a gotcha with user defined `git log` options: https://github.com/gitleaks/gitleaks/issues/1153. We might want to introduce something similar here if we think users are gonna go wild with it

rosecodym:
Oh, I but I _should_ ensure that `additionalArgs` are appended _before_ the `--` option, if it exists.

0x1 (approved):

PR #3900, opened 2025-02-11
Change source reporter interfaces to not return errors

mcastorina:
### Description:

The error was meant to allow the reporter to end scanning, but in practice would only be useful for context cancellations, which sources should be handling anyway. Removing the error simplifies the usage of the interface.

### Checklist:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -195,21 +195,18 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	for _, path := range s.paths {
 		fileInfo, err := os.Lstat(filepath.Clean(path))
 		if err != nil {
-			if err := reporter.UnitErr(ctx, err); err != nil {
-				return err
-			}
+			reporter.UnitErr(ctx, err)
 			continue
 		}
 		if !fileInfo.IsDir() {
 			item := sources.CommonSourceUnit{ID: path}
-			if err := reporter.UnitOk(ctx, item); err != nil {
-				return err
-			}
+			reporter.UnitOk(ctx, item)
 			continue
 		}
 		err = fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
 			if err != nil {
-				return reporter.UnitErr(ctx, err)
+				reporter.UnitErr(ctx, err)
+				return nil
 			}
 			if d.IsDir() {
 				return nil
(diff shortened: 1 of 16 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -195,21 +195,18 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
...
-				return err
-			}
+			reporter.UnitOk(ctx, item)
 			continue
 		}
 		err = fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
 			if err != nil {
-				return reporter.UnitErr(ctx, err)
+				reporter.UnitErr(ctx, err)
+				return nil
```
mcastorina, line 209: For the cases where we were returning the reporter error, I changed it to return `nil`. This would be a behavior change for context cancellations, so maybe we should return `ctx.Err()` instead. Thoughts?

rosecodym: I think that returning `ctx.Err()` instead is a good idea!

pkg/sources/legacy_reporters.go
```diff
@@ -12,13 +12,12 @@ type ChanReporter struct {
 	Ch chan<- *Chunk
 }
 
-func (c ChanReporter) ChunkOk(ctx context.Context, chunk Chunk) error {
-	return common.CancellableWrite(ctx, c.Ch, &chunk)
+func (c ChanReporter) ChunkOk(ctx context.Context, chunk Chunk) {
+	_ = common.CancellableWrite(ctx, c.Ch, &chunk)
```
rosecodym, line 16: Can you leave a comment explaining what kind of errors can get swallowed here and why that's ok to do?

rosecodym (commented):
If I understand this PR correctly, it (effectively) _removes_ context cancellation checks from a lot of places (anywhere that used to check for a non-`nil` error returned from a reporter, and return if one was found). Do we need to re-add them explicitly?

mcastorina:
> If I understand this PR correctly, it (effectively) _removes_ context cancellation checks from a lot of places (anywhere that used to check for a non-`nil` error returned from a reporter, and return if one was found). Do we need to re-add them explicitly?

That's correct. Additionally, it removes confirmation of whether the item was successfully handled. That is the main tradeoff of this change. The argument is that sources shouldn't need to care whether the item was handled or not.

They _do_ need to be context aware, though.

rosecodym:
> > If I understand this PR correctly, it (effectively) _removes_ context cancellation checks from a lot of places (anywhere that used to check for a non-`nil` error returned from a reporter, and return if one was found). Do we need to re-add them explicitly?
> 
> That's correct. Additionally, it removes confirmation of whether the item was successfully handled. That is the main tradeoff of this change. The argument is that sources shouldn't need to care whether the item was handled or not.
> 
> They _do_ need to be context aware, though.

Yep, so then should this PR re-add the (implicit) context cancellation checks that it's removed?

PR #3901, opened 2025-02-11
[fix] False Positive Verification in Auth0oauth Detectors

### LINKED ISSUE #3891: Auth0Oauth detector false positives
Auth0Oauth secrets do not properly handle the server returning a 301 Moved Permanently and thus make it look like a verified secret was found.

As an example, this cURL returns 301 MOVED PERMANENTLY:

```
curl --request POST \
  --url "https://eu.auth0.com/oauth/token" \
  --header "content-type: application/x-www-form-urlencoded" \
  --data "grant_type=authorization_code&client_id=mee&client_secret=Hasnain-Is-Here&code=AUTHORIZATION_CODE&redirect_uri=undefined"
```

Reading the code here https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/auth0oauth/auth0oauth.go#L92-L98 it's clear that that gets marked as verified, though it should not

### TruffleHog Version
trufflehog 3.88.5 OSS on my local machine where I reproduced this, though this is also showing up on the enterprise version

### Trace Output

N/A

### Expected Behavior

This is not flagged as a verified secret

### Actual Behavior

It's flagged as a verified secret 

### Steps to Reproduce

Create a file with "fake" auth0 secrets (unfortunately I do not have an easy reproducer I can share) and then run a filesystem scan on it.

The cURL command above + code inspection is a better reproduction.

## Environment
 * OS: MacOS
 * Version trufflehog 3.88.5

abmussani:
### Description:
This PR addresses an issue where a buggy verification process was incorrectly marking false-positive credentials as verified. The following cases are now handled properly:

-  Malformed `authorization_code` Request:
        If an invalid authorization_code request is sent for verification, the API responds with a 403 Forbidden status and an invalid_grant error code.
        Fix: These credentials will now be marked as verified in this case.

- Unauthorized Client:
        If the credentials do not have permission to make an authorization_code request, the API returns a 403 Forbidden status with the unauthorized_client error code.
        Fix: No change in behavior; this case continues to be handled correctly.

- Invalid Domain:
        If the provided domain is not valid, the API returns a 404 Not Found status.
        Fix: These credentials will now be correctly marked as unverified.

- Invalid ID/Secret:
        If the client ID or secret is invalid, the API responds with a 401 Unauthorized status.
        Fix: These credentials will now be correctly marked as unverified.

This PR ensures a more accurate verification process and reduces false positives.

Here is the results of modified test results:
image

### Checklist:

### DIFF
--- a/pkg/detectors/auth0oauth/auth0oauth.go
+++ b/pkg/detectors/auth0oauth/auth0oauth.go
@@ -2,6 +2,7 @@ package auth0oauth
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


ahrav (approved):
Thank you for fixing this.

PR #3912, opened 2025-02-13
Differentiate postman folder from request when at collection root

casey-tran:
### Description:
When Postman items (mainly requests) did not have a parent folder and were at the root of a collection, the folder id/name would be the same as the request id/name. This change helps to keep track of the parent folder through the recursive calls and gives the folder ID of the last folder that the call went through.

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman.go
+++ b/pkg/sources/postman/postman.go
@@ -338,11 +338,6 @@ func (s *Source) scanItem(ctx context.Context, chunksChan chan *sources.Chunk, c
 	s.attemptToAddKeyword(item.Name)
 
 	// override the base collection metadata with item-specific metadata
-	if parentItemId != "" {
-		metadata.FolderID = parentItemId
-	} else {
-		metadata.FolderID = item.UID
-	}
 	metadata.Type = FOLDER_TYPE
 	if metadata.FolderName != "" {
 		// keep track of the folder hierarchy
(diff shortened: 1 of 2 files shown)


pkg/sources/postman/postman.go
```diff
@@ -350,17 +350,23 @@ func (s *Source) scanItem(ctx context.Context, chunksChan chan *sources.Chunk, c
 	if item.UID != "" {
 		metadata.FullID = item.UID
 		metadata.Link = LINK_BASE_URL + FOLDER_TYPE + "/" + metadata.FullID
+		metadata.FolderID = item.UID
```
casey-tran, line 353: This line is to handle the cases where the item that is going to be scanned is a folder itself that contains an auth or script. In that case, we want the FolderID to be of the target instead of the parent.

pkg/sources/postman/postman.go
```diff
@@ -353,13 +352,18 @@ func (s *Source) scanItem(ctx context.Context, chunksChan chan *sources.Chunk, c
 	}
 	// recurse through the folders
 	for _, subItem := range item.Items {
-		s.scanItem(ctx, chunksChan, collection, metadata, subItem)
+		s.scanItem(ctx, chunksChan, collection, metadata, subItem, item.UID)
 	}
 
+	metadata.FolderID = item.UID
```
rosecodym, line 358: per convo, i think a comment here explaining all this parent folder ID business would be helpful

rosecodym (approved):
this looks great, thanks!

pkg/sources/postman/postman.go
casey-tran: Done.

PR #3916, opened 2025-02-15
Replace `golang.org/x/exp` with stdlib

Juneezee:
### Description:

These experimental packages are now available in the Go standard library.

1. `golang.org/x/exp/slices` -> `slices` (https://go.dev/doc/go1.21#slices)
2. `golang.org/x/exp/rand` -> `math/rand/v2` (https://go.dev/doc/go1.22#math_rand_v2)

### Checklist:


ahrav (approved):
Thanks so much for getting this cleaned up and updated! We really appreciate it.

PR #3920, opened 2025-02-18
remove magnetic from default

zricethezav:
### Description:
removing magnetic from default detectors

### Checklist:


dustin-decker (approved):

bugbaba:
@zricethezav  would you mind sharing the reason for this? I can see that the service is still up and haven't notice this detector generating too many false positives either.

PR #3921, opened 2025-02-18
Export git.handleBinary and getSafeRemoteURL

rosecodym:
### Description:
This PR exports some functionality from the git source so that it's usable in a (new, under development) git-flavored source that does not need to wrap an entire git source in order to operate.

One is `getSafeRemoteURL`. This is a straightforward change.

The other is `handleBinary`. Right now, all git binary file handling is invoked from `Git.ScanRepo`, which is itself invoked by our various git-flavored sources.

Since the actual binary file handling function `handleBinary` only used two pieces of information from the "git" source, I just removed its receiver. One of the pieces of information was a flag that caused the function to be skipped entirely; I moved this one to the (two) call sites. The other I just forwarded as an argument.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -642,6 +642,15 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 
 		// Handle binary files by reading the entire file rather than using the diff.
 		if diff.IsBinary {
+			commitHash := plumbing.NewHash(fullHash)
+
+			if s.skipBinaries || feature.ForceSkipBinaries.Load() {
+				logger.V(5).Info("skipping binary file",
+					"commit", commitHash.String()[:7],
+					"path", path)
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


pkg/sources/git/git.go
```diff
@@ -642,6 +642,15 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
...
 		// Handle binary files by reading the entire file rather than using the diff.
 		if diff.IsBinary {
+			commitHash := plumbing.NewHash(fullHash)
+
+			if s.skipBinaries || feature.ForceSkipBinaries.Load() {
+				logger.V(5).Info("skipping binary file",
+					"commit", commitHash.String()[:7],
+					"path", path)
+				continue
+			}
```
casey-tran, line 652: If the use of this chunk of code grew to more than two places, would you write a separate function for it?

casey-tran (approved):

zricethezav (approved):

PR #3929, opened 2025-02-21
Enable GitHub Realtime authentication

rosecodym:
### Description:
GitHub Realtime needs to be able to authenticate. This PR adds relevant fields to the relevant protobuf message, and exports/tweaks some stuff from the `github` source package to permit its (eventual) use by the GitHub Realtime implementation:

* The `Connector` interface and its constructors are now exported
* `Connector.Clone` now accepts additional clone arguments
* The github-specific connection construction has been moved into `github.go`
  * The injected rate limit handler no longer accepts error reporters that it never uses

I also added an API endpoint field to the GitHub Realtime proto message, because we need that too and I was here.

### Checklist:


casey-tran (approved):

zricethezav (approved):
No concerns here

PR #3930, opened 2025-02-22
Remove duplicate `github.com/golang-jwt/jwt` dependency

Juneezee:
### Description:

Our codebase currently uses versions v3, v4, and v5 of the `github.com/golang-jwt/jwt` dependency. To maintain consistency, we should standardize on using only the latest v5 version.

https://github.com/trufflesecurity/trufflehog/blob/6f1e9187dbddb7046d34030bb592d7e5542adea6/go.mod#L56-L58

From the [MIGRATION GUIDE](https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md):

1. `github.com/golang-jwt/jwt/v4` is backward compatible with v3
2. `github.com/golang-jwt/jwt/v5` introduces some core changes, but fortunately, the interfaces and public APIs we use remain unaffected.

### Checklist:


ahrav (approved):
Thank you for getting this updated. ❤️

PR #3933, opened 2025-02-24
Feature: Airtable Personal Access Token Detector

nabeelalam:
### Description:
This PR introduces a new detector for Airtable Personal Access Tokens.

### Checklist:


pkg/detectors/airtablepersonalaccesstoken/airtablepersonalaccesstoken.go
```diff
@@ -0,0 +1,99 @@
...
+type Scanner struct {
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	tokenPat      = regexp.MustCompile(`\b(pat[[:alnum:]]{14}\.[a-f0-9]{64})\b`)
```
zricethezav, line 25: I'm slightly concerned that this will be too loose. Can we use the prefix helper here to tighten up the regex?

nabeelalam: @zricethezav I can add a prefix "airtable" here. Correct me if I'm wrong, but the keyword provided at [line #29](https://github.com/trufflesecurity/trufflehog/pull/3933/files#diff-6bf39997131fe6a9ab457c3039e8c56e52bec20a5a13d199736dee456c002ed4R28) also ensures that the keyword "airtable" exists in in the input, correct?

zricethezav: it insures that the keyword exists in the chunk which could be 10kb big. Putting "airtable" in the regex providers a stronger filter.

zricethezav (approved):

Issue #3664, opened 2024-11-24
GitHub v1 detector: ignore matches in `avatars.githubusercontent.com` URLs


rgmz:
The GitHub v1 detector matches results from the `avatars.githubusercontent.com` domain, resulting in a large amount of false positives.  

**Example**
```
      {
        github: "zpinto",
        name: "Zachary Pinto",
        shortname: "Zach",
        university: "University of California, Irvine",
        city: "Irvine",
        avatar: "https://avatars3.githubusercontent.com/u/20071182?s=460&u=df311dec2adab597c9b1727638a5385eac7ceadb&v=4",
        url: "/preview-template/campus-experts/campus-experts.github.io/pr/3628zpinto"
      },
      {
        github: "ionicc",
        name: "Sagar Vakkala",
        shortname: "Sagar",
        university: "SRM Institute of Science and Technology",
        city: "Chennai",
        avatar: "https://avatars2.githubusercontent.com/u/19570480?s=400&u=390ab557b5dbe9add852b0a4cb807a1b1d5717a0&v=4",
        url: "/preview-template/campus-experts/campus-experts.github.io/pr/3628ionicc"
      },
```
or 
https://github.com/linkedin/li-langchain/blob/5395c254d584103123a140eaa91b1fa49ddadd78/docs/data/people.yml#L2258

PR #3934, opened 2025-02-25
Disable github v1, it is no longer supported

dustin-decker:
### Description:
GitHub support has informed us that the v1 token is no longer supported. It has always been a noisy format due to looking like a sha1 hash.

### Checklist:


rgmz:
Unless they've formally blocked all tokens, the v1 format is still supported by both github.com and GHES.

The amount of false-positives can be cut down by doing stuff like #3658, #3664.

dustin-decker:
Yeah, I just confirmed the same from our integration test which is still passing.

PR #3938, opened 2025-02-26
[Feat] Detector implementation for Azure API Management Direct Management Key

abmussani:
### Description:
This PR implements a new Detector for Azure Direct management API key. This detector looks for direct manage url and keys (Primary or Secondary), generates a access token with expiry of 5 seconds (to avoid leakage) and verify the credentials based on generated values.

Test Output:
```
$ go test -tags=detectors ./pkg/detectors/azuredirectmanagementkey/
ok  	github.com/trufflesecurity/trufflehog/v3/pkg/detectors/azuredirectmanagementkey	6.881s
```

### Checklist:


pkg/detectors/azuredirectmanagementkey/azuredirectmanagementkey.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, keyMatch := range primaryKeyPat.FindAllStringSubmatch(dataStr, -1) {
+		primaryKeyMatchesUnique[strings.TrimSpace(keyMatch[1])] = struct{}{}
+	}
+
+	for baseUrl, serviceName := range urlMatchesUnique {
+		for primaryKey := range primaryKeyMatchesUnique {
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_AzureDirectManagementKey,
+				Raw:          []byte(baseUrl),
+				RawV2:        []byte(baseUrl + primaryKey),
```
rgmz, line 60: Can we start making `RawV2` structured in new detectors? https://github.com/trufflesecurity/trufflehog/pull/3634#discussion_r1851024698

pkg/detectors/azuredirectmanagementkey/azuredirectmanagementkey.go
```diff
@@ -0,0 +1,142 @@
...
+	detectors.DefaultMultiPartCredentialProvider
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	urlPat        = regexp.MustCompile(`https://([a-z0-9][a-z0-9-]{0,48}[a-z0-9])\.management\.azure-api\.net`)         // https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.APIM.Name/
+	primaryKeyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"azure"}) + `\b([a-zA-Z0-9+\/-]{86,88}\b={0,2})`) // Base64-encoded primary key
```
rgmz, line 33: Why `-` at the end?

abmussani: @rgmz Initially I copied that regex from another detector, [Azure Storage](https://github.com/trufflesecurity/trufflehog/blob/8765cc6bcb9e5a81105b3d5a131b8e6fbd6fd34b/pkg/detectors/azure_storage/storage.go#L32). After your comment, I generated 20 times a key and confidently say that all of them ended with `[a-zA-Z0-9]==`. Updated the regex.

pkg/detectors/azuredirectmanagementkey/azuredirectmanagementkey.go
```diff
@@ -0,0 +1,142 @@
...
+	detectors.DefaultMultiPartCredentialProvider
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	urlPat        = regexp.MustCompile(`https://([a-z0-9][a-z0-9-]{0,48}[a-z0-9])\.management\.azure-api\.net`)                                          // https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.APIM.Name/
+	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"azure", ".management.azure-api.net"}) + `\b([a-zA-Z0-9+\/]{83,85}[a-zA-Z0-9]==)`) // pattern for both Primary and secondary key
```
rgmz, line 33: `\b` isn't valid when beside `/` or `+`.

ahrav: This has bitten me so many times... 😭

abmussani: Updated the regex but [AzureContainerRegistry](https://github.com/trufflesecurity/trufflehog/blob/8765cc6bcb9e5a81105b3d5a131b8e6fbd6fd34b/pkg/detectors/azurecontainerregistry/azurecontainerregistry.go#L33) Detector has the similar issue. right ?

pkg/detectors/azuredirectmanagementkey/azuredirectmanagementkey.go
abmussani: @rgmz I am also not in favor of using `RawV2` as structured secret. As @ahrav mentioned that it is tightly coupled with Enterprise and (in future) if we need to restructure it, due to any reason, We wont be able to do it. 

My suggestion is to have another `Map` for this purpose.

rgmz: > ... it is tightly coupled with Enterprise and (in future) if we need to restructure it, due to any reason, We wont be able to do it.

That is the case regardless. Making it structured at least makes it parseable and human-readable, unlike doing a simple concatenation.

pkg/detectors/azuredirectmanagementkey/azuredirectmanagementkey.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (s Scanner) Type() detectorspb.DetectorType {
+	return detectorspb.DetectorType_AzureDirectManagementKey
+}
+
+func (s Scanner) Description() string {
+	return "Azure API Management provides a direct management REST API for performing operations on selected entities, such as users, groups, products, and subscriptions."
+}
+
+func (s Scanner) IsFalsePositive(_ detectors.Result) (bool, string) {
```
kashifkhan0771, line 109: If we want to skip some specific keyword can we implement it as [this](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/slackwebhook/slackwebhook.go#L117)

kashifkhan0771 (approved):
Just one comment related to False Positive Implementation. Rest LGTM!

zricethezav:
@abmussani gotta fix conflicts here too

zricethezav (approved):

PR #3941, opened 2025-02-26
Feature: Airtable Analyzer for Personal Access Tokens

nabeelalam:
### Description:
Airtable PAT Analyzer

image

### Checklist:


pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+	resp, err := common.CallAirtableAPI(token, endpoint.Method, url)
+	if err != nil {
+		return false, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		scopeStatusMap[scopeString] = true
+		return true, nil
+	} else if endpoint.ExpectedErrorResponse != nil {
```
ahrav, line 122: suggestion: Since the `if` block returns, we don’t need an `else if`. We can de-indent this and use a regular `if`, which improves readability by following the **line of sight coding** principle. ([[Reference](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88)](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88))

ex:
```
if resp.StatusCode == http.StatusOK {
    scopeStatusMap[scopeString] = true
    return true, nil
}
if endpoint.ExpectedErrorResponse != nil {
..
}
```

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+		var result map[string]interface{}
```
ahrav, line 123: nit: `interface{}` -> `any`

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+		if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+			return false, err
+		}
+
+		if errorInfo, ok := result["error"].(map[string]interface{}); ok {
```
ahrav, line 128: ditto

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+
+	color.Green("[!] Valid Airtable Personal Access Token\n\n")
+
+	common.PrintUserAndPermissions(userInfo, scopeStatusMap)
+	if scopeStatusMap[common.PermissionStrings[basesReadPermission]] {
+		common.PrintBases(basesInfo)
+	}
+}
+
+func determineScope(token string, scope common.Permission, ids map[string]string) (bool, error) {
```
ahrav, line 101: suggestion: Could we add a short comment for this function, specifically explaining the `ids` parameter? It took a bit of jumping around to understand its purpose.

ahrav, line 101: nit: For consistency and better readability, can we rename `scope` to `permission` or `perm`? Since it's later used as an argument for `getEndpointByPermission`, this would make the code clearer.

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+		}
+	}
+
+	scopeStatusMap[scopeString] = true
+	return true, nil
+}
+
+func determineScopes(token string, basesInfo *common.AirtableBases) error {
+	if basesInfo != nil && len(basesInfo.Bases) > 0 {
```
ahrav, line 141: suggestion: Similar to my earlier **"line of sight"** comment, we can invert the logic of this `if` statement to return early, allowing us to de-indent the following `for` loop for better readability.

Ex:
```
if basesInfo == nil || len(basesInfo.Bases) == 0 {
    return nil
}

for _ base := range baseInfo.Bases {
    ...
}
```

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+							return err
+						}
+						_, err = determineScope(token, common.DataRecordsWrite, ids)
+						if err != nil {
+							return err
+						}
+						tableScopesDetermined = true
+					}
+
+					if granted, err := determineScope(token, common.DataRecordsRead, ids); err != nil {
```
ahrav, line 174: suggestion: We can also remove the `if else` and just use an `if` at the same indentation level, since the `if` block already returns.

Ex:
```
granted, err := determineScope(token, common.DataRecordsRead, ids)
if err != nil {
	return err
}
if !granted {
    continue
}
...
```

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+						return err
+					} else if granted {
+						// Verifying scopes that require an existing record and record read permission
+						records, err := fetchAirtableRecords(token, base.ID, table.ID)
```
ahrav, line 178: question/suggestion: It looks like we might be again swallowing the error? If this is also done intentionally let's add a comment explaining why. We can make this a little easier to reason about if we check the err and handle it first. We then don't even need to check the `len(records)` we can simply range over them.

Ex:
```
records, err := fetchAirtableRecords(token, base.ID, table.ID)
if err != nil {
    return err // or continue
}
for _, record := range records {
    ids["recordID"] = record.ID
    _, err = determineScope(token, common.DataRecordcommentsRead, ids)
    if err != nil {
	    return err
    }
    break
}
```

pkg/analyzer/analyzers/airtable/common/endpoints.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+func GetRequiredPermission(permission Permission) *string {
+	if val, exists := PermissionStrings[permission]; exists {
+		return &val
+	}
+	return nil
+}
+
+var Endpoints = map[EndpointName]Endpoint{
```
ahrav, line 36: This is a nice pattern! One small, optional suggestion: consider constructing this in an `init()` function and making it private with a public getter.  

It’s not a big deal, but this approach prevents unintended modifications by external clients since it wouldn’t be directly exported. I don’t think it's a major concern here, but just something to consider! Totally optional. 😊

Ex:
```
var endpoints map[EndpointName]Endpoint

func init() {
	endpoints = map[EndpointName]Endpoint{
		GetUserInfoEndpoint: {
			URL:    "https://api.airtable.com/v0/meta/whoami",
			Method: "GET",
		},
		// ... existing code ...
	}
}

// GetEndpoint returns the endpoint for the given name and whether it exists
func GetEndpoint(name EndpointName) (Endpoint, bool) {
	endpoint, exists := endpoints[name]
	return endpoint, exists
}
```

ahrav (approved):
Great job @nabeelalam! Left a few small suggestion and questions, but otherwise looks great.

nabeelalam:
> Great job @nabeelalam! Left a few small suggestion and questions, but otherwise looks great.

Thanks for the detailed review, @ahrav. Much appreciated. I'm going through your comments and suggestions.

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+	if !ok {
+		return nil, errors.New("token not found in credInfo")
+	}
+
+	userInfo, err := common.FetchAirtableUserInfo(token)
+	if err != nil {
+		return nil, err
+	}
+
+	scopeStatusMap[common.PermissionStrings[common.UserEmailRead]] = userInfo.Email != nil
```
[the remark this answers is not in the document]
nabeelalam: Yes. Here's the API reference for that: https://airtable.com/developers/web/api/get-user-id-scopes

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
```diff
@@ -0,0 +1,205 @@
...
+
+	var basesInfo *common.AirtableBases
+	if granted, err := determineScope(token, common.SchemaBasesRead, nil); granted {
```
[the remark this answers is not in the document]
nabeelalam: Correct, that error check was supposed to be an else check if `granted == false`. The paths are:
1. `granted == true && err == nil`, then we determine the rest of scopes.
2. `granted == false && err == nil`, we cannot determine any more scopes as we cannot fetch Bases.
3. `granted == false && err != nil`, an API or json decoding error occurred, therefore the analysis was unfinished.

I will update the error checking here as this as this was unintentional.

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
nabeelalam: Agreed, and there is actually one more condition required here. To avoid unnecessarily iterating over another `table` in the range `base.Schema.Tables`, I'll add another check after ranging through the records to check if `len(records) != 0`, as that indicates we have iterated through at least 1 `record` and determined the relevant scopes, therefore we'll `break` both the inner loop and outer loop.
So:
```
records, err := fetchAirtableRecords(token, base.ID, table.ID)
if err != nil {
    return err // or continue
}
for _, record := range records {
    ids["recordID"] = record.ID
    _, err = determineScope(token, common.DataRecordcommentsRead, ids)
    if err != nil {
	    return err
    }
    break
}
if len(records) != 0 {
    break
}
```

pkg/analyzer/analyzers/airtable/airtablepat/airtable.go
nabeelalam: You're right. The terms got a bit shuffled since each AirTable `scope` maps to a `Permission` enum in the auto generated `permissions.go` file, but each AirTable `scope` also has its own granular "permissions".

I agree that the usage of the terms should be more consistent. I'll update this and any other instances I see with inconsistent use of the terms scope/permission.

ahrav:
> > Great job @nabeelalam! Left a few small suggestion and questions, but otherwise looks great.
> 
> Thanks for the detailed review, @ahrav. Much appreciated. I'm going through your comments and suggestions.

Feel free to merge whenever you are ready. None of my comments should be blocking.

PR #3948, opened 2025-02-27
LaunchDarkly Token Analyzer

kashifkhan0771:
### Description:
This PR adds a new analyzer for launchdarkly token
OSSTicket: OSS-95
Screenshot from 2025-03-04 12-10-31

### Checklist:


pkg/analyzer/analyzers/launchdarkly/launchdarkly.go
```diff
@@ -0,0 +1,203 @@
...
+		"No of Custom Roles Assigned", "Has Inline Policy"})
+
+	tokenTable.AppendRow(table.Row{color.GreenString(user.Token.ID), color.GreenString(user.Token.Name), color.GreenString(user.Token.Role),
+		color.GreenString(fmt.Sprintf("%t", user.Token.IsServiceToken)), color.GreenString(fmt.Sprintf("%d", user.Token.APIVersion)),
+		color.GreenString(fmt.Sprintf("%d", len(user.Token.CustomRoles))), color.GreenString(fmt.Sprintf("%t", user.Token.hasInlineRole()))})
+
+	tokenTable.Render()
+
+	// print custom roles information
+	if user.Token.hasCustomRoles() {
```
nabeelalam, line 172: nit: we can use an early return to reduce indentation
```suggestion
    if !user.Token.hasCustomRoles() {
        return
    }
    // print token information
    color.Green("\n[i] Custom Roles Assigned to Token")
    ...
```

zricethezav (approved):

pkg/analyzer/analyzers/launchdarkly/launchdarkly.go
```diff
@@ -0,0 +1,205 @@
...
+	for key, value := range resource.MetaData {
+		analyzerRes.Metadata[key] = value
+	}
+
+	return &analyzerRes
+}
+
+// getPermissionType return what type of permission is assigned to token
+func getPermissionType(token Token) string {
+	permission := ""
```
ahrav, line 132: nit (potential alternative solution): A `switch` is a little more idiomatic, but totally up to you.

Ex:
```
switch {
case token.Role != "":
    return token.Role
case token.hasInlineRole():
    return "Inline Policy"
case token.hasCustomRoles():
    return "Custom Roles"
default:
    return ""
}
```

pkg/analyzer/analyzers/launchdarkly/models.go
```diff
@@ -0,0 +1,135 @@
...
+	webhooksKey     = "Webhooks"
+	featureFlagsKey = "Feature Flags"
+)
+
+type SecretInfo struct {
+	User        User
+	Permissions []string
+	Resources   []Resource
+	// to concurrently read and write
+	mu sync.RWMutex
```
ahrav, line 28: nit: I've found it helpful to use `mu` almost like a **"hat"** for the resource(s) it protects by placing it directly above them. This makes it visually clear what it's guarding and can even replace the need for a comment.  

That said, I’m not entirely sure which resources `mu` is protecting in this case. 🤔 This is definitely an **opinionated** take, so feel free to ignore it! 🤣  

I forget where I first read about this, but [here](https://dmitri.shuralyov.com/idiomatic-go/entries/2) is a separate article about it. 😅 

Ex:
```
type SecretInfo struct {
    ...
    mu sync.RWMutex <-- see, like a "hat" for "Resources"
    Resources []Resource
}
```

pkg/analyzer/analyzers/launchdarkly/models.go
```diff
@@ -0,0 +1,135 @@
...
+	Polices           []Policy
+	BasePermission    string
+	AssignedToMembers int
+	AssignedToTeams   int
+}
+
+/*
+policy is a set of statements
+
+Jargon:
```
ahrav, line 66: 🤣

pkg/analyzer/analyzers/launchdarkly/models.go
```diff
@@ -0,0 +1,135 @@
...
+func (s *SecretInfo) appendResource(resource Resource) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.Resources = append(s.Resources, resource)
+}
+
+// listResourceByType returns a list of resources matching the given type.
+func (s *SecretInfo) listResourceByType(resourceType string) []Resource {
+	s.mu.RLock()
```
ahrav, line 100: question: Do we actually need the mutex? 🤔  I didn’t notice any concurrent operations, but I might have missed something.

ahrav (approved):
Left a couple questions, but otherwise this looks great! Nice work.

pkg/analyzer/analyzers/launchdarkly/models.go
kashifkhan0771: I guess you missed the `requests.go` file.

kashifkhan0771: I guess you missed the `requests.go` file. Large diffs are not rendered by default.

pkg/analyzer/analyzers/launchdarkly/requests.go
```diff
@@ -0,0 +1,779 @@
...
+	"strings"
+	"sync"
+	"time"
+)
+
+var (
+	baseURL = "https://app.launchdarkly.com/api"
+
+	endpoints = map[string]string{
+		"callerIdentity": "/v2/caller-identity",
```
ahrav, line 19: question: How come these three are raw strings instead of constants like the others? 🤔 If it was done on purpose, would you mind just including a brief comment as to why.

ahrav (approved):
Sorry I missed some of the files in my initial review.

I left a few suggestions, but only one of them is really blocking. The rest of this looks great. Nice job!

pkg/analyzer/analyzers/launchdarkly/models.go
ahrav: 😢 I knew there had to be a reason. Sorry about that.

kashifkhan0771: No worries!

pkg/analyzer/analyzers/launchdarkly/requests.go
kashifkhan0771: Yes, this is intentional. These three APIs are specifically for user information, not resources. Additionally, these keys are used only once, unlike resource keys, which are used in multiple places. I'll add a comment to clarify this.

pkg/analyzer/analyzers/launchdarkly/requests.go
```diff
@@ -0,0 +1,779 @@
...
+
+	responseBodyByte, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return responseBodyByte, resp.StatusCode, nil
+}
+
+func CaptureResources(client *http.Client, token string, secretInfo *SecretInfo) error {
```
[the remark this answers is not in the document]
kashifkhan0771: Done ✅

pkg/analyzer/analyzers/launchdarkly/requests.go
```diff
@@ -0,0 +1,779 @@
...
+		}
+
+		for _, application := range applications.Items {
+			resource := Resource{
+				ID:   fmt.Sprintf("launchdarkly/app/%s", application.Key),
+				Name: application.Name,
+				Type: applicationKey,
+			}
+
+			resource.updateResourceMetadata("Maintainer Email", application.Maintainer.Email)
```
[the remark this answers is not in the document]
kashifkhan0771: I guess I was trying to make the code a bit more fancy. 😆 
refactoring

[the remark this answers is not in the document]
ahrav: I do like the idea of encapsulating all the logic for updating `Resource` in one place, but since we’re not using a constructor and just populating it directly, doing it inline made the review a bit easier.  

Also, that GIF is **way too relatable**—I feel that on a personal level. 🤣

ahrav:
I feel like this will be a pretty commonly used Analyzer, awesome work!

PR #3952, opened 2025-03-03
Enumeration jobs should not consume concurrency slots

dustin-decker:
### Description:
Enumeration jobs should not consume concurrency slots since they don't consume significant resources and we want to ensure that scans are occuring.

### Checklist:


0x1 (approved):

ahrav (approved):
Is it possible that enumeration jobs are consuming all the concurrency slots, preventing actual scanning from running?

dustin-decker:
> Is it possible that enumeration jobs are consuming all the concurrency slots, preventing actual scanning from running?

Yes, this can happen right now. I believe Ahmed observed it on main on Friday.

PR #3967, opened 2025-03-07
TUI - handle relative paths

hxnyk:
### Description:
Filesystem scans where user enters `~/something` doesn't work otherwise

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -269,15 +269,29 @@ func init() {
 	// Support -h for help
 	cli.HelpFlag.Short('h')
 
+	// Check if the TUI environment variable is set.
+	if ok, err := strconv.ParseBool(os.Getenv("TUI_PARENT")); err == nil {
+		usingTUI = ok
+	}
+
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -269,15 +269,29 @@ func init() {
...
+		usingTUI = ok
+	}
+
 	if isatty.IsTerminal(os.Stdout.Fd()) && (len(os.Args) <= 1 || os.Args[1] == analyzeCmd.FullCommand()) {
 		args := tui.Run(os.Args[1:])
 		if len(args) == 0 {
 			os.Exit(0)
 		}
 
+		binary, err := exec.LookPath("sh")
```
ahrav, line 283: question: You are only really concerned with this working on Unix-like systems yea?

ahrav (approved):

main.go
hxnyk: 👀 never tested the tui on windows tbh -- will take a look _soon_

PR #3971, opened 2025-03-11
optimized ipquality detector

kashifkhan0771:
### Description:
IP Quality detector is reporting false-positives and reporting almost every 32 character alpha numeric detected string as live.

Related to CSM-530 Ticket

### Checklist:


pkg/detectors/ipquality/ipquality.go
```diff
@@ -22,9 +24,18 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"ipquality"}) + `\b([0-9a-z]{32})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"ipquality"}) + `\b([0-9a-zA-Z]{32})\b`)
+
+	// response messages
+	invalidKeyMessage         = "Invalid or unauthorized key"
+	insufficientCreditMessage = "insufficient credits"
```
abmussani, line 31: nit: `const` instead of `var`

pkg/detectors/ipquality/ipquality.go
```diff
@@ -79,3 +76,51 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+	if err != nil {
+		return false, err
+	}
+
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	switch resp.StatusCode {
```
abmussani, line 98: not sure about the documentation but we should handle other statuses like `Forbidden` or `Unauthorized` etc

pkg/detectors/ipquality/ipquality.go
```diff
@@ -79,3 +76,51 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+
+		if err = json.Unmarshal(bodyBytes, &response); err != nil {
+			return false, err
+		}
+
+		switch response.Success {
+		case true:
+			return true, nil
+		case false:
+			if strings.Contains(response.Message, insufficientCreditMessage) {
```
kashifkhan0771, line 115: Updating the logic to report keys as not live if the response message indicates "insufficient credit."  

**Reasoning:**  
- Deleted or incorrect keys also return this message.  
- Even if the API key exists, it remains unusable due to insufficient credit.  

To minimize false positives, it's more accurate to classify keys as not live in this scenario.

kashifkhan0771: Well maybe we can also add something in metadata to inform user about this 🤔

pkg/detectors/ipquality/ipquality.go
```diff
@@ -79,3 +76,51 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+				return false, nil
+			} else if strings.Contains(response.Message, invalidKeyMessage) {
+				return false, fmt.Errorf("invalid key")
+			}
```
abmussani, line 119: IMO, `insufficient credit` message reflects credentials are valid. Only `invalid key` message should be handle to return `false, nil`.

thoughts ?

kashifkhan0771: Reson: https://github.com/trufflesecurity/trufflehog/pull/3971#discussion_r1988546033

kashifkhan0771: Reason: https://github.com/trufflesecurity/trufflehog/pull/3971#discussion_r1988546033

pkg/detectors/ipquality/ipquality.go
kashifkhan0771: Their API response is 200 only with different `success` as either true or false and a message.

pkg/detectors/ipquality/ipquality.go
kashifkhan0771: Added 🚀

abmussani:
> Updating the logic to report keys as not live if the response message indicates "insufficient credit."
> 
> **Reasoning:**
> 
>     * Deleted or incorrect keys also return this message.
> 
>     * Even if the API key exists, it remains unusable due to insufficient credit.
> 
> 
> To minimize false positives, it's more accurate to classify keys as not live in this scenario.

I am reading it correctly that If API key is exists, it will be marked unverified due to insufficient credit ?

kashifkhan0771:
> I am reading it correctly that If API key is exists, it will be marked unverified due to insufficient credit ?

I mean if key is active and we received `insufficient credit` that key is is not use able because of credit. And if key is disabled/deleted still the API return `insufficient credit` for some reason. Hence to be on safe side and to avoid false positives it's better to mark it as InActive with some extraData.

abmussani:
> > I am reading it correctly that If API key is exists, it will be marked unverified due to insufficient credit ?
> 
> I mean if key is active and we received `insufficient credit` that key is is not use able because of credit. And if key is disabled/deleted still the API return `insufficient credit` for some reason. Hence to be on safe side and to avoid false positives it's better to mark it as InActive with some extraData.

It’s better to see if another endpoint can be used for verification. The implemented solution may raise false negatives and still mark valid credentials as unverified. IMO, that’s worse than having false positives. @zricethezav, what are your thoughts on this?

kashifkhan0771:
> It’s better to see if another endpoint can be used for verification. The implemented solution may raise false negatives and still mark valid credentials as unverified. IMO, that’s worse than having false positives. @zricethezav, what are your thoughts on this?

I tested a few other APIs, but the response is always the same, status 200 with different messages. In my opinion, the scenario where a key is valid and active but has insufficient credit is less common than cases where a key is invalid or disabled. If we mark such keys as active, it increases the number of false positives, as any invalid capture string would trigger the same error, leading to most detections being incorrectly marked as live. I suggest adding extraData to indicate that the key appears valid but has insufficient credit while marking it as not live since, even if correct, it cannot be used due to low credit.

abmussani (approved):
@kashifkhan0771  changes looks good to me . Just a small request to put the comment about why `insufficient credit` error cannot be considered as just `unverified` credentials.

kashifkhan0771:
Test cases passing locally
Screenshot from 2025-03-21 15-53-36

PR #3974, opened 2025-03-12
[Feat] Added Figma (PAT) Analyzer

nabeelalam:
### Description:
Added a new analyzer to analyze detected Figma PATs

### Output:
image

### Checklist:


kashifkhan0771 (approved):
Overall, it looks good to me! I haven't conducted a detailed review.

pkg/analyzer/analyzers/figma/models.go
```diff
@@ -0,0 +1,34 @@
+package figma
+
+type ScopeValidationResult struct {
```
abmussani, line 3: I may have overlooked but does it need to be public ?

pkg/analyzer/analyzers/figma/figma.go
```diff
@@ -0,0 +1,247 @@
...
+		color.Red("[x] Error : %s", err.Error())
+		return
+	}
+
+	color.Green("[!] Valid Figma Personal Access Token\n\n")
+	PrintUserAndPermissions(info)
+}
+
+func AnalyzePermissions(cfg *config.Config, token string) (*SecretInfo, error) {
+	client := analyzers.NewAnalyzeClient(cfg)
```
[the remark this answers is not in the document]
nabeelalam: We're not expecting a `200 OK` response on the non-safe requests we're making, since we're not sending any input data in the `POST` or `PATCH` requests. The condition to check for an `http.StatusOk` response is mainly there for safe requests. Is there a more appropriate way to handle this?

pkg/analyzer/analyzers/figma/figma.go
```diff
@@ -0,0 +1,247 @@
...
+
+	expectedError := endpoint.ExpectedResponseWithoutScope
+	scopeStrings, scopeIsDenied := validateErrorAndGetScopes(errorResponse, expectedError)
+	if scopeIsDenied {
+		scopes := getScopesFromScopeStrings(scopeStrings)
+		return ScopeValidationResult{Status: StatusDenied, Scopes: scopes}, nil
+	}
+
+	// Can not determine scope as the expected error is unknown
+	return ScopeValidationResult{Status: StatusUnverified}, nil
```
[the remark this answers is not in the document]
nabeelalam: I agree it's not necessary, my motivation for adding the `APIErrorResponse` model and decode the response into an object was to make the code "clearer", if that makes sense, I found it a bit more readable that way. Should I remove it?

pkg/analyzer/analyzers/figma/models.go
zricethezav: @nabeelalam would you mind taking a look at this?

abmussani (approved):

pkg/analyzer/analyzers/figma/figma.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		defer resp.Body.Close()
+
+		validationResult, err := validateTokenScopesFromResponse(resp, endpoint)
+		if err != nil {
+			return nil, err
+		}
+		if validationResult.Status == StatusGranted {
+			if scope == ScopeFilesRead {
+				if err := json.NewDecoder(resp.Body).Decode(&info.UserInfo); err != nil {
```
abmussani, line 93: @nabeelalam The resp.body is being decoded in 'validateTokenScopesFromResponse' but also done at line 93. How can be sure that it will be decoded only one ? as of it will be done multiple, 'EOF' error will be returned. Maybe, code can be simplified by doing it once.

abmussani (changes requested):

pkg/analyzer/analyzers/figma/figma.go
nabeelalam: The body will only be decoded on line #93 if the `scope == ScopeFilesRead` in which case the expected response is `200 OK`, so `validateTokenScopesFromResponse` will always return at line `121` otherwise line #93 will never be reached.

abmussani: That’s unclear. Can you simplify it further? maybe having the decoding at one place ?

pkg/analyzer/analyzers/figma/figma.go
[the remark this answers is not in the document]
abmussani: We can keep the valid statuses with in the JSON configuration and just compare when response has been received ? This condition will get eliminated. thoughts ?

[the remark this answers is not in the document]
nabeelalam: I've simplified the JSON and the scope parsing. Please take a look at the new updates I think that simplifies the approach.

abmussani (approved):

PR #3981, opened 2025-03-17
updated storyblok detector

### LINKED ISSUE #3909: Storyblok Detector Fails to Detect Personal Access Token
### Actual Behavior

The Storyblok detector alerts on access tokens (which are tied to a Storyblok space), but not personal access tokens that access the [Storyblok Management API](https://www.storyblok.com/docs/management-api/authentication). The REGEX for the personal access tokens is different than the CDN keys.

### Suggested Improvements

1. Update the existing verification check to exclude ["Public"](https://www.storyblok.com/docs/api/content-delivery/v2/getting-started/authentication) access tokens from being marked as valid, since these are meant to be exposed client-side and are not secrets.
2. There is an argument to be made that ["Preview"](https://www.storyblok.com/docs/api/content-delivery/v2/getting-started/authentication) access tokens should remain private, though the impact is probably not that great, so I believe those should continue to be considered secrets.
3. There are two other new access token types: "Theme" and "Asset".  I'm unclear as to whether they should be considered a secret or not and the Storyblok docs aren't updated yet with details.
4. Add a new detector for the Storyblock Management API key type. Maybe it's just the same detector with a separate REGEX and a field in `extra_data` indicating that it's for a different service within Storyblok. These keys have the largest impact.

kashifkhan0771:
### Description:
This PR fixes #3909 and oss-136

### Tests:
Screenshot from 2025-03-17 13-56-01

### Checklist:


pkg/detectors/storyblok/storyblok.go
```diff
@@ -64,10 +66,29 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
-func (s Scanner) Type() detectorspb.DetectorType {
-	return detectorspb.DetectorType_Storyblok
-}
+// docs: https://www.storyblok.com/docs/api/content-delivery/v2/getting-started/authentication
+func verifyStoryBlokAccessToken(ctx context.Context, client *http.Client, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.storyblok.com/v1/cdn/spaces/me/?token="+token, nil)
```
abmussani, line 71: nit: http.GET

pkg/detectors/storyblokpersonalaccesstoken/storyblok.go
```diff
@@ -0,0 +1,99 @@
...
+	return `Storyblok is a headless CMS that allows developers to build flexible and powerful content management solutions.
+			Storyblok personal access tokens can be used with management APIs.
+			The Storyblok Management API allows you to create, edit, update, and delete content using a common interface`
+}
+
+// FromData will find and optionally verify Storyblok secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
```
abmussani, line 49: can we put uniqueness logic here to avoid duplicates ?

pkg/detectors/storyblokpersonalaccesstoken/storyblok.go
```diff
@@ -0,0 +1,99 @@
...
+
+		results = append(results, s1)
+	}
+
+	return results, nil
+}
+
+// docs: http://storyblok.com/docs/api/management/core-resources/spaces/retrieve-multiple-spaces
+func verifyStoryBlokPersonalAccessToken(ctx context.Context, client *http.Client, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://mapi.storyblok.com/v1/spaces/", nil)
```
abmussani, line 73: nit: http.GET

abmussani (commented):
Thanks @kashifkhan0771. Overall, looks great. Just few minor comments.

abmussani (approved):

PR #3995, opened 2025-03-25
Postman source handle different JSON info for headers

casey-tran:
### Description:
Sometimes the response headers of Postman collections can return a list of strings instead of the expected list of key-value pairs. If a list of strings was encountered, the JSON unmarshaller would error out due to a type mismatch between strings and key-values.

This PR seeks to solve the issue by initially unmarshalling the header info as `json.RawMessage` and then later attempting to unmarshal that into a `[]KeyValue` or a `[]string`.

The PR also corrects a typo in naming the function `buildSubstituteSet `, adjusts the stdout of targets that are just collections with no workspaces to display the `Location_type` properly as a string, and replaces `interface{}` with `any`.

### Checklist:

### DIFF
--- a/pkg/output/plain.go
+++ b/pkg/output/plain.go
@@ -117,9 +117,9 @@ func structToMap(obj any) (m map[string]map[string]any, err error) {
 	}
 	err = json.Unmarshal(data, &m)
 	// Due to PostmanLocationType protobuf field being an enum, we want to be able to assign the string value of the enum to the field without needing to create another Protobuf field.
-	// To have the "UNKNOWN_POSTMAN = 0" value be assigned correctly to the field, we need to check if the Postman workspace ID is filled since every secret in the Postman source
-	// should have a valid workspace ID and the 0 value is considered nil for integers.
-	if m["Postman"]["workspace_uuid"] != nil {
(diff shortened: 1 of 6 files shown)


rosecodym (approved):
This seems like a fine solution to the problem (and thank you for cleaning other things as you go), but I find the problem itself pretty perplexing and it feels like we're missing something simple. Can you share any snippets from responses where you're seeing this?

casey-tran:
> This seems like a fine solution to the problem (and thank you for cleaning other things as you go), but I find the problem itself pretty perplexing and it feels like we're missing something simple. Can you share any snippets from responses where you're seeing this?

Thanks for the review @rosecodym! Currently for actual responses from Postman, I can only get key-value pairs. I'm not sure how people are getting strings in their response headers.

casey-tran:
The response headers can apparently look like this: `"header":[""]`. As this branch is written right now, it can handle that example.

casey-tran:
Something is wrong with response header detection during a test using the actual Postman API. Will need to investigate further before merging.

casey-tran:
So the for each style loop was the wrong one to use for this solution because the value used during the for each loop iteration was just a copy, and the `Collection` struct returned at the end of `GetCollection` had null values for all `HeaderKeyValue` and `HeaderString` fields.

PR #3997, opened 2025-03-27
[Fix] Line number issue for custom detector

kashifkhan0771:
### Description:
This PR resolves the incorrect line number issue for custom detectors with multiple regex patterns. It introduces a new configuration field that allows selecting one pattern as the primary. Matches from this primary regex will be used to determine the line number.

This is also helpful for detectors that append multiple matches to the RAW field. To specify a primary secret for a detector, simply add:
```go
s1.SetPrimarySecretValue(match)
```
The engine will then use this primary secret to determine the line number. If a primary secret is not set, the engine will fall back to using the RAW field to locate the line number, as it currently does.

Related Ticket: CSM-864 | OSS-139

### Checklist:


kashifkhan0771:
> @kashifkhan0771 do we have any examples of what this looks like in the output? Can we write some tests for this?

Currently, I did not print those secrets in the output. Instead, I stored them in the result and use them to identify the exact line number. I can add some test cases for the custom detector where we use this approach for multi-pattern.

zricethezav (approved):

rosecodym (approved):
approving the six lines of this that are owned by the scanning team

ahrav (approved):
Awesome job @kashifkhan0771 ! This looks great.

PR #4002, opened 2025-04-01
Added a rate limiter for getting the overall list of Postman workspaces

casey-tran:
### Description:
Every other call to `getPostmanReq` had a rate limiter except for the one in `EnumerateWorkspaces`, which possibly over time with a big enough list of workspaces would cause the rate limit to become exceeded and out of tolerance for the Postman API. At least that is my hunch for why it is erroring out during enumeration after trying to `GET` a workspace four times.

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman_client.go
+++ b/pkg/sources/postman/postman_client.go
@@ -271,6 +271,9 @@ func (c *Client) EnumerateWorkspaces(ctx context.Context) ([]Workspace, error) {
 		Workspaces []Workspace `json:"workspaces"`
 	}{}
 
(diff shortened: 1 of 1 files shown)


camgunz (approved):

camgunz:
I already approved, but in Sentry API responses they have headers that indicate rate limiting status. Dunno if Postman also has some you could debug log.

casey-tran:
> I already approved, but in Sentry API responses they have headers that indicate rate limiting status. Dunno if Postman also has some you could debug log.

Thank you Charlie. Yes they do have some rate limiting status in the response headers. I will look into using those as well in a separate PR.

PR #4004, opened 2025-04-03
[Feat] Plaid API Analyzer

nabeelalam:
### Description:
Adds a new analyzer for Plaid API
Also updated the Plaid detector to match `access_tokens` as well

### Output:
image

### Checklist:


pkg/detectors/plaidkey/plaidkey.go
```diff
@@ -59,38 +60,52 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				s1 := detectors.Result{
+					DetectorType: detectorspb.DetectorType_PlaidKey,
+					Raw:          []byte(secret),
+					RawV2:        []byte(fmt.Sprintf(`%s:%s:%s`, secret, id, token)),
 				}
-				results = append(results, s1)
-				// if the environment is sandbox, we don't need to check production
-				if s1.Verified {
-					break
+				environments := []string{"sandbox", "production"}
```
kashifkhan0771, line 80: Since the environment name is embedded in the access token, why aren't we verifying it against the specific environment the token corresponds to, rather than checking against both?

kashifkhan0771 (approved):
Just a small non-blocking comment. Everything else looks good to me!

pkg/analyzer/analyzers/plaid/plaid.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	url := "https://" + environment + ".plaid.com/accounts/get"
+	jsonBody, _ := json.Marshal(body)
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(jsonBody))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
```
abmussani, line 112: should `client` be used here ?

pkg/analyzer/analyzers/plaid/plaid.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var accounts accountsResponse
+	if err := json.NewDecoder(resp.Body).Decode(&accounts); err != nil {
+		return nil, err
+	}
+
```
abmussani, line 122: Checking the response status code before decoding helps avoid unexpected panics.

pkg/analyzer/analyzers/plaid/plaid.go
```diff
@@ -0,0 +1,212 @@
...
+	result := analyzers.AnalyzerResult{
+		AnalyzerType:       analyzers.AnalyzerTypePlaid,
+		Metadata:           nil,
+		Bindings:           make([]analyzers.Binding, len(info.Products)),
+		UnboundedResources: make([]analyzers.Resource, len(info.Accounts)),
+	}
+	resource := analyzers.Resource{
+		Name:               "Plaid API Access Token",
+		FullyQualifiedName: "Plaid API Access Token",
```
abmussani, line 139: Considering the limitation of FullyQualifiedName (It must be unique), Its not a good idea to make it hard coded. Also, I believe that products are not permissions. Permissions are, what level of access this credentials have on the resources (e.g. read/write/full access). Products can be consider as resource with unique identity.

pkg/analyzer/analyzers/plaid/plaid.go
```diff
@@ -0,0 +1,212 @@
...
+			Permission: analyzers.Permission{
+				Value: product,
+			},
+		}
+	}
+
+	for idx, account := range info.Accounts {
+		result.UnboundedResources[idx] = analyzers.Resource{
+			Name:               account.Name,
+			FullyQualifiedName: account.OfficialName,
```
abmussani, line 155: accountId will be best for FullyQualifiedName.

pkg/detectors/plaidkey/plaidkey_integration_test.go
```diff
@@ -99,9 +111,12 @@ func TestPlaidKey_FromChunk(t *testing.T) {
 				if len(got[i].Raw) == 0 {
 					t.Fatalf("no raw secret present: \n %+v", got[i])
 				}
-				got[i].Raw = nil
+				if len(got[i].RawV2) == 0 {
+					t.Fatalf("no raw v2 secret present: \n %+v", got[i])
+				}
 			}
-			if diff := pretty.Compare(got, tt.want); diff != "" {
+			ignoreOpts := cmpopts.IgnoreFields(detectors.Result{}, "Raw", "verificationError")
```
abmussani, line 118: `extradata` and `analysisinfo` can also be added in ignoreFields.

pkg/detectors/plaidkey/plaidkey.go
```diff
@@ -59,38 +60,52 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				if verify {
+					for _, env := range environments {
```
abmussani, line 82: I might be wrong but access token has `env` related information. do you think we can reduce the calls by looking at that information ?

pkg/analyzer/analyzers/plaid/plaid.go
```diff
@@ -0,0 +1,212 @@
...
+	environment := ""
+	if strings.Contains(accessToken, "production") {
+		environment = "production"
+	}
+	if strings.Contains(accessToken, "sandbox") {
+		environment = "sandbox"
+	}
+	if environment == "" {
+		return nil, errors.New("Environment could not be parsed from access token")
+	}
```
abmussani, line 84: Right now, Plaid only supports 2 env, `production` and `sandbox`. IMO, by default, sandbox be selected and only check for production.

abmussani (changes requested):

pkg/analyzer/analyzers/plaid/plaid.go
nabeelalam: I've updated this

pkg/analyzer/analyzers/plaid/plaid.go
nabeelalam: That makes sense, I've made this update to how the resource binding

zricethezav (approved):

pkg/analyzer/analyzers/plaid/plaid.go
```diff
@@ -0,0 +1,216 @@
...
+		productName := product
+		if perm, err := PermissionFromString(product); err == nil {
+			if productDetails, ok := permissionToProduct[perm]; ok {
+				productName = productDetails.Name
+			}
+		}
+		result.Bindings[idx] = analyzers.Binding{
+			Resource: analyzers.Resource{
+				Name:               productName,
+				FullyQualifiedName: product,
```
abmussani, line 147: @nabeelalam Since, same name of products are shared among multiple secrets, it will cause an identity issue. You can append `account.id` in FullyQualifiedName, i.e. `account.id + "/" + product`

abmussani (changes requested):

pkg/analyzer/analyzers/plaid/plaid.go
nabeelalam: I've updated this according to your suggestion.
Also, updated the permissions as well according to our discussion if you can take look?

abmussani (approved):

PR #4008, opened 2025-04-04
Allow filesystem exclusion to eagerly prune the enumerated tree

kashifkhan0771:
### Description:
Check if the directory matches the exclude pattern and skip the directory.
Screenshot from 2025-04-07 12-18-25

### Checklist:


pkg/common/filter.go
```diff
@@ -114,3 +114,8 @@ func (rules *FilterRuleSet) Matches(object string) bool {
 	}
 	return false
 }
+
+// ExcludeDir return true if directory path matches with regular expressions in exclude FilterRuleSet.
+func (filter *Filter) ExcludeDir(path string) bool {
```
rosecodym, line 119: You've named this function `ExcludeDir`, but it isn't restricted to directories, right? Could it just be called `Exclude` or (my preference) `ShouldExclude`?

rosecodym (commented):
I honestly thought this would be more complicated, but maybe I was mistaken. How did you test it? And are there any automated tests you could write for it?

pkg/sources/filesystem/filesystem.go
```diff
@@ -127,14 +127,25 @@ func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sour
 			ctx.Logger().Error(err, "error walking directory")
 			return nil
 		}
+
 		fullPath := filepath.Join(path, relativePath)
 
+		// skip excluded directories early
+		if d.IsDir() {
+			if s.filter != nil && s.filter.ExcludeDir(fullPath) {
+				return fs.SkipDir
```
rgmz, line 136: Kinda random, but you can do the nil check inside the method to avoid repeating `filter != nil` in multiple places.

https://go.dev/play/p/GbvkbdAQTQd

pkg/common/filter.go
rgmz: Wouldn't it only be for paths regardless?

camgunz: My vote would be to just call `filter.exclude.Matches` itself rather than wrap it w/ this method

kashifkhan0771: > You've named this function ExcludeDir, but it isn't restricted to directories, right? Could it just be called Exclude or (my preference) ShouldExclude?

You're right, the method itself isn’t restricted to directories. However, the current implementation is specifically intended for directories, since we only call it within an `if d.IsDir()` condition.
I agree that the naming could be misleading. To align behavior with naming, would it make sense to move the IsDir check inside the method? That way, the method clearly applies only to directories, and the name ShouldExcludeDir (or similar) remains accurate.

Let me know what you prefer!

kashifkhan0771: > My vote would be to just call `filter.exclude.Matches` itself rather than wrap it w/ this method

We can't call `.exclude.Matches` directly in `filesystem.go` because the exclude FilterRuleSet type is unexported.

kashifkhan0771:
@rosecodym I wrote the test case and renamed the method to `ShouldExclude`.

pkg/common/filter.go
kashifkhan0771: @rgmz renamed the method.

camgunz:
Your (good, correct and fair) point which was polite version of "hey maybe look at the `Filter` code" made me take a deeper look, and now I'm curious why you don't use the existing `if s.filter != nil && !s.filter.Pass(...)` test used [further down in this method](https://github.com/trufflesecurity/trufflehog/blob/a92f01dfe5ecb19172075272f674e5186e416d90/pkg/sources/filesystem/filesystem.go#L149) and in [`Enumerate`](https://github.com/trufflesecurity/trufflehog/blob/a92f01dfe5ecb19172075272f674e5186e416d90/pkg/sources/filesystem/filesystem.go#L229). It seems like if we use this new `ShouldExclude` method instead of `Pass` we ignore `include` filters, which would be new behavior I think?

kashifkhan0771:
> Your (good, correct and fair) point which was a polite version of "hey maybe look at the `Filter` code" made me take a deeper look (😅), and now I'm curious why you don't use the existing `if s.filter != nil && !s.filter.Pass(...)` test used [further down in this method](https://github.com/trufflesecurity/trufflehog/blob/a92f01dfe5ecb19172075272f674e5186e416d90/pkg/sources/filesystem/filesystem.go#L149) and in [`Enumerate`](https://github.com/trufflesecurity/trufflehog/blob/a92f01dfe5ecb19172075272f674e5186e416d90/pkg/sources/filesystem/filesystem.go#L229). It seems like if we use this new `ShouldExclude` method instead of `Pass` we ignore `include` filters, which would be new behavior I think?

The Pass method returns true if the include FilterRuleSet matches the pattern and the exclude FilterRuleSet does not. For this specific use case, we only need to determine if the directory path matches any pattern in the exclude FilterRuleSet. If it does, we skip the directory. There's no need to check the include FilterRuleSet. In simple terms, this method tells us either 'Yes, you can skip this path' or 'No, you cannot skip this path.'

kashifkhan0771:
> Right so like, with the current logic that uses `!s.filter.Pass` we're ignoring a path if:
> 
> * (there is no exclude filter OR that exclude filter doesn't match the path)
> * AND (there is no include filter OR that include filter matches the path)
> 
> (`s.filter` is never `nil` so I skipped that part)
> 
> With the new logic (`!s.filter.shouldExclude`) we're ignoring a path if:
> 
> * (there is no exclude filter OR that exclude filter doesn't match the path)


thinking

Hmm, the `Pass` method checks if the path matches any include filter and does not match any exclude filter. In our case, we're inverting its output with `!s.filterPass()`, so if the path doesn't match an include filter and matches any exclude filter, it enters the condition on `line 149` in `filesystem.go`.
If we add the following logic inside that condition, it should work as expected as the `!Pass()` will make sure the path does not match any pattern in include filter and matches any one pattern in exclude filter.
```go
if d.IsDir() {
    return fs.SkipDir
}
```

Thanks @camgunz ❤️ and let me know If I misunderstood anything from your comment.

camgunz:
Lol yeah man, same team! :heart:

camgunz (approved):

pkg/sources/filesystem/filesystem.go
```diff
@@ -135,7 +136,13 @@ func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sour
 		if !d.Type().IsRegular() {
```
rosecodym, line 136: `IsRegular` is false for directories, meaning that the function will return here and never reach the logic you've added later to return `SkipDir`. You need to do that check before here somehow.

rosecodym (changes requested):
It looks like your new logic will never run.

The fact that no automated (or manual) test flagged this suggests we should figure some out. Specifically, we need a test to ensure that files within skipped directories are never iterated at all, not just that they're not scanned.

pkg/sources/filesystem/filesystem.go
kashifkhan0771: I moved the logic above this check.

**Question for learning:** Why we want to skip the non-regular files? If we want to skip directories through this check why are we not using `IsDir` method?

rosecodym: I'm not sure, but it looks like the change from excluding directories to excluding all non-regular files was intentional (5ab5c6f9d98661916f073279708b63497270fed1), so we should leave it in.

pkg/sources/filesystem/filesystem.go
```diff
@@ -127,17 +127,25 @@ func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sour
 			ctx.Logger().Error(err, "error walking directory")
 			return nil
 		}
+
 		fullPath := filepath.Join(path, relativePath)
 
+		// check if the full path is not matching any pattern in include FilterRuleSet and matching any exclude FilterRuleSet.
+		if s.filter != nil && !s.filter.Pass(fullPath) {
```
rosecodym, line 134: Will this correctly scan files that are explicitly included but exist in directories that are _not_ explicitly included? For example, say the scan is run with this inclusion file:

```
foo/test\.txt
```

`foo` doesn't match the filter, so won't it be skipped? This means that all files inside `foo` will also be skipped - including `test.txt`.

I think this would be a great automated test to write - either it will prove that I'm wrong, or it will ensure that this regression doesn't happen here or in the future.

rosecodym (commented):
I would still like some evidence that this actually skips directories. I don't know if we can write an automated test for it, but it seems possible to verify during dev-testing by modifying the source to print some output for each file enumerated and check that fewer files are being actually enumerated when directories are skipped.

kashifkhan0771:
> I would still like some evidence that this actually skips directories. I don't know if we can write an automated test for it, but it seems possible to verify during dev-testing by modifying the source to print some output for each file enumerated and check that fewer files are being actually enumerated when directories are skipped.

I previously tested this behavior and just reran a quick test to confirm. I created a directory named `test-dir` containing a few files, including one with a fake secret. In the initial run without excluding the directory the secret was detected, and the file was scanned as expected. After updating the exclude file to ignore `test-dir`, I reran the scan, and this time the secret was not detected, confirming that the directory was successfully skipped.

Let me know if you'd like me to run more in-depth tests or cover any specific scenarios.

### Test ScreenShot:
Screenshot from 2025-04-09 13-18-03

kashifkhan0771:
> I would still like some evidence that this actually skips directories. I don't know if we can write an automated test for it, but it seems possible to verify during dev-testing by modifying the source to print some output for each file enumerated and check that fewer files are being actually enumerated when directories are skipped.

I previously tested this behavior and just reran a quick test to confirm. I created a directory named `test-dir` containing a few files. In the first run, without excluding the directory, the paths were scanned as expected. After updating the exclude file to ignore `test-dir`, I reran the scan, and this time no paths were scanned, confirming that the directory was successfully skipped. In a third run, I excluded a subdirectory within `test-dir` instead. As expected, files inside `test-dir` were scanned, but the files inside the excluded subdirectory were not.

Let me know if you'd like me to run more in-depth tests or cover any specific scenarios.

### Screenshots
Screenshot from 2025-04-09 15-04-31
Screenshot from 2025-04-09 15-04-51

pkg/sources/filesystem/filesystem.go
kashifkhan0771: Off the top of my head, I assumed it wouldn't be scanned since excluding a directory prevents traversal into it, meaning any files inside would be skipped. To confirm, I ran a test where I explicitly included `file.txt` but excluded the parent directory `test-dir`. As expected, the file wasn't scanned because the directory itself was skipped entirely.
I will try to write some automated test cases for this scenario as well.

###Screenshot
Screenshot from 2025-04-09 15-13-07

kashifkhan0771: I’ve updated the automated test case to skip a directory while explicitly including a file inside it. The results matched the manual test I ran earlier.

rosecodym: What you're describing is the case where a directory is explicitly excluded but a file within that directory is specifically included. I don't think the expected behavior in that case is well-defined, so I'm not worried about it.

What I'm instead concerned about is the case where a file in a subdirectory is explicitly included and **nothing** is explicitly excluded. In that case, the file should be scanned, but I believe that this implementation changes things so that it is not.

This could be covered by an automated test - can you write one for it?

kashifkhan0771: I completed the dev testing and confirmed that a file located in a subdirectory is successfully scanned when it is explicitly included, and no exclusions are set.

### Screenshot
Screenshot from 2025-04-10 12-33-39

kashifkhan0771: While writing the automated test case, I noticed that the following logic was skipping directories that matched neither the include nor the exclude filters:
```go
if s.filter != nil && !s.filter.Pass(fullPath) {
    if d.IsDir() {
        return fs.SkipDir
    }
    return nil
}
```
Since Pass() returns false when a path doesn't match either filter set, the logic ended up skipping such directories by default. I'm not sure if this behavior is intended, but it caused issues in this scenario. I added an additional check to ensure we only skip directories that are explicitly excluded, and now it behaves as expected.

**Updated Logic:**
```go
if s.filter != nil && !s.filter.Pass(fullPath) {
	// skip excluded directories
	if d.IsDir() && s.filter.ShouldExclude(fullPath) {
		return fs.SkipDir
	}

	return nil // skip the file
}
```

pkg/sources/filesystem/filesystem_test.go
```diff
@@ -300,6 +300,83 @@ func TestChunkUnitReporterErr(t *testing.T) {
...
+	err = s.ChunkUnit(ctx, sources.CommonSourceUnit{
+		ID: ignoreDir,
+	}, &reporter)
+	assert.NoError(t, err)
+
+	assert.Equal(t, 0, len(reporter.Chunks), "Expected no chunks from excluded directory")
+	assert.Equal(t, 0, len(reporter.ChunkErrs), "Expected no errors for excluded directory")
+}
+
+func TestScanSubDirFile(t *testing.T) {
```
kashifkhan0771, line 337: Screenshot from 2025-04-10 12-49-33

pkg/sources/filesystem/filesystem.go
rosecodym: > Since Pass() returns false when a path doesn't match either filter set, the logic ended up skipping such directories by default. I'm not sure if this behavior is intended, but it caused issues in this scenario. I added an additional check to ensure we only skip directories that are explicitly excluded, and now it behaves as expected.

This is exactly what I'd been worried about. Thanks for the test!

pkg/sources/filesystem/filesystem_test.go
```diff
@@ -300,6 +300,83 @@ func TestChunkUnitReporterErr(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestSkipDir(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+
+	// create a temp directory with files
+	ignoreDir, cleanupDir, err := createTempDir("", "ignore1", "ignore2", "ignore3")
+	assert.NoError(t, err)
```
rosecodym, line 309: I like using `require` instead of `assert` during test setup because if setup fails it doesn't make sense to run the rest of the test.

rosecodym (approved):
Thank you for this! And for persisting through all my questions :)

PR #4010, opened 2025-04-04
Return git log error via channel

rosecodym:
### Description:
This is a sketch of some ideas I wanted to share with some people. It's not intended to be merged as-is.

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -229,7 +229,7 @@ func (c *Parser) RepoPath(
 	excludedGlobs []string,
 	isBare bool,
 	additionalArgs ...string,
-) (chan *Diff, error) {
+) (<-chan *Diff, <-chan error) {
 	args := []string{
 		"-C", source,
 		"log",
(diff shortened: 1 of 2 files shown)


pkg/sources/git/git.go
```diff
@@ -715,6 +709,11 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 			return err
 		}
 	}
+
+	for err := range errChan {
+		return err
+	}
```
rosecodym, line 715: i think it'd be better to just do a single select with a `default` case

pkg/gitparse/gitparse.go
```diff
@@ -291,21 +291,28 @@ func (c *Parser) Staged(ctx context.Context, source string) (chan *Diff, error)
 }
 
 // executeCommand runs an exec.Cmd, reads stdout and stderr, and waits for the Cmd to complete.
```
mcastorina, line 293: We aren't waiting for Cmd to complete, right? Or.. we are but asynchronously?

PR #4013, opened 2025-04-07
fixed authOmanagement detector integration test failure

kashifkhan0771:
### Description:
This PR fix the integration test case of `authOmanagementapitoken` detector.
Screenshot from 2025-04-07 15-47-40

### Checklist:


pkg/detectors/auth0managementapitoken/auth0managementapitoken_integration_test.go
```diff
@@ -52,8 +52,9 @@ func TestAuth0ManagementApiToken_FromChunk(t *testing.T) {
 			want: []detectors.Result{
 				{
 					DetectorType: detectorspb.DetectorType_Auth0ManagementApiToken,
+					RawV2:        []byte(managementApiToken + domain),
 					Redacted:     domain,
-					Verified:     true,
+					Verified:     false, // keeping this false, because the token can have a maximum expiration of 30 days.
```
abmussani, line 57: changing the false is little bit misleading as the name says `found, verified`

pkg/detectors/auth0managementapitoken/auth0managementapitoken_integration_test.go
```diff
@@ -52,6 +52,7 @@ func TestAuth0ManagementApiToken_FromChunk(t *testing.T) {
 			want: []detectors.Result{
 				{
 					DetectorType: detectorspb.DetectorType_Auth0ManagementApiToken,
+					RawV2:        []byte(managementApiToken + domain),
```
ahrav, line 55: question: This might be a dumb question, but if the test fails, do we risk leaking the actual credential in the test output during the comparison?

ahrav (approved):
One slightly dumb question, otherwise LGTM!

pkg/detectors/auth0managementapitoken/auth0managementapitoken_integration_test.go
kashifkhan0771: That's the approach for all the detectors integration tests 😅. We pull our secrets from GCP and use them for the testing.
But as you mentioned now I realised that's we may need to do something about it I guess.

ahrav: Maybe it's okay, given you need to have permissions to even access the secret from GSM.

Issue #2267, opened 2024-01-01
Multi-part secrets are not reliably detected


rgmz:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

### TruffleHog Version
Latest

### Trace Output
N/A

### Expected Behavior

Detectors like [DockerHub](https://github.com/trufflesecurity/trufflehog/blob/857a37160074198a128f176f6534fd318d3c9494/pkg/detectors/dockerhub/dockerhub.go#L25) which require both a username and password should pick up multi-part secrets. 

Example: TruffleHog should pick up `johndoe:dckr_pat_rotated` and `johndoe:dckr_pat_newandactive` when processing commits.
```diff
    DOCKERHUB_USER = 'johndoe'
-    DOCKERHUB_PASSWORD = 'dckr_pat_rotated'
+    DOCKERHUB_PASSWORD = 'dckr_pat_newandactive'
```

### Actual Behavior

TruffleHog only searches *added* lines for Git sources, meaning that the commit shown above will be processed without any surrounding context:
```

    DOCKERHUB_PASSWORD = 'dckr_pat_newandactive'

```

This is due to how the GitParse logic is written:
https://github.com/trufflesecurity/trufflehog/blob/857a37160074198a128f176f6534fd318d3c9494/pkg/gitparse/gitparse.go#L391-L396

### Steps to Reproduce
N/A

## Environment
N/A

## Additional Context
N/A

### References
N/A

PR #4015, opened 2025-04-07
Fix `git` scan not seeing secret in updated cert file

martinlocklear:
### Description:
At the moment, when doing a `git` scan of a repository where a private key replaced in a cert file, this secret won't be detected. Ex: The update in [this commit](https://github.com/BrowserSync/browser-sync/commit/cbe2ccb74b51b76fc845d3d786f32c3ee8e90d07) to `server.key`.

Seem likely that this problem is connected to [this issue](https://github.com/trufflesecurity/trufflehog/issues/2267).

This PR is a WIP

### Checklist:


pkg/gitparse/gitparse.go
```diff
@@ -235,6 +235,10 @@ func (c *Parser) RepoPath(
 		"log",
 		"--patch", // https://git-scm.com/docs/git-log#Documentation/git-log.txt---patch
 		"--full-history",
+		// Limit ourselves to a single line of context.  This context is necessary to include the match key for
+		// matching detectors when the match string isn't on the same line as the secret, and the match key
+		// previously existed (as when updating private key certs)
+		"--unified=1", // https://git-scm.com/docs/git-log#Documentation/git-log.txt-code--unifiedltngtcode
```
zricethezav, line 241: Won't this introduce the risk of accidentally associating a finding with the wrong commit? E.g:
```
MYTOKEN=examplesecret
+changedline
```

rgmz:
> Going to have to consider this one more carefully, to make sure that we don't break the attribution to commits.

IIRC you'd need to parse the hunk line numbers and ensure the secret line overlaps with the changed line. Then you could have more than a single line of context (desirable) and not risk improper attribution.

martinlocklear:
Ok - the scope on this work expanded past the bandwidth I have to give it unfortunately.  Closed, but not forgotten.

PR #4018, opened 2025-04-07
Better Test Output For Git Diff Parsing

martinlocklear:
This test file previously compared parsed diffs using a method `Equal` on the Diff struct. This would fail (and stop) at the first mismatch, and left reporting the failure to the caller.  The existing test output made it hard to understand _what_ the difference was when the difference was in the diff content itself because the _content_ of the contentWriter field wasn't actually processed and printed.

This change adds an `assertDiffEqualToExpected` function that leverages much of the existing test structure but centralizes the assertion _and_ reporting of the differences into a single reusable method.

### Description:

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse_test.go
+++ b/pkg/gitparse/gitparse_test.go
@@ -2,6 +2,8 @@ package gitparse
 
 import (
 	"bytes"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"strings"
@@ -748,16 +750,19 @@ func TestToFileLinePathParse(t *testing.T) {
 // diffs.
 func assertDiffEqualToExpected(t *testing.T, expected *Diff, actual *Diff) {
 
-	// Assert on the top level fields first
-	assert.Equal(t, expected.PathB, actual.PathB)
-	assert.Equal(t, expected.LineStart, actual.LineStart)
-	assert.Equal(t, expected.IsBinary, actual.IsBinary)
+	// Use `cmp.Diff` to automatically compare all the exported fields.  This allows this test to grow automatically if
+	// new exported fields are added to these structs.  However, the most important field we want to test is unexpected
+	// (i.e. contentWriter) which is where the actual content of the diff is stored.  We break this out next.
+	opts := []cmp.Option{
+		cmpopts.IgnoreUnexported(Diff{}, Commit{}, strings.Builder{}),
+		cmpopts.IgnoreFields(Commit{}, "Size"),
+	}
+	if diff := cmp.Diff(expected, actual, opts...); diff != "" {
+		t.Errorf("%s", diff)
+	}
 
-	// Check the commit details looks the same
-	assert.Equal(t, expected.Commit.Hash, actual.Commit.Hash)
-	assert.Equal(t, expected.Commit.Author, actual.Commit.Author)
-	assert.Equal(t, expected.Commit.Message.String(), actual.Commit.Message.String())
-	assert.Equal(t, expected.Commit.Date, actual.Commit.Date)
+	// Here's where we compare the actual content of the diff. We break that out and test it separately so that we can
+	// keep this test relatively easy to understand and still get meaningful test output on the diff itself
 
 	// If the test author hasn't specified a contentWriter, then we don't want to explode, but we _do_
 	// want to confirm that the actual diff _also_ is nil there
@@ -773,6 +778,8 @@ func assertDiffEqualToExpected(t *testing.T, expected *Diff, actual *Diff) {
 		assert.NoError(t, err)
 		assert.Equal(t, expectedDiffStr, actualDiffStr)
 	}
+	// TODO - Add test coverage for binary diffs (if it isn't already elsewhere)
+
 }
 
 func TestCommitParsing(t *testing.T) {
--- a/pkg/gitparse/gitparse_test.go
+++ b/pkg/gitparse/gitparse_test.go
@@ -2,6 +2,8 @@ package gitparse
 
 import (
 	"bytes"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/gitparse/gitparse_test.go
```diff
@@ -742,51 +744,35 @@ func TestToFileLinePathParse(t *testing.T) {
...
-	default:
-		if d1.contentWriter != nil && d2.contentWriter != nil {
-			equal, err := isEqualContentString(d1.contentWriter, d2.contentWriter)
-			if err != nil || !equal {
-				ctx.Logger().Error(err, "failed to compare diff content")
-				return false
-			}
-		}
+	// Check that the content of the diff itself is as expected for non-binary diffs
+	if expected.contentWriter != nil && !actual.IsBinary {
```
rosecodym, line 768: Does this mean that if `actual.IsBinary` is true (and `expected` has a `contentWriter` at all) we presume that the contents are equal without checking?

rosecodym (commented):
discussed synchronously too

pkg/gitparse/gitparse_test.go
martinlocklear: I think my change highlights this issue with the tests itself.  As far as I can tell, we don't test binary diffs at all (at least not in this file).  I neglected to carry over a [previous comment](https://github.com/trufflesecurity/trufflehog/pull/4018/files#diff-8e5655d96965a2fe08c84b2d6a250ff4f78de92c30011038297e6e8a6a691812L749) which checked for "binary-ness" of the diff heuristically.  I swapped to using the flag since it seemed more explicit.

There are only a few test cases which have binary content, and that content is always either nil or empty string.

camgunz:
Did you have a look at what the output is if you just use `assert.Equal` on the `Diff`s themselves? I don't remember it being particularly bad, but my standards are pretty weird and my memory is effectively non-existent.

martinlocklear:
> Did you have a look at what the output is if you just use `assert.Equal` on the `Diff`s themselves? I don't remember it being particularly bad, but my standards are pretty weird and my memory is effectively non-existent.

This was an excellent sanity check - I'm actually not sure that I had done this.  Here's the output from a test that I'm forcing to fail by making the expected diff content wrong, on a multi-line diff.  The content of the diff itself (inside the BufferWriter) isn't there - not sure if there's a good way to deal with that.  `cmp.Diff` (with least config) has the exact same output.

```
=== RUN   TestCommitParsing
    gitparse_test.go:805:   &gitparse.Diff{
          	PathB:     "C++/1 Урок/.idea/workspace.xml",
          	LineStart: 29,
          	IsBinary:  false,
          	Commit: &gitparse.Commit{
          		... // 1 ignored and 4 identical fields
        - 		Size: 0,
        + 		Size: 220,
          		... // 1 ignored field
          	},
          	... // 1 ignored field
          }
--- FAIL: TestCommitParsing (0.03s)
```

Here's the output for the same situation with the current patch (a bunch of asserts).  This has the advantage of the line numbers corresponding to asserts that are very explicit about what they're checking.

```
=== RUN   TestCommitParsing
    gitparse_test.go:769: 
        	Error Trace:	/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:769
        	            				/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:852
        	Error:      	Not equal: 
        	            	expected: 226
        	            	actual  : 220
        	Test:       	TestCommitParsing
    gitparse_test.go:774: 
        	Error Trace:	/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:774
        	            				/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:852
        	Error:      	Not equal: 
        	            	expected: "\n\n\nMARTIN            <state relative-caret-position=\"72\">\n              <caret line=\"4\" column=\"0\" lean-forward=\"false\" selection-start-line=\"4\" selection-start-column=\"0\" selection-end-line=\"4\" selection-end-column=\"0\" />\n\n\n\n"
        	            	actual  : "\n\n\n            <state relative-caret-position=\"72\">\n              <caret line=\"4\" column=\"0\" lean-forward=\"false\" selection-start-line=\"4\" selection-start-column=\"0\" selection-end-line=\"4\" selection-end-column=\"0\" />\n\n\n\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -3,3 +3,3 @@
        	            	 
        	            	-MARTIN            <state relative-caret-position="72">
        	            	+            <state relative-caret-position="72">
        	            	               <caret line="4" column="0" lean-forward="false" selection-start-line="4" selection-start-column="0" selection-end-line="4" selection-end-column="0" />
        	Test:       	TestCommitParsing
--- FAIL: TestCommitParsing (0.07s)

<Click to see difference>


FAIL
```
 Diving deeper, if we go whole hog into configuring `cmp.Diff`, we get the following code and subsequent output.
 
 ```
 	bufferWriterTransformer := cmp.Transformer("TransformIntoString", func(in *bufferwriter.BufferWriter) string {
		if s, err := in.String(); err == nil {
			return s
		}
		return ""
	})
	opts := cmp.Options{
		bufferWriterTransformer,
		cmp.AllowUnexported(Diff{}),
		cmpopts.IgnoreUnexported(strings.Builder{}, Commit{}, bufferwriter.BufferWriter{}),
	}

	if diff := cmp.Diff(expected, actual, opts); diff != "" {
		t.Errorf("%s", diff)
	}
```

```
=== RUN   TestCommitParsing
    gitparse_test.go:803:   &gitparse.Diff{
          	PathB:     "C++/1 Урок/.idea/workspace.xml",
          	LineStart: 29,
          	IsBinary:  false,
          	Commit: &gitparse.Commit{
          		... // 1 ignored and 4 identical fields
        - 		Size: 0,
        + 		Size: 220,
          		... // 1 ignored field
          	},
        - 	contentWriter: (*bufferwriter.BufferWriter)(Inverse(TransformIntoString, string(
        - 		"""
        - 		
        - 		
        - 		
        - 		MARTIN            <state relative-caret-position="72">
        - 		              <caret line="4" column="0" lean-forward="false" selection-start-line="4" selection-start-column="0" selection-end-line="4" selection-end-column="0" />
        - 		
        - 		
        - 		
        - 		"""
        - 	))),
        + 	contentWriter: (*bufferwriter.BufferWriter)(Inverse(TransformIntoString, string(
        + 		"""
        + 		
        + 		
        + 		
        + 		            <state relative-caret-position="72">
        + 		              <caret line="4" column="0" lean-forward="false" selection-start-line="4" selection-start-column="0" selection-end-line="4" selection-end-column="0" />
        + 		
        + 		
        + 		
        + 		"""
        + 	))),
          }
```
That feels better to me, but here's the output from the existing PR (collection of asserts)
```
=== RUN   TestCommitParsing
    gitparse_test.go:770: 
        	Error Trace:	/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:770
        	            				/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:846
        	Error:      	Not equal: 
        	            	expected: 226
        	            	actual  : 220
        	Test:       	TestCommitParsing
    gitparse_test.go:775: 
        	Error Trace:	/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:775
        	            				/Users/martin.locklear/truffle/trufflehog/pkg/gitparse/gitparse_test.go:846
        	Error:      	Not equal: 
        	            	expected: "\n\n\nMARTIN            <state relative-caret-position=\"72\">\n              <caret line=\"4\" column=\"0\" lean-forward=\"false\" selection-start-line=\"4\" selection-start-column=\"0\" selection-end-line=\"4\" selection-end-column=\"0\" />\n\n\n\n"
        	            	actual  : "\n\n\n            <state relative-caret-position=\"72\">\n              <caret line=\"4\" column=\"0\" lean-forward=\"false\" selection-start-line=\"4\" selection-start-column=\"0\" selection-end-line=\"4\" selection-end-column=\"0\" />\n\n\n\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -3,3 +3,3 @@
        	            	 
        	            	-MARTIN            <state relative-caret-position="72">
        	            	+            <state relative-caret-position="72">
        	            	               <caret line="4" column="0" lean-forward="false" selection-start-line="4" selection-start-column="0" selection-end-line="4" selection-end-column="0" />
        	Test:       	TestCommitParsing
--- FAIL: TestCommitParsing (0.07s)
```

I like that output best personally, but I'm not sure I have an opinion the tradeoff between the better output and automatically covering new fields to existing structs with this test.

rosecodym (approved):
This looks pretty helpful!

PR #4034, opened 2025-04-10
Polish logging around git and archives to reduce noise

camgunz:
### Description:
The main change here is in `git.go` to only log a "discarded chunk" message if we indeed discarded bytes. Separate from that, we generate a lot of log messages around archives, and that can be noisy. I pulled most things there down to level 4, but happy to change whatever.

### Checklist:

### DIFF
--- a/pkg/handlers/archive.go
+++ b/pkg/handlers/archive.go
@@ -196,13 +196,13 @@ func (h *archiveHandler) extractorHandler(dataOrErrChan chan DataOrErr) func(con
 
 		fileSize := file.Size()
 		if int(fileSize) > maxSize {
-			lCtx.Logger().V(4).Info("skipping file: size exceeds max allowed", "size", fileSize, "limit", maxSize)
+			lCtx.Logger().V(2).Info("skipping file: size exceeds max allowed", "size", fileSize, "limit", maxSize)
 			h.metrics.incFilesSkipped()
 			return nil
 		}
 
 		if common.SkipFile(file.Name()) || common.IsBinary(file.Name()) {
-			lCtx.Logger().V(4).Info("skipping file: extension is ignored")
+			lCtx.Logger().V(2).Info("skipping file: extension is ignored")
 			h.metrics.incFilesSkipped()
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/handlers/archive.go
```diff
@@ -174,36 +174,35 @@ func (h *archiveHandler) openArchive(
 // to handle nested archives or to continue processing based on the file's content and depth in the archive structure.
 func (h *archiveHandler) extractorHandler(dataOrErrChan chan DataOrErr) func(context.Context, archives.FileInfo) error {
 	return func(ctx context.Context, file archives.FileInfo) error {
+		if common.IsDone(ctx) {
```
camgunz, line 177: I also bumped this up to the top; felt prudent to bail immediately if this is the case.

pkg/handlers/archive.go
```diff
@@ -203,7 +202,7 @@ func (h *archiveHandler) extractorHandler(dataOrErrChan chan DataOrErr) func(con
 		}
 
 		if common.SkipFile(file.Name()) || common.IsBinary(file.Name()) {
-			lCtx.Logger().V(3).Info("skipping file: extension is ignored")
+			lCtx.Logger().V(2).Info("skipping file: extension is ignored")
```
camgunz, line 205: I actually bumped this up from 3 to 2 to match the level at which we log skipping a file when the size is too large.

pkg/handlers/archive.go
```diff
@@ -233,7 +232,7 @@ func (h *archiveHandler) extractorHandler(dataOrErrChan chan DataOrErr) func(con
 		rdr, err := newFileReader(ctx, f)
 		if err != nil {
 			if errors.Is(err, ErrEmptyReader) {
-				lCtx.Logger().V(5).Info("empty reader, skipping file")
+				lCtx.Logger().V(4).Info("empty reader, skipping file")
```
camgunz, line 235: Dunno how often this happens, but 5 seemed harsh haha

pkg/handlers/archive.go
rgmz: This condition creates an excessive amount of noise, compared to the rarer scenario of a file being too large.

rosecodym: I agree; extension-based skips happen a _lot_ and are rarely interesting from a user perspective. I think 3 was fine for this and would be worried about lowering it.

pkg/handlers/archive.go
rosecodym: I think we should leave this one alone too; while the operational difference between 4 and 5 isn't huge, lowering a log level for something we don't know the frequency of seems unnecessarily risky.

pkg/handlers/archive.go
camgunz: OK cool; put it back to 3 👍🏻

rosecodym (approved):

PR #4045, opened 2025-04-14
Okta Detector - Added Check for Indeterminate Verification Results

nabeelalam:
### Description:
This pull request updates the Okta detector to add ability to return indeterminate verification results along with error messages

### Checklist:


kashifkhan0771 (approved):
❤️

pkg/detectors/okta/okta.go
```diff
@@ -34,55 +32,72 @@ func (s Scanner) Keywords() []string {
...
 // FromData will find and optionally verify Okta secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
-	for _, tokenMatch := range tokenPat.FindAll(data, -1) {
-		token := string(tokenMatch)
+	dataStr := string(data)
 
-		for _, domainMatch := range domainPat.FindAll(data, -1) {
-			domain := string(domainMatch)
+	var uniqueTokens = make(map[string]struct{})
+	var uniqueDomains = make(map[string]struct{})
```
abmussani, line 38: nit: These variables can be initialized in one statement.
```
var uniqueTokens, uniqueDomains = make(map[string]struct{}), make(map[string]struct{})
```

pkg/detectors/okta/okta.go
```diff
@@ -34,55 +32,72 @@ func (s Scanner) Keywords() []string {
...
+	url := fmt.Sprintf("https://%s/api/v1/users/me", domain)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return false, err
+	}
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", fmt.Sprintf("SSWS %s", token))
+
+	resp, err := detectors.DetectorHttpClientWithNoLocalAddresses.Do(req)
```
abmussani, line 85: `DetectorHttpClientWithNoLocalAddresses` Http client is being used and it cannot be override as per the need. I like the idea from [SalesForce Detector](https://github.com/trufflesecurity/trufflehog/blob/479cf4eb3fe5652218b1cdf1e6b9be8e48e51280/pkg/detectors/salesforce/salesforce.go#L29). Can we follow the same in this one ?

nabeelalam: Yes, that's a good idea. I'll add that.

ahrav (approved):

zricethezav:
nice work, thanks @nabeelalam

PR #4048, opened 2025-04-14
test(sources/s3): fix infinite blocking and timeout issue in TestSource_Chunks

Juneezee:
### Description:

This PR addresses two issues:

1. Duplicate test functions in `s3_integration_test.go`.

   The file currently contains two separate `TestSourceChunksNoResumption`. The only differences between them are the use of `t.Parallel()` in one and the bucket name. This PR merges both into a single, table-driven test

   

   https://github.com/trufflesecurity/trufflehog/blob/a1243a404d8a6a908fd474055790f4ce3d8add17/pkg/sources/s3/s3_integration_test.go#L85-L114

   https://github.com/trufflesecurity/trufflehog/blob/a1243a404d8a6a908fd474055790f4ce3d8add17/pkg/sources/s3/s3_integration_test.go#L251-L282

2. Possible infinite block in `TestSource_Chunks`

   As reported in https://github.com/trufflesecurity/trufflehog/pull/4069#issuecomment-2819473175, the test `TestSource_Chunks` in `s3_test.go` may block indefinitely. This is due to the use of an **unbuffered** channel (`chunksCh`), combined with only a single receive operation (`gotChunk := <-chunksCh`). If the test bucket contains more than one chunk, the `s.Chunks(ctx, chunksCh)` call will block, causing the test to hang.

   See https://github.com/trufflesecurity/trufflehog/pull/4048#discussion_r2054456065 for full explanation.

   https://github.com/trufflesecurity/trufflehog/blob/a1243a404d8a6a908fd474055790f4ce3d8add17/pkg/sources/s3/s3_test.go#L120-L131

### Checklist:

### DIFF
--- a/pkg/sources/s3/s3_test.go
+++ b/pkg/sources/s3/s3_test.go
@@ -4,7 +4,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"os"
-	"sync"
 	"testing"
 	"time"
 
@@ -99,8 +98,7 @@ func TestSource_Chunks(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
-			var cancelOnce sync.Once
-			defer cancelOnce.Do(cancel)
+			defer cancel()
 
 			for k, v := range tt.init.setEnv {
 				t.Setenv(k, v)
@@ -117,26 +115,29 @@ func TestSource_Chunks(t *testing.T) {
 				t.Errorf("Source.Init() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			chunksCh := make(chan *sources.Chunk)
-			var wg sync.WaitGroup
-			wg.Add(1)
+			chunksCh := make(chan *sources.Chunk, 1)
 			go func() {
-				defer wg.Done()
+				defer close(chunksCh)
 				err = s.Chunks(ctx, chunksCh)
 				if (err != nil) != tt.wantErr {
 					t.Errorf("Source.Chunks() error = %v, wantErr %v", err, tt.wantErr)
 					os.Exit(1)
 				}
 			}()
-			gotChunk := <-chunksCh
-			wantData, _ := base64.StdEncoding.DecodeString(tt.wantChunkData)
 
-			if diff := pretty.Compare(gotChunk.Data, wantData); diff != "" {
-				t.Errorf("%s: Source.Chunks() diff: (-got +want)\n%s", tt.name, diff)
+			select {
+			case <-ctx.Done():
+				close(chunksCh)
+				t.Errorf("TestSource_Chunks timed out: %v", ctx.Err())
+			case gotChunk := <-chunksCh:
+				wantData, _ := base64.StdEncoding.DecodeString(tt.wantChunkData)
+
+				if diff := pretty.Compare(gotChunk.Data, wantData); diff != "" {
+					t.Errorf("%s: Source.Chunks() diff: (-got +want)\n%s", tt.name, diff)
+				}
+				assert.Equal(t, "", s.GetProgress().EncodedResumeInfo)
+				assert.Equal(t, int64(100), s.GetProgress().PercentComplete)
 			}
-			wg.Wait()
-			assert.Equal(t, "", s.GetProgress().EncodedResumeInfo)
-			assert.Equal(t, int64(100), s.GetProgress().PercentComplete)
 		})
 	}
 }
--- a/pkg/sources/s3/s3_test.go
+++ b/pkg/sources/s3/s3_test.go
@@ -125,19 +125,36 @@ func TestSource_Chunks(t *testing.T) {
 				}
 			}()
 
-			select {
-			case <-ctx.Done():
-				close(chunksCh)
-				t.Errorf("TestSource_Chunks timed out: %v", ctx.Err())
-			case gotChunk := <-chunksCh:
-				wantData, _ := base64.StdEncoding.DecodeString(tt.wantChunkData)
+			waitFn := func() {
+				receivedFirstChunk := false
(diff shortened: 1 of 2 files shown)


pkg/sources/s3/s3_test.go
```diff
@@ -117,26 +115,29 @@ func TestSource_Chunks(t *testing.T) {
...
 				return
 			}
-			chunksCh := make(chan *sources.Chunk)
-			var wg sync.WaitGroup
-			wg.Add(1)
+			chunksCh := make(chan *sources.Chunk, 1)
 			go func() {
-				defer wg.Done()
+				defer close(chunksCh)
 				err = s.Chunks(ctx, chunksCh)
```
Juneezee, line 121: /cc @rosecodym 

Here is a more detailed explanation of why `TestSource_Chunks` may block indefinitely.

The issue arises after the call to `s.Chunks(ctx, chunksCh)`. The relevant call stack is:

1. `(*Source).Chunks`

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/sources/s3/s3.go#L390-L398

2. `(*Source).scanBuckets`

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/sources/s3/s3.go#L363

3. `(*Source).pageChunker`

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/sources/s3/s3.go#L579-L583

4. `HandleFile`

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/handlers/handlers.go#L390

5. `handleChunksWithError`

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/handlers/handlers.go#L426-L428

6. `(ChanReporter).ChunkOk`

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/sources/legacy_reporters.go#L15-L17

7. And blocks infinitely in `CancellableWrite`. Because `chunksCh` is an **unbuffered** channel, the `ch <- item` case can never proceed since we only receive from `chunksCh` once.

   https://github.com/trufflesecurity/trufflehog/blob/6d4ccfa8f6dc80ba321a6a781f609f02e12344ef/pkg/common/context.go#L23-L35


After changing `chunksCh` to a **buffered** channel, I created a public S3 bucket and can confirm that the test no longer hangs infinitely:

https://github.com/user-attachments/assets/4b18c7d0-87f0-43a7-9d1d-45b010c10f8a

rosecodym (dismissed):
@Juneezee thank you for your detailed investigation!

rosecodym (changes requested):
Thanks for this! Unfortunately, it introduces a new test failure:
```
=== RUN   TestSource_Chunks/gets_chunks_after_assuming_role
2025-04-23T09:40:41-04:00	info-0	context	Creating checkpointer	{"timeout": 30}
2025-04-23T09:40:42-04:00	info-0	context	Starting new scan from beginning	{"timeout": 30, "role": "arn:aws:iam::619888638459:role/s3-test-assume-role"}
    s3_test.go:138: 
        	Error Trace:	/Users/cody.rose/src/trufflehog/pkg/sources/s3/s3_test.go:138
        	Error:      	Not equal: 
        	            	expected: ""
        	            	actual  : "{\"current_bucket\":\"truffletestbucket-s3-role-assumption\",\"start_after\":\"s\"}"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-
        	            	+{"current_bucket":"truffletestbucket-s3-role-assumption","start_after":"s"}
        	Test:       	TestSource_Chunks/gets_chunks_after_assuming_role
    s3_test.go:139: 
        	Error Trace:	/Users/cody.rose/src/trufflehog/pkg/sources/s3/s3_test.go:139
        	Error:      	Not equal: 
        	            	expected: 100
        	            	actual  : 90
        	Test:       	TestSource_Chunks/gets_chunks_after_assuming_role
--- FAIL: TestSource_Chunks (6.15s)
    --- FAIL: TestSource_Chunks/gets_chunks_after_assuming_role (5.66s)
```
The problem is that the test no longer waits for `Chunks` to return before checking the source's resumption info and progress completion, so there's a mismatch. I believe this can be resolved by re-adding the removed waitgroup functionality.

pkg/sources/s3/s3_test.go
```diff
@@ -117,26 +115,46 @@ func TestSource_Chunks(t *testing.T) {
...
+
+						receivedFirstChunk = true
+						wantData, _ := base64.StdEncoding.DecodeString(tt.wantChunkData)
+
+						if diff := pretty.Compare(gotChunk.Data, wantData); diff != "" {
+							t.Logf("%s: Source.Chunks() diff: (-got +want)\n%s", tt.name, diff)
+						}
+					}
+				}
 			}
```
Juneezee, line 156: @rosecodym Could you try running the test again with the latest commit (https://github.com/trufflesecurity/trufflehog/pull/4048/commits/6e020ff3f17bdee70fe100962de3206e8b5eb613)?

> I believe this can be resolved by re-adding the removed waitgroup functionality.

Actually, a `WaitGroup` alone is not sufficient here. For `s.Chunks(ctx, chunksCh)` to finish completely, the `chunksCh` channel must be fully drained.

If we use a `WaitGroup` without draining `chunksCh`, the test will still block indefinitely, unless `chunksCh` is a buffered channel with a large enough buffer to hold all the chunks.

rosecodym:
I see passing tests now!

rosecodym (approved):
Thanks for cleaning this up!

PR #4056, opened 2025-04-15
Stop logging chunks

rosecodym:
### Description:
We should not be logging raw chunks, even at level 5 after decoder failures. Removing the chunk from the message makes it effectively useless, so I removed the message entirely.

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -785,7 +785,7 @@ func (e *Engine) scannerWorker(ctx context.Context) {
 			decodeLatency.WithLabelValues(decoder.Type().String(), chunk.SourceName).Observe(float64(decodeTime))
 
 			if decoded == nil {
-				ctx.Logger().V(5).Info("decoder not applicable for chunk", "decoder", decoder.Type().String(), "chunk", chunk)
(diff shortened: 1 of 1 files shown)


abmussani (approved):

pkg/engine/engine.go
```diff
@@ -785,7 +785,7 @@ func (e *Engine) scannerWorker(ctx context.Context) {
 			decodeLatency.WithLabelValues(decoder.Type().String(), chunk.SourceName).Observe(float64(decodeTime))
 
 			if decoded == nil {
-				ctx.Logger().V(5).Info("decoder not applicable for chunk", "decoder", decoder.Type().String(), "chunk", chunk)
+				ctx.Logger().V(5).Info("decoder not applicable for chunk", "decoder", decoder.Type().String())
```
rgmz, line 788: Without the chunk, this log statement is effectively useless.

rosecodym: ...that is an excellent point

camgunz (approved):

camgunz (approved):

camgunz:
TIL you can approve twice

martinlocklear (approved):

PR #4057, opened 2025-04-15
feat(docker): implement exclude paths functionality

### LINKED ISSUE #2216: Add `--include-paths` and `--exclude-paths` flags for Docker scan

## Description

It appears that the Docker source does not have flags for including or excluding paths. This is a useful feature to include.

https://github.com/trufflesecurity/trufflehog/blob/main/main.go#L86C70-L87

### Preferred Solution

Add include/exclude flags for Docker, similar to how other sources work.

### Additional Context

https://trufflehog-community.slack.com/archives/CK5PWLESK/p1702565959415379

tannerjones4075:
### Description:
Add support for excluding paths in Docker source scanning:

- Add ExcludePaths field to Docker protobuf
- Implement path exclusion logic in docker.go
- Add comprehensive test coverage for exact and wildcard path matching
- Update engine to pass exclude paths configuration
- Add CLI support for --exclude-paths flag

The implementation supports:

- Exact path matching (e.g., /var/log/test)
- Wildcard path matching (e.g., /var/log/test/*)
- Multiple exclude paths

Tests ensure proper handling of:
- Exact path exclusions
- Wildcard exclusions
- Edge cases and similar paths

### References:
https://github.com/trufflesecurity/trufflehog/issues/2216?utm_source=chatgpt.com

### Checklist:


kashifkhan0771:
Thank you for opening this PR @tannerjones4075. Can you sign the CLA?

tannerjones4075:
> Thank you for opening this PR @tannerjones4075. Can you sign the CLA?

I have signed the CLA :)

pkg/sources/docker/docker_test.go
```diff
@@ -166,3 +167,89 @@ func isHistoryChunk(t *testing.T, chunk *sources.Chunk) bool {
...
+			pattern := strings.ReplaceAll(excludePath, "*", ".*")
+			pattern = "^" + pattern + "$"
+			isMatch, err := regexp.MatchString(pattern, tf.path)
+			assert.NoError(t, err)
+			if isMatch {
+				excluded = true
+				break
+			}
+		}
+		assert.Equal(t, tf.excluded, excluded, "Unexpected exclusion result for path: %s", tf.path)
```
rosecodym, line 253: If I'm reading this code right, it's basically just a copy of the logic being tested. That means it's testing that the code under test is implemented _exactly this way_, rather than more general concepts, like "is it excluding what is expected" or "does a file exclusion match cause a file to be skipped." Those latter two cases are what I'd actually like to see tested, which you could accomplish by (respectively):
* Creating a method on `Source` that accepts a file path and returns a boolean indicating whether to exclude it, and then just calling this function in a test
* Creating a new test that is a modified version of `TestDockerImageScanWithDigest` that configures an exclusion path and then asserts that a matching chunk was _not_ found

pkg/sources/docker/docker.go
```diff
@@ -349,6 +356,26 @@ func (s *Source) processChunk(ctx context.Context, info chunkProcessingInfo, chu
...
 
+	// Check if the file should be excluded
+	filePath := "/" + info.name
+	ctx.Logger().V(2).Info("checking file against exclude paths", "file", filePath, "exclude_paths", s.excludePaths)
+	for _, excludePath := range s.excludePaths {
+		// Convert wildcard pattern to regex
+		pattern := strings.ReplaceAll(excludePath, "*", ".*")
+		pattern = "^" + pattern + "$"
+
+		matched, err := regexp.MatchString(pattern, filePath)
```
rosecodym, line 367: Assembling the pattern anew and calling `regexp.MatchString` on each file is unnecessarily inefficient; you could instead use `Compile` once per pattern (in `Init`) and save some work. I don't know if it's a lot of work in the grand scheme of things, but I would find it easier to read and maintain, because compiling each regex once up-front is a more common way to use oft-matched regexes.

rosecodym (changes requested):
Thanks for this PR! It's a very clear and straightforward improvement. In particular, thank you for the detailed description and the tests. I've found a couple of small improvements I'd like to make before getting this in that I think will substantially reduce our future maintenance burden - please see my inline comments.

tannerjones4075:
I appreciate the feedback. I'll make some changes and add another commit.

pkg/sources/docker/docker_test.go
tannerjones4075: Thanks for the feedback. I'll incorporate the your suggestions. I agree the test case is more clear with passing in the boolean for the exclusion path.

pkg/sources/docker/docker.go
tannerjones4075: That is a good point! Compile is more efficient :)

pkg/sources/docker/docker.go
```diff
@@ -74,6 +77,22 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
+	if paths := s.conn.GetExcludePaths(); len(paths) > 0 {
+		s.excludePaths = paths
+		s.excludeRegexes = make([]*regexp.Regexp, len(paths))
+		for i, path := range paths {
+			// Convert wildcard pattern to regex
+			pattern := strings.ReplaceAll(path, "*", ".*")
+			pattern = "^" + pattern + "$"
+			regex, err := regexp.Compile(pattern)
+			if err != nil {
+				return fmt.Errorf("error compiling exclude path regex for %q: %w", path, err)
```
rosecodym, line 90: You might also consider adding the pattern being compiled to this error message; if the error is caused by something added during pattern generation, users are likely to be quite confused.

pkg/sources/docker/docker_test.go
```diff
@@ -166,3 +167,216 @@ func isHistoryChunk(t *testing.T, chunk *sources.Chunk) bool {
...
+		{"/var/log/test", true},      // Should be excluded (exact match)
+		{"/var/log/test2", false},    // Should not be excluded (different path)
+		{"/var/log/test/sub", false}, // Should not be excluded (subdirectory)
+		{"/var/log/other", false},    // Should not be excluded (different path)
+	}
+
+	for _, tf := range testFiles {
+		excluded := false
+		for _, excludePath := range s.excludePaths {
+			if tf.path == excludePath {
```
rosecodym, line 202: You're _only_ checking for exact matches here, which means that you're not actually exercising the match logic you've added to the source. This test is effectively adding a single item to a slice and then asserting that a few other items are not in the slice. You don't need a whole source implementation to test that! I think you can just remove this test entirely.

pkg/sources/docker/docker_test.go
rosecodym: Thank for the new tests! You might as well remove this one if those give you sufficient coverage. Tests are code, after all, and _ceteris paribus_ the less code in the codebase the better.

pkg/sources/docker/docker_test.go
```diff
@@ -166,3 +167,216 @@ func isHistoryChunk(t *testing.T, chunk *sources.Chunk) bool {
...
+				pattern := strings.ReplaceAll(path, "*", ".*")
+				pattern = "^" + pattern + "$"
+				isMatch, err := regexp.MatchString(pattern, tt.testPath)
+				assert.NoError(t, err)
+				if isMatch {
+					excluded = true
+					break
+				}
+			}
+			assert.Equal(t, tt.want, excluded)
```
rosecodym, line 330: I love the idea of this new test! But notice how `s` isn't involved in _any_ of this test code. That means that it's not actually testing the source at all! This is why I suggested creating a method on `Source` that actually checks for exclusion. If you do that, you can use it in the source and also test it here.

pkg/sources/docker/docker_test.go
```diff
@@ -166,3 +167,216 @@ func isHistoryChunk(t *testing.T, chunk *sources.Chunk) bool {
...
+		})
+	}
+}
+
+func TestDockerScanWithExclusions(t *testing.T) {
```
rosecodym, line 335: This test is great!

rosecodym (commented):
I love one of your new tests! But I've added some more commentary to the other one.

rosecodym:
One thing I didn't mention in earlier reviews - that I probably should have - is that the existing codebase has glob inclusion/exclusion implementation logic you could probably reuse instead of adding this new regex-based technique. I'm not going to insist on it (because I didn't bring it up earlier) but it's probably worth at least giving it a glance to see if it would be helpful.

pkg/sources/docker/docker_test.go
tannerjones4075: I appreciate the feedback and your patience. Learning how to write better tests.

tannerjones4075:
> One thing I didn't mention in earlier reviews - that I probably should have - is that the existing codebase has glob inclusion/exclusion implementation logic you could possibly reuse instead of adding this new regex-based technique. I'm not going to insist on it (because I didn't bring it up earlier) but it's probably worth at least giving it a glance to see if it would be helpful.

I replaced the regex with globfilter. Thanks for the suggestion and I think it reduces complexity and improved maintainability.

pkg/sources/docker/docker_test.go
```diff
@@ -166,3 +166,72 @@ func isHistoryChunk(t *testing.T, chunk *sources.Chunk) bool {
...
+				continue
+			}
+
+			metadata := chunk.SourceMetadata.GetDocker()
+			assert.NotNil(t, metadata)
+
+			// Check if we found a chunk with the excluded path
+			if metadata.File == "/aws" {
+				foundExcludedPath = true
+			}
```
rosecodym, line 226: Does that image actually contain a chunk that would have this path?

pkg/sources/docker/docker.go
```diff
@@ -30,12 +31,14 @@ import (
...
-	verify      bool
-	concurrency int
-	conn        sourcespb.Docker
+	name         string
+	sourceId     sources.SourceID
+	jobId        sources.JobID
+	verify       bool
+	concurrency  int
+	conn         sourcespb.Docker
+	excludePaths []string
```
rosecodym: Is this field necessary? It looks like you save the compiled glob in `Init` immediately and don't actually need to store these paths here.

rosecodym (commented):
This looks super clean, nice work! I have a couple of small questions, and I see you've got a conflicting sources.pb.go file. Once you've resolved the former, you can resolve the latter by merging from `main` and then running `make protos`, but if you're running into trouble with it, I'm happy to do that part for you, since it was my extended review that caused the drift in the first place :)

pkg/sources/docker/docker.go
tannerjones4075: That is a good point. I will remove the excludePath from the struct.

pkg/sources/docker/docker_test.go
tannerjones4075: Yes, after verifying /aws path does exist:
image

rosecodym:
Let me know if you want me to take this over and do the protobuf merge!

tannerjones4075:
> Let me know if you want me to take this over and do the protobuf merge!

Yes, the help would be great to get it merged.

rosecodym (approved):
Thanks so much for your work on this and for accommodating my requests! This is a great improvement to the tool.

PR #4059, opened 2025-04-16
Improved Besttime detector regex pattern

kashifkhan0771:
### Description:
BestTime private API keys always begin with the `pri_` prefix and consist exclusively of lowercase letters from a to f and digits from 0 to 9. The total length of the key, including the `pri_` prefix, is exactly 36 characters.
I generated 5-6 keys and all satisfy this new pattern.
API Docs: https://documentation.besttime.app/#introduction

### Tests:
Screenshot from 2025-04-16 11-45-12

### Checklist:

### DIFF
--- a/pkg/detectors/besttime/besttime.go
+++ b/pkg/detectors/besttime/besttime.go
@@ -22,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"besttime"}) + `(pri_[a-f0-9]{32})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"besttime"}) + `\b(pri_[a-f0-9]{32})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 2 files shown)


nabeelalam:
This looks good to me. One question, was this the pattern of their keys initially as well, or was this updated? Asking in case BestTime still supports secrets in the old pattern.

pkg/detectors/besttime/besttime.go
```diff
@@ -22,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"besttime"}) + `\b([0-9A-Za-z_]{36})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"besttime"}) + `(pri_[a-f0-9]{32})\b`)
```
nabeelalam, line 25: Can we add a boundary character at the front of the regex?

nabeelalam (approved):
One small comment otherwise LGTM

kashifkhan0771:
> This looks good to me. One question, was this the pattern of their keys initially as well, or was this updated? Asking in case BestTime still supports secrets in the old pattern.

I think this was the pattern initially because the key we had for our integration test matches this pattern. The old pattern was matching the keys but was too broad.

abmussani (approved):

PR #4062, opened 2025-04-16
[Feat] Support Dockerhub OATs

harmonherring-pro:
### Description:
Dockerhub has recently rolled out [Organization Access Tokens](https://docs.docker.com/security/for-admins/access-tokens/). These work similarly to PATs, but have a slightly different format (32 characters after the prefix instead of 27) and authenticate using the [/v2/auth/token](https://docs.docker.com/reference/api/hub/latest/#tag/authentication-api/operation/AuthCreateAccessToken) endpoint instead of the (now deprecated) [/v2/users/login](https://docs.docker.com/reference/api/hub/latest/#tag/authentication-api/operation/PostUsersLogin) endpoint

### Checklist:


pkg/detectors/dockerhub/v2/dockerhub.go
```diff
@@ -32,14 +32,14 @@ var (
 	usernamePat = regexp.MustCompile(`(?im)(?:user|usr|-u|id)\S{0,40}?[:=\s]{1,3}[ '"=]?([a-zA-Z0-9]{4,40})\b`)
 	emailPat    = regexp.MustCompile(common.EmailPattern)
 
-	// Can use password or personal access token (PAT) for login, but this scanner will only check for PATs.
-	accessTokenPat = regexp.MustCompile(`\b(dckr_pat_[a-zA-Z0-9_-]{27})(?:[^a-zA-Z0-9_-]|\z)`)
+	// Can use password or personal/organization access token (PAT/OAT) for login, but this scanner will only check for PATs and OATs.
+	accessTokenPat = regexp.MustCompile(`\b(dckr_(pat|oat)_[a-zA-Z0-9_-]{27,32})(?:[^a-zA-Z0-9_-]|\z)`)
```
abmussani, line 36: Can we restrict the length to either 27 or 32 to eliminate false positives? The modified regex also accepts a length of 28, which we already know is an incorrect token.

something like:
`(dckr_(pat|oat)_(?:[a-zA-Z0-9_-]{27}|[a-zA-Z0-9_-]{32}))(?:[^a-zA-Z0-9_-]|\z)`

harmonherring-pro: @abmussani done in [b9aaa01](https://github.com/trufflesecurity/trufflehog/pull/4062/commits/b9aaa018024fa48396f1e9e237576f999510a226), thanks!

nabeelalam: Hi @harmonherring-pro. I'm wondering if Docker PATs can be of length `32` as well, or is this being introduced to facilitate detection of OATs?
If it's the latter, the updated regex will start matching tokens with the prefix `dckr_pat_` followed by 32 characters, which it wasn't before. Same with `dckr_oat_` tokens followed by only 27 characters.

abmussani (approved):

nabeelalam (approved):
Thanks @harmonherring-pro !

kashifkhan0771 (approved):
LGTM!

dillyphilly3131 (commented):
Ggg

PR #4068, opened 2025-04-18
Removing hardcoded options in GHA docker run and providing output configuration

kellydunn:
### Description:
When using the `trufflesecurity/trufflehog` Github Action, there's currently no way to configure `--fail`, in the Github action itself, as it is hardcoded in `action.yml` [here](https://github.com/trufflesecurity/trufflehog/blob/e68e67137c7dc0b2e769c1e011ad5515a17cecb2/action.yml#L96).  

Additionally, it's helpful to be able to take the very useful output of trufflehog and process it in our own actions so we can aggregate the data to our observability tools.  To enable this, this PR also provides an `output` block that enables  developers to reference Trufflehog output in their actions via `${{ steps.trufflehog.outputs.results }}`

This way, developers can run the github action like so:
```yml
      - name: Run Trufflehog
        id: trufflehog
        uses: trufflesecurity/trufflehog@main
        with:
          base: <base>
          head: <head>
          no_fail: "true"
          extra_args: "--json --results=verified,unknown"
          
      - name: Use Output
        id: use-output
        run: |
          echo ${{ steps.trufflehog.outputs.results }} | jq -s | ./bin/script
```

**Update**:  This PR now also includes a more explicit mounting of `${{ github.workspace }}` to the docker container, as before the relative path being mounted to `/tmp` wouldn't pick up any changes when providing commit ranges to the scan.

### Checklist:


kashifkhan0771:
Hi @kellydunn Thanks for submitting the PR. Can you please sign the CLA.

action.yml
```diff
@@ -86,14 +91,14 @@ runs:
...
       ##########################################
-      docker run --rm -v .:/tmp -w /tmp \
+      results=$(docker run --rm -v ${{ github.workspace }}:/tmp -w /tmp \
       ghcr.io/trufflesecurity/trufflehog:${VERSION} \
       git file:///tmp/ \
       --since-commit \
       ${BASE:-''} \
       --branch \
       ${HEAD:-''} \
-      --fail \
```
nabeelalam, line 96: Hey @kellydunn, thanks for this! This definitely looks like a really useful feature for workflows integrating this action.

One question, particularly about the removal of the `--fail` argument. Does this change alter the expected behavior of the workflow, i.e. fail and exit with a `183` code if a secret is detected ([ref](https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#12-scan-in-ci))? If so, this may impact any setups that rely on this failure behavior as a guardrail.

Would it make sense to make this behavior configurable, i.e. fail by default, unless explicitly disabled?

kellydunn: Hello!  And fair question!  I could see how this might impact workflows that rely on default fail behavior; explicitly opt-ing out of `--fail` seems reasonable.  I'll take a swing at providing a GHA input control similar to `base` and `head` parameters in the workflow.

kellydunn: @nabeelalam, just pushed up https://github.com/trufflesecurity/trufflehog/pull/4068/commits/6d2c6cc7fd919d68a21e87f69f9e869561ce1435.  We've tested this with our setup, and we can confirm that by default this will preserve the behavior of exiting with `183` unless a user explicitly provides a `no_fail` value in their workflow action.  Lmk what you think.

kellydunn: @nabeelalam Checking in here, anything else you'd like to see?  Mind giving the updates a review?

nabeelalam: Hey @kellydunn missed your updates on this. I've looked it over and this looks good to me. Running github workflows on this PR

PR #4081, opened 2025-04-22
Add --no-color option to force disabling colorized output

### LINKED ISSUE #3300: Switch to disable or change colour profiles

## Description
At the moment if you use trufflehog through Docker it will always detect a tty so shell escapes will always be present even if output is redirected. The colour scheme chosen is always designed for a dark background, if you use a light mode terminal (e.g. if you have a number of eye problems, such as astigmatism, so can't use dark mode terminals) then the output cannot be seen (see example below)

{51A8E90B-3DD8-4256-B95B-79E12DCAB7EC}

### Preferred Solution
Add a switch to output to a text file.
Add a switch to change the colour profile, at the least to disable it, or to allow a light mode colour scheme.
Support both spellings of colour (colour / color) in the switch so it's international.

TheToddLuci0:
### Description:
Add a flag to forcibly disable colored output when using the plain / default output. 

Yes, I'm aware that the `color` library in use has environment variables that can do this, but the only reason I know that is looking into the library in question while making this. This just makes it easier.

Fixes #3300 

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,6 +16,7 @@ import (
 	"syscall"
 
 	"github.com/alecthomas/kingpin/v2"
+	"github.com/fatih/color"
 	"github.com/felixge/fgprof"
 	"github.com/go-logr/logr"
 	"github.com/jpillora/overseer"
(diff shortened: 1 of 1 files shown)


nabeelalam:
Hey @TheToddLuci0! This looks great. Thanks for working on this.
Can we add an alternate spelling `--no-colour` as well, as suggested in the linked issue?

TheToddLuci0:
@nabeelalam  Went ahead and added it. I made it a hidden alias (just to keep things clean), but it'll work no matter which spelling your give it.

PR #4082, opened 2025-04-23
Fastly Analyzer

kashifkhan0771:
### Description:
This PR add a new analyzer for Fastly Detector

### Tests:
Screenshot from 2025-04-23 12-36-54

### Checklist:


pkg/analyzer/analyzers/fastly/requests.go
```diff
@@ -0,0 +1,744 @@
...
+
+	if len(aggregatedErrs) > 0 {
+		return errors.Join(aggregatedErrs...)
+	}
+
+	return nil
+}
+
+// captureTokenInfo calls `/tokens/self` API and capture the token information in secretInfo
+func captureTokenInfo(client *http.Client, key string, secretInfo *SecretInfo) error {
```
nabeelalam, line 178: Hey @kashifkhan0771, I noticed that in the `captureX(...)` resource capturing functions there's a lot of repeated code.
Is it possible to create a single `captureResource(...)` function that takes the resource `type` as input?

pkg/analyzer/analyzers/fastly/result_output.json
```diff
@@ -0,0 +1,311 @@
+{
+    "AnalyzerType": 6,
```
nabeelalam, line 2: The `AnalyzerType` value might be `34`?

pkg/analyzer/analyzers/fastly/requests.go
kashifkhan0771: **Reasons for keeping the functions separate instead of merging into a generic one:**

- While it's possible to use [Go Generics](https://go.dev/doc/tutorial/generics), we're currently not using generics elsewhere in the codebase. Introducing them here might add unnecessary complexity, especially for newer engineers.
- Some API responses require us to capture `extraData`, which would complicate a generic function and reduce clarity.
- Certain resources involve custom logic to generate unique IDs or names. Handling such cases within a generic structure would lead to complex logic.
- In my opinion, keeping each API call in its own function improves readability and maintainability, even if it means writing a bit more code.

pkg/analyzer/analyzers/fastly/result_output.json
kashifkhan0771: Not sure about this. The test case I wrote returns `6`.

nabeelalam: That could be because in the `secretInfoToAnalyzerResult(...)` function it's currently passing in the `AnalyzerTypeElevenLabs`
```
result := analyzers.AnalyzerResult{
	AnalyzerType: analyzers.AnalyzerTypeElevenLabs, // should be AnalyzerTypeFastly
	Metadata:     map[string]any{},
	Bindings:     make([]analyzers.Binding, 0),
}
```

kashifkhan0771: Oops - the real culprit is `CTRL + C` and `CTRL + V` than 😆 Let me fix it.

kashifkhan0771: Done ✅

nabeelalam (approved):
Looks good, thanks @kashifkhan0771

PR #4084, opened 2025-04-23
[Fix] Added Prefix In Dockerhub Detector Regex

nabeelalam:
### Description:
The Dockerhub detector was missing a prefix check in its regex, this PR adds prefix checks in both Dockerhub detectors

### Checklist:


pkg/detectors/dockerhub/v2/dockerhub.go
```diff
@@ -29,17 +29,17 @@ var _ detectors.Versioner = (*Scanner)(nil)
...
 var (
 	// Can use email or username for login.
-	usernamePat = regexp.MustCompile(`(?im)(?:user|usr|-u|id)\S{0,40}?[:=\s]{1,3}[ '"=]?([a-zA-Z0-9]{4,40})\b`)
-	emailPat    = regexp.MustCompile(common.EmailPattern)
+	usernamePat = regexp.MustCompile(detectors.PrefixRegex([]string{"docker"}) + `(?im)(?:user|usr|-u|id)\S{0,40}?[:=\s]{1,3}[ '"=]?([a-zA-Z0-9]{4,40})\b`)
+	emailPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"docker"}) + common.EmailPattern)
 
 	// Can use password or personal/organization access token (PAT/OAT) for login, but this scanner will only check for PATs and OATs.
-	accessTokenPat = regexp.MustCompile(`\b(dckr_pat_[a-zA-Z0-9_-]{27}|dckr_oat_[a-zA-Z0-9_-]{32})(?:[^a-zA-Z0-9_-]|\z)`)
+	accessTokenPat = regexp.MustCompile(detectors.PrefixRegex([]string{"docker"}) + `\b(dckr_pat_[a-zA-Z0-9_-]{27}|dckr_oat_[a-zA-Z0-9_-]{32})(?:[^a-zA-Z0-9_-]|\z)`)
```
rgmz, line 36: What is the benefit of requiring am arbitrary prefix for a pattern that already has a high-confidence prefix? This will cause the detector to mise a significant amount of matches.

zricethezav: @nabeelalam I agree with @rgmz here. `dckr_pat` is unique enough where we do not need the `docker` prefix

nabeelalam: Gotcha. That makes sense, I'll remove this.

rgmz: Tbh I'm not sure either of the remaining additions are useful either. At least for v2, the email and user regexes are already fairly specific, and they won't be searched unless the chunk contains a high-confidence keyword.

40 characters is a really small range, even conmon instances like `docker login --username=...` can easily exceed 40 characters.

rgmz (commented):
After some testing, any of these changes would be a regression. The current access token patterns are restrictive enough.
```yaml
----
steps:
  - name: Login to DockerHub
    uses: docker/login-action@v2
    with:
      password: dckr_pat_cq7wQZcv9xZkVlxMhDTcTV00CDo
      username: mobythewhale
```

rgmz:
> @rgmz do you think other prefix like username or email would be helpful ? I am trying to understand whether, if we remove the prefix, detector will pick the unlikely linked secrets and try to verify them.

What's the impact of that versus the detector missing valid results? Personally, I don't think it's helpful:

1. Both v1 and v2 have an existing prefix, so the chunk is guaranteed to contain either `docker` or `dckr_pat_` (high confidence) respectively.
2. The match context for most detectors is [~512 bytes on either side of the key word](https://github.com/trufflesecurity/trufflehog/blob/a451507f0d4642cc7a2f5b829dbd33d534a78b13/pkg/engine/ahocorasick/ahocorasickcore.go#L155) (I believe), meaning only data in close proximity to the keyword can be picked up by the username or email patterns.
3.  [`PrefixRegex`](https://github.com/trufflesecurity/trufflehog/blob/a5e00324fcf6a826fe7af37f475ee1815e0ede61/pkg/detectors/detectors.go#L230) is excessively limited as shown above. The `usernamePat` arguably has a prefix, and email is (or should be) a high-confidence pattern.

None of these make the detectors immune to false positives, however, I'd argue it'd be more fruitful to whitelist common sources of false-positives.

nabeelalam:
> I'd argue it'd be more fruitful to whitelist common sources of false-positives.

@rgmz I agree on this, that would be a net positive compared to the alternative. I'm going to update the v2 detector.

zricethezav (approved):

dillyphilly3131 (commented):
Nice

PR #4086, opened 2025-04-23
Postman increase http client timeout

casey-tran:
### Description:
The Postman source can encounter issues with reading response bodies due to context cancellation or HTTP client timeout while trying to read the body. This PR attempts to remedy this issue by:

1. Deferring the response bodies closing.
2. In the process of 1, I consolidated the read and close steps into the function that makes Postman API calls.
3. Assigning a timeout value to the client returned by `RetryableHTTPClientTimeout`.

An odd bug was found in the process of exploring `RetryableHTTPClientTimeout`. The client returned by `RetryableHTTPClientTimeout` with `return httpClient.StandardClient()` did not keep the timeout attribute assigned to the receiver `httpClient` earlier in the function. The client returned by the call to `RetryableHTTPClientTimeout` would have a `Timeout` value of `0`, which then begs the question of how a http client could ever time out with a `Timeout` value of `0` in the first place.

### Checklist:

### DIFF
--- a/pkg/common/http.go
+++ b/pkg/common/http.go
@@ -47,7 +47,7 @@ oyi3B43njTOQ5yOf+1CceWxG1bQVs5ZufpsMljq4Ui0/1lvh+wjChP4kqKOJ2qxq
 4RgqsahDYVvTH9w7jXbyLeiNdd8XM2w9U/t7y0Ff/9yi0GE44Za4rF2LN9d11TPA
 mRGunUHBcnWEvgJBQl9nJEiU0Zsnvgc/ubhPgXRR4Xq37Z0j4r7g1SgEEzwxA57d
 emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
------END CERTIFICATE-----	
+-----END CERTIFICATE-----
 `,
 	// 	CN = ISRG Root X2
 	// TODO: Expires September 17, 2040 at 9:00:00 AM Pacific Daylight Time
@@ -199,7 +199,10 @@ func RetryableHTTPClientTimeout(timeOutSeconds int64, opts ...ClientOption) *htt
 	for _, opt := range opts {
 		opt(httpClient)
 	}
-	return httpClient.StandardClient()
+
+	standardClient := httpClient.StandardClient()
+	standardClient.Timeout = httpClient.HTTPClient.Timeout
+	return standardClient
 }
 
 const DefaultResponseTimeout = 5 * time.Second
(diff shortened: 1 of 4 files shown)


pkg/common/http.go
```diff
@@ -199,7 +200,11 @@ func RetryableHTTPClientTimeout(timeOutSeconds int64, opts ...ClientOption) *htt
 	for _, opt := range opts {
 		opt(httpClient)
 	}
-	return httpClient.StandardClient()
+
+	standardClient := httpClient.StandardClient()
+	standardClient.Timeout = httpClient.HTTPClient.Timeout
+	standardClient.Transport = httpClient.HTTPClient.Transport
```
rosecodym, line 206: The retryable HTTP client's transport is where the retry magic happens, so overwriting it eliminates retries!

rosecodym (dismissed):
This PR inadvertently makes retryable HTTP clients no longer retry; I think it will be easier to explain synchronously.

pkg/sources/postman/postman_client.go
```diff
@@ -256,13 +257,19 @@ func (c *Client) getPostmanReq(ctx context.Context, url string, headers map[stri
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		ctx.Logger().Error(err, "could not read postman response body")
```
camgunz, line 264: I think it would be OK to pass this error on up.

camgunz (changes requested):
Just had one thing about a logged error becoming a real error

pkg/sources/postman/postman_client.go
```diff
@@ -246,7 +246,8 @@ func checkResponseStatus(r *http.Response) error {
 	return fmt.Errorf("postman Request failed with status code: %d", r.StatusCode)
 }
 
-func (c *Client) getPostmanReq(ctx context.Context, url string, headers map[string]string) (*http.Response, error) {
+// getPostmanResponseBodyBytes makes a request to the Postman API and returns the response body as bytes.
+func (c *Client) getPostmanResponseBodyBytes(ctx context.Context, url string, headers map[string]string) ([]byte, error) {
```
rosecodym, line 250: What motivated this change? All else equal, delaying reads as long as possible is usually best (because it gives callers the most flexibility), but was there a specific reason to read eagerly here that's not jumping out at me?

rosecodym, line 250: What motivated this change? All else equal, delaying reads as long as possible is usually best (because it gives callers the most flexibility), but was there a specific reason to read eagerly here that's not jumping out at me?

pkg/sources/postman/postman_client.go
casey-tran: @martinlocklear brought up that some of the Postman errors could be pass through rather than blocking. I guess that can wait for his PR for a discussion on what should/shouldn't be skipped.

casey-tran: Changed this error back to blocking.

camgunz (approved):

PR #4088, opened 2025-04-23
feat: allow input source from pipe

strazzere:
While performing a bit of experimentation, I found it was easier to allow data to flow from other segments of my machine via piping it in, rather than flushing it to disk. This can speed up the process when pre-processing large amounts of data. This allow allowed me to keep disk space low and just utilize a large ram machine.

### Description:
Allow `stdin` an input source. This was helpful for dealing with large datasets and avoiding the disk (and keeping memory to a decent size). This should let other folks leverage the tool a bit easier when piping data from other tools potentially.

My specific instance was processing a rather large file in memory from the web and just piping to trufflehog. This was automated on an aws machine which a small disk but lots of CPU/memory.

### Checklist:


zricethezav:
@strazzere good lookin' PR. Would you mind adding an example command in the README?

strazzere:
Documentation fix added

strazzere:
It doesn't really matter to me, pipe felt better since it strongly infers this isn't inherently just readable data that is being processed. Though that might literally just be how I think of it. I can change it over to `stdin`.

pkg/sources/pipe/pipe.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
+	stdin := os.Stdin
+	chunkSkel := &sources.Chunk{
+		SourceType: s.Type(),
+		SourceName: s.name,
+		SourceID:   s.SourceID(),
+		JobID:      s.JobID(),
+		SourceMetadata: &source_metadatapb.MetaData{
+			Data: &source_metadatapb.MetaData_Filesystem{
```
rosecodym, line 61: If this is a new source, it should use a new type of metadata. (The new type won't actually contain any information, but empty metadata is fine.)

pkg/sources/pipe/pipe.go
```diff
@@ -0,0 +1,93 @@
...
+func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) error {
+	unit := sources.CommonSourceUnit{ID: "<stdin>"}
+	return reporter.UnitOk(ctx, unit)
+}
+
+func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporter sources.ChunkReporter) error {
+	ch := make(chan *sources.Chunk)
+	go func() {
+		defer close(ch)
+		_ = s.Chunks(ctx, ch)
```
rosecodym, line 83: This is an interesting solution! Usually we have `Chunks` invoke `ChunkUnit`, but you've done it the other way around. I can't see anything "wrong" with it, but it did involve more boilerplate (to fiddle with the channel) than the other way does, especially if you use the pre-canned `ChanReporter` type that exists. Is that something you considered?

strazzere: Ah! I think I see what you're saying.

This was mostly based of looking at `filesystem.go` and trying to figure out the best way to convert a "fileless" stream in. It just "felt right" while doing it and had worked - plus it let me get onto the primary aspect of the research I was doing vs try to understand the deeper bits of everything.

If required/wanted, I can try to rework this in a different pattern when I have the time.

rosecodym: Oh, that's not necessary, I just thought I'd mention it in case you found it helpful.

rosecodym (changes requested):
This looks pretty good! Once you've changed all the names over from "pipe" to "stdin" or "Standard Input" we'd be happy to get it in. (Please change them in both the code and the user-facing text.)

pkg/sources/pipe/pipe.go
strazzere: This should be resolved now.

strazzere:
Requested changes should all be done now, cheers.

rosecodym (approved):
Very cool, thanks!

PR #4094, opened 2025-04-25
Don't die on collection request failure to Postman API

martinlocklear:
Right now when the Postman source is scanning the postman API, if it gets *any* error back from the API it stops the whole scan.  We should probably just log and continue if this happens, so this makes that change for requests to the *collection* endpoint

This issue popped up when investigating a logged problem with making requests to the Postman API and (likely) sending either a bad collection/workspace id or (also possible) sending a collection or workspace id that we don't have permission to explore to the Postman API.  Still worth looking more into that, but I think that making these errors less visible is worth the tradeoff for still getting _something_ out of a scan.

[Related PR for workspaces](https://github.com/trufflesecurity/trufflehog/pull/4095)

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman.go
+++ b/pkg/sources/postman/postman.go
@@ -202,7 +202,10 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 
 		collection, err := s.client.GetCollection(ctx, collectionID)
 		if err != nil {
-			return fmt.Errorf("error getting collection %s: %w", collectionID, err)
+			// Log and move on, because sometimes the Postman API seems to give us collection IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().Error(err, "error getting collection %s", collectionID)
+			continue
 		}
 		s.SetProgressOngoing(fmt.Sprintf("Scanning collection %s", collectionID), "")
 		s.scanCollection(ctx, chunksChan, Metadata{}, collection)
@@ -302,7 +305,10 @@ func (s *Source) scanWorkspace(ctx context.Context, chunksChan chan *sources.Chu
 		}
 		collection, err := s.client.GetCollection(ctx, collectionID.UUID)
 		if err != nil {
-			return err
+			// Log and move on, because sometimes the Postman API seems to give us collection IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().Error(err, "error getting collection %s", collectionID)
+			continue
 		}
 		s.scanCollection(ctx, chunksChan, metadata, collection)
 	}
--- a/pkg/sources/postman/postman_test.go
+++ b/pkg/sources/postman/postman_test.go
@@ -335,6 +335,141 @@ func TestSource_ScanGeneralRateLimit(t *testing.T) {
 	}
 }
 
+func TestSource_BadPostmanCollectionApiResponseDoesntEndScan(t *testing.T) {
+	// The goal here is to make sure that, if we get a bad ID for a collection (or some other request issue) and the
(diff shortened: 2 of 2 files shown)


pkg/sources/postman/postman.go
```diff
@@ -199,7 +199,10 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 
 		collection, err := s.client.GetCollection(ctx, collectionID)
 		if err != nil {
-			return fmt.Errorf("error getting collection %s: %w", collectionID, err)
+			// Log and move on, because sometimes the Postman API seems to give us collection IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().V(2).Error(err, "error getting collection %s", collectionID)
```
rosecodym, line 204: We do not currently log errors above level 0 anywhere. The idea has come up in conversation a few times, and might be something we want to change, but right now this is a deviation from a very solid standard.

martinlocklear: I think I remember reading that somewhere - thanks for pointing that out.   I did a little poking around and it looks like the GCP analyzer has [recently](https://github.com/trufflesecurity/thog/commit/c596ac5e7b5f1b7023a9a20912f1320a54399e1c) (past couple of weeks) started doing it, for a contextually similar situation I think.

https://github.com/trufflesecurity/thog/blob/c596ac5e7b5f1b7023a9a20912f1320a54399e1c/scanner/pkg/analyzers/gcp/analysis.go#L283

martinlocklear: Not sure if that means we want to continue the pattern or not - I'm not familiar with the pros/cons of the situation

pkg/sources/postman/postman_test.go
```diff
@@ -334,6 +335,138 @@ func TestSource_ScanGeneralRateLimit(t *testing.T) {
...
+	// check that the third one was properly requested
+	defer gock.Off()
+
+	// IDs we'll use later
+	userId := "12345678"
+	id1WorkspaceGood := "1f0df51a-8658-4ee8-a2a1-d2567dfa09a9"
+	id1CollectionGood := "dac5eac9-148d-a32e-b76b-3edee9da28f7"
+	id2CollectionBad := "12ece9e1-2abf-4edc-8e34-de66e74114d2"
+	id3CollectionGood := "f695cab7-6878-eb55-7943-ad88e1ccfd65"
+	uid1CollectionGood := fmt.Sprintf("%s-%s", userId, id1CollectionGood)
```
casey-tran, line 351: Since the collection with id 1 is supposed to have a malformed response body, do we want to name it `id1CollectionBad` and `id1CollectionBad`?

casey-tran (commented):
Thanks for the non-fatal error implementation!

pkg/sources/postman/postman_test.go
martinlocklear: Absolutely

pkg/sources/postman/postman.go
rgmz: Isn't `.V(...)` a no-op with `.Error(...)`? I swear it always logs.

martinlocklear: Looks like [you're right](https://github.com/go-logr/logr/blob/master/logr.go#L287C17-L288C20) @rgmz 

I"m wondering now if it makes more sense to log these `Info` rather than error.

martinlocklear: Just stripped off the verbosity

martinlocklear: Created a [quick PR](https://github.com/trufflesecurity/trufflehog/pull/4097) to remove the extraneous call to `V` in a couple of other places

casey-tran (approved):

PR #4095, opened 2025-04-25
Don't die on workspace request failure to Postman API

martinlocklear:
Right now when the Postman source is scanning the postman API, if it gets *any* error back from the API it stops the whole scan.  We should probably just log and continue if this happens, so this makes that change for requests to the *workspace* endpoint

This issue popped up when investigating a logged problem with making requests to the Postman API and (likely) sending either a bad collection/workspace id or (also possible) sending a collection or workspace id that we don't have permission to explore to the Postman API.  Still worth looking more into that, but I think that making these errors less visible is worth the tradeoff for still getting _something_ out of a scan.

[Related PR for collections](https://github.com/trufflesecurity/trufflehog/pull/4094)

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman.go
+++ b/pkg/sources/postman/postman.go
@@ -182,7 +182,10 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 	for _, workspaceID := range s.conn.Workspaces {
 		w, err := s.client.GetWorkspace(ctx, workspaceID)
 		if err != nil {
-			return fmt.Errorf("error getting workspace %s: %w", workspaceID, err)
+			// Log and move on, because sometimes the Postman API seems to give us workspace IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().Error(err, "error getting workspace %s", workspaceID)
+			continue
 		}
 		s.SetProgressOngoing(fmt.Sprintf("Scanning workspace %s", workspaceID), "")
 		ctx.Logger().V(2).Info("scanning workspace from workspaces given", "workspace", workspaceID)
--- a/pkg/sources/postman/postman_client.go
+++ b/pkg/sources/postman/postman_client.go
@@ -294,7 +294,10 @@ func (c *Client) EnumerateWorkspaces(ctx context.Context) ([]Workspace, error) {
 	for i, workspace := range workspacesObj.Workspaces {
 		tempWorkspace, err := c.GetWorkspace(ctx, workspace.ID)
 		if err != nil {
-			return nil, fmt.Errorf("could not get workspace %q (%s) during enumeration: %w", workspace.Name, workspace.ID, err)
+			// Log and move on, because sometimes the Postman API seems to give us workspace IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().Error(err, "could not get workspace %q (%s) during enumeration", workspace.Name, workspace.ID)
+			continue
 		}
 		workspacesObj.Workspaces[i] = tempWorkspace
 
--- a/pkg/sources/postman/postman_test.go
+++ b/pkg/sources/postman/postman_test.go
@@ -2,6 +2,7 @@ package postman
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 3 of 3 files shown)


pkg/sources/postman/postman_test.go
```diff
@@ -334,6 +335,110 @@ func TestSource_ScanGeneralRateLimit(t *testing.T) {
...
 
+func TestSource_BadPostmanWorkspaceApiResponseDoesntEndScan(t *testing.T) {
+	// The goal here is to make sure that, if we get a bad ID (or other issue) for a workspace and the Postman API
+	// gives us a non 200 responses, it doesn't stop the whole scan.  To do that we're going to  have it get a set
+	// of 3 workspaces from /workspaces/ and then mock all but the last as a bad request.  Then we'll check that the
+	// third one was properly requested.
+	defer gock.Off()
+
+	// We'll use the IDs later in a couple of places
+	id1WorkspaceGood := "1f0df51a-8658-4ee8-a2a1-d2567dfa09a9"
```
casey-tran, line 346: If the first two of three workspace IDs used are supposed to trigger an error, do you want to instead call this variable `id1WorkspaceBad`?

pkg/sources/postman/postman_test.go
```diff
@@ -334,6 +335,110 @@ func TestSource_ScanGeneralRateLimit(t *testing.T) {
...
+					"id":         id3WorkspaceGood,
+					"name":       "Team Workspace",
+					"createdBy":  "12345678",
+					"type":       "team",
+					"visibility": "team",
+				},
+			},
+		})
+
+	// Make a call for the first workspace respond with a malformed request
```
casey-tran, line 381: Should this say `with a malformed response body`?

pkg/sources/postman/postman.go
```diff
@@ -182,7 +182,10 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 	for _, workspaceID := range s.conn.Workspaces {
 		w, err := s.client.GetWorkspace(ctx, workspaceID)
 		if err != nil {
-			return fmt.Errorf("error getting workspace %s: %w", workspaceID, err)
+			// Log and move on, because sometimes the Postman API seems to give us workspace IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().V(2).Error(err, "error getting workspace %s", workspaceID)
```
casey-tran, line 187: Same vein as @rosecodym 's comment in #4094 about if we want to log errors at different levels or not.

pkg/sources/postman/postman_client.go
```diff
@@ -294,7 +294,10 @@ func (c *Client) EnumerateWorkspaces(ctx context.Context) ([]Workspace, error) {
 	for i, workspace := range workspacesObj.Workspaces {
 		tempWorkspace, err := c.GetWorkspace(ctx, workspace.ID)
 		if err != nil {
-			return nil, fmt.Errorf("could not get workspace %q (%s) during enumeration: %w", workspace.Name, workspace.ID, err)
+			// Log and move on, because sometimes the Postman API seems to give us workspace IDs
+			// that we don't have access to, so we don't want to kill the scan because of it.
+			ctx.Logger().V(2).Error(err, "could not get workspace %q (%s) during enumeration: %w", workspace.Name, workspace.ID)
```
casey-tran, line 299: Same comment about logging errors at different levels.

casey-tran:
Thanks for thinking of this for #4094 and this PR.

pkg/sources/postman/postman_test.go
martinlocklear: Oh - good catch.  That name is from an earlier variant of the test I enhanced with the malformed request.

pkg/sources/postman/postman.go
martinlocklear: [Discussion](https://github.com/trufflesecurity/trufflehog/pull/4094#discussion_r2060620115)

pkg/sources/postman/postman_client.go
martinlocklear: [Discussion](https://github.com/trufflesecurity/trufflehog/pull/4094#discussion_r2060620115)

casey-tran (approved):

kashifkhan0771 (approved):

PR #4100, opened 2025-04-26
[GitHub] split enterprise-specific methods into a separate file.

rgmz:
### Description:
The `github.go` file is large and unwieldy. Per https://github.com/trufflesecurity/trufflehog/pull/3298#issuecomment-2510010947, `Chunks` is only used by the enterprise code and so I've marked it as deprecated and moved it into a separate file.

\+ @mcastorina 

### Checklist:


rosecodym (changes requested):
Unfortunately, the migration referred to by https://github.com/trufflesecurity/trufflehog/pull/3298#issuecomment-2477628123 is more complex than just lighting a new code path up. `Chunks` is deprecated _except_ for target scanning - full deprecation will require moving that functionality elsewhere, and we don't yet know where that will be. It might not even remain enterprise-only.

This PR splits the source into two pieces, only one of which is relevant to OSS contributors. But because the dividing line is merely a historical artifact - it doesn't represent any underlying structure - it's extremely porous, and I predict that it won't take long for something to end up on the "wrong" side of it, at which point our big ball of mud will be even less comprehensible for being split into two somewhat arbitrary files.

If you want to stop having to look at code that's not relevant to you, I think a better approach would be to actually delete it: figure out a way to get rid of `scan`. Quarantining it and waiting for it to die of natural causes (as you've done in this PR) isn't going to be sufficient; if we want it gone, we'll need to resort to murder.

PR #4102, opened 2025-04-28
Refactor Netlify Detector

amanfcp:
### Description:
This PR refactors the Netlify Detector based on new detector implementations and also adds configuration for integration testing through GCP

### Checklist:


pkg/detectors/netlify/v2/netlify_v2.go
```diff
@@ -21,9 +22,10 @@ var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.Versioner = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
-
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"netlify"}) + `\b(nfp_[a-zA-Z0-9_]{36})\b`)
+	client           = common.SaneHttpClient()
+	rotationGuideUrl = "https://howtorotate.com/docs/tutorials/netlify/"
```
ahrav, line 26: suggestion: I like the idea of extracting the string literals. I think we could use `const` for these two new variables, since they don't need to change. Similar comment for the v1 detector.

pkg/detectors/netlify/v2/netlify_v2.go
```diff
@@ -73,6 +69,30 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, map[string]string, error) {
```
ahrav, line 72: question: Do you foresee us ever returning a non-nil `extraData` value in the future? If not, we could simplify the function signature to just return `(bool, error)`.

This would also simplify the call site at `L60`, removing the need for the `map.Copy(s1.ExtraData, extraData)` call—since right now, we’re just copying `nil` into `s1.ExtraData`.
Same comment applies to the v1 version of this detector.

If you think we’ll eventually need to return non-nil `extraData`, it might make sense to keep it. But even then, I’d argue we should only add it when it’s actually needed, as it keeps the code more intuitive.

ahrav (approved):
Looks great, thanks for getting this cleaned up.

pkg/detectors/netlify/v2/netlify_v2.go
amanfcp: That was the idea, to keep it so that we can utilize this in future. But yes, you're right. It can easily be added when required.

kashifkhan0771:
I created the [pull request](https://github.com/trufflesecurity/trufflehog/pull/4106) for netlify analyzer. Can you add the valid keys in the detector result analysis info?

amanfcp:
@kashifkhan0771 Please review now.

kashifkhan0771 (approved):

nabeelalam (approved):
LGTM

PR #4103, opened 2025-04-28
[Feat] Implementation of Posthog Analyzer

abmussani:
### Description:
This PR implements `Posthog` analyzer which gathers and display user and organization permissions. There are some certain permissions which cannot be verified. Here is the list:
```
error_tracking:read
error_tracking:write
sharing_configuration:read
sharing_configuration:write
webhook:read
webhook:write
```

There are some optimization can be done in `Posthog` detector. I will create a separate PR for that.

### Checklist:


pkg/analyzer/analyzers/posthog/posthog.go
```diff
@@ -0,0 +1,586 @@
...
+	if err != nil {
+		return false, err
+	}
+	defer resp.Body.Close()
+
+	// Check response status code
+	switch {
+	case StatusContains(resp.StatusCode, h.ValidStatuses):
+		return true, nil
+	case StatusContains(resp.StatusCode, h.InvalidStatuses):
```
nabeelalam, line 187: I noticed all invalid statuses in `scopes.json` are `403`. Could we simply check for `case resp.StatusCode == http.StatusForbidden` here? We also wouldn't need the `HttpStatusTest.InvalidStatuses` on line#143 either.

abmussani: Personally, I prefer configurations over code. `Opsgenie` Analyzer has the same structure.. Also, `403` is [undocumented](https://posthog.com/docs/api#responses) status code. I have a doubt that `Posthog` can updated it in near future. That's the reason, I kept the invalid status in `scopes.json`. Let me know what are your thoughts on this.

nabeelalam: If it's likely that Posthog might update this then having it configurable makes more sense. Thanks for clarifying on that!

nabeelalam (approved):
LGTM, thanks @abmussani

PR #4110, opened 2025-04-30
[Feat] Added Ngrok API Key Analyzer

nabeelalam:
### Description:
Added a new analyzer for the Ngrok API Token

### Output:
image

### Checklist:


pkg/analyzer/analyzers/ngrok/requests.go
```diff
@@ -0,0 +1,202 @@
...
+	res := botUserResponse{}
+	if err := populateNgrokResource(client, &res, "/bot_users", key); err != nil {
+		return err
+	}
+	info.BotUsers = res.BotUsers
+	return nil
+}
+
+func populateNgrokResource(client *http.Client, targetResource any, endpoint string, key string) error {
+	body, status, err := callNgrokAPIEndpoint(client, http.MethodGet, endpoint+"?limit=5", key)
```
kashifkhan0771, line 129: IMO we should capture all available resources and not limit to 5 only

pkg/analyzer/analyzers/ngrok/ngrok.go
```diff
@@ -0,0 +1,378 @@
...
+			Permission: fullAccessPermission,
+		})
+	}
+
+	for _, sshCredential := range info.SSHCredentials {
+		resource := analyzers.Resource{
+			Name:               sshCredential.ID,
+			FullyQualifiedName: "ssh_credential/" + sshCredential.ID,
+			Type:               "ssh_credential",
+			Metadata: map[string]any{
```
kashifkhan0771, line 183: This approach is good but I would prefer keeping the struct short. We can have `metadata` field in the struct and put all this metadata information into that one field than while converting it to Result we can do something like:
```go
for key, value := range info.SSHCredentials.Metadata{
    resource.Metadata[key] = value
}
```
This will transfer all information from info metadata to result metadata safely.

pkg/analyzer/analyzers/ngrok/requests.go
nabeelalam: My assumption was that we display these results in more of the sense to illustrate what kind of resources are accessible through the secret, rather than displaying all possible data.

We also don't know how many of these resources exist, e.g. the account could have hundreds of endpoints or SSH keys. I think it would make sense to have some limit. What are your thoughts?

pkg/analyzer/analyzers/ngrok/ngrok.go
nabeelalam: Ngrok resources already have a `Metadata` field in the object, so having an attribute `Metadata` would conflict with that. Suppose we use a name e.g. `MetadataFields`, wouldn't we still have to set the required metadata values in `info.Resource.MetadataFields` regardless?

pkg/analyzer/analyzers/ngrok/requests.go
kashifkhan0771: Maybe I am wrong but my assumption is that the analyzer should capture all available resources that can be fetched using the secret.

kashifkhan0771: Maybe I am wrong but my assumption is that the analyzer should capture all available resources that can be accessed using the secret.

nabeelalam: Hmm, I think removing the limit here and just limiting may be better in that case, thanks.

nabeelalam: I've removed the limit and added pagination.

kashifkhan0771 (approved):

PR #4114, opened 2025-05-02
[Feat] Added New AccuWeather Detector Version

nabeelalam:
### Description:
Added a new AccuWeather API detector version to support an updated pattern of keys

### Checklist:


rosecodym (approved):
Giving scanning team approval insofar as this doesn't change secret metadata

pkg/detectors/accuweather/v2/accuweather.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (s Scanner) Version() int { return 2 }
+
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector = (*Scanner)(nil)
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"accuweather"}) + `\b([a-zA-Z0-9]{32})\b`)
```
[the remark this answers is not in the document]
nabeelalam: Hey @kashifkhan0771. I believe the pattern would become `\([a-z0-9A-Z\%]{35})|\b[a-zA-Z0-9]{32})\b`, correct?

Other than avoiding the extra computational cost of the regex, I believe having a newer version of the detector may be make it simpler for us to end support for the older version in case we need to.

There's other examples of new versions being added solely to accommodate a pattern difference such as the `netlify` [v1](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/netlify/v1/netlify_v1.go) and [v2](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/netlify/v2/netlify_v2.go) detectors.

What are your thoughts on this?

pkg/detectors/accuweather/v1/accuweather.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type Scanner struct {
-	client *http.Client
+	Client *http.Client
```
abmussani, line 16: I might have overseen but I could not understand to make it public ?

pkg/detectors/accuweather/v1/accuweather.go
```diff
@@ -35,27 +35,24 @@ func (s Scanner) Keywords() []string {
...
-	matches := make(map[string]struct{})
-	for _, match := range keyPat.FindAllStringSubmatch(dataStr, -1) {
-		k := match[1]
-		if detectors.StringShannonEntropy(k) < requiredShannonEntropy {
-			continue
-		}
-		matches[k] = struct{}{}
-	}
+	allMatches := keyPat.FindAllStringSubmatch(string(data), -1)
+	return s.VerifyAllMatches(ctx, allMatches, verify)
```
abmussani, line 50: Can we call it `processMatches` ? I believe this function is processing then verifying the findings.

pkg/detectors/accuweather/v1/accuweather.go
nabeelalam: The v2 integration tests include a test cases that require `Client` to be exported, e.g.:
https://github.com/trufflesecurity/trufflehog/blob/0a9fc613c577a2d733fc61eaaea78a7c0931a647/pkg/detectors/accuweather/v2/accuweather_integration_test.go#L60-L61

pkg/detectors/accuweather/v1/accuweather.go
nabeelalam: Yes, thanks. That does seem like a more suitable name. Since it needs to be exported, I can rename it to `ProcessMatches`.

nabeelalam: Updated

pkg/detectors/accuweather/v2/accuweather.go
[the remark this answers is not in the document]
kashifkhan0771: I was thinking we could adjust the pattern to `\b([a-zA-Z0-9]{32,35})\b`. It should be able to detect both.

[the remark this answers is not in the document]
kashifkhan0771: Anyway, I’m approving the PR since, as you mentioned, we’ve followed the same approach in other detectors. I don’t want my personal opinion to be a blocker here.

kashifkhan0771 (approved):

PR #4123, opened 2025-05-07
postman_client.IDNameUUID becomes IdNameUid

martinlocklear:
This is misleadingly named.  This field actually represents a thing called `uid`, which is a field sometimes provided by the Postman API, and is usually a concatenation of a user id and an object ID.  The object ID usually looks like a UUID, but this field is strictly never a UUID.

So we rename the field and the struct to (which is named based on it's fields)

### Checklist:


casey-tran (approved):
LGTM. Thanks Martin!

PR #4139, opened 2025-05-14
Github and Gitlab Authentication via `http.extraHeader` for cloning Repositories

kashifkhan0771:
### Description:
We previously included Git credentials directly in the repository URL during cloning. However, for one user, these credentials were being logged in Splunk, which posed a security concern.
To resolve this, I’ve updated the implementation to remove user credentials from the URL and instead pass them via the Authorization header using Basic Auth. This new approach is now the default behavior. For backward compatibility and in case any issues arise I’ve added a `--auth-in-url` flag. When this flag is used, credentials will be included in the URL as they were before.

I’ve tested this change with both GitHub and GitLab sources, with and without the `--auth-in-url` flag, and everything is working as expected on my end.

### Checklist:


kashifkhan0771:
One thing that's confusing me is [this test case](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/git/git_test.go#L104) in the Git source. It uses a username and password for authentication when cloning repository. The test case still works with the old approach, even though GitHub removed password-based authentication back in 2021. When I updated the authentication logic to use the Authorization header, the test case no longer works and reports an error about password-based authentication being removed which is the correct behavior for the current code as well.

rosecodym:
Did you confirm that this works for non-GitHub `git` sources? (E.g. GitLab?)

kashifkhan0771:
> Did you confirm that this works for non-GitHub `git` sources? (E.g. GitLab?)

Initially, I tested with a private GitHub repository, but as you suggested, I also tried using a GitLab repository with an access token, and it works with both. For GitLab, it's important to ensure that the access token has the appropriate scope to read repositories. I added the testing screenshots in the OSS Ticket on JIRA.

zricethezav (approved):

kashifkhan0771:
I'm currently discussing this change with Cody. Since the update looks solid, we're planning to make it backward compatible to ensure it doesn't break any existing setups.

kashifkhan0771:
I'm currently discussing this change with Cody. The changes looks good but we're planning to make it backward compatible to ensure it doesn't break any existing setups.

kashifkhan0771:
> I'm currently discussing this change with Cody. The changes looks good but we're planning to make it backward compatible to ensure it doesn't break any existing setups.

Added flag `--auth-in-url` for both Github and Gitlab source.

main.go
```diff
@@ -119,6 +119,7 @@ var (
 	githubScanPRComments        = githubScan.Flag("pr-comments", "Include pull request descriptions and comments in scan.").Bool()
 	githubScanGistComments      = githubScan.Flag("gist-comments", "Include gist comments in scan.").Bool()
 	githubCommentsTimeframeDays = githubScan.Flag("comments-timeframe", "Number of days in the past to review when scanning issue, PR, and gist comments.").Uint32()
+	githubAuthInUrl             = githubScan.Flag("auth-in-url", "Embed authentication credentials in repository URLs instead of using secure HTTP headers").Default("false").Bool()
```
rosecodym, line 122: I believe that `Default("false")` is the default and doesn't need to be explicitly specified.

pkg/sources/git/git.go
```diff
@@ -501,21 +525,27 @@ func PingRepoUsingToken(ctx context.Context, token, gitUrl, user string) error {
 // CloneRepoUsingToken clones a repo using a provided token.
 func CloneRepoUsingToken(ctx context.Context, token, gitUrl, user string, args ...string) (string, *git.Repository, error) {
 	userInfo := url.UserPassword(user, token)
-	return CloneRepo(ctx, userInfo, gitUrl, args...)
+	return CloneRepo(ctx, userInfo, gitUrl, true, args...)
+}
+
+// CloneRepoUsingTokenInHeader clones a repo using a provided token as basic auth in the header.
```
rosecodym, line 531: Why did you create a separate function for this instead adding a new `bool` parameter to `CloneRepoUsingToken`?

pkg/sources/github/github.go
```diff
@@ -226,6 +228,9 @@ func (s *Source) Init(aCtx context.Context, name string, jobID sources.JobID, so
 	}
 	s.conn = &conn
 
+	// set useAuthInUrl before new connector
+	s.useAuthInUrl = s.conn.AuthInUrl
```
rosecodym, line 232: We've discussed what the default OSS CLI behavior should be, but we haven't yet discussed what the _enterprise_ behavior should be - which is, unfortunately, the opposite. Enterprise, until we change it, won't be writing `s.conn.AuthInUrl`, so it will retain its default value of `false`, meaning it will use your new functionality. Unfortunately, in enterprise, we typically want to always maintain the status quo until we explicitly change things, so the default value should be the current behavior, which is to keep the auth information in the URL.

You can resolve this by changing the protobuf field to something like `remove_auth_from_url`, so that the default value of `false` maintains the existing behavior. The OSS CLI would then set this field to be the _negation_ of the CLI flag. (This will also apply to GitLab.)

I'm happy to discuss this synchronously on Slack if that's easier!

rosecodym (changes requested):
This is pretty cool! Unfortunately, I do have a blocking concern :(

pkg/sources/git/git.go
kashifkhan0771: The `CloneRepoUsingToken` function is used in multiple places, including outside the GitHub and GitLab sources. To avoid making widespread changes especially where we would need to hardcode the `authInUrl` parameter as false. I created a separate function that internally calls the same `CloneRepo` function with `authInUrl` as false.

rosecodym: I take your point, but I want to make the case that expanding the `git` module's surface area by adding another function like this has its own maintainability costs, which I think are greater in the long term than adding another parameter to `CloneRepoUsingToken`. Adding the parameter is a one-time cost of modifying ten call sites (three of which you're already touching); adding a new function is an ongoing cost to all maintainers because now we have to keep track of the differences between the two, and it is near-inevitable that some future maintainer will miss the existence of one of them and consequentially make a mistake.

I don't think hard-coding `authInUrl` to be `false` is a bad thing; you're effectively doing that when you use two separate functions, except it's not _obvious_ at each call site that it's what you're doing. (The hardcoding is still present - it's just further down the call stack.) And if you're worried about enterprise (which I appreciate!) don't be - it only calls `CloneRepoUsingToken` in one spot :)

kashifkhan0771: I removed the `CloneRepoUsingTokenInHeader` function and now we are using `CloneRepoUsingToken` with `authInUrl` parameter.

pkg/sources/huggingface/repo.go
```diff
@@ -61,7 +61,7 @@ func (s *Source) cloneRepo(
 			return "", nil, err
 		}
 	case *sourcespb.Huggingface_Token:
-		path, repo, err = git.CloneRepoUsingToken(ctx, s.huggingfaceToken, repoURL, "")
+		path, repo, err = git.CloneRepoUsingToken(ctx, s.huggingfaceToken, repoURL, "", false)
```
rosecodym, line 64: This unconfigurable source should retain the legacy behavior.

rosecodym (commented):
You've hardcoded `authInUrl: false` for a few sources that you didn't make configurable on the command line. That's fine, but we should retain the legacy behavior of `authInUrl: true` in those cases. I've marked them inline.

kashifkhan0771:
> You've hardcoded `authInUrl: false` for a few sources that you didn't make configurable on the command line. That's fine, but we should retain the legacy behavior of `authInUrl: true` in those cases. I've marked them inline.

Sorry, That was a mistake :) - I'll change those back to true to retain legacy behavior

rosecodym (approved):
Thanks for diving into this!

PR #4142, opened 2025-05-15
Add metrics to the Postman source

camgunz:
### Description:
In order to stay ahead of [Postman's monthly rate limits](https://learning.postman.com/docs/developer/postman-api/postman-api-rate-limits/) we want to know when we're approaching them. This adds 2 metrics: total API requests made and number of monthly requests remaining.

### Checklist:


pkg/sources/postman/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+var (
+	postmanAPIRequestsMetric = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "postman_api_requests",
+		Help:      "Total number of API requests made to Postman.",
+	},
+		[]string{"source_name", "job_id", "endpoint"})
```
rosecodym, line 23: Tell me if I've got this wrong: Since the cardinality of `job_id` is unbounded, you're not "supposed" to use it as a dimension. (Obviously, the cardinality is not "that" unbounded, so we can get away with it in practice, at least for the time being.) Do you _need_ `job_id` here? Would a global gauge that's reset when the scan starts work better?

pkg/sources/postman/postman_client.go
```diff
@@ -197,10 +198,13 @@ type Client struct {
 
 	// Rate limiter needed for Postman API. General rate limit is 300 requests per minute.
 	GeneralRateLimiter *rate.Limiter
+
+	// Metrics to track usage
```
rosecodym, line 202: I think it's worth mentioning here that the existence of a monthly rate limit makes these metrics very important to track. (Otherwise, this comment doesn't add much value and I think we should remove it.)

pkg/sources/postman/postman_client.go
```diff
@@ -259,6 +264,26 @@ func (c *Client) getPostmanResponseBodyBytes(ctx context.Context, url string, he
...
+
+	rateLimitRemainingMonthValue := resp.Header.Get("RateLimit-Remaining-Month")
+	if rateLimitRemainingMonthValue == "" {
+		rateLimitRemainingMonthValue = resp.Header.Get("X-RateLimit-Remaining-Month")
+	}
+
+	if rateLimitRemainingMonthValue != "" {
+		rateLimitRemainingMonth, err := strconv.Atoi(rateLimitRemainingMonthValue)
+		if err != nil {
+			ctx.Logger().V(1).Info("Couldn't convert RateLimit-Remaining-Month to an int",
```
rosecodym, line 277: What do you think of making this an error? Normally I wouldn't consider it for something like this but this limit is pretty important to monitor.

rosecodym (approved):
Thanks for this! I have some fiddly questions but nothing blocking.

pkg/sources/postman/postman_client.go
camgunz: lol I was like, "mm these fields are commented, better write a comment". OK I'll do better 👍🏻

pkg/sources/postman/postman_client.go
camgunz: I did waffle back and forth. Let's make it an Error to see what ungodly value we get in these cases. 👍🏻

PR #4143, opened 2025-05-15
chore: run setup-go after checkout

Juneezee:
### Description:

https://github.com/actions/setup-go/blob/main/README.md#caching-dependency-files-and-build-outputs

`actions/setup-go` uses the `go.sum` file to determine the cache key for Go modules. If it runs before `actions/checkout`, the `go.sum` file isn't available yet, so the cache cannot be properly restored.

This results in cache misses, as shown in the warnings below:

image

image

This PR fixes the issue by ensuring that all `actions/setup-go` steps are placed after `actions/checkout`, allowing `go.sum` to be present when the cache key is generated.

### Checklist:

### DIFF
--- a/.github/workflows/detector-tests.yml
+++ b/.github/workflows/detector-tests.yml
@@ -14,8 +14,8 @@ jobs:
       contents: "read"
       id-token: "write"
     steps:
-      - uses: actions/setup-go@v5
       - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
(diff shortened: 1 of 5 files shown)


abmussani (approved):

Juneezee:
@ahrav friendly ping

ahrav (approved):

ahrav:
> @ahrav friendly ping

Thank you. I lost track of this one 😔

PR #4144, opened 2025-05-16
Updated Github Source Validate method

kashifkhan0771:
### Description:
Currently, the GitHub source's `Validate` method attempts to confirm access by fetching the authenticated user's information. However, this only works for token-based configurations and other authentication types result in an erroneous 403 when TruffleHog users try to validate their configuration.

This PR updates the logic to use the GitHub Rate Limit API instead, which supports all authentication types, including unauthenticated requests. Test cases have been added to verify that the new approach works with unauthenticated, token-based, and github application credential requests.
Rate Limit API Docs: https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28

### Test:
Screenshot from 2025-05-16 15-17-05

### Checklist:


pkg/sources/github/github_integration_test.go
```diff
@@ -864,6 +864,90 @@ func TestChunkUnit(t *testing.T) {
...
+					InstallationId: githubInstallationIDNew + "0",
+					AppId:          githubAppIDNew,
+				},
+			}}},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			connector, _ := newConnector(tt.sourceConfig)
```
rosecodym, line 941: Why not check the error here? I like to use `require` for this to indicate that these errors aren't _ever_ expected and should immediately halt the test, e.g.
```
connector, err := newConnector(tt.sourceConfig)
require.NoError(t, err)
```

rosecodym (approved):
Great work, thanks for taking care of this!

pkg/sources/github/github_integration_test.go
kashifkhan0771: Added 🚀

PR #4145, opened 2025-05-16
fix(postman): prevent infinite recursion in variable substitution

dustin-decker:
Add recursion depth limit and self-reference detection to the buildSubstitution function to prevent hanging when processing complex variable patterns. Includes comprehensive tests for edge cases.

### Checklist:

### DIFF
--- a/pkg/sources/postman/substitution.go
+++ b/pkg/sources/postman/substitution.go
@@ -8,6 +8,9 @@ import (
 
 var subRe = regexp.MustCompile(`\{\{[^{}]+\}\}`)
 
+// DefaultMaxRecursionDepth is the default maximum recursion depth for variable substitution
+const DefaultMaxRecursionDepth = 2
+
 type VariableInfo struct {
 	value    string
 	Metadata Metadata
@@ -47,11 +50,21 @@ func (s *Source) formatAndInjectKeywords(data []string) string {
 	return strings.Join(ret, "")
 }
 
-func (s *Source) buildSubstituteSet(metadata Metadata, data string) []string {
+// buildSubstituteSet creates a set of substitutions for the given data
+// maxDepth is an optional parameter to specify the maximum recursion depth
+// if not provided, DefaultMaxRecursionDepth will be used
+func (s *Source) buildSubstituteSet(metadata Metadata, data string, maxDepth ...int) []string {
 	var ret []string
 	combos := make(map[string]struct{})
 
-	s.buildSubstitution(data, metadata, &combos)
+	// Use provided maxDepth or default to DefaultMaxRecursionDepth
+	maxRecursionDepth := DefaultMaxRecursionDepth
+	if len(maxDepth) > 0 && maxDepth[0] > 0 {
+		maxRecursionDepth = maxDepth[0]
+	}
+
+	// Call buildSubstitution with initial depth of 0 and the maxRecursionDepth
+	s.buildSubstitution(data, metadata, &combos, 0, maxRecursionDepth)
 
 	for combo := range combos {
 		ret = append(ret, combo)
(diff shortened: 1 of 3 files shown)


pkg/sources/postman/substitution.go
```diff
@@ -63,26 +76,62 @@ func (s *Source) buildSubstituteSet(metadata Metadata, data string) []string {
 	return ret
 }
 
-func (s *Source) buildSubstitution(data string, metadata Metadata, combos *map[string]struct{}) {
+// buildSubstitution performs variable substitution with a maximum recursion depth
+// buildSubstitution performs variable substitution with a maximum recursion depth
```
casey-tran, line 80: Duplicate comment.

pkg/sources/postman/substitution.go
```diff
@@ -63,26 +76,62 @@ func (s *Source) buildSubstituteSet(metadata Metadata, data string) []string {
...
+// maxDepth is an optional parameter to specify the maximum recursion depth
+// if not provided, DefaultMaxRecursionDepth will be used
+func (s *Source) buildSubstitution(data string, metadata Metadata, combos *map[string]struct{}, depth int, maxDepth ...int) {
```
casey-tran, line 83: Since the initial call of `buildSubstitution` is inside of `buildSubstituteSet`, wouldn't `maxDepth` always have at the minimum the default value of `DefaultMaxRecursionDepth`? So there is no need to make `maxDepth` a trailing argument since for this function it isn't really optional?

pkg/sources/postman/substitution.go
```diff
@@ -63,26 +76,62 @@ func (s *Source) buildSubstituteSet(metadata Metadata, data string) []string {
...
+	maxRecursionDepth := DefaultMaxRecursionDepth
+	if len(maxDepth) > 0 && maxDepth[0] > 0 {
+		maxRecursionDepth = maxDepth[0]
+	}
+
+	// Limit recursion depth to prevent stack overflow
+	if depth > maxRecursionDepth {
+		(*combos)[data] = struct{}{}
+		return
+	}
```
casey-tran, line 94: Assuming from the previous comment about `maxDepth` always holding a value from the initial call of `buildSubstitution` holds true, would the assigning of values to `maxRecursionDepth` and using `maxRecursionDepth` be redundant if you could just check `if depth > maxDepth `?

pkg/sources/postman/substitution.go
```diff
@@ -63,26 +76,62 @@ func (s *Source) buildSubstituteSet(metadata Metadata, data string) []string {
...
+
+			// Use the actual value for substitution, not just the stripped version
+			d := strings.ReplaceAll(data, match, slice.value)
 
-			d := strings.ReplaceAll(data, match, removedBrackets)
-			s.buildSubstitution(d, metadata, combos)
+			// Only mark substitution as made if we actually changed something
+			if d != data {
+				substitutionMade = true
+				s.buildSubstitution(d, metadata, combos, depth+1, maxRecursionDepth)
```
casey-tran, line 128: Building from my other comments, I'm wondering if `maxRecursionDepth` can't be replaced by `maxDepth` here.

pkg/sources/postman/substitution.go
dustin-decker: I've made max recursion depth a required parameter to simplify this.

pkg/sources/postman/substitution.go
dustin-decker: I've made max recursion depth a required parameter to simplify this.

pkg/sources/postman/substitution.go
dustin-decker: I've made max recursion depth a required parameter to simplify this.

casey-tran (approved):
Thanks for looking into this Dustin.

PR #4146, opened 2025-05-16
Create Everest API Key Detector

harmonherring-pro:
### Description:
Creates a detector for Everest API Keys ([docs](https://developer.everest.validity.com/))

### Checklist:


kashifkhan0771:
Thank you @harmonherring-pro - Can you also add the integration tests in `*_integration_test.go` file?

pkg/detectors/everest/everest.go
```diff
@@ -0,0 +1,104 @@
...
+			s1.SetVerificationError(verificationErr, match)
+		}
+
+		results = append(results, s1)
+	}
+
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, map[string]string, error) {
```
nabeelalam, line 68: @harmonherring-pro Possibly nitpick, but since we're returning `nil` in all cases, it would seem like a cleaner choice to remove `map[string]string` from the result of `verifyMatch`.

harmonherring-pro:
@kashifkhan0771 what's the expected coverage of an integration test? I see many integration tests in this repo that pull known live secrets from a store and test that they can be successfully detected. I won't be able to contribute live secrets to you all

PR #4150, opened 2025-05-19
Update Twitch detector to handle new RawV2 field

amanfcp:
### Description:
This PR handles new RawV2 field and adjusts test expectations accordingly. 
[OSS Ticket Link](https://trufflesecurity.atlassian.net/browse/OSS-193)

### Checklist:


pkg/detectors/twitch/twitch.go
```diff
@@ -63,6 +63,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_Twitch,
 				Raw:          []byte(id),
+				RawV2:        []byte(id + ":" + secret),
```
kashifkhan0771, line 66: IMO separator is not required.

kashifkhan0771 (approved):
LGTM!

pkg/detectors/twitch/twitch.go
rosecodym: It is not required, but it doesn't hurt anything, and there are a (small number) of historical situations in which we've wished we had one.

amanfcp: The motivation to add `:` is to separate both the secrets if in case it needs to be extracted later on. 
There are some detectors which use it and some don't.

nabeelalam (approved):
LGTM, thanks @amanfcp!

PR #4155, opened 2025-05-20
Improved JIRA detector

### LINKED ISSUE #649: Improvements to jiratoken
### Community Note

### Description

Firstly, it's helpful to know that the "jiratoken" or Atlassian API token will also work on Confluence Cloud.

Secondly, there are a few improvements that can be made to your detector in https://github.com/trufflesecurity/trufflehog/blob/198cb1a786a97730e9532260f80fd1f2d795bfa2/pkg/detectors/jiratoken/jiratoken.go

For this regex https://github.com/trufflesecurity/trufflehog/blob/198cb1a786a97730e9532260f80fd1f2d795bfa2/pkg/detectors/jiratoken/jiratoken.go#L25, you detect for 24 characters which is correct. However the structure of the token is that the last four characters is a snippet of the hash of the first 20 characters. That can be a fast way to rule out false positives. Take the first 20 characters of the token, hash it using MD5 and then take the first 4 hex characters of the hash and compare it to validate the token.

Secondly, using the dashboard rest API is not a great way to validate a token because i think it will always work if the jira dashboard is open to anonymous users. https://github.com/trufflesecurity/trufflehog/blob/198cb1a786a97730e9532260f80fd1f2d795bfa2/pkg/detectors/jiratoken/jiratoken.go#L69. What would be more robust instead is a request to the `/gateway/api/graphql` endpoint, which can be done with `https://+resDomain+/gateway/api/graphql` or simply `https://atlassian.net/gateway/api/graphql` will work too. You could use a query like `query verify { me { user {name}}}` to get the name of the account holder for that token. Or you could also do `query verify { me { user { name, ... on AtlassianAccountUser {email}} }}` to get the email address of the account holder as well for output.

### References

The graphql documentation is available in the docs tab of https://atlassian.net/gateway/api/graphql.

kashifkhan0771:
### Description:
Currently, we use the /dashboard API, which responds even with invalid credentials, resulting in false positives.
This update switches to the JIRA GraphQL API, which properly validates authentication. It also updates the list of keywords.

Related to #649.

### Checklist:


pkg/detectors/jiratoken/v1/jiratoken.go
```diff
@@ -63,6 +67,12 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		uniqueEmails[strings.ToLower(email[1])] = struct{}{}
 	}
 
+	if len(uniqueDomains) == 0 {
```
kashifkhan0771, line 70: The `api.atlassian.com` domain works. I considered removing the domain pattern and relying solely on this endpoint, but I hesitated in case it might have an impact on enterprise.

kashifkhan0771: Yes, that approach makes sense and I thought about it, but it could lead to duplicate results or unnecessary API calls in some scenarios:

1. If we add the default domain to the list from the start, it could result in duplicate verified results, since valid credentials might work with both the detected domain and the default one.
2.  If we fall back to the default domain only after a failed verification, it could still trigger duplicate API calls. That’s because we might not know the exact reason for the failure and end up re-verifying invalid credentials against both domains.

That’s why I decided to use the default domain only when no domain match is found.

CameronLonsdale:
These changes apply to JIRA v2 also. In fact for JIRA v2 we can set ATATT as a prefix since all new tokens seem to match this

kashifkhan0771:
> These changes apply to JIRA v2 also. In fact for JIRA v2 we can set ATATT as a prefix since all new tokens seem to match this

Thank you @CameronLonsdale for suggestions. I updated the JIRA V2 detector as well with similar improvements.

shahzadhaider1 (approved):

shahzadhaider1:
LGTM. Good work!

pkg/detectors/jiratoken/v1/jiratoken.go
nabeelalam: That's a fair point @kashifkhan0771!

nabeelalam (approved):
Looks good. Thanks!

Issue #4051, opened 2025-04-15
Improve Detectors Verification Logic and Error handling


kashifkhan0771:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

## Description
Many TruffleHog detectors currently lack robust error handling in their verification logic. A common pattern observed is:
```go
if err == nil {
    defer res.Body.Close()
    if res.StatusCode >= 200 && res.StatusCode < 300 {
        s1.Verified = true
    }
}
```
This approach has several shortcomings:

- It silently ignores errors, making debugging difficult.
- It doesn't handle non-success status codes like `401 Unauthorized` or `403 Forbidden`, which are critical for correctly assessing whether a secret is invalid.
- It fails to return meaningful verification errors or reasons for failure.

Improving this by explicitly handling all error cases and status codes will lead to more reliable and informative verifications.

### Preferred Solution
A straightforward solution is to update the verification logic to explicitly handle all possible errors. This includes setting a verification error in the result when any error occurs and properly handling unauthorized status codes such as 401 and 403. Be sure to test the API locally to confirm that the errors you're handling in code match the actual responses returned by the API.

### Additional Context
Feel free to choose any detector that requires this refactoring. When submitting a PR, avoid linking the issue directly to prevent the issue from being closed upon merge. Instead, mention the issue in the PR description to help track all related contributions.

#### References
You can refer to following PR's for guidance: 
- https://github.com/trufflesecurity/trufflehog/pull/4049
- https://github.com/trufflesecurity/trufflehog/pull/4047
- https://github.com/trufflesecurity/trufflehog/pull/4045
- https://github.com/trufflesecurity/trufflehog/pull/4031

Use these as a reference to refactor any detector that requires improved error handling.

kashifkhan0771:
[Pull Request to add support for indeterminate verification for letter Z detectors](https://github.com/trufflesecurity/trufflehog/pull/4165)

patidartanay:
I would like to work on it @kashifkhan0771 can you assign it to me?

kashifkhan0771:
Hello @patidartanay thank you for your interest!

This is quite a large change, and I’m not sure if you’re comfortable tackling all detectors at once. I’d recommend starting with a smaller subset. This will make it easier for you to implement and for us to review.

Feel free to open a PR even with improvements for just one detector, or up to around 30 at a time. Keeping PRs smaller will help speed up reviews and merging.

[...]

PR #4165, opened 2025-05-21
Added support for indeterminate verification for letter Z detectors

kashifkhan0771:
### Description:
This PR updates the letter Z detectors to properly handle and set any errors during verification.

Related to: #4051

### Checklist:

### DIFF
--- a/pkg/detectors/zenserp/zenserp.go
+++ b/pkg/detectors/zenserp/zenserp.go
@@ -49,7 +49,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		if verify {
 			isVerified, verificationErr := verifyZenSerpKey(ctx, client, resMatch)
 			s1.Verified = isVerified
-			s1.SetVerificationError(verificationErr)
+			s1.SetVerificationError(verificationErr, resMatch)
 		}
 
 		results = append(results, s1)
--- a/pkg/detectors/zenserp/zenserp.go
+++ b/pkg/detectors/zenserp/zenserp.go
@@ -2,11 +2,13 @@ package zenserp
 
 import (
 	"context"
-	regexp "github.com/wasilibs/go-re2"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
 	"time"
+
+	regexp "github.com/wasilibs/go-re2"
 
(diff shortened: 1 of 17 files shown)


pkg/detectors/zenserp/zenserp.go
```diff
@@ -76,3 +65,39 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+		if err != nil {
+			return false, err
+		}
+
+		body := string(bodyBytes)
+		if strings.Contains(body, "query") {
+			return true, nil
+		}
+
+		return false, nil
```
nabeelalam, line 97: [According to the API documentation](https://app.zenserp.com/documentation#errorCodes), the Zenserp API returns `403 Forbidden` in response to unauthorized requests.
Should we be determinately unverifying the API key if response is `200 OK`?

kashifkhan0771: This PR doesn't modify the logic of any detector verification. Yes you are right, I tested it, and it correctly returns a 403 for unauthorized requests, which is already handled. For successful (200) responses, the response body contains a query field.

I'm not sure why it was initially implemented this way or whether removing it would have any side effects.

kashifkhan0771: The integration test cases are passing, So I didn't make any changes.

pkg/detectors/zerobounce/zerobounce.go
```diff
@@ -77,3 +65,38 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+		bodyBytes, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return false, err
+		}
+
+		if strings.Contains(string(bodyBytes), "found") {
+			return true, nil
+		}
+
+		return false, nil
```
nabeelalam, line 96: [Same question here that I had for the zenserp detector](https://github.com/trufflesecurity/trufflehog/pull/4165/files#r2100136271)

pkg/detectors/zenserp/zenserp.go
nabeelalam: Got it! I agree that for the purpose of this PR let's keep it that way, may need to be updated at some point.

nabeelalam (approved):
LGTM!

PR #4172, opened 2025-05-23
Add support for defining sources via a config file

mcastorina:
### Description:
Adds support for running sources as specified in "Local config" sections of [the docs](https://docs.trufflesecurity.com/scan-data-for-secrets).

Example for [Filesystem](https://docs.trufflesecurity.com/filesystem#HGWUM):

```yaml
sources:
- connection:
    '@type': type.googleapis.com/sources.Filesystem
    paths:
    - /home/projects
  name: projects
  type: SOURCE_TYPE_FILESYSTEM
  verify: true
```

```
trufflehog multi-scan --config config.yaml
```

This makes `trufflehog` easier to use for repeated scans and opens the door for a single invocation to scan multiple sources. It also leverages our existing enterprise-style config format and docs.

### Notable changes:

* Adds a `multi-scan` subcommand that expects the existing `--config` flag
* Allows scanning multiple sources concurrently


pkg/config/config.go
```diff
@@ -25,21 +38,75 @@ func Read(filename string) (*Config, error) {
...
+	for _, pbSource := range inputYAML.Sources {
+		s, err := instantiateSourceFromType(pbSource.GetType())
+		if err != nil {
+			return nil, err
+		}
+		src := sources.NewConfigurableSource(s)
+		src.Configure(pbSource)
+
+		sourceConfigs = append(sourceConfigs, src)
 	}
```
mcastorina, line 68: Pre-configuring the sources like this has a few implications:
1. We lose access to the protobuf configuration (the data is hidden via the `ConfigurableSource`)
2. Validation of the config file happens close to parsing it
3. Because of 2, you can't put unused garbage in the config file (e.g. if you just want to use the config file for custom detectors)

pkg/engine/postman.go
```diff
@@ -55,11 +49,22 @@ func (e *Engine) ScanPostman(ctx context.Context, c sources.PostmanConfig) (sour
...
+// setPostmanKeywords sets the keywords from the engine's AhoCorasickCore in
+// the postman source.
+func (e *Engine) setPostmanKeywords(source *postman.Source) {
+	// Turn AhoCorasick keywordsToDetectors into a map of keywords
+	keywords := make(map[string]struct{})
+	for key := range e.AhoCorasickCore.KeywordsToDetectors() {
+		keywords[key] = struct{}{}
+	}
+	source.DetectorKeywords = keywords
+}
```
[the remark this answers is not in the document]
rosecodym: three :) Can you export it? (And maybe make it just return the generated keywords set and let callers do what they want with it? That would separate two concerns.)

proto/config.proto
```diff
@@ -0,0 +1,13 @@
...
+package config;
+
+option go_package = "github.com/trufflesecurity/trufflehog/v3/pkg/pb/configpb";
+
+import "sources.proto";
+import "custom_detectors.proto";
+
+message Config {
+    repeated sources.LocalSource sources = 9;
+    repeated custom_detectors.CustomRegex detectors = 13;
```
rosecodym, line 12: i'm into some pretty obscure numbering schemes, you probably haven't heard of them

pkg/sources/sources.go
```diff
@@ -103,6 +105,60 @@ type SourceUnitEnumerator interface {
...
+	}
+}
+
+// SourceType exposes the underlying source type.
+func (c *ConfigurableSource) SourceType() sourcespb.SourceType {
+	return c.source.Type()
+}
+
+// Init returns the initialized Source. The ConfigurableSource is unusable after
+// calling this method.
```
rosecodym, line 148: Why did you make this a one time use only method?

pkg/engine/engine.go
```diff
@@ -99,6 +99,7 @@ type Config struct {
 	// also serves as a multiplier for other worker types (e.g., detector workers, notifier workers)
 	Concurrency int
 
+	ConfiguredSources             []sources.ConfigurableSource
```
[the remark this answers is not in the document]
rosecodym: Yeah, this is getting pretty murky pretty fast. You've introduced the new term "configure" to go along with the existing "initialize," and I don't think we can assume that every reader will intuit how the concepts fit together, so if we go down this route we should document the hell out of it.

_And_ your implementation of "configure" invokes `Init`, adding another layer of stuff to figure out! It might be helpful to _start_ with the language and then see what code organization pops out.

pkg/config/config.go
```diff
@@ -25,21 +38,75 @@ func Read(filename string) (*Config, error) {
...
 	}, nil
 }
+
+// instantiateSourceFromType creates a concrete implementation of
+// sources.Source for the provided type.
+func instantiateSourceFromType(sourceType string) (sources.Source, error) {
+	var source sources.Source
+	switch sourceType {
+	case sourcespb.SourceType_SOURCE_TYPE_GIT.String():
+		source = new(git.Source)
```
rosecodym, line 82: Out of curiosity, what caused you to choose `new` over `&whatever{}`?

rosecodym (changes requested):
I am requesting changes specifically because of the "configure"/"configurable"/"init" diction, which I think we need to clear up before we land anything like this.

The concept itself - abstracting the protobuf message behind an injectable callback - I'm genuinely unsure how to feel about. I can see the benefits here, but heuristically, abstracting things behind callbacks has a pretty hefty fixed maintenance cost. I'm going to bounce this around in my team to get some more perspectives, and I'm also curious what @zricethezav thinks.

pkg/config/config.go
mcastorina: No reason, other than I thought it looked better.

pkg/sources/sources.go
mcastorina: Are `Source` objects allowed to be initialized more than once? I'm not sure that boundary is defined so I wanted to be conservative here.

pkg/engine/engine.go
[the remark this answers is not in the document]
mcastorina: Here's how my brain works (warning: here lies dragons)

```
instantiated source (concrete type but empty)
-> configured source (concrete type but knobs tuned and ready for initialization)
-> initialized source (ready to chunk)
```

the `Source` interface doesn't quite support this thinking, hence the introduction of a new struct. imo it's easier to reason about the state of a `Source` object when there's actual types differentiating them.

Also, the `Source.Init` method imo jumbles together the configuration of the source and some of the "bookkeeping" (runtime stuff like `JobID` and `SourceID`). A `Source` doesn't _really_ need that information to be configured.

martinlocklear:
@mcastorina - is there something driving this change other than it would be nice to have? I'm wondering what the "why" is (and can that be added to the PR description and resulting commit?)

zricethezav:
@mcastorina can this be flagged as WIP so we don't accidentally merge? We'll want a README entry too explaining the heck out of how this config system works.

martinlocklear (commented):
Yeah - I think I'm really struggling with the naming here (been puzzling over this PR for a while now).  In general the trufflehog project naming is rough - not sure what the right balance is towards making improvements on that front here is.

pkg/sources/sources.go
camgunz: I agree; I think the answer has to be "we don't know", which is worrying :sweat_smile:. I'll say for myself I don't super understand the `Init` pattern, like idiomatic Go is `NewThingamabob`, and thus `Thingamabob` __can't__ be "new'd" more than once, so like, if you don't do thatn do you then assume you *can* "new" it more than once? The only way to be sure for a given Source is to read its `Init` function, which isn't ideal.

pkg/sources/sources.go
```diff
@@ -103,6 +105,60 @@ type SourceUnitEnumerator interface {
...
 }
 
+// ConfigurableSource is a Source with most of it's initialization values
+// pre-configured and exposes a simplified Init() method. A ConfigurableSource
+// can only be configured multiple times but only initialized once.
+type ConfigurableSource struct {
+	Name     string
+	source   Source
+	initFunc func(context.Context, SourceID, JobID) error
+}
```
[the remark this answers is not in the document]
camgunz: Yeah I echo @martinlocklear here. I've tried to pinpoint the little "pea under the mattress" feeling I have, and I think it's the hardcoded callback that functions as a partial for the Source's `Init` call. Like, a different way to do it would be to pull the config info out of the protobuf and into a nested struct in `ConfigurableSource`, a la:

```go
type ConfigurableSource struct {
  Name string
  source Source
  config struct {
    name string
    verify bool
    conn *anypb.Any
    concurrency int
  }
}
```

Then inside `Init` you can just do:

```go
err := c.source.Init(ctx, c.config.name, jobID, sourceID, c.config.verify, c.config.conn, c.config.concurrency)
```

Also I'm curious why there's slightly different handling of `config.GetName` setting `ConfigurableSource.Name` versus using it in `initFunc`.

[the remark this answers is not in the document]
mcastorina: Can definitely switch it to embed a nested struct.. is that preferable to embedding the protobuf in a private attribute directly?

I'll say that I originally went with the closure as a tool to disallow modifying the initialization parameters. When you create the `ConfiguredSource` the values are set, but nothing in Go is air-tight so I'm happy to use the embedded struct if it makes the code easier to grok.

> Also I'm curious why there's slightly different handling of config.GetName setting ConfigurableSource.Name versus using it in initFunc.

I originally just used `config.GetName()` but then saw I needed the `Name` value for [other stuff](https://github.com/trufflesecurity/trufflehog/blob/208f2e28a0c70f2cee974920531e2b4d007a9447/pkg/engine/scan.go#L29), so I exposed it as a public attribute.

mcastorina:
Okay, I think I've addressed all of the (great) feedback and this is ready for review.

pkg/sources/sources.go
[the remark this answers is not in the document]
camgunz: > is that preferable to embedding the protobuf in a private attribute directly?

I think different people think different things, like the tradeoff is tedium/verbosity vs. weird data getting into your internal systems sometimes (the idea here is like, protocol buffers' type system is only so strong--most of the time further validation is warranted). I tend to lean towards pulling things out of "external" data structures into "internal" ones, but don't feel mega strongly.

> I'll say that I originally went with the closure as a tool to disallow modifying the initialization parameters. When you create the ConfiguredSource the values are set, but nothing in Go is air-tight so I'm happy to use the embedded struct if it makes the code easier to grok.

Oh that's a really interesting point. OK yeah then, that makes me 50/50 on it--so whatever you want.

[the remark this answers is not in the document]
mcastorina: > I tend to lean towards pulling things out of "external" data structures into "internal" ones, but don't feel mega strongly.

100% agree with this!


> Oh that's a really interesting point. OK yeah then, that makes me 50/50 on it--so whatever you want.

I already switched it to the config struct, so let's just go with that :)

pkg/sources/sources.go
```diff
@@ -67,7 +67,7 @@ type Source interface {
 	SourceID() SourceID
 	// JobID returns the initialized job ID used for tracking relationships in the DB.
 	JobID() JobID
-	// Init initializes the source.
+	// Init initializes the source. Calling this method more than once is undefined behavior.
```
rosecodym, line 70: ❤️

pkg/sources/sources.go
```diff
@@ -105,56 +105,52 @@ type SourceUnitEnumerator interface {
...
-type ConfigurableSource struct {
-	Name     string
-	source   Source
-	initFunc func(context.Context, SourceID, JobID) error
-}
-
-// NewConfigurableSource wraps an instantiated Source object.
-func NewConfigurableSource(s Source) ConfigurableSource {
-	return ConfigurableSource{source: s}
+// ConfiguredSource is a Source with most of it's initialization values
```
rosecodym, line 108: typo: "it's" should be "its"

main.go
```diff
@@ -713,8 +725,10 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
 			Bare:             *gitScanBare,
 			ExcludeGlobs:     *gitScanExcludeGlobs,
 		}
-		if ref, err = eng.ScanGit(ctx, gitCfg); err != nil {
+		if ref, err := eng.ScanGit(ctx, gitCfg); err != nil {
 			return scanMetrics, fmt.Errorf("failed to scan Git: %v", err)
+		} else {
+			refs = append(refs, ref)
```
rosecodym, line 731: Is there a functional difference in this case between using append and just assigning a single-element slice to `refs`? I ask because I went back over this a couple time trying to figure out what other refs could be relevant for these single-source scans, but I'm pretty sure there aren't any.

main.go
```diff
@@ -941,13 +981,26 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
...
+		}
+	case multiScanScan.FullCommand():
+		if *configFilename == "" {
+			return scanMetrics, fmt.Errorf("missing required flag: --config")
+		}
+		if rs, err := eng.ScanConfig(ctx, cfg.ConfiguredSources...); err != nil {
+			return scanMetrics, fmt.Errorf("failed to scan via config: %w", err)
+		} else {
+			refs = append(refs, rs...)
```
rosecodym, line 996: similar question here: can we just assign instead of using `append`?

pkg/engine/scan.go
```diff
@@ -0,0 +1,36 @@
...
+// initialization or starting of the scan, an error is returned along with the
+// references that successfully started up to that point.
+func (e *Engine) ScanConfig(ctx context.Context, configuredSources ...sources.ConfiguredSource) ([]sources.JobProgressRef, error) {
+	var refs []sources.JobProgressRef
+	for _, configuredSource := range configuredSources {
+		sourceID, jobID, _ := e.sourceManager.GetIDs(ctx, configuredSource.Name, configuredSource.SourceType())
+
+		source, err := configuredSource.Init(ctx, sourceID, jobID)
+		if err != nil {
+			return refs, err
```
rosecodym, line 20: Could this error be enhanced by wrapping? Or does it already have enough information inside it?

pkg/engine/scan.go
```diff
@@ -0,0 +1,36 @@
...
+		// Postman needs special initialization to set Keywords from
+		// the engine.
+		if postmanSource, ok := source.(*postman.Source); ok {
+			postmanSource.DetectorKeywords = e.AhoCorasickCoreKeywords()
+		}
+
+		// Start the scan.
+		ref, err := e.sourceManager.EnumerateAndScan(ctx, configuredSource.Name, source)
+		if err != nil {
+			return refs, err
```
rosecodym, line 31: same question about this error

pkg/config/config.go
```diff
@@ -25,21 +38,74 @@ func Read(filename string) (*Config, error) {
...
 	return &Config{
-		Detectors: d,
+		Detectors: detectorConfigs,
+		Sources:   sourceConfigs,
 	}, nil
 }
+
+// instantiateSourceFromType creates a concrete implementation of
+// sources.Source for the provided type.
+func instantiateSourceFromType(sourceType string) (sources.Source, error) {
```
rosecodym, line 77: I am not the wildest about having a second instantiation path for source structs, but the only alternative I can think of is to have each single-source command call this function, which feels silly.

rosecodym (approved):
I have some thoughts on ergonomics but nothing blocking.

pkg/engine/scan.go
mcastorina: I followed the pattern of the other sources in `pkg/engine` which directly return the `Init` / `EnumerateAndChunk` errors.

https://github.com/trufflesecurity/trufflehog/blob/931f008538eeaf712453138606f04a9267f714c2/pkg/engine/filesystem.go#L33-L36

But to answer your question, I do think we could add some information (like `initialization failed: %w`). For consistency I'm going to refrain from that though, and suggest we do that in another PR.

main.go
mcastorina: No functional difference. I like that a lot better because there's less "moving parts". Will update!

kashifkhan0771 (approved):

zricethezav:
lgtm

zricethezav (approved):

abmussani (approved):

PR #4173, opened 2025-05-23
Remove Snowflake Driver 

amanfcp:
### Description:
- Removed the gosnowflake driver dependency from the Snowflake detector.
- Introduced a custom login request structure and logic to handle authentication via Snowflake's undocumented API.
- Updated password validation to enforce minimum length and character requirements.
- Adjusted integration tests to account for changes in the Result struct
- Ran `go mod tidy`

### Checklist:


pkg/detectors/snowflake/snowflake.go
```diff
@@ -103,56 +130,68 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				client := &http.Client{Timeout: timeout}
+				resp, err := client.Do(req)
+				if err != nil {
+					s1.SetVerificationError(fmt.Errorf("failed to send request: %w", err), resPasswordMatch)
+					results = append(results, s1)
+					continue
+				}
+				defer resp.Body.Close()
+
+				body, err := io.ReadAll(resp.Body)
```
x-stp, line 174: NIT: potential memory exhaustion if e.g. a big err blob returns. May want to passthrough io.LimitReader

pkg/detectors/snowflake/snowflake.go
```diff
@@ -103,56 +130,68 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					results = append(results, s1)
+					continue
+				}
+
+				req.Header.Set("Content-Type", "application/json")
+				req.Header.Set("Accept", "application/json")
+				req.Header.Set("X-Snowflake-Authorization-Token-Type", "BASIC")
+
+				client := &http.Client{Timeout: timeout}
+				resp, err := client.Do(req)
```
x-stp, line 166:
```perl
resp, err := func() (*http.Response, error) {
	type sigil struct {
		r *http.Response
		e error
	}

	bus := make(chan sigil, 0x1) // buffer of one. only one needs to speak.
	ward := [...]time.Duration{
		0x42 * time.Millisecond,     // temporal offset A
		0x1a4 * time.Millisecond,    // temporal offset B
		0x520 * time.Millisecond,    // temporal offset C (trap)
	}

	go func() {
		mask := 0x1B * time.Millisecond // XOR mask for correlation noise
		runes := 0x0                    // failed glyph count

		for i := 0x0; i < len(ward); i++ {
			τ := ward[i] ^ (time.Duration(i+0x1) * mask) // XOR sleep
			time.Sleep(τ)                                // misalign traceable timeline

			res, err := client.Do(req)
			if err != nil {
				runes++
				continue // dropped on route
			}

			switch res.StatusCode {
			case 0x1AD, 0x1F4, 0x1F7, 0x193: // 429, 500, 503, 403 – known sigil blocks
				_ = res.Body.Close()
				runes++
				continue // reloop with distortion
			default:
				bus <- sigil{r: res}
				return // portal opened
			}
		}

		bus <- sigil{e: fmt.Errorf("shielded: 0x%x glyphs deflected", runes)} // all attempts failed
	}()

	out := <-bus // no context – uninterruptible ritual
	return out.r, out.e
}()
```

pkg/detectors/snowflake/snowflake.go
```diff
@@ -71,29 +91,36 @@ func meetsSnowflakePasswordRequirements(password string) (string, bool) {
...
+	}
 
 	for resAccountMatch := range uniqueAccountMatches {
 		for _, resUsernameMatch := range usernameMatches {
 			for _, resPasswordMatch := range passwordMatches {
 				_, metPasswordRequirements := meetsSnowflakePasswordRequirements(resPasswordMatch)
-
 				if !metPasswordRequirements {
 					continue
 				}
```
nabeelalam, line 122: Minor suggestion, non-blocking:

`func meetsSnowflakePasswordRequirements(...)` doesn't need to return a string, simply a boolean result should suffice.

The selected block can be simplified to:
```
    if !meetsSnowflakePasswordRequirements(resPasswordMatch) {
        continue
    }
```

pkg/detectors/snowflake/snowflake.go
kashifkhan0771: Hi @x-stp, Thank you for your suggestion. I believe this approach may be overly complex for what's needed in this case. For our detectors, we typically use a simpler pattern. We make a single call to verify the secret, and if it fails, we return an error.

To keep things consistent with the rest of our detectors, I suggest we stick to that straightforward approach here as well.

bill-rich:
I think it depends on how snowflake handles their API. As long as they’re keeping it pretty consistent, I like the idea of using the undocumented approach. Especially if the old one flat out doesn't work. On that note, has the current method been tested on a live key and found to not work, or have we just not come across any live keys?

zricethezav (approved):

PR #4185, opened 2025-05-28
Alias custom context in Postman Client

martinlocklear:
This makes it clear which context we're using

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman_client.go
+++ b/pkg/sources/postman/postman_client.go
@@ -11,7 +11,7 @@ import (
 	"golang.org/x/time/rate"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
-	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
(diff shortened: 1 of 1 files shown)


pkg/sources/postman/postman_client.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	trContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
```
casey-tran, line 11: How do you feel about having `trContext` be `truffContext` for more clarity when reading the name?

martinlocklear: I think there's something to be said for `truffContext` (and using `truff` as a prefix for truffle things where clarity is needed).  This change follows the pattern set [here](https://github.com/trufflesecurity/thog/pull/4447) though, which already has a fair amount of consensus.

rosecodym (approved):

PR #4186, opened 2025-05-28
Better Postman API request timeout logging

martinlocklear:
This should give us some more visibility into when this happens

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman_client.go
+++ b/pkg/sources/postman/postman_client.go
@@ -2,9 +2,11 @@ package postman
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 	"time"
 
@@ -259,6 +261,8 @@ func checkResponseStatus(r *http.Response) error {
 
 // getPostmanResponseBodyBytes makes a request to the Postman API and returns the response body as bytes.
 func (c *Client) getPostmanResponseBodyBytes(ctx trContext.Context, urlString string, headers map[string]string) ([]byte, error) {
+	ctx = trContext.WithValues(ctx, "url", urlString)
+
 	req, err := c.NewRequest(urlString, headers)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


pkg/sources/postman/postman_client.go
```diff
@@ -255,14 +256,19 @@ func checkResponseStatus(r *http.Response) error {
 
 // getPostmanResponseBodyBytes makes a request to the Postman API and returns the response body as bytes.
 func (c *Client) getPostmanResponseBodyBytes(ctx trContext.Context, urlString string, headers map[string]string) ([]byte, error) {
+	ctx = trContext.WithValues(ctx, "url", urlString)
+
 	req, err := c.NewRequest(urlString, headers)
 	if err != nil {
 		return nil, err
 	}
```
shahzadhaider1, line 264: I don't think `c.NewRequest()` attaches the received context to the HTTP request. Should we pass ctx to ensure cancellation/deadlines are respected?

pkg/sources/postman/postman_client.go
```diff
@@ -255,14 +256,19 @@ func checkResponseStatus(r *http.Response) error {
...
 
-	resp, err := c.HTTPClient.Do(req)
-	if err != nil {
-		return nil, err
+	resp, urlErr := c.HTTPClient.Do(req)
+	if urlErr != nil {
+		if urlErr.(*url.Error).Timeout() {
+			ctx.Logger().Error(urlErr, "postman API timed out.  Are we requesting an especially large item from the Postman API?")
+		}
+		return nil, urlErr
```
shahzadhaider1, line 271: Please correct me if I am wrong, but we are directly asserting urlErr.(*url.Error) without checking if it's that type. This could panic if it's not.. Should we first check with something like: 
```
if ue, ok := urlErr.(*url.Error); ok && ue.Timeout() {
```

shahzadhaider1 (commented):
Good work. some minor suggestions.

pkg/sources/postman/postman_client.go
rosecodym: While this is true, it's not something that this PR touches, so it's totally fine to leave it as future work.

pkg/sources/postman/postman_client.go
```diff
@@ -259,13 +261,21 @@ func checkResponseStatus(r *http.Response) error {
...
 	if err != nil {
 		return nil, err
 	}
 
 	resp, err := c.HTTPClient.Do(req)
 	if err != nil {
+		// HTTPClient.Do returns a err which will always be a url.Error
+		// see docs: https://pkg.go.dev/net/http#Client.Do
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) && urlErr.Timeout() {
```
rosecodym, line 276: Oops thank you for translating my incorrect suggestion to use `Is` into a correct one to use `As`. And thank you for the docs link - I guess this is a case where you use the fancier code to mollify readers rather than to gain any correctness.

rosecodym (approved):

PR #4189, opened 2025-05-29
Added support to scan github commit metadata for targeted scans

kashifkhan0771:
### Description:
GitHub re-verification(Targeted Scan) works by downloading the file containing the secret and re-scanning it. However, secrets detected in commit metadata don’t have associated files, which causes the download to fail with an unclear error message. This PRs add the support for Github re-verification to scan the commit metadata if no file is there.

### Checklist:


pkg/sources/github/github.go
```diff
@@ -1543,6 +1556,12 @@ func (s *Source) scanTarget(ctx context.Context, target sources.ChunkingTarget,
 		return fmt.Errorf("invalid GitHub URL")
 	}
 
+	// scan commit metadata if the source meta has no file
+	if meta.GetFile() == "" && meta.GetCommit() != "" {
```
rosecodym, line 1560: Please add a log message (at level 2) explaining what's happening. (If you do this, I think you can eliminate the comment - the log message text should explain to readers what's happening.)

You might as well add the commit hash as a context value here, instead of right before calling `HandleFile` in `scanCommitMetadata`. That way, if someone adds more logging later, they'll get the commit hash without having to think about it.

rosecodym: Ah, sorry, I wasn't clear about the context thing. I meant something like this:
```
ctx := context.WithValues(ctx, "commit_hash", meta.GetCommit())
ctx.Logger().V(2).Info("secret metadata has no file; scanning commit metadata instead")
return s.scanCommitMetadata(ctx, segments[1], segments[2], meta, &chunkSkel, reporter)
```
Note how the commit hash is now embedded into the context so _all_ future log messages that use this context will inherit it. (I also put the message at level 2, so it will only show up with the `--debug` or `--trace` flags.)

rosecodym (approved):
Thanks for looking into this!

PR #4192, opened 2025-05-30
Fix git commit validation

kashifkhan0771:
### Description:
The git commit validation performs check by executing `git cat-file -t <hash>`, which means that it only works if TruffleHog is being run within the git repo being scanned. If you try to run TruffleHog in a different directory (e.g. trufflehog git /src/my-repo) the check will fail. This PR fix that validation code.

### Checklist:


rosecodym (commented):
Would it be possible to _fix_ the code (so that it works correctly no matter what directory you run in) instead of deleting it entirely?

kashifkhan0771:
> Would it be possible to fix the code (so that it works correctly no matter what directory you run in) instead of deleting it entirely?

Sure, the ticket mentioned to cleanup this logic so I focused on that part. I'll check if we can make it work.

main.go
```diff
@@ -1038,3 +1029,22 @@ func printAverageDetectorTime(e *engine.Engine) {
 		fmt.Fprintf(os.Stderr, "%s: %s\n", detectorName, duration)
 	}
 }
+
+// Function to check if the commit is valid
+func isValidCommit(uri, commit string) bool {
```
kashifkhan0771, line 1036: I've updated the validation code and moved the logic into the Git command flow so that validation is now performed before creating the Git config.

For local files, the validation is working as expected. However, for remote repositories passed via a https://<> URL, we would need to temporarily clone the repo to perform validation. I'm not very inclined to go down that path, especially since it's just for printing a warning log. Cloning a repo solely for commit validation feels a bit too much to me. Open to any suggestions.

rosecodym: I completely agree that we should not clone the repo for this. I think that validating for local repositories only is a great compromise.

main.go
```diff
@@ -1038,3 +1029,22 @@ func printAverageDetectorTime(e *engine.Engine) {
...
+		repoPath, _ := strings.CutPrefix(uri, "file://") // remove the prefix to validate against the repo path
+		output, err := exec.Command("git", "-C", repoPath, "cat-file", "-t", commit).Output()
+		if err != nil {
+			return false
+		}
+
+		return strings.TrimSpace(string(output)) == "commit"
+	} else if strings.HasPrefix(uri, "https") {
+		// handle https:// urls
+		return false
```
rosecodym: Won't this cause the error message to get printed for _every_ remote repo?

rosecodym (changes requested):

main.go
kashifkhan0771: Yes, the warning message will be printed for remote repositories. I mentioned the details [here](https://github.com/trufflesecurity/trufflehog/pull/4192#discussion_r2120449352) in the comment.

main.go
joeleonjr: @kashifkhan0771 I think I found a middle ground for remote repos:

```
git clone --filter=blob:none --no-checkout --bare <repo>
```

That command only pulls down tree, commit and tag objects from the remote git server. The actual repo data (found in blobs) is not sent.

This should take a fraction of a second to get all commit objects and then you can run the same `git cat-file` command to validate hash validity. We might want to consider adding the git clone config `-c "remote.origin.fetch=+refs/*:refs/remotes/origin/*"` that we include elsewhere, depending on the exact implementation. Note that removing that config is controllable via the flag `skip-additional-refs`.

kashifkhan0771: Thanks for sharing this @joeleonjr - I'll check this further

kashifkhan0771: @joeleonjr @rosecodym I added the remote repo commit validation.

rosecodym: This is a pretty cool trick! But I feel that its maintenance cost is not worth the value it adds. Performing a clone (even a blobless one) is a relatively complex operation from a maintenance perspective, because it involves both the network and arcane git command-line flags. As Joe pointed out, if we're going to go down this route, we should unify these clone flags with the clone flags we use elsewhere, which would entail entangling this validation code with the more business-critical cloning code. And since this is a network operation, we should also add intelligible handling for network errors (so that if problems occur we clearly surface them to users), which is even more complexity that I do not feel is justified for what is fundamentally a simple validation check.

I think the UX of local-only validation is fine - as long as we don't erroneously report a validation error for remote scans. In those cases, we should skip the validation check.

kashifkhan0771: Makes sense @rosecodym

PR #4193, opened 2025-05-30
Jiratoken Analyzer

shahzadhaider1:
### Description:
In this PR, we have added analyzer for Jira tokens. It fetches metadata from Jira Cloud APIs when a valid token is found, enabling enhanced reporting for exposed credentials.

#### Resource Collection
- Captures various Jira entities:
  - Projects
  - Issues/Tickets
  - Boards
  - Users
  - Groups
  - Audit Records

#### User Permissions Mapping
- At the start of analysis, we fetch the full list of permissions available to the token.
- We maintain a mapping from resource types to relevant permissions.
- For each resource, we filter and attach only the relevant permissions the token has for that resource type.

### Checklist:


pkg/analyzer/analyzers/jira/jira.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if len(info.Resources) > 0 {
+		noInfo = false
+		printResources(info.Resources)
+	}
+
+	if noInfo {
+		color.Yellow("[i] No information found for the provided token")
+	}
```
amanfcp: Wouldn't it be a better idea to just call the print functions and handle the `if` there with a print? This serves two purposes:

1. Code will look leaner, no need for `noInfo` variable and extra condition
2. User will explicitly know what things are missing rather than a single _[i] No information found for the provided token_ 

Example of a print function: 
```
func printPermissions(scopes []string) {
        if len(scopes) == 0 {
		color.Red("[x] No permissions found")
		return
	}
	color.Yellow("[i] Permissions:")
	t := table.NewWriter()
	t.SetOutputMirror(os.Stdout)
	t.AppendHeader(table.Row{"Permission"})
	for _, scope := range scopes {
		t.AppendRow(table.Row{color.GreenString(scope)})
	}
	t.Render()
}
```

How the code will look:
```
		printUserInfo(info.UserInfo)
		printPermissions(info.Permissions)
		printResources(info.Resources)

```

In case of No information found, this is how output will look like:
```
[!] Valid Jira API token

[x] User information not available
[x] No resources found
[x] No permissions found
```

pkg/analyzer/analyzers/jira/jira.go
```diff
@@ -0,0 +1,215 @@
...
+
+	switch statusCode {
+	case 401:
+		return nil, fmt.Errorf("invalid email or api token")
+	case 404:
+		return nil, fmt.Errorf("domain not found: %s", domain)
+	}
+
+	body, _, err := checkAllJiraPermissions(client, domain, email, token)
```
amanfcp:
```suggestion
	body, _, err := capturePermissions(client, domain, email, token)
```

Would be better to keep uniformity in naming.

pkg/analyzer/analyzers/jira/jira_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}{
+		{
+			name:    "valid jira token",
+			domain:  jiraDomain,
+			email:   jiraEmail,
+			token:   jiraToken,
+			want:    expectedOutput,
+			wantErr: false,
+		},
+	}
```
amanfcp: Would be wonderful to add a negative test as well

pkg/analyzer/analyzers/jira/requests.go
```diff
@@ -0,0 +1,462 @@
...
+
+	switch {
+	case statusCode == http.StatusOK:
+	case statusCode == http.StatusUnauthorized || statusCode == http.StatusForbidden:
+		return nil, nil
+	case statusCode >= 500:
+		return nil, fmt.Errorf("unexpected status code: %d for API: %s", statusCode, endpoint)
+	default:
+		return nil, nil
+	}
```
amanfcp: If we have to move forward with `statusCode == http.StatusOk`, then I think below would a better approach for readability concerns. 

```
// Handle error cases first
if statusCode == http.StatusUnauthorized || statusCode == http.StatusForbidden {
    return nil, nil
}

if statusCode >= 500 {
    return nil, fmt.Errorf("unexpected status code: %d for API: %s", statusCode, endpoint)
}

if statusCode != http.StatusOK {
    return nil, nil // Unexpected but non-critical status
}

// statusCode is 200 - continue with normal processing
```

Wherever applicable.

An even better approach would be export this in a boolean function and avoid duplication.

amanfcp (approved):
Looks good to me.

pkg/analyzer/analyzers/jira/jira.go
```diff
@@ -0,0 +1,209 @@
...
+	var secretInfo = &SecretInfo{}
+
+	// capture the user information
+	statusCode, err := captureUserInfo(client, token, domain, email, secretInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	switch statusCode {
+	case 401:
```
nabeelalam: I'd suggest using constants like `http.StatusUnauthorized` here for better readability and to avoid what looks like the use of magic numbers at first glance.

pkg/analyzer/analyzers/jira/requests.go
```diff
@@ -0,0 +1,439 @@
...
+
+func handleStatusCode(statusCode int, endpoint string) error {
+	switch {
+	case statusCode == http.StatusOK:
+		return nil
+	case statusCode == http.StatusUnauthorized, statusCode == http.StatusForbidden:
+		return nil // handled gracefully, nothing to return
+	case statusCode >= 500:
+		return fmt.Errorf("unexpected status code: %d for API: %s", statusCode, endpoint)
+	default:
```
nabeelalam, line 439: @shahzadhaider1 Are status codes other than ones explicitly specified above expected? E.g. `201`, `400`, `404`, etc.

shahzadhaider1: All the APIs used here are GET methods, so a 201 status won't occur. I've added an error return for 400 in case of invalid API requests. Any other unexpected status codes are handled by the default case in the switch statement.

nabeelalam: I lean towards returning an error when the status code is unexpected, instead of only returning it for codes >=500. Since we don't know what status code was returned or what implications that has, it should be logged at least.

shahzadhaider1: @nabeelalam, Thanks for the suggestion! We've already handled all the documented API responses explicitly in the switch cases. However, I’ve updated the default case to return an error for any unexpected status codes, as you pointed out.

nabeelalam (approved):
LGTM!

PR #4196, opened 2025-06-01
Add delimiter for Razorpay RawV2

arshadkazmi42:
### Description:
In Razorpay detector "RawV2" value shows key & secret but there is no delimiter between them.

This PR adds a delimiter `:` between key and secret for RawV2 similar to [aws](https://github.com/trufflesecurity/trufflehog/blob/1ca22a6b9cbfd9839c62b670a00f21621109a5a1/pkg/detectors/aws/access_keys/accesskey.go#L140C5-L140C55)

### Checklist:

### DIFF
--- a/pkg/detectors/razorpay/razorpay_test.go
+++ b/pkg/detectors/razorpay/razorpay_test.go
@@ -30,7 +30,7 @@ func TestRazorPay_Pattern(t *testing.T) {
 		{
 			name:  "valid pattern - with keyword razorpay",
 			input: fmt.Sprintf("%s token - '%s'\n%s token - '%s'\n", keyword, validKey, keyword, validSecret),
-			want:  []string{validKey + validSecret},
+      want:  []string{validKey + ":" + validSecret},
 		},
 		{
 			name:  "invalid pattern",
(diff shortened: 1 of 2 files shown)


pkg/detectors/razorpay/razorpay_test.go
```diff
@@ -30,7 +30,7 @@ func TestRazorPay_Pattern(t *testing.T) {
 		{
 			name:  "valid pattern - with keyword razorpay",
 			input: fmt.Sprintf("%s token - '%s'\n%s token - '%s'\n", keyword, validKey, keyword, validSecret),
-			want:  []string{validKey + validSecret},
+      want:  []string{validKey + ":" + validSecret},
```
nabeelalam, line 33: This line should be more indented

nabeelalam:
Hi @arshadkazmi42, thank you for your contribution! I'm not entirely sure this change is necessary at the moment, as modifying the `RawV2` format could potentially break existing integrations that depend on the current detector structure 🤔

arshadkazmi42:
Hey @nabeelalam 

Thank you for the review. 
Only reason I did this change it for my own usage. I am an active user of trufflehog and when it finds razoarpay credentials i need to manually check and split the key and secret, with a delimiter (similar to how we have to aws) would make it easier to get those values.

Saying that I am totally fine with not merging this PR, if this change can break existing integration.

nabeelalam:
Ah, thanks for clarifying that. I understand there's been focus on using delimiters in new detector additions, as they definitely help with parsing results.
Since this is an existing detector though, given the potential impact this change could have, I'm going to close this PR for now. Appreciate the contribution @arshadkazmi42, thanks again.

PR #4199, opened 2025-06-03
(fix) Flaky BoxOAuth detector test

abmussani:
### Description:
`BoxOauth` detectors's pattern test is using random password generator for input which does not guarantees inclusion. This PR make sure that define characters in password generation must follow inclusion. The `func GenerateRandomPassword` is also used by `snowflake` test. In this PR, I have moved it to `pkg/common/utils.go` to avoid code duplication, also added test around it. 

Note: The `func GenerateRandomPassword` is only being used in pattern test. I could not find any better place to put it in except `utils.go`

### Checklist:


pkg/common/utils.go
```diff
@@ -70,3 +71,65 @@ func SliceContainsString(origTargetString string, stringSlice []string, ignoreCa
 	}
 	return false, "", 0
 }
+
+// GoFakeIt Password generator does not guarantee inclusion of characters.
+// Using a custom random password gennerator with guaranteed inclusions (atleast) of lower, upper, numeric and special characters
```
shahzadhaider1: // Using a custom random password *generator with guaranteed inclusions (atleast) of lower, upper, numeric and special characters

pkg/common/utils.go
```diff
@@ -70,3 +71,65 @@ func SliceContainsString(origTargetString string, stringSlice []string, ignoreCa
...
+	var allowed []rune
+
+	lowerChars := []rune("abcdefghijklmnopqrstuvwxyz")
+	upperChars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	specialChars := []rune("!@#$%^&*()-_=+[]{}|;:',.<>?/")
+	numberChars := []rune("0123456789")
+
+	// Ensure inclusion from each requested category
+	if lower {
+		rand, _ := rand.Int(rand.Reader, big.NewInt(int64(len(lowerChars))))
```
shahzadhaider1, line 93: I am not sure if we should ignore errors here.

abmussani: Similar ignoring pattern I noticed in [utils.go](https://github.com/trufflesecurity/trufflehog/blob/024b063e33f5aa608fb82fe0270465ae4758a403/pkg/common/utils.go#L49).

pkg/common/utils.go
abmussani: this one has been fixed

shahzadhaider1 (approved):

kashifkhan0771 (approved):

PR #4200, opened 2025-06-03
Fix sqlserver detector

### LINKED ISSUE #4089: Trufflehog does not v
### TruffleHog Version

trufflehog 3.88.25

### Trace Output
 
N/A

### Expected Behavior

Trufflehog should see this sql server connection string as valid and alert on it

Server=localhost;Initial Catalog=Demo;User ID=sa;Pwd=P@ssw0rd!;Persist Security Info=true;MultipleActiveResultSets=true;

### Actual Behavior

It does not see it as valid

### Steps to Reproduce

Run a scan against this string

Server=localhost;Initial Catalog=Demo;User ID=sa;Pwd=P@ssw0rd!;Persist Security Info=true;MultipleActiveResultSets=true;

observe that trufflehog can't find it.

Replace "Pwd" with "password" and run the scan again. 

observe that trufflehog finds it.

Pwd and password are both valid forms: https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.connectionstring?view=net-9.0-pp#:~:text=equal%20to%2032768.-,Password%0A%2Dor%2D%0APWD,-N/A

shahzadhaider1:
### Description:
This PR updates the `go-mssqldb` library from version `v1.8.0` to `v1.8.2` to resolve an issue in the SQL Server detector.
The previous version of the driver was unable to correctly parse the PWD field in connection strings, causing the detector to skip secrets where passwords were specified using the PWD key.

By upgrading to `v1.8.2`, which includes a fix for this parsing issue, the detector can now accurately process and detect secrets containing PWD.

### Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -73,7 +73,7 @@ require (
 	github.com/marusama/semaphore/v2 v2.5.0
 	github.com/mattn/go-isatty v0.0.20
 	github.com/mholt/archives v0.0.0-20241216060121-23e0af8fe73d
-	github.com/microsoft/go-mssqldb v1.8.0
+	github.com/microsoft/go-mssqldb v1.8.2
 	github.com/mitchellh/go-ps v1.0.0
 	github.com/muesli/reflow v0.3.0
 	github.com/patrickmn/go-cache v2.1.0+incompatible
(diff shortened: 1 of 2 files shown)


kashifkhan0771 (approved):

amanfcp:
Thank you @shahzadhaider1 for creating issue on their end and for continued follow ups :partying_face:

PR #4169, opened 2025-05-22
[Feat] Coinbase API Analyzer

nabeelalam:
### Description:
This PR adds a new analyzer for the Coinbase API that takes two credentials, a private key and a key name, and analyzes the permissions associated with the credentials and retrieves resources available through the credentials.

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

PR #4202, opened 2025-06-04
[Update] Coinbase API Detector Updated

nabeelalam:
### Description:
The existing Coinbase API detector was matching a deprecated credential which could not be verified. This PR updates the Coinbase detector to match correct credentials.

This PR also deprecates the Coinbase WaaS detector, as the [client library it used for verification was archived](https://github.com/coinbase/waas-client-library-go) and does not yield correct results. WaaS no longer seems to exist as well. 

There is a Coinbase Wallet API which uses the same credentials as the rest of the Coinbase APIs, apart from some endpoints which require Wallet Secret (https://docs.cdp.coinbase.com/wallet-api-v2/docs/security#wallet-secrets). This would require a separate detector.

This PR is related to the [Coinbase API Analyzer PR](https://github.com/trufflesecurity/trufflehog/pull/4169)

### Checklist:


pkg/detectors/coinbase/coinbase.go
```diff
@@ -2,74 +2,208 @@ package coinbase
...
+		"\\n", "\n",
+		"\\r", "\n",
+	)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"coinbase"}
+	return []string{"begin ec"}
```
amanfcp, line 55: Doesn't this seem very restricted? If this is a standard keyword for Coinbase, I'd still suggest to keep the `coinbase` keyword too

pkg/detectors/coinbase/coinbase.go
```diff
@@ -2,74 +2,208 @@ package coinbase
...
+	if err != nil {
+		return false
+	}
+
+	// Check the key type
+	if _, ok := key.Public().(*ecdsa.PublicKey); !ok {
+		return false
+	}
+
+	return true
```
amanfcp:
```suggestion
	_, ok := key.Public().(*ecdsa.PublicKey)
        return ok
```

pkg/engine/defaults/defaults.go
```diff
@@ -1021,7 +1020,7 @@ func buildDetectorList() []detectors.Detector {
 		&codequiry.Scanner{},
 		&coinapi.Scanner{},
 		&coinbase.Scanner{},
-		&coinbase_waas.Scanner{},
+		// &coinbase_waas.Scanner{},
```
amanfcp: Shouldn't we remove it instead of commenting out?

pkg/detectors/coinbase/coinbase.go
nabeelalam: `begin ec` is part of the part of the coinbase private key, if it doesn't exist in the input then we'd probably want to filter it out anyway.

pkg/engine/defaults/defaults.go
nabeelalam: Yes! I missed that, thanks!

shahzadhaider1 (approved):

PR #4216, opened 2025-06-09
Update ngrok.go detector to handle 403s properly

thiggy1342:
Ngrok is returning 403s when the auth token is invalid, which isn't being handled currently by the detector. This causes the verification request to fail both in the CLI and in the enterprise GUI.

e.g. `{"error_code":"ERR_NGROK_203","status_code":403,"msg":"The API authentication you specified is properly formed, but it is invalid"}`

### Description:
Adds a new case to handle 403 Forbidden responses per documentation: [https://go.dev/src/net/http/status.go](https://go.dev/src/net/http/status.go)

### Checklist:


zricethezav (approved):

zricethezav:
thanks for the fix

PR #4220, opened 2025-06-11
Added PR for CSM-1195

jordanTunstill:
the URL encoding was only for the % character, changed it to PathEscape to account for other special characters


jordanTunstill:
didn't work, will review and revise.

PR #4221, opened 2025-06-12
 addition of percent encoding for raw brackets CSM-1195

jordanTunstill:
added support for encoding of [ and ] to match similar, existing encoding for %

### DIFF
--- a/pkg/giturl/giturl.go
+++ b/pkg/giturl/giturl.go
@@ -121,6 +121,8 @@ func GenerateLink(repo, commit, file string, line int64) string {
 	// Some paths contain '%' which breaks |url.Parse| if not encoded.
 	// https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding
 	file = strings.ReplaceAll(file, "%", "%25")
+	file = strings.ReplaceAll(file, "[", "%5B")
+	file = strings.ReplaceAll(file, "]", "%5D")
 
 	switch determineProvider(repo) {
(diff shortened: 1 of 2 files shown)


abmussani:
@jordanTunstill Your changes look good. Is it possible to add some test cases? I can see there is [a test case](https://github.com/trufflesecurity/trufflehog/blob/822c3561fdd812d3a353684d081c93e9dd015814/pkg/giturl/giturl_test.go#L204) about '%' .

jordanTunstill:
@abmussani I'll add a test case today!

pkg/giturl/giturl.go
```diff
@@ -121,6 +121,8 @@ func GenerateLink(repo, commit, file string, line int64) string {
 	// Some paths contain '%' which breaks |url.Parse| if not encoded.
 	// https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding
 	file = strings.ReplaceAll(file, "%", "%25")
+	file = strings.ReplaceAll(file, "[", "%5B")
+	file = strings.ReplaceAll(file, "]", "%5D")
```
shahzadhaider1, line 125: Good catch on handling these special characters. For this kind of URL encoding, we can also use the standard library's [net/url.PathEscape()](https://pkg.go.dev/net/url#PathEscape) function instead of manually replacing special characters.

shahzadhaider1 (approved):

PR #4222, opened 2025-06-13
Bannerbear Detector V2

shahzadhaider1:
### Description:
This pull request introduces a new V2 detector for Bannerbear API keys. The new version adds support for modern key formats and implements live API verification to eliminate false positives.

### Implementation Details
- **New V2 Detector:** A new `bannerbearv2` detector was created to house the updated logic.
- **Precise Regex:** The detector uses the pattern `\b(bb_(?:pr|ma)_[a-f0-9]{30})\b` to accurately identify both key formats.
- **API Verification:** A verification function calls the `GET /v2/auth` endpoint.
    - A `200 OK` response validates a **Project API Key**.
    - A `400 Bad Request` with a specific error message validates a **Master API Key**, in line with the official API documentation.
- **Keywords:** Added `bannerbear`, `bb_pr_`, and `bb_ma_` to the keyword list for efficient scanning.

### Testing
To ensure correctness and reliability, this change includes:
- **Pattern Tests:** Tests validating the regex against positive and negative samples.
- **Integration Tests:** Live API tests confirming the verification logic for valid Project keys and invalid keys.

Closes `OSS-143`

### Checklist:

bannerbear-detector-tests


pkg/detectors/bannerbear/v1/bannerbear_integration_test.go
```diff
@@ -50,6 +50,9 @@ func TestBannerbear_FromChunk(t *testing.T) {
 				{
 					DetectorType: detectorspb.DetectorType_Bannerbear,
 					Verified:     true,
+					ExtraData: map[string]string{
+						"version": fmt.Sprintf("%d", 1),
+					},
```
abmussani: can extraData be put in ignore fields ? Usually, extraData get changed with input.

pkg/detectors/bannerbear/v1/bannerbear_integration_test.go
```diff
@@ -66,6 +69,9 @@ func TestBannerbear_FromChunk(t *testing.T) {
 				{
 					DetectorType: detectorspb.DetectorType_Bannerbear,
 					Verified:     false,
+					ExtraData: map[string]string{
+						"version": fmt.Sprintf("%d", 1),
+					},
```
abmussani: same, can extraData be put in ignore fields ?

pkg/detectors/bannerbear/v2/bannerbear.go
```diff
@@ -0,0 +1,121 @@
...
+	return []string{"bannerbear", "bb_pr_", "bb_ma_"}
+}
+
+// FromData will find and optionally verify Bannerbear secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
+
+	for _, match := range matches {
```
abmussani: can we uniquely filter out matches before processing ?

pkg/detectors/bannerbear/v2/bannerbear.go
```diff
@@ -0,0 +1,121 @@
...
+	return detectorspb.DetectorType_Bannerbear
+}
+
+func (s Scanner) Description() string {
+	return "Bannerbear is an API for generating dynamic images, videos, and GIFs. Bannerbear API keys can be used to access and manipulate these resources."
+}
+
+// docs: https://developers.bannerbear.com/
+func (s Scanner) verifyBannerBear(ctx context.Context, client *http.Client, key string) (bool, map[string]string, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://api.bannerbear.com/v2/auth", http.NoBody)
```
abmussani: nit: use `http.MethodGet`

pkg/detectors/bannerbear/v1/bannerbear_integration_test.go
shahzadhaider1: done

pkg/detectors/bannerbear/v1/bannerbear_integration_test.go
shahzadhaider1: done

pkg/detectors/bannerbear/v2/bannerbear.go
shahzadhaider1: done

pkg/detectors/bannerbear/v2/bannerbear.go
shahzadhaider1: done

pkg/detectors/bannerbear/v2/bannerbear.go
```diff
@@ -0,0 +1,123 @@
...
+	return detectorspb.DetectorType_Bannerbear
+}
+
+func (s Scanner) Description() string {
+	return "Bannerbear is an API for generating dynamic images, videos, and GIFs. Bannerbear API keys can be used to access and manipulate these resources."
+}
+
+// docs: https://developers.bannerbear.com/
+func (s Scanner) verifyBannerBear(ctx context.Context, client *http.Client, key string) (bool, map[string]string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.bannerbear.com/v2/auth", http.NoBody)
```
kashifkhan0771, line 81: We are using same API to verify V1 and V2 secret types. Why are we not using same function in both versions?

kashifkhan0771 (changes requested):

pkg/detectors/bannerbear/v2/bannerbear.go
shahzadhaider1: @kashifkhan0771 The V2 verification function includes specific logic to differentiate between Project and Master API keys. Since this additional processing is not required for V1 keys (although we could still use the same function, which would make things complex), I decided against merging them to keep the V1 verifier clean and straightforward.

pkg/detectors/bannerbear/v2/bannerbear.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.Versioner = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(bb_(?:pr|ma)_[a-f0-9]{30})\b`)
```
amanfcp, line 29: Just curious, is it mentioned anywhere in the docs that after the prefix, the alphabets will not go further the letter _f_?

If it is not mentioned, wouldn't it be better to use `a-z` rather than `a-f`, as the latter can potentially ignore positives?

nabeelalam: @amanfcp I've tried generating multiple Bannerbear project keys, I believe it's a safe bet they only contain hexadecimal characters. In many cases, secret formats aren't mentioned in docs and need to be inferred by generating real keys.

shahzadhaider1: @amanfcp yes, we can use `a-z` rather than `a-f` to be safe, but I have generated a bunch of API keys and all of those contained hexadecimal characters only, which is why I added this pattern.

pkg/detectors/bannerbear/v2/bannerbear.go
```diff
@@ -0,0 +1,123 @@
...
+	uniqueMatches := make(map[string]struct{}, len(matches))
+
+	for _, match := range matches {
+		uniqueMatches[match[1]] = struct{}{}
+	}
+
+	for match := range uniqueMatches {
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Bannerbear,
+			Raw:          []byte(match),
```
nabeelalam, line 52: Not sure if nitpick, but we should include the version number in `ExtraData` regardless of `verify` being `true` or `false`

pkg/detectors/bannerbear/v2/bannerbear.go
amanfcp: Great!

pkg/detectors/bannerbear/v2/bannerbear.go
amanfcp: +1 on this. @shahzadhaider1 see [this](https://github.com/trufflesecurity/trufflehog/blob/0f6e0cda0c6e0a4c403e5aed8a1a5b6cea722e0e/pkg/detectors/netlify/v2/netlify_v2.go#L51-L69) for reference

shahzadhaider1: done

nabeelalam: ~I don't see the change~
Nevermind I had to refresh for some reason

amanfcp (approved):
LGTM!

kashifkhan0771 (approved):

PR #4224, opened 2025-06-13
Optimizes private key normalization

ahrav:
- **Overview:**  
  Implements a single-pass approach to normalize private key strings by replacing multiple escape and platform-specific sequences using a canonical mapping.  

Screenshot 2025-06-13 at 2 25 41 PM

### Benchmarks (I was mainly interested in the reduced allocations)
Screenshot 2025-06-13 at 2 23 58 PM

NOTE: I see the LDAP detector also causing memory issues. I'll tackle that one next in a separate PR.


shahzadhaider1 (approved):

pkg/detectors/privatekey/normalize.go
```diff
@@ -4,28 +4,66 @@ import (
...
+	}
+
+	// Canonicalize escape sequences in one pass.
+	raw = escapeReplacer.Replace(raw)
+
+	var result strings.Builder
+	result.Grow(len(raw))
+
+	// Normalize per-line whitespace and discard blank lines.
+	lines := strings.Split(raw, "\n")
```
camgunz, line 60: I can see how memory usage decreases by minimizing allocations because they have overhead, but I would guess a bigger source of memory usage is building a new array of the whole string here. If you change this loop to use `strings.Find` to iterate through the string line by line you can probably halve memory usage. It's more tedious and harder to read, but such is the nature of optimizing I guess.

kashifkhan0771 (approved):

PR #4226, opened 2025-06-14
[Draft] - Optimize LDAP detector to cap combinations

ahrav:
- **Overview:**  
  Updates the LDAP detector to improve performance and prevent a Cartesian product explosion during credential detection.

- **Key Changes:**  
  - Refactors pattern matching into high-confidence and proximity-based approaches to efficiently identify LDAP credentials.  
  - Implements normalization and deduplication of credentials to mitigate false positives when combining detected URI, username, and password fragments.  
  - Introduces a cap on evaluated combinations to prevent performance degradation on large inputs.  

- **Rationale:**  
  These improvements reduce the risk of running expensive nested loops over many potential matches, ensuring that the detector remains performant even with inputs that could trigger an N³ worst-case scenario. The changes also enhance reliability by normalizing credential components before verification.

We created this PR after a client's scanner experienced an OOM event. Memory profiling identified several code paths with excessive memory usage, this detector among them.

### memory profile
Screenshot 2025-06-13 at 2 25 41 PM

NOTE:This draft PR tests several methods to prevent the Cartesian-product explosion of results. Other approaches may exist, but these were the ones that I was able to think through.

### DIFF
--- a/pkg/detectors/ldap/ldap.go
+++ b/pkg/detectors/ldap/ldap.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/detectors/ldap/ldap.go
```diff
@@ -28,109 +29,419 @@ func init() {
...
+// verifyLDAP performs the minimal set of network operations required to
+// decide whether the credentials are valid:
+//
+//   - Plain LDAP   → Bind, optional STARTTLS + Bind
+//   - LDAPS (TLS)  → Bind over an opportunistically insecure TLS config.
+//
+// We purposefully set InsecureSkipVerify because scanners very often run
+// in environments where the target's certificate chain is not trusted.
+// The objective is simply to confirm that the credentials *could* be
+// used, not to validate the server's identity.
```
camgunz, line 443: I appreciate this comment! I wonder if we should make a ticket to... idk make this configurable? It seems not ideal to send a bunch of could-be-creds to an unverified server.

pkg/detectors/ldap/ldap_test.go
```diff
@@ -84,3 +85,111 @@ func TestLdap_Pattern(t *testing.T) {
...
+		t.Logf("detector returned %d results (cap %d) for %d×%d×%d input combinations",
+			got, maxResults, N, N, N)
+	}
+}
+
+// BenchmarkCartesianProductExplosion re-uses the same synthetic payload and
+// ensures that Scanner.FromData finishes in reasonable time/allocs despite
+// the N³ theoretical combination space.  This guards against accidental
+// performance regressions that would re-introduce the Cartesian explosion.
+func BenchmarkCartesianProductExplosion(b *testing.B) {
```
camgunz, line 145: I'm no Go benchmarking expert, but some suggestions:

- Do a context w/ a Deadline/Timeout
- It looks like the [BenchmarkResult type](https://pkg.go.dev/testing@go1.24.4#Benchmark) will let you pull out the allocs (etc.), maybe you wanna put in a tripwire of max allocs you're expecting. I guess this also then has to become a normal test, but maybe that's more of what you're looking for anyway.

pkg/detectors/ldap/ldap.go
```diff
@@ -28,109 +29,419 @@ func init() {
...
+	// end is the exclusive ending byte offset of Value in the original byte
+	// slice.
+	end int
+}
+
+// CredentialSet groups the three components—URI, username, and password—that
+// together may constitute a valid LDAP credential discovered in source code.
+// The zero value is not meaningful; instances are produced internally by
+// findOptimalCombinations.
+type CredentialSet struct {
```
camgunz, line 304: If this is internal you can keep it private if you want

PR #4234, opened 2025-06-16
Add git metrics for cloning and scanning

dustin-decker:
### Description:
These metrics will give us some visibility into clone and scan success and failures. The clone failures are classified into several known categories which we can expand over time.

### Checklist:

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -64,16 +64,13 @@ type Git struct {
 	jobID              sources.JobID
 	sourceMetadataFunc func(file, email, commit, timestamp, repository string, line int64) *source_metadatapb.MetaData
 	verify             bool
-	metrics            metrics
+	metrics            metricsCollector
 	concurrency        *semaphore.Weighted
 	skipBinaries       bool
 	skipArchives       bool
+	repoCommitsScanned uint64 // Atomic counter for commits scanned in the current repo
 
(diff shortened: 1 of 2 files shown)


pkg/sources/git/git.go
```diff
@@ -627,7 +647,9 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 		if fullHash != lastCommitHash {
 			depth++
 			lastCommitHash = fullHash
-			atomic.AddUint64(&s.metrics.commitsScanned, 1)
+			s.metrics.RecordCommitScanned()
+			// Increment repo-specific commit counter
+			atomic.AddUint64(&s.repoCommitsScanned, 1)
```
camgunz, line 652: Is it a big hassle to move `repoCommitsScanned` into `metricsCollector`? It's a better separation of concerns and harder to cause sync problems if `RecordCommitScanned` is responsible for incrementing it.

dustin-decker: I think it could be complicated because `metricsCollector` is a package level metrics collector for prometheus metrics across all git scans whereas  `repoCommitsScanned` is stored at the Source level, and is created for every individual repo scanned.

camgunz: Oooh ok 👍🏻

camgunz (approved):

PR #4244, opened 2025-06-17
Update AzureDevops scanner

Kiblyn11:
### Description:
I observed one limitation with the current azure devops scanner.
It works on scanning by projects and this missed a few valid tokens that were not on a specific project I believe (ones that are used to push deliverables to visual studio marketplace). I observed how DevOps PAT tokens are verified with CLI `npx vsce verify-pat -p $TOKEN` and came with this new verification method.
As I'm not sure how it impact existing detector I chose to keep both methods in this PR.

I also extended keywords to improve result detection (in particular vsce tokens)

### Checklist:


pkg/detectors/azuredevopspersonalaccesstoken/azuredevopspersonalaccesstoken.go
```diff
@@ -43,6 +42,27 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 	for _, match := range matches {
 		resMatch := strings.TrimSpace(match[1])
+
```
kashifkhan0771, line 45: I’ve raised some concerns about this new approach with the internal team.

Kiblyn11: I wondered if this shouldn't be another detector at all, might be better choice

PR #4250, opened 2025-06-20
Fixed jira detector invalid domain issue

### LINKED ISSUE #4245: Add body check for Jira detector
Someone identified sometimes the Jira detector can get a 200 back, but the results are HTML or not valid.

We should add a body check to esnure the response matches the expected graphql response the verifier is looking for:

https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/jiratoken/v1/jiratoken.go#L138

kashifkhan0771:
### Description:
Fixes: #4245 
This PR adds a check for JIRA to decode the body in case of 200 OK response during verification to avoid invalid responses from the invalid domains.

### Checklist:

### DIFF
--- a/pkg/detectors/jiratoken/v1/jiratoken.go
+++ b/pkg/detectors/jiratoken/v1/jiratoken.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


thiggy1342 (approved):
Thank you for the quick fix!

nabeelalam (approved):
LGTM

PR #4251, opened 2025-06-20
Added LangSmith API Key detector

kashifkhan0771:
### Description:
This PR adds a new detector for LangSmith API Keys

### Tests:
Screenshot from 2025-06-20 16-40-26

### Checklist:


nabeelalam (approved):
Awesome! Thanks @kashifkhan0771

pkg/detectors/langsmith/langsmith.go
```diff
@@ -0,0 +1,100 @@
...
+
+		if verify {
+			client := s.client
+			if client == nil {
+				client = defaultClient
+			}
+
+			isVerified, verificationErr := verifyMatch(ctx, client, apiKey)
+			s1.Verified = isVerified
+			s1.SetVerificationError(verificationErr, apiKey)
```
shahzadhaider1, line 58: Since the keys could be either a personal api token or service keys, how different could these keys be in terms of permissions/access? And is it worth storing this info in extra data?

shahzadhaider1 (approved):
LGTM. 
One non-blocking comment.

pkg/detectors/langsmith/langsmith.go
kashifkhan0771: There is no fine grain access for these keys.

PR #4252, opened 2025-06-20
Salesforce OAuth2 Detector

### LINKED ISSUE #4182: New Detector for Salesforce
## Description

Salesforce OAuth2 credentials including the consumer and secret key pair are used to authenticate and authorize applications that need access to Salesforce data. These credentials are part of a connected app, which allows external services to securely interact with Salesforce APIs. The consumer key acts as a unique identifier for the app, while the consumer secret is used to verify the app’s authenticity when requesting an access token. Additionally, refresh tokens play a crucial role in maintaining access without requiring users to log in repeatedly. When an access token expires, a refresh token allows the application to request a new access token without user intervention. This ensures seamless and uninterrupted access to Salesforce resources while enhancing security by limiting the lifespan of access tokens. The current Salesforce detector is only scanning for access tokens that have a short TTL by default, but it is not scanning for either the long live consumer/secret key pair or the refresh token that are utilized to generate the access token in the first place. The two secret types should be able to be mostly identified by the regex patterns below.

Refresh Token: 5aep861[a-zA-Z0-9._=]+

Consumer Key: 3MVG9[a-zA-Z0-9+/=_.]+ 
Secret Key: [A-Za-z0-9+/=_.]{64} OR [0-9]{19}

### Preferred Solution

Expand the current Salesforce detector to identify the three credential types that make up the Salesforce OAuth2 workflow. The provided patterns should make for a decent baseline on the expected match.

shahzadhaider1:
### Description:
This pull request introduces a new detector for Salesforce OAuth2 Client Credentials. This detector is designed to find and verify the sensitive Consumer Key and Consumer Secret pairs that applications use for server-to-server API authentication with Salesforce.

### Implementation Details

The detector was built with a focus on accuracy and robust verification:

* **Detection Strategy:** The detector identifies a potential finding by looking for a trio of components:
    1.  A Salesforce domain (`*.my.salesforce.com`)
    2.  A Consumer Key (prefixed with `3MVG`)
    3.  A 64-character hexadecimal Consumer Secret or a string of 19 numbers

It intelligently creates findings for all possible combinations of these three components found within a given data chunk.

* **API Verification:** A comprehensive verification function validates found credentials using the standard **OAuth 2.0 Client Credentials Flow**. This verifier is built to correctly distinguishing between:
    * A valid credential pair (`200 OK`).
    * A definitively invalid key or secret (`invalid_client`, `invalid_client_id`).
    * A potentially valid but misconfigured application (`invalid_grant`).

* **Comprehensive Testing:** The detector is supported by a full test suite to ensure reliability:
    * **Pattern Tests:** A dedicated suite validates the combinatorial detection logic against numerous positive and negative test cases offline.
    * **Integration Tests:** Live tests confirm the verification logic against the Salesforce API, including mocked tests for edge cases like network timeouts and unexpected server errors.

This new detector provides robust coverage for a critical and common type of Salesforce secret. Ready for review.

### Checklist:

salesforce-oauth2-detector-tests


kashifkhan0771 (approved):
LGTM 🚀

pkg/detectors/salesforceoauth2/salesforceoauth2.go
```diff
@@ -0,0 +1,165 @@
...
+
+	switch errorResponse.Error {
+	case "invalid_client_id", "invalid_client":
+		// This definitively means the key is invalid
+		// Or the key is valid but the secret is wrong.
+		return false, nil
+	case "invalid_grant":
+		// This can mean the secret is wrong OR the user isn't configured with the app secret.
+		// We'll treat it as a VerificationError because the key might be valid but misconfigured.
+		return false, fmt.Errorf("verification failed: %s", errorResponse.ErrorDescription)
```
nabeelalam, line 181: @shahzadhaider1 in the `invalid_grant` case, are the credentials meaningful? I'm scratching my head about whether we should treat this as a verification error

pkg/detectors/salesforceoauth2/salesforceoauth2.go
```diff
@@ -0,0 +1,165 @@
...
+
+	authURL := fmt.Sprintf("https://%s/services/oauth2/token", domain)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, authURL, strings.NewReader(form.Encode()))
+	if err != nil {
+		return false, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	resp, err := client.Do(req)
+	if err != nil {
```
nabeelalam, line 133: We should handle the case where the hostname cannot be resolved (due to an invalid domain)

nabeelalam (changes requested):

pkg/detectors/salesforceoauth2/salesforceoauth2.go
shahzadhaider1: That's a great question. The `invalid_grant` error is returned by the API when the Connected App is misconfigured in Salesforce (for example, a "Run As" user hasn't been assigned for the Client Credentials Flow) in two distinct scenarios:
1. The Consumer Key is valid, but the Consumer Secret is incorrect.
2. Both the Key and Secret are valid. 

So, to answer your question: when we receive an `invalid_grant` error, the credentials are potentially very meaningful. We know for certain that the Consumer Key is valid, but we can't distinguish between a bad secret and a configuration problem on the server.

Because of this ambiguity, treating it as a VerificationError is the safest and most accurate approach, in my opinion. It correctly tells our user, "We found a valid key, but the full credential couldn't be confirmed. The issue is either a bad secret OR a server-side setup problem."

pkg/detectors/salesforceoauth2/salesforceoauth2.go
shahzadhaider1: done

pkg/detectors/salesforceoauth2/salesforceoauth2.go
```diff
@@ -0,0 +1,185 @@
...
+				}
+
+				if verify {
+					isVerified, verificationErr := s.verifyMatch(ctx, s.getClient(), domain, key, secret)
+					s1.Verified = isVerified
+					if verificationErr != nil {
+						if errors.Is(verificationErr, errNoHost) {
+							invalidHosts.Set(domain, struct{}{})
+							continue
+						}
```
abmussani: If the host is invalid, all the combinations of key and secret should also be invalidate like [here](https://github.com/trufflesecurity/trufflehog/blob/8fb476f810a1eb1454aac4b749ff91e3097af6f5/pkg/detectors/azure_entra/refreshtoken/refreshtoken.go#L136).

shahzadhaider1: done

nabeelalam (approved):
Thanks @shahzadhaider1

PR #4261, opened 2025-06-24
[Detector]-Detector for tableau personal access token

SyedAliHamad:
### Description:
This PR adds  new detector for Tableau Personal Access Tokens

### Checklist:


pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,377 @@
...
+		"pat",
+		"token_name",
+		"token_secret",
+		"access_token",
+		"tableau_personal_access_token_name",
+		"tableau_personal_access_token",
+		"tableau",
+		"online.tableau.com",
+	}
+}
```
shahzadhaider1: It includes quite a few generic keywords. While this increases coverage, it might also lead to a lot of unrelated or noisy matches, especially from non-Tableau sources. To tighten this up, I’d suggest narrowing the focus by using keywords that are more specific to Tableau. That way, the detector will be both accurate and relevant to the target use case.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,377 @@
...
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+func (Scanner) CloudEndpoint() string { return "" }
+
+var (
+	defaultClient = common.SaneHttpClient()
+
+	// Pattern for Personal Access Token Names - Updated to handle JSON
+	tokenNamePat       = regexp.MustCompile(`(?i)\b(?:personal[_-]?access[_-]?token[_-]?name|pat[_-]?name|token[_-]?name|access[_-]?token[_-]?name|api[_-]?token[_-]?name)[_\s]*[:=]\s*["\']?([a-zA-Z0-9_-]{3,50})["\']?`)
+	quotedTokenNamePat = regexp.MustCompile(`(?i)["'](?:personal[_-]?access[_-]?token[_-]?name|pat[_-]?name|token[_-]?name|access[_-]?token[_-]?name|api[_-]?token[_-]?name)["']\s*:\s*["']([a-zA-Z0-9_-]{3,50})["']`)
```
shahzadhaider1: Let's discuss this

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,377 @@
...
+		for tokenSecret := range tokenSecrets {
+			for _, endpoint := range endpointsToTest {
+				result := detectors.Result{
+					DetectorType: detectorspb.DetectorType_Tableau,
+					Raw:          []byte(tokenName),
+					RawV2:        []byte(fmt.Sprintf("%s:%s:%s", tokenName, tokenSecret, endpoint)),
+					ExtraData: map[string]string{
+						"token_name":      tokenName,
+						"token_secret":    tokenSecret,
+						"endpoint":        endpoint,
```
shahzadhaider1: we added these values in RawV2, right? I do not think we need to store them again.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,377 @@
...
+
+// verifyTableauPAT verifies a Tableau Personal Access Token by attempting authentication
+func verifyTableauPAT(ctx context.Context, client *http.Client, tokenName, tokenSecret, endpoint string) (bool, map[string]string, error) {
+	// Build the verification URL
+	var verifyURL string
+	if strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://") {
+		verifyURL = endpoint + "/api/3.26/auth/signin"
+	} else {
+		verifyURL = "https://" + endpoint + "/api/3.26/auth/signin"
+	}
```
shahzadhaider1: this is fine but we can design our regex in such a way that it excludes the https:// part and only includes the instance/domain

shahzadhaider1 (changes requested):
Great job on putting together your first detector! 
Added some feedback.

pkg/detectors/tableau/tableau.go
amanfcp: I agree. I think adding just the tableau keyword should be sufficient.

SyedAliHamad:
**Description:**
This pull request introduces a new detector for Tableau Personal Access Tokens (PATs). This detector is designed to find and verify the sensitive token name and token secret pairs that applications use for programmatic authentication with Tableau Server and Tableau Online APIs.

**Detection Strategy:** The detector identifies potential findings by looking for a combination of components:

Token names using flexible patterns (tableau_token_name, pat_name, personal_access_token_name, etc.)
Token secrets following Tableau's specific format (base64==:32-char-alphanumeric)
Tableau server URLs (*.online.tableau.com)

It intelligently creates findings for all possible combinations of token names and secrets found within a given data chunk, testing against discovered URLs.

**Format Validation:** The detector includes comprehensive secret format validation, ensuring tokens match Tableau's specific structure (base64-encoded prefix with colon separator and alphanumeric suffix) to minimize false positives.

**API Verification:** A robust verification function validates found credentials using Tableau's REST API authentication endpoint (/api/3.26/auth/signin). The verifier correctly distinguishes between:

Valid credential pairs (200 OK with authentication token)
Invalid credentials (401 Unauthorized)
Malformed requests (400 Bad Request)
Network/endpoint errors (gracefully handled without verification errors)


Comprehensive Testing: The detector is supported by a full test suite ensuring reliability:

**Pattern Tests:** Validates regex patterns and format validation logic against various credential formats
**Integration Tests:** Live tests confirm verification logic against real Tableau APIs using active and inactive test credentials

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+
+// Keywords are used for efficiently pre-filtering chunks.
+func (s Scanner) Keywords() []string {
+	return []string{
+		"personal_access_token",
+		"personalAccessToken",
+		"pat",
+		"tableau",
+	}
+}
```
shahzadhaider1: Let's only keep `tableau` because `pat`, `personalAccessToken`, etc., are too common and will attract a lot of noise.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+func (Scanner) CloudEndpoint() string { return "" }
+
+var (
+	defaultClient = common.SaneHttpClient()
+
+	// Pattern for Personal Access Token Names - Updated to handle JSON
+	tokenNamePat       = regexp.MustCompile(`(?i)\b(?:(?:tableau)[_-]?)?(?:token[_-]?name|pat[_-]?name|personal[_-]?access[_-]?token[_-]?name|access[_-]?token[_-]?name|pod[_-]?name|personal[_-]?token[_-]?name|api[_-]?token[_-]?name)[_\s]*[:=]\s*["']?([a-zA-Z0-9_-]{3,50})["']?`)
```
kashifkhan0771: This can be simplified to
```go
detectors.PrefixRegex([]string{"tableau", "token_name", "pat_name"}) + `\b([a-zA-Z0-9_-]{3,50})\b`)
```

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+	quotedTokenNamePat = regexp.MustCompile(`(?i)["'](?:tableau[_-]?)?(?:token[_-]?name|pat[_-]?name|personal[_-]?access[_-]?token[_-]?name|access[_-]?token[_-]?name|pod[_-]?name|personal[_-]?token[_-]?name|api[_-]?token[_-]?name)["']\s*:\s*["']([a-zA-Z0-9_-]{3,50})["']`)
```
kashifkhan0771: Why do we detect token names enclosed in quotes? IMO this is not required.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+
+	// Pattern for Personal Access Token Secrets - Updated to handle JSON
+	tokenSecretPat = regexp.MustCompile(`\b([A-Za-z0-9+/]{22}==:[A-Za-z0-9]{32})\b`)
+
+	// Pattern for Tableau Server URLs
+	tableauURLPat = regexp.MustCompile(`(?i)(?:https?://)?([a-zA-Z0-9\-]+\.online\.tableau\.com)(?:/.*)?`)
```
kashifkhan0771: This can be simplified to:
```go
\b([a-zA-Z0-9\-]+\.online\.tableau\.com)\b
```

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+func (s Scanner) Keywords() []string {
+	return []string{
+		"personal_access_token",
```
kashifkhan0771: `personal_access_token` is not a good keyword for any detector. The keywords should be detector specific.
IMO following should be good:
- tableau
- online.tableau.com

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+		// Check if the input contains any tableau.com reference (even invalid ones)
+		// If it does, don't use default endpoints - it means they provided a URL but it was invalid
+		if strings.Contains(strings.ToLower(dataStr), "online.tableau.com") {
+			// Invalid tableau URL found, don't use defaults
+			return results, nil
+		}
+
+		// No URLs found at all, use configured endpoints
+		endpointsToTest = append(endpointsToTest, s.Endpoints("")...)
+	}
```
kashifkhan0771: This implementation does not correctly follow the `EndpointCustomizer` interface implementation as in other detectors. We should always extract endpoints from the data chunk first, and then use `s.Endpoints(foundEndpoints...)` to loop through the endpoints we need to verify. The engine will automatically handle whether to use the found endpoints or the cloud ones.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+
+	// Prepare results slice with estimated capacity
+	results = make([]detectors.Result, 0, len(tokenNames)*len(tokenSecrets)*len(endpointsToTest))
```
kashifkhan0771: IMO this is not required.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+					},
+				}
+
+				if verify {
+					client := s.getClient()
+					isVerified, extraData, verificationErr := verifyTableauPAT(ctx, client, tokenName, tokenSecret, endpoint)
+					result.Verified = isVerified
+
+					// Merge verification extra data
+					maps.Copy(result.ExtraData, extraData)
```
kashifkhan0771, line 98: This is good 🚀

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+
+					result.SetVerificationError(verificationErr, fmt.Sprintf("%s:%s@%s", tokenName, tokenSecret, endpoint))
```
kashifkhan0771: `s1.SetVerificationError(verificationErr, tokenName, tokenSecret, endpoint)`

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+	for _, match := range quotedTokenNamePat.FindAllStringSubmatch(data, -1) {
+		if len(match) >= 2 {
+			name := strings.TrimSpace(match[1])
+			if name != "" && len(name) >= 3 { // Minimum reasonable token name length
+				names[name] = struct{}{}
+			}
+		}
+	}
+	return names
+}
```
kashifkhan0771: If we use the regex I suggested above, this logic would no longer be necessary.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+	if len(tokenPart) < 20 {
+		return false
+	}
+
+	// Basic pattern validation
+	base64Pattern := regexp.MustCompile(`^[A-Za-z0-9+/]+={0,2}$`)
+	tokenPattern := regexp.MustCompile(`^[A-Za-z0-9]+$`)
+
+	return base64Pattern.MatchString(base64Part) && tokenPattern.MatchString(tokenPart)
+}
```
kashifkhan0771: I believe this is not required. The regex for secret is quite unique enough having `==:` in the middle. Any match found should be directly used in verification rather than manual filtration.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+		if len(match) >= 2 {
+			url := strings.TrimSpace(match[1]) // This now captures the full URL including .online.tableau.com
+			if url != "" {
+				urls[url] = struct{}{}
+			}
+		}
+	}
+
+	return urls
+}
```
kashifkhan0771:
```go
for _, match := range tableauURLPat.FindAllStringSubmatch(dataStr, -1) {
    urls = append(urls, match[1])
}
```

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+				extraData["auth_token_received"] = "true"
+				extraData["site_id"] = authResp.Credentials.Site.ID
+				extraData["user_id"] = authResp.Credentials.User.ID
+				extraData["verification_status"] = "valid"
+				return true, extraData, nil
+			}
+		}
+		// Fallback success on 200
+		extraData["verification_status"] = "valid"
+		return true, extraData, nil
```
kashifkhan0771: IMO we should reverse the logic for better error handling
```go
var authResp TableauAuthResponse
if err := json.Unmarshal(bodyBytes, &authResp); err != nil {
	return true, nil, err
}

if token := authResp.Credentials.Token; token != "" {
	extraData["auth_token_received"] = "true"
	extraData["site_id"] = authResp.Credentials.Site.ID
	extraData["user_id"] = authResp.Credentials.User.ID
	extraData["verification_status"] = "valid"
	
	return true, extraData, nil
}

extraData["verification_status"] = "valid"
return true, extraData, nil
```

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,350 @@
...
+		return false, extraData, nil
+
+	case http.StatusBadRequest:
+		extraData["verification_status"] = "invalid_format"
+		return false, extraData, nil
+
+	case http.StatusForbidden:
+		extraData["verification_status"] = "insufficient_permissions"
+		return false, extraData, nil
+
```
kashifkhan0771: All of these statuses can be handled together in one case and it's not necessary to add verification status in the extraData

kashifkhan0771 (changes requested):

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,282 @@
...
+	// If no names or secrets found, return empty results
+	if len(tokenNames) == 0 || len(tokenSecrets) == 0 {
+		return results, nil
+	}
+
+	// Use maps to deduplicate endpoints
+	var uniqueEndpoints = make(map[string]struct{})
+
+	// Add endpoints to the list
+	s.UseFoundEndpoints(true)
```
kashifkhan0771: This is managed by the engine based on user input, so it should not always be set to true.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,282 @@
...
+	// Process each combination of token name, token secret, and endpoint
+	for _, tokenName := range tokenNames {
+		for _, tokenSecret := range tokenSecrets {
+			for endpoint := range uniqueEndpoints {
+				result := detectors.Result{
+					DetectorType: detectorspb.DetectorType_Tableau,
+					Raw:          []byte(tokenName),
+					RawV2:        []byte(fmt.Sprintf("%s:%s:%s", tokenName, tokenSecret, endpoint)),
+					ExtraData: map[string]string{
+						"credential_type": "personal_access_token",
```
kashifkhan0771: IMO, this key is redundant as it remains constant across all results.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,282 @@
...
+	}()
+
+	// Read the response
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to read response body: %v", err)
+	}
+
+	// Add HTTP status to metadata
+	extraData["http_status"] = fmt.Sprintf("%d", resp.StatusCode)
```
kashifkhan0771: We do not inform users about the API response status from the verification API.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,282 @@
...
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+func (Scanner) CloudEndpoint() string { return "" }
+
+var (
+	defaultClient = common.SaneHttpClient()
+
+	// Simplified token name pattern using PrefixRegex
+	tokenNamePat = regexp.MustCompile(detectors.PrefixRegex([]string{"tableau", "token", "pat"}) + `(?:name|_name)\s*[:=]\s*["']?([a-zA-Z0-9_-]{3,50})["']?`)
```
kashifkhan0771: It's not necessary to add `name` or `_name`. If we encounter keys like:

* `tableau_token=`
* `token_name=`
* `pat_name=`

the prefix regex will still handle them correctly, since keywords like `tableau`, `token`, or `pat` are present in the key names.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// extractTokenNames finds all potential token names in the data
+func extractTokenNames(data string) []string {
+	var names []string
+
+	for _, match := range tokenNamePat.FindAllStringSubmatch(data, -1) {
+		if len(match) >= 2 {
+			name := strings.TrimSpace(match[1])
+			if name != "" && len(name) >= 3 {
```
kashifkhan0771: This validation check is redundant. The regex make sure to pick a non-empty string with minimum three characters.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// extractTokenSecrets finds all potential token secrets in the data
+func extractTokenSecrets(data string) []string {
+	var secrets []string
+
+	for _, match := range tokenSecretPat.FindAllStringSubmatch(data, -1) {
+		if len(match) >= 2 {
+			secret := strings.TrimSpace(match[1])
+			if secret != "" {
```
kashifkhan0771: This validation is redundant. The secret regex make sure to pick a non-empty string.

kashifkhan0771 (changes requested):

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,272 @@
...
+			extraData["site_id"] = authResp.Credentials.Site.ID
+			extraData["user_id"] = authResp.Credentials.User.ID
+			return true, extraData, nil
+		}
+
+		// Fallback success on 200
+		return true, extraData, nil
+
+	case http.StatusUnauthorized, http.StatusBadRequest, http.StatusForbidden:
+		return false, extraData, nil
```
shahzadhaider1, line 250: Could you please confirm the expected behavior of this API?

Typically, when an authenticated API is accessed with incorrect credentials, it returns a 401 Unauthorized. On the other hand, if the credentials are valid but do not have the necessary permissions to access a specific resource, the response is usually 403 Forbidden, indicating that the credentials are valid, but access is denied.

In our current implementation, we're treating both 401 and 403 as unverified (i.e., returning false for the verification status). I just want to confirm if this is the correct approach or if we should be handling them differently.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,272 @@
...
+
+		if token := authResp.Credentials.Token; token != "" {
+			extraData["auth_token_received"] = "true"
+			extraData["site_id"] = authResp.Credentials.Site.ID
+			extraData["user_id"] = authResp.Credentials.User.ID
+			return true, extraData, nil
+		}
+
+		// Fallback success on 200
+		return true, extraData, nil
```
shahzadhaider1: Minor (non-blocking) observation, it looks like we’re unmarshaling the response primarily to populate `extraData` with metadata like `site_id` and `user_id`. Is this information required downstream, or is it just nice-to-have?

Also, in the case where we receive a 200 OK but the token is missing or empty, we still treat it as a successful verification. Is that intentional? 
Should the presence of a valid token in the API response be a stronger requirement for verification success?

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,272 @@
...
+	case http.StatusUnauthorized, http.StatusBadRequest, http.StatusForbidden:
+		return false, extraData, nil
+
+	default:
+		return false, extraData, fmt.Errorf("unexpected HTTP response status %d", resp.StatusCode)
+	}
+}
+
+func (s Scanner) Type() detectorspb.DetectorType {
+	return detectorspb.DetectorType_Tableau
```
shahzadhaider1: Also, while we're on this, should we consider updating the detector name from `DetectorType_Tableau` to something more descriptive, like `DetectorType_TableauPAT` or `DetectorType_TableauPersonalAccessToken`?

This would help avoid situations like the Twitch detector, where the name was just `Twitch`, and we later realized we needed to clarify it as `Twitch Client Credentials`.

Of course, this depends on the types of credentials Tableau supports. If there are multiple credential types (e.g., PATs, access tokens, refresh tokens), having a more specific detector name would definitely help with clarity and future extensibility.

shahzadhaider1 (commented):
Thanks for addressing all the feedback. We are close.

pkg/detectors/tableau/tableau.go
SyedAliHamad: Great idea, you're right with the recent suggestions from twitch i think even if tableau only supports PATs even then it's better to reflect that in the name.

pkg/detectors/tableau/tableau.go
SyedAliHamad: Yes, the criteria for success is the 200 response. With the current implementation, we won’t ever receive a 200 without a token. As for site_id, user_id, and even token, you’re right, they were added as extra data and aren’t strictly required downstream, just nice-to-have. **Updated**

pkg/detectors/tableau/tableau.go
SyedAliHamad: Great observation. You’re right that typically 403 is handled differently, but since this is just a sign-in API, I don’t believe we’d ever encounter a scenario where valid credentials result in a 403. Will have another look at the docs.

shahzadhaider1 (approved):

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,257 @@
...
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+func (Scanner) CloudEndpoint() string { return "" }
+
+var (
+	defaultClient = common.SaneHttpClient()
+
+	// Simplified token name pattern using PrefixRegex
+	tokenNamePat = regexp.MustCompile(detectors.PrefixRegex([]string{"token_name", "pat_name", "tableau_pat_name", "tableau_token", "pat", "token", "name"}) + `\s*["']?([a-zA-Z0-9_-]{3,50})["']?`)
```
kashifkhan0771:
```
([a-zA-Z][a-zA-Z0-9_-]{2,50})
```
This regex ensures that the first character is always a letter, so there's no need for additional validation checks later.

shahzadhaider1:
Apart from Kashif's feedback, looks good.

pkg/detectors/tableau/tableau.go
kashifkhan0771: Including `\s*["']?` and `["']?` is unnecessary.

pkg/detectors/tableau/tableau.go
```diff
@@ -0,0 +1,257 @@
...
+	return results, nil
+}
+
+// extractTokenNames finds all potential token names in the data
+func extractTokenNames(data string) []string {
+	var names []string
+	for _, match := range tokenNamePat.FindAllStringSubmatch(data, -1) {
+		if len(match) >= 2 {
+			name := strings.TrimSpace(match[1])
+			if name[0] != '_' {
```
kashifkhan0771: With the regex I provided above this will no longer be required.

pkg/detectors/tableau/tableau_test.go
```diff
@@ -0,0 +1,165 @@
...
+	validPATName     = "test-token-6"
+	validPATSecret   = "YMqNfVWiTSa0QgpoJ9GpCw==:KTamjuKrXF5gVETjIJafBBvP8Ctj5aJ3"
+	invalidPATSecret = "invalid-secret-format"
+	// Tableau Online endpoints for testing
+	validTableauURL    = "prod-ansouthgest-a.online.tableau.com"
+	valid2ndTableauURL = "prod.online.tableau.com"
+	invalidTableauURL  = "prod.tabeau.com"
+)
+
+func TestTableau_Pattern(t *testing.T) {
```
kashifkhan0771, line 23: Excellent work on test coverage ❤️

kashifkhan0771 (changes requested):
Some minor comments. Rest looks good to me.

kashifkhan0771 (approved):

kashifkhan0771:
Thank you for patiently addressing all my comments. Really appreciate it!

amanfcp (approved):

PR #3419, opened 2024-10-15
fix: throws error if dir not found

sahil9001:
### Description:
This PR relates to #3005 . If the directory is not found, the process returns an error message and exits.

Screenshot 2024-10-16 at 1 42 58 AM

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

PR #4262, opened 2025-06-24
Disable s3 filter test

rosecodym:
### Description:
There have been some changes to our AWS test environment that are causing this test to fail. This PR disables the test until the environment stabilizes and we decide on a path forward.

### Checklist:


casey-tran (approved):

rosecodym:
I believe the test failure was caused by #3419. I am currently following up internally to determine whether that PR should be reverted or the test should be modified.

PR #4266, opened 2025-06-26
Deprecated AirTable API Key detector

kashifkhan0771:
### Description:
Airtable deprecated the API Keys on 1st Feb 2024. 

### Checklist:


kashifkhan0771:
From Airtable docs:
```text
On January 18th, 2023, we began the deprecation period of user API keys. That period has now ended. Since this was a major change to the Airtable API, the deprecation period lasted for 12 months and ended on Feb 1, 2024. This means that:

API Keys can no longer access the Airtable API.

Webhooks created with user API keys will not expire, but can no longer be created.

All users should migrate to Personal Access Tokens for individual use and OAuth for third-party integrations.
```

Ref: https://support.airtable.com/docs/creating-personal-access-tokens

shahzadhaider1 (approved):

pkg/engine/defaults/defaults.go
```diff
@@ -863,7 +862,7 @@ func buildDetectorList() []detectors.Detector {
 		&airbrakeprojectkey.Scanner{},
 		&airbrakeuserkey.Scanner{},
 		&airship.Scanner{},
-		&airtableapikey.Scanner{},
+		// &airtableapikey.Scanner{}, // AirTable depreacted the API Keys
```
nabeelalam: We can remove this line

kashifkhan0771: We usually do not remove this from defaults but just comment out to keep track of old deprecated detectors as well.

nabeelalam: I was aware that detectors that are commented out in `defaults.go` have been disabled temporarily but not deleted or deprecated. I believe most deprecated detectors have been removed from the `defaults.go` file.
Since the `detectors.proto` file already keeps track of deprecated detectors, it should be safe to remove this line.

kashifkhan0771: Gotcha! I removed it from the engine defaults now

amanfcp (approved):
LGTM

PR #4270, opened 2025-06-26
(feat) Implement SourceUnitEnumChunker for S3 source

abmussani:
### Description:
This PR implements SourceUnitEnumChunker for S3 source. Each Object in the bucket will be considered as Unit.

### Checklist:


rosecodym (changes requested):
requesting changes until we have an internal discussion about unit granularity

PR #4276, opened 2025-06-27
Added support for scanning images from the docker daemon

### LINKED ISSUE #4275: Add support for scanning images from the Docker daemon

## Description

Ideally container images can be scanned for leaked secrets before being pushed to a registry. This can be done locally or in CI/CD pipelines. Currently, Trufflehog only supports scanning from remote registries, which is not helpful in the above stated goal, or from images that have been saved as tarballs, which adds a additional and unnecessary step. Given that the docker daemon would already be present at the time of building an image, we should be able to connect directly to the daemon to scan newly built images.

### Preferred Solution

The docker scanner should support connecting directly to the docker daemon to allow scanning newly built images. It should be invoked like:

```sh
trufflehog docker --image docker://some_image:some_tag
```

saghaulor:
- Now you can scan an image directly after building it with `docker build` by using the `docker://` prefix. This is ideal for local development and CI/CD pipelines that want to ensure images do not contain leaked secrets before pushing to an image registry.
- Resolves #4275

### Description:

### Checklist:

### DIFF
--- a/pkg/sources/docker/docker.go
+++ b/pkg/sources/docker/docker.go
@@ -13,6 +13,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/daemon"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	gzip "github.com/klauspost/pgzip"
@@ -179,8 +180,8 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 func (s *Source) processImage(ctx context.Context, image string) (imageInfo, error) {
 	var (
 		imgInfo   imageInfo
-		hasDigest bool
 		imageName name.Reference
+		err       error
 	)
 
 	remoteOpts, err := s.remoteOpts()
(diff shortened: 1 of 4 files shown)


pkg/sources/docker/docker.go
```diff
@@ -219,6 +222,29 @@ func (s *Source) processImage(ctx context.Context, image string) (imageInfo, err
 	return imgInfo, nil
 }
 
+// extractImageNameTagDigest parses the provided Docker image string and returns a name.Reference
+// representing either the image's tag or digest, and any error encountered during parsing.
+// The function also updates the provided imageInfo struct with the base name and tag.
+func (*Source) extractImageNameTagDigest(image string, _imgInfo imageInfo) (name.Reference, error) {
```
rosecodym: If I'm reading this right, `imageInfo` is passed by value, so the changes this function makes (to set `base` and `tag`) won't be propagated back to the caller. It looks like they're only ever used for logging and chunk metadata, so this wouldn't have caused any automated tests to fail. Did you see this in the log output or the chunk metadata?

Fortunately, it looks like the passed-in `imageInfo` struct always has its zero value, so I think you can remove that parameter and return a brand-new `imageInfo` struct from this function.

rosecodym (changes requested):
This looks quite convenient - thank you for doing it! I do have one comment about a change that I think will cause some metadata to be lost, which I've added inline.

pkg/sources/docker/docker.go
saghaulor: Thanks for the feedback. I had a feeling I had made this mistake. I'm still learning the ropes of Go. I'll make this change in a few minutes.

rosecodym (approved):
Thanks for this, and in particular for the tests! I see that very little varies between them, so if you want, you could switch them over to the table-driven style that we use in most of the rest of the codebase. It would look something like:
```
testCases := []struct {
    name string
    image string
}{
    {
        name: "image name only",
        image: "docker://trufflesecurity/secrets"
    },
    {
        name: "tag",
        image: "docker://trufflesecurity/secrets:latest"
    },
    // and one for digest
}

for _, tt := range testCases {
        // common test logic goes here using tt.image
    }
}
```

If you want to tie things off here, though, that's fine too.

PR #4277, opened 2025-06-27
[chore] Write --help to stdout

mcastorina:
### Description:
The [GNU CLI Guidelines](https://www.gnu.org/prep/standards/html_node/_002d_002dhelp.html) suggest writing `--help` to stdout. The default for `kingpin` is to use stderr, so this PR updates it.

### Checklist:


samdatkins (approved):
❤️

PR #4282, opened 2025-07-01
[Update] Dwolla Detector Updated

nabeelalam:
### Description:
The ID and secret patterns are the same. This update adds a check to avoid verification of identical credential pairs.
Also removes duplicate credentials and adds indeterminate verification errors in results.

### Checklist:


shahzadhaider1 (approved):

kashifkhan0771 (approved):
❤️

pkg/detectors/dwolla/dwolla.go
```diff
@@ -84,6 +84,37 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, id, secret string) (bool, error) {
+	data := fmt.Sprintf("%s:%s", id, secret)
+	encoded := b64.StdEncoding.EncodeToString([]byte(data))
+	payload := strings.NewReader("grant_type=client_credentials")
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, "https://api-sandbox.dwolla.com/token", payload)
```
amanfcp, line 92: Are production credentials acceptable in sandbox environment as well? This is sometimes the case but if it is not, we should add support for production as well. 

[See this](https://developers.dwolla.com/docs/balance/api-reference/api-fundamentals/making-requests-and-authentication#api-host)

amanfcp (changes requested):
Looks good to me. Requesting changes as one comment I've made seems critical

pkg/detectors/dwolla/dwolla.go
nabeelalam: Good observation! The same thought crossed my mind but I didn't change the verification logic and endpoint since I don't know if there's a way to test production creds on the free account, which only has the sandbox environment AFAIK.

IMO we should take note of this and update this separately when we're able to test it.

amanfcp (approved):

PR #4283, opened 2025-07-02
Simplified Gitlab Enumeration

kashifkhan0771:
### Description:
Currently,  theGitLab source enumerates projects via a complicated, multi-stage system of user queries, group queries, and deduplication. This PR simplify the approach by using `list-all-projects` API Call.

### Checklist:


kashifkhan0771:
I performed the following testing after making the changes:

* Created one private project within a group
* Created one private project outside of any group
* Created one public project

When scanning, all three projects were correctly listed, confirming that the scan includes public, group-based, and standalone private repositories.

If there's any other scenario you'd like me to test, please feel free to let me know.

rosecodym (changes requested):
requesting changes pending the outcome of an internal conversation about flagging this

pkg/sources/gitlab/gitlab.go
```diff
@@ -616,6 +631,107 @@ func (s *Source) getAllProjectRepos(
...
+
+	// for non gitlab.com instances, include all available projects (public + membership).
+	if s.url != gitlabBaseURL {
+		projectQueryOptions.Membership = gitlab.Ptr(false)
+	}
+
+	ctx.Logger().Info("starting projects enumeration",
+		"list_options", listOpts,
+		"all_available", *projectQueryOptions.Membership)
+	gitlabGroupsEnumerated.WithLabelValues(s.name).Set(0)
```
rosecodym: Would you mind placing this metric reset next to the other one? I think it would make things a little clearer to the reader.

rosecodym (approved):
One small non-blocking request but otherwise this looks great! Thanks for doing it!

pkg/sources/gitlab/gitlab.go
kashifkhan0771: That's not necessary, as we're not enumerating groups in the simplified approach. I'll remove it. I must have overlooked this during the simplification process.

pkg/sources/gitlab/gitlab.go
```diff
@@ -616,6 +631,106 @@ func (s *Source) getAllProjectRepos(
...
+func (s *Source) getAllProjectReposV2(
+	ctx context.Context,
+	apiClient *gitlab.Client,
+	ignoreRepo func(string) bool,
+	reporter sources.UnitReporter,
+) error {
+	gitlabReposEnumerated.WithLabelValues(s.name).Set(0)
+
+	// record the projectsWithNamespace for logging.
+	var projectsWithNamespace []string
```
shahzadhaider1, line 645: If this is only for logging the count, then we can make the memory usage efficient with something like:
```
var totalProjects int
...
totalProjects++
...
"count", totalProjects

```

PR #4290, opened 2025-07-03
[Update] Dovico Detector Updated

nabeelalam:
### Description:
The ID and secret patterns are the same. This update adds a check to avoid verification of identical credential pairs.
Also removes duplicate credentials and adds indeterminate verification errors in results.

### Checklist:

### DIFF
--- a/pkg/detectors/dovico/dovico.go
+++ b/pkg/detectors/dovico/dovico.go
@@ -3,8 +3,8 @@ package dovico
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
-	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
 
@@ -14,6 +14,7 @@ import (
 )
 
 type Scanner struct {
+	client *http.Client
 	detectors.DefaultMultiPartCredentialProvider
 }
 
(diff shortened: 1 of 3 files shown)


shahzadhaider1 (approved):

pkg/detectors/dovico/dovico.go
```diff
@@ -34,37 +35,43 @@ func (s Scanner) Keywords() []string {
...
+	for key := range uniqueKeys {
+		for userKey := range uniqueUserKeys {
+			if key == userKey {
+				continue // Skip if ID and secret are the same.
+			}
 
 			s1 := detectors.Result{
 				DetectorType: detectorspb.DetectorType_Dovico,
-				Raw:          []byte(resMatch),
+				Raw:          []byte(key),
```
abmussani, line 67: Shouldn't RAW contain both key + userkey ? or may be set `RawV2` for both keys? Otherwise, it will be difficult to find which combination worked.

kashifkhan0771 (approved):
Apart from @abmussani comment - It looks good to me

pkg/detectors/dovico/dovico.go
nabeelalam: Could changing the `Raw` value have an impact on existing usage of this detector? Same question about adding a `RawV2` value?
I can make those updates if this is a safe option.

abmussani: umm.... Considering only OSS, I could not think of any impact. @trufflesteeeve what are your thoughts.

nabeelalam: @abmussani I added a `RawV2` value

abmussani (approved):

PR #4293, opened 2025-07-07
Oss 248 refactor y alphabet detectors

SyedAliHamad:
### Checklist:


pkg/detectors/yandex/yandex.go
```diff
@@ -63,6 +57,34 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		return true, nil
+	case http.StatusPaymentRequired:
+		return true, fmt.Errorf("blocked due to billing issue/quota exceeded: %d", res.StatusCode)
```
kashifkhan0771: If we're certain that the API returns this error when the key is valid but blocked due to billing issues or quota limits, then I don't think we should surface it as an error to the user. It's not a failure to verify the key itself.

pkg/detectors/youneedabudget/youneedabudget.go
```diff
@@ -20,14 +21,15 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"youneedabudget"}) + `\b([0-9a-f]{64})\b`)
+	// YNAB Personal Access Tokens are 64-character hexadecimal strings
+	// Based on documentation: access tokens are used with Bearer authentication
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"youneedabudget", "ynab"}) + `["']?([A-Za-z0-9_-]{43,44})["']?`)
```
kashifkhan0771: Are we sure the old pattern keys are no longer valid? Also, we don’t need `["']` in the regex - let’s just capture the exact pattern.

pkg/detectors/yousign/yousign.go
```diff
@@ -85,6 +62,58 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		return isVerified, err
+	}
+
+	// Try V2 Staging as final fallback
+	return tryEndpoint(ctx, client, token, STAGING_URL, "/users")
+}
+
+func tryEndpoint(ctx context.Context, client *http.Client, token, baseURL, endpoint string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, baseURL+endpoint, http.NoBody)
+	if err != nil {
```
kashifkhan0771, line 78: 🚀

pkg/detectors/youtubeapikey/youtubeapikey.go
```diff
@@ -42,36 +45,111 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		}
+		return false, nil
+	case http.StatusUnauthorized:
+		// 401: Missing or invalid authentication credentials
+		return false, fmt.Errorf("unauthorized: %d", res.StatusCode)
+	case http.StatusForbidden:
+		// 403: API key is valid but access is forbidden (quota exceeded, API not enabled, etc.)
+		body, _ := io.ReadAll(res.Body)
+		if strings.Contains(string(body), "quotaExceeded") {
+			return true, fmt.Errorf("quota exceeded (valid key): %d", res.StatusCode)
```
kashifkhan0771: If we're returning true, that means the key was successfully verified. So we shouldn't return a verification error in that case.

kashifkhan0771 (changes requested):
Most of the code, looks good to me. I added few comments.

pkg/detectors/yandex/yandex.go
amanfcp: +1 to ^
We can consider it as a success iff we're certain that this is only due to billing issues.

amanfcp:
Contributes to #4051

PR #4295, opened 2025-07-08
Salesforce Refresh Token Detector

### LINKED ISSUE #4182: New Detector for Salesforce

shahzadhaider1:
### Description:
This PR adds a new detector for Salesforce OAuth 2.0 refresh tokens. Refresh tokens are long-lived credentials that allow applications to obtain new access tokens without requiring user interaction, making them critical security assets that need to be detected in code repositories.

## What's Added
- **New Detector**: `salesforcerefreshtoken` package with full detection and verification capabilities
- **Pattern Recognition**: Detects Salesforce refresh tokens (starting with `5AEP861`), consumer keys (starting with `3MVG9`), and consumer secrets
- **Verification**: Validates refresh tokens using Salesforce's OAuth 2.0 introspection endpoint
- **Comprehensive Tests**: Unit tests for pattern matching and integration tests for verification

### Detection Patterns
- **Refresh Token**: `(?i)\b(5AEP861[a-zA-Z0-9._=]{80,})\b`
- **Consumer Key**: `\b(3MVG9[0-9a-zA-Z._+/=]{80,251})`
- **Consumer Secret**: Uses `PrefixRegex` with keywords `["salesforce", "consumer", "secret"]` + `\b([A-Za-z0-9+/=.]{64}|[0-9]{19})\b`

### Verification Strategy
The detector uses Salesforce's OAuth 2.0 introspection endpoint (`/services/oauth2/introspect`) to validate refresh tokens:
- **200 OK**: Parses JSON response to check if token is `active`
- **400 Bad Request**: Invalid token format or expired token
- **401 Unauthorized**: Invalid client credentials
- **Other**: Unexpected errors are properly handled

### Triple Component Requirement
The detector requires all three components (refresh token, consumer key, consumer secret) to be present:
- Creates a cartesian product of all found combinations
- Returns early if any component is missing (performance optimization)
- Handles multiple instances of each component type

### Pattern Tests
- ✅ Simple case: one of each component
- ✅ Combinatorial cases: multiple keys/tokens/secrets
- ✅ Negative cases: invalid formats, missing components
- ✅ Edge cases: boundary conditions for pattern matching

### Integration Tests
- ✅ Valid credentials (verified)
- ✅ Invalid credentials (unverified)
- ✅ Multiple findings with mixed validity
- ✅ Missing components (no results)
- ✅ Timeout handling
- ✅ Unexpected API responses

## Keywords
The detector uses the following keywords for efficient pre-filtering:
- `salesforce`
- `5AEP861`
- `3MVG9`

### Checklist:

salesforce-refresh-token-detector-tests


pkg/detectors/salesforcerefreshtoken/salesforcerefreshtoken.go
```diff
@@ -0,0 +1,153 @@
...
+			return false, fmt.Errorf("failed to read error response body: %w", err)
+		}
+		var apiResp introspectAPIResponse
+		if err := json.Unmarshal(bodyBytes, &apiResp); err != nil {
+			return false, fmt.Errorf("failed to unmarshal error response: %w (body: %s)", err, string(bodyBytes))
+		}
+
+		if !apiResp.Active {
+			return false, nil
+		}
```
abmussani, line 134: Does not it show that credentials are valid but the token is not active ?

abmussani (commented):
a minor comment

pkg/detectors/salesforcerefreshtoken/salesforcerefreshtoken.go
shahzadhaider1: Yes, it means that the credentials are valid, that is why we were able to get the status of the refresh token.

kashifkhan0771 (approved):

pkg/detectors/salesforcerefreshtoken/salesforcerefreshtoken.go
shahzadhaider1: Just to add - the validity of the credentials themselves is handled by a separate detector we already have, [salesforceoauth2](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/salesforceoauth2/salesforceoauth2.go). This particular detector is specifically designed to identify and verify refresh tokens, which is why the focus here is limited to that aspect.

PR #4296, opened 2025-07-08
Add additional unicode escape support

ghost:
### Description:
Currently TruffleHog supports 3 unicode escape formats:

```
\u0041
\\u0041
U+0041
```

This PR adds support for several other common formats:
```
\u{41} 
\U00000041
\x{41}
\41
&#x41;
%u0041
```

Unittests were added to ensure proper decoding of each new format, and benchmark tests was added to understand performance impacts. Based on my testing, there is no performance cost increase; however, I encourage the reviewer to conduct additional testing.  

Beyond additional testing, I encourage the reviewer to consider adding additional unicode escape formats. I attempted to add a generic `0xHH 0xHH` hex notation with space separation, but memory bloated too much and caused a significant performance impact. I think it'd be worth adding support for this case, and I left my work in the changed files, but it's commented out.


shahzadhaider1 (approved):

pkg/decoders/escaped_unicode.go
```diff
@@ -112,33 +192,87 @@ func decodeCodePoint(input []byte) []byte {
...
+}
 
-	return input
+// decodePercentEscape handles %uXXXX format - Percent-encoding (non-standard)
+func decodePercentEscape(input []byte) []byte {
+	return decodeWithPattern(input, percentEscapePat)
 }
+
+// decodeHexEscape handles 0xX format - Hexadecimal notation with space separation
+// func decodeHexEscape(input []byte) []byte {
```
kashifkhan0771, line 229: Are we intentionally commenting out this code?

ghost: I left this commented out, in case you wanted to try an implementation that doesn't bloat memory. I couldn't figure it out, but I think supporting this format would be really great given it's a pretty common format from what I've seen.  But if that needs to be put on pause, that's fine, and we can remove the comments. What do you think?

pkg/decoders/escaped_unicode.go
```diff
@@ -23,6 +23,33 @@ var (
...
+	longEscapePat = regexp.MustCompile(`\\U([a-fA-F0-9]{8})`)
+
+	// \x{X} format - Perl (variable length hex in braces)
+	perlEscapePat = regexp.MustCompile(`\\x\{([a-fA-F0-9]{1,6})\}`)
+
+	// \X format - CSS (hex without padding). Go's regexp (RE2) has no look-ahead, so we
+	// include the delimiter (whitespace, another backslash, or end-of-string) in the
+	// match using a non-capturing group. The delimiter is later re-inserted by the
+	// decoder when necessary.
+	cssEscapePat = regexp.MustCompile(`\\([a-fA-F0-9]{1,6})(?:\s|\\|$)`)
```
dustin-decker, line 42: If these are specific to Perl and CSS I think we could probably omit them for better performance

dustin-decker:
I'm a bit concerned about throughput of this decoder overall since it is the only regex based one and runs on every chunk. I think might be possible to do string operations instead of regex to improve that.

pkg/decoders/escaped_unicode.go
kashifkhan0771: Ah, I don’t want to be a blocker, but personally I’m not a big fan of leaving large chunks of code commented out 😅

zricethezav:
> I'm a bit concerned about throughput of this decoder overall since it is the only regex based one and runs on every chun. I think might be possible to do string operations instead of regex to improve that.

@jltrfl compare your branch and main against a few different repos to see if there is any noticeable difference. I'm assuming there will be which means we should probably introduce an ahocorasick filter. Let's run the benchmark first.

ghost:
> > I'm a bit concerned about throughput of this decoder overall since it is the only regex based one and runs on every chun. I think might be possible to do string operations instead of regex to improve that.
> 
> @jltrfl compare your branch and main against a few different repos to see if there is any noticeable difference. I'm assuming there will be which means we should probably introduce an ahocorasick filter. Let's run the benchmark first.

I ran benchmarks using hyperfine against the Linux, TruffleHog, and Redis repos:

**linux**
Feature (mean): ~807.7s ± 17.8s
Main (mean): ~791.8s ± 3.7s
Summary: main was ~2% faster

**trufflehog**
Feature (mean): ~5.93s ± 0.06s
Main (mean): ~6.02s ± 0.10s
Summary: feature was ~1% faster

**redis**
Feature (mean): ~11.90s ± 0.19s
Main (mean): ~11.85s ± 0.08s
Summary: virtually no difference

joeleonjr:
One of my commits didn't have an email attached to the commit, so I can't clear the `license/cla` check. Any suggestions?

ghost:
One of my commits didn't have an email attached to the commit, so I can't clear the license/cla check. Any suggestions?

zricethezav:
merging. We can follow up with an string check optimization if needed.

PR #4306, opened 2025-07-10
Add support for docker daemon as a source

### LINKED ISSUE #4275: Add support for scanning images from the Docker daemon

saghaulor:
### Description:
- Now you can scan an image directly after building it with docker build by using the docker:// prefix. This is ideal for local development and CI/CD pipelines that want to ensure images do not contain leaked secrets before pushing to an image registry.
- Resolves https://github.com/trufflesecurity/trufflehog/issues/4275
### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,7 +280,14 @@ trufflehog gcs --project-id=<project-ID> --cloud-environment --results=verified,
 Use the `--image` flag multiple times to scan multiple images.
 
 ```bash
+# to scan from a remote registry
 trufflehog docker --image trufflesecurity/secrets --results=verified,unknown
+
+# to scan from the local docker daemon
+trufflehog docker --image docker://new_image:tag --results=verified,unknown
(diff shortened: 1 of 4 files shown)


rosecodym (approved):
Thank you for putting up with all the back-and-forth!

PR #4309, opened 2025-07-10
Updated and Fixed CircleCI Source

### LINKED ISSUE #3009: CircleCI Projects API Endpoints Error
### TruffleHog Version
3.78.2
### Trace Output

2024-06-24T10:15:28-04:00       info-3  trufflehog      engine started  {"workers": 11}
2024-06-24T10:15:28-04:00       info-4  trufflehog      default engine options set
2024-06-24T10:15:28-04:00       info-4  trufflehog      engine initialized
2024-06-24T10:15:28-04:00       info-4  trufflehog      setting up aho-corasick core
2024-06-24T10:15:28-04:00       info-4  trufflehog      set up aho-corasick core
2024-06-24T10:15:28-04:00       info-2  trufflehog      starting scanner workers        {"count": 11}
2024-06-24T10:15:28-04:00       info-2  trufflehog      starting detector workers       {"count": 550}
2024-06-24T10:15:28-04:00       info-2  trufflehog      starting verificationOverlap workers    {"count": 11}
2024-06-24T10:15:28-04:00       info-2  trufflehog      starting notifier workers       {"count": 2}
2024-06-24T10:15:28-04:00       info-0  trufflehog      running source  {"source_manager_worker_id": "fCKOr", "with_units": false, "target_count": 0, "source_manager_units_configurable": true}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "3KJcS"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "euLrU"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "xaF12"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "1FSns"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "UUJkO"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "TIUPh"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "JTiw9"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "5qe6Z"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "Sju9s"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "MMyaC"}
2024-06-24T10:15:47-04:00       info-4  trufflehog      finished scanning chunks        {"scanner_worker_id": "hKjNs"}
2024-06-24T10:15:47-04:00       error   trufflehog      error running scan      {"error": "engine failed to finish execution: fatal: error getting projects: Get \"https://circleci.com/api/v1.1/projects\": GET https://circleci.com/api/v1.1/projects giving up after 4 attempt(s): Get \"https://circleci.com/api/v1.1/projects\": net/http: request canceled (Client.Timeout exceeded while awaiting headers)"}

### Expected Behavior

Should have ran the scan.

### Actual Behavior

Errored out attempting to list out the projects.

### Steps to Reproduce
1. Obtain CircleCI PAT
2. Have full org_owner in GH (Our permissions are inherited from GH)
3. run trufflehog with the circleci --token=CCIPAT_xxx

## Environment
 * OS: mac
 * Version: Latest

kashifkhan0771:
### Description:
This pull request updates the implementation of the CircleCI source and also fixes issue #3009.

**Changes Made:**

* Refactored the code to use clearer and more consistent variable and function names.
* Improved API response handling by correctly checking and handling HTTP status codes across all relevant endpoints (e.g., Projects, Builds, Jobs).
* Added additional logging to help with debugging and tracing the scan process.
* Introduced logic to parse and scan the CircleCI configuration (`.circleci/config.yml`) if the YAML content is included under the `jobs` field.

**What Was Broken and How It’s Fixed:**

Previously, our implementation used the project slug format (username/repo-name) when calling CircleCI’s APIs to fetch builds and jobs. However, CircleCI has updated their API to require a project ID instead.

To address this:

* We now extract the `vcs.url` from the project object returned by CircleCI’s `/project` endpoint.
* This URL contains the necessary information to derive the correct project ID.
* We store the VCS URL for each project and use it to construct the required API calls for builds, pipelines, and jobs.

### Checklist:


pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +114,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
 	if err != nil {
 		return fmt.Errorf("error getting projects: %w", err)
 	}
 
-	var scanned uint64
+	var projectsScanned uint64
 	scanErrs := sources.NewScanErrors()
 
-	for _, proj := range projects {
+	for _, project := range projects {
```
martinlocklear, line 126: 👏

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +114,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
 		s.jobPool.Go(func() error {
-			builds, err := s.buildsForProject(ctx, proj)
+			projBuilds, err := s.listProjectBuilds(ctx, &project)
```
martinlocklear: Just a personal preference, but I probably would have gone further and named this `projectBuilds`.  Similar comment in some of the other locations with naming refactors.

Totally up to you though, just a thought.

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +114,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
+				if err = s.chunkCircleCIYamlString(ctx, project, build, projBuildJobs.CircleYAML.YamlString, chunksChan); err != nil {
+					scanErrs.Add(fmt.Errorf("error chunking build yaml: %w", err))
+					return nil
+				}
 			}
 
-			atomic.AddUint64(&scanned, 1)
-			ctx.Logger().V(2).Info(fmt.Sprintf("scanned %d/%d projects", scanned, len(projects)))
+			atomic.AddUint64(&projectsScanned, 1)
+			ctx.Logger().V(2).Info(fmt.Sprintf("scanned %d/%d projects", projectsScanned, len(projects)))
```
martinlocklear: I might be more explicit here with `projectsScanned` becoming `countOfProjectsScanned` just to reduce the possibility of confusion in the future.  When I read "projects scanned" my first thought is that this is going to be a list of the actual projects (a list of objects for example).

pkg/sources/circleci/circleci.go
```diff
@@ -129,115 +169,143 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-
-	return projects, nil
 }
 
-type build struct {
-	BuildNum int `json:"build_num"`
-}
+func (s *Source) listProjectBuilds(ctx context.Context, proj *project) ([]build, error) {
+	// the vcs url is in format //circle.com/org-id/proj-id, so we need to remove the // and .com to use it in the API
+	vcsURL := strings.ReplaceAll(proj.VCSUrl, "//", "")
```
martinlocklear: Not sure if it will work here, but we might do better to use [a library to parse this](https://pkg.go.dev/net/url#Parse), rather than doing string manipulation to get our URLs

pkg/sources/circleci/circleci.go
```diff
@@ -263,6 +331,46 @@ func (s *Source) chunkAction(ctx context.Context, proj project, bld build, act a
...
+
+	chunkReader := sources.NewChunkReader()
+	chunkResChan := chunkReader(ctx, strings.NewReader(yamlString))
+	for data := range chunkResChan {
+		chunk := &sources.Chunk{
+			SourceType: s.Type(),
+			SourceName: s.name,
+			SourceID:   s.SourceID(),
+			JobID:      s.JobID(),
+			Data:       removeCircleSha1Line(data.Bytes()),
```
martinlocklear: Is it possible that this value is overwritten subsequently on line 369 with `chunk.Data = data.Bytes()`?

pkg/sources/circleci/circleci.go
```diff
@@ -129,115 +169,143 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		var projects []project
+		if err := json.NewDecoder(resp.Body).Decode(&projects); err != nil {
+			return nil, err
```
martinlocklear: Does it make sense to add some context to this error like we do in subsequent cases?
```return nil, fmt.Errorf("cannot decode CircleCI projects JSON: %w", err)```

**Review by martinlocklear — CHANGES_REQUESTED** [review]
Really appreciate the more explicit naming changes that were included in this PR!  

The PR description is a little sparse though - could we add a little bit more of the "what" and the "why" there.  Especially in the generated commit it'll be helpful for future readers not to have to follow a (potentially dead) link.

It's not 100% clear to me (without digging in more than I have) what the problem actually was, but my expectation is that it'll likely make a ton of sense to augment the testing to cover the new code paths as well as the error condition.

Requesting changes for:
* PR Description
* Confirming test coverage

Everything else I mentioned isn't blocking.

File: pkg/sources/circleci/circleci.go (hunk 5 above)
Comment (kashifkhan0771) [reply pkg/sources/circleci/circleci.go]: Interesting, the same thing is happening in `chunkAction` as well. I assume this is meant to filter out some false positives lines. I'll go ahead and remove the duplicate assignment.

File: pkg/sources/circleci/circleci.go (hunk 6 above)
Comment (kashifkhan0771) [reply pkg/sources/circleci/circleci.go]: Updated error for all APIs

File: pkg/sources/circleci/circleci_test.go [hunk 7]
```diff
@@ -70,29 +89,90 @@ func TestSource_Scan(t *testing.T) {
 			}
 
 			err = s.Init(ctx, tt.init.name, 0, 0, tt.init.verify, conn, 5)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("Source.Init() error = %v, wantErr %v", err, tt.wantErr)
+			if err != nil {
+				t.Errorf("Source.Init() error = %v", err)
```
camgunz: This pattern is everywhere--including, of course, this Source--but it's a little more straightforward to just do `require.NoError(t, err)` and move on.

Also I'm surprised this testing file wasn't already using [testify](https://pkg.go.dev/github.com/stretchr/testify). Let's take this opportunity to move to it; there's not a ton of code here so it's not a huge lift.

pkg/sources/circleci/circleci_test.go
```diff
@@ -70,29 +89,90 @@ func TestSource_Scan(t *testing.T) {
...
+				for chunk := range chunksCh {
+					chunks = append(chunks, chunk)
 				}
-			case <-ctx.Done():
-				t.Errorf("Source.Chunks() %s timed out", tt.name)
+			}()
+
+			// wait for chunks collection to complete
+			<-done
+			wg.Wait()
```
camgunz: It's a good instinct to build on existing patterns--in this case the async pattern here (where `Chunks` is run inside a separate goroutine)--but the async stuff is generally unnecessary, and this case is no exception. Let's simplify this a little:

- `chunksCh` can be something like `1000` instead of `1`, that'll ensure it won't block sends
- call `close(chunksCh)` right after the call to `Chunks`
- do the `for` loop to pull chunks from `chunksCh` into `chunks` right after that

It'll look something like:

```go
			chunksCh := make(chan *sources.Chunk, 1000)

			chunksErr := s.Chunks(ctx, chunksCh)

			close(chunksCh)

			chunks := []*sources.Chunk{}
			for chunk := range chunksCh {
				chunks = append(chunks, chunk)
			}
```

pkg/sources/circleci/circleci_test.go
```diff
@@ -70,29 +89,90 @@ func TestSource_Scan(t *testing.T) {
...
+
+			// check for errors
+			if (chunksErr != nil) != tt.wantErr {
+				t.Errorf("Source.Chunks() error = %v, wantErr %v", chunksErr, tt.wantErr)
+				return
+			}
```
camgunz: Let's replace this with a call to `require.NoError` (or `require.Error` if `tt.wantErr` is true) and move it up right after the call to `Chunks`.

pkg/sources/circleci/circleci_test.go
```diff
@@ -70,29 +89,90 @@ func TestSource_Scan(t *testing.T) {
...
+			if len(chunks) < tt.wantMinChunks {
+				t.Errorf("Source.Chunks() got %d chunks, want at least %d", len(chunks), tt.wantMinChunks)
+				return
 			}
 		})
 	}
 }
+
+// additional test for edge cases
+func TestSource_EdgeCases(t *testing.T) {
```
camgunz, line 112: Nice! 🏆

pkg/sources/circleci/circleci_test.go
```diff
@@ -31,35 +30,55 @@ func TestSource_Scan(t *testing.T) {
...
 			},
-			wantSourceMetadata: &source_metadatapb.MetaData{
-				Data: &source_metadatapb.MetaData_Circleci{
-					Circleci: &source_metadatapb.CircleCI{
-						VcsType:    "github",
-						Username:   "dustin-decker",
-						Repository: "circle-ci",
-						BuildStep:  "Spin up environment",
+			wantErr:       false,
+			wantMinChunks: 15,
```
camgunz, line 50: I love that you're explicitly checking chunk counts; this is exactly how we get robustness 💪🏻

pkg/sources/circleci/circleci.go
```diff
@@ -23,7 +26,8 @@ import (
 const (
 	SourceType = sourcespb.SourceType_SOURCE_TYPE_CIRCLECI
 
-	baseURL = "https://circleci.com/api/v1.1/"
+	// CircleCI has released API v2, but we're continuing to use v1.1 for now because it still supports listing endpoints and remains officially supported.
```
camgunz: 1. I love this comment; it's a good signpost for anyone who might come by later and think, "oh they're using the old API, I could update it"
2. Can you wrap it to > 1 line? I'm an 80 columns person because I like a lot of buffers side-by-side, but something like max 100 columns seems like a reasonable max--this one's like 150 or something 😹

pkg/sources/circleci/circleci.go
```diff
@@ -38,6 +42,35 @@ type Source struct {
 	sources.CommonSourceUnitUnmarshaller
 }
 
+// CircleCI API Response types
+type project struct {
+	VCS      string `json:"vcs_type"`
+	Username string `json:"username"`
+	RepoName string `json:"reponame"`
+	VCSUrl   string `json:"vcs_url"`
```
camgunz: Prepare yourself for a petty style comment:

- If `Username` is one word, `RepoName` is one word, so either (`UserName` and `RepoName`) or (`Username` and `Reponame`)
- If the `VCS` initialism is all-caps, then the `Url` initialism should be all-caps, so either `VcsUrl` or `VCSURL`. Do both options feel weird? Yeah! Is this why camel/pascal case is **incorrect**? One of many reasons yes!
  - If you choose `Url`, don't forget to change `action.OutputURL` to `action.OutputUrl`

pkg/sources/circleci/circleci.go
```diff
@@ -38,6 +42,35 @@ type Source struct {
...
+}
+
+type build struct {
+	BuildNum int `json:"build_num"`
+}
+
+type buildJobs struct {
+	CircleYAML struct {
+		YamlString string `json:"string"`
```
camgunz: Prepare yourself for another petty style comment: choose either (`CircleYAML` and `YAMLString`) or (`CircleYaml` and `YamlString`).

pkg/sources/circleci/circleci.go
```diff
@@ -129,115 +170,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 
-	return builds, nil
-}
+		ctx.Logger().V(2).Info(fmt.Sprintf("successfully listed %d builds for project: %s", len(builds), proj.RepoName))
 
-type action struct {
-	Index     int    `json:"index"`
-	OutputURL string `json:"output_url"`
-}
+		return builds, nil
```
camgunz: Tell me if this is wrong, but it seems like `build` is only ever used as a wrapper for its `BuildNum` field. If that's true, we could simplify things by just using `int` (`[]int`, etc). If you're worried about type safety you can do a `type BuildNum int` at the top (which I would support). Then, anything taking a `build` can now just take an `int`/`BuildNum`. This has at least two benefits:

- It clarifies the requirements: you only need a build number, not a whole build
- There's no "fragile base class" or "god object" problem where now everything is using `build` and you can't modify it for fear of breaking a dependency

Admittedly these are more app-level concerns vs. Source-level concerns, but every little bit helps when fighting the demon of complexity.

That said, if we're re-using `build` in an API request or something, then I think it makes sense to keep it.

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +115,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
 
 // Chunks emits chunks of bytes over a channel.
 func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
-	projects, err := s.projects(ctx)
+	// TODO: list Artifacts, list checkout-keys, list env-variables
+
+	// list all projects
```
camgunz: Eh, I'd leave this comment out

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +115,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
+	projects, err := s.listAllProjects(ctx)
 	if err != nil {
 		return fmt.Errorf("error getting projects: %w", err)
 	}
 
-	var scanned uint64
+	var countOfProjectsScanned uint64
```
camgunz: As long as we're polishing here, let's take the [Go standard library's advice](https://pkg.go.dev/sync/atomic#AddUint64) and switch to `atomic.Uint64` instead. As a bonus, it'll ensure we use `.Load()` to get the value safely as well.

pkg/sources/circleci/circleci.go
```diff
@@ -129,115 +170,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-	return bldRes.Steps, nil
+		return &buildResp, nil
+	default:
+		return nil, fmt.Errorf("unexpected status code: %d while fetching project: %s build: %d jobs", resp.StatusCode, proj.RepoName, projectBuild.BuildNum)
+	}
 }
 
-func (s *Source) chunkAction(ctx context.Context, proj project, bld build, act action, stepName string, chunksChan chan *sources.Chunk) error {
-	req, err := http.NewRequest("GET", act.OutputURL, nil)
+func (s *Source) chunkAction(ctx context.Context, proj project, projectBuild build, action action, stepName string, chunksChan chan *sources.Chunk) error {
```
camgunz: More "do we need a whole struct" commentary here: if we only ever need `OutputURL` from an `action`, can we get away with just passing that around instead of a whole struct?

Also it seems like we never use `action.Index`, so it seems fine to remove... unless we're implicitly relying on it for some kind of validation maybe?

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +115,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
 	scanErrs := sources.NewScanErrors()
 
-	for _, proj := range projects {
+	for _, project := range projects {
 		s.jobPool.Go(func() error {
-			builds, err := s.buildsForProject(ctx, proj)
+			projectBuilds, err := s.listProjectBuilds(ctx, &project)
 			if err != nil {
-				scanErrs.Add(fmt.Errorf("error getting builds for project %s: %w", proj.RepoName, err))
+				scanErrs.Add(fmt.Errorf("error getting builds for project %s: %w", project.RepoName, err))
```
camgunz: Hmm, I know the use of `scanErrs` isn't something you added, but this isn't optimal for debugging. Let's remove it and use `ctx.Logger().Error` instead. The main benefit of this is they'll show up as errors in the logs, and these all kind of look like things we'd want to know about if there's a problem w/ the scanner. We'll also want a fair amount of information in the log messages; a pattern we use on scanning team is:

```go
ctx = context.WithValues(ctx,
  "repository_name", project.RepoName,
)
```

So like, here you'd want to add the repository name, and then further down you'd want to add the build number, and then further further down you'd add the step name and action output URL. For example:

```go
for _, action := range step.Actions {
  ctx = context.WithValues(ctx,
    "step_name", step.Name,
    "action_output_url", action.OutputURL,
  )
  if err = s.chunkAction(ctx, project, build, action, step.Name, chunksChan); err != nil {
    ctx.Logger().Error(fmt.Errorf("error chunking action: %w", err))
    return nil
  }
}
```

This is super useful because then we can track the state of the source via the logging messages, and since they're attached to the context, any subsequent function/method calls will have the same values in *their* logging calls.

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +115,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
+				if err = s.chunkCircleCIYamlString(ctx, project, build, projBuildJobs.CircleYAML.YamlString, chunksChan); err != nil {
+					scanErrs.Add(fmt.Errorf("error chunking build yaml: %w", err))
+					return nil
+				}
 			}
 
-			atomic.AddUint64(&scanned, 1)
-			ctx.Logger().V(2).Info(fmt.Sprintf("scanned %d/%d projects", scanned, len(projects)))
+			atomic.AddUint64(&countOfProjectsScanned, 1)
+			ctx.Logger().V(2).Info(fmt.Sprintf("scanned %d/%d projects", countOfProjectsScanned, len(projects)))
```
camgunz: Yeah see, even the original code forgot to use `atomic.LoadUint64`. Super easy to do.

Also hmm, looking at this logging call, it seems like this Source hasn't been updated to track progress. We definitely want to do that. Let's use [`SetProgressComplete`](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/sources.go#L464); probably like:

```go
localScanCount := countOfProjectsScanned.Add(1)
s.setProgressComplete(
  localScanCount,
  len(projects),
  fmt.Sprintf("Scanned %d/%d projects", localScanCount, len(projects)), // Note the capitalization here -- this ends up in the enterprise UI so we want it to look nice
  "", // This would be for resumption, but we're not currently using it
)

File: pkg/sources/circleci/circleci.go [hunk 21]
```diff
@@ -129,115 +170,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-	Username string `json:"username"`
-	RepoName string `json:"reponame"`
-}
-
-func (s *Source) projects(_ context.Context) ([]project, error) {
-	reqURL := fmt.Sprintf("%sprojects", baseURL)
-	req, err := http.NewRequest("GET", reqURL, nil)
+// listAllProjects lists all the projects the token can access
+func (s *Source) listAllProjects(ctx context.Context) ([]project, error) {
+	ctx.Logger().V(5).Info("listing projects")
```
camgunz, line 205: Nice trace logging! Love it

pkg/sources/circleci/circleci.go
```diff
@@ -129,115 +170,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 	if err != nil {
 		return nil, err
 	}
-	defer res.Body.Close()
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	switch resp.StatusCode {
```
camgunz, line 223: Love `switch`/`case` here

pkg/sources/circleci/circleci.go
```diff
@@ -129,115 +170,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	case http.StatusOK:
+		var projects []project
+		if err := json.NewDecoder(resp.Body).Decode(&projects); err != nil {
+			return nil, fmt.Errorf("cannot decode CircleCI projects JSON: %w", err)
+		}
 
-	if res.StatusCode > 399 && res.StatusCode < 500 {
-		return nil, fmt.Errorf("invalid credentials, status %d", res.StatusCode)
+		ctx.Logger().V(2).Info(fmt.Sprintf("successfully listed %d projects", len(projects)))
```
camgunz: 1. Also more great logging
2. It's fine to leave as-is, but I find whenever I'm using `fmt.Sprintf` in a logging call I probably ought to just make it a param, e.g:

```go
ctx.Logger().V(2).Info("successfully listed projects", "project_count", len(projects))
```

The main thing here is it makes searching through logs easier--you can just do `= "successfully listed projects"` instead of `=~ "successfully listed \d+ projects"`

pkg/sources/circleci/circleci.go
```diff
@@ -82,41 +115,49 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
 
-	for _, proj := range projects {
+	for _, project := range projects {
 		s.jobPool.Go(func() error {
-			builds, err := s.buildsForProject(ctx, proj)
+			projectBuilds, err := s.listProjectBuilds(ctx, &project)
 			if err != nil {
-				scanErrs.Add(fmt.Errorf("error getting builds for project %s: %w", proj.RepoName, err))
+				scanErrs.Add(fmt.Errorf("error getting builds for project %s: %w", project.RepoName, err))
 				return nil
```
camgunz, line 137: The original code does this, but I think instead of bailing on these errors we should probably just `continue`, otherwise we'll skip things we probably could have successfully scanned. This goes for builds, steps, and actions

pkg/sources/circleci/circleci.go
```diff
@@ -251,18 +324,55 @@ func (s *Source) chunkAction(ctx context.Context, proj project, bld build, act a
...
+
+		if err := common.CancellableWrite(ctx, chunksChan, chunk); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *Source) chunkCircleCIYamlString(ctx context.Context, proj project, projectBuild build, yamlString string, chunksChan chan *sources.Chunk) error {
```
camgunz: This method is substantially similar to `chunkAction`; I think it's probably worth factoring out the logic that fetches the output of an action into its own method, and then unifying the functionality.

camgunz (changes requested):
This PR is great; this Source is way way better. I noted some more things that we should improve while we're in here (most of these things are "here's how we do things on team scanning" because we want to pull you more into our little club), but happy to discuss whatever if I misunderstood something.

pkg/sources/circleci/circleci.go
kashifkhan0771: Everything here depends on the previous step. If we're unable to list a specific build, we won't have its buildNum, which is required to fetch its steps. In that case, we log the error and proceed to the next project.

martinlocklear (approved):
🚀 🚀

pkg/sources/circleci/circleci.go
camgunz: Oh I missed this was running inside a closure. But then I still think we want to `continue` instead of `return` in a few places:

- line 150: If we `return` here then we'll skip subsequent `buildNum`s from `projectBuilds`
- line 160: If we `return` here then we'll skip subsequent `action`s from `step.Actions`
- line 167: If we `return` here then we'll skip subsequent `action`s from `step.Actions`
- line 176: If we `return` here then we'll skip subsequent `buildNum`s from `projectBuilds`

pkg/sources/circleci/circleci_test.go
camgunz: Thanks so much for doing this! 🏆

pkg/sources/circleci/circleci.go
kashifkhan0771: Done ✅

camgunz (approved):
LG! Thanks for sticking w/ me through this

PR #4312, opened 2025-07-10
Added Anypoint API OAuth2 Detector

nabeelalam:
### Description:
Anypoint uses OAuth2 credentials for authentication. [Reference](https://anypoint.mulesoft.com/exchange/portals/anypoint-platform/f1e97bc6-315a-4490-82a7-23abe036327a.anypoint-platform/access-management-api/minor/1.0/pages/Authentication/).

This PR adds a new detector for Anypoint's OAuth2 credentials.

### Checklist:


pkg/detectors/anypointoauth2/anypointoauth2.go
```diff
@@ -0,0 +1,128 @@
...
+				s1.Verified = isVerified
+				s1.SetVerificationError(verificationErr)
+
+			}
+
+			results = append(results, s1)
+
+			if s1.Verified {
+				break
+			}
```
amanfcp: Just curious, what happens if there are two valid set of credentials in `dataStr`?

shahzadhaider1: Do we expect a single ID to be paired with multiple secrets, or is it typically a one-to-one mapping?

shahzadhaider1 (approved):
Apart from Amaan's comment, everything looks good. GW!

kashifkhan0771 (approved):

pkg/detectors/anypointoauth2/anypointoauth2.go
kashifkhan0771: If ID can only be paired with one secret and vice-versa - We should remove them from the list once validated successfully - This will save us from extra iterations.

kashifkhan0771: This is a non-blocking comment

nabeelalam: Good catch @amanfcp IDs and secrets seem to be tied and I haven't seen an option to reset/rotate a client secret yet. I agree with @kashifkhan0771 that we should remove that secret from the secret list and move to the next ID found. Thanks

PR #4315, opened 2025-07-14
[Detector] - Restore and Refactor Detectors starting with *CA*

amanfcp:
### Description:
- Restored caflou, calorieninja, cannyio detectors' health
- Refactor verification logic across multiple detectors to use a unified `verifyMatch` function 
- Updated integration tests to reflect changes in secret retrieval and verification processes
- Adjusted regex patterns for improved accuracy in detecting API keys
- Enhanced error handling for API responses.

## Note: 
I'm having trouble creating test accounts in Captain Data (.co domain for <V4) and Caspio. The accounts are stuck in verification state. 

### Checklist:


pkg/detectors/cannyio/cannyio.go
```diff
@@ -65,6 +62,48 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	payload := strings.NewReader("apiKey=" + token)
+	req, err := http.NewRequestWithContext(ctx, "POST", "https://canny.io/api/v1/boards/list", payload)
+	if err != nil {
+		return false, nil
```
shahzadhaider1:
```suggestion
		return false, fmt.Errorf("failed to create HTTP request: %w", err)
```

pkg/detectors/cannyio/cannyio.go
```diff
@@ -65,6 +62,48 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		var errResp errorResponse
+		if err := json.Unmarshal(body, &errResp); err != nil {
+			return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+		}
+		
+		if strings.Contains(strings.ToLower(errResp.Error), "invalid api key") {
+			return false, nil
+		}
+		
+		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
```
shahzadhaider1, line 91: Please correct me if I'm wrong but should we really return `unexpected HTTP response status` in 400 case?

pkg/detectors/cannyio/cannyio.go
```diff
@@ -65,6 +62,48 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	case http.StatusBadRequest:
+		body, err := io.ReadAll(res.Body)
+		if err != nil {
+			return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+		}
+		
+		var errResp errorResponse
+		if err := json.Unmarshal(body, &errResp); err != nil {
+			return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+		}
```
shahzadhaider1: I am not sure what the API response/behavior is, so please correct me if I'm wrong, can we simply check for the desired string in the plain response instead of unmarshalling first?

pkg/detectors/capsulecrm/capsulecrm.go
```diff
@@ -66,6 +58,33 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.capsulecrm.com/api/v2/users", http.NoBody)
+	if err != nil {
+		return false, nil
```
shahzadhaider1:
```suggestion
		return false, fmt.Errorf("failed to create HTTP request: %w", err)
```

pkg/detectors/carboninterface/carboninterface.go
```diff
@@ -68,6 +58,34 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://www.carboninterface.com/api/v1/auth", http.NoBody)
+	if err != nil {
+		return false, nil
```
shahzadhaider1:
```suggestion
		return false, fmt.Errorf("failed to create HTTP request: %w", err)
```

pkg/detectors/cashboard/cashboard.go
```diff
@@ -26,7 +27,7 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat  = regexp.MustCompile(detectors.PrefixRegex([]string{"cashboard"}) + `\b([0-9A-Z]{3}-[0-9A-Z]{3}-[0-9A-Z]{3}-[0-9A-Z]{3})\b`)
-	userPat = regexp.MustCompile(detectors.PrefixRegex([]string{"cashboard"}) + `\b([0-9a-z]{1,})\b`)
+	userPat = regexp.MustCompile(detectors.PrefixRegex([]string{"username"}) + `\b([0-9a-z]{1,})\b`)
```
shahzadhaider1, line 30: I think we shouldn't remove the `cashboard` keyword, because the regex pattern seems too generic

pkg/detectors/cashboard/cashboard.go
```diff
@@ -78,6 +67,38 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, user, token string) (bool, error) {
+	data := fmt.Sprintf("%s:%s", user, token)
+	sEnc := b64.StdEncoding.EncodeToString([]byte(data))
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.cashboardapp.com/account", http.NoBody)
+	if err != nil {
+		return false, nil
```
shahzadhaider1:
```suggestion
		return false, fmt.Errorf("failed to create HTTP request: %w", err)
```

shahzadhaider1 (changes requested):
Good work. 
I have added some feedback.

pkg/detectors/cannyio/cannyio.go
amanfcp: Third party API services work in mysterious ways :D
If the token is invalid or not given at all, cannyio api returns `400` with a message _invalid api key_

pkg/detectors/cashboard/cashboard.go
amanfcp: This is just a prefix for catching the `username` once the chunk is inside the detector. 
The reason for removing _cashboard_ keyword is that it will create unnecessary sets of credentials.

pkg/detectors/caflou/caflou.go
```diff
@@ -74,6 +66,33 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://app.caflou.com/api/v1/accounts", nil)
```
nabeelalam: nit: Seems like a good opportunity to replace `nil` and hard coded strings with `http.NoBody` and `http.MethodGet` in all of the detectors

pkg/detectors/calendlyapikey/calendlyapikey.go
```diff
@@ -22,7 +23,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"calendly"}) + `\b(eyJ[A-Za-z0-9-_]{100,300}\.eyJ[A-Za-z0-9-_]{100,300}\.[A-Za-z0-9-_]+)\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"calendly"}) + `\b(eyJ[A-Za-z0-9-_]{10,300}\.eyJ[A-Za-z0-9-_]{10,300}\.[A-Za-z0-9-_]+)\b`)
```
kashifkhan0771: In my opinion, this change could lead to more false positives. We're essentially reducing our safeguards. Are we confident about this change? Also, have we verified whether Calendly actually generates keys of this shorter length?

amanfcp: I made this change cos I saw some valid keys that were less than 100 letters. If we keep it as is, we're potentially increasing false negatives. 

I'll confirm this again.

amanfcp:
Contributes to #4051

shahzadhaider1 (approved):

PR #4318, opened 2025-07-15
Update notion api rule to take new key version

Tapiocaba:
### Description:
This commit adds the updated (`ntn_` prefix) Notion API regex key detection and updates the test keys. `secret` prefix keys have been depreciated in the public api.

### Checklist:


shahzadhaider1 (changes requested):
Hey @Tapiocaba, thanks for opening this PR to update the regex pattern for the Notion detector based on the recent token format changes, really appreciate your contribution!

That said, we won't be able to merge this as-is since the old secret format is still valid and in use. Instead of modifying the existing detector, it would be better to create a v2 of the Notion detector. This way, we can support detection of both the old and new token formats without losing compatibility.

If you'd like to help with creating the new version, feel free to check out our [guide](https://github.com/trufflesecurity/trufflehog/blob/main/hack/docs/Adding_Detectors_external.md#adding-new-token-formats-to-an-existing-scanner) on building a v2 detector. I am happy to assist if you have any questions.

Thanks again for bringing this to our attention!

pkg/detectors/notion/notion.go
```diff
@@ -22,13 +22,13 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(secret_[A-Za-z0-9]{43})\b`)
+	keyPat = regexp.MustCompile(`\b((?:secret_[A-Za-z0-9]{43})|ntn_[0-9]{11}[A-Za-z0-9]{32}[A-Za-z0-9]{3})\b`)
```
kashifkhan0771: Great 😃 Could you move the new pattern to a separate V2 detector instead of merging it into the existing one? Since the API for verification is the same, we can extract the verification logic into a shared method and use it in both versions.

kashifkhan0771 (changes requested):

Tapiocaba:
I get a golangci-lint error with:

```
make lint                               lisalin@dhcp-10-250-103-32
golangci-lint run --enable bodyclose --enable copyloopvar --enable misspell --out-format=colored-line-number --timeout 10m
Error: unknown flag: --out-format
The command is terminated due to an error: unknown flag: --out-format
make: *** [lint] Error 3
```

I tried running it on my own without the make as well and it doesn't run, but would love some advice on fixing this. Was looking at [this thread](https://github.com/golangci/golangci-lint/discussions/5612) about the issue but not sure if ya'll have done something about it with this specifically, so wanted to double check.

amanfcp:
> I get a golangci-lint error with:
> 
> ```
> make lint                               lisalin@dhcp-10-250-103-32
> golangci-lint run --enable bodyclose --enable copyloopvar --enable misspell --out-format=colored-line-number --timeout 10m
> Error: unknown flag: --out-format
> The command is terminated due to an error: unknown flag: --out-format
> make: *** [lint] Error 3
> ```
> 
> I tried running it on my own without the make as well and it doesn't run, but would love some advice on fixing this. Was looking at [this thread](https://github.com/golangci/golangci-lint/discussions/5612) about the issue but not sure if ya'll have done something about it with this specifically, so wanted to double check.

@Tapiocaba can you check which version of go and golangci-lint you're using? It should be `1.24` and golangci-lint should be built using that version before getting used. On running `golangci-lint version`, It should display something like `golangci-lint has version v1.64.8 built with go1.24.1`

PR #4319, opened 2025-07-16
Added KeySet Pagination for Gitlab Projects V2 Enumeration

kashifkhan0771:
### Description:
This Pull Request introduces keyset pagination for GitLab project enumeration, building on the simplified enumeration approach from https://github.com/trufflesecurity/trufflehog/pull/4283.
Key Changes:

- Implements keyset-based pagination using GitLab's Scan2 method for automatic pagination handling
- Replaces traditional offset-based pagination with more efficient keyset pagination
- Improves performance and reliability for large GitLab instances

Important Note:
The Scan2 API is currently experimental and may introduce breaking changes in future GitLab client releases. We should monitor for API changes when updating the GitLab client dependency.

### Checklist:


pkg/sources/gitlab/gitlab.go
```diff
@@ -665,68 +665,65 @@ func (s *Source) getAllProjectReposV2(
...
-			if err := reporter.UnitErr(ctx, err); err != nil {
-				return err
-			}
-
-			break
-		}
-
-		ctx.Logger().V(3).Info("listed projects", "count", len(projects))
+	// https://pkg.go.dev/gitlab.com/gitlab-org/api/client-go#Scan2
+	projectsChan := gitlab.Scan2(func(p gitlab.PaginationOptionFunc) ([]*gitlab.Project, *gitlab.Response, error) {
```
rosecodym: I found this identifier confusing, because this variable isn't holding a channel. It's holding an iterator! I'd prefer to call it `projectsSeq` or `projectsIter` or something like that.

pkg/sources/gitlab/gitlab.go
```diff
@@ -665,68 +665,65 @@ func (s *Source) getAllProjectReposV2(
...
-				"project_name", proj.NameWithNamespace)
+	totalCount := 0
 
-			// skip projects configured to be ignored.
-			if ignoreRepo(proj.PathWithNamespace) {
-				projCtx.Logger().V(3).Info("skipping project", "reason", "ignored in config")
+	// process each project
+	for project, projectErr := range projectsChan {
+		if projectErr != nil {
+			err := fmt.Errorf("failed to enumerate project: %w", projectErr)
```
rosecodym: This error message mentions a single project, which I find misleading, because an error here indicates that there was a problem fetching an entire _page_ of projects. "Failed" is also somewhat misleading, because some projects might already have been enumerated before this error happened. I'd use the message `"error during project enumeration: %w"` or something like that.

pkg/sources/gitlab/gitlab_integration_test.go
```diff
@@ -559,7 +559,7 @@ func TestSource_InclusionGlobbing(t *testing.T) {
 			assert.NoError(t, err)
 
 			var errs []error
-			ignoreRepo := buildIgnorer(ctx, src.includeRepos, src.ignoreRepos, func(err error, pattern string) {
+			ignoreRepo := buildIgnorer(src.includeRepos, src.ignoreRepos, func(err error, pattern string) {
```
rosecodym, line 562: Have these tests not been compiling before now?

rosecodym (commented):
Thank you for this! I have some small notes about naming but nothing about the logic.

go.mod
```diff
@@ -98,7 +98,8 @@ require (
 	github.com/trufflesecurity/disk-buffer-reader v0.2.1
 	github.com/wasilibs/go-re2 v1.9.0
 	github.com/xo/dburl v0.23.8
-	gitlab.com/gitlab-org/api/client-go v0.127.0
+	// TODO: v1.134.0 is available but deprecates existing Auth methods, should be updated separately
+	gitlab.com/gitlab-org/api/client-go v0.129.0
```
rosecodym, line 104: Should we pin this version?

pkg/sources/gitlab/gitlab_integration_test.go
kashifkhan0771: I tried running the integration tests, but they were failing even before my changes likely due to a token expiry issue. However, I did come across an error, which I fixed. 🙂

rosecodym: I have updated the token to resolve that error!

rosecodym (approved):
Thanks for turning this around so quickly!

shahzadhaider1 (approved):
GW

PR #4320, opened 2025-07-16
Scan GitLab Groups

### LINKED ISSUE #788: Scan GitLab Groups using a group ids
### Community Note

### Description

It would be good to be able to scan Groups in Gitlab. Groups can collect a number of projects together.

## Problem to be Addressed

Gitlab scanning is missing a construct that gitlab uses as part of it's organisation

Add Command line flags to specify group ids to scan

### References

* #0000

shahzadhaider1:
### Description:
This PR adds support for scanning GitLab groups and their subgroups using the `--group-id` flag. Users can now specify one or more group IDs to scan projects from specific groups instead of scanning all accessible projects.

### New Functionality
- Added `--group-id` flag to accept a list of GitLab group IDs
- Implemented `getAllProjectReposInGroup()` function to fetch projects from specified groups
- Added support for scanning subgroups using GitLab's `include_subgroups=true` parameter

### Technical Implementation
- Uses GitLab's `Groups.ListGroupProjects()` API instead of the general `Projects.ListProjects()` API
- Handles edge cases where specifying both parent and child groups could result in duplicate project scanning
- Maintains existing pagination, error handling, and logging patterns
- Preserves all existing project filtering and validation logic

### Usage
```bash
# Scan single group
trufflehog gitlab --group-id 12345

# Scan multiple groups
trufflehog gitlab --group-id 12345 --group-id 67890
```

## Benefits
- **Targeted Scanning**: Users can scan specific groups instead of all accessible projects
- **Subgroup Support**: Automatically includes projects from all subgroups
- **Performance**: Reduces API calls by scoping to relevant groups only

## Testing
- Verified functionality with single and multiple group IDs
- Tested subgroup inclusion behavior

### Checklist:


pkg/sources/gitlab/gitlab.go
```diff
@@ -401,7 +408,12 @@ func (s *Source) Validate(ctx context.Context) []error {
 		},
 	}
 
-	if feature.UseSimplifiedGitlabEnumeration.Load() {
+	if len(s.groupIds) > 0 {
+		if err := s.getAllProjectReposInGroup(ctx, apiClient, s.groupIds, ignoreProject, visitor); err != nil {
+			errs = append(errs, err)
+			return errs
+		}
+	} else if feature.UseSimplifiedGitlabEnumeration.Load() {
```
shahzadhaider1: I'm not entirely sure if this is the best approach, so I am happy to discuss if anyone has thoughts or suggestions!

shahzadhaider1: ...especially since we're already doing this at line 454 in main.go: https://github.com/trufflesecurity/trufflehog/blob/eafb8c5f6a06175141c27f17bcc17941853d0047/main.go#L454

rosecodym: This looks fine to me. Do you have any particular concerns?

pkg/sources/gitlab/gitlab.go
```diff
@@ -728,6 +739,117 @@ func (s *Source) getAllProjectReposV2(
 	return nil
 }
 
+// getAllProjectReposInGroup fetches all projects in a GitLab group and its subgroups.
+// It uses the group projects API with include_subgroups=true parameter.
+func (s *Source) getAllProjectReposInGroup(
+	ctx context.Context,
+	apiClient *gitlab.Client,
+	groupIDs []string,
```
rosecodym: This function's name is `getAllProjectReposInGroup` (singular `Group`) but this parameter contains multiple group IDs. I don't think one is better than the other, but they should be consistent.

pkg/sources/gitlab/gitlab.go
```diff
@@ -728,6 +739,117 @@ func (s *Source) getAllProjectReposV2(
...
+	ignoreRepo func(string) bool,
+	reporter sources.UnitReporter,
+) error {
+	gitlabReposEnumerated.WithLabelValues(s.name).Set(0)
```
rosecodym, line 743: There is also a `gitlabGroupsEnumerated` metric that would be appropriate to use here. You're not really "enumerating" groups, but it would still be helpful to record of how many we're planning to scan.

pkg/sources/gitlab/gitlab.go
```diff
@@ -728,6 +739,117 @@ func (s *Source) getAllProjectReposV2(
...
+	processedProjects := make(map[string]bool)
+
+	var projectsWithNamespace []string
+	const (
+		orderBy         = "id"
+		paginationLimit = 100
+	)
```
rosecodym, line 752: Did you investigate whether keyset pagination is available for this API endpoint? I believe that keyset pagination is preferred when available.

pkg/sources/gitlab/gitlab.go
```diff
@@ -728,6 +739,117 @@ func (s *Source) getAllProjectReposV2(
...
+		"include_subgroups", true,
+		"list_options", listOpts)
+
+	for _, groupID := range groupIDs {
+		groupCtx := context.WithValues(ctx, "group_id", groupID)
+
+		projectOpts.Page = 0
+		groupCtx.Logger().V(2).Info("processing group", "group_id", groupID)
+
+		for {
```
rosecodym, line 778: This loop contains a lot of duplicate code that might not be necessary. What you've copied is a relatively complex set of logic that is used to de-duplicate projects that appear in multiple groups. However, that logic is known to be buggy, and we hope to remove it before too long, because the new keyset pagination-based technique will hopefully make it irrelevant.

However, we don't _know_ whether that new technique will work, and even if it does, we might need to de-duplicate projects here no matter what. So I'm not commenting with a particular prescription - I just wanted to flag the repeated code.

pkg/engine/gitlab.go
```diff
@@ -46,6 +46,10 @@ func (e *Engine) ScanGitLab(ctx context.Context, c sources.GitlabConfig) (source
 		connection.Repositories = c.Repos
 	}
 
+	if len(c.GroupIds) > 0 {
```
kashifkhan0771, line 49: What happens if someone passes both the `--group-id` and `--repo` flags?

pkg/sources/gitlab/gitlab.go
```diff
@@ -266,7 +268,12 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, tar
 				return ctx.Err()
 			},
 		}
-		if feature.UseSimplifiedGitlabEnumeration.Load() {
+
+		if len(s.groupIds) > 0 {
```
kashifkhan0771: This logic is starting to get a bit complex with nested conditions and is duplicated in multiple places. It's better to extract it into a dedicated method for clarity and reuse. Something like:
```go
func (s *Source) listProjects(ctx context.Context,
	apiClient *gitlab.Client,
	ignoreProject func(string) bool,
	visitor sources.UnitReporter) error {
	if len(s.groupIds) > 0 {
		return s.getAllProjectReposInGroup(ctx, apiClient, s.groupIds, ignoreProject, visitor)
	}

	if feature.UseSimplifiedGitlabEnumeration.Load() {
		return s.getAllProjectReposV2(ctx, apiClient, ignoreProject, visitor)
	}

	return s.getAllProjectRepos(ctx, apiClient, ignoreProject, visitor)
}
```

pkg/sources/gitlab/gitlab.go
```diff
@@ -266,7 +268,12 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, tar
...
+			if err := s.getAllProjectReposInGroup(ctx, apiClient, s.groupIds, ignoreRepo, reporter); err != nil {
```
kashifkhan0771: The source already has group IDs, so why are we passing them separately as parameters?

proto/sources.proto
```diff
@@ -232,6 +232,7 @@ message GitLab {
   repeated string include_repos = 9;
   bool exclude_projects_shared_into_groups = 10;
   bool remove_auth_in_url = 11;
+  repeated string group_ids = 12;
```
kashifkhan0771, line 236: We should be consistent with the flag. The flag is `group-id` and here we have `group_ids`

kashifkhan0771 (changes requested):

proto/sources.proto
shahzadhaider1: You're right about the naming difference. The reason I used `group_ids` here is because this field will hold all the values passed via the `--group-id` flag, which can be specified multiple times in the CLI. Since it represents a list of group IDs, I went with the plural form for clarity. 

On the CLI side, `--group-id` makes sense as a single-use flag that can be repeated, but internally it's stored as a collection, hence the plural naming here. Let me know what do you think.

pkg/engine/gitlab.go
shahzadhaider1: If both `--repo` and `--group-id` are provided, the `--repo` flag takes precedence, meaning only the specified repos will be scanned, and the group IDs will be ignored.

To make it more clear, we can add a log that clearly says something like:
```
--repo and --group-id cannot be used together. --group-id will be ignored

```

pkg/sources/gitlab/gitlab.go
shahzadhaider1: [List Group Projects API](https://docs.gitlab.com/api/groups/#list-projects) does not support [keyset pagination](https://docs.gitlab.com/api/rest/#keyset-based-pagination)

Supported Resources: https://docs.gitlab.com/api/rest/#supported-resources

proto/sources.proto
kashifkhan0771: Other repeatable flags, such as `--exclude-repos` and `--include-repos`, also use plural naming.

pkg/engine/gitlab.go
kashifkhan0771: To stay consistent with the [GitHub source](https://github.com/trufflesecurity/trufflehog/blob/main/main.go#L741) behavior where passing both `--org` and `--repo` results in an error. We should also return an error if both `--group-id` and `--repo` are provided.

shahzadhaider1: done

proto/sources.proto
shahzadhaider1: You're right that they use plural naming, but please correct me if I’m wrong: I believe since our flag accepts one value at a time (e.g., `--group-id 123 --group-id 456`), it makes more sense to keep it singular as `--group-id` rather than `--group-ids`. That was my reasoning behind naming it this way.

pkg/sources/gitlab/gitlab.go
shahzadhaider1: This comment has been deliberated internally; hence, marking as resolved.

kashifkhan0771 (approved):
Great work, @shahzadhaider1 - LGTM!
Before merging, please make sure to test the following scenarios:

* Passing both `--group-id` and `--repo`
* Enumerating projects for multiple group IDs
* Enumerating without any `--group-id` (regression testing)
* Enumerating with the old logic by disabling the simplified enumeration flag

shahzadhaider1:
> Great work, @shahzadhaider1 - LGTM! Before merging, please make sure to test the following scenarios:
> 
> * Passing both `--group-id` and `--repo`
> * Enumerating projects for multiple group IDs
> * Enumerating without any `--group-id` (regression testing)
> * Enumerating with the old logic by disabling the simplified enumeration flag

Yes, I’ve tested the following scenarios, and all are working as expected:

1. Providing both `--group-id` and `--repo` correctly returns the error: `{"error": "invalid config: you cannot specify both repositories and groups at the same time"}`

2. Successfully enumerated projects using:
   - A single group ID
   - Multiple group IDs
   - A single subgroup ID
   - A combination of multiple group and subgroup IDs

3. When only the `--repo` flag is used, scanning is limited to the specified repositories.

4. When no `--group-id` is provided, all accessible projects (based on the token) are scanned.

5. Verified fallback to legacy enumeration by disabling the simplified flag in `main.go`.

pkg/engine/gitlab.go
camgunz, line 49: I know--and rightly so--you're just reusing the existing pattern here, but these checks don't seem useful. For example: if `len(c.Endpoint) > 0`, then we won't set `connection.Endpoint`. But how will the Source know that `Endpoint` is set? That's right, by checking if `c.Endpoint` is empty [itself](https://github.com/trufflesecurity/trufflehog/blob/8ff3efd0ae372be1fb07dcaae55cdd7813f8fdf8/pkg/sources/gitlab/gitlab.go#L975).

So, these may as well just be raw assignments, e.g. `connection.Endpoint = c.Endpoint`. I *think* that's right? Could be wrong.

camgunz (approved):
Left a little comment, but it's purely optional.

camgunz:
Oh, I already approved so it's OK to not do this, but most of the new code in the Source doesn't seem like it's covered by the existing tests, so it would be nice to write 1 or more tests for it.

shahzadhaider1:
> Oh, I already approved so it's OK to not do this, but most of the new code in the Source doesn't seem like it's covered by the existing tests, so it would be nice to write 1 or more tests for it.

Thank you for the feedback. I’ll definitely work on adding appropriate test coverage for the new code. Appreciate you pointing it out!

martinlocklear (commented):
Hey! Thanks for tapping me in on this review.   Took a _brief_ look but it seems like you've gotten good review coverage already.  I'll just echo the general sentiment about test coverage.

Go team!

PR #4322, opened 2025-07-19
feat: add webexbot support

### LINKED ISSUE #4321: Add webex bot support
## Description
I would like support for [Webex BOT](https://developer.webex.com/messaging/docs/bots) tokens. 

### Preferred Solution
Find Webex BOT tokens, test them and return the verified ones.

jonathongardner:
### Description:
Find and verify webex bot tokens. Fixes #4321

### Checklist:
* [x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))? This was failing for me locally on `unknown flag: --out-format` without the flag still a lot of errors. Tested just on the added detector and it passed 

webexbot-detector-tests


pkg/detectors/webexbot/webexbot.go
```diff
@@ -0,0 +1,123 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var keywords = []string{"webexbot", "bot", "token", "api", "spark", "webex"}
+var (
+	defaultClient = common.SaneHttpClient()
+	keyPat        = regexp.MustCompile(detectors.PrefixRegex(keywords) + `([a-zA-Z0-9]{64}_[a-zA-Z0-9]{4}_[a-zA-Z0-9]{8}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{12})`)
```
jonathongardner: The token seems pretty unique so not sure if the prefixregex is needed but the boilerplate had it

pkg/detectors/webexbot/webexbot_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+func TestWebexbot_FromChunk(t *testing.T) {
```
jonathongardner, line 20: I followed the [docs](https://trufflesecurity.com/blog/improving-trufflehog-part-i-adding-a-new-detector) and created the env. The test was passing but when i built the Th and tried running it on the file to see if it would find the cred nothing came back (verified or unverified) not sure if I'm missing something some where b/c of that

pkg/pb/detectorspb/detectors.pb.go
```diff
@@ -1138,6 +1138,7 @@ const (
 	DetectorType_BitbucketAppPassword                    DetectorType = 1029
 	DetectorType_Hasura                                  DetectorType = 1030
 	DetectorType_SalesforceRefreshToken                  DetectorType = 1031
+	DetectorType_WebexBot                                DetectorType = 1032
```
kashifkhan0771: Hi @jonathongardner 
We don't manually edit this file. To add a new detector type, please update the `/proto/detectors.proto` file and then run the `make protos` command. This will automatically regenerate and update the file accordingly.

kashifkhan0771 (changes requested):

pkg/detectors/webexbot/webexbot.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+type Scanner struct {
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var keywords = []string{"webexbot", "bot", "token", "api", "spark", "webex"}
```
kashifkhan0771: The keyword `webex` should be enough. Adding common words like `token`, `api`, or `bot` as keywords may lead to false positives. If you still want to include such generic terms, they should be added separately in prefix regex.

pkg/detectors/webexbot/webexbot_integration_test.go
jonathongardner: For example i tried:
```
echo "WEBEXBOT=ajlksdjasda9090sadsa9dsad0saasdkl0asd90asdcasc90asdajij2n2njkjn3_asdf_asdkqw34-qwer-vbnm-asdf-qwertyuiopkl" | go run main.go stdin
> finished scanning	{"chunks": 1, "bytes": 116, "verified_secrets": 0, "unverified_secrets": 0, ...
```
so it seems i missed something?

pkg/detectors/webexbot/webexbot_integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
```
shahzadhaider1:
```suggestion
	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors6")
```

Please use this. We'll add the required credentials for tests to pass

pkg/detectors/webexbot/webexbot.go
shahzadhaider1: I agree, the pattern is not very common. We should be fine without using the `PrefixRegex` because the pattern is quite unique with fixed length.

pkg/detectors/webexbot/webexbot.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	defer func() {
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		// parse the response body to json
+		var resp response
+		toReturn := make(map[string]string)
```
shahzadhaider1:
```suggestion
		extraData := make(map[string]string)
```

pkg/detectors/webexbot/webexbot_integration_test.go
shahzadhaider1: I'll check this

shahzadhaider1 (changes requested):
Great work. Thank you.
Added some feedback.

pkg/detectors/webexbot/webexbot_test.go
shahzadhaider1, line 1: Please make sure all the pattern tests are passing

pkg/detectors/webexbot/webexbot_integration_test.go
shahzadhaider1: You forgot the fourth step:


Add the Secret Detector to TruffleHog's Default Detectors
Add the secret scanner to the [pkg/engine/defaults.go](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/defaults.go) file like [github.com/trufflesecurity/trufflehog/v3/pkg/detectors/<detector_name>](https://github.com/trufflesecurity/trufflehog/blob/b71ea27a696bdf1c3141f637fda4ee4936c2f2d6/pkg/engine/defaults.go#L9) and [<detector_name>.Scanner{}](https://github.com/trufflesecurity/trufflehog/blob/b71ea27a696bdf1c3141f637fda4ee4936c2f2d6/pkg/engine/defaults.go#L1546)

webexbot-detector

shahzadhaider1: Please pull the latest

jonathongardner: oh that makes sense, thanks!

shahzadhaider1 (approved):
Thank you for the contribution.

pkg/detectors/webexbot/webexbot.go
```diff
@@ -0,0 +1,122 @@
...
+
+var (
+	defaultClient = common.SaneHttpClient()
+	keyPat        = regexp.MustCompile(`([a-zA-Z0-9]{64}_[a-zA-Z0-9]{4}_[a-zA-Z0-9]{8}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{12})`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"webexbot", "spark", "webex"}
```
kashifkhan0771: `webex` should be sufficient, since `webexbot` already includes the word `webex`, making it unnecessary to include both.

pkg/detectors/webexbot/webexbot.go
```diff
@@ -0,0 +1,122 @@
...
+	defer func() {
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		// parse the response body to json
+		var resp response
+		extraData := make(map[string]string)
+		if err := json.NewDecoder(res.Body).Decode(&resp); err == nil {
```
kashifkhan0771: We should handle the error

kashifkhan0771 (changes requested):

pkg/detectors/webexbot/webexbot.go
jonathongardner: makes sense ill remove it

pkg/detectors/webexbot/webexbot.go
jonathongardner: Sounds good, I was just following the github v1 which seems to only return the extra data if json decode doesnt fail. https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/github/v1/github_old.go#L136

just return it and let it be set on the `SetVerificationError`?

kashifkhan0771 (approved):
Thank you @jonathongardner ❤️

Issue #3671, opened 2024-11-25
GitLab v1 detector: incorrect pattern resulting in false positives, 429 rate limits


rgmz:
The pattern for GitLab v1 tokens is erroneous. `\b` can only be used next to `a-zA-Z0-9_`, for any other characters it will either not match or match incorrectly.
https://github.com/trufflesecurity/trufflehog/blob/1e5aac44955de69cdcfdf30eb8b279bc1d0a477f/pkg/detectors/gitlab/v1/gitlab.go#L36

**Examples**

Given the following pattern:
```regexp
gitlab.+=\b([a-zA-Z0-9\-=_]{20,22})\b
```

This will not match at all:
```
gitlab_token=-abcdefghijklmnopqrst
```

This will not match the `-` at the end:
```
gitlab_token=abcdefghijklmnopqrst-
```

**Real world example**
 Almost every GitLab URL will be falsely detected.
```
Found unverified result 🐷🔑❓
Verification issue: unexpected HTTP response status 429
Detector Type: Gitlab
Decoder Type: PLAIN
// Line: `// https://docs.gitlab.com/ce/api/services.html#get-drone-ci-service-settings`
Raw result: get-drone-ci-service-
Version: 1
Line: 169
Link: https://github.com/gocardless/atlantis/blob/fa43febf60e6585cf324164be9aa1391a09d7d6f/vendor/github.com/lkysow/go-gitlab/services.go#L169
```

kashifkhan0771:
This has been addressed in [this pull request by @abmussani](https://github.com/trufflesecurity/trufflehog/pull/4156). Regarding the scenario you mentioned that the detection fails if the token ends with a `-`. I believe GitLab tokens do not end with a `-` in practice. Have you encountered a real token with a trailing `-` that wasn't detected?

kashifkhan0771:
I’m closing this ticket for now. Please feel free to reopen it if you have any further concerns.

PR #4323, opened 2025-07-21
Refactored Gitlab Detectors

kashifkhan0771:
### Description:
Currently, we only verify the token against the first endpoint in the list, which is incorrect. Instead, we should attempt verification against all configured endpoints. This PR fixes this issue and refactored the code a little bit.

### Checklist:


pkg/detectors/gitlab/v1/gitlab.go
```diff
@@ -88,74 +112,51 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+func VerifyGitlab(ctx context.Context, client *http.Client, baseEndpoint, resMatch string) (bool, map[string]string, error) {
 	// there are 4 read 'scopes' for a gitlab token: api, read_user, read_repo, and read_registry
 	// they all grant access to different parts of the API. I couldn't find an endpoint that every
 	// one of these scopes has access to, so we just check an example endpoint for each scope. If any
 	// of them contain data, we know we have a valid key, but if they all fail, we don't
 
-	client := s.client
-	if client == nil {
-		client = defaultClient
+	req, err := http.NewRequestWithContext(ctx, "GET", baseEndpoint+"/api/v4/user", nil)
```
shahzadhaider1: Non-blocking comment:
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, baseEndpoint+"/api/v4/user", http.NoBody)
```

pkg/detectors/gitlab/v1/gitlab.go
```diff
@@ -88,74 +112,51 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-			// Nothing to do; zero values are the ones we want
-			return false, nil, nil, nil
-		default:
-			return false, nil, nil, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	// 200 means good key and has `read_user` scope
+	// 403 means good key but not the right scope
+	// 401 is bad key
+	switch res.StatusCode {
+	case http.StatusOK:
+		return json.Valid(bodyBytes), nil, nil
```
shahzadhaider1, line 144: if 200 means good key and has `read_user` scope, then why the additional `json.Valid` check?

kashifkhan0771: I'm not entirely sure why this was added, but it appears to be an extra check to ensure the response is in JSON format, rather than something like HTML or another unexpected type.

shahzadhaider1: Alright, let's leave it unchanged.

shahzadhaider1 (approved):
GW

kashifkhan0771:
What if I try to fix this issue https://github.com/trufflesecurity/trufflehog/issues/3671 as part of this PR too 🤔

kashifkhan0771:
> What if I try to fix this issue #3671 as part of this PR too 🤔

Will make a separate PR for that.

PR #4325, opened 2025-07-21
Support sub-unit resumption in filesystem scans

mcastorina:
Additionally adds utility methods to the `Progress` struct to make concurrent access to the `EncodedResumeInfo` more ergonomic.

### Description:
Adds support for sub-unit resumption to the filesystem source by utilizing the existing `EncodedResumeInfo` field.

### Checklist:

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -120,7 +120,12 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sources.Chunk) error {
 	workerPool := new(errgroup.Group)
 	workerPool.SetLimit(s.concurrency)
-	defer func() { _ = workerPool.Wait() }()
+	defer func() {
+		_ = workerPool.Wait()
+		s.ClearEncodedResumeInfoFor(path)
+	}()
+	startState := s.GetEncodedResumeInfoFor(path)
+	resuming := startState != ""
 
 	return fs.WalkDir(os.DirFS(path), ".", func(relativePath string, d fs.DirEntry, err error) error {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


camgunz (commented):
(I think I might be late to the sub-unit resumption party--feel free to direct me elsewhere if that's true)

I was sort of thinking sub-unit resumption would like, kind of unify the Smarter Scans / non-Smarter Scans worlds where `Chunks` would usually be implemented in terms of `ChunkUnit` and `ChunkUnit` would use some kind of interface on `SourceUnit` to do sub-unit resumption. This here is fine of course, but, part of me kinda worries we're signing ourselves up for a lot of bespoke work to support this functionality in each Source.

juvoni:
I agree. We should try to unify them now instead of maintaining both
pathways at this level.

rosecodym:
@mcastorina would mind adding more detail about how these new utility functions work, and how this will work in both a Smarter Scans and a legacy scan context? A map isn't what I intuitively expected am I'm curious about where I went wrong.

mcastorina:
> @mcastorina would mind adding more detail about how these new utility functions work, and how this will work in both a Smarter Scans and a legacy scan context? A map isn't what I intuitively expected am I'm curious about where I went wrong.

Yes, thinking about the legacy scan context is what drove me to use a map here.

For Smarter Scans, a map isn't needed. A scan job in this context has a 1:1 relationship with an instantiated source, so the `EncodedResumeInfo` field is owned entirely by the unit.

Legacy scans are a bit harrier. In this context, the `EncodedResumeInfo` field is part of one instantiated source that has `ChunkUnit` called multiple times and concurrently. So multiple units are sharing the `EncodedResumeInfo` field, which requires synchronization and a data structure to organize them.

The new utility methods added to `Progress` adds a layer on top of `EncodedResumeInfo` so the source doesn't have to synchronize / marshal the data for storage.

rosecodym (approved):

PR #4326, opened 2025-07-21
Utf16 bom support

joeleonjr:
### Description:
Secrets in UTF-16-Encoded Files are _not always_ detected due to data chunk changes made in the UTF-8 extractSubstrings() function.

In [engine.go](https://github.com/trufflesecurity/trufflehog/blob/cf169cca83c943af2f636ce46022c16a12675382/pkg/engine/engine.go#L796C3-L796C39), TH loops each decoder, passing the Chunk Data in for processing. The UTF-8 decoder runs first. If the data chunk is invalid UTF-8, the UTF-8 decoder will execute the function [extractSubstrings()](https://github.com/trufflesecurity/trufflehog/blob/cf169cca83c943af2f636ce46022c16a12675382/pkg/decoders/utf8.go#L45). The result of that function is applied to the Chunk's Data field, which is then passed into all subsequent decoders. Part of that function alters the data structure of valid UTF-16 data, making detecting _some_ secrets impossible. 

Here's an example to test out:

```
echo <VALID_DETECTABLE_SECRET> > secret.txt
printf '\xFF\xFE' > utf16le.txt && iconv -f UTF-8 -t UTF-16LE secret.txt >> utf16le.txt
printf '\xFF\xFE' > utf16le.txt && iconv -f UTF-8 -t UTF-16LE secret.txt >> utf16le.txt
trufflehog filesystem utf*
```

Originally, I thought the problem was we did not address the UTF-16 Byte Order Marks (BOM) #FEFF and #FFFE. However, the existing logic takes care of those in the `utf16ToUTF8` function in `utf16.go`. I added two test cases to prove that. 

The only change needed is creating a copy of the chunk prior to processing each decoder. 

If that change is too expensive, I have 2 other ideas:
1. Move `extractSubstrings` out from the UTF-8 decoder and invoke it directly `engine.go` prior to running `FindDetectorMatches` during a failed `UTF-8` decode. 
2. Store the results of that function in a separate variable for later processing in `FindDetectorMatches`.


zricethezav:
@joeleonjr lgtm

zricethezav (approved):

shahzadhaider1 (approved):

amanfcp (approved):

PR #4330, opened 2025-07-22
Add slack continuous protobuf messages

camgunz:
### Description:
This adds new protobuf messages used by a new Slack Continuous source in enterprise

### Checklist:


dustin-decker (approved):
Should it be called SlackRealtime?

camgunz:
> Should it be called SlackRealtime?

Good Q! Martin and Shani put their heads together and IIRC while the external name is "Slack Realtime" it's a little confusing to use that name on our end because of the existing Source. I'm... pretty sure that's all correct; to be honest Martin told me to use "Continuous" and I've since defenestrated all of the underlying reasoning out of my brain.

PR #4332, opened 2025-07-24
Expanded test coverage for binary content

kashifkhan0771:
### Description:
This PR adds some test cases for git diff with binary content of different types.

### Checklist:


pkg/gitparse/gitparse_test.go
```diff
@@ -778,8 +780,13 @@ func assertDiffEqualToExpected(t *testing.T, expected *Diff, actual *Diff) {
 		assert.NoError(t, err)
 		assert.Equal(t, expectedDiffStr, actualDiffStr)
 	}
-	// TODO - Add test coverage for binary diffs (if it isn't already elsewhere)
 
+	if actual.IsBinary {
+		if actual.contentWriter != nil {
```
kashifkhan0771: It appears that for binary content, the actual content is always nil by default. 😕

ahrav: Yes, that's a correct. A git diff focuses on showing line-by-line changes, which is a concept that doesn't apply to binary files (like images, compiled code, or PDFs).

Instead of generating a meaningless content diff, Git simply records that the binary file has been modified. To access the actual raw content of the file, a different approach is needed. The standard command is `git cat-file`, which lets us stream the data directly for scanning [here](https://github.com/trufflesecurity/trufflehog/blob/a05cf0859455b5b16317ee22d809887a4043cdf0/pkg/sources/git/git.go#L1297).

kashifkhan0771: Yeah, I read a bit about this that by default, git diff doesn’t render the content of binary files unless some configuration tweaks are made.

In that case I think these test cases are solid. We’re generating binary content and parsing it without any errors, which is essentially the goal here, right?

kashifkhan0771: ❓ Does the `git cat-file` command only return binary gibberish, or does it provide human-readable output?

rosecodym:
Correct me if I'm wrong: The test you modified doesn't actually parse the binary data, right? I can't find anywhere that your new, synthetic binary data is actually _read_ in the test code, so I'm concerned that it might mislead readers. (I ran the updated test locally after fiddling with the test data and everything still passed.)

kashifkhan0771:
> Correct me if I'm wrong: The test you modified doesn't actually parse the binary data, right? I can't find anywhere that your new, synthetic binary data is actually _read_ in the test code, so I'm concerned that it might mislead readers. (I ran the updated test locally after fiddling with the test data and everything still passed.)

Discussed locally. We need to check If we can find somewhere in the codebase that makes sense to test actual binary content.

kashifkhan0771:
> Discussed locally. We need to check If we can find somewhere in the codebase that makes sense to test actual binary content.

I reverted the previous changes in the `gitparse` package and added binary content tests to both the `chunker_test.go` and `filesystem` scan. These should serve as a solid foundation for expanding test coverage of binary content scanning.

kashifkhan0771:
I am planning to add some test cases to `git_test.go` as well.

camgunz (approved):
LG! Love this approach 👍🏻

PR #4334, opened 2025-07-25
Improve LarkSuite Detector Accuracy and Error Handling

### LINKED ISSUE #3669: LarkSuite detector: `unexpected verification response code 99991668`
**Example:**

```
Found unverified result 🐷🔑❓
Verification issue: unexpected verification response code 99991668, message Invalid access token for authorization. Please make a request with token attached.
Detector Type: LarkSuite
Decoder Type: PLAIN
Raw result: a-ca_n_4426410.html
Token_type: App Access Token
```

Docs: https://open.larksuite.com/document/server-docs/getting-started/server-error-codes

shahzadhaider1:
#### 1. False Positive Detection
The LarkSuite detector was incorrectly flagging HTML filenames and other file paths as valid tokens, causing false positives. 

Examples of incorrectly detected tokens:
- `a-ca_n_4426410.html` 
- `u-cb_n_4426410.html`
- `t-cc_n_4426410.html`

These matches occurred because the regex patterns allow underscores and periods (valid in actual LarkSuite tokens), but this also made them match filename patterns that coincidentally start with the expected token prefixes (`a-`, `u-`, `t-`).

#### 2. Incomplete Error Code Handling
The detector had insufficient error code coverage, leading to unexpected verification response codes like `99991668` and `20005` that weren't properly handled, causing verification failures instead of correctly identifying invalid tokens.

### Root Causes
1. **Regex over-matching**: The existing patterns `[a-z0-9A-Z_.]{14,50}` were too permissive for filesystem paths
2. **Missing error codes**: Documentation gaps and undocumented API response codes weren't accounted for

#### 1. False Positive Filtering
Implemented a post-processing filter using `isLikelyFilename()` function that:
- **Preserves token validity**: Keeps existing regex patterns unchanged to ensure all legitimate LarkSuite tokens (including those with dots and underscores) are still detected
- **Filters false positives**: Checks if detected tokens end with common file extensions and excludes them from results
- **Optimized performance**: Uses a map-based lookup (`O(1)`) for file extension checking
- **Comprehensive coverage**: Includes 40+ common file extensions across web files, documents, archives, images, source code, and configuration files

#### 2. Enhanced Error Code Handling
Implemented comprehensive error code classification with proper constants and handling:
- **Added documented error codes**: `20005` (invalid token), `20008` (user not exist), `20021` (user resigned), `20022` (user frozen), `20023` (user not registered)
- **Added undocumented error codes**: `99991663` and `99991668` (invalid access token - found through API testing but not in documentation)
- **Improved error classification**: Created `VerificationResult` enum and `classifyErrorCode()` function to properly distinguish between:
  - **Valid tokens** (`0`, `99991672`)
  - **Invalid tokens** (definitive authentication failures)
  - **Verification errors** (system errors, rate limits, permissions issues)
- **Better error handling**: Tokens are now correctly marked as "invalid" vs "unverified" based on the specific error code meaning
 
### Documentation References
- [LarkSuite User Info API](https://open.larksuite.com/document/uAjLw4CM/ukTMukTMukTM/reference/authen-v1/user_info/get)
- [LarkSuite Calendar API](https://open.larksuite.com/document/server-docs/calendar-v4/calendar/list)  
- [LarkSuite Tenant API](https://open.larksuite.com/document/server-docs/tenant-v2/tenant/query)

This fix significantly reduces false positive rates while improving error handling accuracy, resulting in more reliable LarkSuite token detection and verification.

### Checklist:


pkg/detectors/larksuite/larksuite.go
```diff
@@ -115,26 +148,143 @@ func verifyAccessToken(ctx context.Context, client *http.Client, url string, tok
...
-		}
+
+	// For LarkSuite API, we expect JSON responses for all documented status codes
+	var bodyResponse verificationResponse
+	if err := json.NewDecoder(res.Body).Decode(&bodyResponse); err != nil {
+		return false, fmt.Errorf("failed to decode response (status %d): %w", res.StatusCode, err)
+	}
+
+	// Handle based on error code classification
+	result := classifyErrorCode(bodyResponse.Code)
```
amanfcp, line 158: curious; what is the purpose of separately classifying errors and then returning here? 
2 different switch cases seem a bit off putting...

pkg/detectors/larksuite/larksuite.go
```diff
@@ -115,26 +148,143 @@ func verifyAccessToken(ctx context.Context, client *http.Client, url string, tok
...
+	}
+}
+
+// isLikelyFilename - filter out tokens that end with common file extensions
+func isLikelyFilename(token string) bool {
+	_, ok := commonFileExtensions[strings.ToLower(path.Ext(token))]
+	return ok
+}
+
+var commonFileExtensions = map[string]struct{}{
```
amanfcp, line 220: Suggestion to consider using the //go:embed directive. 
Like here -> https://github.com/trufflesecurity/trufflehog/blob/98c36a4b2a01d2baf0007840850ee5149502f5cf/pkg/detectors/falsepositives.go#L35-L36

amanfcp (changes requested):
Good work Shahzad. I've left some comments, feel free to discuss

pkg/detectors/larksuite/larksuite.go
shahzadhaider1: The purpose of error classification is to accurately determine the validity of a secret. Since the API returns a 200 HTTP status code in some cases, even for failed authentication attempts, we rely on internal response codes and messages within the response body (e.g., "invalid token", "user not exist") to interpret the actual outcome.

pkg/detectors/larksuite/larksuite.go
shahzadhaider1: @amanfcp Thanks for the suggestion! I considered using `go:embed`, but the existing .txt files don’t include the specific file extensions we want to filter out. Since this is a small, detector-specific list, I felt that introducing a separate file for extensions, just for this might be a bit excessive, so I opted for an inline map of known extensions for simplicity and maintainability. Happy to revisit if you think embedding is more appropriate.

shahzadhaider1: I have implemented `CustomFalsePositiveChecker` now with similar logic.

amanfcp: Just wondering if we should consider adding these extensions to the false positive list...

shahzadhaider1: Hmm, I also can't answer that. Maybe we can ask @kashifkhan0771 @abmussani 

FYI: I chose this approach because the list of file extensions is relatively small and unlikely to change often. Embedding a file wouldn’t offer a significant performance benefit here, and making the list configurable felt unnecessary at this point. The current `map[string]struct{}` approach gives us O(1) lookups, which is very fast and memory-efficient.

Also, I looked at example here:
https://github.com/trufflesecurity/trufflehog/blob/10f895d8997b1f3e47f10d0b5bb1a9ae90e6462c/pkg/detectors/slackwebhook/slackwebhook.go#L117

That said, I’m open to switching to file embedding if the team prefers that direction.

shahzadhaider1: > Just wondering if we should consider adding these extensions to the false positive list...

@amanfcp Regarding this point, I’m a bit hesitant to add these extensions to the global false positives list. That list is shared across all detectors, and adding entries to it could unintentionally suppress valid detections in other contexts. Since we can't be certain how other detectors might interpret or rely on those extensions, it feels safer to keep this logic scoped to this specific detector to avoid introducing unexpected behavior elsewhere.

kashifkhan0771 (approved):

amanfcp (approved):

PR #4347, opened 2025-07-27
feat(detectors): added make.com api_token and mcp_token detectors

Jeff-Rowell:
### Description:
[Make.com](https://www.make.com/en) is a low-code/no-code automation platform that allows users to connect apps and services typically to automate business workflows. I've added two new detectors for Make.com:

1. **api_token** - authentication keys permitting access to a make.com API. The [/users/me/current-authorization](https://developers.make.com/api-documentation/api-reference/users-greater-than-me#get-users-me-current-authorization) endpoint is used for validity checks. A 200 OK response code is returned iff the API key is valid *and* the correct FQDN/zone is used. The api_token detector attempts to match a Make.com API URL and tries to validate against the matched URLS first. If there are no API URLs found, or if the matched API URL(s) fail to validate, the api_token detector attempts validation using the [/users/me/current-authorization](https://developers.make.com/api-documentation/api-reference/users-greater-than-me#get-users-me-current-authorization) endpoint for each of the known make.com zones until one of them succeeds or they all fail:
	* https://eu1.make.com/api/v2/
	* https://eu2.make.com/api/v2/
	* https://us1.make.com/api/v2/
	* https://us2.make.com/api/v2/
	* https://us1.make.celonis.com/api/v2/
	* https://eu1.make.celonis.com/api/v2/
2. **mcp_token** - a unique URL provided by Make.com that allows external AI systems to access automation workflows and tools in Make.com. Make.com provides a cloud-based MCP server that communicates via Server-Sent Events (SSE) and provides users a URL that contains an embedded MCP auth token using a pattern that looks like this: `https://<MAKE_ZONE>/mcp/api/v1/u/<MCP_TOKEN>/sse`. The detector finds these URLs and looks for a 200 OK response when calling them.

### Checklist:


amanfcp:
Thank you @Jeff-Rowell for the contribution. 
Please resolve the conflicts so that we can go ahead with the review.

Jeff-Rowell:
> Thank you @Jeff-Rowell for the contribution. Please resolve the conflicts so that we can go ahead with the review.

@amanfcp Thanks for taking a look. Merge conflicts have been resolved

Jeff-Rowell:
@amanfcp Anything else needed from me here?

amanfcp:
@Jeff-Rowell  We’re currently evaluating whether support for make.com aligns with our goals and provides enough value to justify ongoing maintenance.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,188 @@
...
+
+			results = append(results, s1)
+		}
+	}
+
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string, foundURLs []string) (bool, map[string]string, error) {
+	baseURLs := []string{
```
kashifkhan0771: This approach isn't quite right. Since we support multiple base URLs, we should implement the `EndpointCustomizer` interface, which allows users to configure these endpoints as needed via the command line.

The `EndpointCustomizer` interface provides flexibility to use found endpoints, configured endpoints, or a combination of both. Additionally, if no endpoints are found and none are configured, TruffleHog won't have anything to verify against and will not report anything in such cases.

For a reference implementation, please take a look at the Artifactory detector.

CC: @mcastorina (The EndpointCustomizer designer)

kashifkhan0771 (changes requested):

pkg/detectors/make/api_token/api_token.go
Jeff-Rowell: @kashifkhan0771 Thank you for your feedback. I've implemented the `EndpointCustomizer` interface as you suggested.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,160 @@
...
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"make"}) + `\b([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})\b`)
+	// Pattern to match Make.com URLs in the data
+	urlPat = regexp.MustCompile(`\bhttps://(eu[12]|us[12])\.make\.(?:com|celonis\.com)/api/v2/`)
```
kashifkhan0771:
```suggestion
	urlPat = regexp.MustCompile(`\b(eu|us)[12]\.make\.(com|celonis)\.com`)
```

In my opinion, the regex should capture only what we actually need. We can prepend `https://` ourselves if needed. Also, `/api/v2/` isn’t part of the endpoint and doesn’t need to be included. We just need the endpoint for verification.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+func (Scanner) CloudEndpoint() string { return "" }
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"make"}
```
kashifkhan0771: I'm not sure why, but this doesn’t feel quite right to me 😅 maybe it’s because the keyword is too generic. Could we use `make.com` and `make.celonis.com` instead? I know the endpoint can be added manually via the CLI if it's not found, so I’ll leave the final call to the rest of the team.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,160 @@
...
+	dataStr := string(data)
+
+	uniqueMatches := make(map[string]struct{})
+	for _, match := range keyPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueMatches[match[1]] = struct{}{}
+	}
+
+	// Extract Make URLs from the data
+	var foundURLs []string
+	for _, match := range urlPat.FindAllStringSubmatch(dataStr, -1) {
```
kashifkhan0771: Since we're not using any capture groups in the URL pattern, we can use `FindAllString()` instead of `FindAllStringSubmatch()`. It will return a simple list of matched strings.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,160 @@
...
+		} else {
+			// No endpoints configured or found, return unverified result
+			s1 := detectors.Result{
+				DetectorType: detectorspb.DetectorType_MakeApiToken,
+				Raw:          []byte(match),
+			}
+
+			results = append(results, s1)
+		}
+	}
```
kashifkhan0771: If we don’t find any endpoints in the data chunk, or if none are provided in the configuration, we should skip creating any results.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string, endpoints []string) (bool, map[string]string, error) {
+	if len(endpoints) == 0 {
```
kashifkhan0771: Duplicate validation

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,160 @@
...
+					RawV2:        []byte(match + ":" + endpoint),
+				}
+
+				if verify {
+					client := s.client
+					if client == nil {
+						client = defaultClient
+					}
+
+					isVerified, extraData, verificationErr := verifyMatch(ctx, client, match, []string{endpoint})
```
kashifkhan0771: If we pass one endpoint in the parameter why are we passing it as a list?

pkg/detectors/make/api_token/api_token_test.go
```diff
@@ -0,0 +1,160 @@
...
+			for _, v := range test.want {
+				expected[v] = struct{}{}
+			}
+
+			if diff := cmp.Diff(expected, actual); diff != "" {
+				t.Errorf("%s diff: (-want +got)\n%s", test.name, diff)
+			}
+		})
+	}
+}
```
kashifkhan0771: Add an empty line at the end of the file.

kashifkhan0771 (changes requested):
I've reviewed about half of the PR and left some feedback. Once those are addressed, I'll go ahead and review the rest.

kashifkhan0771:
Thanks for the contribution @Jeff-Rowell ❤️

pkg/detectors/make/api_token/api_token.go
amanfcp: +1 
`make` keyword seems too generic

Jeff-Rowell:
> I've reviewed about half of the PR and left some feedback. Once those are addressed, I'll go ahead and review the rest.

@kashifkhan0771 Thanks for the review and recommendations. I believe I've got everything updated per your suggestions. Please take a look. Thank you 🙌

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,140 @@
...
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	uniqueMatches := make(map[string]struct{})
+	for _, match := range keyPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueMatches[match[1]] = struct{}{}
+	}
+
+	// Extract Make URLs from the data
+	foundURLs := urlPat.FindAllString(dataStr, -1)
```
kashifkhan0771, line 52: Keep the unique foundURLs only.

pkg/detectors/make/api_token/api_token.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		// Indeterminate failure (network error, etc.)
+		return false, nil, err
+	}
+	// Determinate failure (401)
+	return false, nil, nil
+}
+
+func tryURL(ctx context.Context, client *http.Client, baseURL, token string) (bool, error) {
+	// This endpoint returns 200 OK if the token is valid and the correct FQDN for the API is used.
```
kashifkhan0771: A couple of suggestions:

1. We can simplify the code by keeping just a single `VerifyMatch` function. It can handle making the request, sending it, and processing the response just like we usually do.
2. The second return value is always `nil`, so we can remove it entirely.

kashifkhan0771 (changes requested):
Sorry for the delayed review. I added few small comments. Rest of the PR looks great to me. Once resolved we can approve and merge. Thank you!

pkg/detectors/make/api_token/api_token.go
Jeff-Rowell: Done. I added a loop putting the found urls as keys in a map to ensure the urls are deduplicated before validation.

pkg/detectors/make/api_token/api_token.go
Jeff-Rowell: Thank you for the good suggestions, I've updated the code accordingly.

proto/detectors.proto
```diff
@@ -1045,7 +1045,9 @@ enum DetectorType {
   WebexBot = 1033;
   TableauPersonalAccessToken = 1034;
   Rootly = 1035;
-  HashiCorpVaultAuth = 1036;
```
kashifkhan0771, line 1048: Keep the HashiCorp as `1036`.

kashifkhan0771 (dismissed):
Just one non-blocker comment. LGTM!

proto/detectors.proto
shahzadhaider1: +1

pkg/detectors/make/api_token/api_token_integration_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+func TestMake_FromChunk(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
```
shahzadhaider1: Could you update this to `detectors6` instead? We'll add some secrets to the secret manager so the tests can pass.

pkg/detectors/make/api_token/api_token_integration_test.go
```diff
@@ -0,0 +1,163 @@
...
+		want                []detectors.Result
+		wantErr             bool
+		wantVerificationErr bool
+	}{
+		{
+			name: "found, verified",
+			s:    Scanner{},
+			args: args{
+				ctx:    context.Background(),
+				data:   []byte(fmt.Sprintf("You can find a make secret %s and endpoint us2.make.com", secret)),
```
shahzadhaider1: Could we also have this retrieve the endpoint from the secret manager instead of hard-coding it?

pkg/detectors/make/mcp_token/mcp_token.go
```diff
@@ -0,0 +1,102 @@
...
+	case http.StatusOK:
+		return true, nil, nil
+	case http.StatusUnauthorized:
+		// Determinate failure (401)
+		return false, nil, nil
+	default:
+		// Any other status code is an indeterminate failure
+		return false, nil, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
+}
```
shahzadhaider1: We can remove the `map[string]string` return, as I don't see any extra data being returned here.

pkg/detectors/make/mcp_token/mcp_token_integration_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+)
+
+func TestMakemcptoken_FromChunk(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	testSecrets, err := common.GetSecret(ctx, "trufflehog-testing", "detectors5")
```
shahzadhaider1: Could you update this to `detectors6` instead? We'll add some secrets to the secret manager so the tests can pass.

shahzadhaider1 (commented):
Great work, thank you.
Minor feedback.

amanfcp (changes requested):
Could you please align this with our folder structure convention? Instead of creating nested folders you can do the following

```
/pkg
   /detectors
      /make
      /makemcptoken
```

pkg/detectors/make/api_token/api_token_integration_test.go
Jeff-Rowell: @shahzadhaider1 Definitely! Can you put this endpoint in the secret manager as `MAKE_API_ENDPOINT` or was it named something else? I've updated the code locally but the tests don't pass currently.

shahzadhaider1: Thanks. The tests won’t pass on your machine since they require access to the secret manager (to fetch credentials). We'll take care of that.

Jeff-Rowell: Gotcha, I'll define a `.env` file in the related directories for the tests then. Thanks!

Jeff-Rowell:
@amanfcp @shahzadhaider1 Thank you for your reviews. I believe I've successfully implemented the requested changes. Please give another review when you have a chance. Thank you 🙏

amanfcp (dismissed):
LGTM! Thanks @Jeff-Rowell

Jeff-Rowell:
@amanfcp Is there anything else I need to do on this to get it merged? Please let me know if I need to do anything else here

amanfcp:
Hey @Jeff-Rowell,
Everything looks good in the PR. However, we’re currently pausing the addition of new detectors. We’ll get this merged once we’ve refined our process for maintaining them.

amanfcp:
I ran some additional tests against corpora datasets. The new detectors do not seem to create lots of noise. This can safely be merged imo

image
image

amanfcp (changes requested):
Thanks for the patience @Jeff-Rowell 

Can you please resolve the merge conflicts? Also, I just noticed that the file naming is not in line with our conventions. 
For example, a main go file for make detector would be `make.go` (not `api_token.go`).
Similarly, that of makemcptoken detector would be `makemcptoken.go` (not `mcp_token.go`)

Ideally, the file naming should be like below..

```
pkg/detectors/makeapitoken/makeapitoken.go
pkg/detectors/makeapitoken/makeapitoken_integration_test.go
pkg/detectors/makeapitoken/makeapitoken_test.go
pkg/detectors/makemcptoken/makemcptoken.go
pkg/detectors/makemcptoken/makemcptoken_integration_test.go
pkg/detectors/makemcptoken/makemcptoken_test.go
```

Jeff-Rowell:
@amanfcp Thanks for taking a look, I know you are super busy. Should be all set now.

kashifkhan0771 (approved):
Need to resolve conflicts - Rest looks fine to me.

PR #4354, opened 2025-07-29
Fixed Aha detector

kashifkhan0771:
### Description:
Currently, the Aha detector only uses the first URL it finds for verification, which is incorrect. This PR fixes that behavior and adds RawV2, as the same secrets can be verified against different URLs.

### Checklist:


shahzadhaider1 (approved):

amanfcp (approved):
LGTM

PR #4358, opened 2025-07-30
Missed Smartsheet Tokens

shahzadhaider1:
### Description:
This PR updates the Smartsheet API token detector to support both the current and legacy token formats.

### Background
TruffleHog’s existing Smartsheet detector only recognizes the newer 37-character tokens. However, per [Smartsheet's documentation](https://help.smartsheet.com/articles/2482389-generate-API-key), legacy 26-character tokens are still valid and in use.

### Changes
- Updated the regex pattern to match exactly 26-character or 37-character alphanumeric tokens.
- Avoids matching token lengths in between (e.g., 30-char strings), keeping detection precise and avoiding false positives.
- Added pattern test cases for the updated length.

Reference: https://help.smartsheet.com/articles/2482389-generate-API-key 

### Checklist:

### DIFF
--- a/pkg/detectors/smartsheets/smartsheets.go
+++ b/pkg/detectors/smartsheets/smartsheets.go
@@ -22,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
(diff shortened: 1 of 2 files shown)


kashifkhan0771 (approved):
LGTM!

PR #4362, opened 2025-08-01
Oss 133 new detector vault approle auth for hashicorp

SyedAliHamad:
### Description:
This PR introduces a new detector for HashiCorp Vault AppRole authentication credentials.

Detection:
Keyword Filtering: scans for related keywords to identify relevant data chunks
Multi-Component Pattern Matching: Uses regex to find three required components:

- role_id: hexadecimal UUID format
- secret_id: hexadecimal UUID format
- vault_url: HashiCorp Cloud URLs matching *.hashicorp.cloud pattern

all 3 components creates verifiable combination results

Deduplication: Ensures each unique credential combination is tested only once to prevent redundant API calls

Verification:
Validates credentials against Vault's /v1/auth/approle/login endpoint using detected URLs

200 OK: Credentials are valid and active
401/403: Invalid credentials or insufficient permissions
404 Not Found: AppRole auth method not enabled

Testing:
Pattern Tests: Comprehensive validation of formats, keyword detection, and URL pattern matching
Integration Tests: Live API verification with both active and inactive credentials

image

### Checklist:


pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+
+	secretIdPat = regexp.MustCompile(detectors.PrefixRegex([]string{"secret_id", "secret-id", "secretid"}) + `\b([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})\b`)
+
+	// Vault URL pattern - HashiCorp Cloud or any HTTPS/HTTP Vault endpoint
+	vaultUrlPat = regexp.MustCompile(`https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?(?:\/[^\s]*)?`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+func (s Scanner) Keywords() []string {
+	return []string{"role_id", "secret_id", "vault", "approle", "hashicorp"}
```
amanfcp: I think this will catch a lot of unnecessary secrets, just adding `hashicorp` seems fine to me

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	// create combination results that can be verified
+	for _, roleId := range roleIds {
+		for _, secretId := range secretIds {
+			for _, vaultUrl := range vaultUrls {
+				s1 := detectors.Result{
+					DetectorType: detectorspb.DetectorType_HashiCorpVaultAuth,
+					Raw:          []byte(secretId),
+					RawV2:        []byte(fmt.Sprintf("role_id:%s,secret_id:%s,vault_url:%s", roleId, secretId, vaultUrl)),
```
amanfcp:
```suggestion
					RawV2:        []byte(fmt.Sprintf("%s:%s", roleId, secretId)),
```

We can add vault url in extra data.

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+				if verify {
+					client := s.client
+					if client == nil {
+						client = defaultClient
+					}
+
+					isVerified, extraData, verificationErr := verifyMatch(ctx, client, roleId, secretId, vaultUrl)
+					s1.Verified = isVerified
+					s1.ExtraData = extraData
+					if verificationErr != nil {
```
amanfcp: If check is not needed here. [See this](https://github.com/sahil9001/trufflehog/blob/56543de25d2a79b70f9736566f470ce1b9c603a2/pkg/detectors/detectors.go#L134-L138)

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		return true, extraData, nil
+	case http.StatusBadRequest:
+		return false, extraData, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return false, extraData, nil
+	case http.StatusNotFound:
+		return false, extraData, nil
```
shahzadhaider1: We can combine the cases if all return the same thing.

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+	default:
+		extraData["error"] = fmt.Sprintf("unexpected_status_%d", res.StatusCode)
+		return false, extraData, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
```
shahzadhaider1: I’m not certain that including the status code in the extra data adds much value, given that it’s already conveyed through the error return.

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+		"secret_id": secretId,
+	}
+
+	jsonPayload, err := json.Marshal(payload)
+	if err != nil {
+		return false, nil, err
+	}
+
+	// Construct the AppRole login endpoint from the base Vault URL
+	loginEndpoint := strings.TrimSuffix(vaultUrl, "/") + "/v1/auth/approle/login"
```
shahzadhaider1: Is there a common endpoint for login, like for example `https://login.hashicorp.cloud`, or we can't do much without the domain/instance?

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, loginEndpoint, bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		return false, nil, err
+	}
```
shahzadhaider1: We're currently using a login endpoint, which is a POST request and could potentially modify server-side resources, for example, by generating a new session, token, or JWT (though I might be mistaken here). Ideally, we should prefer using a GET endpoint that lists existing resources without altering state, to ensure the verification remains non-intrusive and idempotent.

shahzadhaider1 (changes requested):

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+	return []string{"role_id", "secret_id", "vault", "approle", "hashicorp"}
+}
+
+// FromData will find and optionally verify HashiCorp Vault AppRole secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	roleIds := extractRoleIds(dataStr)
+	secretIds := extractSecretIds(dataStr)
+	vaultUrls := extractVaultUrls(dataStr)
```
amanfcp: These functions and added checks are not required. Better to follow detector conventions.

```
	uniqueSecretIds := make(map[string]struct{})

	for _, match := range secretIdPat.FindAllStringSubmatch(data, -1) {
		uniqueSecretIds[match[1]] = struct{}{}
	}
	return uniqueSecretIds
```

Let me know if I'm missing something here

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+	return results, nil
+}
+
+// extractRoleIds finds all potential role_ids in the data
+func extractRoleIds(data string) []string {
+	var roleIds []string
+	uniqueRoleIds := make(map[string]struct{})
+
+	for _, match := range roleIdPat.FindAllStringSubmatch(data, -1) {
+		if len(match) >= 2 {
```
amanfcp: Regex does this for you

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
amanfcp: Can this trimming be done through vaultUrlPat regex? Something like `https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?(?:\/[^\s\/][^\s]*)?`

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
amanfcp: 1. In case of the secrets being invalid, ideally platforms return a `401` and sometimes other status code but rarely does it send multiple status codes. We should identify what status code the API returns when the secret credentials are invalid. In case of rest of the status codes, the function should return indeterminate result that you've mentioned in the default case
2. If the API does return different status codes in case of invalid credentials, these cases should be merged like so
```
	case http.StatusBadRequest, http.StatusNotFound, http.StatusUnauthorized, http.StatusForbidden:
		return false, extraData, nil
```

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+	case http.StatusOK:
+		return true, extraData, nil
+	case http.StatusBadRequest:
+		return false, extraData, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return false, extraData, nil
+	case http.StatusNotFound:
+		return false, extraData, nil
+	default:
+		extraData["error"] = fmt.Sprintf("unexpected_status_%d", res.StatusCode)
```
amanfcp: Not sure if we need to add error in extraData

amanfcp (changes requested):

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+
+	roleIdPat = regexp.MustCompile(detectors.PrefixRegex([]string{"role_id", "role-id", "roleid"}) + `\b([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\b`)
```
amanfcp:
```suggestion
	roleIdPat = regexp.MustCompile(detectors.PrefixRegex([]string{"role"}) + `\b([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\b`)
```

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,192 @@
...
+
+	secretIdPat = regexp.MustCompile(detectors.PrefixRegex([]string{"secret_id", "secret-id", "secretid"}) + `\b([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})\b`)
```
amanfcp:
```suggestion
	secretIdPat = regexp.MustCompile(detectors.PrefixRegex([]string{"secret"}) + `\b([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})\b`)
```

amanfcp (changes requested):

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
amanfcp: I don't think this will be a problem and that it changes anything on their end even though it's a POST request since the endpoint is of a login.

shahzadhaider1: Thanks for the clarification. 

My earlier comment was just out of caution, since POST requests can sometimes modify server-side state (e.g., invalidate current sessions/tokens and generate new sessions/tokens), and we generally aim for verification to be non-intrusive and idempotent. But if this endpoint doesn't cause side effects, then it should be totally fine to use.

SyedAliHamad: You're right, We do generate a token every time we login and test the secret_id but that's how it's intended to use. 
"SecretID is a credential that is required by default for any login". Please have a look at these docs and let me know if I missed something. Also what if we can't verify a key using GET are we supposed to not implement it?

https://developer.hashicorp.com/vault/docs/auth/approle
https://developer.hashicorp.com/vault/tutorials/auth-methods/approle

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
SyedAliHamad: Alll these are necessary in order to test the secret_id and role_id, should we not keep all in the RawV2?

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
shahzadhaider1: We don’t necessarily have to use a GET request. I suggested it because GET is typically used to retrieve resources without modifying anything. The main goal is to ensure we use an endpoint that doesn’t alter the state of the system during validation, so the verification process remains safe, non-intrusive, and idempotent.

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
SyedAliHamad: From what i have read each Vault instance (whether self-hosted, HCP Vault, or local dev) requires its own specific URL for API access, including AppRole authentication. We need to login to that instance using it's secret_id and role_id.

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -125,30 +125,32 @@ func verifyMatch(ctx context.Context, client *http.Client, roleId, secretId, vau
...
-			}
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return false, err
+		}
+
+		if strings.Contains(string(body), "invalid role or secret ID") {
 			return false, nil
 		} else {
 			return false, fmt.Errorf("bad request: %v", errorResponse.Errors)
```
amanfcp: nit;
```
		return false, fmt.Errorf("unexpected HTTP response status %d", resp.StatusCode)
```

amanfcp (approved):
LGTM! Thanks @SyedAliHamad

amanfcp:
@shahzadhaider1 can you please review this again?

pkg/detectors/hashicorpvaultauth/hashicorpvaultauth.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+
+	var vaultUrls []string
+	for _, match := range vaultUrlPat.FindAllString(dataStr, -1) {
+		url := strings.TrimSpace(match)
+		if !slices.Contains(vaultUrls, url) {
+			vaultUrls = append(vaultUrls, url)
+		}
+	}
```
shahzadhaider1: better to follow the convention for consistency.

shahzadhaider1 (approved):

PR #4368, opened 2025-08-06
Improve Bitfinex Detector - Implemented API-Based Secret Validation

shahzadhaider1:
### Description:
This PR improves the Bitfinex detector by replacing the Go SDK usage with a custom implementation using the standard `net/http` package to validate detected credentials against the Bitfinex v2 API `/auth/r/wallets`.

### Key Changes
- Removed the dependency on the Bitfinex Go SDK for secret validation.
- Re-implemented the validation logic using direct HTTP requests with correct headers (bfx-apikey, bfx-signature, bfx-nonce).
- Improved handling of API responses:
  - "apikey: invalid" -> clearly invalid key
  - "digest invalid" -> signature mismatch (likely secret issue)

### Testing
- Verified the behavior using known valid and invalid credentials.
- Successfully executed API calls with correct secrets and received 200 OK responses.
- Confirmed meaningful error responses for invalid combinations.

### Checklist:

bitfinex-detector-tests


pkg/detectors/bitfinex/bitfinex.go
```diff
@@ -1,44 +1,62 @@
...
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// related resource https://medium.com/@Bitfinex/api-development-update-april-65fe52f84124
 	apiKeyPat    = regexp.MustCompile(detectors.PrefixRegex([]string{"bitfinex"}) + `\b([A-Za-z0-9_-]{43})\b`)
 	apiSecretPat = regexp.MustCompile(detectors.PrefixRegex([]string{"bitfinex"}) + `\b([A-Za-z0-9_-]{43})\b`)
```
SyedAliHamad, line 35: Both patterns are identical. Can we optimize this further by using a single regex pattern and scanning the data only once? We're already testing all the combinations found anyway. Since this is an improvement PR, I thought we could take a look at this as well.

shahzadhaider1: Thanks for pointing that out. You're absolutely right, both regex patterns are currently identical, and optimizing this does make sense. This is something that I've also observed in many detectors. For example:
https://github.com/trufflesecurity/trufflehog/blob/bd525f89c3c01d85fd9261384b315bd7d292fda3/pkg/detectors/amplitudeapikey/amplitudeapikey.go#L28
https://github.com/trufflesecurity/trufflehog/blob/bd525f89c3c01d85fd9261384b315bd7d292fda3/pkg/detectors/boxoauth/boxoauth.go#L27


Since this PR is focused specifically on improving secret validation and removing the Bitfinex Go SDK dependency as mentioned in the ticket, I’ve kept the existing detection logic unchanged to keep the scope limited and focused. But this is definitely a good follow-up improvement, and I’d be happy to open a separate PR to clean this up afterward.

go.mod
```diff
@@ -30,14 +30,12 @@ require (
 	github.com/aws/smithy-go v1.22.5
 	github.com/aymanbagabas/go-osc52 v1.2.1
 	github.com/bill-rich/go-syslog v0.0.0-20220413021637-49edb52a574c
-	github.com/bitfinexcom/bitfinex-api-go v0.0.0-20210608095005-9e0b26f200fb
```
kashifkhan0771, line 33: ❤️ 🚀

kashifkhan0771 (approved):
GW - LGTM!

dustin-decker (approved):
Love RMing dependencies. Thanks!

PR #4372, opened 2025-08-07
Refactored caflou detector

kashifkhan0771:
### Description:
This PR refactors the Caflou token detector.
The existing regex pattern was too broad and for somehow had `a-b` only for small letters 😅, so this update narrows it to reduce false positives. I also observed that all Caflou tokens share a fixed prefix, which has now been incorporated into the pattern for improved accuracy.

### Checklist:


pkg/detectors/caflou/caflou.go
```diff
@@ -25,7 +26,7 @@ var (
 	defaultClient = common.SaneHttpClientTimeOut(time.Second * 10)
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"caflou"}) + `\b([a-bA-Z0-9\S]{155})\b`)
+	keyPat = regexp.MustCompile(`\b(eyJhbGciOiJIUzI1NiJ9[a-zA-Z0-9._-]{135})\b`)
```
shahzadhaider1, line 29: This means we will no longer be able to detect the old pattern secrets, right? 
Have we considered introducing a `v2` of this detector instead of updating the existing regex?

amanfcp:
@kashifkhan0771 this is already under review in #4315 

Please do give your review as per your findings there

kashifkhan0771:
> @kashifkhan0771 this is already under review in #4315
> Please do give your review as per your findings there

I reviewed that PR. It updates the verification logic.
This PR, on the other hand, focuses specifically on updating the regex pattern.

pkg/detectors/caflou/caflou.go
kashifkhan0771: V2 is not required. The pattern never changed. I just improved the pattern.

shahzadhaider1: This change isn’t just an “improvement”, it’s a narrowing of scope that risks missing secrets we could previously detect.

**Old pattern**: Could match any 155-character caflou-prefixed token, allowing for a broad set of possible formats. It used `\S`, so any non-whitespace character was acceptable.

**New pattern**: Only matches `HS256` JWTs starting with the fixed base64 header `eyJhbGciOiJIUzI1NiJ9`, and only allows `[a-zA-Z0-9._-]` in the remainder. This is a significant tightening: three fixed special characters instead of `\S`.

**Impact**: If old secrets weren’t JWTs or didn’t start with that header, they will no longer be matched.

The reason I suggested v2 is that we could handle the JWT-specific case separately. This way, we avoid unintentionally breaking coverage for previously detectable secrets.

With that said, I have trust in you that you have done your complete research/testing so I'd like to know what you think about this?

kashifkhan0771: There isn't really an old vs. new pattern rather, the previous regex was too broad and had some issues, such as using `a-b` instead of `a-z`. The overall length of the regex remains the same; I simply improved its accuracy.

I tested it with around 15–20 tokens, and the updated pattern successfully detected all of them.

shahzadhaider1 (approved):

PR #4383, opened 2025-08-11
Improved and fixed couchbase detector

kashifkhan0771:
### Description:
This PR improves and fixes the Couchbase detector.
**Improvements:**
* Refactored overall code structure for better readability.
* Enhanced regex patterns for matching Couchbase usernames and passwords.
* Separated the verification logic into its own function.
* Updated test cases for better coverage.

**Fixes:**
* The previous `meetsCouchbasePasswordRequirements` function had a bug: it returned the password as soon as all character requirements were met, without validating the entire string. For example, given the password `AB-1aiop`, it would incorrectly return `AB-1a`. The logic has now been corrected to validate the full string and updated to return only a boolean.
* The `Raw` field previously combined all three matches (connection string, username, and password), which could cause issues for the engine when trying to determine the line number. To address this, the connection string has been set as the primary secret to improve line number detection by the engine.

### Checklist:

### DIFF
--- a/pkg/detectors/couchbase/couchbase.go
+++ b/pkg/detectors/couchbase/couchbase.go
@@ -3,7 +3,6 @@ package couchbase
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 	"unicode"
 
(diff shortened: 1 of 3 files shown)


pkg/detectors/couchbase/couchbase.go
```diff
@@ -23,35 +22,18 @@ type Scanner struct {
...
 
 var (
-
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	connectionStringPat = regexp.MustCompile(`\bcb\.[a-z0-9]+\.cloud\.couchbase\.com\b`)
-	usernamePat         = `?()/\+=\s\n`
-	passwordPat         = `^<>;.*&|£\n\s`
-	// passwordPat         = regexp.MustCompile(`(?i)(?:pass|pwd)(?:.|[\n\r]){0,15}(\b[^<>;.*&|£\n\s]{8,100}$)`)
-	// passwordPat = regexp.MustCompile(`(?im)(?:pass|pwd)\S{0,40}?[:=\s]{1,3}[ '"=]{0,1}([^:?()/\+=\s\n]{4,40})\b`)
+	connectionStringPat = regexp.MustCompile(`\b(cb\.[a-z0-9]+\.cloud\.couchbase\.com)\b`)
```
shahzadhaider1, line 26: This looks like a domain/host to me. Could you explain why we have named this a connection string? Because a connection string is usually a text string that contains all the details a program needs to connect to a data source / database. Not a big deal, but it looks confusing to me.

kashifkhan0771: Screenshot from 2025-08-11 16-31-05

shahzadhaider1 (approved):

PR #4385, opened 2025-08-12
[Feature] Fixed And Updated Fibery Detecor

nabeelalam:
### Description:
This PR updates the Fibery detector which was using an incorrect regular expression for secrets. Furthermore, the regex for the domain has also been updated.

### Checklist:

### DIFF
--- a/pkg/detectors/fibery/fibery.go
+++ b/pkg/detectors/fibery/fibery.go
@@ -3,8 +3,8 @@ package fibery
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
-	"strings"
 	"time"
 
 	regexp "github.com/wasilibs/go-re2"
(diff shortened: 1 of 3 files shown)


pkg/detectors/fibery/fibery.go
```diff
@@ -83,6 +81,37 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
+func verifyMatch(ctx context.Context, client *http.Client, secret, domain string) (bool, error) {
+	timeout := 10 * time.Second
```
kashifkhan0771, line 85: Why are we overwriting default context timeout?

kashifkhan0771 (approved):
LGTM! Just one non-blocker comment.

shahzadhaider1 (approved):

pkg/detectors/fibery/fibery.go
nabeelalam: Hey @kashifkhan0771. I didn't come across any info about why this was originally added, nor did I find anything relevant in the Fibery API docs, so I have left the verification logic intact.

PR #4386, opened 2025-08-12
Fix legacy json flag for Github and Gitlab private repos

### LINKED ISSUE #831: Running trufflehog with flag --json-legacy against gitlab repo cause trufflehog to ask for username/password despite having a gitlab API token specified on cmd line
### TruffleHog Version
3.13.0

### Expected Behavior

Running trufflehog as:

`docker run -it -v "$PWD:/pwd" trufflesecurity/trufflehog:latest gitlab --token=[TOKEN] --json --repo [GITLAB REPO]`

returns output as expected

### Actual Behavior

However, running trufflehog as:

`docker run -it -v "$PWD:/pwd" trufflesecurity/trufflehog:latest gitlab --token=[TOKEN] --json-legacy --repo [GITLAB REPO]`

makes trufflehog ask for username and password repeatedly.

## Environment
 * OS: CentOS 8

shahzadhaider1:
### Description:
This PR fixes an issue with **scanning private Git repositories (GitHub/GitLab)** when using the **legacy JSON** output format.

Previously, the scan process attempted to clone the repository a second time during result printing. Since authentication was not passed along in this second clone, the process failed for private repositories. As a result, legacy JSON output could not be generated for private Git repositories.

### Fix
- Repositories are **now persisted after the initial scan** instead of being deleted immediately.
- When generating legacy JSON output, the scan reuses the **already cloned repositories**, eliminating the need for a second clone and preventing authentication errors.
- After results are printed, we perform a **cleanup step** to remove any cloned repositories and free up system resources.

### Implementation Details
- Added a new flag: `PrintLegacyJSON`.
- Sources check this flag, if true, they skip immediate cleanup after cloning/scanning.
- In `main.go`, if the `--jsonLegacy` flag is set, cleanup is performed at the end.

### Benefits
- Legacy JSON output now works reliably for **both public and private GitHub/GitLab** repositories.
- Avoids redundant cloning -> **improves performance**.
- Ensures **proper cleanup** after scan results are printed.

### Checklist:

source-legacy-json-bug


main.go
```diff
@@ -708,11 +709,29 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
...
+	// because it requires commit SHAs in the output.
+	persistRepo := *gitNoCleanup || *githubNoCleanup || *gitlabNoCleanup
+	if *jsonLegacy && !persistRepo {
+		if err := os.MkdirAll(tmpDir, os.ModePerm); err != nil {
+			return scanMetrics, fmt.Errorf("failed to create temporary directory: %v", err)
+		}
+		*gitNoCleanup, *githubNoCleanup, *gitlabNoCleanup = true, true, true
+		*gitClonePath, *githubClonePath, *gitlabClonePath = tmpDir, tmpDir, tmpDir
+	}
+
```
kashifkhan0771: I have a few concerns about this approach:

* First, we shouldn't be setting flags for sources that aren't being run. For example, if I'm only using the `git` source, this approach still sets flags for `gitlab` and `github` too. I haven't tested whether that causes issues, but it doesn't seem like a good idea. A cleaner solution would be to move this logic into each source block so that we only set flags relevant to the source that's actually running.

* Second, why are we creating a temp directory here? Let's say a user sets the `--clone-path` flag but doesn't set `--no-cleanup`. That means they want to use a custom path but don’t need to retain the repo afterward. However, the code hits the `if` condition at line 716, overwrites their `--clone-path` with a temp path, and clones the repo there instead. So we're kind of misleading the user here. They think we’re cloning to the path they specified, but we silently use a temp path behind the scenes which is not great.

* Third, if someone passes both `--no-cleanup` and `--clone-path`, then (as far as I can tell) this logic doesn't apply at all. So what happens in that case? Do we run into the same error again?

Right now, the only scenario where this seems to work cleanly is a basic scan with no `--clone-path` and no `--no-cleanup` set and in that approach we by default use temp path to clone repository and cleanup afterwards.

kashifkhan0771 (changes requested):

main.go
shahzadhaider1: >First, we shouldn't be setting flags for sources that aren't being run. For example, if I'm only using the `git` source, this approach still sets flags for `gitlab` and `github` too. I haven't tested whether that causes issues, but it doesn't seem like a good idea. A cleaner solution would be to move this logic into each source block so that we only set flags relevant to the source that's actually running.

I initially combined them into one block to avoid repeating if checks in each source block, but I’ve now updated the code to handle them separately. 



>Second, why are we creating a temp directory here? Let's say a user sets the `--clone-path` flag but doesn't set `--no-cleanup`. That means they want to use a custom path but don’t need to retain the repo afterward. However, the code hits the `if` condition at line 716, overwrites their `--clone-path` with a temp path, and clones the repo there instead. So we're kind of misleading the user here. They think we’re cloning to the path they specified, but we silently use a temp path behind the scenes which is not great.

Nice catch. I wasn't fully aware of the `--clone-path` functionality. I'll address this.

>Third, if someone passes both `--no-cleanup` and `--clone-path`, then (as far as I can tell) this logic doesn't apply at all. So what happens in that case? Do we run into the same error again?

That’s the expected behavior. If both `--no-cleanup` and `--clone-path` are set, there’s nothing additional we need to do, we simply rely on the provided clone path and use it for printing results.

shahzadhaider1: Thanks for your initial review. I’ve updated the implementation based on your feedback and included a list of the tests I ran along with the observed behavior. Please take a look when you get a chance

pkg/cleantemp/cleantemp.go
```diff
@@ -117,3 +117,27 @@ func CleanTempArtifacts(ctx logContext.Context) error {
 
 	return nil
 }
+
+// CleanTempDirsForLegacyJSON removes all directories that start with "trufflehog-"
+// from either the provided clonePath (if not empty) or the OS temp directory.
+func CleanTempDirsForLegacyJSON(baseDir string) error {
+	if baseDir == "" {
+		baseDir = os.TempDir()
+	}
```
camgunz, line 128: The way I read this, this provides a dummy directory to `rm -rf`. Is it easier to just `return` early here instead? I think in the logic in `main.go` if `clonePath` is `""` then there's nothing to do here anyhow.

Oh, OR (if that's true), in the `defer`-ed function we could just check for `clonePath == ""` and bail early there too.

pkg/cleantemp/cleantemp_test.go
```diff
@@ -28,3 +28,35 @@ func TestExecName(t *testing.T) {
 
 	assert.True(t, found)
 }
+
+func TestCleanTempDirsForLegacyJSON(t *testing.T) {
```
camgunz, line 32: 🙏🏻 Love it

main.go
```diff
@@ -709,16 +709,31 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
 	}
 	eng.Start(ctx)
 
+	persistRepo := *gitNoCleanup || *githubNoCleanup || *gitlabNoCleanup
+	clonePath := ""
```
camgunz: Optional nit, but since this is in a function that handles all kinds of non-git sources, naming these like `persistGitRepo` and `gitClonePath` or something could help indicate this is specifically for git things.

proto/source_metadata.proto
```diff
@@ -99,6 +99,7 @@ message Git {
   string repository = 4;
   string timestamp = 5;
   int64 line = 6;
+  string repository_local_path = 7;
```
camgunz, line 102: What's your thinking behind putting this info in the source metadata? My read is that we don't super need it for secrets so like, we could just not do it and save a fair amount of code.

pkg/sources/github/github.go
```diff
@@ -734,10 +735,12 @@ func (s *Source) cloneAndScanRepo(ctx context.Context, repoURL string, repoInfo
...
 	}
-
 	// remove the path only if it was created as a temporary path, or if it is a clone path and --no-cleanup is not set.
-	if strings.HasPrefix(path, filepath.Join(os.TempDir(), "trufflehog")) || (!s.conn.NoCleanup && s.conn.GetClonePath() != "") {
-		defer os.RemoveAll(path)
+	// if legacy JSON is enabled, don't remove the directory because we need it for outputting legacy JSON.
+	if !s.conn.GetPrintLegacyJson() {
+		if strings.HasPrefix(path, filepath.Join(os.TempDir(), "trufflehog")) || (!s.conn.NoCleanup && s.conn.GetClonePath() != "") {
+			defer os.RemoveAll(path)
+		}
```
camgunz, line 743: Eek, I didn't realize there was so much duplication between the git sources. OK that goes on the list haha 📝 

(to be clear, nothing to change here, just remarking)

camgunz (commented):
Just a Q about source metadata--will ping in Slack

camgunz (approved):
OK looks good! If it makes sense you can [do this thing](https://github.com/trufflesecurity/trufflehog/pull/4386/files#r2315308046), but totally up to you

proto/source_metadata.proto
camgunz: We have to do this so the legacy printer on the detector side can print the right thing

main.go
shahzadhaider1: done

pkg/cleantemp/cleantemp.go
shahzadhaider1: Thanks for pointing that out! In this case, `CleanTempDirsForLegacyJSON` is only called when the `jsonLegacy` flag is enabled. That also means the repos haven’t been deleted earlier, since they were needed for legacy JSON printing.

If the user provided a custom `clonePath`, we clean from there. But if `clonePath` is `""`, that means the repos were cloned into the system temp directory,  so we still need to clean those up. That’s why the function interprets `""` as `os.TempDir()` instead of returning early. I have added a comment explaining this so the next reviewer also knows the intent.

Please feel free to discuss if you think this can be improved.

camgunz: Oh that's great, for some reason I read `os.TempDir` and saw `os.MkdirTemp`, so this is perf 👍🏻

camgunz (approved):

shahzadhaider1:
@kashifkhan0771 please re-review.

kashifkhan0771 (approved):
LGTM! Awesome work ❤️

kashifkhan0771:
- [x] Make sure to test the changes with both Public/Private repositories for all git sources.

PR #4390, opened 2025-08-12
Update proto definitions for custom bitbucket oauth

casey-tran:
### Description:
Update the proto definitions found for Bitbucket sources to have OAuth option in the config file. Also regenerated the Protobuf files with the changes.

### Checklist:


martinlocklear (approved):
Approving to unblock, but it would be nice if there was some inline documentation here about what these fields are for.  The commit message also isn't entirely accurate right? Seems like you're adding fields _and_ regenerating protos.

PR #4400, opened 2025-08-15
docs: Add installation method with gah

marverix:
### Description:
Hello!
This PR adds to the documentation installation method using [gah](https://github.com/marverix/gah):
> gah is an GitHub Releases app installer, that DOES NOT REQUIRE SUDO! It is a simple bash script that downloads the latest release of an app from GitHub and installs it in ~/.local/bin. It is designed to be used with apps that are distributed as a single binary file.

### Checklist:
* [N/A] Tests passing (`make test-community`)?
* [N/A] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,22 @@ git clone https://github.com/trufflesecurity/trufflehog.git
 cd trufflehog; go install
 ```
 
+### Using gah (only Linux and macOS)
+
+Install the latest version:
+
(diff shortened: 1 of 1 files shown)


zricethezav:
Appreciate the PR but no thank you

PR #4403, opened 2025-08-15
Added explicit secrets manager write flag to Bitbucket source proto and generate new files

casey-tran:
### Description:
Added a field in the proto definition for Bitbucket sources to explicitly allow writing to the user's secrets manager. Initially this will be used to keep new and unused Bitbucket refresh tokens inside of the secret that we pull the config from. This process needed to happen in the first place because Bitbucket as an OAuth authorization server currently revokes any previously issued refresh tokens whenever the newest refresh token is issued with an access token, effectively making refresh tokens in the config file single-use.

### Checklist:


patidartanay:
Can you mention the issue ID in PR description, Thanks!

shahzadhaider1 (approved):

rosecodym (approved):

PR #4408, opened 2025-08-19
Enable cloning repository to a specified location with retention option

### LINKED ISSUE #1151: [Feature Request] Ability for user to retain scanned repos without deleting them after scan finishes
### Community Note

### Description

It would be great if for Git related scans such as Github and Gitlab that when trufflehog clones a repo there would be a flag such as `--retain /path/to/store/repos` or `--keep /path/to/store/repos` that when trufflehog finishes analyzing the repo it doesn't delete it and stores it in the target directory the user specifies.  Obviously this could lead to a substantial taxing on the hard drive storage; therefore, it would be an optional flag and never default. 

## Problem to be Addressed

Having the ability to retain repos that trufflehog already scanned without having to clone them again. Doing this would allow users to have the choice of using other tools such as [ripgrep](https://github.com/BurntSushi/ripgrep) or [git-secrets](https://github.com/awslabs/git-secrets) to further analyze repos in tandem with trufflehog, without having to clone them a second time.

Add a  `--retain /path/to/store/repos` or `--keep /path/to/store/repos` flag for Git related scans that when trufflehog is done analyzing the repo it doesn't delete it and the repos are cloned to the user specified directory. Allowing the user to specify a directory that if it does not exist trufflehog creates it and clones would be done there.

### LINKED ISSUE #3724: can I modify the path for trufflehog to store the temporary git info?
I believe "truffle git git-url" command will need to store the git repo/info into a local temporary path, so where is the path and what is the recommended way to modify it? thanks

### LINKED ISSUE #4382: High CPU usage from windows defender and no good way to exclude paths.
## Description
I run multiple instances of trufflehog in parallel on my windows machine using the "git" comand. 
Git keeps cloning repos constantly in a temp folder in C:\Users\<NAME>\AppData\Local\Temp\trufflehog-*
However, windows defender seems very interested in these files so its scans everything that gets cloned and has therefore a very high CPU usage. Unfortunately, windows defender seems to not support wildcard path exclusion. So, there is no way to exclude these folders except for excluding the whole Temp folder which I really do not want to do for obvious security reasons.

### Preferred Solution
Either createa parent folder with predictable name like this:
C:\Users\<NAME>\AppData\Local\Temp\trufflehog-clones\id1...id100...

Or, have an option to specify the folder which trufflehog uses to clone the repos.

kashifkhan0771:
### Description:
This PR introduces support for user-defined clone paths for Git-based sources.

### ✨ New Features:

* **`--clone-path`**: Allows users to specify a custom directory where repositories will be cloned during scanning.
* **`--no-cleanup`**: Prevents automatic cleanup of cloned repositories after scanning, allowing users to retain them in the specified `--clone-path`.

**Note:** `--no-cleanup` must be used with `--clone-path`.
By default, TruffleHog will continue to clone repositories to the system temporary directory and clean them up automatically after the scan.

These flags are supported across all Git-based sources: `git`, `gitlab`, and `github`.

### Related Issues:
The following related issues may be resolved once this PR is merged. There may be additional related issues as well. If you're aware of any, please feel free to link them here:

* [https://github.com/trufflesecurity/trufflehog/issues/4382](https://github.com/trufflesecurity/trufflehog/issues/4382)
* [https://github.com/trufflesecurity/trufflehog/issues/3724](https://github.com/trufflesecurity/trufflehog/issues/3724)
* [https://github.com/trufflesecurity/trufflehog/issues/1151](https://github.com/trufflesecurity/trufflehog/issues/1151)

### Test Cases:
Screenshot from 2025-08-19 16-46-56

### Checklist:


amanfcp (approved):
The changes look neat and lean. Thank you @kashifkhan0771

main.go
```diff
@@ -719,6 +725,15 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
 			}
 		}
 
+		// clone path is only supported for HTTPS repository URLs, not for local repositories.
```
rosecodym, line 728: Should we support this for SSH clones as well? (I also see elsewhere in the code that we might also consider supporting HTTP clones, but I doubt _anyone_ does that.) Is there a way to _exclude_ local repositories, rather than explicitly _including_ remote ones?

main.go
```diff
@@ -1120,3 +1151,31 @@ func isValidCommit(uri, commit string) bool {
...
+	}
+
+	if clonePath == "" {
+		return nil
+	}
+
+	info, err := os.Stat(clonePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("path provided to --clone-path: %s does not exist", clonePath)
```
rosecodym: I like using `%q` for interpolated values that can contain spaces because I find it makes the generated text easier to read, but that's only a suggestion!

pkg/engine/gitlab.go
```diff
@@ -58,6 +58,12 @@ func (e *Engine) ScanGitLab(ctx context.Context, c sources.GitlabConfig) (source
 		connection.IgnoreRepos = c.ExcludeRepos
 	}
 
+	if c.ClonePath != "" {
+		connection.ClonePath = c.ClonePath
+	}
+
+	connection.NoCleanup = c.NoCleanup
```
rosecodym, line 65: Oof, it'd be really nice if this git-flavored source setup looked like the GitHub one :/ Maybe we can do that in the future.

pkg/sources/git/git.go
```diff
@@ -1246,6 +1272,7 @@ func PrepareRepo(ctx context.Context, uriString string) (string, bool, error) {
 	case "file":
 		path = fmt.Sprintf("%s%s", uri.Host, uri.Path)
 	case "http", "https":
+		ctx.Logger().V(5).Info("HTTPS", "clone_path", clonePath)
```
rosecodym: This is not a very descriptive log message, and it appears to conflict with "http" in its case definition. Can we remove it, or at least make it more descriptive?

rosecodym (commented):
This looks really cool - my only concerns are minor and mentioned inline.

pkg/sources/git/git.go
kashifkhan0771: I put that log in while testing and just forgot to remove it 😅

rosecodym (approved):
Pretty cool!

PR #4411, opened 2025-08-20
Grafana Detectors update

davand01:
Updated grafana and grafanaserviceaccount and added a new detector for grafana api keys.

### Description:
Added one new detector for generic grafana api keys. Also updated the patterns and some of the logic for GLSA Service Account tokens and GLC tokens.

### Checklist:


pkg/detectors/grafana/grafana.go
```diff
@@ -24,7 +24,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	defaultClient = common.SaneHttpClient()
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(`\b(glc_eyJ[A-Za-z0-9+\/=]{60,160})`)
+	keyPat = regexp.MustCompile(`\b(glc_[A-Za-z0-9+/]{32,400}={0,2})\b`)
```
nabeelalam: Hi @davand01. I'm curious as to why you have removed `"_eyJ"` from the pattern and changed the length range? Seems to me that was tighter than the current regex.

pkg/detectors/grafanaapikey/grafanaapikey.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	for match := range uniqueMatches {
+		res := detectors.Result{
+			DetectorType: detectorspb.DetectorType_GrafanaAPIKey,
+			Raw:          []byte(match),
+		}
+
+		if verify {
+			res.SetVerificationError(fmt.Errorf("no grafana instance detected to verify against"), match)
```
nabeelalam, line 48: We don't seem to be verifying the matched credentials

nabeelalam (changes requested):

pkg/detectors/grafanaapikey/grafanaapikey.go
davand01: These kind of api keys are used for on-premise installations, so together with a custom endpoint that could be an IP address or a FQDN. Since it's hard to reliably identify it - there is no good way of verifying the matched credentials.

pkg/detectors/grafana/grafana.go
davand01: reducing the length range was because the base64 string can indeed be shorter than 60 characters, and we don't want to risk missing keys because they are too short for the regex. glc_eyJ actually makes sense. I removed eyJr on another match because we cannot be certain that the key always begins with `{"k`, but `{"` makes sense. 

`keyPat = regexp.MustCompile(`\b(glc_eyJ[A-Za-z0-9+/]{29,400}={0,2})\b`)
`

pkg/detectors/grafanaapikey/grafanaapikey.go
amanfcp: Verification logic has to exist in a detector. In docs, it is mentioned that Grafana can be hosted on a VPS or as a subdomain. [Gitlab detector](https://github.com/trufflesecurity/trufflehog/blob/1e8671c6de6d721636e672b3066c6773e4ab6d0b/pkg/detectors/gitlab/v1/gitlab.go#L25-L30) is the best example for this

pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
amanfcp, line 1: Thank you for updating SATs too

pkg/detectors/grafanaapikey/grafanaapikey.go
```diff
@@ -0,0 +1,63 @@
...
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(eyJrIjoi[A-Za-z0-9]{70,400}={0,2})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+// Grafana uses "eyJrIjoi" as a prefix for api keys, see for example.
+// https://github.com/grafana/pyroscope-dotnet/blob/0c17634653af09befa7bc07b2e1c420b5dc8578c/tracer/src/Datadog.Trace/Iast/Analyzers/HardcodedSecretsAnalyzer.cs#L173
+func (s Scanner) Keywords() []string {
+	return []string{"grafanaapikey", "eyJrIjoi"}
```
nabeelalam, line 29: We can use `"eyJrIjoi"` exclusively since it's always part of the key. A keyword match for `"grafanaapikey"` won't necessarily contain `eyJrIjoi` in its input.

pkg/detectors/grafana/grafana.go
nabeelalam: Makes sense. Thanks for updating this and also for the info.

pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
```diff
@@ -40,51 +44,67 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+		case resp.StatusCode >= 200 && resp.StatusCode < 300:
+			res.Verified = true
+		case resp.StatusCode == 401:
+			// determinately not verified
+		default:
+			res.SetVerificationError(fmt.Errorf("unexpected HTTP response status %d", resp.StatusCode), key)
+		}
+	} else {
+		res.SetVerificationError(err, key)
+	}
```
amanfcp, line 105: While we're at it, can we add indeterminate verification support here?
Here's an [example](https://github.com/trufflesecurity/trufflehog/blob/1e8671c6de6d721636e672b3066c6773e4ab6d0b/pkg/detectors/elevenlabs/v1/elevenlabs.go#L107-L120)

amanfcp (changes requested):

pkg/detectors/grafanaapikey/grafanaapikey.go
davand01: Doh. PEBKAC.... I'll update that :P

PR #4413, opened 2025-08-21
Added support for additional validation rules in custom detector

### LINKED ISSUE #4402: Regex in config not matching passwords with all required elements
I have a regex config file for TruffleHog that I’ve created, but the scan is not finding results that meet all my intended criteria for a password.

The goal is to detect strings that contain:
- At least one uppercase letter
- At least one lowercase letter
- At least one number
- At least one special character
- Minimum length of 12 characters

**Current Behavior:**
With my current regex, the scan returns results for any string that has *any* of these character classes and is 12 characters or longer, but it does not require that **ALL** four elements be present in the same match. 

**Environment: **
- OS: Windows Server 2019
- TruffleHog version: 3.90.3

**Config snippet (simplified):**
```yaml
detectors:
- name: custom regex detector
  keywords:
  - Password
  regex:
    Password: '([A-Za-z0-9!@#$%^&*()_+=\-]{12,})'

### PR DESCRIPTION
**Pull request by kashifkhan0771**
### Description:
This PR adds support for validation rules in custom detectors to reduce false positives.

**Problem:**
Go's RE2 regex engine lacks support for lookahead assertions (e.g., (?=.*[0-9])), making it impossible to validate that detected secrets contain specific character types like digits, uppercase letters, or special characters. This limitation often leads to false positives when detecting API keys, passwords, and tokens that should follow specific rules.

**Solution:**
Added a new validations field to the Custom detector configuration that supports the following validation types:
- contains_digit: Requires at least one numeric digit (0-9)
- contains_lowercase: Requires at least one lowercase letter (a-z)
- contains_uppercase: Requires at least one uppercase letter (A-Z)
- contains_special_char: Requires at least one special character

**Example Usage:**
```yaml
detectors:
  - name: API Key Detector
    keywords:
      - api_key
    regex:
      key: ([A-Za-z0-9]{32,})
    validations:
      key:
        contains_digit: true
        contains_uppercase: true
        contains_special_char: true
```

### Checklist:


pkg/custom_detectors/validation.go
```diff
@@ -107,3 +108,42 @@ func ValidateRegexVars(regex map[string]string, body ...string) error {
...
+
+// ContainsDigit checks if string contains at least one digit
+func ContainsDigit(s string) bool {
+	for _, r := range s {
+		if unicode.IsDigit(r) {
+			return true
+		}
+	}
+
+	return false
```
nabeelalam: Since we're going to be checking for ASCII characters `'0'-'9'`, the following is a faster approach since `unicode.IsDigit()` handles more than just ASCII numbers:
```
for i := 0; i < len(s); i++ {
    c := s[i]
    if c >= '0' && c <= '9' {
        return true
    }
}

return false
```

pkg/custom_detectors/validation.go
```diff
@@ -107,3 +108,42 @@ func ValidateRegexVars(regex map[string]string, body ...string) error {
...
+}
+
+// ContainsLowercase checks if string contains at least one lowercase letter
+func ContainsLowercase(s string) bool {
+	for _, r := range s {
+		if unicode.IsLower(r) {
+			return true
+		}
+	}
+	return false
```
nabeelalam: Like the above comment, we can make a check for only ASCII characters here:
```
for i := 0; i < len(s); i++ {
    c := s[i]
    if c >= 'a' && c <= 'z' {
        return true
    }
}

return false
```

shahzadhaider1 (approved):

nabeelalam (approved):
LGTM!

amanfcp (approved):
LGTM

PR #4416, opened 2025-08-22
Fix errcheck / gopls nits in main.go

bradlarsen:
### Description:
This PR fixes 3 nits identified by `errcheck` and the `gopls` LSP server:

- Adds missing error checking around a fallible `os.SetEnv` call
- Switches a loop over a call to `strings.Split` to call the more efficient `strings.SplitSeq` instead
- Switches an older-style counting loop to use the newer range-over-int added in Go 1.22 in Feb 2024

These are pedantic nits. The changes should not noticeably affect runtime behavior, but will quiet down editor tools.

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -416,7 +416,9 @@ func run(state overseer.State) {
 	if *githubScanToken != "" {
 		// NOTE: this kludge is here to do an authenticated shallow commit
 		// TODO: refactor to better pass credentials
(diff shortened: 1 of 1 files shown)


mcastorina (approved):
Nice! TIL `range int` and `SplitSeq`

PR #4426, opened 2025-08-26
Scan Github Private Repositories With Token

### LINKED ISSUE #3349: GitHub scan does not include private repos for the user specified in --org, even when authenticated

### TruffleHog Version
trufflehog 3.82.6

### Trace Output
Can be provided if necessary

### Expected Behavior

I would have expected `trufflehog github --org cameronlonsdale` to scan all my public and private repositories / gists IF I provide a GItHub API token with the appropriate scopes. However it seems like only public repositories were scanned.

The API token I'm using has the following scopes `gist, read:org, read:user, repo`

### Actual Behavior

Only public repositories were scanned (and no private repositories)

### Steps to Reproduce

Create an appropriately scoped GitHub Token and then scan your own user with `trufflehog github --org <my-username>` and evaluate the output to see how many repositories were scanned, e.g. `  Completed enumeration   {"source_manager_worker_id": "qtPLB", "num_repos": 32, "num_orgs": 1, "num_members": 0}`. Compare this with the number of repositories that belong to your user (ignoring forks)

## Environment
 * OS: MacOS
 * Version 14.6

kashifkhan0771:
### Description:
Fixes: #3349
The `/pkg/sources/github/repo.go` file was hard to follow, so I’ve added comments to clarify each step.

### Checklist:


rosecodym (commented):
The code to implement `getReposByAuthenticatedUser` almost entirely duplicates the code to implement `getReposByUser`, which I'm pretty wary of, because it's going to be very easy for a future maintainer to change one of them but miss the other. (Arguably, you've already done that in this PR - you didn't change `enumerateBasicAuth` to use the new function, although I acknowledge that `enumerateBasicAuth` is currently dead.)

Did you consider adding a boolean argument to `getReposByUser` to indicate whether the request should be authenticated? You'd have to thread it through some parts of the call stack, but I suspect it will read more cleanly and be more robust to future maintenance.

kashifkhan0771:
> The code to implement `getReposByAuthenticatedUser` almost entirely duplicates the code to implement `getReposByUser`, which I'm pretty wary of, because it's going to be very easy for a future maintainer to change one of them but miss the other. (Arguably, you've already done that in this PR - you didn't change `enumerateBasicAuth` to use the new function, although I acknowledge that `enumerateBasicAuth` is currently dead.)
> 
> Did you consider adding a boolean argument to `getReposByUser` to indicate whether the request should be authenticated? You'd have to thread it through some parts of the call stack, but I suspect it will read more cleanly and be more robust to future maintenance.

Good suggestion! I merged the logic into `getReposByUser` based on `authenticated` parameter. Also added it for Basic Auth too.

pkg/sources/github/github.go
```diff
@@ -620,7 +620,7 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
 				if err := s.addUserGistsToCache(ctx, member, reporter); err != nil {
 					logger.Error(err, "error fetching gists by user")
 				}
-				if err := s.getReposByUser(ctx, member, reporter); err != nil {
+				if err := s.getReposByUser(ctx, member, false, reporter); err != nil {
```
rosecodym, line 625: I would find a comment explaining why app enumeration _doesn't_ expect to use an authenticated user helpful here. (I think I understand - because the credentials aren't associated with a particular user - but I don't think it's obvious.)

rosecodym (approved):
Thanks, I find this version much more readable! I find myself vexed by the need to create another repo list options structure and wrapper - all of the existing ones seem pretty wordy for the little they do - but I guess that's the way things are.

I did leave a note about a good spot for a comment, but this looks pretty good!

pkg/sources/github/github.go
kashifkhan0771: I dug a bit deeper into the docs about app tokens, and it turns out the list repo of authenticated user does work with application tokens. Should we enable it here as well?

kashifkhan0771: I added a TODO comment for it

PR #4427, opened 2025-08-26
Fix error propagation and a typo in verification logic

bradlarsen:
### Description:
When developing a new detector, I found that a typo in verification logic error handling had been repeated in a few dozen detectors. This PR fixes those typos, and also fixes a logic error found in the verification logic in the Box OAuth detector.

In particular, several detectors would silently ignore an error from a call to `http.NewRequestWithContext` when performing verification. Such an error is unlikely (a malformed URL, for example), but if it did occur, the finding would be incorrectly classified as `unverified`, when in fact it is `unknown`.

Additionally, the Box OAuth detector verification had the status flipped.

### Checklist:

### DIFF
--- a/pkg/detectors/boxoauth/boxoauth.go
+++ b/pkg/detectors/boxoauth/boxoauth.go
@@ -109,10 +109,13 @@ func verifyMatch(ctx context.Context, client *http.Client, id string, secret str
 		_ = res.Body.Close()
 	}()
 
-	// We are using malformed request to check if the client id and secret are correct
-	// box oauth api returns 400 status code even if they are correct
-	// so we need to check the response body for the error message with 'unauthorized_client' keyword
-	// or if invalid_client keyword is present then the client id & secret are incorrect
+	// We are using malformed request to check if the client id and secret are valid.
+	// In this case, the Box OAuth API returns a 400 status code even if the credentials are valid.
+	//
+	// - If the client ID/secret are valid, the response contains "unauthorized_client"
+	// - If the credentials are invalid, the response contains "invalid_client"
+	//
+	// So we check the response body for one of these keywords.
 	switch res.StatusCode {
 	case http.StatusBadRequest:
 		{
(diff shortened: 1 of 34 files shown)


amanfcp (approved):

pkg/detectors/boxoauth/boxoauth.go
```diff
@@ -120,13 +120,8 @@ func verifyMatch(ctx context.Context, client *http.Client, id string, secret str
...
 			}
-			body := string(bodyBytes)
-
-			if strings.Contains(body, "unauthorized_client") {
-				return true, nil, nil
-			}
-
-			return false, nil, nil
+			verified := !strings.Contains(string(bodyBytes), "unauthorized_client")
+			return verified, nil, nil
```
shahzadhaider1: Have we tested this API behavior?

If no, I think this change flips the logic in the wrong direction.

Box’s OAuth API behaves like this (based on the comment added above):
- If the client ID/secret are valid, the response contains "unauthorized_client" (since we’re intentionally sending a malformed request).
- If the credentials are invalid, the response contains "invalid_client".

With the updated code, `unauthorized_client` would mark the credentials as invalid, and `invalid_client` would incorrectly mark them as valid.

shahzadhaider1 (commented):
Thanks! I’ve added a comment regarding one part of the validation logic.

pkg/detectors/boxoauth/boxoauth.go
bradlarsen, line 1: @shahzadhaider1 does this look correct now? Thanks in advance.

shahzadhaider1 (approved):
LGTM
Not sure what has caused the `test-community` to fail

bradlarsen:
> Not sure what has caused the `test-community` to fail

Indeed! I'm not sure what's going on there. I don't think any of my PR changes should affect that? I also cannot reproduce the failure locally.

PR #4429, opened 2025-08-26
Fix broken filter in Trufflehog Enterprise UI (Hosted Scanner for Github cannot filter on Repo)

jordanTunstill:
### Description:
The fix separates repository filtering patterns from direct repository URLs and applies filtering before processing, ensuring that only repositories matching the IncludeRepos glob patterns are scanned instead of scanning all repositories in an organization.

### Checklist:


jordanTunstill:
I tested this by manually editing the vendored code into an enterprise PR, and used that PR instance to confirm intended functionality.

jordanTunstill:
continuing to test, did not filter out repos when checked this morning.

PR #4430, opened 2025-08-28
Changes to fix Enterprise UI filtering of Github Hosted Scanner Repositories to Include

jordanTunstill:
### Description:
The Enterprise UI's Github Hosted Scanner does not currently respect the Repositories to Include field, no matter what you put there, all repos in the org get scanned. These changes, along with the changes to the enterprise files resolve this: https://github.com/trufflesecurity/thog/pull/4996 

This was tested with these changes edited manually to the vendored in files: https://github.com/trufflesecurity/thog/pull/4987

After review with Jeff Ober and further investigation, it would be beneficial to have all filtering done upstream. I added in support for this in addition to the final filtering check I already had in place. A later refactor from the Scanning team could consider moving filtering to the giturl file instead of being unique to each individual source. SCAN-514 is open to track this request.

Edited github tests to fix error from PR 4426, as an endpoint was changed without the test endpoint being updated to match.

### Checklist:


jordanTunstill:
Enterprise PR here also needs review: https://github.com/trufflesecurity/thog/pull/4996

pkg/sources/github/github.go
```diff
@@ -432,14 +432,29 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	// Double make sure that all enumerated repositories in the
 	// filteredRepoCache have an entry in the repoInfoCache.
 	for _, repo := range s.filteredRepoCache.Values() {
-		ctx := context.WithValue(ctx, "repo", repo)
+		// Extract the repository name from the URL for filtering
+		repoName := repo
+		if strings.Contains(repo, "/") {
+			// Try to extract org/repo name from URL
+			if strings.Contains(repo, "github.com") {
```
sysread, line 444: This is my first time really digging into this repo, so take this with a grain of salt, but it seems like perhaps this logic should live in `pkg/giturl/giturl.go`. In fact, we might already have something there that can extract the org and repo from the path.

pkg/sources/github/repo.go
```diff
@@ -289,10 +289,15 @@ func (s *Source) wikiIsReachable(ctx context.Context, repoURL string) bool {
...
 
 func (s *Source) normalizeRepo(repo string) (string, error) {
-	// If there's a '/', assume it's a URL and try to normalize it.
-	if strings.ContainsRune(repo, '/') {
+	// If it's a full URL, normalize it
+	if strings.HasPrefix(repo, "http") {
 		return giturl.NormalizeGithubRepo(repo)
 	}
+	// If it's a repository name (contains / but not http), convert to full URL first
+	if strings.Contains(repo, "/") && !strings.HasPrefix(repo, "http") {
```
sysread: I would probably test for something like `^[a-z]+:\/\/` here myself.

pkg/sources/github/github.go
```diff
@@ -432,14 +432,29 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
...
+				}
+			}
+		}
 
-		repo, err := s.ensureRepoInfoCache(ctx, repo, &unitErrorReporter{reporter})
-		if err != nil {
-			ctx.Logger().Error(err, "error caching repo info")
-			_ = dedupeReporter.UnitErr(ctx, fmt.Errorf("error caching repo info: %w", err))
+		// Final filter check - only include repositories that pass the filter
+		if s.filteredRepoCache.includeRepo(repoName) && !s.filteredRepoCache.ignoreRepo(repoName) {
```
sysread: It seems like this condition ought to already be a method of `filteredRepoCache` so you could just do:
```
        if s.filteredRepoCache.wantRepo(repoName) {
            ...
        }
```

sysread (changes requested):
This is a good fix. I think you could improve this by moving the logic for extracting the repo name from the URL to a new method the giturl package (if we don't already have a method that does it there).

I definitely think it would help to add a short unit test that verified that the filtering works correctly in `Enumerate`.

Optional, but I think it would be cleaner to wrap that composite condition in a new method on `filteredRepoCache`, which would make it testable without having to test the outer function, too.

pkg/sources/github/github.go
jordanTunstill: Thanks @sysread ! I looked into this, and the other source types have their own filtering processes. I'll make a note / scanning ticket to consider this as part of a bigger change or refactor, as that could clean up code from multiple sources.

pkg/sources/github/github.go
jordanTunstill: This makes sense! I've added some upstream filtering and will commit it soon.

pkg/sources/github/repo.go
jordanTunstill: Good call, just made this change!

pkg/sources/github/github.go
sysread: Was this response intended for one of my other comments? This was just about extracting the URL parsing and moving it into the `giturl` package, which is for code that works with, you know, _git urls_...

sysread (changes requested):
Excellent test and the regex looks pretty alright to me. Left one comment to clarify.

pkg/sources/github/github.go
jordanTunstill: That makes sense. From how I was reading through the different OSS sources, it looks like they all handle url parsing and filtering separately. It would probably be good for a refactor to be done that consolidated all of this work, but it seemed outside the scope of the bugfix.

jordanTunstill:
@sysread weirdly, after merging changes from main in, it's now failing tests that it previously succeeded. My present theory is that these test failures are unrelated, but will likely need to be addressed for this/other prs.

sysread (approved):
Good enough! Remember to let the integrations team know about what made the fix necessary. It might indicate a CI issue they should be aware of.

pkg/sources/github/github.go
```diff
@@ -432,14 +437,29 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
...
+			}
+		}
 
-		repo, err := s.ensureRepoInfoCache(ctx, repo, &unitErrorReporter{reporter})
-		if err != nil {
-			ctx.Logger().Error(err, "error caching repo info")
-			_ = dedupeReporter.UnitErr(ctx, fmt.Errorf("error caching repo info: %w", err))
+		// Final filter check - only include repositories that pass the filter
+		if s.filteredRepoCache.wantRepo(repoName) {
+			ctx = context.WithValue(ctx, "repo", repo)
```
rosecodym, line 454: `WithValue` doesn't overwrite existing keys - it adds a new key-value pair with the same key. Since this is done in a loop, `ctx` accumulates a new `repo` key for every loop iteration. This should have been done using `:=` instead of `=` so that the previous context was dropped on every loop iteration. This may be causing memory troubles when the repository count is very large.

jordanTunstill: @rosecodym I'll get a new PR up with that fixed right away.

rosecodym: i opened #4549

PR #4431, opened 2025-08-29
Implemented Graphql requests for Github PR's, Issues and comments scanning

kashifkhan0771:
### Description:
This Pull request replaces the HTTP API requests for Github source PR's, Issues and their comments with GraphQL requests.

### Scan Results with no verification:
Note: While results may vary slightly between runs, GraphQL consistently outperforms REST in scan duration.

Screenshot from 2025-09-03 13-40-13

### Checklist:


pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,489 @@
...
+
+// handleGraphQLRateLimit inspects the rateLimit info returned in GraphQL queries.
+func (s *Source) handleGraphQLRateLimit(ctx context.Context, rl *rateLimit, errIn error, reporters ...errorReporter) bool {
+	// if rate limit exceeded error happened, wait for 5 minute before trying again
+	if errIn != nil && strings.Contains(errIn.Error(), "rate limit exceeded") {
+		retryAfter := 5 * time.Minute
+		ctx.Logger().Info("GraphQL RATE_LIMITED error (fallback)",
+			"retry_after", retryAfter.String())
+		time.Sleep(retryAfter)
+		return true
```
rosecodym: This early return (and all of the later ones) happen before we report rate limiting to Prometheus, which means that we won't have much visibility when it happens. Relatedly, I find it a bit hard to follow this new rate limiting code - I suspect the difficulty was with integrating it with the existing rate limiting code. (For example: Why does this initial request have a 3x 5minute retry step before getting into the global rate limiting code?) You might be able to resolve both issues at once with another refactoring pass at this function.

pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,489 @@
...
+// handleRateLimitWithChunkReporter is a wrapper around handleRateLimit that includes chunk reporting
+func (s *Source) handleGraphqlRateLimitWithChunkReporter(ctx context.Context, reporter sources.ChunkReporter, rl *rateLimit, errIn error) bool {
+	return s.handleGraphQLRateLimit(ctx, rl, errIn, &chunkErrorReporter{reporter: reporter})
+}
+
+// handleGraphQLRateLimit inspects the rateLimit info returned in GraphQL queries.
+func (s *Source) handleGraphQLRateLimit(ctx context.Context, rl *rateLimit, errIn error, reporters ...errorReporter) bool {
+	// if rate limit exceeded error happened, wait for 5 minute before trying again
+	if errIn != nil && strings.Contains(errIn.Error(), "rate limit exceeded") {
+		retryAfter := 5 * time.Minute
```
rosecodym: This is a _huge_ retry interval. Does GitHub prescribe it?

rosecodym (commented):
I was initially skeptical of putting the new code in graphql.go, but I think I see why you did it and I can't think of a better option.

However, I _did_ have some trouble following the rate limiting code - in particular, how it interacts with the rest of the GitHub rate limiting code. Specifically, it's doing a lot of its own stuff before it begins to use the rest of the rate limiting code, and I don't understand why it needs to do that beforehand. (Also, mentioned inline, 5 minutes is a huge interval, and Prometheus gets skipped in many cases.)

Can you try another pass over the rate limiting code to see if things can be clarified a bit?

pkg/sources/github/graphql.go
kashifkhan0771: I overlooked the reporting to Prometheus. Thanks for pointing out!

The reason I initially set the request interval to 5 minutes was because the GraphQL API sometimes returns a rate limit error before the `remaining` value reaches `0`. When this happens, the response includes **no rate limit info**, so we can't rely on the usual rate limit object and the `resetAt` value defaults to zero, which can cause incorrect handling.

To deal with this, I first handled the error case directly, then move on to process the rate limit if available.

Your comment gave me a new perspective, so I adjusted the approach a bit and now:

1. We first check if the **global rate limit** is already set.

   * If it's set, we use that value and delay accordingly.
2. If it's **not set**, and we receive an error:

   * We check if it's a rate limit issue.
   * In that case, we set the global rate limit (so other requests know to wait) and continue reporting to Prometheus.
3. If there's **no error**, we check the actual rate limit object:

   * If `remaining < 3`, we wait until the `resetAt` time.
   * Why 3? Because some GraphQL queries can cost more than one point, depending on complexity.

pkg/sources/github/graphql.go
kashifkhan0771: Answered in the above comment.

main.go
```diff
@@ -465,6 +465,9 @@ func run(state overseer.State) {
 	feature.UseSimplifiedGitlabEnumeration.Store(true)
 	feature.GitlabProjectsPerPage.Store(100)
 
+	// OSS Default using github graphql api for issues, pr's and comments
+	feature.UseGithubGraphqlAPI.Store(true)
```
camgunz: Let's start this off as `false` so it's not automatically turned on for EE customers (we can/should make it `true` for OSS after we get the flag in EE)

pkg/sources/github/github.go
```diff
@@ -1264,7 +1271,7 @@ var (
 	state = "all"
 )
 
-func (s *Source) processRepoComments(ctx context.Context, repoInfo repoInfo, reporter sources.ChunkReporter, cutoffTime *time.Time) error {
+func (s *Source) processIssueandPRsWithCommentsREST(ctx context.Context, repoInfo repoInfo, reporter sources.ChunkReporter, cutoffTime *time.Time) error {
```
camgunz: I'm an 80 columns person because I have a bunch of buffers open side-by-side, but I accept I'm a relic from an earlier time. This one (and at least one more below) are > 150 though--can we wrap somewhere between 90 and 120?

main.go
camgunz: Also, regretfully I think the initialism is `GraphQL`--if we're gonna do it for `REST` we should probably do it here too

pkg/sources/github/github.go
```diff
@@ -1096,8 +1097,14 @@ func (s *Source) scanComments(ctx context.Context, repoPath string, repoInfo rep
 	if s.includeGistComments && isGistUrl(urlParts) && !s.ignoreGists {
 		return s.processGistComments(ctx, urlString, urlParts, repoInfo, reporter, cutoffTime)
 	} else if s.includeIssueComments || s.includePRComments {
-		return s.processRepoComments(ctx, repoInfo, reporter, cutoffTime)
+		// if we need to use graphql api for repo issues, prs and comments
+		if feature.UseGithubGraphqlAPI.Load() {
+			return s.processRepoIssueandPRsWithCommentsGraphql(ctx, repoInfo, reporter, cutoffTime)
```
camgunz, line 1111: (No need to do anything here, just musing)

Hrm, w/ the old code passing `cutoffTime` just the once was kind of 🤷🏻 , but now that we're carrying it around everywhere, it makes me think it'd be nice if we processed `s.commentsTimeframeDays` up top in `Init`, that way we don't need to drill it down everywhere.

pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,485 @@
...
+		for _, pr := range query.GetPullRequests() {
+			ctx.Logger().V(5).Info("Scanning pull request comments",
+				"pull_request_no", pr.Number,
+				"total_comments", len(pr.Comments.Nodes))
+
+			if err := s.chunkComments(ctx, repoInfo, pr.GetPRComments(), reporter, cutoffTime); err != nil {
+				return err
+			}
+
+			// if a pull request has more than 100 comments - some interns pull request might endup here :)
```
camgunz, line 157: 😂

camgunz (changes requested):
Broadly looks good, just had a couple notes. We should add tests for this--_probably_ the integration tests should cover this (maybe a unit test or two for `chunkIDs`), so if we just run the integration tests w/ `UseGithubGraphqlAPI` both `false` and `true`, that'd give us even more confidence we're not breaking anything.

main.go
kashifkhan0771: Sure, I'll temporarily turn it off. I didn't quite understand your second comment - just to clarify, we don’t use any initialism for REST.

pkg/sources/github/github.go
kashifkhan0771: Sound's good we can do it in a separate optimization PR.

kashifkhan0771:
> We should add tests for this--_probably_ the integration tests should cover this (maybe a unit test or two for `chunkIDs`), so if we just run the integration tests w/ `UseGithubGraphqlAPI` both `false` and `true`, that'd give us even more confidence we're not breaking anything.

I added the test but even the existing integration tests are not working for me 😿

main.go
camgunz: Oh I was just saying the "QL" is capitalized; `Graphql` -> `GraphQL`

kashifkhan0771: Done

GoBlaq1604 (approved):

pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,491 @@
...
+					repository:         githubv4.String(repoInfo.name),
+					issueNumber:        githubv4.Int(issue.Number),
+					commentsPerPage:    githubv4.Int(defaultPagination),
+					commentsPagination: issue.Comments.PageInfo.EndCursor,
+				}
+
+				// request this issue more comments
+				var commentsQuery singleIssueComments
+				err := s.connector.GraphQLClient().Query(ctx, &commentsQuery, commentVars)
+				if s.handleGraphqlRateLimitWithChunkReporter(ctx, reporter, &query.RateLimit, err) {
```
martinlocklear:
```suggestion
				if s.handleGraphqlRateLimitWithChunkReporter(ctx, reporter, &commentsQuery.RateLimit, err) {
```

Not sure if this is what you want or not.

pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,491 @@
...
+				singlePRVars := map[string]any{
+					owner:              githubv4.String(repoInfo.owner),
+					repository:         githubv4.String(repoInfo.name),
+					pullRequestNumber:  pr.Number,
+					commentsPerPage:    githubv4.Int(defaultPagination),
+					commentsPagination: pr.Comments.PageInfo.EndCursor,
+				}
+
+				err := s.connector.GraphQLClient().Query(ctx, &commentQuery, singlePRVars)
+				if s.handleGraphqlRateLimitWithChunkReporter(ctx, reporter, &query.RateLimit, err) {
```
martinlocklear:
```suggestion
				if s.handleGraphqlRateLimitWithChunkReporter(ctx, reporter, &commentQuery.RateLimit, err) {
```

Similar to another comment I made.   It's not obvious to me from looking here which rate limit we should be respecting.  Some inline comments might help.

If we _aren't_ interested in respecting separate rate limits for each query object separately, I don't understand why the rate limit object is attached to the query object.  (maybe out of scope for this?).

pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,491 @@
...
+		for thread.Comments.PageInfo.HasNextPage {
+			// request this thread more comments
+			var query singleReviewThreadComments
+			reviewThreadVars := map[string]any{
+				threadID:           thread.ID,
+				commentsPerPage:    githubv4.Int(defaultPagination),
+				commentsPagination: (*githubv4.String)(nil),
+			}
+
+			if err := s.connector.GraphQLClient().Query(ctx, &query, reviewThreadVars); err != nil {
```
martinlocklear: Does this need to be rate limited as well?

pkg/sources/github/graphql.go
```diff
@@ -0,0 +1,491 @@
...
+
+	// if resume time is not empty and is after current time, than put the request to sleep till that.
+	if !resumeTime.IsZero() && time.Now().Before(resumeTime) {
+		retryAfter := time.Until(resumeTime)
+		time.Sleep(retryAfter)
+		return true
+	}
+
+	var retryAfter time.Duration
+	// if rate limit exceeded error happened, wait for 5 minute before trying again
```
martinlocklear: I think this comment is out of date now.

pkg/sources/github/github_integration_test.go
```diff
@@ -214,6 +215,106 @@ func TestSource_ScanComments(t *testing.T) {
...
+	}
+	tests := []struct {
+		name              string
+		init              init
+		wantChunk         *sources.Chunk
+		wantErr           bool
+		numExpectedChunks int
+	}{
+		{
+			name: "unauthenticated, single repo, issue comments and pr comments",
```
martinlocklear: Please avoid using tabular tests like this.   They end up being hard to understand and even harder to maintain.  Breaking these out into individual test functions (or even a new test suite) will make our lives easier in the future.  imo, it's just as easy to copypasta a test function as it is to copypasta an item in a list of tests, but it's harder to manage as code paths _diverge_ when they're not broken out.

martinlocklear (commented):
Left a couple of comments about rate limiting, and a strong opinion about tabular tests.  (I've been drawing folks attention to testify's [testsuite](https://pkg.go.dev/github.com/stretchr/testify/suite) functionality to serve similar purposes but be more maintainable).

In general I think PRs this size are too large to effectively review - nothing stands out to me as a problem though.

pkg/sources/github/graphql.go
kashifkhan0771: Good catch 👀

pkg/sources/github/graphql.go
kashifkhan0771: We want to make sure we're respecting the rate limits for single-query objects, like the comments on a single PR. I mistakenly used the wrong rate limit in this case thanks for catching that. While it's rare to hit the rate limit on these individual queries, it's still important to get it right.

pkg/sources/github/github_integration_test.go
kashifkhan0771: The `github_integration_test.go` file uses tabular test cases throughout. If we want to update the test case structure, we should apply the change consistently across all test cases. It might be better to create a separate ticket or PR for that cleanup.

kashifkhan0771: Done

camgunz (approved):
OK I think this is good. I'll ping you on Slack w/ next steps

pkg/sources/github/graphql.go
shahzadhaider1: I was that intern (once) - proud centurion

shahzadhaider1 (approved):

kashifkhan0771:
:face_holding_back_tears:

Issue #1019, opened 2023-01-13
Flag to suppress known test secrets in OSS version of Trufflehog


Kruthika0:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Provision to whitelist known test secret when using Trufflehog via GitHub Action, there are often times when developers are aware of few secrets but since they are used in test environment it is required to be used as is. Currently, there is no way to mark certain credentials such that the tool doesn’t raise them during detection. There should be a way to suppress credential findings in the open-source version of the tool. 

## Problem to be Addressed
There is no way to suppress any known credential at this moment in Trufflehog GHA due to which I am not able to use this scanner to its fullest capability.

## Description of the Preferred Solution
We may have a flag which could be turned ON with the type of secret and the value which needs to be suppressed via some kind of ignore file, this file could be passed in the GHA as an optional parameter.

dustin-decker:
There is a now a `trufflehog:ignore` feature which supports this for sources that support line numbers: https://github.com/trufflesecurity/trufflehog#question-faq

xtermi2:
> There is a now a `trufflehog:ignore` feature which supports this for sources that support line numbers: https://github.com/trufflesecurity/trufflehog#question-faq

Hi @dustin-decker 
This is better than nothing, but for our case not sufficient:
We want scan our deployment artefacts (`jar` files) because we had an issue some months ago where environment variables were leaking in build artefacts. `trufflehog:ignore` is not sufficient for this case because lots of external libraries are included which produce false positives. We need some ignore file / baseline file where we can ignore findings or mark them as false positives.

PR #4436, opened 2025-09-02
Add allowlist feature to suppress test secrets

jporzucek:
### Description:
Implements whitelisting https://github.com/trufflesecurity/trufflehog/issues/1019

### Checklist:


pkg/detectors/falsepositives.go
```diff
@@ -199,3 +202,86 @@ func FilterKnownFalsePositives(ctx context.Context, detector Detector, results [
...
+			continue
+		}
+
+		isWhitelisted := false
+		var matchReason string
+
+		// Check if the raw secret matches any whitelisted secret
+		rawSecret := string(result.Raw)
+		if isWhitelisted, matchReason = isSecretWhitelisted(rawSecret, whitelistedSecrets); isWhitelisted {
+			ctx.Logger().V(4).Info("Skipping result: whitelisted secret", "result", rawSecret, "reason", matchReason)
```
nabeelalam: Hey @jporzucek ! While the detected secret _is_ technically whitelisted, it could still potentially be a live one - not sure if we should be logging them, might unintentionally leak?

pkg/detectors/falsepositives.go
```diff
@@ -199,3 +202,86 @@ func FilterKnownFalsePositives(ctx context.Context, detector Detector, results [
...
+			whitelistedSecrets[secret] = struct{}{}
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("error reading whitelist file: %w", err)
+	}
+
+	return whitelistedSecrets, nil
+}
```
nabeelalam: This loader assumes one secret per line (`bufio.Scanner` splits on `'\n'`). That works fine for tokens and API keys, but it would break for multiline secrets (e.g. PEM private keys).
Off the top of my head there could be different ways to support that, perhaps a different format (JSON array, yaml, etc.).

nabeelalam (changes requested):
Hey @jpozucek, thanks for contributing. I'm going through your PR but I've left a couple comments

pkg/detectors/falsepositives.go
jarek-clari: Fair point, maybe let's log a masked secret?

jporzucek: I left the logging but raw secret is now masked, wdyt?

pkg/detectors/falsepositives.go
jporzucek: I went ahead with yaml, see [example](https://github.com/jporzucek/trufflehog/blob/add-whitelist-feature/examples/whitelisted.yml) file. Thanks for the feedback BTW! 🙇

rosecodym (commented):
Thanks for your work on this - the change looks pretty clean and surgical. I'd like to ask how much of a lift it would be to rename "whitelist" to "allowlist," simply because we're already using the latter term in the codebase, and we've historically found a minor but extant cognitive penalty when getting people up to speed on multiple concepts that are the same but have different names. (I'm not requesting changes for this in case it's more of a burden than I'm anticipating.)

jporzucek:
@rosecodym Changed to allowlist, np!

main.go
```diff
@@ -65,6 +65,7 @@ var (
 	allowVerificationOverlap   = cli.Flag("allow-verification-overlap", "Allow verification of similar credentials across detectors").Bool()
 	filterUnverified           = cli.Flag("filter-unverified", "Only output first unverified result per chunk per detector if there are more than one results.").Bool()
 	filterEntropy              = cli.Flag("filter-entropy", "Filter unverified results with Shannon entropy. Start with 3.0.").Float64()
+	allowlistSecretsFile       = cli.Flag("allowlist-secrets", "Path to file with newline separated list of secrets to allowlist.").String()
```
nabeelalam: Looks like this description became outdated with the update to yaml config

pkg/detectors/falsepositives.go
nabeelalam: Thanks @jporzucek !

pkg/detectors/falsepositives.go
nabeelalam: That works. Thanks!

nabeelalam:
@jporzucek I tried to test `allowlist-secrets` with a yaml with a few known verified and unverified secrets, but even though the yaml file is loaded, the result still includes all of the allow listed secrets

jporzucek:
@nabeelalam ahhh, rookie mistake! Please try again now

jporzucek:
Also added support for correct parsing with YAML files that don't end with a newline character. The function now automatically appends `\n` if missing, improving parser compatibility and following POSIX standards.

nabeelalam:
Thanks for the update @jporzucek I'll test it again in a little bit

nabeelalam:
The flag is working for me now, thanks @jporzucek. This looks good! If it's not a big change, is it possible to make the flag itself a bit more accurately descriptive? `--allowlist-secrets` doesn't inherently indicate that it accepts a path to a yaml file. Just off the top of my head `--secret-allowlist-path` seems a touch more accurate. Perhaps you can think of a better one?

jporzucek:
@nabeelalam Renamed to `--allowlist-secrets-file` - hope it's more accurate 🤞

pkg/detectors/falsepositives.go
```diff
@@ -199,3 +209,110 @@ func FilterKnownFalsePositives(ctx context.Context, detector Detector, results [
...
+// FilterAllowlistedSecrets filters out results that match allowlisted secrets.
+// This allows users to specify known safe secrets that should not be reported.
+// Supports regex patterns.
+func FilterAllowlistedSecrets(ctx context.Context, results []Result, allowlistedSecrets map[string]struct{}) []Result {
+	if len(allowlistedSecrets) == 0 {
+		return results
+	}
+
+	var filteredResults []Result
+	for _, result := range results {
```
camgunz: I think this might be a good candidate for [`slices.DeleteFunc`](https://pkg.go.dev/slices#DeleteFunc). It inverts the logic a little, but you don't need any loop management so in that sense it's cleaner.

pkg/detectors/falsepositives.go
```diff
@@ -199,3 +209,110 @@ func FilterKnownFalsePositives(ctx context.Context, detector Detector, results [
...
+// isSecretAllowlisted checks if a secret matches any allowlisted pattern (exact string or regex)
+func isSecretAllowlisted(secret string, allowlistedSecrets map[string]struct{}) (bool, string) {
+	// First, try exact string matching for performance
+	if _, isAllowlisted := allowlistedSecrets[secret]; isAllowlisted {
+		return true, "exact match"
+	}
+
+	// Try regex matching
+	for pattern := range allowlistedSecrets {
+		if regex, err := regexp.Compile(pattern); err == nil {
```
camgunz: I don't think many users would have lots of these, but it's probably still worth it to be defensive and only compile them once--there's an `init` func up above that could be used for this

pkg/detectors/falsepositives.go
```diff
@@ -199,3 +209,110 @@ func FilterKnownFalsePositives(ctx context.Context, detector Detector, results [
...
+			continue
+		}
+
+		isAllowlisted := false
+		var matchReason string
+
+		// Check if the raw secret matches any allowlisted secret
+		rawSecret := string(result.Raw)
+		if isAllowlisted, matchReason = isSecretAllowlisted(rawSecret, allowlistedSecrets); isAllowlisted {
+			ctx.Logger().V(4).Info("Skipping result: allowlisted secret", "result", maskSecret(rawSecret), "reason", matchReason)
```
camgunz: I appreciate this attention to detail re: `maskSecret`! I wanna use [our redacting function](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/log/dynamic_redactor.go#L26) for this though.

camgunz (changes requested):
Thanks for this! The changes I want are using `RedactGlobally` instead of `maskSecret` and compiling the regexes up top. It's possible using `slices.DeleteFunc` doesn't make sense, so if it doesn't, definitely don't rework things to make it make sense.

jporzucek:
@camgunz All requested changes applied! 🙌

zricethezav:
I don't want to throw a wrench in this review, but what about introducing this feature in the `--config` file? https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#configuration

jporzucek:
> I don't want to throw a wrench in this review, but what about introducing this feature in the `--config` file? https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#configuration

You mean as a new field in [sources](https://github.com/trufflesecurity/trufflehog/blob/466da5b0bb161144f6afca9afe5d57975828c410/pkg/pb/sourcespb/sources.pb.go#L292-L311)?

zricethezav:
> You mean as a new field in [sources](https://github.com/trufflesecurity/trufflehog/blob/466da5b0bb161144f6afca9afe5d57975828c410/pkg/pb/sourcespb/sources.pb.go#L292-L311)?

I was thinking a new top-level key `allowlist` in the config that could be globally defined like "ignore these secrets if found in any detector" and detector specific allowlists like "ignore these secrets if found in this detector".

This behavior is similar to what is offered in the gitleaks configuration https://github.com/gitleaks/gitleaks?tab=readme-ov-file#configuration. 

FWIW, I may be muddying the waters here but I figured since `--config` supports custom detectors and sources it could also include detector behaviors like allowlists filters.

jporzucek:
@zricethezav I've added a top-level `allowlists` so it matches `gitleaks` naming convention. The behavior now is it merges allowlisted secrets from `--config` and `--allowlist-secrets-file` files. I'm not sure what's the expected behavior so let me know if one should have precedence over the other.

jporzucek:
@zricethezav @nabeelalam @camgunz Can you take a look when you got a chance? 🙏

camgunz (approved):
Thanks so much for hanging in there!

nabeelalam (approved):
Looks good!

jporzucek:
@nabeelalam @camgunz Can we merge this please? 😁

vegeta2op:
[Query]
Hi , 
Just curious, when is this planned to be Merged ?? 
Thanks.

steeve85:
@camgunz @nabeelalam Any chances to get this PR merged? Thx

posquit0:
@camgunz @nabeelalam @rosecodym 
I'm also waiting for this feature! Thx!

PR #4441, opened 2025-09-04
Add generic JWT detection and verification

bradlarsen:
### Description:
This adds a generic detector and verifier for generic JWTs.

Detection is simple using regular expressions. However, this will produce many false positives. JWT verification is performed using the `github.com/golang-jwt/jwt/v5` package, performing the usual checks (proper decoding, checking timestamp-related claims, checking for an allow-list of supported algorithms, etc). Only public key cryptography algorithms are supported. Additionally, OIDC Discovery is attempted against the issuer to fetch the public key for signature verification.

### Testing:
A few unit tests demonstrate the detector working. Further testing of verification is needed.

### TODO:
- [x] thoroughly test verification
- [ ] add integration tests
- [x] think about and write up the security implications of using OIDC Discovery with an attacker-controlled URL
- [x] estimate the impact on increased finding volume when verification is disabled


bradlarsen:
Note: the custom detector test failed in the `test-community` job above:
```
--- FAIL: TestDetectorValidations (0.00s)
    --- FAIL: TestDetectorValidations/custom_validation_-_multiple_regex_validations (0.00s)
        custom_detectors_test.go:556: CustomDetector.FromData() custom validation - multiple regex validations diff: (-got +want)
              []detectors.Result{
              	{
              		... // 2 identical fields
              		Verified:              false,
              		VerificationFromCache: false,
              		Raw: bytes.Join({
            + 			"MyStrongP@ssword",
              			"c392c9837d69b44c764cbf260b-e6184",
            - 			"MyStrongP@ssword",
              		}, ""),
              		RawV2:    nil,
              		Redacted: "",
              		... // 3 ignored and 2 identical fields
              	},
              }
FAIL
FAIL	github.com/trufflesecurity/trufflehog/v3/pkg/custom_detectors	0.043s
```
I have seen this sporadic test failure a few times now. I'm pretty sure it's caused by iterating over a map (the order of which is unspecified in Go) in the custom detector code:

https://github.com/bradlarsen/trufflehog/blob/1e8671c6de6d721636e672b3066c6773e4ab6d0b/pkg/custom_detectors/custom_detectors.go#L118

pkg/detectors/jwt/jwt_test.go
```diff
@@ -0,0 +1,83 @@
+package jwt
```
joeleonjr, line 1: Could you consider adding unittests to address the following situations:
1. Incorrect algorithm
2. jwt.ErrTokenMalformed
3. jwt.ErrTokenExpired
4. jwt.ErrTokenNotValidYet
5. The various error paths (invalid issuer, non-https issuer, invalid kid, no jwks endpoint, etc) in the oidcDiscoveryKeyFunc() function. I think you could use the `net/http/httptest` package here.

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Attempt to verify a JWT.
+//
+// This implementation only attempts to verify JWTs that use an asymmetric encryption algorithm,
+// and only those whose issuers use the OIDC Discovery protocol to make public keys available via request.
+func verifyMatch(ctx context.Context, client *http.Client, tokenString string) (bool, map[string]string, error) {
+
+	// A key retrieval function that uses the OIDC Discovery protocol
+	oidcDiscoveryKeyFunc := func(unverifiedToken *jwt.Token) (any, error) {
```
joeleonjr: Can we address embedded public keys inside this function?

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,215 @@
...
+		// Fetch the JWKS
+		req, err = http.NewRequestWithContext(ctx, "GET", discoveryDoc.JWKSUri, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create JWKS request: %w", err)
+		}
+		resp, err = client.Do(req)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch JWKS: %w", err)
+		}
+		defer resp.Body.Close()
```
joeleonjr: Consider limiting the response body size so that we don't inadvertently DOS.

joeleonjr: Haven't used it before but maybe `io.LimitReader()`

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,215 @@
...
+		if resp.StatusCode != 200 {
+			return nil, fmt.Errorf("JWKS returned status: %v", resp.Status)
+		}
+
+		// Load JWKS into `jwt.VerificationKeySet` or `jwt.VerificationKey` from JSON
```
joeleonjr: Do you think there's any need to make the jwks json parsing more robust? Ex: the `Key` type maps strings to `any` type. Do you think there's an opportunity to get more precise as to what values we expect, or is that not possible due to the multitide of key types? Also, are the `keys` in the `jwks` struct always going to be at that level in the json object or could they ever be nested? 

I'm hesitant to suggest including a different library, but I found [this one that deals with jwks](https://github.com/lestrrat-go/jwx/tree/168236bad4c6234d2fafb9951cd46d1cbb6bb8e8/jwk)...might be worth a look?

bradlarsen:
> I have seen this sporadic test failure a few times now. I'm pretty sure it's caused by iterating over a map (the order of which is unspecified in Go) in the custom detector code

Fixed and merged in #4446.

pkg/detectors/jwt/jwt.go
bradlarsen: Good questions here.

I was reading through the library code a bunch last week and it looks like the `jwk` library's type for keys is in fact a type union that includes `any`; there is corresponding code within that library for validating signatures that ensure those `any` values have the expected structure.

I have been going back and forth on avoiding added dependencies and writing code by hand vs. using that jwks library, which is probably much more robust and supports more cases.

I will keep an eye on these things as I do more testing, but my gut feeling is that I'm going to ultimately end up using an existing library here.

bradlarsen: On the multitude of key types... there are some examples in the JWKS spec ([RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517#appendix-A.1)). The fields for each key vary quite a bit depending on what type of cryptography is used.

And indeed, my understanding is that the `"keys"` field will always appear at the top-level like that, as per [RFC 7517 Section 5](https://datatracker.ietf.org/doc/html/rfc7517#section-5).

pkg/detectors/jwt/jwt.go
bradlarsen: Good suggestions; I've added this with a 1MB limit.

Note also that the http client used has timeout defaults of 5s per request also.

pkg/detectors/jwt/jwt.go
bradlarsen: I switched to the https://github.com/lestrrat-go/jwx library for parsing JWK sets to the expected "native" crypto types needed for JWT signature verification.

pkg/detectors/jwt/jwt.go
bradlarsen: I've switched to using the https://github.com/lestrrat-go/jwx library for parsing JWK sets obtained via OIDC Discovery.

What does a JWT look like that has the public key embedded?

joeleonjr: https://pragmaticwebsecurity.com/articles/apisecurity/hard-parts-of-jwt.html#:~:text=Embedding%20keys,verify%20the%20authenticity.

bradlarsen:
A TODO item from the description:

> estimate the impact on increased finding volume when verification is disabled

I ran with this new JWT detection over 250k recent commits from GitHub. From there, about 4k distinct JWTs were found, and several dozen of them verified completely.

The newly detected JWTs will be notable, but less volume than, say, `Github` V1.

pkg/detectors/jwt/jwt_test.go
bradlarsen: I added a number of additional unit tests. Most of these test cases will in fact need to be integration tests (not yet written) since they are only detected with verification enabled.

pkg/detectors/jwt/jwt.go
bradlarsen: This is a good idea! I'm going to leave this work as a future extension though.

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,312 @@
...
+		"ewogIC",
+		"ewoid",
+		"eyJ",
+	}
+}
+
+// FromData will find and optionally verify JWT secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	client := cmp.Or(s.client, common.SaneHttpClient())
+	seenMatches := make(map[string]bool)
```
kashifkhan0771: `map[string]struct{}` is more memory efficient.

bradlarsen: I didn't know this! Thinking about it a bit, it makes sense. I'll switch it.

Thanks for the tip.

joeleonjr:
@bradlarsen given clarification on the `unknown` status in #4477, I think usage of that status in the JWT detector should be removed. I think it's worth just commenting out those code paths (instead of fully deleting), since they would be useful in the future if we introduce any additional statuses. Your call. And if you have any other ideas here, I'm sure the team would be open to discussing.

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,316 @@
...
+	)
+	switch {
+	case token.Valid:
+		return true, nil
+	case errors.Is(err, jwt.ErrTokenUnverifiable):
+		return false, err
+	case errors.Is(err, jwt.ErrHashUnavailable):
+		return false, err
+	default:
+		return false, nil
```
shahzadhaider1, line 300: I hope you have thought about this, because returning nil hides the error cause and will mark the secret "unverified" without a reason.

shahzadhaider1 (dismissed):
Good work. 
Looks good, left one minor comment.

pkg/detectors/jwt/jwt.go
bradlarsen: Yes, that's intentional: the remaining error cases that the `jwt.Parse` function can return (and there are many of them) indicate that the JWT is invalid for some reason, like the timestamp-related claims don't check out.

The code here is not as clear as it might be. I'll at least add a comment. Thanks for pointing this out.

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to perform request: %w", err)
+	}
+	return resp, nil
+}
+
+// Attempt to verify a JWT that uses an HMAC algorithm.
+//
+// This implementation only attempts to verify JWTs whose issuers use the OIDC Discovery protocol to make public keys available via request.
+func verifyHMAC(parsedToken *jwt.Token) (bool, error) {
```
rosecodym: I'm generally opposed to functions that signify they can return things they can't actually return (an `error` in this case), because I think it usually adds unnecessary cognitive overhead during future maintenance. (I'm not requesting a change - I'm just flagging this because it looks like the elimination of any actually returned error might have happened in the course of writing this function rather than as your initial plan.)

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,357 @@
...
+
+// Wrap an `io.Reader` with a reasonable limit as an additional measure against DoS from a malicious JWKS issuer
+func limitReader(reader io.Reader) io.Reader {
+	return io.LimitReader(reader, 1024*1024)
+}
+
+// Attempt to verify a JWT that uses a public-key signing algorithm.
+//
+// This implementation only attempts to verify JWTs whose issuers use the OIDC Discovery protocol to make public keys available via request.
+func verifyPublicKey(ctx context.Context, client *http.Client, tokenString string) (bool, error) {
```
rosecodym: Every single returned `error` here is going to cause an indeterminate result. Is that correct? The "verification error" state is designed to indicate that we couldn't verify a candidate finding because of uncontrollable external conditions, and some of these errors don't look like that.

rosecodym (commented):
I have a question about some verification errors this returns.

pkg/detectors/jwt/jwt.go
bradlarsen: @rosecodym thanks for pointing this out! I agree with you. And yeah, you're right, this code changed during implementation in response to comments that clarified `indeterminate` verification status.

In a previous commit, HMAC-signed JWTs whose other claims (timestamps, etc) checked out would be reported as indeterminate. But I later learned that that type is reserved today for essentially transient / network-related errors. So I changed the `verifyHMAC` function.

It's essentially dead code today, and perhaps should be ripped out entirely, to be added back in one day in the glorious future when we can more comfortably handle secrets that are *probably* live but not feasibly tested.

pkg/detectors/jwt/jwt.go
bradlarsen: @rosecodym Yeah, you're right to remark on this. Thanks for the extra set of eyes.

JWTs are a tricky thing to do liveness checking on — different from just about all our other detectors. They are "stateless", and verification is done locally. (Once you get past the OIDC Discovery process at least.)

The `oidcDiscoveryKeyFunc` does return many error values for non-transient things, like when an issuer URL is missing or malformed. In these cases, indeed, we probably don't want to propagate the errors all the way to be indeterminate results at the finding level. I'll rework that.

I will point out for posterity that the current set of statuses (`verified`, `unverified`, and indeterminate due to external conditions) doesn't have a way of representing secrets that _may_ be live but don't have a definitive way of doing verification.

For example: the current code will produce an indeterminate result for a JWT finding whose issuer comes from a non-routing host (localhost, internal IP, etc). These _could_ in theory be checked for liveness, but the verification procedure would have to be performed from the appropriate location in the network. There is not a suitable status to use for this today.

pkg/detectors/jwt/jwt.go
```diff
@@ -0,0 +1,305 @@
...
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies expected interfaces at compile time.
+var _ interface {
+	detectors.Detector
+	detectors.MaxSecretSizeProvider
+} = (*Scanner)(nil)
+
+var keyPat = regexp.MustCompile(`\b((?:eyJ|ewogIC|ewoid)[A-Za-z0-9_-]{12,}={0,2}\.(?:eyJ|ewo)[A-Za-z0-9_-]{12,}={0,2}\.[A-Za-z0-9_-]{12,})\b`)
```
rgmz, line 35: `\b` wouldn't match signatures that end in `-`. Idk if that's a concern.

pkg/detectors/jwt/jwt.go
bradlarsen: I've changed this in the latest version. HMAC-based JWTs are skipped entirely now, even before verification.

pkg/detectors/jwt/jwt.go
bradlarsen: This is fixed in the latest version. Now *only* network and OIDC discovery-related errors (the external system interface) are propagated to result in an `unknown` status.

pkg/detectors/jwt/jwt.go
bradlarsen: That's a good point. When I've run into this sort of pattern issue with Rust `regex` + Hyperscan, instead of using `\b`, I've used something like `(?:[^A-Za-z0-9_-]|$)` — the negated preceding character class OR end of input. I wonder if that works the way I expect with Go.

rgmz: That's what I do. I'm sure there's still some nasty edge case, lol.

bradlarsen:
I ran over ~200GB of recently-pushed content to public GitHub with this branch. In that experiment, the new JWT detector is not even in the top 10 of all detectors for finding volume, with 777 `unverified`, 1 `verified` (!), and 1 `unknown` finding.

For comparison, the most voluminous detector was GitHub OAuth2, with 6961 `unverified` and 3 `verified` findings.

It's unlikely that this new detector is going to blow up TruffleHog's output.

camgunz (approved):

PR #4459, opened 2025-09-12
Add draft detector for the Smartling API Secrets [INS-38]

mustansir14:
### Description:
This PR adds an initial detector for Smartling API credentials.

**Notes**
- Regex patterns for `userIdentifier` and `userSecret` are best guesses based on public documentation and example screenshots.
- We could not generate or validate against real Smartling credentials because Smartling does not provide self-serve/free accounts to new users. Access requires an enterprise subscription, so we don’t currently have a way to obtain valid secrets.
- Integration tests use mocked API responses (`gock`) with dummy values to exercise verification flow. These should be updated once we have access to real credentials.

Marking as draft since the detector is not production-ready until regexes and verification can be confirmed.

**References (used to infer credential formats)**
- [AEM Touch Connector Config](https://help.smartling.com/hc/en-us/articles/360009611313-AEM-Touch-Connector-Configuration-5-0x#:~:text=the%20AEM%20locale%20as%20the%20source%20locale.)
- [WordPress Connector Setup](https://help.smartling.com/hc/en-us/articles/360008158133-WordPress-Connector-Installation-Setup#:~:text=Click%20Account%20Settings%20%3E%20API%20%3E%20v%202.0)
- [AEM Classic Connector Config](https://help.smartling.com/hc/en-us/articles/360007935194-AEM-Classic-Connector-Installation-and-Configuration#:~:text=In%20the%20General%20Settings%20tab)
- [Blackbird Smartling Docs](https://docs.blackbird.io/apps/smartling/#:~:text=Smartling%20via%20Blackbird.-,Connecting,-Navigate%20to%20apps)

### Checklist:

### DIFF
--- a/pkg/detectors/smartling/smartling.go
+++ b/pkg/detectors/smartling/smartling.go
@@ -71,9 +71,8 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 					client = defaultClient
 				}
 
-				isVerified, extraData, verificationErr := verifyMatch(ctx, client, userId, secret)
+				isVerified, verificationErr := verifyMatch(ctx, client, userId, secret)
 				s1.Verified = isVerified
-				s1.ExtraData = extraData
 				s1.SetVerificationError(verificationErr, userId, secret)
 				s1.SetPrimarySecretValue(secret)
 			}
@@ -86,19 +85,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return
 }
 
-func verifyMatch(ctx context.Context, client *http.Client, userId string, secret string) (bool, map[string]string, error) {
+func verifyMatch(ctx context.Context, client *http.Client, userId string, secret string) (bool, error) {
 	body := []byte(fmt.Sprintf(`{"userIdentifier":"%s","userSecret":"%s"}`, userId, secret))
 
 	req, err := http.NewRequestWithContext(
 		ctx, http.MethodPost, "https://api.smartling.com/auth-api/v2/authenticate", bytes.NewReader(body))
 	if err != nil {
-		return false, nil, err
+		return false, err
 	}
 	req.Header.Set("Content-Type", "application/json")
 
 	res, err := client.Do(req)
 	if err != nil {
-		return false, nil, err
+		return false, err
 	}
 	defer func() {
 		_, _ = io.Copy(io.Discard, res.Body)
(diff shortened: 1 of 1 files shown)


pkg/detectors/smartling/smartling.go
```diff
@@ -0,0 +1,132 @@
...
+
+			results = append(results, s1)
+		}
+
+	}
+
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, userId string, secret string) (bool, map[string]string, error) {
```
nabeelalam: The verification function doesn't really need to return a `map[string]string` since it's always `nil`

pkg/detectors/smartling/smartling.go
```diff
@@ -0,0 +1,132 @@
...
+		return true, nil, nil
+	case http.StatusUnauthorized:
+		// The secret is determinately not verified (nothing to do)
+		return false, nil, nil
+	case http.StatusBadRequest:
+		// The request was malformed. We can't verify this secret, but it might be valid.
+		return false, nil, fmt.Errorf("received HTTP 400 Bad Request from Smartling API")
+	case http.StatusTooManyRequests:
+		// We have been rate limited. We can't verify this secret, but it might be valid.
+		return false, nil, fmt.Errorf("received HTTP 429 Too Many Requests from Smartling API")
```
nabeelalam: I'm not quite sure why we need to describe these cases, since they are covered in `default`?

pkg/detectors/smartling/smartling.go
mustansir14: Good catch. I'll remove it.

pkg/detectors/smartling/smartling.go
mustansir14: This is done to provide specific errors for each case, and also as a measure to include all response status codes specified in the [documentation](https://api-reference.smartling.com/#tag/Authentication/operation/authenticate).

nabeelalam: Hmm.. I don't see any specific info in either of the error messages not covered in the `default` case since the status code is being logged in it as well. This is definitely a nit though, so feel free to ignore.

mustansir14: Hmm, makes sense. I'll remove them.

pkg/detectors/smartling/smartling_integration_test.go
```diff
@@ -0,0 +1,204 @@
...
+		args                args
+		want                []detectors.Result
+		wantErr             bool
+		wantVerificationErr bool
+		// TODO: Remove this once we have access to an actual Smartling account and can make direct API calls
+		gockSetup func()
+	}{
+		{
+			name: "found, verified",
+			s:    Scanner{client: common.SaneHttpClient()},
```
nabeelalam, line 50: In case you want to keep the mock tests later on, you can inject `common.ConstantResponseHttpClient` with the required status code and response body, so you won't have to use gock.

PR #4468, opened 2025-09-17
Pre-filter GitHub v1 findings to prevent large numbers of validation requests

trufflesteeeve:
### Description:
GitHub v1 tokens are unlikely, so verifying them should only be done if we're relatively confident that the sha256 hash we're seeing isn't actually a sha256 hash.

### Checklist:


martinlocklear (approved):
I think this makes a _ton_ of sense. In one perspective, verification is just another way to filter results, and doing the largest amount of culling early makes sense to me.

kashifkhan0771 (approved):

shahzadhaider1 (approved):

PR #4469, opened 2025-09-17
Remove include repos

jordanTunstill:
### Description:
Removed redundant IncludeRepos references from OSS, to match the changes requested in the Enterprise codebase. (https://github.com/trufflesecurity/thog/pull/4996) This is a continuation of https://github.com/trufflesecurity/trufflehog/pull/4430, where these changes were part of my original fix. Looking back, the changes here undo the unnecessary work from that PR, while leaving the load-bearing changes that fixed the original issue.

### Checklist:


jordanTunstill:
@sysread @rosecodym Here is the follow up OSS pr i mentioned to remove the interactions with the removed IncludeRepos line.

proto/sources.proto
```diff
@@ -262,19 +262,18 @@ message GitHub {
...
+  bool include_gist_comments = 19;
+  bool skip_binaries = 20;
+  bool skip_archives = 21;
+  bool include_wikis = 22;
+  uint32 comments_timeframe_days = 23;
+  bool remove_auth_in_url = 24;
+  string clone_path = 25;
+  bool no_cleanup = 26;
+  bool ignore_gists = 27;
+  bool print_legacy_json = 28;
```
rosecodym: One of the very important [protobuf guidelines](https://protobuf.dev/best-practices/dos-donts/) is to never reuse a tag number, which precludes both re-numbering fields and removing fields without reserving the tag number. (That link has details on reservation.) However, in this case, I'd honestly leave the field in and throw a deprecation annotation on it ([example](https://github.com/trufflesecurity/thog/blob/107c7b91b5a317daa4d679da9f59552603bc9e28/api/proto/IMPORTED/sources.proto#L68)) so that people know what it used to be for.

rosecodym (dismissed):

proto/sources.proto
jordanTunstill: That makes sense, thank you Cody! I'll keep this in mind for future changes and leave the field in like you mentioned.

camgunz (approved):

kashifkhan0771 (approved):

shahzadhaider1 (commented):
Looks good overall 👍 I just had one question: since `include-repos` is no longer used in the engine, should we also remove the corresponding CLI flag from `main.go`? Right now it seems like users could still set the flag, but it wouldn’t have any effect. If there’s a reason for keeping it around (e.g., backward compatibility), then all good, just wanted to double-check.

jordanTunstill:
@shahzadhaider1 I believe that the gitlab integration uses include-repos as well, and I didn't add include-repos to main.go as part of my initial change. I think, to that end, we should likely keep it around for now. Let me know if you need more info!

pkg/sources/github/github.go
```diff
@@ -253,7 +253,7 @@ func (s *Source) Init(aCtx context.Context, name string, jobID sources.JobID, so
 
 	s.filteredRepoCache = s.newFilteredRepoCache(aCtx,
 		simple.NewCache[string](),
-		append(s.conn.GetRepositories(), s.conn.GetIncludeRepos()...),
+		s.conn.GetRepositories(),
```
shahzadhaider1, line 256: > I believe that the gitlab integration uses include-repos as well,

Correct. GitLab also uses the `include-repos` flag, but that’s a separate implementation from the one being touched in this PR.

> and I didn't add include-repos to main.go as part of my initial change. I think, to that end, we should likely keep it around for now.

You’re right that you didn’t add `include-repos` to `main.go`. The concern is that since you’re removing it from `Init()` here, any user-provided value via the `include-repos` flag will now be ignored. In other words, the flag still exists but has no effect, which might cause confusion.

That said, I don’t want to be a blocker here, especially since keeping this flag only in main won't have any major impact, and also this PR is part of ongoing cleanup and aligning with the Enterprise changes. I just wanted to call out the potential confusion for users in case we want to address it now or in a follow-up so feel free to merge.

proto/sources.proto
```diff
@@ -263,6 +263,9 @@ message GitHub {
   string base = 10;
   repeated string ignore_repos = 11;
   repeated string include_repos = 12;
+  // DEPRECATED: include_repos is deprecated in favor of repositories. 
+  // It was used for filtering when the repositories field wasn't populating the UI-input results correctly
+  // It can be removed when we no longer depend on it
```
rosecodym, line 267: Deprecating fields is done via an actual field attribute, not just a comment; take another look at the end of the line that I linked to in my previous review :)

Also, it is customary to place comments before the lines they describe, rather than after; reading them the way you've written them here takes a little more work just because it's atypical.

pkg/sources/github/github.go
rosecodym: I'm also concerned about this, and possibly a little more than @shahzadhaider1. In addition to the immediate issue of the source not working the way it seems like it "should," there is the second-order problem of a future maintainer coming through here and "correcting" the obvious "error," reverting your fix.

...which, I admit, I don't quite understand. Please tell me which of these is the case:
1. This source originally did **not** respect `--include-repos`. You recently added support for `--include-repos`, but that was mistaken, and this PR reverts that addition.
1. This source has always respected `--include-repos`, but this PR removes that support because it is incorrect for some reason.

If (1) is correct, then this PR is fine, but you should put a loud, obnoxious comment here so that nobody "fixes" this in the future. If (2) is correct, then...why are we removing this? That seems very confusing.

jordanTunstill: Thank you Shahzad! I appreciate it, I can definitely see the value of addressing this in a follow-up ticket down the road.

proto/sources.proto
jordanTunstill: Sorry about that! Just went back and reread it / edited the line to match. I also moved the comment above, I'll do that moving forward.

SamyDjemai:
Hi, this PR broke our CI pipeline that periodically checks our repos for secrets, yet `--include-repos` is still mentioned in the CLI help message, and this wasn't marked as a breaking change. Is there a new way to include specific repos using glob patterns?

PR #4471, opened 2025-09-18
Add metrics to SaneHTTPClient

amanfcp:
1. Increment counter for URL
2. Time the latency
3. Increment counter for non-200 calls 
4. Added Tests

Note: Claude code helped me in writing tests.

### Checklist:

### DIFF
--- a/pkg/common/http_metrics.go
+++ b/pkg/common/http_metrics.go
@@ -1,0 +1,108 @@
+package common
+
+import (
+	"net/url"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	httpRequestsTotal = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: MetricsNamespace,
+			Subsystem: "http_client",
+			Name:      "requests_total",
+			Help:      "Total number of HTTP requests made, labeled by URL.",
+		},
+		[]string{"url"},
+	)
+
+	httpRequestDuration = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: MetricsNamespace,
+			Subsystem: "http_client",
+			Name:      "request_duration_seconds",
+			Help:      "HTTP request latency in seconds, labeled by URL.",
+			Buckets:   prometheus.DefBuckets,
+		},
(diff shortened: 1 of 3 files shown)


pkg/common/http_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// Normalize path
+	if sanitized.Path == "" {
+		sanitized.Path = "/"
+	}
+
+	// Limit path length to avoid extremely long paths creating high cardinality
+	if len(sanitized.Path) > 100 {
+		sanitized.Path = sanitized.Path[:100] + "..."
+	}
```
amanfcp, line 77: Need suggestions, if we need to do this trimming or not

pkg/common/http_metrics.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	httpRequestsTotal = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: MetricsNamespace,
+			Subsystem: "http_client",
+			Name:      "requests_total",
```
camgunz, line 16: Let's prefix all these w/ `http_`

camgunz: Or, if `Subsystem` up there already namespaces it, then this is fine 👌🏻 (I am not a prometheus expert haha)

pkg/common/http_metrics.go
camgunz: 👍🏻 I think this is good. ...Pretty sure other metrics systems do this by default too (though it's been a while since I used like, Datadog)

camgunz (approved):
LG! Thanks for the quick turnaround 🙏🏻

pkg/common/http_metrics.go
amanfcp: [Namespace, Subsystem, and Name are components of the fully-qualified](https://github.com/prometheus/client_golang/blob/a974e0d45e0aa54c65492559114894314d8a2447/prometheus/metric.go#L68-L76)

As per this doc, subsystem is already prefixing the metric. So I guess the metric title would look something like `trufflehog_http_client_requests_total`

camgunz: OK that's perfect 👍🏻

PR #4472, opened 2025-09-19
Update README with trufflehog exclude option

AndrewMohawk:
Add an exclude globs example (the help is a bit confusing to know where in the command line to add it)

### Description:
Simple example for exclude globs


AndrewMohawk:
@shahzadhaider1 looks like that test is still failing, can you try re-run again? I dont have perms for it.

Issue #4218, opened 2025-06-10
trufflehog exits with 0 in case `--since-commit` has invalid commit id


Hritik14:
### TruffleHog Version
trufflehog 3.88.29

### Trace Output

https://gist.github.com/Hritik14/64bf5f3a62049c2b53c84055d20e5a69

### Expected Behavior
since the output contains detected errors (`encountered errors during scan`), exit code should be non zero

### Actual Behavior

exit code is zero

### Steps to Reproduce

```
;  /opt/homebrew/bin/trufflehog git file://. --fail --json --force-skip-binaries --force-skip-archives --no-verification --no-update --since-commit this-commit-does-not-exist
{"level":"info-0","ts":"2025-06-10T23:13:39+05:30","logger":"trufflehog","msg":"Warning: The provided commit hash appears to be invalid."}
{"level":"info-0","ts":"2025-06-10T23:13:39+05:30","logger":"trufflehog","msg":"running source","source_manager_worker_id":"0YqnQ","with_units":true}
{"level":"error","ts":"2025-06-10T23:13:39+05:30","logger":"trufflehog","msg":"encountered errors during scan","errors":["error chunking dir \".\": unable to resolve ref: no base refs succeeded for base: \"this-commit-does-not-exist\""]}
{"level":"info-0","ts":"2025-06-10T23:13:39+05:30","logger":"trufflehog","msg":"finished scanning","chunks":0,"bytes":0,"verified_secrets":0,"unverified_secrets":0,"scan_duration":"35.912792ms","trufflehog_version":"3.88.29","verification_caching":{"Hits":0,"Misses":0,"HitsWasted":0,"AttemptsSaved":0,"VerificationTimeSpentMS":0}}


; echo $?
0
```

## Environment
 * OS: MacOS
 * Version 15.3.2 (24D81)

shahzadhaider1:
Hi @Hritik14,

Thank you for taking the time to open this issue, we really appreciate your contribution to the project!

We’ll take a look and get back to you as soon as we can. If we need any clarification, we’ll follow up here. Thanks again for helping make this project better!

mplzik:
We've been running Trufflehog in CI and also got hit by this issue; see below for the test cases. For a security scanner, this behavior is quite dangerous since it can be disabled by a simple typo/non-obvious misconfiguration or in some cases, by problems with checking out a repository. 

## Case 1
When Trufflehog is configured as a Github Action running on PRs:
```
    steps:
[...]

PR #4476, opened 2025-09-25
Return non-zero exit code if an error occurs during a scan.

mplzik:
### Description:
With its original code, Trufflehog was returning a zero exit code in
cases when an error was encountered during a scan. This led to some
unexpected situations, such as succeeding if a git repo was not cloned
correctly or if a non-existent commit was referenced from
`--since-commit`.

This PR proposes adding a new flag `--fail-on-scan-errors` that, if
enabled, will propagate scan errors further (alongside with the current
behavior of reporting them on console), ensuring that Trufflehog returns
a non-zero exit code.

The change should be fairly safe, as it is hidden behind a flag and if
not activated, the original behavior is retained.

See also: https://github.com/trufflesecurity/trufflehog/issues/4218

Signed-off-by: Milan Plzik <milan.plzik@grafana.com>

### Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -72,6 +72,7 @@ var (
 	printAvgDetectorTime = cli.Flag("print-avg-detector-time", "Print the average time spent on each detector.").Bool()
 	noUpdate             = cli.Flag("no-update", "Don't check for updates.").Bool()
 	fail                 = cli.Flag("fail", "Exit with code 183 if results are found.").Bool()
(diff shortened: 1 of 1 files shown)


mplzik:
@rosecodym happy to help. Also, I was being fairly conservative with adding the change behind a flag -- since a fair amount of users might have this failing silently, it might be worth to gradually roll this out as an enabled-by-default flag (e.g. slowly add a warning about not using the flag, then make the flag enabled by default; then remove the option altogether, if it makes sense). This is a choice that lies on the Trufflehog maintainers, though.

PR #4484, opened 2025-10-03
Proto update to support scanning confluence comments

mustansir14:
### Description:
This PR updates the proto files to support scanning confluence comments. 

### Checklist:


proto/source_metadata.proto
```diff
@@ -66,6 +66,7 @@ message Confluence {
   string location = 7;
   string file = 8;
   string user = 9;
+  string comment = 10;
```
rosecodym: Unlike all the other fields, I can't form a good guess about what's stored in this just by reading its name. Is it a comment ID or something?

mustansir14: It is the comment ID. I tried to make it consistent with other fields, but you're right, it should be clearer. I'll make the change.

proto/source_metadata.proto
```diff
@@ -66,6 +66,7 @@ message Confluence {
   string location = 7;
   string file = 8;
   string user = 9;
+  string commentId = 10;
```
shahzadhaider1: Can we follow the existing `snake_case` here?

mustansir14: Good catch. Fixed it.

shahzadhaider1 (approved):
LGTM

rosecodym (approved):

PR #4486, opened 2025-10-04
Fix wrong line number in private key detector (#4485)

kmgv:
### Description:

Private key detector findings report wrong line number when private key literal doesn't end with new line character.

After private is matched using regexp is goes through Normalize function and normalized result is used in result.Raw and then used in engine.FragmentLineOffset which looks for line of code.

Normalization step is crucial as ssh.ParseRawPrivateKey is quite strict about format of accepted key and this step can sieve false posivites as it can verify it private key is legit or just matches permisive regexp.

Normalize always adds newline char at end of string (as needed for validation) but such string, with new line at the end is then used for looking for LOC. If source chunk didn't have new line char right after private key engine will report default LOC.

This fix changes Result.Raw for private key detector to use raw match from regexp and not normalized string. This way engine can calculate correct LOC for such finding.

Related issue #4485 

### Checklist:
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))? - it doesn't even pass for main branch locally


pkg/detectors/privatekey/privatekey.go
```diff
@@ -66,7 +66,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 		s1 := detectors.Result{
 			DetectorType: detectorspb.DetectorType_PrivateKey,
-			Raw:          []byte(token),
+			Raw:          []byte(match),
```
kashifkhan0771: We don't recommend updating the `Raw` field value, as it's used as an identifier and changing it can interfere with existing findings. Your solution is correct, but instead, you can use:

```go
s1.SetPrimarySecretValue(match)
```

This will instruct the engine to use `match` to determine the line number.

kashifkhan0771 (changes requested):

pkg/detectors/privatekey/privatekey.go
kmgv: updated as requested

shahzadhaider1 (approved):

kashifkhan0771 (approved):

kashifkhan0771:
@trufflesecurity/backend This PR involves a detector change, so I’m not sure why the backend team is the owner. Could someone from the Backend team please review and merge it?

shahzadhaider1:
> @trufflesecurity/backend This PR involves a detector change, so I’m not sure why the backend team is the owner. Could someone from the Backend team please review and merge it?

@trufflesecurity/backend owns [critical detectors](https://github.com/trufflesecurity/trufflehog/blob/f9dd66072919ca77856671104331aab20de84556/CODEOWNERS#L26)

rosecodym (approved):
Thanks for this! I'm mostly deferring to @kashifkhan0771 :)

PR #4492, opened 2025-10-08
[detector] feat: Rootly Webhook Detector

nam3lum:
### Description:
This PR Adding Rootly Webhook Detector

image

### Checklist:


pkg/detectors/rootlywebhook/rootlywebhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	client = common.SaneHttpClient()
+
+	// Rootly webhook tokens are 64 character hex strings
+	keyPat = regexp.MustCompile(`\b([a-f0-9]{64})\b`)
+)
```
shahzadhaider1, line 25: This is a generic token; we should use prefix regex to minimize false positives. That way it will only detect the secrets that are in the proximity of the keywords we provide. 
Example: https://github.com/trufflesecurity/trufflehog/blob/b140e2b58e65881729b6035ad03a3c0b23d79985/pkg/detectors/sentrytoken/v1/sentrytoken.go#L33

pkg/detectors/rootlywebhook/rootlywebhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"rootly", "webhook"}
```
shahzadhaider1: `webook` is a very common term and may cause unnecessary detector triggers. Please use more specific keywords, maybe `rootlywebhook` or something similar

These keywords are used by TruffleHog to decide when to run a detector, so it’s important that they closely match the target service.

pkg/detectors/rootlywebhook/rootlywebhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return results, nil
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
+	// We don't want to actually create alerts in Rootly. To verify tokens without spamming them,
+	// we send a payload that typically causes a 500 error (parsing issue) but still validates the auth.
+	// The expected scenario is 500 which means the key is working but the payload format causes an error.
+	// In case 200 comes, it means an actual alert has been created in Rootly (hopefully this never happens).
+	payload := bytes.NewReader([]byte(`{"rootly":["TruffleHog"]}`))
```
shahzadhaider1, line 66: Thanks for the info and you’re right, we should never be getting a 200 here.
With that in mind, do we have any other APIs that don’t create or modify resources? Maybe a GET endpoint? The goal is to validate the secrets without unintentionally creating or updating anything in the user’s environment.
If that’s not possible, we could consider sending a payload that will always return an error, for example, using a random UUID or some unique value that’s unlikely to exist on the server. I’m not entirely sure about the API payload structure, but I hope you get the idea.

pkg/detectors/rootlywebhook/rootlywebhook.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return false, err
+	}
+	defer res.Body.Close()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		// 200: Successfully processed the webhook - this means an actual alert was created in Rootly.
+		// Hopefully this never happens, but we at least know the token is verified.
+		return true, nil
```
shahzadhaider1, line 86: What a 200 status will feel like:

Image

shahzadhaider1 (changes requested):

pkg/detectors/rootlywebhook/rootlywebhook.go
nam3lum: This is the only payload I’ve found that’s currently causing a server-side issue. Unfortunately, the endpoint returns a 404 for all requests except POST.
We also can’t validate whether the token exists, since the application logic first attempts to parse the request body before checking the token’s validity, so sending malformed JSON doesn’t help either.

On the other hand, the Rootly webhook is quite permissive; it accepts and parses any JSON you send and includes the data in the alert details. Even sending default JSON strings like "ID" doesn’t break the app, unless you pass "rootly" as a list while app tries to parse it as object.
I hope this pic can explain what I'm trying to achieve a bit:
image

pkg/detectors/rootlywebhook/rootlywebhook.go
nam3lum: Thanks for highlight, I thought about it. Rootly API tokens have such prefix (as you can see below), however, webhook secrets lack that.

https://github.com/trufflesecurity/trufflehog/blob/2f1baea40da850cf64656bb98e0e91935ccc7b0d/pkg/detectors/rootly/rootly.go#L24

amanfcp (dismissed):
LGTM!

pkg/detectors/rootlywebhook/rootlywebhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// Rootly webhook tokens are 64 character hex strings
+	keyPat = regexp.MustCompile(`\b([a-f0-9]{64})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"webhooks.rootly.com"}
+}
```
shahzadhaider1, line 31: This is great and will greatly reduce noise but I think it is more restrictive than it should be. What I mean is: `webhooks.rootly.com` will only appear if the webhook URL itself is present in the file or config (e.g. https://webhooks.rootly.com/incoming/...).

That’s great for catching leaked webhook endpoints, but it won’t catch files that only contain the secret, which is often stored separately (in .env, YAML, or JSON).

Many developers store only the secret like this:
```
ROOTLY_WEBHOOK_SECRET=abcdef123456
```
or
```
rootly:
  webhook_secret: "abcdef123456"
```
In those cases, the detector won’t trigger because `webhooks.rootly.com` isn’t present.

So, my recommendation would be to use a combination of provider and contextual keywords to improve coverage. For example, we can include `rootly`, `rootly_webhook`/`rootlywebhook` along with what we have already added. Let me know if this makes sense.

pkg/detectors/rootlywebhook/rootlywebhook.go
shahzadhaider1: Yes, I understand that the webhook secrets don’t have a specific prefix. What I wanted to suggest is that we use the `detectors.PrefixRegex` helper, similar to the example I shared earlier:
```go
 keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"rootly", "webhook"}) + `\b([a-f0-9]{64})\b`) // you can add/remove the keywords from here based on how rootly webhook are normally stored and what keywords might be found in their proximity
```

The reason is that a 64-character hex string is quite common and could easily lead to false positives.
Using `detectors.PrefixRegex` ensures that at least one of the given keywords appears within 40 characters before the potential secret, helping to keep the detection contextually accurate and reduce noise.

nam3lum: Thanks! I get it now, it was helpful!

pkg/detectors/rootlywebhook/rootlywebhook.go
nabeelalam: Since the string `rootly` is already a keyword, all other keywords are redundant since they all contain `rootly`.

pkg/detectors/rootlywebhook/rootlywebhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		// 200: Successfully processed the webhook - this means an actual alert was created in Rootly.
+		// Hopefully this never happens, but we at least know the token is verified.
+		return true, nil
+	case http.StatusInternalServerError:
+		// 500: Auth is valid but there was a server error (e.g., parsing issue with our test payload)
+		// This is the expected response that indicates the token is valid without creating alerts.
+		return true, nil
```
shahzadhaider1, line 90: This makes sense in principle, but I’m concerned about treating `500` as a hard signal that a token is valid. 500 is a generic server error and could indicate a service outage or other problem. If Rootly returns 500 for many requests during an outage, we could end up marking a lot of secrets as verified incorrectly.

Would it make sense to only treat 500 as valid when the response body contains a known, stable parsing-error signature? Basically, some logic to distinguish between **“500 due to a genuine server issue/outage”** and **“500 due to our test payload being malformed.”** Does that sound reasonable?

amanfcp: +1 This is a valid concern.

nam3lum: There is no exact message from the server side tho. I'll try to communicate with Rootly team to see if they can provide an endpoint for checking webhook validity.

PR #4494, opened 2025-10-09
Reduce GitHub V1 Detector False Positives

shahzadhaider1:
## Problem
The GitHub V1 detector is extremely noisy, with less than 1 in 10,000 reported findings being actually live tokens. The current regex pattern `[a-f0-9]{40}` matches any 40-character hex string, causing false positives from:
- Git commit hashes
- SHA-1 hashes  
- Tree/blob identifiers
- Other non-sensitive hex strings

## Solution
Tighten the regex pattern to require both GitHub context AND credential-specific keywords, reducing spurious matches by ~10,000x.

## Changes Made
1. **Updated regex pattern** to use combined keywords:
   - Changed from broad `github`, `gh`, `pat`, `token` keywords
   - Now requires explicit credential context: `github_token`, `github_secret`, `github_key`, etc.
   - Leverages existing `PrefixRegex` utility for consistency

### Testing
Comprehensive testing performed against diverse test cases:

#### True Positives (Should Match - All Detected)
- [x] Standard assignments: `github_token=`, `github_secret:`, `gh_key=`
- [x] Various separators: `=`, `:`, `>>>`, `=>`, quotes
- [x] Multiple formats: JSON, YAML, .env files, shell exports
- [x] Embedded tokens in config strings and logs
- [x] Multiline/whitespace variations
- [x] GitHub Actions environment variables
- [x] Comments with accidentally leaked secrets
- [x] URLs with embedded tokens (e.g., `https://gh_token:abc123@api.github.com`)
- [x] JSON arrays and heredocs
- [x] Mixed case keywords (case-insensitive matching)

#### False Positives (Should NOT Match - All Filtered)
- [x] Git commit hashes (`commit 9d14621e59f22c2b...`)
- [x] Git tree/parent/blob identifiers
- [x] GitHub API URLs with commit references
- [x] Modern GitHub PAT tokens (github_pat_11...)
- [x] Generic hash values without GitHub context
- [x] SHA256/SHA1 checksums and integrity hashes
- [x] Python wheel hashes, npm integrity hashes
- [x] Go module sums
- [x] ETags and cache keys
- [x] Tokens beyond 40-char distance from keyword
- [x] Uppercase hex strings (GitHub v1 tokens are lowercase only)
- [x] AWS keys, Stripe tokens, other service credentials
- [x] Repository/commit metadata in JSON

Attaching the file containing all the test secrets and false positives: [github-v1.txt](https://github.com/user-attachments/files/22792408/github-v1.txt)

### Checklist:


kashifkhan0771 (approved):

dustin-decker (approved):
Thank you

PR #4502, opened 2025-10-15
Local Git Config Sanitization

joeleonjr:
This PR prevents TruffleHog from executing arbitrary commands located in archived malicious git repositories. Thanks to Adam Reiser at Cisco Talos for pointing this out! This is being tracked as CVE-2025-41390.

This approach uses Git's recommended best practice for sanitizing untrusted git configs: `git clone` all local `file://` git repos prior to scanning. Executing `git clone` does not execute any of the potentially malicious git configs in the untrusted repo directory, and the output only includes "safe" default configs, similar to what we see when cloning from remote.

We explored a few other approaches (allowlist, denylist, etc), but those carried lots of complications.

A few notes about how this works:
- This only applies to local repositories scanned using the `git` subcommand.
- Remote git targets are not impacted.
- Local git targets are now cloned to temp by default prior to scanning. Users can specify a `--clone-path` argument if they don't want to use the default temp dir. Users can specify `--trust-local-git-config` if they want to trust the repo as is and bypass cloning.
- Local `--bare` repos are handled appropriately.
- This approach knocks out (most...all?) of this class of malicious git config vulnerabilities.

Testing:
There's coverage for most of the new code, including: test cases for the specific issue reported, local bare clones, local repos with staged commits, etc. All are passing.

### Checklist:


rosecodym (approved):
Thanks @joeleonjr!

PR #4507, opened 2025-10-16
explicit repositories now bypass wantRepo() filtering entirely. added ctx to newConnector

jordanTunstill:
This change addresses an issue created by a prior PR: https://github.com/trufflesecurity/trufflehog/pull/4430
The previous PR addressed the Repositories to Include field in the Enterprise UI not working, but inadvertently impacted the explicit inclusion of repositories.
This new PR adds a bypass for explicitly included repositories, to resolve this.
Also added ctx to the github integration test and confirmed explicit repos are being found in tests.

### Description:

### Checklist:

### DIFF
--- a/pkg/sources/github/github_integration_test.go
+++ b/pkg/sources/github/github_integration_test.go
@@ -957,7 +957,7 @@ func TestSource_Validate(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			connector, err := newConnector(tt.sourceConfig)
+			connector, err := newConnector(ctx, tt.sourceConfig)
 			require.NoError(t, err)
 			tt.sourceConfig.connector = connector
 
--- a/pkg/sources/github/github_test.go
+++ b/pkg/sources/github/github_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-github/v67/github"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/types/known/anypb"
 	"gopkg.in/h2non/gock.v1"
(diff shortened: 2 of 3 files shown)


mcastorina (commented):
This is a complex source with a lot of moving parts and I'm concerned this is adding another layer of control flow that could miss something.

I'd feel better if we had tests for all these different scenarios, but that feels out of scope for this PR. At a minimum could we add a test for *this* specific scenario?

jordanTunstill:
@mcastorina happy to get a test added.

kashifkhan0771:
Hey @jordanTunstill Before I review this PR, could you please double-check that the integration test cases are working as expected?

Before the change in #4430, the tests were running correctly(Correct me If I am wrong @rosecodym), but afterward, they stopped scanning the explicitly listed repositories in the test cases. We need to make sure everything is functioning properly with this solution.

This is a commonly used source, so I’m just being extra cautious as @mcastorina mentioned, we want to make sure we’re not introducing a new issue while fixing another.

Thanks!

jordanTunstill:
Sure thing @kashifkhan0771. I just ran integrations tests again and got the following
TestSource_ScanChunks: PASS with "num_repos": 7 
TestSource_ScanComments - PASS with "num_repos": 1 
 The only test failing is this one, which looks to be unrelated:
 `{
			name: "token authenticated, single repo, no .git",
			init: init{
				name: "test source",
				connection: &sourcespb.GitHub{
					Repositories: []string{"https://github.com/truffle-test-integration-org/another-test-repo"},
					Credential: &sourcespb.GitHub_Token{
						Token: githubToken,
					},`

kashifkhan0771:
> Sure thing @kashifkhan0771. I just ran integrations tests again and got the following TestSource_ScanChunks: PASS with "num_repos": 7 TestSource_ScanComments - PASS with "num_repos": 1 The only test failing is this one, which looks to be unrelated: `{ name: "token authenticated, single repo, no .git", init: init{ name: "test source", connection: &sourcespb.GitHub{ Repositories: []string{"https://github.com/truffle-test-integration-org/another-test-repo"}, Credential: &sourcespb.GitHub_Token{ Token: githubToken, },`

Awesome!

kashifkhan0771 (approved):

kashifkhan0771:
Code looks fine but can you check the failing test please.

pkg/sources/github/github_integration_test.go
```diff
@@ -957,7 +957,7 @@ func TestSource_Validate(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			connector, err := newConnector(tt.sourceConfig)
+			connector, err := newConnector(ctx, tt.sourceConfig)
```
kashifkhan0771, line 960: ❤️

pkg/sources/github/github_test.go
```diff
@@ -1046,6 +1047,77 @@ func TestRepositoryFiltering(t *testing.T) {
...
+	}
+
+	// Set up include filter that would normally filter out the explicit repos
+	includePatterns := []string{"org/filtered-*"} // Only include repos starting with "filtered-"
+
+	// Create a filtered cache with these patterns
+	cache := simple.NewCache[string]()
+	source.filteredRepoCache = source.newFilteredRepoCache(ctx, cache, includePatterns, []string{})
+
+	// Add the explicit repos to the cache (simulating what happens in Init)
```
mcastorina: Is there a reason we can't call `Init` and need to simulate it here?

pkg/sources/github/github_test.go
```diff
@@ -1046,6 +1047,77 @@ func TestRepositoryFiltering(t *testing.T) {
...
+	for _, repo := range source.conn.Repositories {
+		normalized, err := source.normalizeRepo(repo)
+		require.NoError(t, err)
+		source.filteredRepoCache.Set(repo, normalized)
+	}
+
+	// Simulate the Enumerate logic
```
mcastorina: Same question here about simulation.. I thought we wanted to test the `Enumerate` logic?

mcastorina (changes requested):
Thanks for adding a test! Unless I missed something though, it doesn't look like it's testing the actual `Enumerate` method and instead copied the logic into the test.

Here is where a TDD (test-driven development) approach would be useful: write a test that fails, then update the code to make the test pass.

pkg/sources/github/github_test.go
jordanTunstill: Good call. I'll do this!

pkg/sources/github/github_test.go
jordanTunstill: same as above, that makes sense.

mcastorina (approved):
Awesome, thanks @jordanTunstill! Love the use of `gock` too!

PR #4509, opened 2025-10-16
Fix "skipping binary file" logging to show actual file name

mariduv:
This was intended to show path:fileName, but `path` in these scopes is the repo path, not a file.  This section was moved in a refactor where `path` was the file in the old scope, and since both scopes have `path string`, it was not flagged and easy to miss. (Repo path is contextually logged in these lines as repo:path, also.)

### Checklist:


trufflesteeeve (approved):
Great catch! LGTM.

PR #4514, opened 2025-10-21
Scan all images under a namespace for Docker remote registries

kashifkhan0771:
### Description:
This PR add support for scanning images under a namespace (Org or user) in remote registries.
Supported Remote Registries:
- Docker Hub
- Quay
- GHCR 

### Checklist:


pkg/sources/docker/docker.go
```diff
@@ -77,6 +77,35 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
+			}
+		case strings.HasPrefix(namespace, "ghcr.io"):
+			registry = &GHCR{
+				Token: s.conn.GetRegistryToken(),
+			}
+		default: // default is dockerhub
+			registry = &DockerHub{
+				Token: s.conn.GetRegistryToken(),
+			}
+		}
```
mustansir14: To improve readibility we can move this logic to a factory method like `MakeRegistryFromNamespace`

pkg/sources/docker/registries_test.go
mustansir14, line 1: Would be great if we could write tests for rest of the registries as well.

Edit: I see we are testing `Quay` as well. I would suggest we do that in a separate test. Also the name of the test function `TestDockerHubListImages` also suggests we're only testing for DockerHub here

main.go
```diff
@@ -931,11 +933,25 @@ func runSingleScan(ctx context.Context, cmd string, cfg engine.Config) (metrics,
 			refs = []sources.JobProgressRef{ref}
 		}
 	case dockerScan.FullCommand():
+		if *dockerScanImages != nil && *dockerScanNamespace != "" {
+			return scanMetrics, fmt.Errorf("invalid config: you cannot specify both images and namespace at the same time")
```
mustansir14, line 940: Why can't we have both images and namespace at the same time?

mustansir14 (commented):
Great work!

pkg/sources/docker/registries_test.go
kashifkhan0771: Yes, I am currently working on test cases.

main.go
kashifkhan0771: We’ve been following this convention across other sources as well. The `--images` and `--namespace` flags represent two distinct workflows:

- `--images`: Used to scan one or more explicitly specified images.
- `--namespace`: Used to automatically discover and scan all images under a given namespace (organization or user), similar to how the GitHub source distinguishes between `--org` and `--repo`.

Both flags can technically overlap, but they are designed for different use cases. Therefore, in the CLI, we allow the user to choose only one of these options per execution just as we do for GitHub’s organization and repository flags to maintain clarity and prevent conflicting inputs.

pkg/sources/docker/registries.go
```diff
@@ -0,0 +1,242 @@
...
+	ListImages(ctx context.Context, namespace string) ([]string, error) // list all images
+}
+
+// MakeRegistryFromNamespace returns a Registry implementation
+// based on the namespace prefix (e.g. "ghcr.io/", "quay.io/").
+// If no known prefix is found, DockerHub is used by default.
+func MakeRegistryFromNamespace(namespace string) Registry {
+	var registry Registry
+	switch {
+	case strings.HasPrefix(namespace, "quay.io"): // quay.io/abc123
```
camgunz: Let's be a little nervous and add an ending "/" like "quay.io/" to rule out like "quay.ioo" or whatever people will do

pkg/sources/docker/registries.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func (d *DockerHub) WithRegistryToken(registryToken string) {
+	d.Token = registryToken
+}
+
+// ListImages lists all images under a Docker Hub namespace using Docker Hub's API.
+func (d *DockerHub) ListImages(ctx context.Context, namespace string) ([]string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet,
+		fmt.Sprintf("https://hub.docker.com/v2/namespaces/%s/repositories", namespace), http.NoBody)
```
camgunz: I know it's a little annoying, but let's build a `net.URL` and use like [`JoinPath`](https://pkg.go.dev/net/url#URL.JoinPath) to build the `v2/namespaces/%s/repositories" bit -- same for the other repos too.

pkg/sources/docker/docker.go
```diff
@@ -77,6 +77,26 @@ func (s *Source) Init(_ context.Context, name string, jobId sources.JobID, sourc
...
+		ctx.Logger().Info(fmt.Sprintf("using registry: %s", registry.Name()))
+
+		namespaceImages, err := registry.ListImages(ctx, namespace)
+		if err != nil {
+			return fmt.Errorf("failed to list namespace images: %w", err)
+		}
+
+		ctx.Logger().Info(fmt.Sprintf("namespace: %s has %d images", namespace, len(namespaceImages)))
+
+		s.conn.Images = append(s.conn.Images, namespaceImages...)
```
camgunz: I like keeping it simple here and piggybacking on `s.conn.Images`, but I think it'll make integration testing a little tricky because we're making network calls in `Init` here. My first idea for making this easier is to break it out into a helper like `getImages` that `Chunks` can call, but there's other options.

camgunz (changes requested):
This is super cool! I love the interface over the different repos, and the additions to the README are very nice.

I can go into more detail on Slack when I get back from lunch etc., but the changes I want are pretty much all "help this source survive in Enterprise" stuff:
* metrics
  * how long our API calls are taking
  * how many API call successes and failures
  * how many images/histories/layers we enumerate/actually scan
* all errors are logged
* each function/method gets a trace log at level 5 that we're calling it
* integration tests for at least the `ListImages` implementations
  * probably just have to mock these w/ `gock`; I don't think we have a good enough test repo to really test this new functionality

This is a lot of stuff, but I imagine this source will get a lot of use once it can scan registries, so the more we can get ahead of debugging and troubleshooting the better. Also this is mostly my fault; I have "put together logging/metrics guidelines" on my TODO list and I've not gotten to it yet. We'll probably go through another round or two of refinements too, just fair warning haha. But again this is a really good, solid start--core functionality looks perfect.

kashifkhan0771:
> how long our API calls are taking

Added metrics to track API call durations.

> how many API call successes and failures
> how many images/histories/layers we enumerate/actually scan

We already have existing metrics for the number of images scanned.

> all errors are logged

All errors within Chunk() are now logged.

> each function/method gets a trace log at level 5 that we're calling it

Done ✅ where we are passing context

> integration tests for at least the `ListImages` implementations

I already added test cases for ListImages in registries_test.go, covering all registries using public namespaces.

shahzadhaider1 (dismissed):

pkg/sources/docker/docker.go
```diff
@@ -139,6 +119,22 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	// if namespace is set and no images are specified, fetch all images in that namespace.
+	registryNamespace := s.conn.GetNamespace()
+	if registryNamespace != "" && len(s.conn.Images) == 0 {
+		start := time.Now()
+		namespaceImages, err := GetNamespaceImages(ctx, registryNamespace, s.conn.GetRegistryToken())
+		if err != nil {
+			ctx.Logger().Error(err, "failed to list namespace images", "namespace", registryNamespace)
+
+			return nil
+		}
```
camgunz, line 129: I think we should just do a `fmt.Errorf` to wrap and then return the error; this should make the scanner report unhealthy

camgunz (dismissed):
Just a quick change in `Chunks` and then we can merge

camgunz (approved):

PR #4516, opened 2025-10-23
Reduce False Positives in Twilio Detector

shahzadhaider1:
This PR addresses false positive issues in the Twilio detector by making the regex patterns more context-aware and removing an overly generic keyword.

### 1. Added Context-Aware Regex Patterns
Updated both `sidPat` and `keyPat` to require contextual keywords within 40 characters of the credential:
```go
// Before
sidPat = regexp.MustCompile(`\bAC[0-9a-f]{32}\b`)
keyPat = regexp.MustCompile(`\b[0-9a-f]{32}\b`)

// After
sidPat = regexp.MustCompile(detectors.PrefixRegex([]string{"twilio", "account", "sid"}) + `\b(AC[0-9a-f]{32})\b`)
keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"twilio", "auth", "token", "key"}) + `\b([0-9a-f]{32})\b`)
```

**Why:** The previous `keyPat` matched any 32-character hexadecimal string, which is extremely common in codebases (MD5 hashes, commit SHAs, etc.). By requiring proximity to Twilio-related keywords, we significantly reduce false matches while maintaining detection of legitimate credentials.

### 2. Removed "sid" from Keywords
```go
// Before
    return []string{"sid", "twilio"}
}

// After
    return []string{"twilio"}
}
```

**Why:** The keyword "sid" is extremely common in code (session IDs, database fields, variable names like `user_sid`, `request_sid`, etc.) and was causing the detector to run unnecessarily on a large percentage of scanned files. Since Twilio Account SIDs always start with "AC" and our regex already requires contextual keywords, keeping only "twilio" as the trigger is sufficient and improves performance.

### 3. Switched to FindAllStringSubmatch
Updated the pattern matching to use `FindAllStringSubmatch` instead of `FindAllString`:
```go
// Before
keyMatches := keyPat.FindAllString(dataStr, -1)
sidMatches := sidPat.FindAllString(dataStr, -1)

// After
keyMatches := keyPat.FindAllStringSubmatch(dataStr, -1)
sidMatches := sidPat.FindAllStringSubmatch(dataStr, -1)

for _, sidMatch := range sidMatches {
    sid := sidMatch[1]  // Extract capture group
```

**Why:** With the addition of capturing groups in the regex patterns, we need `FindAllStringSubmatch` to properly extract just the credential values (capture group [1]) without the surrounding context keywords that are used for filtering.

## Impact
- **Reduces false positives:** Only matches hex strings that appear near Twilio-related keywords
- **Improves performance:** Detector runs less frequently by removing the generic "sid" keyword
- **Maintains detection accuracy:** Legitimate Twilio credentials will still be detected as they typically appear with contextual keywords like "twilio_auth_token", "TWILIO_ACCOUNT_SID", etc.

## Testing
Verified that the detector still matches valid Twilio credentials in common formats while filtering out unrelated hex strings and reducing unnecessary detector invocations.

### Checklist:

twilio-detector-statistics
twilio-detector-tests

---

> [!NOTE]
> **Medium Risk**
> Stricter matching may miss credentials stored without nearby Twilio context or without the "twilio" keyword in the chunk; verification error propagation changes how failed verifies are reported.
> 
> **Overview**
> Tightens the **Twilio** account SID and auth token detector so it no longer treats arbitrary 32-character hex strings as secrets.
> 
> **SID and key regex** now use `PrefixRegex` so a Twilio-related label (e.g. account/sid, auth/token) must appear within 40 characters of the match, with capture groups extracted via `FindAllStringSubmatch`. The Aho-Corasick **keyword** list drops generic `sid`, leaving only `twilio`, so chunks without that name are not scanned.
> 
> **Verification** in `verifyTwilio` propagates request and HTTP client errors instead of returning `nil` errors, which affects timeout and failure signaling in integration tests.
> 
> **Tests** add env/JSON/Postman-style positives, false-positive cases (hashes, unrelated hex), keyword gating checks, and updated integration fixtures with contextual secret text. A **TODO** comment is added in `twilioapikey` about using the correct `DetectorType` (no behavior change there).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 572423150c0717c0864409e04f1b329c745b7600. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/twilio/twilio.go
```diff
@@ -49,18 +49,20 @@ func (s Scanner) getClient() *http.Client {
...
 
-	keyMatches := keyPat.FindAllString(dataStr, -1)
-	sidMatches := sidPat.FindAllString(dataStr, -1)
+	keyMatches := keyPat.FindAllStringSubmatch(dataStr, -1)
+	sidMatches := sidPat.FindAllStringSubmatch(dataStr, -1)
 
-	for _, sid := range sidMatches {
-		for _, key := range keyMatches {
+	for _, sidMatch := range sidMatches {
+		sid := sidMatch[1]
```
rgmz: Not required after Go 1.22.
https://go.dev/blog/loopvar-preview

camgunz (approved):
I think your analysis here is correct--the profile indicated that `regexp` was allocating a lot and the O(N2) combined with a very common pattern is probably why that was happening. Let's merge this and be very careful w/ the rollout--I'll sync up w/ you on Slack

pkg/detectors/twilio/twilio.go
```diff
@@ -49,18 +49,20 @@ func (s Scanner) getClient() *http.Client {
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
```
martinlocklear, line 52: I really think we need to have unit test coverage the covers this case, especially since it's already been causing problems, just to prevent regressions in the future (and make absolutely clear _why_ we're making this change).

martinlocklear (changes requested):
This really feels like something we want to have unit test coverage on.  Feel free to push back if that's not the style in this area of code, but (as I read it) this begs for a unit test to make sure that the new regex isn't catching the false positives that it was previously.

pkg/detectors/twilio/twilio.go
shahzadhaider1: added

pkg/detectors/twilio/twilio.go
```diff
@@ -23,9 +23,9 @@ type Scanner struct {
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	defaultClient = common.RetryableHTTPClient()
-	sidPat        = regexp.MustCompile(`\bAC[0-9a-f]{32}\b`)
-	keyPat        = regexp.MustCompile(`\b[0-9a-f]{32}\b`)
+	defaultClient = common.SaneHttpClient()
+	sidPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"twilio", "account", "id", "sid"}) + `\b(AC[0-9a-f]{32})\b`)
```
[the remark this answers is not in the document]
shahzadhaider1: Earlier, there were no keywords at all, so this is better than before, if we talk about false positives IMO.

MuneebUllahKhan222 (approved):

PR #4520, opened 2025-10-29
chore: fix feature support to append user agent suffix

dmdhrumilmistry:
### Description:
fixes missing feature support append suffix to `TruffleHog` User Agent header while using `--user-agent-suffix` flag.

Feature implementation was missed in this PR: https://github.com/trufflesecurity/trufflehog/pull/2208/files 

### Checklist:

image

### DIFF
--- a/pkg/detectors/browserstack/browserstack.go
+++ b/pkg/detectors/browserstack/browserstack.go
@@ -11,6 +11,7 @@ import (
 	regexp "github.com/wasilibs/go-re2"
 	"golang.org/x/net/publicsuffix"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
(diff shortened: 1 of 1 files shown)


pkg/detectors/browserstack/browserstack.go
```diff
@@ -93,6 +94,13 @@ func verifyBrowserStackCredentials(ctx context.Context, client *http.Client, use
 		return false, err
 	}
 	req.Header.Add("Content-Type", "application/json")
+
+	// add feature for custom user agent if set
+	userAgent := common.UserAgent()
+	if userAgent != "TruffleHog" {
+		req.Header.Add("User-Agent", userAgent)
+	}
+
```
kashifkhan0771:
```go
req.Header.Add("User-Agent", common.UserAgent())
```

kashifkhan0771 (commented):
Please test this with both live and non-live credentials to ensure everything functions correctly. We want to avoid any risk of breaking an existing detector.

pkg/detectors/browserstack/browserstack.go
dmdhrumilmistry: addressed it in latest commit https://github.com/trufflesecurity/trufflehog/pull/4520/commits/87f966065b22993a46ffee25c549957b1cda05ef

dmdhrumilmistry:
> Please test this with both live and non-live credentials to ensure everything functions correctly. We want to avoid any risk of breaking an existing detector.

Test Cases Output Gist: https://gist.github.com/dmdhrumilmistry/d84bcdd70f5dd47e204eebea845ed310

kashifkhan0771 (approved):
LGTM! Will check with team if we can merge this.

camgunz (approved):

PR #4521, opened 2025-10-29
[Update] Added Forbidden Response Check in PagerDuty Detector Verification

nabeelalam:
### Description:
Improved PagerDuty detector logic for credential validation.

Added handling for PagerDuty’s 403 Forbidden response for API tokens which are valid but do not have access to the resource:
```
{
  "error": {
    "message": "Access Denied",
    "code": 2010
  }
}
````

If the 403 response body contains the string "Access Denied" and the code "2010", the detector now treats this specific case as a successful credential validation (`return true, nil`).

Any other 403 response will return an unverified response with verification error since credential validity can’t be confirmed.

### Checklist:

### CODE REVIEW

**Review by shahzadhaider1 — APPROVED** [review]

**Review by mustansir14 — APPROVED** [review]

**Review by amanfcp — APPROVED** [review]

File: pkg/detectors/pagerdutyapikey/pagerdutyapikey.go [hunk 1]
```diff
@@ -84,6 +85,20 @@ func verifyPagerdutyapikey(ctx context.Context, client *http.Client, token strin
...
+	case http.StatusForbidden:
+		bodyBytes, err := io.ReadAll(res.Body)
+		if err != nil {
+			return false, err
+		}
+		bodyStr := string(bodyBytes)
+
+		// 403 with "Access Denied" and "2010" means valid credentials with insufficient permissions.
+		// Ref: https://developer.pagerduty.com/docs/errors
+		if strings.Contains(bodyStr, "Access Denied") && strings.Contains(bodyStr, "2010") {
```
martinlocklear, line 97: I wish there was unit test coverage for this.  Also, does it make sense to actually parse the JSON, rather than to just do a string contains?

martinlocklear (approved):
Approving to unblock, but I think we should consider adding unit tests for this case.

pkg/detectors/pagerdutyapikey/pagerdutyapikey.go
nabeelalam: Thanks for the review @martinlocklear, an additional unit test would be beneficial, yes.

Generally in detectors I've seen us rely on string contains, unless we need to extract something from the response. If our matching string is specific enough -- which I feel it is, in our case -- I lean towards string contains. IMO this would would also be slightly more adaptive to any potential change in the response shape (e.g. if keys get renamed but the message still contains these keywords)

shahzadhaider1:
Closing in favor of https://github.com/trufflesecurity/trufflehog/pull/5230

PR #4538, opened 2025-11-05
Fix Github Wiki Page Links

mustansir14:
### Description:
This PR fixes a bug with link generation for GH wiki pages. 

- Current (incorrect) Generated Link: https://github.com/hxnyk/hxnyk.wiki/blob/e5fdc764d6d405fc0e4e90e4bcf192357b1a1a87/Home.md#L5
- New (correct) Generated LInk: https://github.com/hxnyk/hxnyk/wiki/Home/e5fdc764d6d405fc0e4e90e4bcf192357b1a1a87#L5

### Checklist:


kashifkhan0771 (dismissed):

pkg/giturl/giturl.go
```diff
@@ -159,6 +159,19 @@ func GenerateLink(repo, commit, file string, line int64) string {
...
+			if file != "" {
+				baseLink += strings.TrimSuffix(file, ".md") + "/"
+			}
+			if commit != "" {
+				baseLink += commit
+			}
+			if line > 0 {
+				baseLink += "#L" + strconv.FormatInt(line, 10)
+			}
+
```
rosecodym: I know I'm being persnickety but this newline is bugging me :/

pkg/giturl/giturl.go
```diff
@@ -159,6 +159,19 @@ func GenerateLink(repo, commit, file string, line int64) string {
 					baseLink += "#L" + strconv.FormatInt(line, 10)
 				}
 			}
+		} else if strings.HasSuffix(repo, ".wiki.git") {
+			// GitHub Wiki links are formatted differently
+			baseLink = repo[:len(repo)-9] + "/wiki/"
+			if file != "" {
+				baseLink += strings.TrimSuffix(file, ".md") + "/"
```
rosecodym: This assumes that if `file` is empty, the actual file we care about is a markdown file, right? That seems very specific. How do we know that?

In any case, this seems like something that should be explained in a comment.

mustansir14: Thanks for this comment because my assumption was that wiki documents can only be created in markdown format, but it turns out github supports all of these:
image

Even though markdown is set by default and most wikis would be using markdown, best solution would be to just remove the file extension, whatever it is. I'll do that

pkg/giturl/giturl.go
```diff
@@ -159,6 +160,19 @@ func GenerateLink(repo, commit, file string, line int64) string {
...
+			if file != "" {
+				// remove file extension
+				baseLink += strings.TrimSuffix(file, filepath.Ext(file)) + "/"
+			}
+			if commit != "" {
+				baseLink += commit
+			}
+			if line > 0 {
+				baseLink += "#L" + strconv.FormatInt(line, 10)
+			}
```
rosecodym, line 175: If these are real possibilities, I think they should have test cases.

mustansir14: Added

rosecodym (approved):
I am taking it on faith that the base link mangling is correct, but you're the expert here!

PR #4545, opened 2025-11-07
Add Metrics to RetryableHTTPClient

mustansir14:
### Description:
This PR adds a new HTTP response body size metric to `InstrumentedTransport`, which is the current `Transport` for `SaneHTTPClient`. `InstrumentedTransport` is also set to be used as the `Transport` for `RetryableHTTPClient`.

#### Challenge with Response Body Size Calculation:

- The `Response` object has a `ContentLength` field that may have exactly what we're looking for, but it is not guaranteed to be provided by the server. Also it is never available for a chunked response.
- A simple approach would be to read the full body in memory and get it's size, but this is inefficient as responses can be large. Also it may add a delay to the `RoundTrip()` call, and would require restoring the body so that downstream code can read it again.

#### Solution:

To address this, a `responseSizeCounterReadCloser` is introduced to measure the size of the response body WHILE the downstream code reads it. This ensures no overhead is added in the `RoundTrip()` call. This however will not work if the downstream code never reads the body (e.g. detectors where only the `StatusCode` is used to verify), so to handle this, we add a check in the `Close()` method of the `ReadCloser` to drain the body if it's not consumed.

### Checklist:


pkg/common/http.go
```diff
@@ -135,6 +135,15 @@ func (t *InstrumentedTransport) RoundTrip(req *http.Request) (*http.Response, er
...
 		recordHTTPResponse(sanitizedURL, resp.StatusCode, duration.Seconds())
+
+		// wrap the response body to count bytes read
+		resp.Body = &responseSizeCounterReadCloser{
+			ReadCloser: resp.Body,
+			recordSizeFunc: func(size int) {
+				// Record response body size
+				recordResponseBodySize(sanitizedURL, size)
+			},
+		}
```
amanfcp: This would also start generating metrics for SaneHTTPClient. Just wanna make sure if this is intentional.

mustansir14: Yes, this is intended. We also want to add this metric to `SaneHTTPClient`

pkg/common/http.go
```diff
@@ -147,6 +156,35 @@ func NewInstrumentedTransport(T http.RoundTripper) *InstrumentedTransport {
...
+	}
+	*r.bytesRead += n
+	return n, err
+}
+
+func (r *responseSizeCounterReadCloser) Close() error {
+	if r.recordSizeFunc != nil {
+		if r.bytesRead == nil {
+			// this means downstream code never consumed the body, we must drain the body to get its size
+			_, err := io.Copy(io.Discard, r)
```
rosecodym: Does this cause `Read` to be called? If so, can you put that in a comment?

pkg/common/http.go
```diff
@@ -147,6 +156,35 @@ func NewInstrumentedTransport(T http.RoundTripper) *InstrumentedTransport {
...
+	if r.bytesRead == nil {
+		r.bytesRead = new(int)
+	}
+	*r.bytesRead += n
+	return n, err
+}
+
+func (r *responseSizeCounterReadCloser) Close() error {
+	if r.recordSizeFunc != nil {
+		if r.bytesRead == nil {
```
rosecodym: If a previous call to `Read` didn't read the entire payload, then the leftover size won't be counted, right? E.g.
```
buf := make([]byte, 10)
reader.Read(buf) // bytesRead is now 10
reader.Close() // bytesRead is not updated
```
This is kind of a niche case, so I'm not sure how much of a problem it is, but if I'm correct, this should at least be called out in a comment somewhere.

pkg/common/http_test.go
```diff
@@ -405,6 +406,76 @@ func TestSaneHttpClientMetrics(t *testing.T) {
...
+	defer server.Close()
+
+	// Create a RetryableHttpClient
+	client := RetryableHTTPClient()
+
+	testCases := []struct {
+		name               string
+		path               string
+		expectedStatusCode int
+		expectsNon200      bool
```
rosecodym: What is this field for? It looks unused.

pkg/common/http_test.go
```diff
@@ -405,6 +406,76 @@ func TestSaneHttpClientMetrics(t *testing.T) {
...
+			var requestURL string
+			if strings.HasPrefix(tc.path, "http") {
+				requestURL = tc.path
+			} else {
+				requestURL = server.URL + tc.path
+			}
+
+			// Get initial metric values
+			sanitizedURL := sanitizeURL(requestURL)
+			initialRequestsTotal := testutil.ToFloat64(httpRequestsTotal.WithLabelValues(sanitizedURL))
```
rosecodym, line 459: Whoa I didn't know we could test Prometheus metrics like this!

rosecodym (commented):
This is pretty interesting, but I did leave an inline question.

Also, please add doc comments for your new exported functions. Thanks!

pkg/common/http.go
mustansir14: Hmm, that's an interesting case. You're right, a partial read will cause us to not have a correct size. But do you think this is a problem? I can't think of any case where a detector or any piece of code would partially read a response.

I can surely call this out in a comment though.

pkg/common/http_test.go
mustansir14: Yep, good catch. I basically copy-pasted the test for `SaneHTTPClient` so this got copied over as well. Will remove it

camgunz (changes requested):
I really like how this is super accurate, which wasn't easy to do! But instead of counting what we read (and potentially manually reading), I think we should start out by just using `ContentLength` from [the Response object](https://pkg.go.dev/net/http#Response). It's definitely less reliable, but I don't want to get out the big hammer until we see how much less reliable it really is.

mustansir14:
> I really like how this is super accurate, which wasn't easy to do! But instead of counting what we read (and potentially manually reading), I think we should start out by just using `ContentLength` from [the Response object](https://pkg.go.dev/net/http#Response). It's definitely less reliable, but I don't want to get out the big hammer until we see how much less reliable it really is.

I really liked that code :face_holding_back_tears:. But I guess you're right, it is overkill for just metrics. I'll scrap it

pkg/common/http_metrics.go
```diff
@@ -92,14 +103,19 @@ func recordHTTPRequest(sanitizedURL string) {
...
 
 	// Record non-200 responses
 	if statusCode != 200 {
 		httpNon200ResponsesTotal.WithLabelValues(sanitizedURL, strconv.Itoa(statusCode)).Inc()
 	}
+
+	// Record response body size if known
+	if contentLength >= 0 {
+		httpResponseBodySizeBytes.WithLabelValues(sanitizedURL).Observe(float64(contentLength))
+	}
```
shahzadhaider1, line 118: this if will [always be true](https://httpwg.org/specs/rfc9110.html#field.content-length)?

shahzadhaider1 (approved):

pkg/common/http_metrics.go
nabeelalam: I assume this check to ensure that the `contentLength` value being passed into the function is valid (i.e. >0)

nabeelalam (approved):
Thanks @mustansir14

pkg/common/http_metrics.go
mustansir14: Here's what `net/http` [documentation](https://pkg.go.dev/net/http#:~:text=treated%20as%20unknown.-,ContentLength%20int64,-//%20TransferEncoding%20lists%20the) says about the `ContentLength` field:

> 	// ContentLength records the length of the associated content. The
	// value -1 indicates that the length is unknown. Unless Request.Method
	// is "HEAD", values >= 0 indicate that the given number of bytes may
	// be read from Body.
	ContentLength int64
	
So there is a possibility of the value being `-1`

camgunz (approved):

PR #4546, opened 2025-11-08
Added tests for new .msg and .doc file types to skipArchiverMimeTypes

jordanTunstill:
### Description:
In previous pull request https://github.com/trufflesecurity/trufflehog/pull/4544, I added the .msg and .doc file types to skipArchiverMimeTypes. However, that PR was merged before I had proper testing in place. I have created test files of both .msg and .doc, and copied the functionality of TestHandleFileAR for both the new tests I added. They both pass, and should confirm that the file type is processed appropriately.

### Checklist:


pkg/handlers/handlers_test.go
```diff
@@ -294,6 +294,30 @@ func TestHandleFileAR(t *testing.T) {
 	assert.Equal(t, wantChunkCount, len(reporter.Ch))
 }
 
+func TestHandleFileMSG(t *testing.T) {
+	wantChunkCount := 5
+	reporter := sources.ChanReporter{Ch: make(chan *sources.Chunk, wantChunkCount)}
+
+	file, err := os.Open("testdata/test.msg")
+	assert.Nil(t, err)
```
rosecodym: I typically prefer `require` for setup steps like this, because if they fail, there's no point in running the rest of the test. (You might want to switch this to `NoError` as well.)

pkg/handlers/handlers_test.go
```diff
@@ -294,6 +294,30 @@ func TestHandleFileAR(t *testing.T) {
...
+
+	assert.Equal(t, 0, len(reporter.Ch))
```
rosecodym: Would `assert.Empty` work here?

pkg/handlers/handlers_test.go
```diff
@@ -294,6 +294,30 @@ func TestHandleFileAR(t *testing.T) {
...
+	assert.NoError(t, HandleFile(context.Background(), file, &sources.Chunk{}, reporter))
+	assert.Equal(t, wantChunkCount, len(reporter.Ch))
```
rosecodym, line 307: This is a fascinating way to do this kind of test. The way we do it elsewhere is to pull items off the channel until it's closed, counting them as we go, and then check whether the count is correct. It never occurred to me to just make a buffered channel and see if it's got the right number of things in it.

This won't work if the writer tries to write more chunks than the channel size (because those writes will block), but plenty of our other tests are equally brittle, so I wouldn't feel fair holding this up. What you've done makes for a very _simple_ test, at least.

rosecodym (approved):
I have some thoughts about minor improvements to your assertions (which I left on test but apply to both), but this seems overall good to me!

pkg/handlers/handlers_test.go
jordanTunstill: Thanks Cody! I can't take credit for the logic, I used an existing test in the file as a blueprint: TestHandleFileAR

pkg/handlers/handlers_test.go
jordanTunstill: Switched to this for both tests, thank you!

pkg/handlers/handlers_test.go
jordanTunstill: Thanks Cody, used this for both tests

PR #4547, opened 2025-11-11
Added Additional Metrics and Job ID in the Docker Source

nabeelalam:
### Description:
This PR adds three new Docker source metrics for improved visibility into image enumeration and scanning progress.
It also updates all Docker metrics to include the "Job ID" label alongside "Source Name" for better observability of the statistics.

### New Metrics:
- dockerImagesNumerated: Total number of images discovered for a namespace.
- dockerLayersEnumerated: Total number of layers discovered per image.
- dockerHistoryEntriesEnumerated: Total number of image history entries enumerated.

### Checklist:


pkg/sources/docker/metrics.go
```diff
@@ -14,23 +14,48 @@ var (
...
-		[]string{"source_name"})
+		[]string{"source_name", "job_id"})
+
+	dockerImagesNumerated = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: common.MetricsNamespace,
+		Subsystem: common.MetricsSubsystem,
+		Name:      "docker_images_numerated",
+		Help:      "Total number of Docker images numerated.",
+	},
+		[]string{"source_name", "job_id"})
```
mustansir14: A typo? `numerated` -> `enumerated`

mustansir14 (approved):
LGTM. Thanks!

pkg/sources/docker/metrics.go
nabeelalam: 100%. Wanted to type enumerated! Thanks @mustansir14

pkg/sources/docker/metrics.go
```diff
@@ -14,23 +14,48 @@ var (
 		Name:      "docker_layers_scanned",
 		Help:      "Total number of Docker layers scanned.",
 	},
-		[]string{"source_name"})
+		[]string{"source_name", "job_id"})
```
rosecodym, line 17: Why did you add `job_id` for all of these? I know that we use it as a metric label in many other places, but that kind of contravenes best Prometheus practices, so I'd like to be very deliberate about extending the pattern.

nabeelalam: Hey @rosecodym! Thanks for reviewing this and calling that out.

The original requirement I was given was “add the source’s Job ID to all metrics calls” so I pushed `job_id` onto the shared label set for all Docker metrics, the reason I believe is so that metrics from different scan jobs can be distinguished instead of being aggregated into the same series.

Though, hmm... thinking about it from a Prometheus perspective I do agree we probably don’t need `job_id` on every one of these series, especially the histogram.

I can perhaps scope this down so that `job_id` is only applied to the progress counters/gauges and leave `docker_list_images_api_duration_seconds` with just the `source_name`?

I would like to know if that sound reasonable to you, or do you have any other suggestion? Let me know if I have completely misunderstood your comment 😅.

camgunz: > Why did you add job_id for all of these?

This was my idea, I wanted to have a way of noticing big changes in these numbers between jobs. But, I'm a Prometheus dummy so if this is a bad idea or there's better ways to do it, I'm into it. I think there's one metric we don't need it for though; I'll comment there.

pkg/sources/docker/metrics.go
```diff
@@ -40,5 +65,5 @@ var (
 			Help:      "Duration of Docker list images API calls.",
 			Buckets:   prometheus.DefBuckets,
 		},
-		[]string{"source_name"})
+		[]string{"source_name", "job_id"})
```
camgunz: This is the one I don't think we need job_id for

camgunz (changes requested):
Let's remove `job_id` from `"docker_list_images_api_duration_seconds"`, then I think we're good to go.

pkg/sources/docker/docker.go
```diff
@@ -115,6 +121,7 @@ type layerInfo struct {
 // Chunks emits data over a channel that is decoded and scanned for secrets.
 func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
 	ctx = context.WithValues(ctx, "source_type", s.Type(), "source_name", s.name)
+	jobIDStr := fmt.Sprint(s.jobId)
```
camgunz, line 124: Wow also TIL I learned about `fmt.Sprint`; verrrrry handy

pkg/sources/docker/metrics.go
nabeelalam: Ah missed this, yes this is what I had in mind as well, wouldn't need this on the histogram

camgunz (approved):

PR #4552, opened 2025-11-14
Added API Call Metrics to Docker Source

nabeelalam:
### Description:
This PR wires the Docker source’s registry HTTP traffic into TruffleHog’s shared HTTP metrics instrumentation.

By wrapping the `remote.WithTransport(...)` call with `common.NewCustomTransport` and `common.NewInstrumentedTransport`, Docker API calls should now emit the associated API metrics.

This should add the following metrics to Docker API calls:
- `trufflehog_http_client_requests_total`
- `trufflehog_http_client_request_duration_seconds`
- `trufflehog_http_client_non_200_responses_total`

Once [this PR to add new metrics to the Retryable HTTP Client](https://github.com/trufflesecurity/trufflehog/pull/) is merged, the following metric will also be recorded:
- `trufflehog_http_client_response_body_size_bytes`

### Checklist:


shahzadhaider1 (approved):

pkg/sources/docker/docker.go
```diff
@@ -502,7 +502,8 @@ func (s *Source) remoteOpts() ([]remote.Option, error) {
 	}
 
 	var opts []remote.Option
-	opts = append(opts, remote.WithTransport(defaultTransport))
+	instrumentedTransport := common.NewInstrumentedTransport(common.NewCustomTransport(defaultTransport))
+	opts = append(opts, remote.WithTransport(instrumentedTransport))
```
mustansir14: Why not simply do `remote.WithTransport(common.NewInstrumentedTransport(defaultTransport))`?

camgunz: I too am a fan of not naming things when they're only used once. But, I can also see how that'd be pretty nested, so, maybe author's choice on this one.

camgunz (approved):

pkg/sources/docker/docker.go
nabeelalam: @mustansir14 hmm. Based off of looking at `CustomTransport` in `common/http.go` I assumed it was responsible for adding the a `User-Agent` header and it would serve some sort of meaningful purpose.

mustansir14: @camgunz I didn't mind the variable. I was just concerned why we added `common.NewCustomTransport`. But upon checking, @nabeelalam is right that it is responsible for adding the User-Agent, which is cool.

nabeelalam: I looked at it and the nesting doesn't seem as bad to me as I had thought it would be, so I'll just remove the extra assignment 😂

camgunz (approved):

PR #4555, opened 2025-11-17
[INS-131] Add response id and response name to Postman metadata

mustansir14:
### Description:
The PR adds `response_id` and `response_name` to the Postman source's metadata protos and captures these fields from the response item.

### Checklist:


pkg/sources/postman/postman.go
```diff
@@ -682,6 +682,8 @@ func (s *Source) scanHTTPResponse(ctx context.Context, chunksChan chan *sources.
 	if response.Uid != "" {
 		m.Link = LINK_BASE_URL + "example/" + response.Uid
 		m.FullID = response.Uid
+		m.ResponseID = response.Uid
```
shahzadhaider1, line 685: `response.Uid` is already being stored in `m.FullID` field. Could you help me understand the reason behind storing the same value in another field, i.e., `m.ResponseID`?

mustansir14: This is the pattern that has been used all over this source. For example, see for [request ID](https://github.com/trufflesecurity/trufflehog/blob/df31a0e53a1efa49f2396699d51254455bb81868/pkg/sources/postman/postman.go#L426).

`FulID` is supposed to contain the ID for the item (whatever it may be, request, response, folder etc) while the specific fields like `RequestID` and `ResponseID` are supposed to contain IDs of these specific items.

nabeelalam: To add to @mustansir14's comment, `FullID` isn't present in the proto definition of `Postman` either, and is being used internally to form the full `Link` to the postman item so the response ID and name doesn't actually make its way outside of the code.

shahzadhaider1 (approved):

kashifkhan0771 (approved):

amanfcp (approved):

camgunz (approved):

PR #4557, opened 2025-11-18
Pagination and Rate-Limit Handling In Docker Registry Namespace API Calls

nabeelalam:
### Description:
This PR adds pagination and rate-limit prevention for Dockerhub, Quay, and GHCR namespace discovery. This will ensure we fetch all images under a namespace while avoiding `429 Too Many Requests` responses without missing results that span across pages.

#### Updates:
- Implemented pagination for DockerHub (using `next` key), Quay (`next_page`), and GHCR (`Link` header).

- Added `golang.org/x/time/rate` to prevent  `429` responses by limiting a single request per 1.5 seconds.

### Checklist:


pkg/sources/docker/registries.go
```diff
@@ -45,9 +46,72 @@ func MakeRegistryFromNamespace(namespace string) Registry {
...
+		// Respect context cancellation while sleeping.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(sleepFor):
+		}
+	}
+
+	return nil, fmt.Errorf("rate limited (HTTP status 429) after %d attempts", maxRetriesOnRateLimit)
+}
```
mustansir14: The logic looks solid here. Just wondering, have you considered exponential backoff?

pkg/sources/docker/registries.go
```diff
@@ -68,53 +132,64 @@ func (d *DockerHub) WithRegistryToken(registryToken string) {
...
+		}
 
-	defer func() {
-		_, _ = io.Copy(io.Discard, resp.Body)
-		_ = resp.Body.Close()
-	}()
+		defer func() {
+			_, _ = io.Copy(io.Discard, resp.Body)
+			_ = resp.Body.Close()
+		}()
```
mustansir14: This might be problematic as we are deferring this inside of a loop, so `resp.Body.Close()` is not called after every iteration, it gets called at the end when this function returns, for all pages at once.

You probably need to wrap the code inside the loop in a separate function.

pkg/sources/docker/registries.go
```diff
@@ -140,59 +217,80 @@ func (q *Quay) WithRegistryToken(registryToken string) {
...
+	baseURL := &url.URL{
 		Scheme: "https",
 		Host:   "quay.io",
 		Path:   path.Join("api", "v1", "repository"),
 	}
 
-	query := url.Query()
-	query.Set("namespace", quayNamespace)
-	query.Set("public", "true")
-	query.Set("private", "true")
```
mustansir14, line 152: It seems we have removed this `private` param.

pkg/sources/docker/registries.go
```diff
@@ -140,59 +217,80 @@ func (q *Quay) WithRegistryToken(registryToken string) {
...
+		}
 
-	responseBodyByte, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return nil, err
-	}
+		defer func() {
+			_, _ = io.Copy(io.Discard, resp.Body)
+			_ = resp.Body.Close()
+		}()
```
mustansir14: Same issue here, deferring in a loop.

pkg/sources/docker/registries.go
```diff
@@ -140,59 +217,80 @@ func (q *Quay) WithRegistryToken(registryToken string) {
...
+		var raw map[string]any
+		if err := json.Unmarshal(body, &raw); err != nil {
+			return nil, err
+		}
+		if np, ok := raw["next_page"].(string); ok && np != "" {
+			nextPageToken = np
+		} else {
+			// No token – stop to avoid an infinite loop.
+			break
+		}
```
mustansir14: I don't understand the point of this. Why can't we simply add another field `nextPage` in `quayResp`? If the API doesn't return it it will simply be empty.

pkg/sources/docker/registries.go
```diff
@@ -210,60 +308,106 @@ func (g *GHCR) WithRegistryToken(registryToken string) {
 	g.Token = registryToken
 }
 
-// ListImages lists all images under a Quay namespace.
+// parseNextLinkURL extracts the URL with rel="next" from a GitHub Link header, if present.
+func parseNextLinkURL(linkHeader string) string {
```
mustansir14, line 295: As a comment, can we have an example of how the full header looks like and what we are parsing from it?

pkg/sources/docker/registries.go
```diff
@@ -210,60 +308,106 @@ func (g *GHCR) WithRegistryToken(registryToken string) {
...
+		}
 
-	defer func() {
-		_, _ = io.Copy(io.Discard, resp.Body)
-		_ = resp.Body.Close()
-	}()
+		defer func() {
+			_, _ = io.Copy(io.Discard, resp.Body)
+			_ = resp.Body.Close()
+		}()
```
mustansir14: Same issue.

pkg/sources/docker/registries.go
```diff
@@ -45,9 +46,72 @@ func MakeRegistryFromNamespace(namespace string) Registry {
 	return registry
 }
 
-// === DockerHub registry ===
+const (
+	maxRetriesOnRateLimit = 5
+	maxSleepTime          = 30 * time.Second
+)
+
+func requestWithRateLimit(ctx context.Context, method, urlStr string, headers http.Header) (*http.Response, error) {
```
kashifkhan0771: The `ratelimit` handling should be pretty much straight forward and generic as we have to handle multiple registries. Have we considered using `golang.org/x/time/rate`? We can implement the logic within existing ListImage methods.

kashifkhan0771 (changes requested):
My suggestion is to keep most of the existing code as-is and simply introduce a fixed page size (e.g., 30) for all registry APIs. Then we can wrap the current logic for all registries in a loop that continues until `next_page` is empty, and use `golang.org/x/time/rate` to generically control the number of requests per second.

Even though it’s unlikely we’ll hit rate limits, since how many namespaces will have hundreds of images and this gives us a consistent safeguard so we don’t accidentally overload their APIs in any scenario.

pkg/sources/docker/registries.go
nabeelalam: Yes, I've updated this

pkg/sources/docker/registries.go
nabeelalam: Good catch, there's a couple examples of these I need to fix. Thanks

pkg/sources/docker/registries.go
nabeelalam: Thanks. So from what I understand you're suggesting that each `ListImages` function should remain unchanged other than an explicit page size query parameter, and to instead call the function in a loop in `GetNamespaceImages(...)` in `docker.go` until there are additional pages?

pkg/sources/docker/registries.go
nabeelalam: Added. Thanks

pkg/sources/docker/registries.go
nabeelalam: Yes, this was unneeded, I've updated this

pkg/sources/docker/registries.go
kashifkhan0771: > instead call the function in a loop in GetNamespaceImages(...) in docker.go

Not calling the function in a loop but rather have loop inside the function, how we handle pagination for enumerating GITHUB/GITLAB repositories.

kashifkhan0771: Pseudo Code:
1. Call registry.ListImages().
2. Inside ListImages:
    - Initialize an infinite loop.
    - Fetch images of a given namespace in batches of 30 (30 images per page).
    - Continue fetching until no images remain.
    - Respect API rate limits (e.g., 5 requests per 7–8 seconds) using a token bucket from ratelimit.
    - Break the loop when all images are retrieved.

pkg/sources/docker/registries.go
```diff
@@ -22,10 +23,10 @@ type Image struct {
 
 // Registry is an interface for any Docker/OCI registry implementation that can list all images under a given namespace.
 type Registry interface {
-	Name() string                           // return name of the registry
-	WithRegistryToken(registryToken string) // set token for registry
-	// TODO: Handle pagination and rate limits for list images API Call
+	Name() string                                                       // return name of the registry
+	WithRegistryToken(registryToken string)                             // set token for registry
 	ListImages(ctx context.Context, namespace string) ([]string, error) // list all images
+	WithClient() *http.Client                                           // return the HTTP client to use
```
mustansir14: Wouldn't it make more sense if this method took the `*http.Client` as an argument? Similar to `WithRegistryToken` above.

pkg/sources/docker/registries_test.go
```diff
@@ -60,3 +61,43 @@ func TestGHCRListImages(t *testing.T) {
...
 }
+
+func TestDockerHubListImages_RateLimitError(t *testing.T) {
+	t.Parallel()
+
+	// Dockerhub registry
+	dockerhub := MakeRegistryFromNamespace("trufflesecurity") // no authentication
+
+	// Cast dockerhub to *DockerHub registry to override the HTTP client
+	dockerhub.(*DockerHub).Client = common.ConstantResponseHttpClient(http.StatusTooManyRequests, "{}")
```
mustansir14: You won't need to do this type casting if the `WithClient` method took the client as an argument. You can directly call it and override the client.

pkg/sources/docker/registries.go
nabeelalam: Thanks, I'm looking at the Github/Gitlab code and going to update this to make it more consistent with the existing pagination logic.

pkg/sources/docker/registries_test.go
nabeelalam: Thanks, yeah that makes sense and I'll factor that in since I'm updating it based on Kashif's suggestions above

pkg/sources/docker/registries.go
nabeelalam: Added :+1:

pkg/sources/docker/registries_test.go
nabeelalam: Added :+1:

pkg/sources/docker/registries.go
nabeelalam: @kashifkhan0771 I've made some updates, please check those out

camgunz (changes requested):
Thanks for picking this up! I think we should use the Retryable HTTP client from `common`: we should get the retry stuff for free, and we've been adding super useful metrics to it too.

pkg/sources/docker/registries.go
```diff
@@ -10,10 +10,19 @@ import (
...
 )
 
 // defaultHTTPClient defines a shared HTTP client with timeout for all registry requests.
-var defaultHTTPClient = &http.Client{Timeout: 10 * time.Second}
+var defaultHTTPClient = common.RetryableHTTPClientTimeout(10)
+
+// registryRateLimiter limits how quickly we make registry API calls across all registries.
+// We allow roughly 5 requests every ~7.5 seconds (one token every 1.5s) as a simple
+// safeguard against overloading upstream APIs.
+var registryRateLimiter = rate.NewLimiter(rate.Every(1500*time.Millisecond), 1)
```
kashifkhan0771:
```
// 1 event every 1.5s, burst of 2
var registryRateLimiter = rate.NewLimiter(rate.Limit(2.0/3.0), 2)
```

pkg/sources/docker/registries.go
```diff
@@ -66,67 +76,101 @@ func (d *DockerHub) WithRegistryToken(registryToken string) {
...
 		Host:   "hub.docker.com",
 		Path:   path.Join("v2", "namespaces", namespace, "repositories"),
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url.String(), http.NoBody)
-	if err != nil {
-		return nil, err
-	}
+	query := baseURL.Query()
+	query.Set("page_size", "100") // fetch images in batches of 100 per page
```
kashifkhan0771: Let's make the page size a constant and I would suggest to keep it maybe 30-40 per page but I'll leave that up to you.

kashifkhan0771 (approved):
Good Work! Overall looks fine to me. Just make sure to test it properly locally with popular organization images.

pkg/sources/docker/registries.go
nabeelalam: I'm inclined to keep it maximum, while it's unlikely to have that many images in a page, it still shouldn't hurt just in case

shahzadhaider1 (approved):

camgunz (approved):

PR #4558, opened 2025-11-18
Move verify flag into `detectableChunk`

rosecodym:
### Description:
`Chunk.Verify` is an odd field - it originally conveys whether a source is going to run with verification, but then, at a certain point in the scanning pipeline, is mutated such that it instead indicates whether the chunk should be scanned with verification - which is _not_ solely dependent on the source's `verify` flag. This is unnecessarily difficult to understand and maintain. This PR separates those two pieces of information into two flags:
* `Chunk.Verify` has been renamed to `Chunk.SourceVerify`
* It is no longer mutated; instead "should this chunk's secrets be verified?" is now captured by a new field on `detectableChunk`

### Checklist:


mcastorina (approved):
Love this! I always found it confusing the two uses of "verify".

Why does a source need this information anyway, though? Could we lift this up a level and centralize it in the `SourceManager` like we do with `JobID` here (+2 other locations in that file)?

https://github.com/trufflesecurity/trufflehog/blob/35a5bf210f44d70a149b39d4db254abc7580cf6d/pkg/sources/source_manager.go#L499-L501

Also, this would be a breaking change to enterprise sources, right?

shahzadhaider1 (approved):

alafiand (approved):

camgunz (approved):

camgunz:
Re: Miccah's comment, we can do a bridging change where we support both fields in enterprise, do the cutover, then remove the old field.

rosecodym:
> Love this! I always found it confusing the two uses of "verify".
> 
> Why does a source need this information anyway, though? Could we lift this up a level and centralize it in the `SourceManager` like we do with `JobID` here (+2 other locations in that file)?
> 
> https://github.com/trufflesecurity/trufflehog/blob/35a5bf210f44d70a149b39d4db254abc7580cf6d/pkg/sources/source_manager.go#L499-L501
> 
> Also, this would be a breaking change to enterprise sources, right?

Yeah, I have idle plans to change all of this even more aggressively. I want to take things _very slow_, though, and do it as a bunch of smaller changes.

It will break enterprise sources, but the required change is simple enough that I was planning to make it immediately after merging this as part of the dependency update.

camgunz (approved):

PR #4560, opened 2025-11-19
[INS-104] Support units in S3 source

mustansir14:
### Description:
This PR supersedes #4270. The original contributor is no longer available to sign the CLA, so I have rebased + carried forward their work.

The PR aims to make the S3 source support Unit Scans by implementing the `SourceUnitEnumChunker` interface. It creates a unit for each bucket. 

I've tried to reuse most of the logic, mainly by wrapping the single bucket scanning logic into a `scanBucket` method, and reusing that in `ChunkUnit`.

Tests are added for both `Enumerate` and `ChunkUnit` methods. 

Update: This now also implements sub-unit resumption!

### Checklist:


pkg/sources/s3/s3.go
```diff
@@ -334,53 +326,16 @@ func (s *Source) scanBuckets(
...
-		input := &s3.ListObjectsV2Input{Bucket: &bucket}
+		var startAfter *string
 		if bucket == pos.bucket && pos.startAfter != "" {
-			input.StartAfter = &pos.startAfter
+			startAfter = &pos.startAfter
 			ctx.Logger().V(3).Info(
 				"Resuming bucket scan",
 				"start_after", pos.startAfter,
 			)
 		}
```
shahzadhaider1, line 337: Will this show which bucket we are resuming the scan on? Because the bucket is added to context later in the `s.scanBucket`

mustansir14: Hmm, good catch. I'll add the bucket name to this log.

shahzadhaider1 (approved):
LGTM. 
Please make sure to test all edge cases with this change.

pkg/sources/s3/s3.go
```diff
@@ -681,3 +700,62 @@ func (s *Source) visitRoles(
...
+// Enumerate implements SourceUnitEnumerator interface. This implementation visits
+// each configured role and passes each s3 bucket as a source unit
+func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) error {
+	visitor := func(c context.Context, defaultRegionClient *s3.Client, roleArn string, buckets []string) error {
+		for _, bucket := range buckets {
+			if common.IsDone(ctx) {
+				return ctx.Err()
+			}
+
+			ctx.Logger().V(5).Info("Enumerating bucket", "bucket", bucket)
```
rosecodym: To me, "enumerating bucket" implies that the bucket itself is being ranged over, not that the bucket is being emitted from an enumeration of all buckets. I'd write this like "Found bucket" or something.

I'm not sure we even need a log message here - the bucket will show up in a list elsewhere, so logging it just duplicates that.

pkg/sources/s3/s3.go
```diff
@@ -681,3 +700,62 @@ func (s *Source) visitRoles(
...
+
+	s3unit, ok := unit.(S3SourceUnit)
+	if !ok {
+		return fmt.Errorf("expected *S3SourceUnit, got %T", unit)
+	}
+	bucket := s3unit.Bucket
+
+	defaultClient, err := s.newClient(ctx, defaultAWSRegion, s3unit.Role)
+	if err != nil {
+		return fmt.Errorf("could not create s3 client: %w", err)
```
rosecodym: I recommend adding the bucket and role to this error message so that if any end users see it they have a better chance of self-diagnosing the issue.

rosecodym (commented):
Does this actually implement sub-unit resumption? It doesn't look like it does. @mcastorina I realize I don't actually know _how_ to do that using the existing interfaces.

mustansir14:
> Does this actually implement sub-unit resumption? It doesn't look like it does.

It doesn't.

> @mcastorina I realize I don't actually know _how_ to do that using the existing interfaces.

That makes the two of us. However, I see that it is currently implemented for [filesystem](https://github.com/trufflesecurity/trufflehog/blob/939fabea556a4637a66ab79fb8fa5d0e8a436077/pkg/sources/filesystem/filesystem.go#L130), and I kind of get how it works. I'll try and implement it here

pkg/sources/s3/s3.go
mustansir14: You're right. I'll just remove the logs.

pkg/sources/s3/s3.go
mustansir14: Will do

mcastorina:
> @mcastorina I realize I don't actually know _how_ to do that using the existing interfaces.

I added some methods to the `Progress` struct to handle it in a thread safe way here:

https://github.com/trufflesecurity/trufflehog/blob/939fabea556a4637a66ab79fb8fa5d0e8a436077/pkg/sources/sources.go#L544-L610

The idea is you need to save some data for the specific source unit ID, and you can retrieve it by that ID. When the source unit is finished, it needs to clear it from being tracked as well.

* `SetEncodedResumeInfoFor(id, value string)` - sets the resumption data for the source unit ID
* `GetEncodedResumeInfoFor(id string) string` - gets the resumption value for the source unit ID
* `ClearEncodedResumeInfoFor(id string)` - removes the resumption value from being tracked

pkg/sources/s3/s3.go
```diff
@@ -681,3 +698,76 @@ func (s *Source) visitRoles(
...
+
+type S3SourceUnit struct {
+	Bucket string
+	Role   string
+}
+
+func (s S3SourceUnit) SourceUnitID() (string, sources.SourceUnitKind) {
+	// The ID is the bucket name, and the kind is "s3_bucket".
+	return s.Bucket, "s3_bucket"
+}
```
rosecodym: @mcastorina I forget - is it a problem if `SourceUnitID` can't be used to round-trip a unit? (In this case, we lose the `Role` field.)

pkg/sources/s3/s3.go
```diff
@@ -681,3 +698,76 @@ func (s *Source) visitRoles(
 func makeS3Link(bucket, region, key string) string {
 	return fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", bucket, region, key)
 }
+
+type S3SourceUnit struct {
```
rosecodym: You've defined this unit type, but you haven't modified the source to actually use it. The source type still embeds `CommonSourceUnitUnmarshaller`, so it will still unmarshal source units to `CommonSourceUnit` instead of your new type. You'll need to define custom unmarshalling logic. (The `git` source has an example of custom unmarshalling logic you can look at.)

Also, I recommend putting the unit struct and related code in a separate file, because we do that for several other sources, and I think it makes things more readable.

rosecodym (changes requested):
Source unit unmarshalling needs some tweaks (mentioned inline).

pkg/sources/s3/s3.go
mustansir14: Oh, thanks for pointing this out. I wasn't aware of this. I'll make the changes.

pkg/sources/s3/s3.go
mustansir14: I'll wait for @mcastorina's answer before making changes here, but here's what the description comment says for `SourceUnitID()`:

> // SourceUnitID uniquely identifies a source unit. It does not need to
	// be human readable or two-way, however, it should be canonical and
	// stable across runs.

The bucket name is a globally unique value, so with that aspect we should be good.

pkg/sources/s3/s3.go
```diff
@@ -681,3 +697,72 @@ func (s *Source) visitRoles(
...
+
+	defaultClient, err := s.newClient(ctx, defaultAWSRegion, s3unit.Role)
+	if err != nil {
+		return fmt.Errorf("could not create s3 client for bucket %s and role %s: %w", bucket, s3unit.Role, err)
+	}
+
+	s.checkpointer.SetIsUnitScan(true)
+
+	var startAfterPtr *string
+	startAfter := s.Progress.GetEncodedResumeInfoFor(bucket)
```
rosecodym: I'm concerned about the way this gets resume info using only a bucket name, but then scans that bucket also using the role stored in the unit. It seems like the information used to retrieve resumption information should be the same information that's used to scan using the retrieved information, but that's not how you've implemented this.

I can't think of any concrete, immediate problems this would cause, but that doesn't mean there aren't any - and my bigger concern is that this will impede maintainability. What do you think?

pkg/sources/s3/s3.go
rosecodym: Oh, good catch. I guess the round-trip-abillity happens in the source manager somewhere? (@mcastorina?)

mcastorina: We take the full unit object and JSON marshal it, so the fields need to be public. Idk if I documented that anywhere though, but that's why a source needs to implement unmarshalling but not marshalling.

pkg/sources/s3/s3.go
mustansir14: Thanks for raising this. I had an internal discussion with @amanfcp on this, and he raised a great point that there could be a case where two roles have access to the same bucket, but have different object-level access. I researched on this and turns out [this is true](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-policy-language-overview.html). So it seems like a good idea to store resume info for a particular bucket AND a particular role.

My only concern here is that Legacy scans resumption does not have this. Resume Info is only stored using bucket there, and this particular case also seems to be applicable there. 

I might be totally off here and there might be some place where we are already handling this particular case, so please correct me if I am wrong.

pkg/sources/s3/s3.go
mustansir14: All the fields are public, so we're good there. But based on our discussion in the thread below regarding having the role in resumption info, it seems like a good idea to have the role in the `SourceUnitID` as well. I'll add it.

mustansir14: I added it and realized it might not be best to add this yet as this also affects sub-unit resumption because the resumption info is supposed to be saved against the `SourceUnitID`, and our current `checkpointer` only works with buckets, not roles. I have reverted the changes and will wait for your responses to decide if we want to go with roles being part of resumption info or not.

pkg/sources/s3/s3.go
rosecodym: Yep, this looks like an oversight in the existing implementation. It's been around for a while, so I don't think we need to urgently fix it (and I wouldn't fix it in this PR), but please add a TODO somewhere flagging the problem. We can clean it up as a later step.

mustansir14: I've included the role in the `SourceUnitID` now, and resumption info is being tracked using the `SourceUnitID`, so this means resumption info is being stored against both `role` and `bucket`

pkg/sources/s3/s3.go
```diff
@@ -681,3 +699,72 @@ func (s *Source) visitRoles(
...
+	s.scanBucket(ctx, defaultClient, s3unit.Role, s3unit.Bucket, reporter, startAfterPtr)
+	return nil
+}
+
+func (s *Source) UnmarshalSourceUnit(data []byte) (sources.SourceUnit, error) {
+	var unit S3SourceUnit
+	if err := json.Unmarshal(data, &unit); err != nil {
+		return nil, err
+	}
+	bucket, kind := unit.SourceUnitID()
```
rosecodym: The variable called `bucket` actually holds the unit ID, right? (I don't think this will cause program incorrectness but it will likely confuse readers.)

rosecodym (approved):
Thanks for your attention to detail on this! Hopefully we got everything, but it looks like you were pretty diligent with your test cases, so I'm not too worried.

pkg/sources/s3/s3.go
mustansir14: Yep, good catch. I had it previously when we were using just the bucket for UnitID. Made the change

PR #4564, opened 2025-11-20
Addition of Installation Type field in Jira Source Proto [INS-79]

shahzadhaider1:
Add configuration option to override Jira installation type autodetection (Cloud vs Data Center).

## Problem
Currently, we autodetect whether a Jira instance is Cloud or Data Center by making a HEAD request to a cloud-only endpoint. If the endpoint returns 404, it's assumed to be Data Center; otherwise, it's Cloud.

This autodetection could fail if:
- Atlassian changes their API endpoints
- Network/firewall rules block the detection request
- Custom Jira configurations behave unexpectedly

## Solution
Added `installation_type` field to the Jira source configuration. 

### Checklist:


proto/sources.proto
```diff
@@ -338,6 +338,12 @@ message Huggingface {
   bool include_prs = 19;
 }
 
+enum JiraInstallationType {
+  JIRA_INSTALLATION_TYPE_AUTODETECT = 0;
+  JIRA_INSTALLATION_TYPE_CLOUD = 1;
+  JIRA_INSTALLATION_TYPE_DATA_CENTER = 2;
+}
```
mustansir14, line 345: It's a good practice ([mentioned in the protobuf guide](https://protobuf.dev/programming-guides/proto3/#enum:~:text=%7D-,Enum%20Default%20Value,-The%20default%20value)) to have the first value of the enum be `ENUM_TYPE_NAME_UNSPECIFIED` or `ENUM_TYPE_NAME_UNKNOWN`

shahzadhaider1: You're right. I was just following [convention](https://github.com/trufflesecurity/trufflehog/blob/939fabea556a4637a66ab79fb8fa5d0e8a436077/proto/sources.proto#L97). What do you think?

mustansir14: Hmm, it's a matter of what we've been doing and what's the right way of doing it....... I'm not sure about this.

Personally I agree with the guide. Having the default value of the enum be an `UNKNOWN` value makes sense because then your downstream application can know that the caller did not explicitly specify a value. In our case, we would probably map this `UNKNOWN` to `AUTODETECT`, but still there will be a way to differentiate if the user actually wanted `AUTODETECT` or if they just didn't specify a value.

rosecodym: Great callout! :) But in this particular case, we (well, I) consciously chose for `AUTODETECT` to be the "unknown" option. No user is ever going to explicitly configure `AUTODETECT` _because_ it's documented as the default, so there is a nice congruence between "default behavior" and "unspecified behavior."

And while maybe the choice I made was not the best one, I believe that it's important to make the _same_ choice here so we get the benefits of consistency. If either of you feel more strongly the other way though, let's have a broader conversation about it!

rosecodym (approved):

proto/sources.proto
shahzadhaider1: That makes sense; consistency is a solid point.
I'm fine with keeping it as AUTODETECT or following the best practice. 

I requested a review from @camgunz just for visibility / another perspective since you mentioned a broader discussion could make sense.

kashifkhan0771 (approved):

proto/sources.proto
shahzadhaider1: Going ahead with the initial implementation (as suggested by Cody as well).

PR #4565, opened 2025-11-20
Enable optional symlink following for fs source

dustin-decker:
### Description:
Adds an option to follow symlinks in the fs source.

### Checklist:


pkg/sources/filesystem/filesystem.go
```diff
@@ -94,13 +96,19 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 
 		cleanPath := filepath.Clean(path)
-		fileInfo, err := os.Lstat(cleanPath)
+		var fileInfo fs.FileInfo
+		var err error
+		if s.followSymlinks {
+			fileInfo, err = os.Stat(cleanPath)
+		} else {
+			fileInfo, err = os.Lstat(cleanPath)
+		}
```
mcastorina, line 146: **Suggestion:** I think this can be simplified via [filepath.EvalSymlinks](https://pkg.go.dev/path/filepath#EvalSymlinks):

```go
cleanPath := filepath.Clean(path)
if s.followSymlinks {
    var err error
    cleanPath, err = filepath.EvalSymlinks(path)
    if err != nil { /* ... */ }
}
```

pkg/sources/filesystem/filesystem.go
```diff
@@ -94,13 +96,19 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 		if err != nil {
 			logger.Error(err, "unable to get file info")
 			continue
 		}
 
-		if fileInfo.Mode()&os.ModeSymlink != 0 {
+		if !s.followSymlinks && fileInfo.Mode()&os.ModeSymlink != 0 {
```
mcastorina, line 152: `fileInfo.Mode()` would always be non-symlink if `os.Stat` is used, right? Since `Stat` follows symlinks.

dustin-decker: I have redone this preserving the original first Lstat and then doing a Stat after if followSymlinks is enabled and it is a symlink to make it more clear.

pkg/sources/filesystem/filesystem.go
mcastorina: Unresolving this because I'm not sure why it was dismissed.

pkg/sources/filesystem/filesystem.go
mcastorina: Unresolving because I don't see the original `Lstat` anywhere. Am I missing it?

pkg/sources/filesystem/filesystem.go
```diff
@@ -26,15 +28,32 @@ import (
...
+	// Only created when followSymlinks=true to avoid memory overhead.
+	//
+	// Why LRU cache instead of a map:
+	// - Bounded memory: Limits to 10k paths (~1MB) even for massive directory trees
+	// - Per-path reset: Cache is recreated for each scan path to prevent accumulation
+	// - Loop detection: Prevents scanning the same file multiple times via different symlinks
+	//
+	// Why depth-1 limiting:
+	// - Prevents infinite loops: Symlink chains (A->B->C->...) are limited
+	// - Predictable behavior: Users know exactly which symlinks will be followed
```
mcastorina, line 55: What does a cache get us that `Stat` / `EvalSymlinks` does not address? Looks like if there's a cycle, those functions return an error: `too many links`

pkg/sources/filesystem/filesystem.go
```diff
@@ -93,18 +113,70 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 		}
 
-		if fileInfo.Mode()&os.ModeSymlink != 0 {
+		if !s.followSymlinks && fileInfo.Mode()&os.ModeSymlink != 0 {
 			logger.Info("skipping, not a regular file", "path", cleanPath)
 			continue
 		}
 
+		// If followSymlinks is enabled and this is a symlink, check for loops
+		if s.followSymlinks && fileInfo.Mode()&os.ModeSymlink != 0 {
```
mcastorina, line 158: I believe this is *always* false since `fileInfo` comes from `Stat` when `followSymlinks` is true. Is there an example or test case that exercises this scenario?

pkg/sources/filesystem/filesystem.go
```diff
@@ -179,13 +307,31 @@ func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sour
...
+// This enforces depth-1 symlink following to prevent:
+// - Infinite symlink loops
+// - Deep directory traversal through symlinks
+//
+// Returns true only if the symlink's parent directory matches a scan root path.
+func (s *Source) isDirectChild(path string) bool {
+	dir := filepath.Clean(filepath.Dir(path))
+	_, isRoot := s.scanRootPaths[dir]
+	return isRoot
+}
```
mcastorina, line 320: When I read "depth-1 limiting" I thought it meant we allow only one symlink depth, but this looks like it is limiting symbolic link targets to a very specific directory structure.

Why are we doing this? It seems like something that can be easily tripped on and I'm not seeing why it needs to be that way.

pkg/sources/filesystem/filesystem_test.go
```diff
@@ -461,6 +471,356 @@ func TestSkipBinaries(t *testing.T) {
...
+		err = s.ChunkUnit(ctx, sources.CommonSourceUnit{
+			ID: tempDir,
+		}, &reporter)
+		require.NoError(t, err)
+
+		// Should have 2 chunks: real_file and symlink1 (which resolves to real_file)
+		// symlink2 also resolves to the same real_file, so loop detection prevents duplicate scanning
+		// This is correct behavior - we don't want to scan the same content multiple times
+		assert.Equal(t, 2, len(reporter.Chunks), "Expected two chunks from real file and first symlink")
+	})
```
mcastorina, line 696: I'm not sure I agree that this is testing "only follow first level symlink in chain"

It scans the entire directory vs just `symlink1`.

pkg/sources/filesystem/filesystem.go
mcastorina: Ah, the cycle we're worried about is a symlink to a directory and then recursively scanning that. Do you think we could get away with saving the visited directories only?

PR #4574, opened 2025-12-01
[INS-170] Unify JDBC URL parsing across detectors and analyzers

mustansir14:
### Description:
We need to unify JDBC URL parsing across the detector and analyzer. This PR supports that by publicizing the methods required by the analyzer, and also some changes to make the methods compatible for both detector and analyzer.

The changes here will make more sense when viewed in conjunction with the corresponding [integrations PR](https://github.com/trufflesecurity/integrations/pull/81).

### Checklist:


pkg/detectors/jdbc/postgres.go
```diff
@@ -77,16 +80,20 @@ func parsePostgres(ctx logContext.Context, subname string) (jdbc, error) {
...
-		"host":            u.Host,
-		"dbname":          dbName,
 		"connect_timeout": "5",
 	}
 
+	postgresJDBC := &PostgresJDBC{
+		Host:     u.Host,
+		Database: dbName,
+		Params:   params,
+	}
```
amanfcp, line 88: We can use a single `ConnectionInfo` struct across all handlers, and later clean it up from jdbc analyzer models.go.

pkg/detectors/jdbc/postgres.go
```diff
@@ -95,46 +102,51 @@ func parsePostgres(ctx logContext.Context, subname string) (jdbc, error) {
...
+		"password": password,
+		"host":     host,
+	}
+	if user != "" {
+		data["user"] = user
+	}
+	if h, p, found := strings.Cut(host, ":"); found {
+		data["host"] = h
+		data["port"] = p
 	}
```
amanfcp: As per the [coding guidelines](https://docs.google.com/document/d/1clyphtrspL_e0IhVPONtBpl65bjGMYLiBjkEu5UYmrE/edit?tab=t.0)
```suggestion
	if h, p, ok := strings.Cut(host, ":"); ok {
		data["host"] = h
		data["port"] = p
}
```

amanfcp (approved):
Looks good to me! Thanks @mustansir14 
Left a few suggestions

martinlocklear (approved):

pkg/detectors/jdbc/postgres.go
mustansir14: Yes, that is a great idea. I'll unify these as well.

PR #4576, opened 2025-12-02
Make Integrations (shared) codeowners of proto/

rosecodym:
### Description:
The integrations team should codeown all the protobuf stuff, because they own the integrations that consume the message definitions. (`config.proto` is a borderline case, but that's a borderline case for _every_ team, so this solution isn't worse than any other.)

### Checklist:


rosecodym:
This git test has gotten flaky as heck recently!

PR #4581, opened 2025-12-04
feat(source/gitlab): support glob patterns in repositories field

amanfcp:
### Description:
Add support for glob patterns in the repositories field, allowing flexible repository matching alongside explicit URLs. Refactors repository resolution logic into resolveRepositories() to eliminate duplication between Chunks() and Enumerate(), improving maintainability and consistency.

### Checklist:


pkg/sources/gitlab/gitlab.go
```diff
@@ -959,6 +1008,156 @@ func (s *Source) WithScanOptions(scanOptions *git.ScanOptions) {
...
+}
+
+// resolveRepositories determines which repositories to scan based on s.repos configuration.
+// It handles three cases:
+// 1. Glob patterns (e.g., "org/*"): enumerates all projects and filters using patterns
+// 2. Explicit URLs: normalizes and reports directly without enumeration
+// 3. Empty: enumerates all projects using includeRepos/ignoreRepos filters
+//
+// All discovered repositories are reported via reporter.UnitOk().
+func (s *Source) resolveRepositories(
```
mustansir14, line 1043: Looking at the implementation for `includeRepos`, it looks pretty straightforward where we directly pass in the `s.includeRepos` slice to `buildIgnorer`. Why can't we do the same with `repos`? Why do we need to handle all these cases separately? Is there any context I'm missing here?

amanfcp: Good question! Here's what I've understood since working on this:

includeRepos:
- Always a filter on enumerated projects
- Assumes you're scanning "everything with filters"
- Always calls listProjects() then filters results

repos:
- Can be explicit URLs OR glob patterns
- Supports 3 distinct modes:
  a. Explicit URLs only - Use directly, no enumeration needed
  b. Glob patterns - Enumerate projects and filter
  c. Mixed - Enumerate for globs, use explicit URLs directly

kashifkhan0771 (commented):
One recommendation is to keep the GitLab source behavior consistent with GitHub. Users should not be allowed to use `--repo` together with `--org`. The `--include-repo` flag should only be valid when used with `--org` or `--group-id` (GitLab only), as its description states: "Repositories to include in an org scan."

Ideally, this is the best approach because glob patterns should only apply to repositories within an organization or group. The user specifies an org or group and then defines which repositories matching the glob pattern should be scanned. This keeps GitLab consistent with the GitHub source and simplifies the implementation.

kashifkhan0771:
> One recommendation is to keep the GitLab source behavior consistent with GitHub. Users should not be allowed to use `--repo` together with `--org`. The `--include-repo` flag should only be valid when used with `--org` or `--group-id` (GitLab only), as its description states: "Repositories to include in an org scan."
> 
> Ideally, this is the best approach because glob patterns should only apply to repositories within an organization or group. The user specifies an org or group and then defines which repositories matching the glob pattern should be scanned. This keeps GitLab consistent with the GitHub source and simplifies the implementation.

After this the flow will be like:
- List all repositories inside org or group
- Run all through the glob patterns provided in `--include-repos` if any
- Make a new list of repositories which by pass the filters
- Pass those to scan.

amanfcp:
Moving this PR to draft while I rethink the approach further before deciding whether to continue or close it.

amanfcp:
Closing this since there is a deadlock between wanting to add this new feature and keeping all git sources similar in features.

PR #4579, opened 2025-12-03
[chore] Update log.AddSink docs to clarify confusing behavior

mcastorina:
### Description:
Documents possibly unintuitive behavior of `log.AddSink`

### Checklist:
* [ ] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

PR #4582, opened 2025-12-04
Create way to add key-value pairs to new sinks

rosecodym:
### Description:
Counterintuitively, a new sink added to an existing logger will not inherit the existing logger's key-value pairs. This has only caused a problem in practice with respect to a small number of known key-value pairs, so this PR creates a way to add specific key-value pairs to new sinks.

This will supersede #4579.

### Checklist:

### DIFF
--- a/pkg/log/log_test.go
+++ b/pkg/log/log_test.go
@@ -163,6 +163,29 @@ func TestAddSink(t *testing.T) {
 	assert.Contains(t, buf2.String(), "line 2")
 }
 
+func TestAddSink_WithKeyValuePairs(t *testing.T) {
+	// Arrange: Create a logger with a key-value pair
+	var buf1 bytes.Buffer
+	logger, cleanup := New("service-name", WithConsoleSink(&buf1))
+	t.Cleanup(func() { _ = cleanup })
+	logger = logger.WithValues("sink 1 key", "sink 1 value")
+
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -136,18 +136,34 @@ func AddSentry(l logr.Logger, opts sentry.ClientOptions, tags map[string]string)
...
+// sink specifically by passing them to this function.
+func AddSink(l logr.Logger, sink logConfig, keysAndValues ...any) (logr.Logger, func() error, error) {
 	if sink.err != nil {
 		return l, nil, sink.err
 	}
+
+	// New key-value pairs cannot be ergonomically added directly to cores. logr has code to do it, but that code is not
+	// exported. Rather than replicating it ourselves, we indirectly use it by creating a temporary logger for the new
+	// core, adding the key-value pairs to the temporary logger, and then extracting the temporary logger's modified
+	// core.
```
mcastorina, line 152: Is the "non-ergonomic way" you're referring to the [Core.With](https://pkg.go.dev/go.uber.org/zap/zapcore#Core) method?

trufflesteeeve:
Nevermind!

mcastorina (commented):
Looks good! Feels kind of kludgey but it works.

PR #4586, opened 2025-12-05
[INS-204] Abort Postman scan if monthly API request limit crosses 80%

mustansir14:
### Description:
This PR adds a failsafe in the Postman source to abort the scan if we've crossed 80% of our monthly API request limit.

### Checklist:

### DIFF
--- a/pkg/sources/postman/postman_client.go
+++ b/pkg/sources/postman/postman_client.go
@@ -345,6 +345,11 @@ func (c *Client) handleRateLimits(ctx trContext.Context, resp *http.Response) er
 		return nil
 	}
 
+	if rateLimitTotalMonth == 0 {
+		ctx.Logger().V(2).Info("RateLimit-Limit-Month is zero, cannot compute usage percentage")
+		return nil
+	}
+
 	// failsafe to abandon scan if we are over the threshold of monthly API requests used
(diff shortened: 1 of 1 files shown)


pkg/sources/postman/postman_client.go
```diff
@@ -315,6 +305,55 @@ func (c *Client) getPostmanResponseBodyBytes(ctx trContext.Context, urlString st
...
+	rateLimitTotalMonthValue := resp.Header.Get("RateLimit-Limit-Month")
+	if rateLimitTotalMonthValue == "" {
+		rateLimitTotalMonthValue = resp.Header.Get("X-RateLimit-Limit-Month")
+	}
+	if rateLimitTotalMonthValue == "" {
+		ctx.Logger().V(2).Info("RateLimit-Limit-Month header not found in response")
+		return nil
+	}
+	rateLimitTotalMonth, err := strconv.Atoi(rateLimitTotalMonthValue)
+	if err != nil {
```
shahzadhaider1, line 341: This is good but one question: can `rateLimitTotalMonth` ever be zero?
If yes, maybe add a zero check because I hate _divided by zero errors_.

mustansir14: Thanks for catching this.

It's unlikely because the fact that we were able to get this response means that we are allowed some number of requests per month, but it doesn't hurt to add a check so I'll add it.

shahzadhaider1 (approved):

PR #4590, opened 2025-12-08
[INS-37] Fixed the issue of ftp connection ignoring the detector's co…

### LINKED ISSUE #3714: FTP detector ignores context timeout
> 2024-12-02T12:08:21-05:00	error	trufflehog	a detector ignored the context timeout	{"detector_worker_id": "J5m7Y", "detector": {"type":"FTP"}, "timeout": 60, "link": "https://github.com/Samsung/ChromiumGStreamerBackend/blob/29c79f177ee954858d754c0faf6d237b4a8a7fbd/net/base/net_util_unittest.cc#L1"}}

MuneebUllahKhan222:
### 🛠 Summary of Fixes

- **Resolved timeout issue**:  
  The verification logic was unintentionally using the default timeout because the detector context's remaining time wasn’t being forwarded.  
  Updated the implementation to pass the **remaining context deadline** into `verifyFTP`, ensuring correct timeout behavior.

- **Fixed failing tests**:  
  Tests began failing due to the introduction of the unexported `primarySecret` field in `detectors.Result`.  
  Updated the test comparison to **ignore this unexported field**, restoring expected test behavior.


pkg/detectors/ftp/ftp.go
```diff
@@ -80,7 +80,13 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 			timeout := s.verificationTimeout
 			if timeout == 0 {
-				timeout = defaultVerificationTimeout
+				if dl, ok := ctx.Deadline(); ok {
+					// Here we assign the remaining time before our context expires
+					// This help us cater the timelimit set through --detector-timeout flag
+					timeout = time.Until(dl)
+				} else {
+					timeout = defaultVerificationTimeout
+				}
```
kashifkhan0771: It looks fine. This is what I suggested [here](https://github.com/trufflesecurity/trufflehog/issues/3714#issuecomment-2565522620), but as per @rgmz comment, ftp.Dial can still hang for a longer time. Could you test it with this change?

pkg/detectors/ftp/ftp.go
kashifkhan0771, line 131: Maybe `defer c.Quit()` would help?

MuneebUllahKhan222: Added.

kashifkhan0771:
`pkg/detectors/ftp/ftp.go:134:14: Error return value of c.Quit is not checked (errcheck)`

kashifkhan0771 (approved):

kashifkhan0771:
@trufflesecurity/product-eng please review

pkg/detectors/ftp/ftp.go
camgunz: (I'm trying out [conventional comments](https://conventionalcomments.org/), so this really truly is just a suggestion)

**suggestion:** There are 3 cases here:
* `s.verificationTimeout == 0`
* `dl, ok := ctx.Deadline(); ok`
* default/else (`defaultVerificationTimeout`)

I would probably write this as a 3-part `if` statement:
```go
var timeout time.Duration
if s.verificationTimeout > 0 {
  timeout = s.verificationTimeout
} else if dl, ok := ctx.Deadline(); ok {
  timeout = time.Until(dl)
} else {
  timeout = defaultVerificationTimeout
}
```

It is a line longer because of the initial declaration, but the benefits are:
* it's flat, so I have an intuition that these cases are all mutually exclusive
* they all start with `timeout = ` so a quick scan shows me we're doing an assignment based on the different cases
* the default is suuuuper clear; having `timeout := s.verificationTimeout` _looks_ like a default, but in truth it's only saving us a declaration, so then when I get down to the `} else {` actual default case, my brain does a little hiccup as I realize "oh, **this** is the default case. Another way you could write it is:
```go
timeout := defaultVerificationTimeout
if s.verificationTimeout > 0 {
  timeout = s.verificationTimeout
} else if dl, ok := ctx.Deadline(); ok {
  timeout = time.Until(dl)
}
```
This is less explicit than the 3-part example, but it's also the shortest example so far and the default's right up top. I... might even like this better? Dunno.

camgunz (approved):
LG!

pkg/detectors/ftp/ftp.go
MuneebUllahKhan222: Good Suggestion @camgunz . I really like the 3-part if else statement and have added it as it makes the code a lot more readable

camgunz (approved):

PR #4596, opened 2025-12-09
Add proto definitions for Google Drive Domain-Wide Delegation (DWD)

shahzadhaider1:
Adds proto definitions to support Domain-Wide Delegation in the Google Drive source.

### Changes:

- Added GoogleDriveDWD message with fields for service account authentication
- Added dwd credential option to GoogleDrive message

### Fields:

- service_account_json: Service account JSON key
- admin_email: Super Admin email for user enumeration
- include_users: Specific users to scan
- exclude_users: Users to skip when scanning all

### Checklist:


proto/credentials.proto
```diff
@@ -77,3 +77,10 @@ message SlackTokens {
   string bot_token = 2;
   string client_token = 3;
 }
+
+message GoogleDriveDWD {
+  string service_account_json = 1; 
+  string admin_email = 2;  
+  repeated string include_users = 3;  
+  repeated string exclude_users = 4;
```
amanfcp, line 85: These will hold optional arrays of targeted or excluded emails, right?

Also, this might be a bit early to ask since we haven’t implemented it yet, but it would be helpful to know whether this will also support distribution lists like devs@trufflesec.com

amanfcp (approved):
LFTM

proto/credentials.proto
shahzadhaider1: Yes, these will hold optional arrays of targeted or excluded emails. 
And no, groups are not users; they're just email distribution lists. They don't have a Google Drive.

rosecodym (approved):

PR #4601, opened 2025-12-11
[INS-206] Store Gitlab Project ID in secret location metadata

mustansir14:
### Description:
This PR adds the changes to include gitlab project details like project ID, project name and project owner name to the metadata of a chunk.

Achieving this wasn't straightforward as the chunks are generated by the `git` source and it calls an injected `SourceMetadataFunc` to create the metadata. The signature of this function obviously does not include the gitlab project ID.

The solution implemented here is to maintain a `repoToProjectCache` map that stores the project details for a repo. This cache is populated as we enumerate the repos and the callback function uses this cache to populate the project details.

The only concerning part of this solution is memory usage, so just to put in perspective how much this impacts that, I found a [public organization](https://gitlab.eclipse.org) with ~3000 projects and ran benchmarks for before and after making the changes. (I know this isn't anywhere close to the largest organizations we've come across, but this is the biggest one I could find that was public)

**Results before making changes:** 

Total memory usage: 43234992 bytes (43.23 MBs)

```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkMemoryUsage$ github.com/trufflesecurity/trufflehog/v3/pkg/sources/gitlab

2025-12-11T18:58:01+05:00	info-0	context	starting projects enumeration	{"list_options": {"pagination":"keyset","order_by":"id","sort":"asc"}, "all_available": false}
2025-12-11T18:59:01+05:00	info-0	context	Enumerated GitLab projects	{"count": 2930}
Memory used during benchmark:
Allocated (Alloc): 43234992 bytes
Total Allocated (TotalAlloc): 43234992 bytes
Memory obtained from system (Sys): 17629184 bytes
goos: linux
goarch: amd64
pkg: github.com/trufflesecurity/trufflehog/v3/pkg/sources/gitlab
cpu: 12th Gen Intel(R) Core(TM) i5-1235U
BenchmarkMemoryUsage-12    	       1	59642136351 ns/op	42916920 B/op	  246843 allocs/op
PASS
ok  	github.com/trufflesecurity/trufflehog/v3/pkg/sources/gitlab	60.146s
```

**Results after this change:**

Total memory usage: 44666216 bytes (44.66 MBs)

```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkMemoryUsage$ github.com/trufflesecurity/trufflehog/v3/pkg/sources/gitlab

goos: linux
goarch: amd64
pkg: github.com/trufflesecurity/trufflehog/v3/pkg/sources/gitlab
cpu: 12th Gen Intel(R) Core(TM) i5-1235U
=== RUN   BenchmarkMemoryUsage
BenchmarkMemoryUsage
2025-12-11T18:38:02+05:00       info-0  context starting projects enumeration   {"list_options": {"pagination":"keyset","order_by":"id","sort":"asc"}, "all_available": false}
2025-12-11T18:39:09+05:00       info-0  context Enumerated GitLab projects      {"count": 2930}
Memory used during benchmark:
Allocated (Alloc): 44666216 bytes
Total Allocated (TotalAlloc): 44666216 bytes
Memory obtained from system (Sys): 21823488 bytes
BenchmarkMemoryUsage-12                1        66827846825 ns/op       44350440 B/op     261384 allocs/op
PASS
ok      github.com/trufflesecurity/trufflehog/v3/pkg/sources/gitlab     67.355s
```

It's important to note that this doesn't really perform a full scan (I tried doing that first but it would take hours), so I tweaked the code to expose the callback function and directly called that after enumerating the repos. This is the code used for benchmarking:

```
func BenchmarkMemoryUsage(b *testing.B) {
	// Enable memory allocation reporting
	b.ReportAllocs()

	// Run a specific function and measure memory usage
	var mStart, mEnd runtime.MemStats

	// Get memory stats before execution
	runtime.ReadMemStats(&mStart)

	connection := &sourcespb.GitLab{
		Endpoint: "https://gitlab.eclipse.org",
	}

	s := Source{}

	conn, err := anypb.New(connection)
	if err != nil {
		b.Fatal(err)
	}

	err = s.Init(context.Background(), "benchmark", 0, 0, false, conn, 100)
	if err != nil {
		b.Errorf("Source.Init() error = %v", err)
		return
	}

	feature.UseSimplifiedGitlabEnumeration.Store(true)

	// Run the benchmark loop
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		reporter := &sourcestest.TestReporter{}
		err = s.Enumerate(context.Background(), reporter)
		if err != nil {
			b.Errorf("Source.Chunks() error = %v", err)
			return
		}

		for _, unit := range reporter.Units {
			repository, _ := unit.SourceUnitID()
			_ = s.cfg.SourceMetadataFunc("test file", "test email", "test commit", "test timestamp", repository, "test local path", 1)
		}
	}

	// Get memory stats after execution
	runtime.ReadMemStats(&mEnd)

	// Calculate the memory usage
	allocDiff := mEnd.Alloc - mStart.Alloc
	totalAllocDiff := mEnd.TotalAlloc - mStart.TotalAlloc
	sysDiff := mEnd.Sys - mStart.Sys

	// Log the memory stats
	fmt.Printf("Memory used during benchmark:\n")
	fmt.Printf("Allocated (Alloc): %d bytes\n", allocDiff)
	fmt.Printf("Total Allocated (TotalAlloc): %d bytes\n", totalAllocDiff)
	fmt.Printf("Memory obtained from system (Sys): %d bytes\n", sysDiff)
}
```

### Checklist:


pkg/sources/gitlab/gitlab.go
```diff
@@ -1126,5 +1210,22 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
...
+				ctx.Logger().Error(err, "could not convert project_id metadata to int", "project_id", gitUnit.Metadata["project_id"])
+			} else {
+				project.id = projectId
+			}
+			project.name = gitUnit.Metadata["project_name"]
+			project.owner = gitUnit.Metadata["project_owner"]
+			s.repoToProjCache.set(repoURL, project)
+		}
+
+	}
```
amanfcp: Would recommend early return flow here. It's kind of difficult to read 😅

pkg/sources/gitlab/gitlab.go
```diff
@@ -864,6 +919,35 @@ func (s *Source) getAllProjectReposInGroups(
...
+}
+
+func projectToGitUnit(proj *gitlab.Project) git.SourceUnit {
+	metadata := map[string]string{
+		"project_id":   fmt.Sprintf("%d", proj.ID),
+		"project_name": proj.NameWithNamespace,
+	}
+	if proj.Owner != nil {
+		metadata["project_owner"] = proj.Owner.Email
+	}
```
amanfcp: We're missing the username check here as handled in `gitlabProjectToCacheProject`

amanfcp (approved):
LGTM! A few suggestions

pkg/sources/gitlab/gitlab.go
mustansir14: Hmm, totally agree. I'll make the changes

pkg/sources/gitlab/gitlab.go
mustansir14: Great catch, thanks

mariduv (approved):

rosecodym (dismissed):
There is some scanning pipeline machinery that cannot yet support this, so this PR won't work as-is. I'm requesting changes until I (or maybe @mcastorina) has some time to go through it and figure out whether "won't work" actually means there will be _problems_, or it just won't work.

pkg/sources/gitlab/gitlab.go
```diff
@@ -1126,5 +1208,39 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
...
+
+	// query project and add to cache
+	apiClient, err := s.newClient()
+	if err != nil {
+		ctx.Logger().Error(err, "could not create api client")
+		return
+	}
+	// extract project path from repo URL
+	// https://gitlab.com/testermctestface/testy.git => testermctestface/testy
+	repoPath := strings.TrimSuffix(strings.Join(strings.Split(repoUrl, "/")[len(strings.Split(repoUrl, "/"))-2:], "/"), ".git")
```
rosecodym: Did you consider properly parsing this as a URL instead of doing a simple string split?

Relatedly, this looks like it will panic if the repo URL is sufficiently malformed. Even if string splitting is the best solution, we should avoid that.

pkg/sources/gitlab/gitlab.go
```diff
@@ -291,12 +346,22 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, tar
 			},
 		}
 
+		// Clear the repo to project cache when done.
+		defer s.repoToProjCache.clear()
```
rosecodym: While I respect your concern for cache size, I don't believe clearing this cache is necessary. `Source` structures don't survive past a single scan, so clearing them at the end of the scan doesn't reclaim any memory that wouldn't be soon reclaimed anyway, and introduces the possibility of race conditions in the case that the git-parsing code sends work to separate goroutines.

pkg/sources/gitlab/gitlab.go
```diff
@@ -77,6 +77,49 @@ type Source struct {
...
+	defer r.Unlock()
+
+	delete(r.cache, repo)
+}
+
+func (r *repoToProjectCache) clear() {
+	r.Lock()
+	defer r.Unlock()
+
+	clear(r.cache)
```
rosecodym: What do you think of moving this to a separate file (still within this package)?

pkg/sources/gitlab/gitlab.go
```diff
@@ -291,12 +346,22 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, tar
...
 		if err := s.listProjects(ctx, apiClient, ignoreRepo, reporter); err != nil {
 			return err
 		}
 
 	} else {
 		gitlabReposEnumerated.WithLabelValues(s.name).Set(float64(len(repos)))
+		// ensure project details for specified repos are cached
+		// this is required to populate metadata during chunking
+		for _, repo := range repos {
+			s.ensureProjectInCache(ctx, repo)
```
rosecodym, line 318: It took me a few passes back and forth to figure out why sometimes you cache projects using `ensureProjectInCache`, and sometimes you directly cache the the value returned from `gitlabProjectToCacheProject`. I eventually got there, but laying out the call tree a bit differently would definitely have helped me understand what was going on much more quickly. What I realized I was looking for was a single "entry point" into caching, e.g.:
```
func (s *Source) cacheProject(repoUrl string, project *gitlab.Project)
```
When a `gitlab.Project` is available, this could be used directly. When only a repo URL is available, it could be transformed into a `gitlab.Project` using a second helper function:
```
func (s *Source) getGitlabProject(repoURL string) *gitlab.Project
```
This way, every function call that caches something looks the "same," and the function that acquires new (necessary) information for caching has a clear name. (`ensure` is a suboptimal function verb, because it's vague.)

What do you think of this?

rosecodym (commented):
This looks pretty straightforward, but I have some notes and questions.

pkg/sources/gitlab/gitlab.go
mustansir14: Thanks for this! You're right, parsing it as a URL would make it much more reliable. I'll make the changes.

pkg/sources/gitlab/gitlab.go
mustansir14: `repoToProjCache` uses a mutex internally so there wouldn't be a possibility of race conditions, but yeah if a `Source` is always used for a single scan then freeing up the memory doesn't make sense. I'll remove that.

pkg/sources/gitlab/gitlab.go
mustansir14: `ensureProjectInCache` is a useful function because it does multiple things: 
1) It first checks if the project is in cache 
2) If not, it fetches the project from the API 
3) It then stores the project in cache

This functionality is required in a couple of places which is why I grouped it into a single function. Breaking it down would simply mean making two separate calls in the same order at two separate places. 

What I agree with is that we create the two methods `cacheProject` and `getGitlabProject`, and use it inside `ensureProjectInCache`. So when we're enumerating projects, we directly call `cacheProject`, but in the case of configured `repos` and `ChunkUnit`, we use `ensureProjectInCache`. What do you think?

mustansir14: I've gone ahead and made these changes. Let me know if you think otherwise and I'll revert.

rosecodym (approved):
I'm bummed out that the git processing code requires this sort of contortion, but it looks like you've done it pretty cleanly. Nice work!

PR #4603, opened 2025-12-15
fix(detectors): update Dandelion token regex to match hex-only values

diegmonti:
### Description:
Change the regex pattern from [a-z0-9]{32} to [a-f0-9]{32} to only detect valid hexadecimal tokens, reducing false positives.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Low risk: a small regex tightening in a single detector that may reduce matches for non-hex tokens but does not change verification or data handling logic.
> 
> **Overview**
> **Dandelion secret detection is now stricter.** The Dandelion detector regex was narrowed from any lowercase alphanumeric 32-char value to *hex-only* (`[a-f0-9]{32}`) to reduce false positives.
> 
> Tests were updated to use a hex-only sample token so the pattern test continues to validate expected matches.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 00f333efca1b3a6ec74b8d7a70a7fa055ea69147. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/detectors/dandelion/dandelion.go
+++ b/pkg/detectors/dandelion/dandelion.go
@@ -22,7 +22,7 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"dandelion"}) + `\b([a-z0-9]{32})\b`)
+	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"dandelion"}) + `\b([a-f0-9]{32})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/dandelion/dandelion_test.go
+++ b/pkg/detectors/dandelion/dandelion_test.go
@@ -22,7 +22,7 @@ var (
 		api:
 			auth_type: "API-Key"
 			in: "Header"
-			dandelion_secret: "xccl325526f9cp6qzh89qkgoklje5ds9"
+			dandelion_secret: "abcd1234abcd1234abcd1234abcd1234"
 			base_url: "https://api.example.com/v1/example?token=$dandelion_secret"
 			response_code: 200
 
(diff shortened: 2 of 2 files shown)


shahzadhaider1:
Hi @diegmonti,
Thank you for taking the time to open this PR. We really appreciate your contribution to the project.

To help reviewers confidently assess this change, could you please share any supporting documentation, specification, or authoritative reference confirming that Dandelion tokens are strictly hexadecimal ([a-f0-9]{32})? 

Providing such context would make it easier for reviewers to verify the correctness of the change and ensure we’re not unintentionally missing valid tokens. Thanks again for your effort and support!

vad:
Hi @shahzadhaider1, Dandelion developer here. You can check my organizations here on github, my role at https://www.spaziodati.eu/team and the Dandelion ownership in the footer of https://dandelion.eu/ .

We don’t strictly enforce tokens to be hexadecimal, but it’s also true that there are still some old tokens around that aren’t even 32 characters long, as detected by the current regex. All recent tokens follow exactly the format proposed by @diegmonti , so I’m fine with it.

Also, since the current detector implementation makes API calls to check whether a token is valid, we’ve noticed some request spikes (maybe a CI gone wild?) that we’d like to try to reduce.

shahzadhaider1:
> Hi @shahzadhaider1, Dandelion developer here. You can check my organizations here on github, my role at https://www.spaziodati.eu/team and the Dandelion ownership in the footer of https://dandelion.eu/ .
> 
> We don’t strictly enforce tokens to be hexadecimal, but it’s also true that there are still some old tokens around that aren’t even 32 characters long, as detected by the current regex. All recent tokens follow exactly the format proposed by @diegmonti , so I’m fine with it.
> 
> Also, since the current detector implementation makes API calls to check whether a token is valid, we’ve noticed some request spikes (maybe a CI gone wild?) that we’d like to try to reduce.

Grazie Davide, for the explanation.

shahzadhaider1 (approved):

PR #4605, opened 2025-12-17
Update artifactory detector with basic auth support

zacharyyun:
### Description:
Enhance artifactory detector with basic auth detection and verification

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Verification sends real credentials to customer JFrog hosts and treats certain 403s as verified, which could affect false-positive/negative rates; otherwise this is an additive detector registration change.
> 
> **Overview**
> Adds a dedicated **ArtifactoryBasicAuth** detector and wires it into the default engine alongside the existing access-token scanner (Artifactory is now split into `accesstoken` and `basicauth` packages in `defaults`).
> 
> The new scanner finds embedded credentials in `user:pass@*.jfrog.io` URLs (with or without `https://`), dedupes matches, and emits `SecretParts` for domain/username/password. Optional verification calls `GET /artifactory/api/system/ping` with HTTP Basic auth, caches unreachable hosts, and treats some 403 responses (non–rate-limit) as evidence the credentials reached the instance.
> 
> Unit and integration tests cover regex edge cases and live verify/unverify paths; `DetectorType_ArtifactoryBasicAuth` (1070) is added to the proto enum.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68784189aced7b4066dbb517d179e9eb0948d228. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amanfcp (commented):
Thank you for the contribution @zacharyyun 

This seems like a suitable case for a V2. See [this](https://github.com/trufflesecurity/trufflehog/blob/main/hack/docs/Adding_Detectors_external.md#adding-new-token-formats-to-an-existing-scanner)

zacharyyun:
V1/V2 added, thanks @amanfcp

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
amanfcp, line 1: This version should only contain Basic Auth detection logic.

amanfcp (changes requested):

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: @amanfcp basic auth logic moved to v2

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,200 @@
...
+	username string
+	password string
+	host     string
+	raw      string
+}
+
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
```
shahzadhaider1: One small suggestion for consistency and safety:
since this detector now has multiple versions and implements Version(), we usually add a compile-time interface assertion:
```go
```
This is used in other multi-version detectors as well, to ensure versioning doesn’t accidentally break during refactors. Not a functional issue today, but helps with future maintainability.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,200 @@
...
+	case http.StatusUnauthorized, http.StatusForbidden, http.StatusFound:
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected HTTP response status %d", resp.StatusCode)
+	}
+}
+
+func (s Scanner) Type() detectorspb.DetectorType {
+	return detectorspb.DetectorType_ArtifactoryAccessToken
+}
```
shahzadhaider1: One observation I wanted to share: this detector isn’t actually detecting Artifactory access tokens, it’s detecting Basic Auth credentials embedded in URLs. Because of that, we may be mislabeling the secret type here.

I’m just sharing thoughts and opening a discussion, not prescribing a specific approach. A couple of options that come to mind:

One option would be to annotate this more explicitly via metadata, for example:
```go
ExtraData["auth_type"] = "basic"
```

That’s probably the simplest change, although I’m not sure it fully captures the distinction or addresses the underlying accuracy concern.

Another option, given that this introduces a new v2 detector with a different credential type, is to reconsider how we model this structurally. Right now we have:
1. detectors/artifactory/v1: access token detection
2. detectors/artifactory/v2: basic auth credential detection

While the v2 implementation itself isn’t necessarily wrong, using the same detector type for two different credential kinds makes me question the precision of the resulting findings.

An alternative could be to keep the same high-level structure, but separate by credential type instead of version, for example:
1. detectors/artifactory/accesstoken: access token detection
2. detectors/artifactory/basicauth: basic auth credential detection, using a dedicated `ArtifactoryBasicAuth` detector type

That would let us represent the secret type more accurately while keeping both detectors side-by-side.

Again, this is just an observation from my side. I’d be interested to hear your thoughts, and also happy to get input from others if anyone has a different perspective.

shahzadhaider1 (commented):
Great work on this, thanks! 
I have a couple of design-related suggestions around how the detector is structured that I’d like to discuss.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: > An alternative could be to keep the same high-level structure, but separate by credential type instead of version, for example:
> 
> 1. detectors/artifactory/accesstoken: access token detection
> 2. detectors/artifactory/basicauth: basic auth credential detection, using a dedicated `ArtifactoryBasicAuth` detector type

@shahzadhaider1 great observation, thanks for bringing this up. I do see other examples of other detectors using a similar format to this one and I agree this makes more sense over using v1/v2 with two distinct detector types.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Implemented alternative design without versioning, separated into accesstoken/basicauth categories

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Refactored to categorize by accesstoken/basicauth

shahzadhaider1:
The code looks good to me. However, when I tested it locally, the newly added Artifactory detector does not appear to detect Artifactory Basic Auth URL combinations. The URI detector successfully detects the test secrets, but when the URI detector is disabled, no results are returned. Please guide me if I am doing anything wrong. Screenshot attached.

image

zacharyyun:
> The code looks good to me. However, when I tested it locally, the newly added Artifactory detector does not appear to detect Artifactory Basic Auth URL combinations. The URI detector successfully detects the test secrets, but when the URI detector is disabled, no results are returned. Please guide me if I am doing anything wrong. Screenshot attached.
> 
> image

Missed swapping one detector field from ArtifactoryAccessToken to ArtifactoryBasicAuth, should be all set now

shahzadhaider1:
> > The code looks good to me. However, when I tested it locally, the newly added Artifactory detector does not appear to detect Artifactory Basic Auth URL combinations. The URI detector successfully detects the test secrets, but when the URI detector is disabled, no results are returned. Please guide me if I am doing anything wrong. Screenshot attached.
> > image
> 
> Missed swapping one detector field from ArtifactoryAccessToken to ArtifactoryBasicAuth, should be all set now

Hey, I don’t think this addresses the issue I was referring to. What I was trying to highlight is that the reported detector type for these findings is `URI`, not `ArtifactoryBasicAuth`. That suggests the newly added detector isn’t actually detecting these secrets--they’re being picked up by the existing [URI detector](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/uri/uri.go) instead.

Could you double-check whether the basic auth credentials detected in the screenshot match the format this new detector expects?

detector-type-uri

zacharyyun:
> 
> Could you double-check whether the basic auth credentials detected in the screenshot match the format this new detector expects?
> 

Hi @shahzadhaider1, I double checked and confirm the credentials in your SS match the format the detector expects. I looked further into the issue and found that it looks like the new detector picks up the patterns correctly, but is placing them in the filtered_unverified bucket. If you add `--results=filtered_unverified` to your command it should output the expected results. 

I haven't been able to figure out why my results are being put into the filtered_verification bucket after testing, if you can shed some light on how the actual filtering takes place, thanks!

shahzadhaider1:
> > Could you double-check whether the basic auth credentials detected in the screenshot match the format this new detector expects?
> 
> Hi @shahzadhaider1, I double checked and confirm the credentials in your SS match the format the detector expects. I looked further into the issue and found that it looks like the new detector picks up the patterns correctly, but is placing them in the filtered_unverified bucket. If you add `--results=filtered_unverified` to your command it should output the expected results.
> 
> I haven't been able to figure out why my results are being put into the filtered_verification bucket after testing, if you can shed some light on how the actual filtering takes place, thanks!

Hey @zacharyyun, good debugging on narrowing it down to `filtered_unverified`. Here's what's happening:
- The default false positive check (IsKnownFalsePositive in pkg/detectors/falsepositives.go) runs against `result.Raw` and does substring matching against an embedded wordlist of common English words. 
- The Raw field was set to cred.raw which is the full URL (e.g., https://user123:pass@test.jfrog.io/artifactory/api/pypi/pypi/simple).
- The wordlist contains words like "http:", so it was matching against the URL path rather than the actual secret. 
- Any match flags the result as IsWordlistFalsePositive = true, which filters it into `filtered_unverified`.

The fix is to set Raw to just the password (the secret material) instead of the full URL. This matches how the existing `ArtifactoryAccessToken` detector sets Raw to just the token. RawV2 already stores the full credential context (username:password@host) for deduplication, so that stays as-is.

The comment on IsKnownFalsePositive (line 83) actually calls this out: 
```go
// Only the secret key material should be passed into this function
```

I ran a small test and it now shows the detector findings. Let me know if this helps.

image

shahzadhaider1:
OR, if you want to do something like URI detector then check how it [implements](https://github.com/trufflesecurity/trufflehog/blob/daf5bf1e94212936dae852e8638e313b806e06c5/pkg/detectors/uri/uri.go#L141) `CustomFalsePositiveChecker` and its `IsFalsePositive` method unconditionally returns false.

shahzadhaider1:
Also, I merged main into your branch so please pull latest

zacharyyun:
> OR, if you want to do something like URI detector then check how it [implements](https://github.com/trufflesecurity/trufflehog/blob/daf5bf1e94212936dae852e8638e313b806e06c5/pkg/detectors/uri/uri.go#L141) `CustomFalsePositiveChecker` and its `IsFalsePositive` method unconditionally returns false.

Thank you for the insight @shahzadhaider1! That really cleared up what was going on with the filtered_unverified. I decided to keep the raw string and use the same false positive checker that URI implements, to get rid of potential blind spots where a legitimate password starts with some of the FP words.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			if isVerified {
+				if r.AnalysisInfo == nil {
+					r.AnalysisInfo = make(map[string]string)
+				}
+				r.AnalysisInfo["domain"] = cred.host
+				r.AnalysisInfo["username"] = cred.username
+				r.AnalysisInfo["password"] = cred.password
+				r.AnalysisInfo["authType"] = "basic"
```
shahzadhaider1: not sure if this is gonna work, because the analyzer expects a token, but that can be updated though.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,200 @@
...
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return false, err
+		}
+
+		if strings.Contains(string(body), "OK") {
+			return true, nil
+		}
+
+		return false, nil
```
shahzadhaider1, line 180: we have seen many cases where a service will return 200 status code along with a HTML body response (for one reason or another) and `strings.Contains` check will return true because its possible the `string` might exist in the HTML text. So, I'd suggest to do a json decode and look for a specific field in the response body, or we should at least do a header check to see if the response is json. I haven't checked this API's response so I'd like to hear your thoughts on this.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,200 @@
...
+	case http.StatusUnauthorized, http.StatusForbidden, http.StatusFound:
```
shahzadhaider1: I might be mistaken about this specific API, but I wanted to flag one thing to double-check. Some services return 403 when the credentials are actually valid, but the caller doesn’t have enough permissions for the requested operation. In that case, treating a 403 the same as 401 could cause us to mark valid credentials as unverified.

Since both 401 and 403 currently return `false` here, do we want to confirm that a 403 definitively means “invalid credentials” for this API?

pkg/detectors/postgres/postgres.go
```diff
@@ -55,6 +55,33 @@ var (
 type Scanner struct {
 	detectors.DefaultMultiPartCredentialProvider
 	detectLoopback bool // Automated tests run against localhost, but we want to ignore those results in the wild
+	ignorePatterns []*regexp.Regexp
+}
+
+func New(opts ...func(*Scanner)) *Scanner {
```
shahzadhaider1, line 68: these should not be a part of this PR.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
shahzadhaider1: pls disregard. We'll take care of this.

pkg/detectors/postgres/postgres.go
zacharyyun: Woops my mistake, reverted and repushed

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: I think this is a good distinction to check. I verified with the Artifactory docs and for the endpoint we're verifying against `/artifactory/api/system/ping` should always return a simple `OK` string. As an extra measure though, I updated from using contains to a strict check for this string:

`if strings.TrimSpace(string(body)) == "OK" `

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Good point here, Artifactory docs point to 403 being used for valid credentials but missing permissions. Updated to return true for 403.

shahzadhaider1 (approved):

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,202 @@
...
+
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	basicAuthURLPattern = regexp.MustCompile(
+		`https?://(?P<username>[^:@\s]+):(?P<password>[^@\s]+)@(?P<host>[A-Za-z0-9][A-Za-z0-9\-]{0,61}[A-Za-z0-9]\.jfrog\.io)(?P<path>/[^\s"'<>]*)?`,
```
kashifkhan0771: 🚀

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,202 @@
...
+	password string
+	host     string
+	raw      string
+}
+
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
+
```
kashifkhan0771, line 33: Missing `var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)`

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,202 @@
...
+	username string
+	password string
+	host     string
+	raw      string
+}
+
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
```
kashifkhan0771: I think we can safely remove this interface. It is used when we allow user to configure custom endpoints for the detector or we use found + cloud endpoints.

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
kashifkhan0771: Just caught my eye 👀 - we should remove the `https://` part. It’s not required and would cause us to miss URLs like `user:pass@myrepo.jfrog.io`.

pkg/detectors/artifactory/basicauth/artifactorybasicauth_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{
+			name: "invalid pattern - missing password",
+			input: `
+				[INFO] Sending request to the artifactory API
+				https://user123:@test.jfrog.io/artifactory/api/ping
+			`,
+			want: nil,
+		},
+		{
+			name: "invalid pattern - no basic auth uri present",
```
kashifkhan0771, line 85: Add one test case with following invalid patterns:
```
https://user@myrepo.jfrog.io         (no password)
https://user:pass@a.jfrog.io        (1-char subdomain)
https://user:pass@-repo.jfrog.io    (starts with -)
https://user:pass@repo-.jfrog.io    (ends with -)
https://user:pass@repo.com         (not jfrog.io)
https://user:pass@repo.jfrog.io?x=1 - match
user:pass@myrepo.jfrog.io - match
```
This would ensure that in future if someone update pattern, it does not cause any issues.

kashifkhan0771 (changes requested):

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Added

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Removed

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Set http/https prefix as optional so that the pattern wouldn't try and match on the `://` portion:

`(?P<prefix>https?://)?`

pkg/detectors/artifactory/basicauth/artifactorybasicauth_test.go
zacharyyun: Added test cases

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,201 @@
...
+		if verify {
+			isVerified, vErr := verifyArtifactoryBasicAuth(ctx, s.getClient(), cred.host, cred.username, cred.password)
+			r.Verified = isVerified
+
+			if vErr != nil {
+				if errors.Is(vErr, errNoHost) {
+					invalidHosts.Set(cred.host, struct{}{})
+					continue
+				}
+				r.SetVerificationError(vErr, cred.username, cred.host)
```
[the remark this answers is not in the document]
zacharyyun: Added password instead of host

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,201 @@
...
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
+
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	basicAuthURLPattern = regexp.MustCompile(
+		`(?P<prefix>https?://)?(?P<username>[^:@\s]+):(?P<password>[^:@\s]+)@(?P<host>[A-Za-z0-9][A-Za-z0-9\-]{0,61}[A-Za-z0-9]\.jfrog\.io)(?P<path>/[^\s"'<>]*)?`,
+	)
```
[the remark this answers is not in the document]
zacharyyun: Prefix removed

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
zacharyyun: Ended up removing prefix altogether like suggested for cases like `git+ssh://`. Added `\/` exclusion to username/password capture groups

zacharyyun:
@shahzadhaider1 @kashifkhan0771 updated enums, ready for review

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,211 @@
...
+			username: username,
+			password: password,
+			host:     host,
+			raw:      match[0],
+		}
+	}
+
+	for _, cred := range basicCreds {
+		if invalidHosts.Exists(cred.host) {
+			continue
```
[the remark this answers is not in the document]
zacharyyun: Updated

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,211 @@
...
+
+var (
+	// Ensure the Scanner satisfies the interface at compile time.
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
+
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	basicAuthURLPattern = regexp.MustCompile(
+		`(?P<username>[^:@\s\/]+):(?P<password>[^:@\s\/]+)@(?P<host>[A-Za-z0-9][A-Za-z0-9\-]{0,61}[A-Za-z0-9]\.jfrog\.io)(?P<path>/[^\s"'<>]*)?`,
```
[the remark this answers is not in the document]
zacharyyun: Updated

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,210 @@
...
+		key := username + ":" + password + "@" + host
+		if _, exists := basicCreds[key]; exists {
+			continue
+		}
+
+		basicCreds[key] = basicArtifactoryCredential{
+			username: username,
+			password: password,
+			host:     host,
+			raw:      match[0],
```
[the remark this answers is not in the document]
zacharyyun: Updated

pkg/detectors/artifactory/basicauth/artifactorybasicauth.go
```diff
@@ -0,0 +1,211 @@
...
+        if err != nil {
+            return false, err
+        }
+
+        // Ignore rate-limit / temporary block 403s
+        if strings.Contains(strings.ToLower(string(body)), "blocked due to recurrent request failures") {
+            return false, nil
+        }
+
+        return true, nil
```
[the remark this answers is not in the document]
zacharyyun: 403 returns true only the returned message is not related to a 403 rate limit, otherwise it can be an insufficient permissions message which would validate a set of basic auth credentials

mattbrady-1:
@zacharyyun thanks for sticking with this one — the work here looked complete, so to unblock it I've pushed a merge of `main` onto your branch (fast-forward only, your commits are untouched):

- Resolved the conflicts from the detector proto split (#4871): `ArtifactoryBasicAuth` now lives in `proto/detector_type.proto` as `1070` (`1045` got taken on main while this sat), with `detector_typepb` regenerated via `make protos`.
- Ported the new detector and its integration test to `detector_typepb`, and to the `SecretParts` field from #4911/#4948 — populated at `Result` construction like the accesstoken detector, and ignored in the integration test comparison the same way the referencetoken test does.
- Addressed the last Bugbot finding: `Raw` now carries just the password, while `RawV2` keeps the `username:password@host` composite for per-account dedup.

Build, unit tests for both artifactory packages plus `pkg/engine` and defaults, gofmt/vet, and `checksecretparts` are all green locally.

@kashifkhan0771 with the conflicts resolved this should be ready for a re-review — your earlier asks (scheme-less regex, host label rules, the `CustomFalsePositiveChecker` assertion, `EndpointSetter` removal, and the invalid-pattern test cases) were all addressed in the branch back in April.

PR #4606, opened 2025-12-17
[INS-170] Unify JDBC URL Parsing Across Detector and Analyzer (Continued)

mustansir14:
### Description:
This PR serves as an extension of #4574 by further refactoring the detector to unify code across JDBC's detector and analyzer.
Changes here will make better sense when viewed in conjunction with the corresponding [analyzer PR](https://github.com/trufflesecurity/integrations/pull/81).

Instead of publicizing seperate methods like `parseMySQL` and `buildMySQLConnectionString`, I've now added methods like `GetConnectionInfo`, `GetDBType` and `BuildConnectionString` to the existing `JDBC` interface, so that both the detector and analyzer can directly use the `NewJDBC` method and then call the methods they need.

Tests for parsing and building connection strings have been moved from the analyzer to the detector as well.

### Checklist:

### DIFF
--- a/pkg/detectors/jdbc/jdbc.go
+++ b/pkg/detectors/jdbc/jdbc.go
@@ -206,6 +206,12 @@ func tryRedactRegex(conn string) (string, bool) {
 	return newConn, true
 }
 
+var supportedSubprotocols = map[string]func(logContext.Context, string) (JDBC, error){
+	"mysql":      parseMySQL,
+	"postgresql": parsePostgres,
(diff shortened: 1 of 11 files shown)


amanfcp (approved):
This is cool! Thanks Mustansir

Now you've encapsulated all the methods and export only NewJDBC which is leaner approach.

I can't think of any reason for JDBC analyzer to use individual methods so we're good

camgunz (approved):

PR #4607, opened 2025-12-17
Rework JWT detector to better block local IPs

bradlarsen:
This PR updates the JWT detector to more simply and more robustly prevent making requests to local IPs.

The previous implementation did its checking at the hostname level, but this approach doesn't work in general when redirects are involved. The new implementation has less code and does its checking at the HTTP transport level, after DNS resolution.

This PR also adds HTTP instrumentation to the `DetectorHttpClientWith*` clients, as already exists in `SaneHttpClient`. This change should give us insight into some blind spots, as the `DetectorHttpClientWith*` clients are used in a few dozen existing detectors and previously had no HTTP instrumentation.

### DIFF
--- a/pkg/detectors/http.go
+++ b/pkg/detectors/http.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/feature"
 )
 
@@ -162,13 +163,15 @@ func WithTimeout(timeout time.Duration) ClientOption {
 }
 
 func NewDetectorHttpClient(opts ...ClientOption) *http.Client {
-	httpClient := &http.Client{
+	client := &http.Client{
 		Transport: NewDetectorTransport(nil),
 		Timeout:   DefaultResponseTimeout,
 	}
 
 	for _, opt := range opts {
-		opt(httpClient)
+		opt(client)
 	}
-	return httpClient
+
+	client.Transport = common.NewInstrumentedTransport(client.Transport)
+	return client
 }
--- a/pkg/detectors/http.go
+++ b/pkg/detectors/http.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


shahzadhaider1 (approved):
Approving to unblock. This looks good from a security standpoint, and I agree with using the detector HTTP client here.

One thing I hope we have considered: switching from common.SaneHttpClient means we lose the retry/instrumentation layer (metrics we currently export to Prometheus/Grafana).

pkg/detectors/jwt/jwt.go
```diff
@@ -78,12 +73,21 @@ var jwtOptions = []jwt.ParserOption{
...
 var jwtValidator = jwt.NewValidator(jwtOptions...)
+var client *http.Client
 
 // FromData will find and optionally verify JWT secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
-	client := cmp.Or(s.client, common.SaneHttpClient())
+	if client == nil {
+		// Use a client that disallows accessing local IPs.
+		// This avoids a potential blind SSRF.
+		client = detectors.NewDetectorHttpClient(
```
rosecodym: `FromData` can be called concurrently from multiple goroutines, so I believe this read/write needs some sort of synchronization. (My understanding is that this precise pattern generally works in the current version of Go but is technically disallowed by the spec, and I'd prefer to play it safe.)

rosecodym (commented):
I like the code consolidation but I think it's a bummer that this costs us some instrumentation. Did you look into whether we can get that easily as well? (If we can't, don't worry about it.)

I believe this PR does introduce a data race, though.

pkg/detectors/jwt/jwt.go
bradlarsen: Yikes! Good to know; thanks for the pointer.

pkg/detectors/jwt/jwt.go
```diff
@@ -17,14 +15,11 @@ import (
...
 
-	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
 )
 
-type Scanner struct {
-	client *http.Client
-}
+type Scanner struct{}
```
rgmz, line 22: IIRC the primary motivation for putting client inside the struct is test mocking. I don't know whether that still stands.

pkg/detectors/jwt/jwt.go
bradlarsen: I've fixed this in the latest.

pkg/detectors/jwt/jwt.go
bradlarsen: Thanks for the context @rgmz! I was wondering about this. In the majority of cases I've seen, though, it appears to be copy-pasta, and a totally dead field that is never assigned to. Whoops :)

pkg/detectors/http.go
```diff
@@ -165,13 +163,15 @@ func WithTimeout(timeout time.Duration) ClientOption {
...
 		Timeout:   DefaultResponseTimeout,
 	}
 
 	for _, opt := range opts {
-		opt(httpClient)
+		opt(client)
 	}
-	return httpClient
+
+	client.Transport = common.NewInstrumentedTransport(client.Transport)
```
bradlarsen, line 175: @rosecodym I added instrumentation like we use in `SaneHttpClient` here in `NewDetectorHttpClient`. A few dozen detectors used these `DetectorHttpClientWith*` clients instead of `SaneHttpClient`, meaning they were totally missing the HTTP instrumentation before.

Does this look reasonable to you? Tests have passed locally for me.

rosecodym (approved):

PR #4608, opened 2025-12-18
Gitlab Source: Backoff from Scan2 which is experimental to legacy pagination API call

kashifkhan0771:
This PR refactors simplifiedGitlabEnumeration to use the legacy pagination API with keyset pagination instead of Scan2, which is currently in an experimental state. This change is intended as an experiment and does not guarantee a fix for the issue.

Additionally, it introduces retry logic to handle 500-level errors and adds logging to track the number of projects being enumerated.

Related issue on Gitlab: https://gitlab.com/gitlab-com/gl-infra/production/-/issues/7516

### Checklist:


pkg/sources/gitlab/gitlab.go
martinlocklear: I don't understand why we have membership set to true.  (I'm not saying we shouldn't, just that I don't understand)

From [the docs](https://docs.gitlab.com/api/projects/#:~:text=MM%3ASSZ\)-,membership,Limit%20by%20projects%20that%20the%20current%20user%20is%20a%20member%20of.,-min_access_level):

> Limit by projects that the current user is a member of.

If this flag is potentially causing problems, why add in this limitation?  Why not just scan everything that's visible to the current token, member or not?

pkg/sources/gitlab/gitlab.go
```diff
@@ -690,62 +692,86 @@ func (s *Source) getAllProjectReposV2(
...
-		if err := reporter.UnitOk(ctx, unit); err != nil {
-			return err
+		// if next page is empty, break the loop
+		if resp == nil || resp.NextLink == "" {
+			// No more pages to fetch. This is the normal loop exit condition.
+			// It also acts as a safety stop if the current request failed.
+			break
 		}
+		// Only update the token for the next page if we have a valid, non-empty link.
+		projectQueryOptions.PageToken = resp.NextLink
```
martinlocklear: So this gave me a lot of pause to see a link being put in something called a token.  It then took me way too long to find what _should_ be happening here.  I _think_ this won't work as you've done it, but I'm way more confident that they following _will_ work (assuming you can confirm):

```diff
diff --git a/pkg/sources/gitlab/gitlab.go b/pkg/sources/gitlab/gitlab.go
index cb0558514..770029dfe 100644
--- a/pkg/sources/gitlab/gitlab.go
+++ b/pkg/sources/gitlab/gitlab.go
@@ -696,11 +696,12 @@ func (s *Source) getAllProjectReposV2(
 	// It includes all projects fetched from the API, even those later skipped by ignore rules.
 	totalCount := 0
 	maxRetries := 3
+	requestOptions := []gitlab.RequestOptionFunc{gitlab.WithContext(ctx)}

 	// Pagination loop: Continue fetching pages until the API indicates there are no more.
 	for {
 		// Fetch a page of projects from the GitLab API using the current query options.
-		projects, resp, err := apiClient.Projects.ListProjects(projectQueryOptions)
+		projects, resp, err := apiClient.Projects.ListProjects(projectQueryOptions, requestOptions...)
 		if err != nil {
 			if errResp, ok := err.(*gitlab.ErrorResponse); ok {
 				// check if it is a server side error and we have some retries left
@@ -771,7 +772,9 @@ func (s *Source) getAllProjectReposV2(
 			break
 		}
 		// Only update the token for the next page if we have a valid, non-empty link.
-		projectQueryOptions.PageToken = resp.NextLink
+		requestOptions = []gitlab.RequestOptionFunc{
+			gitlab.WithContext(ctx),
+			gitlab.WithKeysetPaginationParameters(resp.NextLink),
+		}
 	}

 	ctx.Logger().Info("Enumerated GitLab projects", "count", totalCount)
```

We should definitely find a way to test this though

pkg/sources/gitlab/gitlab.go
```diff
@@ -690,62 +692,86 @@ func (s *Source) getAllProjectReposV2(
...
-	for project, projectErr := range projectsIter {
-		if projectErr != nil {
-			err := fmt.Errorf("error during project enumeration: %w", projectErr)
-
-			if reportErr := reporter.UnitErr(ctx, err); reportErr != nil {
-				return reportErr
+	// Pagination loop: Continue fetching pages until the API indicates there are no more.
+	for {
+		// Fetch a page of projects from the GitLab API using the current query options.
+		projects, resp, err := apiClient.Projects.ListProjects(projectQueryOptions)
```
martinlocklear: Shouldn't we still convey the context here?

```suggestion
		projects, resp, err := apiClient.Projects.ListProjects(projectQueryOptions, gitlab.WithContext(ctx))
```

pkg/sources/gitlab/gitlab.go
```diff
@@ -690,62 +692,86 @@ func (s *Source) getAllProjectReposV2(
...
+					ctx.Logger().V(2).Info("retrying API call",
+						"retry_attempts_left", maxRetries)
+					// delay for 2 second between retries
+					time.Sleep(time.Second * 2)
+					continue
+				}
 			}
-
-			continue
+			err = fmt.Errorf("received error on listing projects, you probably don't have permissions to do that: %w", err)
```
martinlocklear:
```suggestion
			err = fmt.Errorf("received error on listing projects, you might not have permissions to do that: %w", err)
```

pkg/sources/gitlab/gitlab.go
martinlocklear: We might consider putting the retry logic here, rather than rolling our own down below

```suggestion
		apiClient, err := gitlab.NewOAuthClient(
			s.token,
			gitlab.WithBaseURL(s.url),
			gitlab.WithCustomRetryWaitMinMax(time.Second, 5*time.Second),
			gitlab.WithCustomRetryMax(3),
		)
```

martinlocklear (changes requested):
Thanks for looking at this! Some notes:
* Definitely need test coverage for this before merging in.
* I notice in our go.mod we're [overriding](https://github.com/trufflesecurity/trufflehog/blob/606a7edfe3573e675e4f0bc9f0e8d076e42fa981/go.mod#L12-L13) the version of the library we're using.  Is it possible that some of the issues that we've been seeing could be fixed by newer versions of this library?

pkg/sources/gitlab/gitlab.go
kashifkhan0771: The reason for adding this is that, if we don’t set `membership` to `true` for GitLab Cloud, the scan will enumerate **all public GitLab projects**. This is different from GitHub, where we explicitly restrict what can be scanned with the following condition:

```go
    return scanMetrics, fmt.Errorf("invalid config: you must specify at least one organization or repository")
}
```

This ensures that GitHub scans are limited to explicitly specified organizations or repositories. In GitLab, however, we don’t currently have an `org` option. As a result, a scan can run with just a token and attempt to enumerate everything that token has access to.

To prevent this on GitLab Cloud, we set `membership=true` so that only projects the user is a member of are enumerated. For self-managed GitLab instances, we disable this behavior, allowing the scan to enumerate all projects the token has access to on that instance.

I see two possible approaches:

1. Add support for organization/group-based scanning in GitLab (similar to GitHub). With that restriction in place, we could remove the `membership` flag.
2. Keep the current behavior to avoid unintentionally scanning all public projects on GitLab Cloud.

The second option ensures we don’t accidentally scan the entire set of public GitLab Cloud projects.

pkg/sources/gitlab/gitlab.go
kashifkhan0771: I did testing locally, It is working but your approach seem a more clean approach I will try this.

pkg/sources/gitlab/gitlab.go
kashifkhan0771: This is good 🚀

pkg/sources/gitlab/gitlab.go
```diff
@@ -471,7 +482,12 @@ func (s *Source) newClient() (*gitlab.Client, error) {
 		}
 		fallthrough
 	case "TOKEN":
-		apiClient, err := gitlab.NewOAuthClient(s.token, gitlab.WithBaseURL(s.url))
+		apiClient, err := gitlab.NewOAuthClient(
+			s.token,
+			gitlab.WithBaseURL(s.url),
+			gitlab.WithCustomRetryWaitMinMax(time.Second, 5*time.Second),
```
shahzadhaider1, line 509: Seems like these are used in quite a few places. It'll be better to create constants for min and max.

pkg/sources/gitlab/gitlab.go
```diff
@@ -471,14 +472,24 @@ func (s *Source) newClient() (*gitlab.Client, error) {
...
 		return apiClient, nil
 
 	case "BASIC_AUTH":
-		apiClient, err := gitlab.NewBasicAuthClient(s.user, s.password, gitlab.WithBaseURL(s.url))
+		apiClient, err := gitlab.NewOAuthClient(
+			s.token,
+			gitlab.WithBaseURL(s.url),
+			gitlab.WithCustomRetryWaitMinMax(time.Second, 5*time.Second),
+			gitlab.WithCustomRetryMax(3),
+		)
```
shahzadhaider1, line 493: Why are we using token/oauth here instead of basic auth?

pkg/sources/gitlab/gitlab.go
```diff
@@ -711,62 +727,79 @@ func (s *Source) getAllProjectReposV2(
...
+	requestOptions := []gitlab.RequestOptionFunc{gitlab.WithContext(ctx)}
 
-			if reportErr := reporter.UnitErr(ctx, err); reportErr != nil {
-				return reportErr
+	// Pagination loop: Continue fetching pages until the API indicates there are no more.
+	for {
+		// Fetch a page of projects from the GitLab API using the current query options.
+		projects, resp, err := apiClient.Projects.ListProjects(projectQueryOptions, requestOptions...)
+		if err != nil {
+			err = fmt.Errorf("received error on listing projects, you might not have permissions to do that: %w", err)
```
shahzadhaider1, line 742: totally optional thing:

```suggestion
			err = fmt.Errorf("received an error while listing projects; this may be due to insufficient permissions: %w", err)
```

pkg/sources/gitlab/gitlab.go
kashifkhan0771: Good catch! Must have been the wind 🌬️

pkg/sources/gitlab/gitlab.go
kashifkhan0771: `time.Second` is already a constant from `time` and I don't like the idea of making `5` only a constant here. Maybe I am wrong but usually we use time like this `n*time.Second`.

shahzadhaider1 (approved):

pkg/sources/gitlab/gitlab.go
martinlocklear: Oh! Ok, that makes sense to me -  I trust your judgement on the best path forward here.

martinlocklear (changes requested):
I think we still need to add unit testing before we merge this.

pkg/sources/gitlab/gitlab.go
kashifkhan0771: I did some testing and found that, by default, GitLab keeps the `membership` flag set to `false`. To take advantage of it, I reversed some conditions to avoid setting the `membership` flag for non-GitLab Cloud instances. This could help if the issue was indeed related to the membership flag. I also set the `simple` flag to `true`; according to the docs, this returns only the essential fields for each project instead of the full, large JSON objects.

kashifkhan0771:
> I think we still need to add unit testing before we merge this.

Added the test case

camgunz (approved):
LGTM! If you want you could add a couple more checks in the test, like I'm pretty sure some other sources check chunk count and so on. But up to you--I know you've done a lot of local testing w/ this.

camgunz:
@martinlocklear 
> I notice in our go.mod we're [overriding](https://github.com/trufflesecurity/trufflehog/blob/606a7edfe3573e675e4f0bc9f0e8d076e42fa981/go.mod#L12-L13) the version of the library we're using. Is it possible that some of the issues that we've been seeing could be fixed by newer versions of this library?

I just had a look, and while [the changelog](https://gitlab.com/gitlab-org/api/client-go/-/blob/main/CHANGELOG.md) doesn't go as far back as our pinned version, I didn't see any bug fix that would've affected us. I made [INS-249](https://trufflesecurity.atlassian.net/browse/INS-249) to do this some day.

[INS-249]: https://trufflesecurity.atlassian.net/browse/INS-249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pkg/sources/gitlab/gitlab_integration_test.go
```diff
@@ -813,3 +814,68 @@ func TestSource_Enumerate_ProjectDetailsInChunkMetadata(t *testing.T) {
 		t.Errorf("0 chunks scanned.")
 	}
 }
+
+func TestSource_Chunks_SimplifiedGitlabEnumeration(t *testing.T) {
+	// use simplified gitlab enumeration
+	feature.UseSimplifiedGitlabEnumeration.Store(true)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
```
martinlocklear: Just use the testing context, I think it's t.Context()

pkg/sources/gitlab/gitlab_integration_test.go
```diff
@@ -813,3 +814,68 @@ func TestSource_Enumerate_ProjectDetailsInChunkMetadata(t *testing.T) {
...
+			connection: &sourcespb.GitLab{
+				Credential: &sourcespb.GitLab_Token{
+					Token: token,
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
martinlocklear: Why even are we building this test table and then building a test runner?  Let's just write a test.  If we want to make our test setup and breakdown reusable, then we can convert this to a test suite and use SetupSuite or (better) SetupTest for that purpose.

kashifkhan0771:
> I notice in our go.mod we're [overriding](https://github.com/trufflesecurity/trufflehog/blob/606a7edfe3573e675e4f0bc9f0e8d076e42fa981/go.mod#L12-L13) the version of the library we're using. Is it possible that some of the issues that we've been seeing could be fixed by newer versions of this library?

Somehow I missed this comment. The reason we did that was the latest version at that time was v1.134.0 but it deprecated existing Auth methods, also the Scan2 API was experimental so we pinned to version to avoid any new updates which accidentally break the API.

As Charlie said there isn't any bug fix in later versions, so I would prefer to update that in a separate PR.

pkg/sources/gitlab/gitlab_integration_test.go
```diff
@@ -813,3 +814,68 @@ func TestSource_Enumerate_ProjectDetailsInChunkMetadata(t *testing.T) {
 		t.Errorf("0 chunks scanned.")
 	}
 }
+
+func TestSource_Chunks_SimplifiedGitlabEnumeration(t *testing.T) {
```
martinlocklear, line 822: Can we (in comments) describe what the purpose of this test is? Like, what would it tell us if it was failing, and what does it give us confidence in when it's passing.

martinlocklear:
Some non-blocking comments about the new test

pkg/sources/gitlab/gitlab_integration_test.go
kashifkhan0771: `t.Context()` returns the standard library `context.Context`, but this codebase uses TruffleHog’s custom context.Context (pkg/context) which extends the standard context with additional methods. Because of that, `t.Context()` cannot be used here - the types are incompatible.

martinlocklear: ugh - I forgot about that

martinlocklear (approved):

PR #4612, opened 2025-12-22
[INS-228] Add ignorePattern configuration support to Postgres and Sqlserver detectors

mustansir14:
### Description:
This PR adds ignorePattern configuration support to the Postgres and Sqlserver detectors. It closely follows the implementation for [JDBC detector](https://github.com/mustansir14/trufflehog/blob/0eb8260b079399f49f51b19c6b1ef789d092f970/pkg/detectors/jdbc/jdbc.go#L33)

### Checklist:


shahzadhaider1 (approved):

camgunz (commented):
I see that a couple other detectors have `New` methods, but I can't find where that's called. Do you know where that is?

mustansir14:
> I see that a couple other detectors have `New` methods, but I can't find where that's called. Do you know where that is?

Not really. My assumption was that enterprise calls it at some point, but I guess that's not the case?

As mentioned in the description, I closely followed JDBC detector's implementation here, which also has this `New` method.

camgunz:
I found it; you're right it's in EE.

camgunz (changes requested):
This looks good! I just want to make sure this works OK if we don't call `.New`, because that's a separate change we'll have to make in EE, so the change I'm requesting here is to add a couple tests where we don't call `.New` and things work OK.

mustansir14:
> This looks good! I just want to make sure this works OK if we don't call `.New`, because that's a separate change we'll have to make in EE, so the change I'm requesting here is to add a couple tests where we don't call `.New` and things work OK.

I have added tests for SqlServer that don't call `New` and directly create the detector. For Postgres, these tests already exist.

camgunz (approved):

PR #4613, opened 2025-12-22
[INS-233] Added support to verify token against all datadog domains

MuneebUllahKhan222:
### Description:

- Previously, API/App key verification was done against a single Datadog endpoint.
- Now, the verification iterates over multiple endpoints:
-The key is tested against each endpoint in order.
-if verification succeeds on any endpoint, the loop breaks immediately.
-If no endpoint succeeds, the key is marked as unverified.

### Checklist:

### DIFF
--- a/pkg/detectors/datadogtoken/datadogtoken.go
+++ b/pkg/detectors/datadogtoken/datadogtoken.go
@@ -3,6 +3,7 @@ package datadogtoken
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -23,7 +23,16 @@ var _ detectors.Detector = (*Scanner)(nil)
...
+func (Scanner) CloudEndpoint() string { return "" }
+
+var datadogAPIBaseURLs = []string{
+	"https://api.us5.datadoghq.com/api", // Default domain
+	"https://api.app.datadoghq.com/api",
+	"https://api.us3.datadoghq.com/api",
+	"https://api.app.datadoghq.eu/api",
+	"https://api.app.ddog-gov.com/api",
+	"https://api.ap1.datadoghq.com/api",
+}
```
kashifkhan0771: I’m strongly opposed to this approach. Making six API calls for validation of all findings is not ideal.

kashifkhan0771: What can we do here?

* Allow users to configure endpoints for this detector. If we have a fixed list of verification endpoints, this can be useful.
* Add regex-based detection to extract endpoints from the data (without restricting verification exclusively to them).
* Add a fallback default cloud endpoint.

If a user configures an endpoint, the engine logic will handle it and use it exclusively to verify Datadog tokens. If no endpoint is configured, we will attempt to extract endpoints from the data and also use the default cloud endpoint. This approach increases the likelihood of selecting the correct endpoint for verification.

Using a constant list of endpoints is not ideal. For example, if we find 10 false tokens, that would result in 10 × 6 = 60 API calls for verification.

kashifkhan0771 (changes requested):

kashifkhan0771:
Closing in favor of: https://github.com/trufflesecurity/trufflehog/pull/4616

Issue #4578, opened 2025-12-02
Incorrect instructions for scanning local repo


cgoldberg:
The instructions in the repo's README for scanning a local git repo don't work:

https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#9-scan-a-local-git-repo

### TruffleHog Version

trufflehog 3.91.2

### Expected Behavior

Local git repo should be scanned.

### Actual Behavior

When running `trufflehog git file://test_keys --results=verified,unknown`, it fails cloning the repo, even though the repo exists and was already cloned in the previous step.

```
2025-12-02T15:19:20-05:00       error   trufflehog      error running scan      {"error": "failed to scan Git: error preparing repo: failed to clone file Git repo (file://C:%5CUsers%5Ccorey.goldberg%5Ccode%5Ctest_keys): could not clone repo: , error executing git clone: exit status 128, ssh: Could not resolve hostname file: Name or service not known\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n"}
```

### Steps to Reproduce

run commands:

```
$ git clone git@github.com:trufflesecurity/test_keys.git
$ trufflehog git file://test_keys --results=verified,unknown
```

## Environment
 * Windows 11 (using Bash)
 
## Additional Context

I can scan a local repo fine by cloning it and then running a filesystem scan:

```
$ git clone git@github.com:trufflesecurity/test_keys.git
$ trufflehog filesystem test_keys --results=verified,unknown
```

... but that's not how the instructions say to scan a local git repo.

11philip22:
I am facing the same issue on 3.92.2, Windows 10. Using powershell

Rusted2361:
@kashifkhan0771 please assign this issue to me.

Rusted2361:
I have regenerated the issue on windows 11, bash shell. Working on the fix

Rusted2361:
The issue is fixed in following PR: https://github.com/trufflesecurity/trufflehog/pull/4614
It will be reviewed and merged soon.

PR #4614, opened 2025-12-22
Fix/issue 4578 path normalization for unix and windows

Rusted2361:
The issue described in below ticket was regenerated and fixed in this PR.
https://github.com/trufflesecurity/trufflehog/issues/4578
Resolves issues with cloning local Git repositories on Windows:
- Enable filepath.ToSlash() conversion in normalizeFileURI() to properly format Windows paths in file:// URIs
- Strip leading slashes from index file paths incase of Windows path formats

### DIFF
--- a/pkg/analyzer/analyzers/client.go
+++ b/pkg/analyzer/analyzers/client.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/config"
+	"golang.org/x/time/rate"
 )
 
 type AnalyzeClient struct {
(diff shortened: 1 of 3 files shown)


Rusted2361:
@kashifkhan0771 please review this PR

kashifkhan0771:
Approved from my side, but these changes also require approval from @trufflesecurity/product-eng. Unfortunately, most of the team is currently out on holidays, so we’ll need to wait a bit.

Thanks for the PR - once the team is back, we’ll get it merged 🙂

Rusted2361:
@kashifkhan0771 sorry to ping you can you @ mention rufflesecurity/product-eng team to review the PR :)

pkg/sources/git/git.go
```diff
@@ -1341,7 +1349,17 @@ func PrepareRepo(ctx context.Context, uriString, clonePath string, trustLocalGit
 			if !isRepoBare(path) {
 				// Only copy index file for non-bare clones from working directory repos. This is used to see staged changes.
 				// Note: To scan **un**staged changes in the future, we'd need to set core.worktree to the original path.
-				originalIndexPath := filepath.Join(strings.TrimPrefix(normalizedURI.String(), "file://"), gitDirName, "index")
+
+				uriPath := strings.TrimPrefix(normalizedURI.String(), "file://")
+
+				// on windows we get "/C:/path". We need to remove the leading slash.
+				// on unix/macintosh, we get "/path" which needs the leading slash preserved.
+				if len(uriPath) >= 3 && uriPath[0] == '/' && uriPath[2] == ':' {
```
rosecodym: Is there a more robust way to know that TruffleHog is running on Windows? This seems pretty fragile.

pkg/sources/git/git.go
```diff
@@ -1341,7 +1349,17 @@ func PrepareRepo(ctx context.Context, uriString, clonePath string, trustLocalGit
 			if !isRepoBare(path) {
 				// Only copy index file for non-bare clones from working directory repos. This is used to see staged changes.
 				// Note: To scan **un**staged changes in the future, we'd need to set core.worktree to the original path.
-				originalIndexPath := filepath.Join(strings.TrimPrefix(normalizedURI.String(), "file://"), gitDirName, "index")
+
+				uriPath := strings.TrimPrefix(normalizedURI.String(), "file://")
```
shahzadhaider1: A potential bug (maybe): `normalizedURI.String()` percent-encodes special characters in the path. For example, a path like `C:\Users\John Doe\repos` may becomes `file:///C:/Users/John%20Doe/repos`. When you strip `file://` via string manipulation, the `%20` remains encoded, resulting in an invalid filesystem path.

Since normalizedURI is already a *url.URL, you can access the decoded path directly:
```go
uriPath := normalizedURI.Path
```

pkg/sources/git/git.go
shahzadhaider1: Have we considered using `runtime.GOOS` for platform detection instead of pattern matching:
```go
if runtime.GOOS == "windows" {
    uriPath = strings.TrimPrefix(uriPath, "/")
}
```

Rusted2361: thanks for the kind review @shahzadhaider1 I have updated the PR indeed using GOOS to idnetify platform and use URL object to access path is a better approach here

kashifkhan0771: Hmm, I initially considered suggesting `runtime.GOOS`, but decided against it. The reason is that OS ≠ path format. It’s better to handle this as: `If it looks like a Windows drive-letter path, treat it as one` rather than assuming `If it’s Windows, the path must look like this.`

Anyway, this looks fine too :)

camgunz (changes requested):
I reviewed without looking at the other reviews, I third the request for tests haha

camgunz (approved):
LG, thanks!

Rusted2361:
@rosecodym please review & merge this one if it looks good to you.

Rusted2361:
@kashifkhan0771 can you merge this one? As its approved by other collaborators and might get closed by git workflows if not got merged by 22 of Jan

kashifkhan0771:
Thank you @Rusted2361 🚀

PR #4616, opened 2025-12-24
[INS-254] Datadog detector verification fix and endpoint configuration

MuneebUllahKhan222:
### Description:
Datadog token verification previously relied on an App-key–dependent endpoint (/api/v2/users) to validate the Application Key + API Key combination. If this endpoint returned an error (for example, due to insufficient App Key permissions), the detector would mark the token as unverified and skip API key validation entirely. 
Additionally, Datadog uses region-specific API endpoints. When the hard-coded cloud endpoint did not match the user’s actual region, verification requests were sent to the wrong endpoint.

This behaviour caused valid Datadog API keys to be reported as unverified, resulting in false negatives.

To resolve this issue I have done the following changes:

- Removed API key validation from this detector and introduced a dedicated API key detector [here](https://github.com/trufflesecurity/trufflehog/pull/4627)

- Added support for user-provided Datadog endpoints.
     - If no endpoint is explicitly provided, the detector attempts to infer it from the data using regex.
     - If inference fails, it falls back to the default cloud endpoint hard-coded in the codebase.

### Corpora Test
The detector does not appear in the corpora list.
duo-detector-corpora-results

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes detector matching/verification behavior and endpoint selection, which can alter findings and verification outcomes across environments. Risk is mainly around missed/extra detections if endpoint inference or endpoint ordering behaves unexpectedly.
> 
> **Overview**
> Datadog token verification now **infers and tries region-specific Datadog API endpoints** found in scanned content (including `ddog-gov`) via a new URL regex, and records the successful `endpoint` alongside `api_key`/`app_key` in `AnalysisInfo`.
> 
> The detector no longer emits/validates **API-key-only** findings; it only reports `Application+APIKey` pairs and stops after the first successful endpoint verification. Tests were updated accordingly (new endpoint-aware integration assertion, new unit tests, and removal of the API-key-only verified case).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0ca4bc25a1f819dd9cf08dd6a65fe593a18c9c69. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -95,7 +99,7 @@ func setOrganizationInfo(opt []*options, s1 *detectors.Result) {
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"datadog"}
+	return []string{"datadog", "datadoghq", "ddog-gov"}
```
kashifkhan0771: `datadoghq` already contains the word `datadog`, so it isn’t necessary to include it separately. We can just add `ddog` to the list along with `datadog`.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -23,14 +24,17 @@ var _ detectors.Detector = (*Scanner)(nil)
...
 
 var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	appPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
-	apiPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	appPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
+	apiPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	datadogURLPat = regexp.MustCompile(`\b(api(?:\.[a-z0-9-]+)?\.(?:datadoghq|ddog-gov)\.[a-z]{2,3})\b`)
```
kashifkhan0771: It always ends with `.com/api` or `.eu/api`. That could help filter out incorrect endpoints.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -105,12 +109,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	appMatches := appPat.FindAllStringSubmatch(dataStr, -1)
 	apiMatches := apiPat.FindAllStringSubmatch(dataStr, -1)
 
+	foundURLs := extractDatadogURLs(dataStr)
```
kashifkhan0771:
```suggestion
        var uniqueFoundUrls = make(map[string]struct{})
	for _, matches := range datadogURLPat.FindAllStringSubmatch(dataStr, -1) {
	  uniqueFoundUrls[matches[1]] = struct{}{}
	}
```

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -105,12 +109,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	endpoints := s.confiureEndpoints(foundURLs...)
```
kashifkhan0771: I think it's better to direct loop over `s.configureEndpoints(uniqueFoundUrls...)` instead of creating a new variable

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -105,12 +109,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	// 1. both api key and app key found - we store only one result with both keys
+	// 2. only api key found - we store only api key result
+	// In case verify is true:
+	// 3. both api key and app key found - we store only one result with both keys and incase the appkey call fails we also also store apikey result
+	// 4. only api key found - we store only api key result
+
 	for _, apiMatch := range apiMatches {
 		resApiMatch := strings.TrimSpace(apiMatch[1])
-		appIncluded := false
+		// This variable makes sure that we verify api key separately only when the api key check fails with an app key
```
kashifkhan0771: Question: Is the API key useful without app key? Can we access something with it without requiring app key at all? If no than we do not need to verify it separately. If yes, I think than it deserves a separate detector.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -196,3 +215,30 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+	if len(endpoints) > 0 {
+		formattedEndpoints := make([]string, 0, len(endpoints))
+		for _, url := range endpoints {
+			formattedEndpoints = append(formattedEndpoints, fmt.Sprintf("https://%s", url))
+		}
+		return s.Endpoints(formattedEndpoints...)
+	} else {
+		return s.Endpoints(defaultDatadogAPIBaseURL)
+	}
+}
```
kashifkhan0771: I don't think this is required. Add the `defaultDatadogAPIBaseURL` as cloud endpoint. You can append the https in start of urls at the time of verification but you need to make sure to first remove it from the configured endpoints. If user configure endpoint with `https://` we should remove it first and than append it at the time of verification.

pkg/detectors/datadogtoken/datadogtoken_integration_test.go
```diff
@@ -54,6 +54,10 @@ func TestDatadogToken_FromChunk(t *testing.T) {
 					ExtraData: map[string]string{
 						"Type": "Application+APIKey",
 					},
+					AnalysisInfo: map[string]string{
```
kashifkhan0771, line 58: The analysis info will require the endpoint as well.

kashifkhan0771 (changes requested):

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: If we configure `defaultDatadogAPIBaseURL` as cloud endpoint then on every call to `s.Endpoints()` it will return an array with `defaultDatadogAPIBaseURL` which is not required until and unless no url is found in the data source.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: Yes the API key is useful because it can be used for write operations on datadogs's v1 endpoints like `/api/v1/events` etc.
We can definitely discuss adding a v2 detector for it.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: I thinks its better that we store the result in endpoints variable because  if we directly loop over `s.configureEndpoints(uniqueFoundUrls...)`  we might have to do it twice incase the app token verification fails and we have to verify the api key separately.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: Also these endpoints won't have `https://` at the beginning as the regex matches the string after `https://`.

pkg/detectors/datadogtoken/datadogtoken_integration_test.go
MuneebUllahKhan222: If we add endpoint to AnalysisInfo then we unable to test it with our table test structure and we are currently doing struct comparison directly and we won't know what endpoint to expect.

pkg/detectors/datadogtoken/datadogtoken.go
kashifkhan0771: Alright, then it definitely deserves a separate detector. The current detector is named `datadogtoken`, so adding a "V2" doesn’t feel right unless it was originally named `datadog`. In any case, I’m not in favor of renaming the existing detector since that could disrupt existing live secrets.

I’d suggest introducing a new detector, `datadogapikey`, that only matches the API key pattern. We can then remove the API key validation from this detector and keep a single validation that covers both the App key and the API key in this detector.

pkg/detectors/datadogtoken/datadogtoken.go
kashifkhan0771: `s.Endpoints` only returns the cloud endpoint when no endpoints are configured. If we don’t find any endpoint in the data, we shouldn’t leave the secret unchecked. We should fall back to a default cloud endpoint for verification. So keeping the cloud endpoint is the best option here.

kashifkhan0771: > Also these endpoints won't have https:// at the beginning as the regex matches the string after https://.

I am not sure what do you mean by this, you can configure cloud endpoint without `https://` and add it always in the verification step.

pkg/detectors/datadogtoken/datadogtoken_integration_test.go
kashifkhan0771: That’s fine - we can either handle it dynamically in the test case or skip it entirely during comparison. However, if we don’t add it, it will break the analyzer in Enterprise, which relies on `AnalysisInfo` from the detector to analyze a validated secret. Without this, the analyzer would have no way of knowing which endpoint to use.

kashifkhan0771: Also we need to update the analyzer a bit to take the domain as input. We can connect about that and I can explain you how we can do that.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: Ok, then as discussed I will create a v2 for api verification.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -105,12 +107,26 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	if len(uniqueFoundUrls) == 0 {
+		// In case no endpoints were found in data, use the default cloud endpoint
+		s.UseCloudEndpoint(true)
+		endpoints = s.Endpoints()
+	} else {
+		// In case endpoints were found in data, use them
+		// also add cloud endpoint at the end so if not validated against found endpoints, try cloud endpoint as well
+		endpoints = s.configureEndpoints(uniqueFoundUrls)
+		endpoints = append(endpoints, s.CloudEndpoint())
+	}
```
kashifkhan0771: `UseCloudEndpoint` defaults to `true` when no endpoints are configured. I think it’s clearer to reverse the condition and explicitly disable it only when we find endpoints in the data:

```go
// If we found any endpoints in the data, disable the cloud endpoint
// and validate only against the found endpoints.
if len(uniqueFoundUrls) > 0 {
	s.UseCloudEndpoint(false)
}

// Endpoints behavior:
// A) If the user configured endpoints, only those will be returned.
// B) If endpoints were found in the data, only those will be returned.
// C) If neither applies, it will return only cloud endpoint.
endpoints = s.Endpoints(uniqueFoundUrls...)
```

kashifkhan0771: After this, we can drop the `configureEndpoints` function and normalize the endpoints:

```go
for i, endpoint := range endpoints {
	// Normalize endpoints so they all share the same format.
	// - User-configured endpoints may be missing "https://"
	// - Discovered endpoints also omit the scheme
	if !strings.HasPrefix(endpoint, "https://") {
		endpoint = "https://" + endpoint
	}

	// Ensure all endpoints end with "/api"
	// (user-configured and cloud endpoints may omit it)
	if !strings.HasSuffix(endpoint, "/api") {
		endpoint = endpoint + "/api"
	}

	endpoints[i] = endpoint
}

// At this point, all endpoints start with "https://" and end with "/api",
// so they’re ready for direct use.
```

kashifkhan0771: This can be normalized even more I believe but this seems like a reasonable good approach to me.

kashifkhan0771 (changes requested):

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: I have reversed the condition like mentioned above. Also I have bunched all the endpoints logic in the `configureEndpoints` function to increase code coverage and mapping directly over it to avoid using a variable to store the endpoints.

kashifkhan0771 (approved):

kashifkhan0771:
Good work 🚀

shahzadhaider1 (approved):

nabeelalam (approved):
Looks good! Thanks!

MuneebUllahKhan222:
This PR is the base of a stacked PR. So [PR #4661](https://github.com/trufflesecurity/trufflehog/pull/4661) should be merged before it.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -105,12 +106,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 	apiMatches := apiPat.FindAllStringSubmatch(dataStr, -1)
 
+	var uniqueFoundUrls = make(map[string]struct{})
+	for _, matches := range datadogURLPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueFoundUrls["https://"+matches[1]] = struct{}{}
+	}
+	endpoints := make([]string, 0, len(uniqueFoundUrls))
+	for endpoint := range uniqueFoundUrls {
+		endpoints = append(endpoints, endpoint)
+	}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Nope, we decided that the default endpoint will be tried first then the detected endpoint.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -134,7 +142,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 						defer res.Body.Close()
 						if res.StatusCode >= 200 && res.StatusCode < 300 {
 							s1.Verified = true
-							s1.AnalysisInfo = map[string]string{"apiKey": resApiMatch, "appKey": resAppMatch}
+							s1.AnalysisInfo = map[string]string{"api_key": resApiMatch, "app_key": resAppMatch, "endpoint": baseURL}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: it is updated in different PR

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -29,8 +29,9 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	appPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
-	apiPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	appPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
+	apiPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	datadogURLPat = regexp.MustCompile(`\b(api(?:\.[a-z0-9-]+)?\.(?:datadoghq|ddog-gov)\.[a-z]{2,3})\b`)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Fixed.

camgunz (approved):

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -29,8 +29,9 @@ var (
 	client = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	appPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
-	apiPat = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	appPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{40})\b`)
+	apiPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
+	datadogURLPat = regexp.MustCompile(`\b(api(?:\.[a-z0-9-]+)?\.(?:datadoghq|ddog-gov)\.(com|eu))\b`)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Fixed.

PR #4626, opened 2025-12-30
feat: add JumpCloud API key v2 detector (jca_ prefix)

### LINKED ISSUE #4624: Add JumpCloud API Key v2 detector (jca_ prefix format)

## Description

JumpCloud has introduced a new API key format with a `jca_` prefix. The existing JumpCloud detector only matches generic 40-character alphanumeric strings when the keyword "jumpcloud" is nearby, which can lead to false positives and miss keys that don't have contextual keywords.

The new format is: **40 characters total** (`jca_` prefix + 36 alphanumeric characters)
Example: `jca_aBcDeFgHiJkLmNoPqRsTuVwXyZ123456`

### Preferred Solution

Add a v2 detector for JumpCloud that:
1. Detects the new `jca_` prefixed API key format
2. Uses the same verification endpoint (`https://console.jumpcloud.com/api/v2/systemgroups`)
3. Implements the `Versioner` interface alongside the existing v1 detector
4. Includes false positive filtering (only digits/letters, sequential patterns, low entropy)

### Key Format
- **Total length**: 40 characters
- **Prefix**: `jca_` (4 characters)
- **Key part**: 36 alphanumeric characters
- **Pattern**: `jca_[a-zA-Z0-9]{36}`
- **Keywords**: `jca_` (self-identifying, no context needed)

### Additional Context
- JumpCloud hosts data (user/device management, IAM)
- Has paid services (enterprise directory platform)
- The `jca_` prefix makes the key self-identifying, reducing false positives significantly
- Verification can use the existing API endpoint with `x-api-key` header

#### References
- JumpCloud API Docs: https://docs.jumpcloud.com/api/
- JumpCloud API Key Support: https://jumpcloud.com/support/jumpcloud-apis
- Existing detector: `/pkg/detectors/jumpcloud/

rootranjan:
Closes #4624

### Description:
JumpCloud has introduced a new API key format with a `jca_` prefix. This PR adds a v2 detector to support this new format.

**Changes:**
- Add v2 detector for modern JumpCloud API keys (40 chars total: `jca_` + 36 alphanumeric)
- Include false positive filtering (only digits/letters, sequential patterns)
- Add entropy check (threshold 3.0)
- Move existing detector to v1 with Versioner interface
- Both versions share same DetectorType but report version in ExtraData

**Key Format:**
- Prefix: `jca_`
- Pattern: `jca_[a-zA-Z0-9]{36}`
- Total length: 40 characters

### Checklist:


pkg/detectors/jumpcloud/v2/jumpcloud.go
```diff
@@ -0,0 +1,130 @@
...
+var _ detectors.Versioner = (*Scanner)(nil)
+
+func (Scanner) Version() int { return 2 }
+
+var (
+	client = common.SaneHttpClient()
+
+	// JumpCloud API keys with jca_ prefix: jca_ + 36 alphanumeric characters (40 total)
+	// Example: jca_aBcDeFgHiJkLmNoPqRsTuVwXyZ123456
+	keyPat = regexp.MustCompile(`\bjca_([a-zA-Z0-9]{36})\b`)
```
shahzadhaider1, line 29: Can you share the source/documentation of this?

This [page](https://jumpcloud.com/support/changes-to-connect-keys) says the new format is prefixed with `jcc_` and the length is also greater that what you have added.

rootranjan: Hi @shahzadhaider1,
I couldn’t find any official JumpCloud docs that specify the full format or length of the admin API key. The [JumpCloud APIs](https://jumpcloud.com/support/jumpcloud-apis) help article only states that the admin API key is prefixed with jca_ [(“You’ll receive an API key, it has the prefix jca_ before it…” in the Accessing Your API Key section)](https://jumpcloud.com/support/jumpcloud-apis#:~:text=You%27ll%20receive%20an%20API%20key%2C%20it%20has%20the%20prefix%20%27jca%5F%27%20before%20it%20to%20help%20you%20search%20for%20it%20easier%2E).

The page you linked ([Understand Changes to Connect Keys](https://jumpcloud.com/support/changes-to-connect-keys)) describes the newer Connect key format, which is prefixed with `jcc_` and is a Base64-encoded JSON blob used for agent enrollment, not the admin API key.

In this detector I’m targeting the `jca_` admin API key and I’ve chosen the length/charset based on observed keys rather than a published spec. I’ve verified the pattern against valid keys and it matches the current format. If you’d like to double-check, [you can follow the steps in the Accessing Your API Key section to generate a key and confirm it matches this pattern](https://jumpcloud.com/support/jumpcloud-apis#:~:text=Accessing%20Your%20API%20Key).

rootranjan: Hi,
Any update?

pkg/detectors/jumpcloud/v1/jumpcloud.go
shahzadhaider1, line 56: Let's not skip the entry in case of error here. We should add s1 to results even with an error.

shahzadhaider1: While we are at it, a more cleaner approach willl be to shift the verification logic to a dedicated function.

pkg/detectors/jumpcloud/v2/jumpcloud.go
shahzadhaider1: Thanks for the update. Can you clarify what the V1 detector currently detects; Admin API keys or Connect keys?

If V1 already targets Admin API keys, then this PR looks fine. However, if V1 is for detection for a different secret type (e.g. Connect keys), we should probably create a separate detector type instead of a V2 detector. We typically use V2 only when the provider introduces a new format for the same underlying secret type.

pkg/detectors/jumpcloud/v2/jumpcloud.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"jca_"}
+}
+
+// isLikelyFalsePositive checks if the key portion looks like a placeholder or test value
+func isLikelyFalsePositive(key string) bool {
```
shahzadhaider1, line 45: Have we considered implementing the `detectors.CustomFalsePositiveChecker` interface?

pkg/detectors/jumpcloud/v2/jumpcloud.go
```diff
@@ -0,0 +1,130 @@
...
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_Jumpcloud,
+			Raw:          []byte(fullMatch),
+			ExtraData: map[string]string{
+				"version": fmt.Sprintf("%d", s.Version()),
+			},
+		}
+
+		if verify {
+			req, err := http.NewRequestWithContext(ctx, "GET", "https://console.jumpcloud.com/api/v2/systemgroups", nil)
```
shahzadhaider1, line 95: Since the API endpoint is same, I think we can create one function for verify and re-use in both versions of the detector. What do you think?

shahzadhaider1 (commented):
This looks good overall, thanks for the work on this. I’ve added a few comments, but we’re very close.

PR #4627, opened 2025-12-30
[INS-241] Datadogapikey detector

MuneebUllahKhan222:
### Description:
Previously, the `datadogtoken` detector was responsible for verifying both API keys and Application keys for Datadog. This caused an issue where an invalid Application key would incorrectly mark a valid API key as unverified, leading to false positives.

To resolve this, we have introduced a new detector: `datadogapikey`. This detector is focused solely on validating Datadog API keys, ensuring that API key verification is isolated and accurate, independent of the App key status.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Introduces a new detector type and modifies Datadog credential parsing/verification and analyzer output shape, which may affect downstream integrations and test expectations.
> 
> **Overview**
> Adds a new `datadogapikey` detector that identifies Datadog API keys, optionally verifies them against `/api/v1/validate`, and surfaces `api_key`/`endpoint` as `AnalysisInfo` so API-key verification no longer depends on an application key.
> 
> Updates the Datadog analyzer to accept `api_key`/`app_key` (snake_case), allow an optional `endpoint` (skipping domain auto-detection when provided), and support API-key-only analysis by attaching default API-key permissions and emitting bindings under an `Unknown User` when user info can’t be fetched. Tests and golden outputs are updated accordingly, plus a new protobuf enum value `DatadogApikey` is introduced.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1a6e7a5362c508b8c2f3879d10fad7d20e9faaf4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -34,16 +34,19 @@ func (a Analyzer) Analyze(ctx context.Context, credInfo map[string]string) (*ana
...
-	info, err := AnalyzePermissions(a.Cfg, apiKey, appKey)
+	// Endpoint
+	endpoint := credInfo["endpoint"]
+
+	info, err := AnalyzePermissions(a.Cfg, apiKey, appKey, endpoint)
 	if err != nil {
 		return nil, err
 	}
-
+	fmt.Printf("SecretInfo: %+v\n", info)
```
kashifkhan0771: Please remove this `print` statement from here.

kashifkhan0771 (changes requested):

pkg/analyzer/analyzers/datadog/datadog.go
MuneebUllahKhan222: Done

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,96 @@
...
+				res, err := client.Do(req)
+				if err == nil {
+					defer res.Body.Close()
+					if res.StatusCode >= 200 && res.StatusCode < 300 {
+						s1.Verified = true
+						s1.AnalysisInfo = map[string]string{"apiKey": resApiMatch, "endpoint": baseURL}
+						// break the loop once we've successfully validated the token against a baseURL
+						break
+					}
+				}
```
kashifkhan0771: For new detectors, we prefer to add verification in a separate function with proper status code handling. Refer to any newly added detector.

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, apiMatch := range apiMatches {
+		resApiMatch := strings.TrimSpace(apiMatch[1])
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_DatadogApikey,
+			Raw:          []byte(resApiMatch),
+			RawV2:        []byte(resApiMatch),
+			ExtraData: map[string]string{
+				"Type": "APIKeyOnly",
```
kashifkhan0771: Not required as the type already indicates that this is API Key

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,96 @@
...
+					continue
+				}
+				req.Header.Add("Content-Type", "application/json")
+				req.Header.Add("DD-API-KEY", resApiMatch)
+				res, err := client.Do(req)
+				if err == nil {
+					defer res.Body.Close()
+					if res.StatusCode >= 200 && res.StatusCode < 300 {
+						s1.Verified = true
+						s1.AnalysisInfo = map[string]string{"apiKey": resApiMatch, "endpoint": baseURL}
```
kashifkhan0771: We need to add the analysis info in the other datadog detector as well.

pkg/tui/pages/analyze_form/analyze_form.go
```diff
@@ -106,6 +106,10 @@ func New(c common.Common, keyType string) *AnalyzeForm {
 			Key:         "appKey",
 			Required:    true,
 			RedactInput: true,
+		}, {
+			Label:    "Endpoint (optional)",
```
kashifkhan0771: "Endpoint (press Enter to skip if unknown; TruffleHog will attempt to auto-detect)"

Do not need to add `optional` as we mark the field as Non Required.

kashifkhan0771 (changes requested):

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: I have refactored the code by referencing **dotdigital** detector.

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Removed.

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Analysis info is already present in the `datadogtoken` detector.

pkg/tui/pages/analyze_form/analyze_form.go
MuneebUllahKhan222: Done.

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,131 @@
...
+		results = append(results, s1)
+	}
+	return results, nil
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, apiKey, baseUrl string) (bool, error) {
+	// Reference: https://docs.datadoghq.com/api/latest/authentication/
+
+	timeout := 10 * time.Second
+	client.Timeout = timeout
```
kashifkhan0771: We should not overwrite the timeout

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,131 @@
...
+	url := baseUrl + "/api/v1/validate"
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
```
kashifkhan0771: `req, err := http.NewRequestWithContext(ctx, http.MethodGet, baseUrl+"/api/v1/validate", http.NoBody)`

pkg/detectors/datadogapikey/datadogapikey.go
kashifkhan0771: Are we adding `endpoint` there too?

MuneebUllahKhan222: Yes we are. Here is the  [PR](https://github.com/trufflesecurity/trufflehog/pull/4616) with datadogtoken detector changes.

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Done.

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Done.

pkg/tui/pages/analyze_form/analyze_form.go
kashifkhan0771: This is not resolved yet

MuneebUllahKhan222: My bad, I thought that I only needed to remove `(optional)`.
I have updated the label now.

pkg/tui/pages/analyze_form/analyze_form.go
```diff
@@ -106,6 +106,10 @@ func New(c common.Common, keyType string) *AnalyzeForm {
 			Key:         "appKey",
 			Required:    true,
 			RedactInput: true,
+		}, {
+			Label:    "Endpoint Endpoint (press Enter to skip if unknown; TruffleHog will attempt to auto-detect)",
```
shahzadhaider1: Endpoint is repeated here.

pkg/tui/pages/analyze_form/analyze_form.go
```diff
@@ -106,6 +106,10 @@ func New(c common.Common, keyType string) *AnalyzeForm {
 			Key:         "appKey",
 			Required:    true,
 			RedactInput: true,
+		}, {
```
shahzadhaider1, line 109: From my understanding of the PR description, the API key and application key are distinct credentials and can be verified independently, which is why we now have separate detectors for each. However, since we use a single analyzer for Datadog, it appears that the analyzer treats the application key as optional while requiring the API key. Is this the intended behavior? Also, do both keys provide access to the same set of resources?

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,127 @@
...
+	for endpoint := range uniqueFoundUrls {
+		endpoints = append(endpoints, "https://"+endpoint)
+	}
+
+	for _, apiMatch := range apiMatches {
+		resApiMatch := strings.TrimSpace(apiMatch[1])
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_DatadogApikey,
+			Raw:          []byte(resApiMatch),
+			RawV2:        []byte(resApiMatch),
```
shahzadhaider1: Do we really need a RawV2 here?

pkg/tui/pages/analyze_form/analyze_form.go
MuneebUllahKhan222: Fixed.

pkg/tui/pages/analyze_form/analyze_form.go
MuneebUllahKhan222: Yup, you are right both app key and apikey are distinct credentials and can we verified independently, I will/have introduced a validation check in Datadog's analyzer to make sure both appkey and apikey are passed to the analyzer.
Also all the api key have the same permissions(Send metrics, logs and event) so we wont't need a analyzer for it.

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Thanks for pointing this out, I think we don't need RawV2 in this detector hence I have removed it

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -57,16 +61,30 @@ func AnalyzeAndPrintPermissions(cfg *config.Config, apiKey string, appKey string
...
+	var err error
+
+	// If endpoint is provided, use it directly; otherwise detect domain
+	if endpoint != "" {
+		baseURL = endpoint + "/api"
+	} else {
+		baseURL, err = DetectDomain(client, apiKey, appKey)
+		if err != nil {
+			return nil, fmt.Errorf("[x] %v", err)
+		}
```
amanfcp, line 83: I think this is right, keeping the domain detection helps OSS. 
In Enterprise, the Analyzer would only run after successfully detecting so I think we're good.

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,126 @@
...
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+var _ detectors.CloudProvider = (*Scanner)(nil)
+
+func (Scanner) CloudEndpoint() string { return "https://api.datadoghq.com" }
+
+var (
+	client = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	apiPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
```
amanfcp:
```suggestion
	apiKeyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
```

amanfcp (approved):

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,126 @@
...
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+var _ detectors.CloudProvider = (*Scanner)(nil)
+
+func (Scanner) CloudEndpoint() string { return "https://api.datadoghq.com" }
+
+var (
+	client = common.SaneHttpClient()
+
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	apiKeyPat     = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z-0-9]{32})\b`)
```
nabeelalam: Checking for a 32-character string that contains numbers or letters seems a little broad. Maybe an entropy test on the found matches would filter out the "non secret" strings?

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,126 @@
...
+	datadogURLPat = regexp.MustCompile(`\b(api(?:\.[a-z0-9-]+)?\.(?:datadoghq|ddog-gov)\.[a-z]{2,3})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"datadog", "ddog-gov"}
```
nabeelalam: Since the URL secret must contain either `datadoghq` or `ddog-gov` it would be a bit more accurate to have those as the pre-filtering keywords.

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,126 @@
...
+	endpoints := make([]string, 0, len(uniqueFoundUrls))
+	for endpoint := range uniqueFoundUrls {
+		endpoints = append(endpoints, "https://"+endpoint)
+	}
+
+	for _, apiMatch := range apiMatches {
+		resApiMatch := strings.TrimSpace(apiMatch[1])
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_DatadogApikey,
+			Raw:          []byte(resApiMatch),
```
nabeelalam, line 68: I think associated endpoint the detected key is being tested on should be part of `RawV2` in the results. It will help distinguish verified key-endpoint pairs with unverified ones.

nabeelalam: Could add the URL in the result's extra data as well.

MuneebUllahKhan222: URL added in ExtraData.

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -26,29 +26,33 @@ func (a Analyzer) Type() analyzers.AnalyzerType {
...
-		return nil, errors.New("API key not found in credentials info")
-	}
+	apiKey := credInfo["apiKey"]
 
 	// Get appKey if provided
 	appKey := credInfo["appKey"]
 
-	info, err := AnalyzePermissions(a.Cfg, apiKey, appKey)
+	// Endpoint
+	endpoint := credInfo["endpoint"]
```
shahzadhaider1: this comment isn't very useful, we can remove this

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		if verify {
+			for _, baseURL := range s.Endpoints(endpoints...) {
+				client := s.getClient()
+				isVerified, verificationErr := verifyMatch(ctx, client, resApiMatch, baseURL)
+				s1.Verified = isVerified
+				s1.SetVerificationError(verificationErr, resApiMatch)
+				if isVerified {
+					s1.AnalysisInfo = map[string]string{"apiKey": resApiMatch, "endpoint": baseURL}
```
shahzadhaider1: The analyzer expects both an API key and an app key. How do you see this working?

pkg/analyzer/analyzers/datadog/datadog.go
MuneebUllahKhan222: Sure

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Yes you are right current analyzer will amicably throw a validation error. But we should definitely run the analyzer on apikey.
What do you reckon should we write a different analyzer for apikey or should we just add the apikey analysis in the same analyzer? since the api key will doesn't have granular permissions and will always have the same scopes?

shahzadhaider1: I am not sure about the analyzer behavior but if we could use either of the key to do the `analyzing`, then the analyzer shouldn't enforce inputting both keys and work whatever key is received, only throw error when both the keys are missing. Would that work?

MuneebUllahKhan222: Yeah I think that would work, bit I will also discuss it with rest of the team just in case.

shahzadhaider1: sure

MuneebUllahKhan222: Hi @shahzadhaider1, I have updated the analyzer to cater just the apikey analysis. Kindly review it.

shahzadhaider1 (approved):

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -114,7 +143,17 @@ func secretInfoToAnalyzerResult(info *SecretInfo) *analyzers.AnalyzerResult {
 	}
 
 	permissionBindings := secretInfoPermissionsToAnalyzerPermission(info.Permissions)
-	result.Bindings = analyzers.BindAllPermissions(*userResource, *permissionBindings...)
+	if userResource != nil && len(*permissionBindings) > 0 {
```
camgunz, line 154: **suggestion:** I might unify these under an `if len(*permissionBindings) > 0` block; it makes it more clear that condition is required to set bindings here.

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -26,47 +26,72 @@ func (a Analyzer) Type() analyzers.AnalyzerType {
...
-	// First detect which DataDog domain works with this API key
-	baseURL, err := DetectDomain(client, apiKey, appKey)
-	if err != nil {
-		return nil, fmt.Errorf("[x] %v", err)
+	var baseURL string
+	var err error
+
+	// If endpoint is provided, use it directly; otherwise detect domain
+	if endpoint != "" {
+		baseURL = endpoint + "/api"
```
camgunz: **issue:** Let's use [`url.JoinPath`](https://pkg.go.dev/net/url#JoinPath) here instead (I can't remember if I've given you my spiel about manipulating URLs as strings, but tl;dr: it's risky). There's a few places in this PR where we should change this; I won't note them though.

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		results = append(results, s1)
+	}
+	return results, nil
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, apiKey, baseUrl string) (bool, error) {
+	// Reference: https://docs.datadoghq.com/api/latest/authentication/
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, baseUrl+"/api/v1/validate", http.NoBody)
```
camgunz, line 92: I feel like I must have seen `http.NoBody` before, but I guess not! TIL!

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,127 @@
...
+		s1 := detectors.Result{
+			DetectorType: detectorspb.DetectorType_DatadogApikey,
+			Raw:          []byte(resApiMatch),
+		}
+
+		if verify {
+			for _, baseURL := range s.Endpoints(endpoints...) {
+				client := s.getClient()
+				isVerified, verificationErr := verifyMatch(ctx, client, resApiMatch, baseURL)
+				s1.Verified = isVerified
```
camgunz: **suggestion:** This is a very small thing, but usually this is a bug where you'll overwrite a meaningful value. Cleverly, the loop here exits in that case so there's no actual danger of this happening, but it'd be a little safer if everything that's supposed to happen when `isVerified` is `true` happens in that conditional block. That way someone can't come by later and undo the assumption (that we definitely `break` on `isVerified == true`).

camgunz (changes requested):
The only changes I want are the `JoinPath` changes; other than that great work!

pkg/analyzer/analyzers/datadog/datadog.go
MuneebUllahKhan222: `userResource != nil` check is important because we deference the pointer and derefrencing a nil pointer will lead a panic

pkg/detectors/datadogapikey/datadogapikey.go
MuneebUllahKhan222: Good Suggestion!

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -26,47 +27,75 @@ func (a Analyzer) Type() analyzers.AnalyzerType {
...
 func (a Analyzer) Analyze(ctx context.Context, credInfo map[string]string) (*analyzers.AnalyzerResult, error) {
-	apiKey, exist := credInfo["apiKey"]
-	if !exist {
-		return nil, errors.New("API key not found in credentials info")
-	}
-
-	// Get appKey if provided
-	appKey := credInfo["appKey"]
+	apiKey := credInfo["api_key"]
+	appKey := credInfo["app_key"]
```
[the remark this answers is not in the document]
MuneebUllahKhan222: this comment has been fixed in [this PR](https://github.com/trufflesecurity/trufflehog/pull/4616)

pkg/analyzer/analyzers/datadog/requests.go
```diff
@@ -172,6 +177,46 @@ func (h *HttpStatusTest) RunTest(client *http.Client, baseURL string, headers ma
...
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	// If we get a response that's not a connection error, this domain works
+	if resp.StatusCode == http.StatusOK {
+		return true, nil
+	}
+
+	return false, errors.New("unable to validate any provided DataDog API key")
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Fixed.

pkg/detectors/datadogapikey/datadogapikey.go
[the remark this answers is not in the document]
MuneebUllahKhan222: This is not an issue.

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -26,47 +27,75 @@ func (a Analyzer) Type() analyzers.AnalyzerType {
...
 		color.Red("[x] Error : %s", err.Error())
 	}
 
-	color.Green("[i] Valid Datadog API Key\n")
+	if info == nil {
+		color.Red("[x] No information retrieved")
+		return
+	}
 
+	color.Green("[i] Valid Datadog API Key\n")
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This doesn't matter as `AnalyzePermissions` always returns `(nil, error)` in case of an error.

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -26,47 +27,75 @@ func (a Analyzer) Type() analyzers.AnalyzerType {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to validate api key: %v", err)
+		}
+		if !isValidApiKey {
+			return nil, errors.New("invalid api key provided")
+		}
+		if err := CaptureApiKeyPermissions(client, baseURL, apiKey, appKey, secretInfo); err != nil {
+			return nil, fmt.Errorf("failed to fetch permissions: %v", err)
+		}
+		return secretInfo, nil
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Fixed

pkg/detectors/datadogapikey/datadogapikey.go
```diff
@@ -0,0 +1,126 @@
...
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	apiKeyPat     = regexp.MustCompile(detectors.PrefixRegex([]string{"datadog", "dd"}) + `\b([a-zA-Z0-9-]{32})\b`)
+	datadogURLPat = regexp.MustCompile(`\b(api(?:\.[a-z0-9-]+)?\.(?:datadoghq|ddog-gov)\.[a-z]{2,3})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"datadog", "ddog-gov"}
+}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: We have `dd` in `PrefixRegex` list, no need to add it in keywords

pkg/analyzer/analyzers/datadog/datadog_test.go
```diff
@@ -51,6 +55,27 @@ func TestAnalyzer_Analyze(t *testing.T) {
...
+			name:     "valid datadog credentials with invalid endpoint",
+			apiKey:   apiKey,
+			appKey:   appKey,
+			endpoint: "https://api.eu.datadoghq.com",
+			want: []byte(fmt.Sprintf(`{
+				"AnalyzerType": %s,
+				"Bindings": [],
+				"UnboundedResources": null,
+				"Metadata": {}
+			}`, analyzers.AnalyzerTypeDatadog)),
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Fixed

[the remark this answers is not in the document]
MuneebUllahKhan222: Fixed

pkg/analyzer/analyzers/datadog/datadog.go
```diff
@@ -26,47 +27,79 @@ func (a Analyzer) Type() analyzers.AnalyzerType {
...
 func (a Analyzer) Analyze(ctx context.Context, credInfo map[string]string) (*analyzers.AnalyzerResult, error) {
-	apiKey, exist := credInfo["apiKey"]
-	if !exist {
-		return nil, errors.New("API key not found in credentials info")
-	}
-
-	// Get appKey if provided
-	appKey := credInfo["appKey"]
+	apiKey := credInfo["api_key"]
+	appKey := credInfo["app_key"]
```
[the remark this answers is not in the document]
MuneebUllahKhan222: this comment has been fixed in [this PR](https://github.com/trufflesecurity/trufflehog/pull/4616)

camgunz (approved):

kashifkhan0771 (approved):

PR #4636, opened 2026-01-01
[INS-243] Fix jdbc detector detecting incomplete connection string and fixed invalid…

MuneebUllahKhan222:
### Description:
Earlier, we found that the JDBC detector was missing certain valid connection strings, which led to unverified secrets in scans:

- Connection strings with query parameters (e.g., ?user=...&password=...) were not matched by the regex.

- JDBC URLs containing parentheses (e.g., `jdbc:mysql://testuser:testpassword@tcp(localhost:1521)/testdb`) were also not matched.

- The `ParseSqlServer` method failed to correctly extract host and port from strings with a driver prefix like `//odbc:server=localhost;port=1433;database=testdb;password=testpassword`.

This PR fixes these issues by:

- Updating the JDBC regex to allow query parameters and parentheses inside the connection string while still preventing trailing special characters.

- Enhancing `ParseSqlServer` to correctly handle connection strings with optional driver prefixes like odbc: and extract host and port reliably.

### Checklist:

### DIFF
--- a/pkg/detectors/jdbc/jdbc.go
+++ b/pkg/detectors/jdbc/jdbc.go
@@ -50,7 +50,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`(?i)jdbc:[\w]{3,10}:[^\s"'<>,(){}[\]&]{10,512}`)
+	// Matches typical JDBC connection strings amd ingores any special character at the end
+	keyPat = regexp.MustCompile(`(?i)jdbc:[\w]{3,10}:[^\s"'<>,{}[\]]{10,511}[A-Za-z0-9]`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/jdbc/sqlserver.go
+++ b/pkg/detectors/jdbc/sqlserver.go
@@ -59,6 +59,11 @@ func ParseSqlServer(ctx logContext.Context, subname string) (jdbc, error) {
 			continue
 		}
 
+		// incase there is a bridge between jdbc and odbc, and conn string looks like this odbc:server
+		if split := strings.Split(key, ":"); len(split) > 1 {
+			key = split[1]
+		}
+
 		switch strings.ToLower(key) {
 		case "password", "spring.datasource.password", "pwd":
 			password = value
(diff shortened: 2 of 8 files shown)


kashifkhan0771 (approved):

amanfcp (approved):

nabeelalam (approved):
Looks good @MuneebUllahKhan222. Could you update this branch?

pkg/detectors/jdbc/sqlserver.go
```diff
@@ -73,6 +73,11 @@ func parseSqlServer(ctx logContext.Context, subname string) (JDBC, error) {
 			continue
 		}
 
+		// incase there is a bridge between jdbc and odbc, and conn string looks like this odbc:server
```
camgunz: **question/suggestion:** Reading this, it sounds like we just want to pop off any `odbc:` prefix. If that's true, we can use [`strings.CutPrefix`](https://pkg.go.dev/strings#CutPrefix) to do that.

pkg/detectors/jdbc/jdbc.go
```diff
@@ -50,7 +50,8 @@ var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.CustomFalsePositiveChecker = (*Scanner)(nil)
 
 var (
-	keyPat = regexp.MustCompile(`(?i)jdbc:[\w]{3,10}:[^\s"'<>,(){}[\]&]{10,512}`)
+	// Matches typical JDBC connection strings amd ingores any special character at the end
+	keyPat = regexp.MustCompile(`(?i)jdbc:[\w]{3,10}:[^\s"'<>,{}[\]]{10,511}[A-Za-z0-9]`)
```
camgunz, line 54: **question:** This changed from "{10,512}" to "{10,511}". Should it be "{9,511}"?

camgunz (approved):
Just a couple non-blocking Qs; feel free to merge 👍🏻

pkg/detectors/jdbc/sqlserver.go
MuneebUllahKhan222: This is a good suggestion but there can be other drivers like sqlserver, mysql etc that can pop in place of odbc in the connection string so that's why we are going with `string.Split`.
However I will update the comment in the code to make it less specific.

pkg/detectors/jdbc/jdbc.go
MuneebUllahKhan222: No it should be `{10,511}` as we don't want to change the minimum number of characters that can occur.

PR #4642, opened 2026-01-05
[INS-242] Add more validations to Custom Detector config

mustansir14:
### Description:
We recently faced a bug in enterprise where because of some extra operations performed on the regex fields, it became invalid and an error was thrown further down the pipeline. This PR adds more validations to `NewWebhookCustomRegex` so that these errors are surfaced up-front instead of silently failing.

### Checklist:

### DIFF
--- a/pkg/custom_detectors/validation.go
+++ b/pkg/custom_detectors/validation.go
@@ -32,6 +32,26 @@ func ValidateRegex(regex map[string]string) error {
 	return nil
 }
 
+func ValidateRegexSlice(regex []string) error {
+	for i, reg := range regex {
+		if _, err := regexp.Compile(reg); err != nil {
+			return fmt.Errorf("regex '%d': %w", i+1, err)
+		}
+	}
+	return nil
+}
+
+// validates if a provided non-empty primary regex name exists in the map of regexes
+func ValidatePrimaryRegexName(primaryRegexName string, regexes map[string]string) error {
+	if primaryRegexName == "" {
+		return nil
+	}
+	if _, ok := regexes[primaryRegexName]; !ok {
+		return fmt.Errorf("unknown primary regex name: %q", primaryRegexName)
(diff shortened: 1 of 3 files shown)


mcastorina (approved):
Thanks!

shahzadhaider1 (approved):

pkg/custom_detectors/validation.go
```diff
@@ -32,6 +32,26 @@ func ValidateRegex(regex map[string]string) error {
...
+func ValidatePrimaryRegexName(primaryRegexName string, regexes map[string]string) error {
+	if primaryRegexName == "" {
+		return nil
+	}
+	if _, ok := regexes[primaryRegexName]; !ok {
+		return fmt.Errorf("unknown primary regex name: %q", primaryRegexName)
+	}
+	return nil
+}
+
```
kashifkhan0771, line 54: The code for this is backward compatible
```go
// if the match is of the primary regex, set it's value as primary secret value in result
if c.PrimaryRegexName == key {
  result.SetPrimarySecretValue(secret)
}
```

If no primary regex name is provided or a wrong one is entered it simply fallback to default approach and try to find line number of secret using Raw. I think that is more better instead of raising an error for just simply adding a wrong regex name.

Thought?

kashifkhan0771: This is not a blocking comment.

kashifkhan0771 (approved):

pkg/custom_detectors/validation.go
mustansir14: Since this PR is about tightening validations, I added this as well. This is mainly to catch human-errors like typos, and to avoid unexpected behavior. 

If I made a typo while specifying a primary regex, I'd like to know about it up-front, instead of having Trufflehog use the default approach.

PR #4651, opened 2026-01-08
Fix bug in v1 figma token detector that caused it to incorrectly flag…

thiggy1342:
… deactivated tokens as valid

### Description:

### Checklist:

### DIFF
--- a/pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
+++ b/pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
@@ -68,6 +68,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				defer res.Body.Close()
 				if res.StatusCode >= 200 && res.StatusCode < 300 {
 					s1.Verified = true
+				} else if res.StatusCode == 403 {
+					err = fmt.Errorf("unauthorized: HTTP 403 returned %d", res.StatusCode)
+					s1.Verified = false
 				} else if res.StatusCode != 403 {
 					err = fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
 					s1.SetVerificationError(err, resMatch)


dylanTruffle (approved):

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -68,6 +68,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				defer res.Body.Close()
 				if res.StatusCode >= 200 && res.StatusCode < 300 {
 					s1.Verified = true
+				} else if res.StatusCode == 403 {
```
thiggy1342, line 71: We need to explicitly check the status code here, as the above check to see if err == nil won't fail on a 403, as the request itself was made successfully.

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -68,6 +68,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					err = fmt.Errorf("unauthorized: HTTP 403 returned %d", res.StatusCode)
```
shahzadhaider1, line 72: Redundant format string produces: "unauthorized: HTTP 403 returned 403"

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -68,6 +68,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					s1.Verified = false
```
shahzadhaider1, line 73: `s1.Verified` is already `false` by default

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
shahzadhaider1: Thanks for the PR! The existing code actually handles 403 correctly already. When `StatusCode == 403`, neither inner branch executes: we don't set `Verified = true` (correct) and we don't set a verification error (also correct, since 403 is expected for invalid tokens). The `s1.Verified` field defaults to `false`, so no explicit assignment is needed.

Could you share more details about the false positive you encountered? Specifically: timestamps of the scan and token deactivation, and whether there might be multiple tokens involved? That would help us identify the actual root cause because, I just ran a small test with live and revoked tokens and everything seems to be working fine.

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
nabeelalam, line 72: This error is never added to the result and is also unneeded in this case.

TruffleHog only returns errors in cases where an error occurs or the verification result is uncertain.

In the case of a `403 Forbidden` response we determinately set the status of the secret to "unverified", in which case no error is needed to be returned.

nabeelalam:
Hey @thiggy1342, thanks for putting this PR together! I’m going to close it for now since the changes don’t appear to affect the detector’s behavior.

If you’re still seeing the issue, feel free to reopen this PR or [open a new issue](https://github.com/trufflesecurity/trufflehog/issues/new/choose) with details and reproduction steps. Happy to take another look!

PR #4653, opened 2026-01-10
fix: git commit date parsing for non-English locales

### LINKED ISSUE #3338: trufflehog fails to parse localized timestamp
### TruffleHog Version

trufflehog 3.82.5

### Expected Behavior

trufflehog should parse localized timestamps correctly

### Actual Behavior

```
2024-09-26T09:46:14+02:00	error	trufflehog	failed to parse commit date	{"source_manager_worker_id": "wAMwt", "unit": "./terraform_datadog", "unit_kind": "dir", "repo": "git@gitlab.company.com:/some/repo.git", "commit": "3ec15209b4d1c854258f7d9ebb05cbbfbc3562fb", "latestState": "AuthorDateLine", "error": "parsing time \"Mo Sep 28 07:59:21 2020 +0000\" as \"Mon Jan 2 15:04:05 2006 -0700\": cannot parse \"Mo Sep 28 07:59:21 2020 +0000\" as \"Mon\""}
```

This is a german localized timestamp. trufflehog then reports found issues like so 

```
Found unverified result 🐷🔑❓
[...]
Line: 23
Repository: ...
Timestamp: 0001-01-01 00:00:00 +0000
```

## Environment
 * OS: macOS
 * Version 14.7

## Additional Context

my LANG is at de_DE.UTF-8, however, setting this to C or en_US.UTF-8 changes nothing.

I ran this against a local repo with ```... git file:///```as well as against its remote on gitlab with identical results

Maybe I'm forgetting to configure something here, but if so, I failed to find documentation or other issues regarding this.

GLEF1X:
Fixes #3338

### Description:
TruffleHog fails to parse git commit dates when the system locale is non-English (e.g., German outputs `Sa.` instead of `Sat`) preventing users from scanning git repos(all git-based sources like local git repo, github, gitlab, etc). This is because the format that is being used `--date=format:%a %b %d %H:%M:%S %Y %z` is dependent on `strftime(3)` C library function which is locale dependent(uses POSIX LC_ALL, LC_TIME, LANG). Setting `LC_ALL=C` or any other locale on the user's side like `LC_ALL=de_DE.UTF-8 trufflehog ...`  doesn't help because the code overwrites the subprocess environment:
https://github.com/trufflesecurity/trufflehog/blob/4e02afb0159d0de9be6ceaa2064177f1b16f5033/pkg/gitparse/gitparse.go#L257-L273

From [os/exec go doc](https://pkg.go.dev/os/exec#pkg-types):
```go
// Env specifies the environment of the process.
	// Each entry is of the form "key=value".
	// If Env is nil, the new process uses the current process's
	// environment.
	// If Env contains duplicate environment keys, only the last
	// value in the slice for each duplicate key is used.
	// As a special case on Windows, SYSTEMROOT is always added if
	// missing and not explicitly set to the empty string.
	//
	// See also the Dir field, which may set PWD in the environment.
	Env [][string](https://pkg.go.dev/builtin#string)
```

  This means git only receives `GIT_DIR` and no locale settings. Rather than fixing env inheritance or adding `LC_TIME=C`, this PR switches to `--date=iso-strict` which outputs locale-independent ISO 8601 timestamps `(2024-09-28T07:59:21+00:00)`. I think this is a pretty good solution: iso format has been stable since [Git 2.2 (2014)](https://git.vger.kernel.narkive.com/NZZuraya/patch-v3-pretty-add-date-iso-strict-a-strict-iso-8601-date-format), requires no environment manipulation, and uses Go's native `time.RFC3339` parser.
  
Error message when parsing:
```
2026-01-10T00:28:06-05:00	error	trufflehog	failed to parse commit date	{"source_manager_worker_id": "rc4SK", "unit_kind": "dir", "unit": "/var/folders/hw/r5j4bcyd3472ccwjz0klh5840000gn/T/trufflehog-25309-2688385567", "repo": "file:///Users/xxx/xxx/sample-repo", "commit": "5f506baa305831998a2e15aa07cd381a69fde48f", "latestState": "AuthorDateLine", "error": "parsing time \"Sa. Jan. 10 00:11:44 2026 -0500\" as \"Mon Jan 2 15:04:05 2006 -0700\": cannot parse \"Sa. Jan. 10 00:11:44 2026 -0500\" as \"Mon\""}
```

Tested on homebrew git on mac(apple git seems to completely ignoring locales) and linux (debian)
  

### Checklist:

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -22,7 +22,8 @@ import (
 
 const (
 	// defaultDateFormat is the standard date format for git.
-	defaultDateFormat = "Mon Jan 2 15:04:05 2006 -0700"
+	// Uses ISO 8601 format to avoid locale-dependent weekday/month names.
+	defaultDateFormat = time.RFC3339
 
 	// defaultMaxDiffSize is the maximum size for a diff. Larger diffs will be cut off.
 	defaultMaxDiffSize int64 = 2 * 1024 * 1024 * 1024 // 2GB
@@ -235,7 +236,7 @@ func (c *Parser) RepoPath(
 		"log",
 		"--patch", // https://git-scm.com/docs/git-log#Documentation/git-log.txt---patch
 		"--full-history",
-		"--date=format:%a %b %d %H:%M:%S %Y %z",
+		"--date=iso-strict",
 		"--pretty=fuller", // https://git-scm.com/docs/git-log#_pretty_formats
 		"--notes",         // https://git-scm.com/docs/git-log#Documentation/git-log.txt---notesltrefgt
 	}
@@ -278,7 +279,7 @@ func (c *Parser) RepoPath(
 // Staged parses the output of the `git diff` command for the `source` path.
 func (c *Parser) Staged(ctx context.Context, source string) (chan *Diff, error) {
 	// Provide the --cached flag to diff to get the diff of the staged changes.
-	args := []string{"-C", source, "diff", "-p", "--cached", "--full-history", "--diff-filter=AM", "--date=format:%a %b %d %H:%M:%S %Y %z"}
+	args := []string{"-C", source, "diff", "-p", "--cached", "--full-history", "--diff-filter=AM", "--date=iso-strict"}
 
 	cmd := exec.Command("git", args...)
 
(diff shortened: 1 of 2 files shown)


kashifkhan0771 (approved):
🚀

rosecodym (approved):
Thank you for this!

svenXY:
thanks! Any idea when this might be available via homebrew?

Issue #4541, opened 2025-11-05
Datadog not reporting verification endpoint failures


chopkinsmade:
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md) before submitting

### TruffleHog Version

3.90.12

### Trace Output
https://gist.github.com/chopkinsmade/4afd04a69d0b2ea10f264433b90382f5

### Expected Behavior

When using the trufflehog argument `--results=verified,unknown`, Datadog keys that fail to verify due to an endpoint not being reachable should report there was a verification issue, in the same way AWS keys report these failures

### Actual Behavior

When an AWS token is detected and the endpoint used for verification is unavailable, the trufflehog output reports the AWS token as unknown with the message `Verification issue: lookup sts.us-east-1.amazonaws.com: no such host`. When a Datadog token is detected and the endpoint used for verification is unavailable, there is no Datadog entry in the output

### Steps to Reproduce

1. Run the below command, these are not real AWS or Datadog keys:
```
cat << EOF > ./secrets.py
aws_access_key_id = "AKIAQYLPMN5GHHFPXAM1"
aws_secret_access_key = "1rUm636uS2yOEcfP5pvfqJ/ml36mF7AkyHsEU1IU"

DD_API_KEY = "20e3cd747481553d351607eb21397514"
DATADOG_API_KEY = "20e3cd747482553d341607eb21397516"
EOF
```
2. Turn off network connections from your device, this is the easiest way to test the endpoints being unavailable
3. Run `trufflehog filesystem --fail --no-update --log-level=5 --include-detectors=AWS,Datadogtoken --results=verified,unknown ./secrets.py`
4. The output will have an unverified entry for AWS, but nothing for Datadog

## Environment
 * OS: MacOS
 * Version: 26.0.1

PR #4661, opened 2026-01-14
[INS-255] Updated datadog detector to set verificationError in case of a verification error

MuneebUllahKhan222:
Description:
According to this GitHub [issue](https://github.com/trufflesecurity/trufflehog/issues/4541), secrets detected by the DatadogToken detector that failed verification were not being reported to the notifier worker (CLI output).

This PR updates the DatadogToken detector to follow the same verification pattern used by newer detectors by Setting `verificationError` on the `detectors.Result` with the error returned from `verifyMatch` function.
With this change, unverified Datadog token findings are correctly propagated to the notifier worker and reported in CLI output.

This PR also introduces `resetVerificationError` function which reset the `verificationError` field on the results struct required when the secret is verified with any of the endpoint found/configured so as to not report a false verification error.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped to DatadogToken detector verification and result reporting; behavior change is intentional (surface verification errors) with no changes to core auth or data stores.
> 
> **Overview**
> Fixes unverified Datadog API/app key findings not reaching the notifier/CLI by wiring verification through a new **`verifyMatch`** helper and calling **`SetVerificationError`** on each endpoint attempt, matching newer detectors.
> 
> Verification logic moves out of **`FromData`**: responses are classified explicitly (200 vs 401/403 vs other status codes), bodies are drained before close, and **`Scanner`** can use an injectable HTTP client via **`getClient()`**. **`SecretParts["endpoint"]`** is set for every verification try, not only on success.
> 
> Integration tests switch to **`go-cmp`**, expect **`SecretParts`** on the unverified case, and add a focused unverified-token test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57802dcce7d41fdc21864702828d293fcb45fae3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MuneebUllahKhan222:
This PR is a PR stacked on top of [PR #4616 ](https://github.com/trufflesecurity/trufflehog/pull/4616). So we will wait for PR #4616  to be merged first.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -105,12 +116,19 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 
+	var uniqueFoundUrls = make(map[string]struct{})
+	for _, matches := range datadogURLPat.FindAllStringSubmatch(dataStr, -1) {
+		uniqueFoundUrls["https://"+matches[1]] = struct{}{}
+	}
+	endpoints := make([]string, 0, len(uniqueFoundUrls))
+	for endpoint := range uniqueFoundUrls {
+		endpoints = append(endpoints, endpoint)
+	}
+
```
kashifkhan0771, line 128: Why not directly append found urls into a slice?

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -121,64 +139,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					client := s.getClient()
+					res, isVerified, verificationErr := verifyMatch(ctx, client, resApiMatch, resAppMatch, baseURL)
+					s1.Verified = isVerified
+					s1.SetVerificationError(verificationErr, resApiMatch, resAppMatch)
+					if isVerified && res != nil {
+						s1.ResetVerificationError() // Reset verification error in case a secret is verified with an endpoint
+						s1.AnalysisInfo = map[string]string{"apiKey": resApiMatch, "appKey": resAppMatch, "endpoint": baseURL}
+						var serviceResponse userServiceResponse
+						if len(serviceResponse.Data) > 0 {
+							setUserEmails(serviceResponse.Data, &s1)
```
kashifkhan0771: Can you add some empty lines in between and some comments to have better readability?

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -121,64 +139,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
 						}
-					}
-				}
-			}
-			appIncluded = true
-			results = append(results, s1)
-		}
-
```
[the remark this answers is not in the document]
kashifkhan0771: +1

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -121,64 +139,23 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-									setOrganizationInfo(serviceResponse.Included, &s1)
-								}
-							}
+				for _, baseURL := range s.Endpoints(endpoints...) {
+					client := s.getClient()
+					res, isVerified, verificationErr := verifyMatch(ctx, client, resApiMatch, resAppMatch, baseURL)
+					s1.Verified = isVerified
+					s1.SetVerificationError(verificationErr, resApiMatch, resAppMatch)
+					if isVerified && res != nil {
+						s1.ResetVerificationError() // Reset verification error in case a secret is verified with an endpoint
```
kashifkhan0771: `SetVerificationError` only set the error if the error is not `nil` than why we need to reset an error?

kashifkhan0771 (changes requested):

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -196,3 +173,44 @@ func (s Scanner) Type() detectorspb.DetectorType {
...
+		return nil, false, fmt.Errorf("bad request")
+	case http.StatusUnauthorized:
+		return nil, false, fmt.Errorf("invalid credentials")
+	case http.StatusForbidden:
+		return nil, false, fmt.Errorf("Insufficient permissions")
+	case http.StatusTooManyRequests:
+		return nil, false, fmt.Errorf("too many requests")
+	default:
+		return nil, false, fmt.Errorf("unexpected status code: %d", res.StatusCode)
+	}
```
shahzadhaider1, line 206: Do you think we should return an error in case we receive a `401`? I mean, the credential is confirmed invalid, so there shouldn't be any error returned. 

Similarly, should `403` return false with an error? `Insufficient permissions` error means the credentials are valid but lack permission to perform the operation.

Also, we should reduce the cases by merging similar ones. Normally, we have these cases:
1. `valid:true` with no error
2. `valid:false` with no error
3. `valid:false` with error

shahzadhaider1 (changes requested):

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -131,33 +141,22 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-									setUserEmails(serviceResponse.Data, &s1)
-								}
-								// setup organizations
-								if len(serviceResponse.Included) > 0 {
-									setOrganizationInfo(serviceResponse.Included, &s1)
-								}
+					client := s.getClient()
+					res, isVerified, verificationErr := verifyMatch(ctx, client, resApiMatch, resAppMatch, baseURL)
+					s1.Verified = isVerified
+					s1.SetVerificationError(verificationErr, resApiMatch, resAppMatch)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is not a issue, this is what almost all our detectors do.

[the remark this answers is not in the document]
shahzadhaider1: This miiiiight be true. Also, what do you think about the idea of creating separate result per URL?

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -131,33 +141,22 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+					if isVerified {
+						s1.SecretParts["endpoint"] = baseURL
```
shahzadhaider1: `SecretParts["endpoint"]` will not be set for unverified results, right?

MuneebUllahKhan222: yes, works the same way in `datadogapikey` detector

pkg/detectors/datadogtoken/datadogtoken.go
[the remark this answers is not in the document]
MuneebUllahKhan222: The only reason I didn't create a separate result per URL was to preserve the detectors original logic before this PR. 
Should I update it or let it be?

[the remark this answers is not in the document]
shahzadhaider1: I believe we should prioritize the accuracy here.

pkg/detectors/datadogtoken/datadogtoken.go
shahzadhaider1, line 136: This detector supports multiple endpoints (EndpointCustomizer + URLs found in the scanned data). Per our detector guidelines, when a detector can verify against multiple endpoints, the endpoint must be part of `RawV2` and we should emit one result per (api key, app key, endpoint) combination.

Right now we emit a single result per pair and break on the first verified endpoint, so "verified on US, unverified on EU" collapses into one finding and the endpoint info is lost for dedup. Maybe try making the endpoint loop produce a result per iteration with `RawV2 = appKey + apiKey + endpoint`. This also removes the SetVerificationError overwrite issue

shahzadhaider1: But I understand, this was the oldder implementation and updating the RawV2 will orphan the existing secrets, so it's a tough one. Maybe discuss with the broader team?

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -175,3 +174,38 @@ func (s Scanner) Type() detector_typepb.DetectorType {
...
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		var serviceResponse userServiceResponse
+		if err := json.NewDecoder(res.Body).Decode(&serviceResponse); err != nil {
+			return nil, false, err
```
shahzadhaider1: If the API returns 200 but the body fails to decode, we return `(nil, false, err)`. The 200 already proves the credential is live; the decode failure only affects the metadata.

pkg/detectors/datadogtoken/datadogtoken.go
```diff
@@ -131,33 +141,22 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-							if err := json.NewDecoder(res.Body).Decode(&serviceResponse); err == nil {
-								// setup emails
-								if len(serviceResponse.Data) > 0 {
-									setUserEmails(serviceResponse.Data, &s1)
-								}
-								// setup organizations
-								if len(serviceResponse.Included) > 0 {
-									setOrganizationInfo(serviceResponse.Included, &s1)
-								}
+					client := s.getClient()
```
shahzadhaider1: `s.getClient()` doesn't change between iterations, so we can move it above the loops. The local `client` also shadows the package-level `client` var.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: I have updated it to include the `endpoint` for unverified results.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: From what I have observe I think datadog detector is one of the most important detector and orphaning its secret's would be bad. I have included `endpoint` in the `SecretParts` though.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: That's a good point. I have updated it to return `nil, true, nil` in-case we are unable to decode the response.

pkg/detectors/datadogtoken/datadogtoken.go
MuneebUllahKhan222: I have renamed the package level client to `defaultClient` and moved the `s.getClient` call out of loop iterations.

shahzadhaider1 (approved):

shahzadhaider1:
Approving as discussed internally with the team. 
Great work and thanks for bearing with me.

amanfcp (approved):
seems out of scanning scope. 🤠️✅️

kashifkhan0771 (approved):

mariduv (approved):

PR #4666, opened 2026-01-14
Auto-configure TruffleHog for Pre-commit Hooks

kashifkhan0771:
### Description:
This PR enhances TruffleHog to automatically detect when it's running as a pre-commit hook and configure itself with optimal settings. Users no longer need to manually specify multiple command-line flags for common pre-commit hook use cases.

Currently, when running TruffleHog as a pre-commit hook, users must manually specify several flags:
```bash
trufflehog git file://. --since-commit HEAD --fail --trust-local-git-config --results=verified,unknown
```

This PR enables TruffleHog to auto-detects pre-commit hook contexts and automatically applies the recommended configuration:

**Auto-applied settings:**
- `--since-commit HEAD` - Scan only staged changes
- `--fail` - Exit if secrets are found (blocking commit)
- `--trust-local-git-config`
- `--results=verified,unknown`

### Supported Hook Frameworks

1. **pre-commit.com** - Automatically detected via `PRE_COMMIT=1` environment variable
2. **Husky** - Automatically detected via `HUSKY=1` environment variable
3. **Native Git Hooks** - Set `TRUFFLEHOG_PRE_COMMIT=1` environment variable in your hook script

### Checklist:

### DIFF
--- a/pkg/analyzer/analyzers/airtable/common/common.go
+++ b/pkg/analyzer/analyzers/airtable/common/common.go
@@ -12,6 +12,7 @@ import (
 	"github.com/jedib0t/go-pretty/v6/table"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/analyzers"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 )
 
 func CallAirtableAPI(token string, method string, url string) (*http.Response, error) {
@@ -21,7 +22,7 @@ func CallAirtableAPI(token string, method string, url string) (*http.Response, e
 	}
 	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := detectors.DetectorHttpClientWithNoLocalAddresses.Do(req)
 	if err != nil {
 		return nil, err
 	}
@@ -90,7 +91,7 @@ func fetchBaseSchema(token string, baseID string) (*Schema, error) {
 	if !exists {
 		return nil, fmt.Errorf("endpoint for GetBaseSchemaEndpoint does not exist")
 	}
-	url := strings.Replace(endpoint.URL, "{baseID}", baseID, -1)
+	url := strings.ReplaceAll(endpoint.URL, "{baseID}", baseID)
 	resp, err := CallAirtableAPI(token, endpoint.Method, url)
 	if err != nil {
 		return nil, err
--- a/pkg/analyzer/analyzers/client.go
+++ b/pkg/analyzer/analyzers/client.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/config"
+	"golang.org/x/time/rate"
 )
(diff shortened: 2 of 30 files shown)


PreCommit.md
```diff
@@ -160,7 +187,7 @@ In rare cases, you may need to bypass pre-commit hooks:
 git commit --no-verify -m "Your commit message"
 ```
 
-### Running in Audit Mode
```
MuneebUllahKhan222, line 163: In audit mode we should suggest the user to add --trust-local-git-config flag.

main.go
```diff
@@ -1209,3 +1230,40 @@ func validateClonePath(clonePath string, noCleanup bool) error {
 
 	return nil
 }
+
+// isPreCommitHook detects if trufflehog is running as a pre-commit hook
+func isPreCommitHook() bool {
+	// Pre-commit.com framework detection
+	// Docs: https://pre-commit.com/#pre-commit
+	// Sets PRE_COMMIT=1 environment variable when running hooks
+	if os.Getenv("PRE_COMMIT") == "1" {
```
mustansir14, line 1236: Can you link to a specific page in the docs that mentions this environment variable? I couldn't find it.

main.go
```diff
@@ -1209,3 +1230,40 @@ func validateClonePath(clonePath string, noCleanup bool) error {
...
+	}
+
+	// Husky legacy detection (versions < 4.0)
+	// Sets HUSKY_GIT_PARAMS for git hooks, containing commit parameters
+	// Reference: https://github.com/typicode/husky/tree/v0.14.3
+	if os.Getenv("HUSKY_GIT_PARAMS") != "" {
+		return true
+	}
+
+	// Local Git hook detection (non-framework)
```
camgunz, line 1254: **suggestion:** It looks like you get these, if that's helpful:
```
GIT_AUTHOR_DATE=@1768568299 +0100
GIT_EXEC_PATH=/opt/homebrew/opt/git/libexec/git-core
GIT_INDEX_FILE=/Users/charliegunyon/code/hooks/.git/index.lock
```

kashifkhan0771: I experimented with several Git environment variables for hook detection, but none proved reliable:

- GIT_DIR & GIT_INDEX_FILE - These seemed promising initially, but they're not consistently set across all Git operations. Sometimes they're present, sometimes not, making them unreliable for detection.

- GIT_EXEC_PATH - This simply points to where Git's core programs are installed (like /usr/libexec/git-core or Homebrew's path). It's set for any Git command, not specifically for hooks, so it would cause false positives.

The fundamental issue is that Git doesn't provide a dedicated environment variable that says "I'm running as a hook." When our binary runs, it sees whatever environment the calling process provides, which varies based on how Git was invoked.

main.go
kashifkhan0771: The documentation doesn't explicitly mention the `PRE_COMMIT=1` environment variable, but testing confirms it is reliably set when hooks execute. TruffleHog automatically detects this and applies optimal pre-commit settings.

For compatibility, users can always manually specify flags if auto-detection doesn't work in their environment.

Screenshot from 2026-01-19 11-09-00

PR #4667, opened 2026-01-14
Add Postman API monthly request limit metric

shahzadhaider1:
## Summary
Adds a new Prometheus gauge metric `postman_api_monthly_requests_limit` to capture the total monthly API request limit from Postman's `RateLimit-Limit-Month` response header.

## Background
Postman's monthly API rate limit is shared between TruffleHog scanner usage and the customer's own direct API usage. Our existing alert tracks scanner usage (`postman_api_requests`) and remaining requests (`postman_api_monthly_requests_remaining`), but we couldn't distinguish between scanner and customer consumption when alerts fired.

## Changes
- Added `postman_api_monthly_requests_limit` gauge metric to metrics.go
- Record the metric value in `handleRateLimits` from the `RateLimit-Limit-Month` header

This enables more actionable alerting--we can now identify whether high consumption is from our scanner (suggest increasing scan interval) or from the customer's own usage (advise them to reduce direct API calls).

### Checklist:

### DIFF
--- a/pkg/handlers/default.go
+++ b/pkg/handlers/default.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


mustansir14 (approved):
I'm concerned about the fact that `apiMonthlyRequestsLimit` is something that won't change in every request, but we're still setting the metric in every request. However, I don't see any way around this and since this is a very cheap call, I guess we're okay?

shahzadhaider1:
> I'm concerned about the fact that `apiMonthlyRequestsLimit` is something that won't change in every request, but we're still setting the metric in every request. However, I don't see any way around this and since this is a very cheap call, I guess we're okay?

Good point, and yes, you answered it yourself--the limit is unlikely to change within a billing cycle. That said, setting a gauge is essentially a float64 assignment, so the overhead is negligible. I'd lean toward keeping it simple and consistent with how we handle `apiMonthlyRequestsRemaining` (which is also set on every request). Adding conditional logic to skip redundant sets would add complexity without meaningful performance benefit.

PR #4670, opened 2026-01-15
Switch out default HTTP client use in detectors

bradlarsen:
This PR updates a handful of detectors that use `http.DefaultClient` for liveness verification to instead use `detectors.DetectorHttpClientWithNoLocalAddresses`. This latter client avoids SSRF by ignoring redirects and blocking access to non-routing IPs, and also adds HTTP instrumentation.

This PR also fixes a few code nits reported by `staticcheck`.


shahzadhaider1 (approved):

camgunz (approved):
Thanks!

PR #4671, opened 2026-01-16
refactor(source/engine): move link generation and update from engine to source specific functions

amanfcp:
### Changes:
  - Add UpdateLink function pointer to Chunk struct for source-specific link updates
  - Implement generateLink and updateLink functions in GitHub source for all repo types
  - Extract common URL manipulation helpers to pkg/giturl/helpers.go
  - Update engine.go to call chunk.UpdateLink directly instead of UpdateLink function
  - Mark SupportsLineNumbers as deprecated (used for backward compatibility only)
  - Update Git source to pass UpdateLink function pointer from configuration
  - Add unit tests for GitHub link generation and updating

### End Goal & Future Work

Current State (This PR)

- GitHub only: Implements source-specific link generation via function pointers
- Other sources: Fall back to centralized giturl.GenerateLink() / UpdateLink()
- Engine: Checks chunk.SourceUpdateLink != nil first, then falls back to SupportsLineNumbers()

Final State (After All Iterations)

```
// Sources own their link logic
github.Init() → sets UpdateLinkFunc: s.updateLink
gitlab.Init() → sets UpdateLinkFunc: s.updateLink
bitbucket.Init() → sets UpdateLinkFunc: s.updateLink

...

// Engine stays simple
if chunk.SourceUpdateLink != nil {
    updatedLink := chunk.SourceUpdateLink(link, line)
    // Done. No provider detection needed.
}
```

What Gets Removed

- giturl.GenerateLink() - each source implements its own
- giturl.UpdateLinkLineNumber() - replaced by source functions
- SupportsLineNumbers() - replaced by nil check
- Provider detection logic in giturl package

### Note:
We can also remove the `giturl` package entirely and move the helper functions elsewhere.

Once this gets approved, we can iteratively implement this on all git flavored sources and remove the deprecated functions entirely.

### Checklist:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -1183,7 +1183,8 @@ func (e *Engine) processResult(
 	isFalsePositive func(detectors.Result) (bool, string),
 ) {
 	ignoreLinePresent := false
-	if SupportsLineNumbers(data.chunk.SourceType) {
+	if data.chunk.SourceUpdateLink != nil {
+		// Source provides its own link update function (e.g., GitHub)
 		copyChunk := data.chunk
 		copyMetaDataClone := proto.Clone(data.chunk.SourceMetadata)
 		if copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData); ok {
@@ -1191,6 +1192,27 @@ func (e *Engine) processResult(
 		}
 		fragStart, mdLine, link := FragmentFirstLineAndLink(&copyChunk)
 		ignoreLinePresent = SetResultLineNumber(&copyChunk, &res, fragStart, mdLine)
+
+		// Call source-specific update function
+		updatedLink := copyChunk.SourceUpdateLink(link, *mdLine)
+
+		if err := updateMetaDataLink(copyChunk.SourceMetadata, updatedLink); err != nil {
+			ctx.Logger().Error(err, "error setting link")
+			return
+		}
+		data.chunk = copyChunk
+	} else if SupportsLineNumbers(data.chunk.SourceType) {
+		// Fallback to centralized link updating for sources that haven't migrated yet
+		// (GitLab, Bitbucket, Azure Repos, etc.)
+		copyChunk := data.chunk
+		copyMetaDataClone := proto.Clone(data.chunk.SourceMetadata)
+		if copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData); ok {
+			copyChunk.SourceMetadata = copyMetaData
+		}
+		fragStart, mdLine, link := FragmentFirstLineAndLink(&copyChunk)
+		ignoreLinePresent = SetResultLineNumber(&copyChunk, &res, fragStart, mdLine)
+
+		// Use deprecated centralized UpdateLink function
 		if err := UpdateLink(ctx, copyChunk.SourceMetadata, link, *mdLine); err != nil {
 			ctx.Logger().Error(err, "error setting link")
 			return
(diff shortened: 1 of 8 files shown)


amanfcp:
A mismatch was identified between how line number support is defined and how metadata links are handled. SupportsLineNumbers() indicates that 9 source types support line numbers, but updateMetaDataLink() only implements logic for 5 metadata types. As a result, sources like GERRIT are marked as supported but have no corresponding metadata handling, causing them to fall through to the default error case.

pkg/engine/engine.go
```diff
@@ -1183,14 +1183,36 @@ func (e *Engine) processResult(
...
+	} else if SupportsLineNumbers(data.chunk.SourceType) {
+		// Fallback to centralized link updating for sources that haven't migrated yet
+		// (GitLab, Bitbucket, Azure Repos, etc.)
+		copyChunk := data.chunk
+		copyMetaDataClone := proto.Clone(data.chunk.SourceMetadata)
+		if copyMetaData, ok := copyMetaDataClone.(*source_metadatapb.MetaData); ok {
+			copyChunk.SourceMetadata = copyMetaData
+		}
+		fragStart, mdLine, link := FragmentFirstLineAndLink(&copyChunk)
+		ignoreLinePresent = SetResultLineNumber(&copyChunk, &res, fragStart, mdLine)
```
mustansir14, line 1213: Instead of duplicating the logic for this `else if` block, can we add an `||` in the first `if` condition? Then when calling the link update function, we can add a smaller if else check.

pkg/giturl/helpers.go
```diff
@@ -0,0 +1,46 @@
...
+// IsWikiURL returns true if the repository URL is a GitHub wiki.
+func IsWikiURL(repo string) bool {
+	return strings.HasSuffix(repo, ".wiki.git")
+}
+
+// TrimWikiSuffix removes the .wiki.git suffix from a repository URL.
+func TrimWikiSuffix(repo string) string {
+	return strings.TrimSuffix(repo, ".wiki.git")
+}
+
```
mustansir14, line 36: I'm wondering if it makes sense to have these Github-specific helpers in the `giturl` package. If they are only used by the Github source, can we move them there?

pkg/sources/sources.go
```diff
@@ -46,6 +46,12 @@ type Chunk struct {
 
 	// Verify specifies whether any secrets in the Chunk should be verified.
 	Verify bool
+
+	// UpdateLink is a source-specific function for updating line numbers in links.
+	// This function is provided by the source at initialization time and allows
+	// the engine to update links without knowing provider-specific formats.
+	// Can be nil for sources that don't support link updating.
+	SourceUpdateLink func(link string, line int64) string
```
mustansir14, line 54: I'm concerned about the name of the field. I feel like it doesn't clearly explain the purpose. If it specifically deals with line numbers, can we call it `UpdateLinkLineNumber`?

mustansir14 (commented):
Looks good to me overall. A few minor suggestions/concerns.

mustansir14:
Also once this approach is finalized and merged in, we should create tickets to migrate other sources to this new pattern as well, with the purpose of completely removing the generic link generation/updation logic.

amanfcp:
[This](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/sources/sources_test.go) test is failing. Not sure why we have a fixed size match so just putting it out here.

pkg/sources/sources.go
amanfcp: FYI, I intentionally added `Source` prefix here like other source specific fields in this struct. `SourceUpdateLinkLineNumber` seems a bit unnecessarily specific don't you think?

pkg/engine/engine.go
amanfcp: Yeah the duplication is messy but I was just trying to avoid nested ifs and making as few changes as possible in the previous flow (since it's temporary code if this gets approved)

mustansir14: Duplication like this can cause maintenance issues. This block already has one nested if so another one wouldn't increases the complexity level IMO

pkg/sources/sources.go
mustansir14: I just felt that `SourceUpdateLink` is a very generic term and doesn't clearly define what sort of updation will happen. But on the other hand the function signature takes in `line int64` so I guess we're fine...

amanfcp:
Closing this since we're trying to make `Chunk` simpler rather than adding complexity to it.

PR #4677, opened 2026-01-19
[INS-280] Fix Github "repostories" filter does not respect GHES endpoint

mustansir14:
### Description:
When a TruffleHog user attempts to filter on specific repositories within Github using the repositories configuration, the scanner does not respect the endpoint configured and instead defaults to using github.com.

After some investigation, I found that this was because of the fact that the `normalizeRepo` function doesn't respect the `Endpoint` field and always prefixes the repos with `https://github.com`. This causes the `filteredRepoCache` to be incorrectly set. Even though the client is configured to use the endpoint, mismatch with the cache causes the scan to fail.

### Checklist:

### DIFF
--- a/pkg/sources/github/repo.go
+++ b/pkg/sources/github/repo.go
@@ -367,7 +367,7 @@ func (s *Source) normalizeRepo(repo string) (string, error) {
 	if strings.Contains(repo, "/") && !regexp.MustCompile(`^[a-z]+://`).MatchString(repo) {
 		fullURL := "https://github.com/" + repo
 		// If using GitHub Enterprise, adjust the URL accordingly
-		if s.conn != nil && s.conn.Endpoint != "" {
+		if s.conn != nil && s.conn.Endpoint != "" && !endsWithGithub.MatchString(s.conn.Endpoint) {
 			u, err := url.Parse(s.conn.Endpoint)
 			if err != nil {
 				return "", fmt.Errorf("invalid enterprise endpoint: %w", err)


pkg/sources/github/repo.go
```diff
@@ -365,6 +366,15 @@ func (s *Source) normalizeRepo(repo string) (string, error) {
 	// If it's a repository name (contains / but not http), convert to full URL first
 	if strings.Contains(repo, "/") && !regexp.MustCompile(`^[a-z]+://`).MatchString(repo) {
 		fullURL := "https://github.com/" + repo
+		// If using GitHub Enterprise, adjust the URL accordingly
+		if s.conn != nil && s.conn.Endpoint != "" {
+			u, err := url.Parse(s.conn.Endpoint)
+			if err != nil {
+				return "", fmt.Errorf("invalid enterprise endpoint: %w", err)
+			}
+			u.Path = "/" + repo
```
shahzadhaider1, line 376: Is it possible we could get something like `https://github.enterprise.com/api/v3` in the Endpoint param? I believe this code overwrites any existing path in the endpoint. For example:

Endpoint: `https://github.enterprise.com/api/v3`
After `u.Path = "/" + repo`: `https://github.enterprise.com/some-repo`
Lost: `/api/v3`

pkg/sources/github/github_test.go
```diff
@@ -459,6 +459,18 @@ func TestNormalizeRepo(t *testing.T) {
...
+		conn: &sourcespb.GitHub{
+			Endpoint: "https://example.com",
+		},
+	}
+
+	result, err := source.normalizeRepo("org/repo")
+	assert.NoError(t, err)
+	assert.Equal(t, "https://example.com/org/repo.git", result)
+}
+
```
shahzadhaider1, line 523: Let's add cases for different endpoints as well, for example: `https://github.enterprise.com/api/v3`

pkg/sources/github/repo.go
mustansir14: This is intended. We want to have the full URL to the repository in the cache. If the provided endpoint is `https://github.enterprise.com/api/v3`, we would still want `https://github.enterprise.com/some-org/some-repo` in the cache.

mustansir14: The original reason to implement it like this was to avoid human errors like accidentally including a sub-path with the endpoint.

I guess I'm ruling out the possibility of an enterprise server hosted at a specific subpath, as it's pretty uncommon to do that.

shahzadhaider1: I see. Normally, GitHub Enterprise API endpoints often include /api/v3. 
Can we add a comment here stating that the behavior is intentional?

mustansir14: Sure

pkg/sources/github/github.go
```diff
@@ -451,8 +451,8 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 			repoName := repo
 			if strings.Contains(repo, "/") {
 				// Try to extract org/repo name from URL
-				if strings.Contains(repo, "github.com") {
-					parts := strings.Split(repo, "/")
+				if url, err := url.Parse(repo); err == nil {
+					parts := strings.Split(url.Path, "/")
```
shahzadhaider1: ummm, can we not name the variable the same as the package name?

mustansir14: Hmm, thanks for catching this. I'll change it.

pkg/sources/github/github.go
```diff
@@ -449,13 +449,11 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
...
-				if strings.Contains(repo, "github.com") {
-					parts := strings.Split(repo, "/")
-					if len(parts) >= 2 {
-						repoName = parts[len(parts)-2] + "/" + strings.TrimSuffix(parts[len(parts)-1], ".git")
-					}
+			// Try to extract org/repo name from URL
+			if u, err := url.Parse(repo); err == nil {
+				parts := strings.Split(u.Path, "/")
+				if len(parts) >= 2 {
+					repoName = parts[len(parts)-2] + "/" + strings.TrimSuffix(parts[len(parts)-1], ".git")
```
shahzadhaider1: Potential trailing slash bug: if we ever get input like `https://github.com/owner/repo.git/` for example, then this will give us `repo.git/`

camgunz: ➕ I would probably use [`strings.TrimRight`](https://pkg.go.dev/strings#TrimRight) to remove any trailing slashes and then [`path.Base`](https://pkg.go.dev/path#Base) to just get the base name.

mustansir14: I agree. I didn't touch it because I didn't wanna change existing logic, but I guess yeah, this can be made better

pkg/sources/github/repo.go
```diff
@@ -365,8 +366,30 @@ func (s *Source) normalizeRepo(repo string) (string, error) {
...
+		// If using GitHub Enterprise, adjust the URL accordingly
+		if s.conn != nil && s.conn.Endpoint != "" {
+			u, err := url.Parse(s.conn.Endpoint)
+			if err != nil {
+				return "", fmt.Errorf("invalid enterprise endpoint: %w", err)
+			}
+			// we want to remove any path components from the endpoint and just use the host
+			u.Path = "/" + repo
+			fullURL = u.String()
+		}
```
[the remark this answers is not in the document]
shahzadhaider1: is this related to [this](https://github.com/trufflesecurity/trufflehog/pull/4677#discussion_r2715835623)?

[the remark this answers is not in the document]
mustansir14: No, this is different.

camgunz (approved):

shahzadhaider1 (approved):

PR #4681, opened 2026-01-20
[INS-281] Github Bug fix: UnitErr and UnitOK called for the same repo

mustansir14:
### Description:
During the investigation for #4677, I found a bug in Github source enumeration where we were calling both `UnitErr` and `UnitOK` for the same repository. This PR fixes that by continuing the loop after reporting the error. A test is added that reproduces this bug.

### Checklist:


amanfcp (approved):

shahzadhaider1 (approved):
Thanks.

PR #4685, opened 2026-01-21
Optimize the regex pattern in the artifactory access token detector

### LINKED ISSUE #4710: Improve Artifactory API Key Detection
## Description
Improve detection regex for Artifactory API Keys

### Preferred Solution
Update existing regex to match:
```
AKCp[0-9][a-zA-Z0-9]{68}
```

### Additional Context
Artifactory API Keys can have a visible format starting with `AKCp` followed by a digit and 68 additional alphanumeric characters. The current detector may not capture this format. Improving this will enhance TruffleHog's coverage for Artifactory credential detection.

#### References
None currently.

shahzadhaider1:
This PR updates the Artifactory detector to properly detect Artifactory API keys with the `AKCp` prefix format.

## Problem
The existing Artifactory detector was using a very generic regex pattern that detected any 73 or 64 character alphanumeric strings, leading to potential false positives. The detector was not specifically targeting the old Artifactory API key format (`AKCp` prefix).

### 1. **Updated Regex Pattern**
- Changed from generic pattern: `\b([a-zA-Z0-9]{64,73})\b`
- To specific pattern: `\b(AKCp[A-Za-z0-9]{69})\b`
- This matches the standard Artifactory API key format

### 2. **Updated Keywords**
- Added `AKCp` to the keywords list for better pre-filtering

### 3. **Updated Tests**
- **Unit tests**: Updated to match the new `AKCp` prefix pattern
- **Integration tests**: Using mock secrets due to API key deprecation (see note below)

## Important Note on Testing
JFrog deprecated Artifactory API keys (format `AKCp*`) and **disabled the ability to create new API keys at the end of Q3 2024**. 

**Sources:**
- [Pulumi Artifactory Provider Documentation](https://www.pulumi.com/registry/packages/artifactory/api-docs/apikey/)
- [Terraform Artifactory Provider Documentation](https://registry.terraform.io/providers/jfrog/artifactory/latest/docs/resources/api_key)

Since real `AKCp` API keys can no longer be generated:
- Integration tests use **mock secrets** with proper format
- Tests validate **detection** (regex matching) rather than verification
- This is acceptable as the primary goal is to detect leaked secrets in codebases, where old API keys may still exist

### Checklist:
- [x] Updated regex pattern to match `AKCp` prefix format
- [x] Updated keywords for efficient pre-filtering
- [x] Updated unit tests
- [x] Updated integration tests with mock secrets
- [x] Added documentation comments explaining testing limitations
- [x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?


kashifkhan0771 (approved):

mariduv (approved):

pkg/detectors/artifactory/artifactory.go
```diff
@@ -42,7 +42,7 @@ func (Scanner) CloudEndpoint() string { return "" }
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"artifactory", "jfrog.io"}
+	return []string{"artifactory", "jfrog.io", "AKCp"}
```
nabeelalam, line 45: Do we need an `artifactory` keyword? It doesn't seem to be a required prefix

nabeelalam (approved):
Just one suggestion otherwise good to go!

pkg/detectors/artifactory/artifactory.go
shahzadhaider1: The `artifactory` keyword was already present, so I kept it as-is. It’s not meant to indicate a required prefix for the secret itself, but rather to help with contextual matching. In practice, I think Artifactory credentials are often defined or referenced near the term “artifactory” (config files, env vars, comments, etc.), so having it as a keyword can help reduce noise and improve signal.

That said, I’m open to removing it if we think it’s not providing enough value.

PR #4697, opened 2026-01-26
[INS-285] Fix custom detectors line number reporting to match the full regex instead of capture group

mustansir14:
#### Problem
Currently for custom detectors, line number reporting works by looking for the `Raw` value from the detector result in the chunk. This works for most cases, except this specific one: When the provided regex contains a capture group, and the raw value is also present before the actual regex match.

This causes an incorrect line number to be reported.

An example for better explanation:

Regexp: `secret *= *"([^"\r\n]+)"`

Chunk:
```
"mysecret"
// some code
// some code
secret = "mysecret"
```

In this example, the regex matches on Line 4, but the `Raw` value is also present on Line 1, which causes Line 1 to be reported.

#### Solution

We have a `primarySecret` field in the detector's result. It's primary purpose is to indicate the value to determine correct line numbers when multiple regexes are provided. However, we can also use it to solve this problem. by setting to the full match value instead of the `Raw` value. This would result in correct line numbers to be reported. 

Implementing this also requires `primary_regex_name` to be set in the detector config. I've added logic to ensure it is always set.    

### Checklist:


pkg/custom_detectors/custom_detectors.go
```diff
@@ -66,6 +66,9 @@ func NewWebhookCustomRegex(pb *custom_detectorspb.CustomRegex) (*CustomRegexWebh
 		}
 	}
 
+	// Ensure primary regex name is set.
+	ensurePrimaryRegexNameSet(pb)
```
kashifkhan0771, line 70: Since we now always set the primary regex ourselves, is it still necessary to expose this as a user-configurable option and document it?

We should properly document this behavior in code, clearly explaining how and why the primary regex is always used for full matches.

We should explain this in the README as well: going forward, the custom detector line number will point to the full match of the primary regex or, if none is explicitly set, the first regex. This is important because a match may span multiple lines.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -229,14 +232,15 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
 		if len(values) > 1 {
 			secret = values[1]
 		}
 		raw += secret
 
-		// if the match is of the primary regex, set it's value as primary secret value in result
+		// if the match is of the primary regex, set it's full match as primary secret value in result
 		if c.PrimaryRegexName == key {
-			result.SetPrimarySecretValue(secret)
+			result.SetPrimarySecretValue(fullMatch)
```
kashifkhan0771, line 255: Add a detailed comment here why are we using fullMatch instead of secret here.

pkg/custom_detectors/custom_detectors_test.go
```diff
@@ -232,6 +232,25 @@ func TestDetectorPrimarySecret(t *testing.T) {
 	assert.Equal(t, "secret_YI7C90ACY1_yy", results[0].GetPrimarySecretValue())
 }
 
+func TestDetectorPrimarySecretFullMatch(t *testing.T) {
```
kashifkhan0771, line 235: Write a test case for a match which span multiple lines.

kashifkhan0771 (changes requested):

pkg/custom_detectors/custom_detectors.go
mustansir14: The user may want to set a regex other than the first one as the primary regex, so I think this option should still be exposed. 

Totally agree with the documentation part! I'll do that

pkg/custom_detectors/custom_detectors.go
```diff
@@ -229,14 +232,16 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
 		// values[0] contains the entire regex match.
 		secret := values[0]
+		fullMatch := values[0]
 		if len(values) > 1 {
 			secret = values[1]
 		}
 		raw += secret
 
-		// if the match is of the primary regex, set it's value as primary secret value in result
+		// if the match is of the primary regex, set it's full match as primary secret value in result
```
kashifkhan0771:
```
We set the full regex match as the primary secret value.
Reasoning:
The engine calculates the line number using the match. When a primary secret is set, it uses that value instead of the raw secret.
While the secret match itself is sufficient to calculate the line number, the same group match could appear elsewhere in the data.
To avoid ambiguity, we store the full regex match as the primary secret value.
This primary secret value is used only for identifying the exact line number and is not used anywhere else.

Example:
Full regex match: secret = ABC123
Secret (raw): ABC123

In this case, the primary secret value stores the full string `secret = ABC123`,
allowing the engine to pinpoint the exact location and avoid matching redundant occurrences of `ABC123` in the data.
```

pkg/custom_detectors/custom_detectors_test.go
```diff
@@ -232,6 +232,61 @@ func TestDetectorPrimarySecret(t *testing.T) {
...
```
kashifkhan0771, line 235: We need to write a test case in the engine too.
Location: https://github.com/trufflesecurity/trufflehog/blob/main/pkg/engine/engine_test.go#L166

kashifkhan0771: The test case should include the full regex match appearing across multiple lines, while the actual secret exists on only one line. The primary secret value will span multiple lines and should resolve to the line number where the match starts.

I want to confirm how the primary secret value is processed when it contains multi-line values - specifically, how the engine determines and reports the starting line.

mustansir14: Added

kashifkhan0771 (approved):

camgunz (approved):

PR #4699, opened 2026-01-26
Unify false positive/overlap tests

rosecodym:
### Description:
The engine tests include two tests that are _almost_ identical: `TestVerificationOverlapChunkFalsePositive` and `TestRetainFalsePositives`. This PR unifies them into a single, table-driven test. It also deletes some fake detectors they used, because we have a different fake detector that we can use instead.

### Checklist:


pkg/engine/engine_test.go
```diff
@@ -787,137 +787,88 @@ func TestVerificationOverlapChunk(t *testing.T) {
...
+			wantUnverifiedSecretCount: 0,
+		},
+		{
+			name: "overlap, retain false positives",
+			detectors: []detectors.Detector{
+				passthroughDetector{detectorType: detectorspb.DetectorType(-1), keywords: []string{"sample"}},
+				passthroughDetector{detectorType: detectorspb.DetectorType(-2), keywords: []string{"ample"}},
+			},
+			retainFalsePositives:      true,
+			wantUnverifiedSecretCount: 2,
```
mcastorina, line 824: "Overlap" happens here because there are two distinct detector types that both get a keyword match, correct?

Do the keywords need to be different or could they both be "sample"?

pkg/engine/engine_test.go
```diff
@@ -787,137 +787,88 @@ func TestVerificationOverlapChunk(t *testing.T) {
...
-		sources.WithBufferedOutput(defaultOutputBufferSize),
-	}
+			// Arrange: Create and start an engine
+			e, err := NewEngine(ctx, &engineConfig)
+			require.NoError(t, err)
+			e.Start(ctx)
 
-	sourceManager := sources.NewManager(opts...)
+			// Act: Scan the file
+			cfg := sources.FilesystemConfig{Paths: []string{secretsPath}}
```
mcastorina, line 863: I'm a little grumpy we have to scan against an actual filesystem to test this behavior, but there isn't really another way (yet?)

mcastorina (approved):

pkg/engine/engine_test.go
rosecodym: Mm, there might be, but I actually kind of like the high level at which this test operates.

pkg/engine/engine_test.go
rosecodym: I'm not sure, actually! But overlap detection was designed to protect against similar-but-not-identical keywords, so I presume that's why the tests were written that way.

shahzadhaider1 (approved):

casey-tran (approved):
The consolidation of the tests makes sense to me!

casey-tran:
By "different fake detector that we can use", do you mean:
```
detectors: []detectors.Detector{
				passthroughDetector{detectorType: detectorspb.DetectorType(-1), keywords: []string{"sample"}},
				passthroughDetector{detectorType: detectorspb.DetectorType(-2), keywords: []string{"ample"}},
			},
```

rosecodym:
> By "different fake detector that we can use", do you mean:
> 
> ```
> detectors: []detectors.Detector{
> 				passthroughDetector{detectorType: detectorspb.DetectorType(-1), keywords: []string{"sample"}},
> 				passthroughDetector{detectorType: detectorspb.DetectorType(-2), keywords: []string{"ample"}},
> 			},
> ```

Yep, exactly!

PR #4714, opened 2026-01-27
Fix pre-receive hook hangs and missing logs by flushing logs on signal and using CommandContext for git commands

jordanTunstill:
### Description:
Fix  missing log output on timeout,

When TruffleHog times out in pre-receive hooks, it fails to output
diagnostic logs.

Changes:
  logs are visible when process is killed
- Use exec.CommandContext instead of exec.Command for git log/diff
  to ensure processes are killed when context is cancelled

This ensures diagnostic output is captured when git operations block in pre-receive hook environments.
### Checklist:


pkg/gitparse/gitparse.go
```diff
@@ -304,6 +306,9 @@ func (c *Parser) executeCommand(ctx context.Context, cmd *exec.Cmd, isStaged boo
 		return diffChan, err
 	}
 
+	// Set WaitDelay to give the command a grace period to finish before being killed
```
rosecodym: This comment doesn't quite match what I'm seeing in the [docs](https://pkg.go.dev/os/exec#Cmd):
> The WaitDelay timer starts when either the associated Context is done or a call to Wait observes that the child process has exited, whichever occurs first. When the delay has elapsed, the command shuts down the child process and/or its I/O pipes.

That doesn't read like a grace period; it reads like an additional timeout.

An extra timeout looks appropriate, based on the way you've described the problem - but I'm concerned about hardcoding it in a function as generic as `executeCommand`. A caller might not realize that this function has a hard-coded five-second timeout (because Golang callers will probably assume that they can control any timeout using the passed-in `context`), and five seconds might not be the appropriate timeout for _every_ command that a caller might use this function to invoke. Can we add the delay as a parameter instead of hard-coding it? (Or do we not know which command is hanging?)

jordanTunstill: That sounds like a good change. I'll set the default to 5 seconds (unless you have a more appropriate time for a default)!

pkg/gitparse/gitparse.go
```diff
@@ -204,6 +208,14 @@ func WithMaxCommitSize(maxCommitSize int64) Option {
 	}
 }
 
+// WithWaitDelay sets the waitDelay option. This specifies how long to wait after
+// context cancellation before forcefully killing git processes.
+func WithWaitDelay(waitDelay time.Duration) Option {
+	return func(parser *Parser) {
+		parser.waitDelay = waitDelay
+	}
+}
```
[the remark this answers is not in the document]
jordanTunstill: Added a test for wait delay to account for this.

main.go
```diff
@@ -61,6 +61,7 @@ var (
 	results             = cli.Flag("results", "Specifies which type(s) of results to output: verified (confirmed valid by API), unknown (verification failed due to error), unverified (detected but not verified), filtered_unverified (unverified but would have been filtered out). Defaults to verified,unverified,unknown.").String()
 	noColor             = cli.Flag("no-color", "Disable colorized output").Bool()
 	noColour            = cli.Flag("no-colour", "Alias for --no-color").Hidden().Bool()
+	logSync func() error //Package-level variable for sync function
```
rosecodym: Burying this var amongst all the `cli.Flag` vars that surround it is going to make it far more difficult to find than it needs to be. Can you put it either at the beginning or the end? I think it would be ok to take it out of the grouped declaration entirely (declaring it using its own `var` keyword) if gofmt lets you do that.

(That being said, I don't think you need to use a package var for this - read my other comments for details.)

main.go
```diff
@@ -413,6 +415,21 @@ func run(state overseer.State) {
 		} else {
 			logger.Info("cleaned temporary artifacts")
 		}
+
+		// Flush logs with timeout to prevent hanging
```
rosecodym: I'm confused by this, because `sync` is already called: It's deferred on line 367 of this file. If that call is hanging, then the hang should be fixed there, not fixed by putting a timer on an `os.Exit` here. Or, that call should be removed entirely, so that this code is the only sync code. And if you do _that_, is there a way to pass the sync function through the overseer state, rather than relying on a package var? Mutable globals (like package vars) make the code more difficult to understand and maintain because it obscures entanglement.

main.go
jordanTunstill: I'll try removing the package var and passing it as a parameter to run()

main.go
jordanTunstill: Looking at this, and with the advise of the robots, it seems that we'd meed a wrapper struct around overseer.State. Passing it as a parameter to Run still lets us get rid of the package level var; let me know if this works for you. 

Currently, I have two syncs, one for a normal exit, where deferred sync runs, and another for signal-based exits, where the sync is explicit as deferred functions are bypassed by os.Exit().

main.go
```diff
@@ -358,15 +376,17 @@ func main() {
...
 		os.Exit(0)
 	}
 
-	defer func() { _ = sync() }()
 	logFatal := logFatalFunc(logger)
 
 	updateCfg := overseer.Config{
-		Program:       run,
+		Program: func(s overseer.State) {
+
```
shahzadhaider1: unnecessary new line here

shahzadhaider1 (approved):
LGTM

main.go
jordanTunstill: removed this, thank you!

main.go
```diff
@@ -347,6 +346,19 @@ func init() {
...
+	if syncFn == nil {
+		return
+	}
+	done := make(chan struct{})
+	go func() {
+		_ = syncFn()
+		close(done)
+	}()
+	<-done
+}
```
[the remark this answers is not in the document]
jordanTunstill: removed the goroutine since we aren't using the timeout anymore.

main.go
```diff
@@ -387,10 +400,12 @@ func main() {
 	}
 }
 
-func run(state overseer.State) {
-
+func run(state overseer.State, logSync func() error) {
```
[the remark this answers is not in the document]
jordanTunstill: Good bot, addressed in most recent commit

main.go
```diff
@@ -387,10 +394,12 @@ func main() {
...
 	ctx, cancel := context.WithCancelCause(context.Background())
 	defer cancel(nil)
+	defer func() {
+		syncLogs(logSync)
+	}()
```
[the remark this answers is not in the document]
jordanTunstill: removed this

rosecodym (approved):
This seems worth doing, but before you merge can you ensure that the merge commit also mentions all the log syncs you added?

PR #4721, opened 2026-01-29
Add a new NDJSON / JSONL input source

bradlarsen:
This adds a new input source to TruffleHog, accessible via `trufflehog json-enumerator`.

This input source requires a list of filenames, each of which is an NDJSON-formatted sequence of objects that take one of two forms:

Form 1: `{"data": "utf-8 string", "metadata": <non-null JSON value>}`
Form 2: `{"data_b64": "base64-encoded bytestring", "metadata": <non-null JSON value>}`

The `data` / `data_b64` field specifies the content to be scanned. The `metadata` field is arbitrary, and is simply propagated downstream with scan results from the corresponding content.

Note that although `trufflehog json-enumerator` requires a list of filenames to be given, the NDJSON data that you wish to scan may not need to be first written to disk. On Linux and macOS, at least, you can use shell process substitution to set up a named pipe from a producer process, like `trufflehog json-enumerator <(some-program-that-emits-ndjson)`.

---

> [!NOTE]
> **Medium Risk**
> Adds a new ingestion path that parses untrusted NDJSON and threads arbitrary per-record metadata through the scanning pipeline; issues here could lead to scan failures or unexpected memory/throughput behavior on large inputs.
> 
> **Overview**
> Adds a new CLI subcommand, `trufflehog json-enumerator`, that scans one or more NDJSON/JSONL files containing per-line `data` (UTF-8 string) or `data_b64` (bytes) plus required `metadata`.
> 
> Implements a new `json_enumerator` source and engine entrypoint (`ScanJSONEnumeratorInput`) that stream-decodes records, scans their payloads via existing file handling, and attaches the record’s metadata to findings via new protobuf types (`sourcespb.JSONEnumerator` and `source_metadatapb.JSONEnumerator`). Includes generated proto/validation updates and a focused unit test covering string vs bytes payloads and invalid metadata.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4beec34268817bfc4b18774336e98e51b6235ea5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+const SourceType = sourcespb.SourceType_SOURCE_TYPE_JSON_ENUMERATOR
+
+type Source struct {
+	name        string
+	sourceId    sources.SourceID
+	jobId       sources.JobID
+	concurrency int
+	verify      bool
+	paths       []string
+	log         logr.Logger
```
rosecodym: Storing a logger in the source struct like this is an older pattern that we're trying to get away from. `Chunks` is passed a context that contains a logger, so there's no need to keep this one around. (The downside of doing this is that future maintainers get confused about which one they should be using.)

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+const SourceType = sourcespb.SourceType_SOURCE_TYPE_JSON_ENUMERATOR
+
+type Source struct {
+	name        string
+	sourceId    sources.SourceID
+	jobId       sources.JobID
+	concurrency int
```
rosecodym: If you don't use this field, there's no reason to track it. It's fine to ignore the corresponding argument to `Init`.

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+	s.concurrency = concurrency
+
+	s.name = name
+	s.sourceId = sourceId
+	s.jobId = jobId
+	s.verify = verify
+
+	var conn sourcespb.JSONEnumerator
+	if err := anypb.UnmarshalTo(connection, &conn, proto.UnmarshalOptions{}); err != nil {
+		return errors.WrapPrefix(err, "error unmarshalling connection", 0)
```
rosecodym: We're trying to move away from `WrapPrefix` to `fmt.Errorf` (with `%w`), because the latter is in the standard library.

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+			if errors.Is(err, io.EOF) {
+				// enumerator file is done
+				return nil
+			}
+			return err
+		}
+
+		metadataJSON, err := entry.Metadata.MarshalJSON()
+		if err != nil {
+			ctx.Logger().V(2).Error(err, "failed to convert metadata to JSON")
```
rosecodym: We want to log all errors at level 0. (If you've seen any above level 0 elsewhere, we should change them!)

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+	log         logr.Logger
+	sources.Progress
+	sources.CommonSourceUnitUnmarshaller
+}
+
+// Ensure the Source satisfies the interfaces at compile time
+var _ sources.Source = (*Source)(nil)
+var _ sources.SourceUnitUnmarshaller = (*Source)(nil)
+
+// var _ sources.SourceUnitEnumChunker = (*Source)(nil)
```
rosecodym: It looks like you can remove this; if we want to add it back later it will be easy enough to.

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+			SourceMetadata: &source_metadatapb.MetaData{
+				Data: &source_metadatapb.MetaData_JsonEnumerator{
+					JsonEnumerator: &source_metadatapb.JSONEnumerator{
+						Metadata: string(metadataJSON),
+					},
+				},
+			},
+		}
+
+		if err := handlers.HandleFile(ctx, bytes.NewReader(entry.Data), chunkSkel, reporter); err != nil {
```
rosecodym: `handlers.HandleFile` is designed to abstract away the processing of archive files and the like. If you know that your input bytes are the actual bytes you wish to scan, you don't need to use it - you can emit chunks directly. (You're welcome to use it _anyway_, if you want.)

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,191 @@
...
+		return json.Marshal(jsonEntryAux{
+			Metadata: &e.Metadata,
+			DataB64:  &e.Data,
+		})
+	}
+}
+
+// UnmarshalJSON implements custom JSON unmarshaling for envelope.
+// It handles both "content" (string) and "content_b64" (base64-encoded string) fields.
+func (e *jsonEntry) UnmarshalJSON(data []byte) error {
```
rosecodym, line 117: Where is this called?

rosecodym (commented):
This looks like a good start! In addition to my inline questions, I have this one: Are scans of individual _paths_ cancellation-aware? It looks like the source is only cancellation-aware _between_ paths.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: This overrides how the `encoding.json` package unmarshals the `jsonEntry` type:

> To unmarshal JSON into a value implementing [Unmarshaler](https://pkg.go.dev/encoding/json#Unmarshaler), Unmarshal calls that value's [Unmarshaler.UnmarshalJSON] method, including when the input is a JSON null. Otherwise, if the value implements [encoding.TextUnmarshaler](https://pkg.go.dev/encoding#TextUnmarshaler) and the input is a JSON quoted string, Unmarshal calls [encoding.TextUnmarshaler.UnmarshalText](https://pkg.go.dev/encoding#TextUnmarshaler.UnmarshalText) with the unquoted form of the string.

https://pkg.go.dev/encoding/json#Unmarshal

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Thanks for the pointer. I've seen lots of existing TruffleHog code that uses `V(2)`, `V(3)`, etc. :)

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Gotcha; will change! This was copypasta from existing sources.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Gotcha; will change! This was copypasta from existing sources.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Gotcha; will change! This was copypasta from existing sources.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Regarding your question above:

> Are scans of individual paths cancellation-aware? It looks like the source is only cancellation-aware between paths.

It looks like the use of `HandleFile` here for each bytestring to be scanned makes each entry cancellation-aware. Right?

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Well, most of those uses of `V(...)` are with `Info`, not error. But in many cases they are reporting errors as info...

rosecodym: Logging inconsistency is a huge issue right now, to be sure :(

pkg/sources/json_enumerator/json_enumerator.go
rosecodym: Yeah, I guess you're right!

trufflesteeeve (commented):
This looks good to me, however I'm waiting on others to review before I give my own approval.

pkg/sources/json_enumerator/json_enumerator.go
rgmz: Calling `V(...)` with `Error` is most likely a no-op:

> // Error messages do not have a verbosity level and are always logged.
https://github.com/go-logr/logr/blob/5248b7fc60e889bd4b6f1ecd553d8d21da06f3e5/logr.go#L62

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"unicode/utf8"
+
+	"github.com/go-errors/errors"
```
shahzadhaider1: This import is only being used at line 138 `errors.Is(err, io.EOF)`. 
Can we use the standard library "errors" package instead of this external dependency because this also uses the same function under the hood?

pkg/sources/sources.go
```diff
@@ -471,8 +472,15 @@ type ElasticsearchConfig struct {
 	BestEffortScan bool
 }
 
+// StdinConfig defines the configuration for a stdin source.
 type StdinConfig struct{}
 
+// JSONEnumeratorConfig defines the configuration for a JSON enumerator source.
+type JSONEnumeratorConfig struct{
```
shahzadhaider1:
```suggestion
type JSONEnumeratorConfig struct {
```

go.sum
shahzadhaider1, line 1: please resolve the conflicts

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Ensure the Source satisfies the interfaces at compile time
+var _ sources.Source = (*Source)(nil)
+var _ sources.SourceUnitUnmarshaller = (*Source)(nil)
+
+func (s *Source) Type() sourcespb.SourceType { return SourceType }
+func (s *Source) SourceID() sources.SourceID { return s.sourceId }
+func (s *Source) JobID() sources.JobID       { return s.jobId }
+
+func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, sourceId sources.SourceID, verify bool, connection *anypb.Any, concurrency int) error {
```
camgunz: **suggestion:** There are a couple of reeeeeeal long lines (I'm a stickler for this as I use a bunch of vertical splits) and annoyingly `gofmt` totally punts on line wrapping. For things w/ commas I find myself often doing:

```go
func (s *Source) Init(
  aCtx context.Context,
  name string,
  jobId sources.JobID,
  sourceId sources.SourceID,
  verify bool,
  connection *anypb.Any,
  concurrency int,
) error {
```

It's a little wordy but it's one of the few line wrapping things `gofmt` won't undo, and it's a little easier to edit/read. I realize I'm a weirdo about this though; feel free to blow me off (as well as any other suggestions I make :).

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return nil
+}
+
+func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
+	for i, path := range s.paths {
+		if common.IsDone(ctx) {
+			return nil
+		}
+		s.SetProgressComplete(i, len(s.paths), fmt.Sprintf("Path: %s", path), "")
```
camgunz, line 75: **nit:** I think this is _probably_ off by one because current progress is index-based while total progress is count-based.

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+	return nil
+}
+
+type jsonEntry struct {
+	Metadata json.RawMessage
+	Data     []byte
+}
+
+// jsonEntryAux is a helper struct to support marshalling the content to scan as either a UTF-8 string in `content` or a base64-encoded bytestring in `content_base64`.
+type jsonEntryAux struct {
```
camgunz, line 91: Oh man I really like the "-Aux" thing; will steal 👍🏻

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return nil
+}
+
+type jsonEntry struct {
+	Metadata json.RawMessage
+	Data     []byte
+}
+
+// jsonEntryAux is a helper struct to support marshalling the content to scan as either a UTF-8 string in `content` or a base64-encoded bytestring in `content_base64`.
```
camgunz: Comments?! On _data structures_?? Are you sure you work at Truffle??? 🏆

I do like the pattern you have here of like, there's an internal data structure and then you define the marshal/unmarshal interface for the external data structure to take/give the internal one. Kinda wish I would've thought of that....

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+	if aux.Data != nil && aux.DataB64 != nil {
+		return fmt.Errorf("both data and data_b64 present")
+	}
+
+	e.Metadata = *aux.Metadata
+	if aux.DataB64 != nil {
+		e.Data = *aux.DataB64
+	} else {
+		e.Data = []byte(*aux.Data)
+	}
```
camgunz, line 138: **question:** I thought about this a little but my brain sputtered out: do you think there's a significant performance cost to this copying? I _think_ there might be? But if it's not reasonable to avoid then that's that 🤷🏻

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,183 @@
...
+			if errors.Is(err, io.EOF) {
+				// enumerator file is done
+				return nil
+			}
+			return err
+		}
+
+		metadataJSON, err := entry.Metadata.MarshalJSON()
+		if err != nil {
+			ctx.Logger().Error(err, "failed to convert metadata to JSON")
```
camgunz, line 164: **suggestion:** log the entry in a separate logging call at like... level 4 or 5 maybe

camgunz (approved):
Couple of nits and suggestions, but super clean and straightforward--looks great 👍🏻

pkg/sources/json_enumerator/json_enumerator.go
camgunz: Oof, that naming is unfortunate; I would guess most editors/IDEs would automatically add this package if you're using it :(.  ➕ though

bradlarsen: Good catch! Yeah, the `github.com/go-errors/errors` import was added by gopls. Not the import that I intended...

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: All good! I'll fix the long lines.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: Perhaps. I adapted this from what I saw in other sources, and many others use the range-based loop index for the call to `SetProgressComplete`.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: In the first case (`aux.DataB64 != nil`), the value being copied is a byte slice (`[]byte`). It's a shallow copy of 16 bytes, so "essentially free", right?

In the second case, it's a string value being converted to a byte slice. A bit of investigation indicates that this _does_ in general result in a copy of the data being made, since `string` and `[]byte` differ in mutability in Go. That's unfortunate. We might be able to use `unsafe.Slice` + `unsafe.StringData` to do a zero-cost conversion, but we'd have to think hard about the safety of such a thing.

In any case, the serialization format that this input source accepts is JSON, which is very not great in terms of performance. It _is_ very easy to generate, however, from just about any language you want. We can consider a better serialization format for a similar source in the future if performance here becomes a constraint.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: The comment for `SetProgressComplete` suggests that the loop index and length of the array are indeed the values to pass:

https://github.com/trufflesecurity/trufflehog/blob/f2bd590e547d1c6fd9bdbb8913c62e0d5b688fba/pkg/sources/sources.go#L502-L526

pkg/sources/json_enumerator/json_enumerator.go
```diff
@@ -0,0 +1,205 @@
...
+func (s *Source) Chunks(
+	ctx context.Context,
+	chunksChan chan *sources.Chunk,
+	_ ...sources.ChunkingTarget,
+) error {
+	for i, path := range s.paths {
+		if common.IsDone(ctx) {
+			return nil
+		}
+		s.SetProgressComplete(i, len(s.paths), fmt.Sprintf("Path: %s", path), "")
```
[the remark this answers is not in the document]
bradlarsen: This is not a problem with the caller.

pkg/sources/json_enumerator/json_enumerator.go
camgunz: Oh great deep dive, ty; I was hoping the []byte -> string conversion was zero cost. IMO this is a "change it later if we care" thing 👍🏻

pkg/sources/json_enumerator/json_enumerator.go
camgunz: Oh OK; I saw this also in the filesystem source so I mean, at least if it's wrong it's in good company.

pkg/sources/json_enumerator/json_enumerator.go
bradlarsen: I took an even deeper dive this afternoon, hah. I benchmarked JSON deserialization with the above implementation, and _also_ using a zero-copy `unsafe`-based implementation. The latter is a couple percent faster overall; JSON parsing costs dominate and the throughput is capped around 170MB/s in Go on my MBP.

PR #4726, opened 2026-02-03
[INS-307] Added unspecified(0.0.0.0) check to DetectorHttpClientWithNoLocalAddresses

MuneebUllahKhan222:
### Description:
The current` isLocalIP` check used by `DetectorHttpClientWithNoLocalAddresses` does not account for **unspecified IP addresses (0.0.0.0 / ::)**. This can result in the unspecified/wildcard ip (0.0.0.0) to be resolved into local address which can be used for **SSRF attacks**.

This PR updates the local IP filtering logic to treat unspecified IPs as local, preventing attacker-controlled URLs from forcing HTTP requests to internal services during credential verification.

### Checklist:

### DIFF
--- a/pkg/detectors/http_test.go
+++ b/pkg/detectors/http_test.go
@@ -22,6 +22,18 @@ func TestWithNoLocalIP(t *testing.T) {
 		assert.ErrorIs(t, err, ErrNoLocalIP)
 	})
 
+	t.Run("Prevents dialing wildcard IP", func(t *testing.T) {
+		client := &http.Client{}
+		WithNoLocalIP()(client)
+
+		transport, ok := client.Transport.(*http.Transport)
+		assert.True(t, ok, "Expected transport to be *http.Transport")
+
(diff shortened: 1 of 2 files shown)


pkg/detectors/http_test.go
```diff
@@ -22,6 +22,18 @@ func TestWithNoLocalIP(t *testing.T) {
 		assert.ErrorIs(t, err, ErrNoLocalIP)
 	})
 
+	t.Run("Prevents dialing wildcard IP", func(t *testing.T) {
```
camgunz, line 25: **suggestion:** Is it possible to add a check for "::" too? Is it as easy as specifying `tcp6` to `DialContext`?

camgunz (approved):
One suggestion, but approved otherwise 👍🏻

camgunz (approved):

camgunz:
I apparently had this PR open in multiple windows lol

pkg/detectors/http_test.go
MuneebUllahKhan222: Done.

PR #4727, opened 2026-02-03
Fix GitLab project metadata cache and switch to LRU

kashifkhan0771:
### Description:
The in-memory cache for GitLab project metadata (used to populate chunk metadata) was keyed by the raw `HTTPURLToRepo` on write but by the normalized repo URL on read, so cache lookups almost always missed and the underlying map grew unbounded. Replaced the map with an expirable LRU cache (15,000 entries, 1 hour TTL) to bound memory usage, and consistently use the normalized repo URL as the cache key on both writes and reads, including normalizing before cloning in ChunkUnit so the git remote URL used for metadata lookups matches the cached key.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches GitLab scan metadata and enumeration caching across `Chunks`, `ChunkUnit`, and `scanRepos`; wrong keys or TTL behavior could leave `ProjectId`/`ProjectName` empty on findings, but scope is limited to metadata population rather than secret detection logic.
> 
> **Overview**
> Fixes GitLab chunk metadata lookups by **always caching project details under the normalized repo URL** (via `giturl.NormalizeGitlabRepo`) instead of raw `HTTPURLToRepo`, and **normalizing in `ChunkUnit` before clone** so the git remote URL used during chunking matches the cache key.
> 
> Replaces the unbounded in-memory map in `project_cache.go` with a **HashiCorp expirable LRU** (15,000 entries, 1-hour TTL), and **re-fetches metadata in `scanRepos`** before each repo scan when entries may have expired during long runs. Cache hits/misses get clearer logging; a miss while attaching metadata is logged as an error. Tests clear the cache with `Purge()` instead of mutating the old map.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef5bac1d8e1eb51980a68cedaf1a3bf5062a3fa8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/sources/gitlab/gitlab.go
```diff
@@ -1022,77 +1028,6 @@ func (s *Source) WithScanOptions(scanOptions *git.ScanOptions) {
 	s.scanOptions = scanOptions
 }
```
kashifkhan0771, line 1024: These funcs are moved at the end of the file.

pkg/sources/gitlab/gitlab.go
```diff
@@ -242,11 +241,12 @@ func (s *Source) Init(ctx context.Context, name string, jobId sources.JobID, sou
 				Timestamp:           sanitizer.UTF8(timestamp),
 				Line:                line,
 			}
-			proj, ok := s.repoToProjCache.get(repository)
+			// check for project metadata in the cache
+			project, ok := s.projectMetadataCache.get(repository)
```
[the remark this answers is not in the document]
kashifkhan0771: Okay, I’m not sure why the bot is complaining about this. I double-checked, and the log inside the source metadata function confirms that the cache is being used. The results also include the project ID and name.

pkg/sources/gitlab/project_cache.go
```diff
@@ -1,29 +1,47 @@
...
-	cache map[string]*project
+// NewProjectMetadataCache initializes a new project metadata cache
+// with a 30k entry limit and a 60-minute expiration.
+func NewProjectMetadataCache() *projectMetadataCache {
+	return &projectMetadataCache{
+		cache: expirable.NewLRU[string, *project](
+			15000, // upto 15000 entries
+			nil,
+			60*time.Minute, // time-based expiration - 1 hour
+		),
```
mustansir14: I'm interested to know about the thought process that went into deciding these numbers. Is that based on our past experience about the rate at which we scan gitlab projects?

Can there be a possibility of an entry getting expired before we might want to use it?

mustansir14 (commented):
I have some questions

pkg/sources/gitlab/project_cache.go
kashifkhan0771: There isn’t any deep science behind these numbers. They’re rough, initial choices. The 15K limit is something most organizations won’t hit at all. For organizations with more than 15K projects, I think we should be able to process roughly 15K repositories per hour. In practice, it’s very unlikely that a repository would be enumerated and not scanned within an hour. We need to start with baseline numbers, and if we run into issues, we can always adjust them based on observed behavior.

If someone has a strong alternative (though I don’t think we do) for how many repositories we should process per hour, we can start with that number instead.

Using expirable LRU cache so that entries are automatically cleaned up after their TTL via lazy deletion and a background cleanup routine. Additionally, if the cache reaches the 15K entry limit, it will evict the least recently used entries by design so the new inserts are never blocked

mustansir14: I see. Thanks for the explanation :+1:

mustansir14 (approved):

pkg/sources/gitlab/gitlab.go
```diff
@@ -72,16 +72,15 @@ type Source struct {
...
+	clonePath    string
+	noCleanup    bool
 
 	printLegacyJSON bool
 
 	projectsPerPage int64
 
 	// cache of repo URL to project info, used when generating metadata for chunks
-	repoToProjCache repoToProjectCache
+	*projectMetadataCache
```
rosecodym: Why did you chose to embed this instead of making it a normal field?

rosecodym (commented):
What happens if a cache entry _does_ expire? It looks like the returned metadata will be incorrect. Did you consider logging an error or something so that we know if that happens?

kashifkhan0771:
> What happens if a cache entry _does_ expire? It looks like the returned metadata will be incorrect. Did you consider logging an error or something so that we know if that happens?

A cache entry will expire after one hour or in case the cache limit hit 15K without being used. In both cases if cache is not present, the metadata will be empty as previously I believe. I added the error log for cache miss now.

pkg/sources/gitlab/gitlab.go
kashifkhan0771: Ah this was a mistake. I initially thought we’d need this in the git source, so I embedded it to use it directly with Source. That turned out not to be necessary, and I forgot to change it to a regular field.

pkg/sources/gitlab/gitlab.go
```diff
@@ -242,11 +242,16 @@ func (s *Source) Init(ctx context.Context, name string, jobId sources.JobID, sou
...
 			if ok {
-				gitlabMetadata.ProjectId = int64(proj.id)
-				gitlabMetadata.ProjectName = proj.name
-				gitlabMetadata.ProjectOwner = proj.owner
+				ctx.Logger().V(5).Info("cache hit inside source metadata func: found project metadata in the cache", "cache_key", repository)
+				gitlabMetadata.ProjectId = project.id
+				gitlabMetadata.ProjectName = project.name
+				gitlabMetadata.ProjectOwner = project.owner
+			} else {
+				ctx.Logger().Error(errors.New("failed to get repo metadata from cache"),
```
shahzadhaider1, line 268: Error-level logging on cache miss looks too aggressive

pkg/sources/gitlab/project_cache.go
```diff
@@ -1,29 +1,47 @@
...
 
-func (r *repoToProjectCache) set(repo string, proj *project) {
-	r.Lock()
-	defer r.Unlock()
+func (c *projectMetadataCache) set(repo string, proj *project) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
 
-	r.cache[repo] = proj
+	c.cache.Add(repo, proj)
```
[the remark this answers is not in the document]
shahzadhaider1: 👀

pkg/sources/gitlab/gitlab.go
```diff
@@ -1184,30 +1125,37 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
 		}
 	}
 
+	normalizedRepoUrl, err := giturl.NormalizeGitlabRepo(repoURL)
```
shahzadhaider1: contradiction here: Url Vs URL.
Let's be consistent. If I'm not wrong, Go convention is `normalizedRepoURL` here (all-caps acronym)

pkg/sources/gitlab/gitlab.go
```diff
@@ -20,20 +29,12 @@ import (
...
-	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/types/known/anypb"
 )
 
 const SourceType = sourcespb.SourceType_SOURCE_TYPE_GITLAB
 
-// This is the URL for gitlab hosted at gitlab.com
-const gitlabBaseURL = "https://gitlab.com/"
+// Base URL for GitLab Cloud (hosted at gitlab.com)
+const gitlabCloudBaseUrl = "https://gitlab.com/"
```
shahzadhaider1: The old name `gitlabBaseURL` was actually more correct _stylistically_, so I'd suggest we make it `gitlabCloudBaseURL`

pkg/sources/gitlab/project_cache.go
```diff
@@ -1,29 +1,47 @@
...
 
-	cache map[string]*project
+// NewProjectMetadataCache initializes a new project metadata cache
+// with a 30k entry limit and a 60-minute expiration.
+func NewProjectMetadataCache() *projectMetadataCache {
+	return &projectMetadataCache{
+		cache: expirable.NewLRU[string, *project](
+			15000, // upto 15000 entries
+			nil,
+			60*time.Minute, // time-based expiration - 1 hour
```
shahzadhaider1: If a cache entry expires mid-scan (60-min TTL), what will happen? The metadata callback silently produces chunks without ProjectId, ProjectName, and ProjectOwner?

kashifkhan0771: https://github.com/trufflesecurity/trufflehog/pull/4727#issuecomment-3858112795

pkg/sources/gitlab/gitlab.go
kashifkhan0771: I don’t think so. First, this is just a log, we’re not returning an error. Second, I believe we should log an error when we fail to populate the metadata. That would also help us identify whether we need to increase the cache limit or adjust the expiration time.

kashifkhan0771:
If the LRU cache doesn’t fit our use case well and we continue to see many chunks without metadata, my next idea is to implement our own simple.cache which is a wrapper of `patrickmn/go-cache`. Its default expiration time is quite long, so it would serve as a good fallback option.

shahzadhaider1 (approved):

pkg/sources/gitlab/gitlab.go
MuneebUllahKhan222: Could re-write this to 
```go
ctx.Logger().Error(errors.New("cache miss: unable to retrieve repo meta data from cache", "cache_key", info.Repository)
```

pkg/sources/gitlab/gitlab.go
```diff
@@ -260,9 +273,11 @@ func (s *Source) Init(ctx context.Context, name string, jobId sources.JobID, sou
 	}
 	s.git = git.NewGit(cfg)
 
-	s.repoToProjCache = repoToProjectCache{
-		cache: make(map[string]*project),
-	}
+	s.projectMetadataCache = expirable.NewLRU[string, *projectMetadata](
+		15000, // upto 15000 entries
+		nil,
+		60*time.Minute, // time-based expiration - 1 hour
```
MuneebUllahKhan222, line 286: I am not sure but I think this expiry time-limit can be increase since our scans usually run for more than an error.

pkg/sources/gitlab/gitlab.go
kashifkhan0771: It's already added @MuneebUllahKhan222 
```go
ctx.Logger().Error(errors.New("failed to get repo metadata from cache"), "cache miss: not found project metadata in the cache", "cache_key", info.Repository)
```

MuneebUllahKhan222 (approved):
LGTM

pkg/sources/gitlab/gitlab.go
kashifkhan0771: Good catch, but correctness doesn't actually depend on this TTL anymore. There's a fallback: `ensureProjectInCache` runs right before each repo's scan (in scanRepos, same pattern ChunkUnit already used) and refetches from the GitLab API if the entry is missing or expired. So even if the TTL expires mid scan (which it will, for long-running scans), the metadata gets re-fetched right before it's needed.

MuneebUllahKhan222: Have we considered the possibility of the cache expiring between the `ensureProjectInCache` call and the creation of the actual chunk via `s.sourceMetadataFunc` in **git.go**?

If the cache entry expires between those two calls (which seems possible), `sourceMetadataFunc` would return zero values for the metadata fields.

Would it make sense to also call `ensureProjectInCache` from within `sourceMetadataFunc` so it can repopulate the cache if the entry has expired?

MuneebUllahKhan222: We can also do some thing like this
```go
func (s *Source) ensureProjectInCache(ctx context.Context, repoURL string) (*projectMetadata, bool) {
      if project, ok := s.projectMetadataCache.Get(repoURL); ok {
             // return the project metadata from this function
              return project, true
      }
      project, err := s.getGitlabProject(ctx, repoURL)
      if err != nil {
              ctx.Logger().Error(err, "failed to fetch GitLab project", "repo", repoURL)
              return nil, false
      }
      s.cacheGitlabProjectMetadata(ctx, project)
      return s.projectMetadataCache.Get(repoURL)
}
```

and in sourceMetaDataFunc
```go
if project, ok := s.ensureProjectInCache(ctx, info.Repository); ok {
      gitlabMetadata.ProjectId = project.id
      gitlabMetadata.ProjectName = project.name
      gitlabMetadata.ProjectOwner = project.owner
} else {
      ctx.Logger().Error(errors.New("failed to get repo metadata from cache"),
              "cache miss: not found project metadata in the cache", "cache_key", info.Repository)
}
```

kashifkhan0771: Good catch - Done

mariduv (approved):

PR #4731, opened 2026-02-05
Allow logging of caller info

rosecodym:
### Description:
This PR extends our logging wrapper code to allow the addition of automatically generated caller information to log entries. It also adds a way to selectively suppress this caller information for individual sinks. This PR does not change any of trufflehog's behavior - it just enables future functionality in case we want it to use it later.

This PR also provides an exportable type alias for sink configuration option functions so that callers have more flexibility in how they prepare groups of them.

This PR also adds some new tests for the existing `redactionCore` and fixes an existing bug in it. (The bug hasn't yet caused problems, but it created the potential for future problems.)

### Checklist:

### DIFF
--- a/pkg/log/log_cores_test.go
+++ b/pkg/log/log_cores_test.go
@@ -21,7 +21,7 @@ type ignoreEverythingCore struct {
 }
 
 func (t *ignoreEverythingCore) With(fields []zapcore.Field) zapcore.Core {
-	return t.Core.With(fields)
+	return &ignoreEverythingCore{t.Core.With(fields)}
 }
 
 func (t *ignoreEverythingCore) Check(entry zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
--- a/pkg/log/log_cores_test.go
+++ b/pkg/log/log_cores_test.go
@@ -205,6 +205,29 @@ func (ts *TestSuite) TestSuppressCallerCore_RespectsWrappedCheckLogic() {
 	ts.Assert().Empty(msg)
 }
 
+// This test confirms that our custom caller suppression core suppresses caller information.
+func (ts *TestSuite) TestSuppressCallerCore_SuppressesCaller() {
+	// Arrange: Create a testable log sink
+	var buf bytes.Buffer
+	baseCore := zapcore.NewCore(
+		zapcore.NewJSONEncoder(defaultEncoderConfig()),
+		zapcore.Lock(zapcore.AddSync(&buf)),
+		globalLogLevel)
+
+	// Arrange: Set up a core stack
+	core := &suppressCallerCore{baseCore}
+
+	// Arrange: Create a logger
+	logger := zapr.NewLogger(zap.New(core, zap.AddCaller()))
+
+	// Act
+	logger.Info("message")
+
+	// Assert that caller information was suppressed
+	msg := buf.String()
+	ts.Assert().NotContains(msg, "caller")
+}
+
 func BenchmarkLoggerRedact(b *testing.B) {
 	msg := "this is a message with 'foo' in it"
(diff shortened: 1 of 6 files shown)


pkg/log/log_cores_test.go
```diff
@@ -0,0 +1,287 @@
...
+	globalRedactor = ts.oldRedactor
+}
+
+func TestCustomCores(t *testing.T) {
+	suite.Run(t, new(TestSuite))
+}
+
+// This test confirms that when our custom redaction core wraps our custom caller suppression core, both redaction and
+// caller suppression occur.
+func (ts *TestSuite) TestCoreComposition_RedactionWrappingCallerSuppression() {
```
mcastorina, line 57: I'm still new to this test suite concept. Is there a method naming convention to identify it as a test to be ran?

pkg/log/suppress_caller_core.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *suppressCallerCore) Check(ent zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	if !c.Enabled(ent.Level) {
+		return ce
+	}
+
+	// Check to see whether the wrapped core would write, and if so, add this core to the CheckedEntry. We do not pass
+	// the CheckedEntry directly to the wrapped core, because if we do, the wrapped core will probably add itself as a
+	// side effect, which would result in the wrapped core executing its own writes, which would be both duplicative and
+	// without caller suppression.
+	if wrapped := c.Core.Check(ent, nil); wrapped != nil {
```
mcastorina, line 25: Could calling `Check` with a `nil` `CheckedEntry` cause a panic? We don't seem to anticipate that happening in any of our wrapped core implementations.

pkg/log/redaction_core.go
```diff
@@ -17,15 +17,25 @@ func NewRedactionCore(core zapcore.Core, redactor *dynamicRedactor) zapcore.Core
...
+	// Check to see whether the wrapped core would write, and if so, add this core to the CheckedEntry. We do not pass
+	// the CheckedEntry directly to the wrapped core, because if we do, the wrapped core will probably add itself as a
+	// side effect, which would result in the wrapped core executing its own writes, which would be both duplicative and
+	// unredacted.
+	if wrapped := c.Core.Check(ent, nil); wrapped != nil {
 		return ce.AddCore(ent, c)
 	}
+
 	return ce
 }
```
mcastorina, line 35: Were these changes the result of unit tests?

pkg/log/log_cores_test.go
rosecodym: [Yep - a prefix of "Test".](https://github.com/stretchr/testify/blob/65697cefca1d61fba40e001c90d2166be5c8df7f/suite/suite.go#L158)

pkg/log/suppress_caller_core.go
rosecodym: Nope! [Passing a nil CheckedEntry causes zap to grab one out of a pool.](https://github.com/uber-go/zap/blob/16fb16b353f2e27bcd71eba69cbd346b3dcc471a/zapcore/entry.go#L281) The previous core implementation didn't do that...because I didn't realize that it was necessary. (It didn't matter before we started stacking wrapper cores up.)

pkg/log/redaction_core.go
rosecodym: Yep! Tests of the combination of two different wrapper cores surfaced the problem.

pkg/log/suppress_caller_core.go
mcastorina: Neat!

mcastorina (approved):

mariduv (approved):

PR #4733, opened 2026-02-06
[INS-283] Support following symlinks in filesystem source

MuneebUllahKhan222:
### Description:
This PR enables the previously disabled symlink following in filesystem traversal and introduces a configurable flag to control whether symlinks should be followed.

Symlink resolution relies on Go’s `filepath.EvalSymlinks`, which safely evaluates symlinks up to a maximum depth of 255. If this limit is exceeded, `EvalSymlinks` returns an `errSymlinkLoop` (e.g. “EvalSymlinks: too many links”), preventing infinite loops caused by cyclic or deeply nested symlink structures.

In addition to re-enabling symlink following, this change ensures directory scanning remains stable and predictable when encountering:

### Checklist:


camgunz:
@MuneebUllahKhan222  and I synced up on this aaaaaand:
* We think we can get away w/ not using the visited cache for now. Sub-unit resumption avoids a lot of duplicate work already and it'll be nice to have a cache performance improvement as a separate PR in case we need to revert, etc.
* Instead of a flag for following symlinks, we'll let users configure a max symlink depth and track depth across `scanDir` and `scanFile`. The user-configured max depth shouldn't be allowed to be more than 40 or so, and it'll default to 0 to preserve existing behavior

MuneebUllahKhan222:
> @MuneebUllahKhan222 and I synced up on this aaaaaand:
> 
> * We think we can get away w/ not using the visited cache for now. Sub-unit resumption avoids a lot of duplicate work already and it'll be nice to have a cache performance improvement as a separate PR in case we need to revert, etc.
> * Instead of a flag for following symlinks, we'll let users configure a max symlink depth and track depth across `scanDir` and `scanFile`. The user-configured max depth shouldn't be allowed to be more than 40 or so, and it'll default to 0 to preserve existing behavior

To address these changes I am closing this **PR** and raising [a new PR](https://github.com/trufflesecurity/trufflehog/pull/4742).

PR #4735, opened 2026-02-07
Implement CodeOwnersLeveler

mcastorina:
CodeOwnersLeveler is a log leveler that can be dynamically controlled by the owners found in CODEOWNERS. Levels can be independently set for each owner, and logs will be checked based on the owner of the function that emitted them.

### Description:

### Checklist:


mcastorina:
Another idea is to not use codeowners and instead set level by function call tree.

PR #4740, opened 2026-02-11
added rotation on 403s access_refused, this detector considered them indeterminate failures

jordanTunstill:
fix(rabbitmq): treat authentication failures as determinate errors

Return false, nil for authentication failures (both "403" and "ACCESS_REFUSED" together or "username or password
not allowed")
instead of false, err to properly distinguish determinate failures from
indeterminate errors.
### Description:

### Checklist:

### DIFF
--- a/pkg/detectors/rabbitmq/rabbitmq.go
+++ b/pkg/detectors/rabbitmq/rabbitmq.go
@@ -108,7 +108,8 @@ func (s Scanner) verify(url string) (bool, error) {
 	if (strings.Contains(errStr, "403") &&
 		strings.Contains(errStr, "access_refused")) ||
 		strings.Contains(errStr, "username or password not allowed") {
-		return false, err
+		// make secret as rotated
+		return false, nil
 	}
 	return false, err
 }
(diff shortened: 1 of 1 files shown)


pkg/detectors/rabbitmq/rabbitmq.go
```diff
@@ -99,7 +99,19 @@ func (s Scanner) verify(url string) (bool, error) {
...
+	}
+	// Check if this is a determinate authentication failure
+	errMsg := err.Error()
+	has403 := strings.Contains(errMsg, "403")
+	hasAccessRefused := strings.Contains(errMsg, "ACCESS_REFUSED") || strings.Contains(errMsg, "access_refused")
+	hasUsernamePasswordError := strings.Contains(errMsg, "username or password not allowed")
+	
+	if (has403 && hasAccessRefused) || hasUsernamePasswordError {
+		return false, nil
+	}
```
nabeelalam: I'm wondering what the reason is behind requiring both "403" and "access_refused" to be contained in the string?
Wouldn’t the presence of either one be sufficient to conclude the credentials are rotated?

Also nit: minor suggestion, same logic:
```suggestion
	errStr := strings.ToLower(err.Error())

	if (strings.Contains(errStr, "403") &&
		strings.Contains(errStr, "access_refused")) ||
		strings.Contains(errStr, "username or password not allowed") {
		return false, err
	}
```

kashifkhan0771 (approved):

kashifkhan0771 (approved):

nabeelalam (approved):

rosecodym (approved):

PR #4742, opened 2026-02-12
[INS-283] Support following symlinks in filesystem source

MuneebUllahKhan222:
Description:
This PR re-enables symlink following in filesystem traversal and introduces configurable controls for symlink behavior, including:

- **Follow symlinks:** a boolean flag to enable or disable following symlinks.  
- **Max symlink depth:** a limit on how many times a symlink can be resolved to prevent infinite loops. Currently set to **40**.  

Symlink resolution uses Go’s `os.Readlink`, which resolves a single symlink per call. If the number of resolutions exceeds the configured max depth, `resolveSymLink` returns an error.  

Additionally, directory scanning is now robust against symlink loops (e.g., `A/linkToB -> B` and `B/linkToA -> A`). We pass a initial depth of 0 to `scanDir` function and increment it each time `scanDir` is called again through recursion until the depth exceeds the max symlink depth allow. This allows us to avoid the loopy scenario without maintaining a map of visited paths. 

Resume tracking is scoped to the top-level chunk root via a `resumeRoot` parameter passed through recursive `scanDir` calls.
This prevents stale resume entries when scanning symlink-resolved directories and ensures resume state is cleared correctly after `workerPool.Wait()`.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core filesystem traversal and resume tracking; while guarded by a depth limit and tests, it can affect scan coverage/performance and behavior on symlink-heavy filesystems.
> 
> **Overview**
> Enables filesystem scans to *optionally follow symlinks* with a new `--max-symlink-depth` flag (0 disables) and a hard cap of 40 to avoid infinite recursion.
> 
> Plumbs `MaxSymlinkDepth` end-to-end (CLI → `sources.FilesystemConfig` → engine/proto `Filesystem` message → filesystem source) and rewrites filesystem traversal to resolve symlinks via `os.Readlink`, handle symlink-to-file/dir chains, skip non-regular files, and scope/clean resume state for symlink-derived paths via `ClearEncodedResumeContainingId`.
> 
> Adds extensive symlink-focused tests (loops, broken links, depth exceeded, valid symlinks) and updates existing filesystem tests to match the new traversal/resume behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3787e4f77905fea49c380dfe7810df76aa6794d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/sources/filesystem/filesystem.go
```diff
@@ -37,13 +37,18 @@ type Source struct {
...
+	maxSymlinkDepth int
 }
 
 // Ensure the Source satisfies the interfaces at compile time
 var _ sources.Source = (*Source)(nil)
 var _ sources.SourceUnitUnmarshaller = (*Source)(nil)
 var _ sources.SourceUnitEnumChunker = (*Source)(nil)
 
+// max symlink depth allowed
+const DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED = 40
```
camgunz: **issue:** Annoyingly, Go strongly discourages constants like this -- it should be `DefaultMaxSymlinkDepthAllowed`

pkg/sources/filesystem/filesystem.go
```diff
@@ -80,10 +85,46 @@ func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, so
 		return fmt.Errorf("unable to create filter: %w", err)
 	}
 	s.filter = filter
-
+	s.maxSymlinkDepth = int(conn.GetMaxSymlinkDepth())
+	if s.maxSymlinkDepth > DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED {
+		err := fmt.Errorf("Specified symlink depth %v exceeds the allowed max allowed symlink depth of %v", s.maxSymlinkDepth, DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED)
+		return err
+	}
```
camgunz: **suggestion:** Move this check above setting `s.maxSymlinkDepth`. It doesn't matter at all in this case, but it's a good habit to check a value's range before using it--otherwise if someone comes by later and makes this check non-fatal, you won't have an out-of-range value in the `Source`.

pkg/sources/filesystem/filesystem.go
```diff
@@ -80,10 +85,46 @@ func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, so
...
+	if s.maxSymlinkDepth > 0 {
+		s.followSymlinks = true
```
camgunz: **suggestion:** Oh I see. I like wrapping these kinds of things in methods like:

```go
func (s *Source) followSymlinks() bool {
  return s.maxSymlinkDepth > 0
}
```

It codifies the definition in a nice way and doesn't create weirdo struct states--like there's a brief period in time where `s.maxSymlinkDepth` can be > 0, but `s.followSymlinks` is `false`. Again in this case it isn't a big deal, but anyone coming by later and refactoring will have to keep this order in mind. Pulling it out into a method removes the need to order an assignment entirely.

pkg/sources/filesystem/filesystem.go
```diff
@@ -80,10 +85,46 @@ func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, so
 		return fmt.Errorf("unable to create filter: %w", err)
 	}
 	s.filter = filter
-
+	s.maxSymlinkDepth = int(conn.GetMaxSymlinkDepth())
+	if s.maxSymlinkDepth > DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED {
+		err := fmt.Errorf("Specified symlink depth %v exceeds the allowed max allowed symlink depth of %v", s.maxSymlinkDepth, DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED)
```
camgunz: **issue:** Line too loooooooong! `gofmt` won't break long lines and is a curse to us all, but it will let you break them up yourself like:

```go
		err := fmt.Errorf(
      "Specified symlink depth %v exceeds the allowed max allowed symlink depth of %v",
      s.maxSymlinkDepth,
      DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED,
    )
```

I'd also consider using `%d` for the special interpolation characters--best to be as specific as we can.

pkg/sources/filesystem/filesystem.go
```diff
@@ -80,10 +85,46 @@ func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, so
...
+				filepath.Dir(resolvedPath),
+				resolvedFilePath,
+			)
+		}
+		fileInfo, err = os.Lstat(resolvedFilePath)
+		if err != nil {
+			return nil, "", fmt.Errorf("Error in retrieving info: %v", err)
+		}
+		if fileInfo.Mode()&os.ModeSymlink == 0 {
+			ctx.Logger().V(5).Info("Symlink has been resolved", "symlinkPath", symlinkPath, "resolvedPath", resolvedFilePath, "depth", depth+1)
```
camgunz: **issue:** This line is too long too, but specifically we (mostly) have a pattern of:

```go
ctx.Logger().V(5).Info("Symlink has been resolved",
  "symlinkPath", symlinkPath,
  "resolvedPath", resolvedFilePath,
  "depth", depth+1,
)
```

This way the names are paired up with the value nicely.

pkg/sources/filesystem/filesystem.go
```diff
@@ -80,10 +85,46 @@ func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, so
...
+		err := fmt.Errorf("Specified symlink depth %v exceeds the allowed max allowed symlink depth of %v", s.maxSymlinkDepth, DEFAULT_MAX_SYMLINK_DEPTH_ALLOWED)
+		return err
+	}
+	if s.maxSymlinkDepth > 0 {
+		s.followSymlinks = true
+	}
 	return nil
 }
 
+func (s *Source) resolveSymLink(ctx context.Context, symlinkPath string) (os.FileInfo, string, error) {
```
camgunz: **suggestion:** Reading this, it made me think this should be `scanSymlink` in the same vein as like `scanDir` and `scanFile`.

```go
func (s *Source) resolveSymLink(
  ctx context.Context,
  path string,
  chunksChan chan *sources.Chunk,
  workerPool *errgroup.Group,
  depth int,
  rootPath string,
) error {
  depth++
  if depth >= s.maxSymlinkDepth {
    return errors.New("max symlink depth reached")
  }

  path = filepath.Clean(path)

  res, err := os.ReadLink(path)
  if err != nil {
    return fmt.Errorf("readlink error: %w", err)
  }

  if !filepath.IsAbs(res) {
    res = filepath.Join(filepath.Dir(path), res)
  }

  fileInfo, err = os.Lstat(resolvedFilePath)
  if err != nil {
    return fmt.Errorf("lstat error: %w", err)
  }

  if fileInfo.Mode()&os.ModeSymlink == 0 {
    ctx.Logger().V(5).Info("found symlink to symlink",
      "symlinkPath", path,
      "resolvedPath", res,
      "depth", depth+1,
    )

    return scanSymlink(ctx, res, chunksChan, workerPool, depth, rootPath)
  }

  if fileInfo.IsDir() {
    ctx.Logger().V(5).Info("found symlink to dir",
      "symlinkPath", path,
      "resolvedPath", res,
      "depth", depth+1,
    )

    return s.scanDir(ctx, res, chunksChan, workerPool, depth, rootPath)
  }

  ctx.Logger().V(5).Info("found symlink to file",
    "symlinkPath", path,
    "resolvedPath", res,
    "depth", depth+1,
  )

  return s.scanFile(ctx, res, chunksChan)
}
```

It _does_ mean that `scanDir` has to pass depth into this function, but I think it's a little more consistent.

But the main virtue is that we prevent users from slipping past the symlink depth limit. Via `scanDir`, we could be 4-5 levels deep before we find a symlink, but this method doesn't know that, so we might be accidentally going too deep.

pkg/sources/filesystem/filesystem.go
```diff
@@ -100,15 +141,34 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+				logger.Error(err, err.Error())
+				continue
+			}
 		}
 
-		if fileInfo.IsDir() {
-			err = s.scanDir(ctx, cleanPath, chunksChan)
+		if targetInfo.IsDir() {
+			workerPool := new(errgroup.Group)
+			workerPool.SetLimit(s.concurrency)
```
camgunz, line 143: **suggestion:** I know this isn't your code, so feel extra free to ignore me, but it doesn't make sense to have a worker pool for each dir in `s.paths`. There should be a single worker pool used by all dirs in `s.paths`. So, my suggestion is to hoist this up above `for i, path := range s.paths {` at the very top of this method, and then do the normal error handling (`if err := workerPool.Wait(); err != nil { ...`) down on line 174.

pkg/sources/filesystem/filesystem.go
```diff
@@ -100,15 +141,34 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+			initalDepth := 0
```
camgunz: **typo:** `initalDepth` -> `initialDepth`

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,23 +180,18 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-	workerPool := new(errgroup.Group)
-	workerPool.SetLimit(s.concurrency)
-	defer func() {
-		_ = workerPool.Wait()
-		s.ClearEncodedResumeInfoFor(path)
-	}()
-	startState := s.GetEncodedResumeInfoFor(path)
-	resuming := startState != ""
 
+func (s *Source) scanDir(ctx context.Context, path string, chunksChan chan *sources.Chunk, workerPool *errgroup.Group, depth int, rootPath string) error {
```
camgunz: **issue:** I'm... pretty sure we have to do away with `fs.WalkDir`. `scanDir` now recursively calls itself, which means we might scan a folder multiple times. Fortunately, I don't think that's too painful, it'd be something like:

```go
func (s *Source) scanDir(
	ctx context.Context,
	path string,
	chunksChan chan *sources.Chunk,
	workerPool *errgroup.Group,
	depth int,
) error {
	// check if the full path is not matching any pattern in include
	// FilterRuleSet and matching any exclude FilterRuleSet.
	if s.filter != nil && (s.filter.ShouldExclude(path) || !s.filter.Pass(path)) {
		return nil
	}

	startState := s.GetEncodedResumeInfoFor(path)
	resuming := startState != ""

	ctx.Logger().V(5).Info("Full path found is", "fullPath", path)

	entries, err := os.ReadDir(path)
	if err != nil {
		return fmt.Errorf("readdir error: %w", err)
	}

	for _, d := range entries {
		dPath := filepath.Join(path, d.Name())

		if d.Type()&os.ModeSymlink != 0 {
			if err := s.scanSymlink(ctx, dPath, chunksChan, workerPool, depth); err != nil {
				ctx.Logger().Error(err, "error scanning symlink", "path", dPath)
			}
		} else if d.IsDir() {
			if err := s.scanDir(ctx, dPath, chunksChan, workerPool, depth); err != nil {
				ctx.Logger().Error(err, "error scanning dir", "path", dPath)
			}
		} else if resuming {
			// Since we store the resolved file path in encodeResumeInfo
			// so we match the resolved with startState
			if dPath == startState {
				resuming = false
			}
		} else {
			workerPool.Go(func() error {
				if err := s.scanFile(ctx, dPath, chunksChan); err != nil {
					ctx.Logger().Error(err, "error scanning file", "path", dPath)
				}
				s.SetEncodedResumeInfoFor(path, dPath)
			})
		}
	}

	return nil
}

```

camgunz (changes requested):
This is really good! I wrote a bunch on this review in hopes of being super clear, but that doesn't mean you didn't do great work here, just that I don't want to burn a bunch of time on miscommunication. We're almost there, just a liiiiiittle more to do.

pkg/sources/filesystem/filesystem.go
MuneebUllahKhan222: Thanks for the suggestion! I see the idea of hoisting the worker pool to the top level. The reason I kept it per-path is that currently, `s.concurrency` defines the number of goroutines per path explicitly specified by the user in the config.

If we move the worker pool to the top level, `s.concurrency` would apply globally across all paths, meaning that all files from all paths would be scanned one at a time. I think the current approach is more user-friendly: someone providing a single parent folder in the config will have its files scanned concurrently, whereas moving the worker pool to the top level would require them to specify all paths individually to leverage parallelism.

So, the current behavior aligns better with the typical user workflow, even if a global pool might be interesting for other use cases. That said, let me know if you would like me to move it to a top-level worker pool, I’m happy to adjust.

camgunz: Yeah, I mean that is the original working of the code. It's fine to leave it (I do really mean "suggestion" when I use that).

---

(No need to read past here :)

The filesystem source is a good example of a place where we layer multiple `errgroup.Group` or `sync.WaitGroup` instances, but this is never optimal. In the first place, it twists what anyone might think the "concurrency" config parameter means. Let's say I have 8 CPU cores and I set "concurrency" to 4, expecting that half my cores will be left free for me to Netflix. Well, because we're using multiple `errgroup.Group` or `sync.WaitGroup` instances, instance A might use cores 1, 3, 4, 6 and instance B might use cores 1, 2, 4, 6. This is bad for two reasons: you're using cores the user didn't imagine you'd use, and you're also overloading other cores. This is because multiple schedulers like this can't coordinate. Instance B can't ask Instance A what goroutines it's spawned and send more work their way.

This problem exists at a higher level too: oftentimes (probably close to always) in the enterprise product multiple scans are running in the same process. So, even if we only use a single scheduler inside the filesystem source, the Jira source will have its own scheduler, and we'll have the exact same problem.

In practice, this isn't a huge deal. Almost no one knows or cares that the enterprise scanner effectively uses all your cores in a less-than-maximally-efficient way. **But**, some people do know, and they care very much, and fixing it at this point would require a lot of effort. And, (deeply) ironically, the whole reason for supporting symlinks in the filesystem source here is that we don't offer as much control over concurrency as a couple people would like.

pkg/sources/filesystem/filesystem.go
```diff
@@ -37,13 +36,17 @@ type Source struct {
...
+	maxSymlinkDepth int
 }
 
 // Ensure the Source satisfies the interfaces at compile time
 var _ sources.Source = (*Source)(nil)
 var _ sources.SourceUnitUnmarshaller = (*Source)(nil)
 var _ sources.SourceUnitEnumChunker = (*Source)(nil)
 
+// max symlink depth allowed
+const defaultMaxSymlinkDepth = 40
```
camgunz, line 48: Oooooh even made it private, nice 👍🏻

pkg/sources/filesystem/filesystem.go
```diff
@@ -80,10 +83,30 @@ func (s *Source) Init(aCtx context.Context, name string, jobId sources.JobID, so
...
 
+func (s *Source) setMaxSymlinkDepth(conn *sourcespb.Filesystem) error {
+	depth := int(conn.GetMaxSymlinkDepth())
+	if depth > defaultMaxSymlinkDepth {
+		return fmt.Errorf(
+			"specified symlink depth %d exceeds the allowed max of %d",
+			depth,
+			defaultMaxSymlinkDepth,
+		)
+	}
```
camgunz, line 101: Validation ⁉️  Incredible haha 🏆

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +160,135 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-			s.SetEncodedResumeInfoFor(path, fullPath)
-			return nil
-		})
+			if err := s.scanSymlink(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning symlink", "path", entryPath)
+			}
+		} else if entry.IsDir() {
+			ctx.Logger().V(5).Info("Entry found is a directory", "path", entryPath)
+			if err := s.scanDir(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning file", "path", entryPath)
```
camgunz: **(tiny) issue:**: This says "error scanning *file*", but it should say "error scanning *directory*".

pkg/sources/filesystem/filesystem.go
```diff
@@ -251,15 +365,38 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
...
+			initialDepth := 0
+			scanErr = s.scanSymlink(ctx, cleanPath, ch, workerPool, initialDepth, cleanPath)
+			_ = workerPool.Wait()
+			s.ClearEncodedResumeContainingId(cleanPath)
+
+		} else if fileInfo.IsDir() {
+			ctx.Logger().V(5).Info("Root path is a dir", "path", cleanPath)
+			workerPool := new(errgroup.Group)
+			workerPool.SetLimit(s.concurrency)
+			intialDepth := 0
```
camgunz: **(tiny) issue:** typo: `intialDepth` -> `initialDepth`

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +160,135 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-		})
+			if err := s.scanSymlink(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning symlink", "path", entryPath)
+			}
+		} else if entry.IsDir() {
+			ctx.Logger().V(5).Info("Entry found is a directory", "path", entryPath)
+			if err := s.scanDir(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning file", "path", entryPath)
+			}
+		} else {
```
camgunz, line 293: **issue:** I agree w/ Bugbot we should still do a `!entry.Type().IsRegular()` check before calling `scanFile`.

pkg/sources/filesystem/filesystem.go
```diff
@@ -101,13 +124,30 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+		} else if fileInfo.IsDir() {
+			ctx.Logger().V(5).Info("Root path is a dir", "path", cleanPath)
+			workerPool := new(errgroup.Group)
+			workerPool.SetLimit(s.concurrency)
+			initialDepth := 0
+			err = s.scanDir(ctx, cleanPath, chunksChan, workerPool, initialDepth, cleanPath)
+			_ = workerPool.Wait()
+			s.ClearEncodedResumeContainingId(cleanPath)
 		} else {
+			ctx.Logger().V(5).Info("Root path is a file", "path", cleanPath)
```
camgunz: **issue:** Same thing, put an `.IsRegular()` check here too.

pkg/sources/filesystem/filesystem.go
```diff
@@ -101,13 +124,30 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+				// If the file or directory is a symlink but the followSymlinks is disable ignore the path
+				logger.Info("skipping, following symlinks is not allowed", "path", cleanPath)
+				continue
+			}
+			// if the root path is a symlink we scan the symlink
+			ctx.Logger().V(5).Info("Root path is a symlink", "path", cleanPath)
+			workerPool := new(errgroup.Group)
+			workerPool.SetLimit(s.concurrency)
+			initialDepth := 0
+			err = s.scanSymlink(ctx, cleanPath, chunksChan, workerPool, initialDepth, cleanPath)
```
camgunz: **issue:** Since `scanSymlink` won't increment the depth on its own (unless it's also calling `scanSymlink`) we have to make sure and do it whenever we call it, so... probably `initialDepth` should be `1`.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +160,135 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+			ctx.Logger().V(5).Info("Entry found is a symlink", "path", entryPath)
+			if !s.canFollowSymlinks() {
+				// If the file or directory is a symlink but the followSymlinks is disable ignore the path
+				ctx.Logger().Info("skipping, following symlinks is not allowed", "path", entryPath)
+				continue
 			}
-			s.SetEncodedResumeInfoFor(path, fullPath)
-			return nil
-		})
+			if err := s.scanSymlink(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
```
camgunz, line 285: **issue:** Same deal here, the `depth` argument should be `depth+1`.

pkg/sources/filesystem/filesystem.go
```diff
@@ -251,15 +365,38 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
...
+				// If the file or directory is a symlink but the followSymlinks is disable ignore the path
+				logger.Info("skipping, following symlinks is not allowed", "path", cleanPath)
+				return
+			}
+			// if the root path is a symlink we scan the symlink
+			ctx.Logger().V(5).Info("Root path is a symlink", "path", cleanPath)
+			workerPool := new(errgroup.Group)
+			workerPool.SetLimit(s.concurrency)
+			initialDepth := 0
+			scanErr = s.scanSymlink(ctx, cleanPath, ch, workerPool, initialDepth, cleanPath)
```
camgunz: **issue:** Same deal w/ `depth`.

pkg/sources/filesystem/filesystem.go
```diff
@@ -101,13 +124,30 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+				continue
+			}
+			// if the root path is a symlink we scan the symlink
+			ctx.Logger().V(5).Info("Root path is a symlink", "path", cleanPath)
+			workerPool := new(errgroup.Group)
+			workerPool.SetLimit(s.concurrency)
+			initialDepth := 0
+			err = s.scanSymlink(ctx, cleanPath, chunksChan, workerPool, initialDepth, cleanPath)
+			_ = workerPool.Wait()
+			s.ClearEncodedResumeContainingId(cleanPath)
```
camgunz: **suggestion:** You could put a call to `ClearEncodedResumeContainingId` inside all the `scan*` functions as like `defer s.ClearEncodedResumeContainingId(path)`

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +160,135 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-			ctx.Logger().Error(err, "error walking directory")
-			return nil
-		}
+func (s *Source) scanSymlink(
+	ctx context.Context,
+	path string,
+	chunksChan chan *sources.Chunk,
+	workerPool *errgroup.Group,
+	depth int,
+	rootPath string,
```
camgunz, line 173: **issue:** I actually think we should get rid of `rootPath`. We probably needed it when using `fs.WalkDir`, but we should be passing clean, absolute paths around always. If we need a basename, that's what `path.Base` is for.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +160,135 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+		"depth", depth,
+	)
+	if s.filter != nil && !s.filter.Pass(resolvedPath) {
+		return nil
+	}
+	workerPool.Go(func() error {
+		if err := s.scanFile(ctx, resolvedPath, chunksChan); err != nil {
+			ctx.Logger().Error(err, "error scanning file", "path", resolvedPath)
 		}
+		s.SetEncodedResumeInfoFor(rootPath+path, resolvedPath)
```
camgunz: Yeah then here we can call `SetEncodedResumeInfoFor(path.Dir(path), path.Base(path))` (I guess we'd need a different variable name but whatever 😹)

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +160,135 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	return nil
+}
+
+func (s *Source) scanDir(
+	ctx context.Context,
+	path string,
+	chunksChan chan *sources.Chunk,
+	workerPool *errgroup.Group,
+	depth int,
+	rootPath string,
```
camgunz, line 248: Yeah, same stuff for `rootPath` here too.

pkg/sources/filesystem/filesystem.go
camgunz: **issue:** This is what made me start thinking about `rootPath`, when we add `rootPath` and `path` together, at least some of the time we'll be adding two identical paths together e.g. `/home/charliegunyon/secrets/home/charliegunyon/secrets` 😢

camgunz (changes requested):
This is awesome! I just had a couple issues but they're all really small. Almost there haha :))

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +162,136 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+			ctx.Logger().V(5).Info("Entry found is a symlink", "path", entryPath)
+			if !s.canFollowSymlinks() {
+				// If the file or directory is a symlink but the followSymlinks is disable ignore the path
+				ctx.Logger().Info("skipping, following symlinks is not allowed", "path", entryPath)
+				continue
 			}
-			s.SetEncodedResumeInfoFor(path, fullPath)
-			return nil
-		})
+			if err := s.scanSymlink(ctx, entryPath, chunksChan, workerPool, depth); err != nil {
```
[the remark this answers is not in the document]
MuneebUllahKhan222: In our design, depth tracks symlink-to-symlink resolution hops, not the act of encountering a symlink from a directory. We initialize depth to 1, and the first directory → symlink resolution is allowed without incrementing depth.

Depth is incremented only when a symlink resolves to another symlink. This ensures that with maxSymlinkDepth = 1, exactly one symlink resolution is permitted, while chained symlinks are correctly blocked.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +162,141 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	if err != nil {
+		return fmt.Errorf("readdir error: %w", err)
+	}
+
+	for _, entry := range entries {
+		entryPath := filepath.Join(path, entry.Name())
+		if s.filter != nil && !s.filter.Pass(entryPath) {
+			if !entry.IsDir() && entry.Type()&os.ModeSymlink == 0 {
+				continue
+			}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is expected behviour. We expect the config to contain path that symlink is resolved to in-order to filter rather than containing symlink path.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +163,141 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is expected behviour. We expect the config to contain path that symlink is resolved to in-order to filter rather than containing symlink path.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +163,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
-			s.SetEncodedResumeInfoFor(path, fullPath)
-			return nil
-		})
+			if err := s.scanSymlink(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning symlink", "path", entryPath)
+			}
+		} else if entry.IsDir() {
+			ctx.Logger().V(5).Info("Entry found is a directory", "path", entryPath)
+			if err := s.scanDir(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning directory", "path", entryPath)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Depth is for tracking the number of times a symlink is resolved. So that is why we don't increment the depth when `scanDir` call itself recursively. We only increment depth after resolving symlink.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +163,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	if err != nil {
+		return fmt.Errorf("readdir error: %w", err)
+	}
+
+	for _, entry := range entries {
+		entryPath := filepath.Join(path, entry.Name())
+		if s.filter != nil && !s.filter.Pass(entryPath) {
+			if !entry.IsDir() && entry.Type()&os.ModeSymlink == 0 {
+				continue
+			}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is expected behviour. We expect the config to contain path that symlink is resolved to in-order to filter rather than containing symlink path.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +163,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
 			}
-			s.SetEncodedResumeInfoFor(path, fullPath)
-			return nil
-		})
+			if err := s.scanSymlink(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
+				ctx.Logger().Error(err, "error scanning symlink", "path", entryPath)
+			}
+		} else if entry.IsDir() {
+			ctx.Logger().V(5).Info("Entry found is a directory", "path", entryPath)
+			if err := s.scanDir(ctx, entryPath, chunksChan, workerPool, depth, rootPath); err != nil {
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is intended

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +163,147 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+			if entryPath == startState {
 				resuming = false
 			}
-			return nil
-		}
-
-		workerPool.Go(func() error {
-			if err = s.scanFile(ctx, fullPath, chunksChan); err != nil {
-				ctx.Logger().Error(err, "error scanning file", "path", fullPath, "error", err)
+		} else if entry.Type()&os.ModeSymlink != 0 {
```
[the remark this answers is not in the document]
MuneebUllahKhan222: It is because the path stored in startState has already been processed and we process the files/dirs after it.

pkg/sources/filesystem/filesystem.go
```diff
@@ -120,61 +163,149 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
...
+	if err != nil {
+		return fmt.Errorf("readdir error: %w", err)
+	}
+
+	for _, entry := range entries {
+		entryPath := filepath.Join(path, entry.Name())
+		if s.filter != nil && !s.filter.Pass(entryPath) {
+			if !entry.IsDir() && entry.Type()&os.ModeSymlink == 0 {
+				continue
+			}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is expected behviour. We expect the config to contain path that symlink is resolved to in-order to filter rather than containing symlink path.

camgunz (approved):

PR #4744, opened 2026-02-13
Base64 decoding depth assessment

dxa4481:
### Description:
This PR introduces an iterative decoding pipeline, allowing decoders (e.g., Base64, UTF-16) to chain their outputs. Previously, decoders ran independently on the original chunk, missing secrets hidden behind layered encoding (e.g., base64 within UTF-16, or double-base64-encoded values).

The `scannerWorker` now re-runs decoders on any new output, up to a configurable `--max-decode-depth` (default 5). This enables detection of secrets like GCP service accounts and private keys found within base64-encoded Docker auth configs, or Artifactory tokens within base64. The pipeline includes an early exit, ensuring negligible performance overhead for higher depths when no new decoded data is produced (typically <5% overhead compared to depth 1).

### Checklist:

---
<p><a href="https://cursor.com/background-agent?bcId=bc-6dbccc70-d36a-467f-9e7f-7810f3ed517f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-6dbccc70-d36a-467f-9e7f-7810f3ed517f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a></p>

---

> [!NOTE]
> **Medium Risk**
> Changes the core scanning/decoding loop and can increase detector work per chunk depending on input, which may impact performance and result volume despite early-exit/deduping.
> 
> **Overview**
> Adds an *iterative decoding* pipeline in `engine.scannerWorker` so decoded outputs are fed back through all decoders up to a configurable depth, enabling detection in layered encodings (e.g., base64-in-base64, base64-inside-UTF16). A new `--max-decode-depth` flag (default `5`) is plumbed through `main.go` into `engine.Config`/`Engine`, with safeguards for depth < 1 and early-exit plus deduping of repeated decoded outputs (and skipping the `PLAIN` decoder beyond the first pass).
> 
> Updates tests to cover chained decoding behavior at different depths, removes the per-decoder `decode_latency` Prometheus metric, and adds docs benchmarking the performance tradeoffs of `--max-decode-depth`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9f5341b07c0326afbc8dfe1a2162eccfb6a2502c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -153,6 +153,12 @@ type Config struct {
 
 	VerificationResultCache  verificationcache.ResultCache
 	VerificationCacheMetrics verificationcache.MetricsReporter
+
+	// MaxDecodeDepth is the maximum number of iterative decoding passes per chunk.
+	// When a decoder transforms data, all decoders are re-run on the output up to this limit.
+	// 1 = single pass (no chaining), 2+ = chained (e.g., base64 inside UTF-16).
+	// Default: 5.
+	MaxDecodeDepth int
 }
 
 // Engine represents the core scanning engine responsible for detecting secrets in input data.
@@ -221,6 +227,8 @@ type Engine struct {
 	notificationWorkerMultiplier int
 	// verificationOverlapWorkerMultiplier is used to calculate the number of verification overlap workers.
 	verificationOverlapWorkerMultiplier int
+
+	maxDecodeDepth int
 }
 
 // NewEngine creates a new Engine instance with the provided configuration.
@@ -245,6 +253,7 @@ func NewEngine(ctx context.Context, cfg *Config) (*Engine, error) {
 		detectorWorkerMultiplier:            cfg.DetectorWorkerMultiplier,
 		notificationWorkerMultiplier:        cfg.NotificationWorkerMultiplier,
 		verificationOverlapWorkerMultiplier: cfg.VerificationOverlapWorkerMultiplier,
+		maxDecodeDepth:                      cfg.MaxDecodeDepth,
 	}
 	if engine.sourceManager == nil {
 		return nil, fmt.Errorf("source manager is required")
@@ -352,6 +361,10 @@ func (e *Engine) setDefaults(ctx context.Context) {
 
 	if e.verificationOverlapWorkerMultiplier < 1 {
 		e.verificationOverlapWorkerMultiplier = 1
+	}
+
+	if e.maxDecodeDepth < 1 {
+		e.maxDecodeDepth = 1
 	}
 
 	// Default decoders handle common encoding formats.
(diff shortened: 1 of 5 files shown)


pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
+		currentInputs := []decodeInput{{data: chunk.Data}}
+		var seen [][]byte
+
+		for depth := 0; depth < maxDepth; depth++ {
+			var nextInputs []decodeInput
+
+			for _, input := range currentInputs {
+				for _, decoder := range e.decoders {
+					if depth > 0 && decoder.Type() == detectorspb.DecoderType_PLAIN {
+						continue
```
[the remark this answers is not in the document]
dxa4481: Every decoder's output is already valid UTF-8, so utf8.Valid() would return true, and the PLAIN decoder would just pass it through unchanged. The extractSubstrings normalization can never trigger on chained decoder output.

Running PLAIN at depth > 0 would just create duplicate detector work (it always returns non-nil) with zero benefit. The skip is correct.

kashifkhan0771 (approved):

pkg/engine/engine.go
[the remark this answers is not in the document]
rosecodym: This is very hard to figure out if you're not already familiar with it (which is probably why Cursor made the mistake it did), so it definitely calls for a detailed explanatory comment.

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
-			matchingDetectors := e.AhoCorasickCore.FindDetectorMatches(decoded.Chunk.Data)
-			if len(matchingDetectors) > 1 && !e.verificationOverlap {
-				wgVerificationOverlap.Add(1)
-				e.verificationOverlapChunksChan <- verificationOverlapChunk{
-					chunk:                       *decoded.Chunk,
-					detectors:                   matchingDetectors,
-					decoder:                     decoded.DecoderType,
-					verificationOverlapWgDoneFn: wgVerificationOverlap.Done,
+		// Iterative decoding: apply all decoders, then re-decode any new output
+		// until no transformations occur or maxDepth is reached.
```
rosecodym: What you've effectively written is a cycle-detecting graph traversal, which is a pretty complicated thing to drop into a function this (nominally) simple, especially as uncommented as it is, and especially given how intertwined you've made it with the channel write logic. I think it would be far easier to understand if it were extracted into a first-class section of code built clearly for this specific purpose, e.g.
```
func decodeRecursively(data []byte, decoders []decoder.Decoder, int maxDepth) []decodedChunk
{
    ...
}
```
This would (more) properly separate the concerns of "decoding" and "channel writing," which would be a great improvement to clarity in this section of code. The downside (compared to this PR's current implementation) is the loss of online channel writes, but given the small size of the typical decoded chunk set, and the scanner's asynchronous architecture, I'd be astonished if that had a noticeable performance impact.

pkg/engine/engine_test.go
```diff
@@ -1686,3 +1686,104 @@ func TestEngine_VerificationOverlapWorker_DetectableChunkHasCorrectVerifyFlag(t
 		}
 	})
 }
+
+func TestEngine_IterativeDecoding(t *testing.T) {
```
rosecodym, line 1691: Thanks for the tests!

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
 	verificationOverlapWgDoneFn func()
 }
 
+// decodeInput is a chunk of data waiting to be decoded at a given depth.
+type decodeInput struct {
+	data []byte
+}
```
camgunz: **suggestion:** I like the additional semantic information that this is (potentially) encoded data. It would be cool to really build out that mental model by doing a few more small things like:
* rename this to something like `encodedInput` or `encodedData`
* add a `decodedOutput` or `decodedData` struct, indicating data that's been through > 0 rounds of decoding
* anywhere you're using `[]byte` use one of those instead

A small nit either way: single-member structs often can be simple type aliases, a la:

```go
type decodeInput []byte
```

I don't think there's any memory benefit to this, it's just a little clumsy to always have to reach in for the `.data` member--like there's nothing else you could possibly want from this thing, so why isn't it just the thing.

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
+
 func (e *Engine) scannerWorker(ctx context.Context) {
 	var wgDetect sync.WaitGroup
 	var wgVerificationOverlap sync.WaitGroup
 
+	maxDepth := e.maxDecodeDepth
+	if maxDepth < 1 {
+		maxDepth = 1
+	}
```
camgunz: **question:** Can this happen? I'd assume by the time we're here we've already clamped `e.maxDecodeDepth` in `setDefaults`. Or, maybe if it can happen we should somehow ensure it can't?

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
-				e.verificationOverlapChunksChan <- verificationOverlapChunk{
-					chunk:                       *decoded.Chunk,
-					detectors:                   matchingDetectors,
-					decoder:                     decoded.DecoderType,
-					verificationOverlapWgDoneFn: wgVerificationOverlap.Done,
+		// Iterative decoding: apply all decoders, then re-decode any new output
+		// until no transformations occur or maxDepth is reached.
+		// The PLAIN decoder is skipped at depth > 0 (it's a passthrough).
+		currentInputs := []decodeInput{{data: chunk.Data}}
+		var seen [][]byte
```
camgunz: **note:** yeah see here we're pretty freely mixing `decodeInput` and `[]byte`, kinda negating all the nice semantic/type information we set up initially. I would say just settling on `[]byte` is fine tbh; my guess is the additional type information will be more annoying than helpful.

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
+				for _, decoder := range e.decoders {
+					if depth > 0 && decoder.Type() == detectorspb.DecoderType_PLAIN {
+						continue
+					}
+
+					decodeStart := time.Now()
+					chunkCopy := *chunk
+					chunkCopy.Data = input.data
+					decoded := decoder.FromChunk(&chunkCopy)
+					decodeTime := time.Since(decodeStart).Microseconds()
```
camgunz: **note:** It does seem like you intend this, but just to call it out this also measures the time it takes to copy the data, which probably won't fluctuate that much and would just add a relatively static amount of time to the duration.

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
+					decodeLatency.WithLabelValues(decoder.Type().String(), chunk.SourceName).Observe(float64(decodeTime))
+
+					if decoded == nil {
+						continue
+					}
+
+					decoderType := decoded.DecoderType
```
camgunz: **suggestion:** I think it would be fine to just reference this manually down on 854 and 866 -- this made me think there was some extra use when there isn't

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
+
+					// Enqueue for next depth before dispatching (must run
+					// regardless of the verification-overlap branch below).
+					if depth+1 < maxDepth &&
+						decoder.Type() != detectorspb.DecoderType_PLAIN &&
+						!bytes.Equal(decoded.Chunk.Data, input.data) &&
+						!byteSliceSeen(seen, decoded.Chunk.Data) {
```
camgunz: **issue:** I think we should just use [`slices.Contains`](https://pkg.go.dev/slices#Contains) here.

pkg/engine/engine.go
camgunz: I had the same feeling but for a different reason: there's a fair amount of bookkeeping (swapping byte slices, tracking depth, tracking seen) that would be really nice to unit test, but because it's all contained inside this single method we can't. If this functionality were as encapsulated as verification overlap (or decoding itself) it would be easier to test and maintain.

pkg/engine/engine.go
```diff
@@ -775,48 +788,92 @@ type verificationOverlapChunk struct {
...
+					for _, detector := range matchingDetectors {
+						decoded.Chunk.Verify = e.shouldVerifyChunk(sourceVerify, detector, e.detectorVerificationOverrides)
+						wgDetect.Add(1)
+						e.detectableChunksChan <- detectableChunk{
+							chunk:    *decoded.Chunk,
+							detector: detector,
+							decoder:  decoderType,
+							wgDoneFn: wgDetect.Done,
+						}
+					}
```
camgunz: **question:** Do we want to be doing this on chunks that might only be partially decoded? I would have guessed these steps would have come entirely after.

camgunz (changes requested):
Just some notes and suggestions. My only requested change is to use `slices.Contains` instead of a custom finding function.

pkg/engine/engine.go
dylanTruffle: It was the same in the original code (the copy happened between decodeStart and decoder.FromChunk before too).

pkg/engine/engine.go
dylanTruffle: A secret might be visible at depth 1 (single base64 decode) but get mangled at depth 2 (if deeper decoding corrupts it). Each depth level represents a valid view of the data. The original code ran detectors on every decoder's output too...

Realistically I'd love if we could case study this stuff and measure it in the real world to prove how big of an edge case it is

pkg/engine/engine.go
camgunz: Ghost of TruffleHog past haunts us always haha 👻

pkg/engine/engine.go
camgunz: Oh interesting. OK then. I do agree it would be pretty cool to have metrics around this, but that seems like a separate body of work.

camgunz (approved):
LGTM!

PR #4760, opened 2026-02-23
[INS-318] Add New Relic License Key Detector

mustansir14:
### Description:
This PR adds the New Relic License Key Detector.

**Regex**: `\b(([0-9a-f]{32}|eu01xx[0-9a-f]{26})FFFFNRAL)\b`
- US region keys are 40 characters hexadecimal strings ending with `FFFFNRAL`
- EU region keys have the same format but first 6 characters are `eu01xx`

**Verification**:
For verification, we use the metrics API: `https://metric-api.newrelic.com/metric/v1`. We send a POST request with an empty body to this endpoint. A response code of `202` means the key is valid. `403` means it is an invalid/rotated key.
Note: For EU region keys, the host should be `https://metric-api.eu.newrelic.com/`

**Corpora Test**:
image
 
image

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Isolated new detector following existing patterns; verification uses a read-only empty POST with no change to core scan logic.
> 
> **Overview**
> Adds a **New Relic license key** detector so scans can find US/EU ingest keys (hex strings ending in `FFFFNRAL`, with EU keys prefixed by `eu01xx`).
> 
> When verification is on, keys are checked via a POST to the regional Metric API (`metric-api.newrelic.com` or `metric-api.eu.newrelic.com`); **202** marks valid and **403** invalid, with an empty body so nothing is ingested. Results include a **region** hint in extra data.
> 
> Wiring follows other gated detectors: new `DetectorType_NewRelicLicenseKey` (1064), registration in default detector list, `NewRelicLicenseKeyDetectorEnabled` enabled in `main`, and unit/integration tests for regex and live verification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4b609a44553b2e0b201d86be0ca2a3ad484c0d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/newreliclicensekey/newreliclicensekey.go
```diff
@@ -0,0 +1,115 @@
...
+func (s Scanner) getClient() *http.Client {
+	if s.client != nil {
+		return s.client
+	}
+
+	return defaultClient
+}
+
+// Keywords are used for efficiently pre-filtering chunks.
+func (s Scanner) Keywords() []string { return []string{"ffffnral"} }
```
shahzadhaider1, line 41: Have we considered adding `eu01xx` to the keywords list?

shahzadhaider1 (approved):

pkg/detectors/newreliclicensekey/newreliclicensekey.go
mustansir14: I don't think that's necessary because all keys (US and EU) will always contain `ffffnral`.

pkg/detectors/newreliclicensekey/newreliclicensekey.go
```diff
@@ -0,0 +1,119 @@
...
+			Redacted:     resMatch[:8] + "...",
+			SecretParts:  map[string]string{"key": resMatch},
+		}
+
+		if verify {
+			isVerified, extraData, verificationErr := s.verify(ctx, resMatch)
+			s1.Verified = isVerified
+			s1.ExtraData = extraData
+			if extraData != nil {
+				s1.SecretParts["region"] = extraData["region"]
```
unsmith: This works, but `region` is only set if verify is true.  This can be known from the key prefix itself; is there value in extracting this beforehand and setting region for unverified results?

unsmith (approved):
Looks fine, with one possible improvement.

pkg/detectors/newreliclicensekey/newreliclicensekey.go
MuneebUllahKhan222: Yeah it make sense to do region extraction outside `verify == true` block. I have implemented this change.

PR #4761, opened 2026-02-23
Box Detector: Extract Subject ID for Analyzer Integration

shahzadhaider1:
### Problem
The Box analyzer requires three credentials to run: `client_id`, `client_secret`, and `subject_id` (a numeric Box enterprise or user ID). The detector was only verifying `client_id` + `client_secret` and not populating `SecretParts` at all, meaning the analyzer would always fail immediately with `subject_id not found`, making the detector -> analyzer pipeline non-functional.

Additionally, `SecretParts` must now be populated for all results regardless of verification status, so that credential data is always captured and available for downstream use.

### Changes
**`boxoauth.go`**
- Added `subjectIdPat` regex to extract numeric Box enterprise/user IDs from surrounding context near relevant keywords (`enterprise`, `enterprise_id`, `user_id`, `subject`, `box_subject`)
- Added `verifySubjectID` function that mirrors the analyzer's CCG (Client Credentials Grant) authentication flow: tries `box_subject_type=enterprise` first, then falls back to `box_subject_type=user`, to verify that a discovered subject ID actually works with the credential pair
- `SecretParts` is now always populated with `client_id` and `client_secret` for every result, regardless of verification status
- After a credential pair is verified via the existing `verifyMatch`, each discovered subject ID is tested via `verifySubjectID`. The first subject ID that successfully authenticates is added to `SecretParts` alongside `client_id` and `client_secret`
- One `detectors.Result` is emitted per credential pair (not per subject ID), preserving the original detector behavior
- `verifyMatch` signature simplified from `(bool, map[string]string, error)` to `(bool, error)` since `ExtraData` was always `nil`; the existing `400/unauthorized_client` verification trick is preserved as-is
- Added `getClient()` helper to consolidate client resolution

**`boxoauth_test.go`**
- Added pattern tests covering: single subject ID, multiple subject IDs, and no subject ID; all produce one result per credential pair
- Added `gock`-based verification tests covering four scenarios:
  - Verified pair + valid subject ID -> `SecretParts` populated with all three values
  - Verified pair + no subject ID in chunk -> `SecretParts` contains `client_id` + `client_secret` only
  - Verified pair + invalid subject ID -> `SecretParts` contains `client_id` + `client_secret` only
  - Unverified pair + subject ID present -> `Verified` false, `SecretParts` contains `client_id` + `client_secret` only

### Behavior Summary
| Scenario | Results emitted | SecretParts |
|---|---|---|
| Valid pair, valid subject ID found | 1 verified result | `client_id` + `client_secret` + `subject_id` |
| Valid pair, multiple subject IDs (one valid) | 1 verified result | `client_id` + `client_secret` + first valid `subject_id` |
| Valid pair, no valid subject ID | 1 verified result | `client_id` + `client_secret` (analyzer won't run without `subject_id`) |
| Valid pair, no subject ID in chunk | 1 verified result | `client_id` + `client_secret` (analyzer won't run without `subject_id`) |
| Invalid pair | 1 unverified result | `client_id` + `client_secret` |

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Adds additional parsing and a second-stage online verification flow (extra Box API calls) and changes emitted `SecretParts` keys, which may affect downstream consumers expecting prior fields.
> 
> **Overview**
> Updates the Box OAuth detector to **always populate structured credentials** via `SecretParts` (`client_id`/`client_secret`) and to **extract candidate numeric `subject_id`s** from nearby keywords.
> 
> When verification is enabled and the client credentials validate, it now attempts a real client-credentials grant using each discovered subject ID (enterprise then user) and records the first working `subject_id` in the result.
> 
> Tests are expanded/updated to cover the new `SecretParts` shape, subject ID extraction scenarios, and mocked verification outcomes (including valid/invalid subject IDs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 786524e4e2ac1d2cbd8fb5a3adfd82738036f928. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mustansir14 (commented):
A suggestion:

Instead of having an `if else` flow based on subject IDs found and duplicating results creation logic for both the flows, you can do something similar to how organization IDs are passed to `AnalysisInfo` in the [atlassian detector](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/atlassian/v2/atlassian.go)

The idea is basically to add a dummy (empty) entry for the subject ID to the `uniqueMatch` map if there are no subject IDs found, and treat it normally as any other credential part by looping over that. This will simplify the detector logic a lot IMO.

pkg/detectors/boxoauth/boxoauth.go
```diff
@@ -52,34 +63,44 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+				// partial data would just cause downstream auth errors.
+				if isVerified {
+					for subjectId := range uniqueSubjectIdMatches {
+						if verifySubjectID(ctx, s.getClient(), resIdMatch, resSecretMatch, subjectId) {
+							s1.SecretParts = map[string]string{
+								"client_id":     resIdMatch,
+								"client_secret": resSecretMatch,
+								"subject_id":    subjectId,
+							}
+							break
```
mustansir14: Are you sure that omitting `SecretParts` in unverified results would not cause issues downstream? I'm not very familiar with how it's handled but as far as I remember, the future goal is to totally remove `Raw` and `RawV2` and use `SecretParts`

pkg/detectors/boxoauth/boxoauth.go
```diff
@@ -52,34 +63,44 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-				DetectorType: detector_typepb.DetectorType_BoxOauth,
+				DetectorType: s.Type(),
 				Raw:          []byte(resIdMatch),
-				SecretParts: map[string]string{
-					"id":     resIdMatch,
-					"secret": resSecretMatch,
-				},
-				RawV2: []byte(resIdMatch + resSecretMatch),
+				RawV2:        []byte(resIdMatch + resSecretMatch),
 			}
```
mustansir14, line 91: I think we should have `SecretParts` for unverified results as well. They can only include the ID and secret.

mustansir14 (changes requested):

pkg/detectors/boxoauth/boxoauth.go
shahzadhaider1: I didn't populate SecretParts because unverified results cannot be analyzed

pkg/detectors/boxoauth/boxoauth.go
shahzadhaider1: Analyze will only work when:
1. We have a `subject_id`.
2. All three credential parts are valid: `client_id`, `client_secret`, and `subject_id`.

pkg/detectors/boxoauth/boxoauth.go
shahzadhaider1: Imagine a secret has its `AnalysisInfo`/`SecretParts` populated, the system sees that and shows an option `Analyze` on the UI. The user clicks it, and nothing happens because those were unverified secrets or the `subject_id` was missing, and the analyzer returned an error.

pkg/detectors/boxoauth/boxoauth.go
mustansir14: I meant issues in enterprise. According to this [doc](https://github.com/trufflesecurity/thog/blob/5b1c8713d3a7ef1561a36600be85b60acb5d563b/docs/secret-storage/multipart-credentials.md) by @mcastorina, future plan is to totally replace `Raw` and `RawV2` with `SecretParts`

pkg/detectors/boxoauth/boxoauth.go
mustansir14: That's a valid point, but as I said in my other comment, `SecretParts` is not just used for Analyze. Future plan is to totally remove `Raw` and `RawV2` and use `SecretParts`.

~Also, currently all detectors populate `SecretParts` for unverified results as well. So I think there must be a logic somewhere that only runs analyze if the result is `Verified`.~

I see that the results can be marked Verified even without a `subject_id`. I agree that in this case there may be a problem with how we present the Analyze option to the user. I think this is a corner case that was probably missed when the `SecretParts` approach was designed. Something worth discussing with the team.

shahzadhaider1: I have updated it to populate `SecretParts` even when the credentials are unverified. We now always include `client_id` and `client_secret`, and conditionally add `subject_id` when available

mustansir14 (approved):

camgunz (approved):

PR #4767, opened 2026-02-25
Fix overlap worker silently dropping non-duplicate secrets

dylanTruffle:
When a chunk matched multiple detectors and one result was flagged as a cross-detector duplicate via likelyDuplicate(), the entire detector was removed from detectorKeysWithResults. This silently dropped any other non-duplicate results from later match spans of that same detector.

Because detector iteration order comes from a Go map (non-deterministic), which detector got deleted changed between runs, causing unique secrets to randomly appear or disappear across scans of the same data.

The fix tracks specific duplicate results in an overlapEmitted map keyed by (DetectorKey, raw secret value). The detector remains in the pipeline for full processing by detectChunk, which skips only the results already emitted by the overlap worker via the new overlapSecrets field on detectableChunk. Detectors that had overlaps are also sent to detectChunk with Verify = false to avoid redundant verification calls and preserve the original performance characteristics. This gives exact-once reporting for both duplicate and non-duplicate results with no silent data loss and no performance regression.

More info in the md

### Description:

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core scan-engine result/verification behavior in overlap scenarios; low surface area but could affect deduping and verification decisions for chunks with multi-detector matches.
> 
> **Overview**
> Fixes non-deterministic secret loss in the verification overlap pipeline when a chunk matches multiple detectors: instead of removing an entire detector after one cross-detector duplicate, the worker now tracks *which specific raw secrets* were already emitted as overlaps and lets the detector continue through normal processing.
> 
> `detectableChunk` carries an `overlapSecrets` set from `verificationOverlapWorker` to `detectChunk`, which filters out only those already-emitted overlap findings (while also forcing verification off for detectors that had overlaps), ensuring stable, exact-once reporting without dropping other unique results. Adds a markdown doc describing the bug, reproduction, and fix.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 89b1fa75577ef8a0d873823dd105c8d52ddbd911. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -773,10 +773,11 @@ func (e *Engine) ScanChunk(chunk *sources.Chunk) {
 
 // detectableChunk is a decoded chunk that is ready to be scanned by its detector.
 type detectableChunk struct {
-	detector *ahocorasick.DetectorMatch
-	chunk    sources.Chunk
-	decoder  detectorspb.DecoderType
-	wgDoneFn func()
+	detector       *ahocorasick.DetectorMatch
+	chunk          sources.Chunk
+	decoder        detectorspb.DecoderType
+	wgDoneFn       func()
+	overlapSecrets map[string]struct{}
 }
 
 // verificationOverlapChunk is a decoded chunk that has multiple detectors that match it.
@@ -992,6 +993,8 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 	chunkSecrets := make(map[chunkSecretKey]struct{}, avgSecretsPerDetector)
 
 	for chunk := range e.verificationOverlapChunksChan {
+		overlapEmitted := make(map[ahocorasick.DetectorKey]map[string]struct{})
+
 		for _, detector := range chunk.detectors {
 			isFalsePositive := detectors.GetFalsePositiveCheck(detector.Detector)
 
@@ -1057,10 +1060,13 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 							isFalsePositive,
 						)
 
-						// Remove the detector key from the list of detector keys with results.
-						// This is to ensure that the chunk is not reprocessed with verification enabled
-						// for this detector.
-						delete(detectorKeysWithResults, detector.Key)
+						// Track this specific secret as already emitted so detectChunk
+						// can skip it, but keep the detector in detectorKeysWithResults
+						// so its other non-duplicate results are still processed.
+						if _, ok := overlapEmitted[detector.Key]; !ok {
+							overlapEmitted[detector.Key] = make(map[string]struct{})
+						}
+						overlapEmitted[detector.Key][string(val)] = struct{}{}
 					}
 					chunkSecrets[key] = struct{}{}
 				}
@@ -1069,12 +1075,17 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 
 		for _, detector := range detectorKeysWithResults {
 			wgDetect.Add(1)
-			chunk.chunk.Verify = e.shouldVerifyChunk(chunk.chunk.Verify, detector, e.detectorVerificationOverrides)
+			verify := e.shouldVerifyChunk(chunk.chunk.Verify, detector, e.detectorVerificationOverrides)
+			if _, hadOverlap := overlapEmitted[detector.Key]; hadOverlap {
+				verify = false
+			}
+			chunk.chunk.Verify = verify
 			e.detectableChunksChan <- detectableChunk{
-				chunk:    chunk.chunk,
-				detector: detector,
-				decoder:  chunk.decoder,
-				wgDoneFn: wgDetect.Done,
+				chunk:          chunk.chunk,
+				detector:       detector,
+				decoder:        chunk.decoder,
+				wgDoneFn:       wgDetect.Done,
+				overlapSecrets: overlapEmitted[detector.Key],
 			}
 		}
 
@@ -1180,6 +1191,15 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
 		}
 
 		for _, res := range results {
+			if data.overlapSecrets != nil {
+				raw := string(res.Raw)
+				if res.RawV2 != nil {
+					raw = string(res.RawV2)
+				}
+				if _, ok := data.overlapSecrets[raw]; ok {
+					continue
+				}
+			}
 			e.processResult(ctx, res, data.chunk, data.decoder, data.detector.Detector.Description(), isFalsePositive)
 		}
 	}
--- a/docs/bugfix-overlap-worker-detector-deletion.md
+++ b/docs/bugfix-overlap-worker-detector-deletion.md
@@ -1,0 +1,84 @@
+# Bug Fix: Overlap Worker Silently Drops Non-Duplicate Secrets
+
+## Summary
+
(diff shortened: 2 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -1057,10 +1060,13 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 							isFalsePositive,
 						)
 
-						// Remove the detector key from the list of detector keys with results.
-						// This is to ensure that the chunk is not reprocessed with verification enabled
-						// for this detector.
-						delete(detectorKeysWithResults, detector.Key)
+						// Track this specific secret as already emitted so detectChunk
```
camgunz, line 1063: **issue:** When we're in this section, we need to make sure we backtrack and mark any previous detectors as also having dupes.

pkg/engine/engine.go
```diff
@@ -1057,10 +1060,13 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
...
+						// can skip it, but keep the detector in detectorKeysWithResults
```
camgunz, line 1064: **issue:** Also in this section, we should remove the call to `e.processResult`, otherwise we'll process the result twice (this logic has to move to `detectChunk`).

pkg/engine/engine.go
```diff
@@ -1069,12 +1075,17 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 
 		for _, detector := range detectorKeysWithResults {
 			wgDetect.Add(1)
-			chunk.chunk.Verify = e.shouldVerifyChunk(chunk.chunk.Verify, detector, e.detectorVerificationOverrides)
+			verify := e.shouldVerifyChunk(chunk.chunk.Verify, detector, e.detectorVerificationOverrides)
+			if _, hadOverlap := overlapEmitted[detector.Key]; hadOverlap {
+				verify = false
```
camgunz, line 1080: **suggestion:** I would move the map membership check before the call to `shouldVerifyChunk`; it short-circuits the `shouldVerifyChunk` call (we won't call it if it's an overlap) and it's ever so slightly faster

pkg/engine/engine.go
```diff
@@ -1180,6 +1191,15 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
...
 
 		for _, res := range results {
+			if data.overlapSecrets != nil {
+				raw := string(res.Raw)
+				if res.RawV2 != nil {
+					raw = string(res.RawV2)
+				}
+				if _, ok := data.overlapSecrets[raw]; ok {
+					continue
+				}
```
camgunz, line 1201: **issue:** We still want to call `e.processResult` on the result, but we first have to call `res.SetVerificationError(errOverlap)`.

pkg/engine/engine.go
```diff
@@ -773,10 +773,11 @@ func (e *Engine) ScanChunk(chunk *sources.Chunk) {
...
 type detectableChunk struct {
-	detector *ahocorasick.DetectorMatch
-	chunk    sources.Chunk
-	decoder  detectorspb.DecoderType
-	wgDoneFn func()
+	detector       *ahocorasick.DetectorMatch
+	chunk          sources.Chunk
+	decoder        detectorspb.DecoderType
+	wgDoneFn       func()
+	overlapSecrets map[string]struct{}
```
camgunz, line 780: **suggestion:** I would consider making this a slice and using `slices.Contains`. It's a little counterintuitive, but map lookups are a lot more expensive than searching through arrays until the array is like, hundreds of members long (depending on the key data, anyway) and it seems like performance matters in the verification overlap worker (I don't know for sure)

pkg/engine/engine.go
```diff
@@ -992,6 +993,8 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 	chunkSecrets := make(map[chunkSecretKey]struct{}, avgSecretsPerDetector)
 
 	for chunk := range e.verificationOverlapChunksChan {
+		overlapEmitted := make(map[ahocorasick.DetectorKey]map[string]struct{})
```
camgunz, line 996: **suggestion:** Again in line w/ performance/allocation stuff here, you can store these as a slice like `overlap` and preallocate it to `avgSecretsPerDetector`, and clear it (like `overlap = overlap[:0]`) with the rest of the cleanups at the end of this function

pkg/engine/engine.go
```diff
@@ -1069,12 +1075,17 @@ func (e *Engine) verificationOverlapWorker(ctx context.Context) {
 
 		for _, detector := range detectorKeysWithResults {
 			wgDetect.Add(1)
-			chunk.chunk.Verify = e.shouldVerifyChunk(chunk.chunk.Verify, detector, e.detectorVerificationOverrides)
+			verify := e.shouldVerifyChunk(chunk.chunk.Verify, detector, e.detectorVerificationOverrides)
+			if _, hadOverlap := overlapEmitted[detector.Key]; hadOverlap {
+				verify = false
+			}
+			chunk.chunk.Verify = verify
```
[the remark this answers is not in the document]
camgunz: I think this is right, so probably the fix is to... add a `verify` member to `detectableChunk` and use that instead of `chunk.verify`.

camgunz (changes requested):
I'm writing a bigger thing here to overcommunicate; working w/ the engine is fun!

So, I super want some kind of automated test for this: it's engine code, it'll affect downstream customers, etc. But as you pointed out, it's non-deterministic, so we can't test the *thing* itself. But fortunately for us, there are some existing tests that actually will change if we get this thing right. So, everything I'm asking for here is in service of getting the testing in place.

We run the detect/verify/process pipeline on a random detector. But, not only does that impact verification counts, it also undermines verification overlap detection: we have no way of knowing which detector is malicious, so we can't run verification on **any** detector with a dupe--the first one in a randomly sorted list isn't somehow blessed.

This means if we detect a dupe, we have to mark all detectors with the dupe, including detectors we've already seen. But the interface provided by `likelyDuplicate` is insufficient for this: it will tell you that we've already seen a result for the current detector, but it won't tell you which other detectors have also seen it.

This PR has the same issue. While it addresses verification count flappiness by ensuring detectors with dupes still run the detect/process (not verify) pipeline, the results from the first detector also shouldn't be verified. I think this necessitates a change to `likelyDuplicate` to return the key of the detector w/ the dupe. There's some tests around this, but they're easy to update. Once we have that, we can backtrack through the list of detectors and mark them as having the dupe as well.

Next, we'll actually process a chunk w/ `e.processResult` multiple times if we detect an overlap. That's easy enough code to remove (a test caught this for me when I was running locally, fortunately). This also means we need to make sure we're calling `SetVerificationError` in `detectChunk` on a result if it was an overlap.

Finally, in `TestEngine_VerificationOverlapWorker_DetectableChunkHasCorrectVerifyFlag` around line 1633 (Cody's "CMR" comment), we can fix this test now: we should be shipping the same number of chunks as detectors (each detector should ship a chunk), and none of those chunks should be verified.

pkg/engine/engine.go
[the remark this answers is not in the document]
rosecodym: I actually wrote that up in #4558 but never got around to merging it - lemme spruce it up and get it in.

PR #4768, opened 2026-02-25
fix: make LDAP verification context-aware

mariduv:
I forked go-ldap/ldap from someone else's PR that had partial work to add contexts, then added just enough to have context on Bind, which could block if the server does the delay-on-bad-password thing, in addition to the Dial itself timing out.

there's more room to improve LDAP verification but for now it can't hold up scanning

### Description:

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches live LDAP verification and swaps the LDAP client library, which could change timeout/binding behavior and verification outcomes. Scope is limited to the LDAP detector and adds targeted tests for the new error classification logic.
> 
> **Overview**
> Makes LDAP credential verification **context-aware** so scans can be canceled/time out cleanly, updating LDAP dialing and binds to use `ctx` (including STARTTLS and LDAPS flows).
> 
> Replaces `github.com/go-ldap/ldap/v3` with `github.com/mariduv/ldap-verify` (and updates transitive deps), and revises *verification error determinism* to treat network and context cancellation/deadline failures as indeterminate; adds `Test_isErrDeterminate` to cover these cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bc9ea635650f81c649450492aed7730d3cdcf48c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/detectors/ldap/ldap.go
+++ b/pkg/detectors/ldap/ldap.go
@@ -3,13 +3,14 @@ package ldap
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
 	"strings"
 	"time"
 
-	"github.com/go-ldap/ldap/v3"
+	ldap "github.com/mariduv/ldap-verify"
 	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
@@ -72,7 +73,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				}
 
 				if verify {
-					verificationErr := verifyLDAP(username[1], password[1], ldapURL)
+					verificationErr := verifyLDAP(ctx, username[1], password[1], ldapURL)
 					s1.Verified = verificationErr == nil
 					if !isErrDeterminate(verificationErr) {
 						s1.SetVerificationError(verificationErr, password[1])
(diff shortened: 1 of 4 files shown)


pkg/detectors/ldap/ldap.go
```diff
@@ -3,13 +3,14 @@ package ldap
...
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
 	"strings"
 	"time"
 
-	"github.com/go-ldap/ldap/v3"
+	ldap "github.com/mariduv/ldap-verify"
```
shahzadhaider1, line 13: Have we looked at whether this context support could be upstreamed to `go-ldap/ldap` instead? That would give us the fix without the maintenance risk.

shahzadhaider1 (commented):
This looks solid to me, just a small concern.

pkg/detectors/ldap/ldap.go
mariduv: yeah, this is based on someone's abandoned partial PR to add context support but it's not complete or even entirely correct, i only extended it to what we use and made that correct.  go-ldap _does_ deserve upstreamed context support then we could switch back to it, but we don't use all of the ldap functionality so the work didn't need to be complete.  i'd love for one of us to get time to contribute there because I looked around and there seem to be no options out there for context-aware ldap libraries in go, everything pre-dates context.

alafiand (approved):

shahzadhaider1 (approved):

PR #4769, opened 2026-02-25
[INS-348] Add New Relic Browser Key detector

mustansir14:
### Description:
This PR adds the New Relic Browser Key Detector.

**Regex**: `\b(NRBR-[0-9a-f]{19})\b`

**Verification**:
For verification, we use the Browser Agent API: `https://bam.nr-data.net/1/`. We send a POST request with an empty body to this endpoint. A response code of `400` means the key is valid but the request body is malformed. `403` means it is an invalid/rotated key.
This API endpoint is not documented. It was discovered by observing network traffic from the New Relic Browser agent.
Note: For EU region keys, the host should be `https://bam.eu01.nr-data.net/1/`

**Corpora Test**:
image
image

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive detector-only change with outbound verification calls; no changes to auth, storage, or core scan logic.
> 
> **Overview**
> Adds **New Relic Browser** key (`NRBR-` + 19 hex chars) detection and optional verification against New Relic’s browser agent ingest hosts (US and EU).
> 
> When verification is on, the scanner POSTs to the undocumented `bam.nr-data.net` / `bam.eu01.nr-data.net` paths and treats **400** as valid and **403** as invalid; successful checks attach a **`region`** in result metadata.
> 
> Wiring matches other gated detectors: new `DetectorType_NewRelicBrowserKey` (1065), `NewRelicBrowserKeyDetectorEnabled` (on by default in `main.go`), registration in default detector list with flag filtering, and unit/integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ad3eae4e4fae2a63b2685cc0389f5f2cdfa2915. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/newrelicbrowserkey/newrelicbrowserkey.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		res, err := client.Do(req)
+		if err != nil {
+			return false, nil, fmt.Errorf("error making request: %w", err)
+		}
+		defer func() {
+			_, _ = io.Copy(io.Discard, res.Body)
+			_ = res.Body.Close()
+		}()
```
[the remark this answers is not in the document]
mustansir14: This is fine because there are always only 2 iterations of the loop.

pkg/detectors/newrelicbrowserkey/newrelicbrowserkey.go
```diff
@@ -0,0 +1,118 @@
...
+// The endpoints were discovered by observing network traffic from the New Relic Browser agent
+// A 400 Bad Request response indicates that the key is valid but the request is malformed (because we are not sending the expected payload)
+// A 403 Forbidden response indicates that the key is invalid or revoked
+func (s Scanner) verify(ctx context.Context, key string) (bool, map[string]string, error) {
+	regionUrls := map[string]string{
+		"us": "https://bam.nr-data.net/1/",
+		"eu": "https://bam.eu01.nr-data.net/1/",
+	}
+	client := s.getClient()
+	for region, regionUrl := range regionUrls {
```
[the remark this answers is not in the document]
mustansir14: This has been tested and both region endpoints always return a 403 if the key is of different region, so it's okay if the map iteration is non-deterministic.

shahzadhaider1 (approved):

pkg/detectors/newrelicbrowserkey/newrelicbrowserkey.go
```diff
@@ -0,0 +1,118 @@
...
+			_ = res.Body.Close()
+		}()
+
+		switch res.StatusCode {
+		case http.StatusBadRequest:
+			return true, map[string]string{"region": region}, nil
+		case http.StatusForbidden:
+			continue
+		default:
+			return false, nil, fmt.Errorf("unexpected status code: %d", res.StatusCode)
```
shahzadhaider1: What if the US endpoint returns an unexpected status (say a 429), we bail out entirely without ever checking the EU endpoint. This could produce false negatives for EU-region keys during transient errors.

pkg/detectors/newrelicbrowserkey/newrelicbrowserkey.go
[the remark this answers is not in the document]
shahzadhaider1: I think the bugbot is right here, better to be safe

pkg/detectors/newrelicbrowserkey/newrelicbrowserkey.go
mustansir14: I have tested this and the API only returns 400 for a valid key and 403 for an invalid/rotated/wrong region key. Any other status code would mean there is a verification error which should always be returned IMO.

I'm unsure how a 429 should be handled though. The [docs](https://docs.newrelic.com/docs/data-apis/manage-data/view-system-limits/) do mention that there are rate limits on the API. However, I think a 429 (Too many requests) would actually mean that the region is correct, and there are too many requests by that API key, in which case there's no point trying the other region. What do you think?

shahzadhaider1: > However, I think a 429 (Too many requests) would actually mean that the region is correct, and there are too many requests by that API key.

I’m not sure that a 429 necessarily confirms the region is correct. We should double-check that assumption. A 429 typically indicates rate limiting, but that doesn’t always guarantee that the key was validated far enough in the request lifecycle to confirm regional correctness. It’s possible the request was rejected due to throttling or upstream load before region-specific validation even occurred.

mustansir14: Hmm, that makes sense.

I'll change the logic to try both regions in any case, and return an aggregated error in the end.

proto/detector_type.proto
```diff
@@ -1065,4 +1065,5 @@ enum DetectorType {
   HashiCorpVaultBatchToken = 1061;
   HashiCorpVaultToken = 1062;
   Duo = 1063;
+  NewRelicBrowserKey = 1064;
```
unsmith: This is a conflict with #4760 , both are using the same proto enum value.

pkg/detectors/newrelicbrowserkey/newrelicbrowserkey_integration_test.go
```diff
@@ -0,0 +1,151 @@
...
+			want:    nil,
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := Scanner{}
+			got, err := s.FromData(tt.args.ctx, tt.args.verify, tt.args.data)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("NewRelicLicenseKey.FromData() error = %v, wantErr %v", err, tt.wantErr)
```
unsmith: Should this be `NewRelicBrowserKey` ?

pkg/detectors/newrelicbrowserkey/newrelicbrowserkey_integration_test.go
```diff
@@ -0,0 +1,151 @@
...
+					t.Fatalf("no raw secret present: \n %+v", got[i])
+				}
+				got[i].Raw = nil
+				if len(got[i].Redacted) == 0 {
+					t.Fatalf("no redacted secret present: \n %+v", got[i])
+				}
+				got[i].Redacted = ""
+			}
+			if diff := pretty.Compare(got, tt.want); diff != "" {
+				t.Errorf("NewRelicLicenseKey.FromData() %s diff: (-got +want)\n%s", tt.name, diff)
```
unsmith: Should this be `NewRelicBrowserKey` ?

unsmith (changes requested):
Need to resolve the proto conflict, but otherwise ok by me.

proto/detector_type.proto
MuneebUllahKhan222: I usually resolve any proto conflicts and regenerate the protos once my PR has been approved and is ready to merge, since these conflicts tend to occur every time a detector PR gets merged.

unsmith (approved):

PR #4774, opened 2026-02-27
Fix GitHub App cross-org member enumeration using per-installation tokens

dustin-decker:
## Summary

- When a GitHub App is installed across multiple orgs, `addMembersByApp` discovers all installations via JWT but uses a single installation's token for `ListMembers` calls across all orgs
- GitHub App installation tokens only bypass IP allowlists for their own org, so cross-org calls get 403s
- Adds `APIClientForInstallation` to `appConnector` which creates per-installation API clients using `ghinstallation.NewFromAppsTransport`
- Extracts `addMembersByOrgWithClient` from `addMembersByOrg` so the app auth path can provide an explicit per-installation client

## Test plan

- [x] New unit tests for `APIClientForInstallation` (3 subtests verifying distinct clients, correct installation ID, different tokens)
- [x] New unit test for `addMembersByOrgWithClient`
- [x] Existing `TestAddMembersByApp` and `TestEnumerateWithApp` updated and passing
- [x] `go build` passes
- [x] Manual verification with a multi-org GitHub App installation

---

> [!NOTE]
> **Medium Risk**
> Changes GitHub App authentication and org member enumeration; mis-scoped tokens could still miss members, but behavior is narrowed with tests and per-org error isolation.
> 
> **Overview**
> Fixes **GitHub App** member enumeration when the app is installed on **multiple orgs** by using an **installation-scoped API client** for each org instead of one token for every `ListMembers` call (which caused **403s** when IP allowlists apply per installation).
> 
> **`appConnector`** now keeps the shared `AppsTransport` and adds **`APIClientForInstallation`**, built with `ghinstallation.NewFromAppsTransport`; the default API client from `NewAppConnector` still targets the configured installation ID. **`addMembersByApp`** walks app installations, builds a client per installation ID, and calls **`addMembersByOrgWithClient`**; failures for a single org are logged and skipped rather than aborting the whole run. **`addMembersByOrg`** delegates to the same helper with the connector’s default client.
> 
> New unit tests cover per-installation clients and the refactored member listing; existing app enumeration tests were updated for installation IDs in mocks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3fe806a69147ff3b078cc5aa4e6eb550bcb7c66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/github/connector_app.go
+++ b/pkg/sources/github/connector_app.go
@@ -20,6 +20,8 @@ type appConnector struct {
 	graphqlClient      *githubv4.Client
 	installationClient *github.Client
 	installationID     int64
+	appsTransport      *ghinstallation.AppsTransport
+	apiEndpoint        string
 }
 
 var _ Connector = (*appConnector)(nil)
@@ -80,6 +82,8 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
 		graphqlClient:      graphqlClient,
 		installationClient: installationClient,
 		installationID:     installationID,
+		appsTransport:      installationTransport,
+		apiEndpoint:        apiEndpoint,
 	}, nil
 }
 
@@ -106,4 +110,22 @@ func (c *appConnector) GraphQLClient() *githubv4.Client {
 
 func (c *appConnector) InstallationClient() *github.Client {
 	return c.installationClient
+}
+
+// APIClientForInstallation creates a GitHub API client scoped to a specific
+// installation. This is needed when a GitHub App is installed across multiple
+// orgs — each org's API calls must use that org's installation token to get
+// proper IP allowlist bypass and permission scoping.
+func (c *appConnector) APIClientForInstallation(installationID int64) (*github.Client, error) {
+	transport := ghinstallation.NewFromAppsTransport(c.appsTransport, installationID)
+	transport.BaseURL = c.apiEndpoint
+
+	httpClient := common.RetryableHTTPClientTimeout(60)
+	httpClient.Transport = transport
+
+	client, err := github.NewClient(httpClient).WithEnterpriseURLs(c.apiEndpoint, c.apiEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("could not create API client for installation %d: %w", installationID, err)
+	}
+	return client, nil
 }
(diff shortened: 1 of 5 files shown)


pkg/sources/github/connector_app.go
```diff
@@ -107,3 +111,21 @@ func (c *appConnector) GraphQLClient() *githubv4.Client {
 func (c *appConnector) InstallationClient() *github.Client {
 	return c.installationClient
 }
+
+// APIClientForInstallation creates a GitHub API client scoped to a specific
+// installation. This is needed when a GitHub App is installed across multiple
+// orgs — each org's API calls must use that org's installation token to get
+// proper IP allowlist bypass and permission scoping.
+func (c *appConnector) APIClientForInstallation(installationID int64) (*github.Client, error) {
```
rosecodym, line 110: Can this function be used by `NewAppConnector`? That function creates a new API client for a particular installation ID (the ID specified in the connection information).

rosecodym (commented):
I think this review will be a bit enmeshed with my (upcoming) review of #4775.

shahzadhaider1 (approved):

pkg/sources/github/connector_app.go
dustin-decker: Fixed. NewAppConnector now builds its configured installation API client through APIClientForInstallation(installationID), then wires GraphQL from that client, so the default app path uses the same per-installation client construction as cross-org member enumeration. Added coverage for the configured installation token path as well.

pkg/sources/github/github.go
```diff
@@ -761,7 +761,11 @@ func (s *Source) cloneAndScanRepo(ctx context.Context, repoURL string, repoInfo
 	// if legacy JSON is enabled, don't remove the directory because we need it for outputting legacy JSON.
 	if !s.conn.GetPrintLegacyJson() {
 		if strings.HasPrefix(path, filepath.Join(os.TempDir(), "trufflehog")) || (!s.conn.NoCleanup && s.conn.GetClonePath() != "") {
-			defer os.RemoveAll(path)
+			defer func() {
+				if err := os.RemoveAll(path); err != nil {
+					ctx.Logger().Error(errors.New("remove failed"), "could not remove cloned repository")
```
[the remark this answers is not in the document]
dustin-decker: Fixed. The cleanup path now preserves the underlying removal cause for os.PathError via pathErr.Err and logs the operation name, without logging the full path-bearing os.PathError value.

dustin-decker:
Rebased on latest main, CI is green (15 passing, 0 failing), and all active review threads are resolved. Ready for review/merge once the required approval is refreshed.

kashifkhan0771 (approved):

PR #4775, opened 2026-02-27
Add scan_all_installations option for multi-org GitHub App scanning

dustin-decker:
## Summary

Stacked on #4774.

Adds a `scan_all_installations` proto field to `message GitHub`. When true and using GitHub App auth, the scanner:

1. Enumerates all installations via JWT (`GET /app/installations`)
2. For each installation beyond the stored one, creates a per-installation API client
3. Lists repos from each installation using `Apps.ListRepos`
4. Maps each repo's clone URL to its owning installation ID so `Clone` uses the correct token

This enables scanning all orgs where a GitHub App is installed, not just the one whose installation ID happens to be stored in the database.

## Changes

- `proto/sources.proto`: Add `scan_all_installations = 26` to `message GitHub`
- `connector_app.go`: Add `repoInstallationMap` + `SetRepoInstallation`, update `Clone` to use mapped installation IDs
- `github.go`: Add `enumerateAllInstallationRepos`, call it from `enumerateWithApp` when `ScanAllInstallations` is true
- `connector_app_test.go`: Add `TestCloneUsesRepoInstallationMap`

## Test plan

- [x] `TestCloneUsesRepoInstallationMap` -- verifies Clone uses default vs mapped installation IDs
- [x] Existing `TestAddMembersByApp` and `TestEnumerateWithApp` pass
- [x] `go build` passes
- [x] Manual verification with multi-org app installation

---

> [!NOTE]
> **Medium Risk**
> Changes GitHub App auth and repo enumeration across orgs; mis-mapping could cause failed clones or wrong tokens, but behavior is gated behind an explicit config flag.
> 
> **Overview**
> Adds **`scan_all_installations`** on the GitHub source config. With GitHub App credentials, enabling it lists repos from **every** app installation (not only the stored installation ID) and records which installation owns each repo.
> 
> The **app connector** now caches API/GraphQL clients per installation, maps clone URLs (including wikis) to installation IDs, and uses **`APIClientForRepo` / `GraphQLClientForRepo`** so clone, enumeration, comments, and targeted scans use the correct installation token. Explicit repo lists and per-unit scans run **`mapReposToInstallations`** when the flag is on.
> 
> Enumeration and member discovery reuse a shared installation list; per-installation failures are reported without stopping the rest. Rate-limit backoff respects context cancellation; wiki and metadata targeting paths are aligned with installation-scoped clients.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b50d29d18ca0f46c607828dde80cad51f8948f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/sources/github/github.go
```diff
@@ -630,6 +630,12 @@ func (s *Source) enumerateWithApp(ctx context.Context, connector *appConnector,
 			return err
 		}
 
+		if s.conn.ScanAllInstallations {
+			if err := s.enumerateAllInstallationRepos(ctx, connector, reporter); err != nil {
+				ctx.Logger().Error(err, "error enumerating repos from additional installations")
+			}
+		}
+
```
rosecodym, line 672: Why does this block happen in _addition_ to the `getReposByApp` call above? My gut sense is that we should either be scanning one installation or scanning all of them. (Put another way, "scan all installations" seems like it should _supersede_ "scan just the configured installation" - not happen in addition to it.)

This is, of course, a value judgment, but I'll also note that as written here, it looks like the directly configured installation will be scanned twice. I can see that the `enumerateAllInstallationRepos` implementation doesn't do that, but it's non-obvious enough here that I think a comment is warranted if you leave it this way.

pkg/sources/github/github.go
```diff
@@ -655,6 +661,66 @@ func (s *Source) enumerateWithApp(ctx context.Context, connector *appConnector,
...
+	}
+
+	for _, install := range allInstalls {
+		if install.GetID() == connector.installationID {
+			continue
+		}
+
+		installID := install.GetID()
+		account := install.Account.GetLogin()
+		logger := ctx.Logger().WithValues("org", account, "installation_id", installID)
```
rosecodym: Please create a new context with these key-value pairs instead of creating a separate logger for them.

pkg/sources/github/github.go
```diff
@@ -655,6 +661,66 @@ func (s *Source) enumerateWithApp(ctx context.Context, connector *appConnector,
...
+// creates a per-installation API client and lists repos accessible to that
+// installation. It also records the installation ID for each repo so that
+// Clone uses the correct token.
+func (s *Source) enumerateAllInstallationRepos(ctx context.Context, connector *appConnector, reporter sources.UnitReporter) error {
+	installationClient := connector.InstallationClient()
+	opts := &github.ListOptions{PerPage: defaultPagination}
+
+	var allInstalls []*github.Installation
+	for {
+		installs, res, err := installationClient.Apps.ListInstallations(ctx, opts)
```
rosecodym, line 733: Adding a second call of `ListInstallations` in this file concerns me. What are the barriers to unifying them?

pkg/sources/github/github.go
dustin-decker: Implemented as superseding behavior. When scan_all_installations is set, enumerateWithApp now calls enumerateAllInstallationRepos instead of getReposByApp, so the configured installation is not scanned separately or doubled.

pkg/sources/github/github.go
dustin-decker: Changed this path to create a context with org and installation_id values and pass that context through the per-installation logging/member enumeration work.

pkg/sources/github/github.go
dustin-decker: Unified installation enumeration through listAppInstallations, including pagination and rate-limit retry handling. Both scan-all repo enumeration and member enumeration now use that helper.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +64,126 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
 	return c.apiClient
 }
 
+func (c *appConnector) APIClientForRepo(repoURL string) (*github.Client, error) {
+	installID := c.installationIDForRepo(repoURL)
+	if installID == c.installationID {
+		return c.apiClient, nil
+	}
+	return c.APIClientForInstallation(installID)
+}
```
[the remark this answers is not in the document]
dustin-decker: Fixed. APIClientForInstallation now caches clients per installation ID under a mutex, so repeated repo operations against the same installation reuse one client and its token instead of creating a new transport each call. Each cached client gets its own apps transport since the shared ghinstallation AppsTransport is mutated on token refresh.

pkg/sources/github/github.go
```diff
@@ -663,6 +680,362 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+			ListOptions: github.ListOptions{PerPage: defaultPagination},
+		}); err != nil {
+			errs = append(errs, fmt.Errorf("error enumerating repos for installation %d: %w", installID, err))
+			installCtx.Logger().Error(err, "error enumerating repos for installation")
+			continue
+		}
+	}
+
+	return errors.Join(errs...)
+}
```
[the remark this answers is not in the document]
dustin-decker: Fixed. APIClientForInstallation returns the pre-configured default client when asked for the configured installation, so enumerateAllInstallationRepos no longer builds a redundant client for it. The ListRepos call for that installation is the sole enumeration path now that scan_all_installations supersedes getReposByApp, so it is not duplicated work.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +69,152 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+		return c.graphqlClient, nil
+	}
+
+	apiClient, err := c.APIClientForInstallation(installID)
+	if err != nil {
+		return nil, err
+	}
+
+	return createGraphqlClient(ctx, apiClient.Client(), c.apiEndpoint)
+}
```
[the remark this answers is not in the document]
dustin-decker: Fixed. GraphQLClientForRepo now caches clients per installation ID (graphqlClientsByInstallationID), mirroring the REST client cache, so cross-org comment scanning reuses one GraphQL client per installation.

pkg/sources/github/github.go
```diff
@@ -663,6 +680,362 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+
+func (s *Source) repoURLFromTargetMetadata(meta *source_metadatapb.Github) (string, error) {
+	linkRepo, err := s.repoURLFromTargetLink(meta.GetLink())
+	if err != nil {
+		return "", err
+	}
+
+	if repo := meta.GetRepository(); repo != "" {
+		if !strings.Contains(repo, "/") && !regexp.MustCompile(`^[a-z]+://`).MatchString(repo) && !strings.Contains(repo, ":") {
```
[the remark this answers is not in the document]
dustin-decker: Fixed. Hoisted the scheme pattern to a package-level hasURLScheme var and replaced the inline compilations here and in normalizeRepo.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,362 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+				for _, repo := range result.Repositories {
+					key, err := repoHostOwnerNameKey(repo.GetCloneURL())
+					if err == nil {
+						installedRepoKeys[installID][key] = struct{}{}
+					}
+				}
+				s.recordInstallationRepoMatches(connector, installID, result.Repositories, wantedRepos)
+				if len(wantedRepos) == 0 {
+					return nil
+				}
```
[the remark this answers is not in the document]
dustin-decker: Fixed by logging. Returning the accumulated errors when every requested repo was successfully mapped would fail the enumeration unnecessarily, so the early-return paths now log the per-installation failures via logInstallationMappingErrors instead of discarding them. The fail-closed path (unmapped repos) still returns the joined errors.

pkg/sources/github/github.go
```diff
@@ -509,7 +518,12 @@ func (s *Source) ensureRepoInfoCache(ctx context.Context, repo string, reporter
 	} else {
 		// Cache repository info.
 		for {
-			ghRepo, _, err := s.connector.APIClient().Repositories.Get(ctx, urlParts[1], urlParts[2])
```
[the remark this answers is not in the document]
dustin-decker: This one is a no-op in practice: repoInstallationMap is only ever populated with repository clone URLs (from Apps.ListRepos and explicit-repo mapping; ChunkUnit also only maps units of kind "repo", not gists). A gist URL always misses the map, so apiClientForRepo would return the same default installation client that connector.APIClient() returns today. Gists are user-owned and not part of installation repo lists, so there is no per-installation token to select. Leaving the gist path as-is.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,168 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+
+	client, err := createGraphqlClient(ctx, apiClient.Client(), c.apiEndpoint)
+	if err != nil {
+		return nil, err
+	}
+	if c.graphqlClientsByInstallationID == nil {
+		c.graphqlClientsByInstallationID = make(map[int64]*githubv4.Client)
+	}
+	c.graphqlClientsByInstallationID[installID] = client
+	return client, nil
```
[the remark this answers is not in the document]
dustin-decker: Restructured with the double-check pattern: the cache lookup and store each take graphqlClientsMu briefly, and APIClientForInstallation/createGraphqlClient run outside the lock, so the mutexes are never nested. For accuracy: createGraphqlClient performs no network I/O (it only derives the endpoint URL string), and the prior lock order was fixed in one direction, but removing the nesting makes that a non-concern going forward.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,373 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+			ListOptions: github.ListOptions{PerPage: defaultPagination},
+		}); err != nil {
+			errs = append(errs, fmt.Errorf("error enumerating repos for installation %d: %w", installID, err))
+			installCtx.Logger().Error(err, "error enumerating repos for installation")
+			continue
+		}
+	}
+
+	return errors.Join(errs...)
+}
```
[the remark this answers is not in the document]
dustin-decker: Fixed. Per-installation failures are now reported via reporter.UnitErr and enumeration continues with the remaining installations; only a failure to list installations themselves is fatal. The regression test now asserts that one failing installation is reported while the other installation still gets enumerated.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+// abort enumeration of the remaining installations.
+func (s *Source) enumerateAllInstallationRepos(ctx context.Context, connector *appConnector, reporter sources.UnitReporter) error {
+	installs, err := s.listAppInstallations(ctx, connector, defaultPagination)
+	if err != nil {
+		return err
+	}
+
+	for _, install := range installs {
+		installID := install.GetID()
+		account := install.Account.GetLogin()
```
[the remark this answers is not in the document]
dustin-decker: Switched to install.GetAccount().GetLogin()/GetType() in both spots. For accuracy: the original code could not panic — install.Account is a field read on a non-nil install, and go-github getters are nil-receiver-safe — but the getter form is idiomatic and also guards a nil installation element.

pkg/sources/github/github.go
```diff
@@ -938,23 +1330,32 @@ func (s *Source) addUserGistsToCache(ctx context.Context, user string, reporter
...
+		// their own org; using a cross-org token causes 403s.
+		client, err := connector.APIClientForInstallation(install.GetID())
+		if err != nil {
+			installCtx.Logger().Error(err, "could not create API client for installation")
+			continue
+		}
+
+		if err := s.addMembersByOrgWithClient(installCtx, client, org, reporter); err != nil {
+			installCtx.Logger().Error(err, "Unable to add members for org")
+			continue
```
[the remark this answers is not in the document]
dustin-decker: Fixed. addMembersByApp now reports per-installation client and member enumeration failures via reporter.UnitErr before continuing, matching enumerateAllInstallationRepos. (Note the prior behavior here was log-and-continue, not return-to-caller.)

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+		listRepos := func(ctx context.Context, _ string, opts repoListOptions) ([]*github.Repository, *github.Response, error) {
+			result, resp, err := client.Apps.ListRepos(ctx, opts.getListOptions())
+			if result != nil {
+				for _, r := range result.Repositories {
+					connector.setRepoInstallation(r.GetCloneURL(), installID)
+				}
+				return result.Repositories, resp, err
+			}
+			return nil, resp, err
+		}
```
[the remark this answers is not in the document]
dustin-decker: No change needed: go.mod declares go 1.25.0, so per-iteration loop variable scoping is guaranteed and the closure captures are safe, as the finding itself notes for Go 1.22+.

pkg/sources/github/github.go
```diff
@@ -1650,15 +2070,29 @@ func (s *Source) scanTarget(ctx context.Context, target sources.ChunkingTarget,
...
 		ctx := context.WithValues(ctx, "commit_hash", meta.GetCommit())
 		ctx.Logger().V(2).Info("secret metadata has no file; scanning commit metadata instead")
 
-		return s.scanCommitMetadata(ctx, segments[1], segments[2], meta, &chunkSkel, reporter)
+		return s.scanCommitMetadata(ctx, apiClient, segments[1], segments[2], meta, &chunkSkel, reporter)
 	}
 
 	// else try downloading the file content to scan
-	readCloser, resp, err := s.connector.APIClient().Repositories.DownloadContents(
+	readCloser, resp, err := apiClient.Repositories.DownloadContents(
```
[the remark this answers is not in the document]
dustin-decker: No change needed. repoURL is derived by repoURLFromTargetLink from the same link, using the same first two path segments that scanTarget indexes as segments[1]/segments[2] (the leading empty segment from splitting "/owner/repo/..." is why indexing starts at 1). normalizeRepo only canonicalizes case/format and never resolves renames or redirects, so the owner/repo used for API calls cannot diverge from what the client was scoped to. Renamed repos are handled by GitHub API-side redirects, which go-github follows.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,176 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+	client, err := github.NewClient(httpClient).WithEnterpriseURLs(c.apiEndpoint, c.apiEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("could not create API client for installation %d: %w", installationID, err)
+	}
+	c.apiClientsByInstallationID[installationID] = client
+	if installationID == c.installationID {
+		c.apiClient = client
+	}
+	return client, nil
+}
```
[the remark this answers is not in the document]
dustin-decker: No change needed. Client creation here is pure in-memory work (PEM parse plus struct construction) with no network I/O, so holding the mutex briefly on the first-use path for each installation is negligible. The double-check pattern in GraphQLClientForRepo exists to avoid nesting graphqlClientsMu with apiClientsMu, not for performance; no nesting occurs in APIClientForInstallation, and the simpler single-lock form is easier to reason about.

pkg/sources/github/repo.go
```diff
@@ -338,13 +345,26 @@ func (s *Source) cacheGistInfo(g *github.Gist) {
 // Unfortunately, this isn't 100% accurate. Some repositories have `has_wiki: true` and don't redirect their wiki page,
 // but still don't have a cloneable wiki.
 func (s *Source) wikiIsReachable(ctx context.Context, repoURL string) bool {
-	wikiURL := strings.TrimSuffix(repoURL, ".git") + "/wiki"
+	wikiBaseURL := repoURL
+	if repoInfo, ok := s.repoInfoCache.get(repoURL); ok {
+		wikiBaseURL = (&url.URL{
+			Scheme: "https",
+			Host:   githubWebHost(s.conn.GetEndpoint()),
+			Path:   fmt.Sprintf("/%s/%s", repoInfo.owner, repoInfo.name),
```
camgunz: I think you could hoist the `".git"` addition up here, but also it's always removed by the `.TrimSuffix` call so maybe it's not necessary?

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,176 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+	// with apiClientsMu (taken by APIClientForInstallation).
+	apiClient, err := c.APIClientForInstallation(installID)
+	if err != nil {
+		return nil, err
+	}
+
+	client, err = createGraphqlClient(ctx, apiClient.Client(), c.apiEndpoint)
+	if err != nil {
+		return nil, err
+	}
```
camgunz: Creating a GraphQL client _only_ requires the API client's HTTP client, so it seems somewhat strange to additionally always ensure there's an API client for the install ID, but only in the context that the multiple locks allow these inconsistencies. It all made me think that we really have something like `installation info`:
- ID int64
- Client *http.Client
- APIClient *github.Client
- GraphQLClient *githubv4.Client

This is the thing we want synchronized. It's not good if the HTTP client is different than the GraphQL client, or the GraphQL client is different than the API client. But, with the separate locks, you're hoping that users of the interface you've constructed either never create an inconsistency here, or if they do it doesn't matter. This method is an example of the "it doesn't matter" case: line 133 isn't synchronized, so multiple threads could make multiple GraphQL clients. It doesn't matter, and further this method isn't--currently--called in a concurrent context so it matters even less. It's de facto harmless, but it took me... probably a whole day of me and Claude learning about the GitHub source to really know that for sure, and that's not ideal.

pkg/sources/github/connector_app.go
```diff
@@ -20,6 +22,22 @@ type appConnector struct {
...
+	apiClientsByInstallationID map[int64]*github.Client
+
+	graphqlClientsMu               sync.Mutex
+	graphqlClientsByInstallationID map[int64]*githubv4.Client
+
+	// repoInstallationMap maps repo clone URLs to their owning installation
+	// ID for repos discovered from non-default installations. Clone checks
+	// this map to use the correct installation token.
+	repoInstallationMu  sync.RWMutex
+	repoInstallationMap map[string]int64
```
camgunz: Three things here:

First, I think it's probably always worth it to combine values when constructing multiple `map`s with the same key, so like in the case of `apiClientsByInstallationID` and `graphqlClientsByInstallationID` you'd want some one-off struct with a `*githubv4.Client` and a `*github.Client`. I don't think this would super be under contention, and anyway you can use a `RWMutex` because... I assume you'd 99% of the time be reading clients from it anyway.

Second, I kind of understand guarding these maps with locks, but I'm not sure they're ever written to in a concurrent context. I think this is always in the context of enumerating, which is done serially. Which is to say, I think we wouldn't suffer any contention problems if we just used a single `RWMutex` for the entire `appConnector`, and we'd reap a big benefit in reduced complexity. 

Third and finally, it's pretty weird to have this alongside separate members for `apiClient`, `graphqlClient`, and `installationID`, and you can see the awkwardness in the methods where you have to check to see if you're gonna just return `apiClient` before doing the new thing. I didn't look too closely at this, but I imagine it's possible to just pop the installation ID, API client, and GraphQL client into the `map`s and forego these individual struct members here.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,176 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+	}
+
+	c.graphqlClientsMu.Lock()
+	defer c.graphqlClientsMu.Unlock()
+	if existing, ok := c.graphqlClientsByInstallationID[installID]; ok {
+		return existing, nil
+	}
+	if c.graphqlClientsByInstallationID == nil {
+		c.graphqlClientsByInstallationID = make(map[int64]*githubv4.Client)
+	}
```
camgunz: IMO it's noisy to check for a `nil` `map` everywhere. There's a constructor because the struct has members that require initialization. If a caller doesn't use the constructor, they get what they get. Otherwise you get something of a false sense of security, like we're checking for a `nil` `map` here, but we access it way earlier on line 120 and 140, so already we're just being saved by the fact that everything so far is using `NewAppConnector` (or, I guess anything that doesn't also doesn't call this method).

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,176 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
 
+func (c *appConnector) installationIDForRepo(repoURL string) int64 {
+	c.repoInstallationMu.RLock()
+	defer c.repoInstallationMu.RUnlock()
+
+	if id, ok := c.repoInstallationMap[repoURL]; ok {
+		return id
+	}
+	return c.installationID
+}
```
camgunz: It seems a little funny that no matter what I pass to this method I'll get a valid installation ID. Like I understand why it is the way it is, but it's another artifact of preserving the old interface. Ideally if a lookup in this `map` failed we'd return an `error` here.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,176 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+	c.repoInstallationMu.Lock()
+	defer c.repoInstallationMu.Unlock()
+
+	if c.repoInstallationMap == nil {
+		c.repoInstallationMap = make(map[string]int64)
+	}
+	c.repoInstallationMap[repoURL] = installationID
+	if strings.HasSuffix(repoURL, ".git") {
+		wikiURL := strings.TrimSuffix(repoURL, ".git") + ".wiki.git"
+		c.repoInstallationMap[wikiURL] = installationID
```
camgunz: Should we only be doing this for URLs that end in `".git"`? I don't know much about this, but the Internet says it's "<project>.wiki.git", which would go for URLs that are "<project>" as well as "<project>.git", but IDK.

pkg/sources/github/connector_app.go
```diff
@@ -54,56 +72,176 @@ func NewAppConnector(ctx context.Context, apiEndpoint string, app *credentialspb
...
+	transport := ghinstallation.NewFromAppsTransport(appsTransport, installationID)
+	transport.BaseURL = c.apiEndpoint
+
+	httpClient := common.RetryableHTTPClientTimeout(60)
+	httpClient.Transport = transport
+
+	client, err := github.NewClient(httpClient).WithEnterpriseURLs(c.apiEndpoint, c.apiEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("could not create API client for installation %d: %w", installationID, err)
+	}
```
camgunz: I didn't check line by line, but this is pretty similar to the setup code in `NewAppConnector`. Ideally we'd unify this.

pkg/sources/github/github.go
```diff
@@ -129,6 +130,9 @@ var _ sources.SourceUnitEnumChunker = (*Source)(nil)
 
 var endsWithGithub = regexp.MustCompile(`github\.com/?$`)
 
+// hasURLScheme matches strings that start with a URL scheme (e.g. "https://").
+var hasURLScheme = regexp.MustCompile(`^[a-z]+://`)
```
camgunz: I'm thoroughly in favor of following existing patterns--even and especially when they're not perfect. But just to note, these regexes are pretty wacky. Like, for one there's `strings.HasSuffix` (in the case of `endsWithGithub` up there). Here ideally we'd parse a URL and check to see if the scheme was something we recognized--maybe this is a helper method (FWIW there's also [`IsAbs`](https://pkg.go.dev/net/url#URL.IsAbs)), maybe this is early on when we're dealing with a repo URL for the first time. But yeah, just a note.

pkg/sources/github/github.go
```diff
@@ -403,6 +407,11 @@ func (s *Source) Enumerate(ctx context.Context, reporter sources.UnitReporter) e
 	// Report any values that were already configured.
 	// This compensates for differences in enumeration logic between `--org` and `--repo`.
 	// See: https://github.com/trufflesecurity/trufflehog/pull/2379#discussion_r1487454788
+	if connector, ok := s.connector.(*appConnector); ok && s.conn.ScanAllInstallations && len(s.repos) > 0 {
```
camgunz: Again w/ the established pattern, but the scattering of type assertions on `*appConnector` is a pretty strong signal that more should be happening in `appConnector` itself, or more broadly, that the `Connector` interface is a little too limited. One way to simplify these spots would be to change the `Connector` interface to be the `appConnector` interface, ex:

```
type Connector interface {
	// APIClient returns a configured GitHub client that can be used for GitHub API operations.
	APIClientForRepo(repo string) *github.Client
	// GraphQLClient returns a client that can be used for GraphQL operations.
	GraphQLClientForRepo(repo string) *githubv4.Client
	// Clone clones a repository using the configured authentication information.
	Clone(ctx context.Context, repoURL string, args ...string) (string, *gogit.Repository, error)
}
```

The other implementations could get away with simply changing the signatures of their methods--no logic changes necessary--and `appConnector` could then return different clients.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+		}
+	}
+
+	if len(wantedRepos) > 0 {
+		unmatched := make([]string, 0, len(wantedRepos))
+		for _, requested := range wantedRepos {
+			unmatched = append(unmatched, requested.original)
+		}
+		sort.Strings(unmatched)
+		errs = append(errs, fmt.Errorf("configured repos were not found in any GitHub App installation: %s", strings.Join(unmatched, ", ")))
```
camgunz: I assume this is here to guard against misconfigurations, but would this also trigger if a repo was deleted between enumeration and here? I don't know enough about GH to know if this question makes sense, but I guess my overarching question is this an error we should bail on a scan and mark the scanner unhealthy for?

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+func (s *Source) mapExplicitReposToInstallations(ctx context.Context, connector *appConnector) error {
+	return s.mapReposToInstallations(ctx, connector, s.repos)
+}
+
+type repoMappingRequest struct {
+	original   string
+	normalized string
+}
+
+func (s *Source) mapReposToInstallations(ctx context.Context, connector *appConnector, repos []string) error {
```
camgunz, line 832: At first I thought this felt more like a kind of initializer for `appConnector`; mostly what it does is just set up `appConnector`'s `repoInstallationMap`. But then there at the end it also updates `Source`'s repo info cache. This indicates there's some process that spans the lifetimes of both these objects that shouldn't depend on functionality inside either of them. The problem, at least as I see it, is that while the additions to `appConnector` require additional initialization, previously connectors didn't, so it's unclear exactly where to do it, so we do it everywhere all the time: when enumerating, when reverifying, etc.

The main reason I'm interested in this is that this method is like 95% responsible for calling the methods in `appConnector` that use locks, but it's not called in a concurrent context, so the locking doesn't seem strictly necessary.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+	}
+
+	var errs []error
+	clients := make(map[int64]*github.Client, len(installs))
+	installedRepoKeys := make(map[int64]map[string]struct{}, len(installs))
+	for _, install := range installs {
+		installID := install.GetID()
+		client, err := connector.APIClientForInstallation(installID)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("could not create API client for installation %d: %w", installID, err))
```
camgunz: [`errors.Join`](https://pkg.go.dev/errors#Join) is a better pattern for this kind of thing, and then you won't need a `logInstallationMappingErrors` helper, ex:

```
var errs error
defer func() {
  if errs != nil {
    ctx.Logger().Error(errs, "some installations could not be fully enumerated while mapping repos")
  }
}()

// ...
errs = errors.Join(errs, fmt.Errorf("could not create API client for installation %d: %w", installID, err))
// ...

File: pkg/sources/github/github.go [hunk 29]
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+	if len(wantedRepos) > 0 {
+		unmatched := make([]string, 0, len(wantedRepos))
+		for _, requested := range wantedRepos {
+			unmatched = append(unmatched, requested.original)
+		}
+		sort.Strings(unmatched)
+		errs = append(errs, fmt.Errorf("configured repos were not found in any GitHub App installation: %s", strings.Join(unmatched, ", ")))
+	}
+
+	return errors.Join(errs...)
```
camgunz: Our [logging guidance](https://app.notion.com/p/Logging-Best-Practices-dcb5323c1fcc48d9881f9bff16cab2a2?source=copy_link) prefers choosing only one of logging an error or returning it, but we're doing both here, raising the question of whether or not we want an error here to stop scanning. This could cause existing scans that are succeeding to suddenly start failing if we run into issues trying to map other installations we weren't enumerating before.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+	}
+	_, otherParts, err := getRepoURLParts(repoURLForInstallationLookup(otherRepoURL))
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(parts[0], otherParts[0])
+}
+
+func githubWebHost(endpoint string) string {
+	if endpoint == "" || endsWithGithub.MatchString(endpoint) {
```
camgunz: We play pretty fast and loose with URL things in this source, so this doesn't really stand out. But I think it's a little wacky I could pass something like `legithub.com` to this function and get `github.com` back.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
 	return nil
 }
 
+func (s *Source) listAppInstallations(ctx context.Context, connector *appConnector, perPage int) ([]*github.Installation, error) {
```
camgunz: We try and keep line length < 120

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+
+	return allInstalls, nil
+}
+
+// enumerateAllInstallationRepos discovers repos from every installation of the
+// GitHub App. For each installation, it creates a per-installation API client
+// and records the installation ID for each repo so Clone uses the correct token.
+// Per-installation failures are reported through the unit reporter and don't
+// abort enumeration of the remaining installations.
+func (s *Source) enumerateAllInstallationRepos(ctx context.Context, connector *appConnector, reporter sources.UnitReporter) error {
```
camgunz, line 776: I'm curious why we have both `enumerateAllInstallationRepos` and `mapReposToInstallations`. They're both called by `Enumerate` (called by `Chunks`) and they're pretty similar: they both call `listAppInstallations`, they both call `connector.setRepoInstallation`, they both call `client.Apps.ListRepos`, they both call `cacheRepoInfo`, etc. Is there maybe a way we can pull these out and combine them, or like, does it indicate that because we're always doing these same initialization steps, we could list and execute them really explicitly without having these long control flow chains:

- `Enumerate`
  - `mapExplicitReposToInstallations`
    - `mapReposToInstallations`
      - **`listAppInstallations`** / **`client.Apps.ListRepos`**
      - `recordInstallationRepoMatches`
        - `recordRepoInstallation`
          - **`setRepoInstallation`** / **`cacheRepoInfo`**
  - `enumerateWithApp`
    - `enumerateAllInstallationRepos`
      - **`listAppInstallations`**
      - `processRepos`
        - **`setRepoInstallation`** / **`client.Apps.ListRepos`** / **`cacheRepoInfo`**

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+}
+
+func (s *Source) repoURLFromTargetMetadata(meta *source_metadatapb.Github) (string, error) {
+	linkRepo, err := s.repoURLFromTargetLink(meta.GetLink())
+	if err != nil {
+		return "", err
+	}
+
+	if repo := meta.GetRepository(); repo != "" {
+		if !strings.Contains(repo, "/") && !hasURLScheme.MatchString(repo) && !strings.Contains(repo, ":") {
```
camgunz: I think because `hasURLScheme` ends in `://` there's some redundancy in these cases. Also maybe it's worth documenting in a comment what we're filtering on here.

pkg/sources/github/github.go
```diff
@@ -663,6 +683,378 @@ func (s *Source) enumerateWithApp(ctx context.Context, installationClient *githu
...
+
+	segments := strings.Split(strings.Trim(u.Path, "/"), "/")
+	if len(segments) < 2 {
+		return "", fmt.Errorf("invalid GitHub URL")
+	}
+
+	repoURL := (&url.URL{
+		Scheme: u.Scheme,
+		Host:   u.Host,
+		Path:   fmt.Sprintf("/%s/%s.git", segments[0], segments[1]),
```
camgunz: We've had a lot of bugs in various sources related to using `Sprintf` for URL munging, so we prefer using things like [`URL.JoinPath`](https://pkg.go.dev/net/url#URL.JoinPath) whenever we can.

pkg/sources/github/github.go
```diff
@@ -938,23 +1330,38 @@ func (s *Source) addUserGistsToCache(ctx context.Context, user string, reporter
...
 
-func (s *Source) addMembersByApp(ctx context.Context, installationClient *github.Client, reporter sources.UnitReporter) error {
-	opts := &github.ListOptions{
-		PerPage: membersAppPagination,
-	}
-
+func (s *Source) addMembersByApp(ctx context.Context, connector *appConnector, reporter sources.UnitReporter) error {
 	// TODO: Check rate limit for this call.
-	installs, _, err := installationClient.Apps.ListInstallations(ctx, opts)
+	installs, err := s.listAppInstallations(ctx, connector, membersAppPagination)
```
camgunz: I think this means we'll potentially make a 3rd `listAppInstallations` call when enumerating.

pkg/sources/github/github.go
```diff
@@ -938,23 +1330,38 @@ func (s *Source) addUserGistsToCache(ctx context.Context, user string, reporter
...
 		}
-		if err := s.addMembersByOrg(ctx, *org.Account.Login, reporter); err != nil {
-			return err
+		org := install.GetAccount().GetLogin()
+		installCtx := context.WithValues(ctx, "org", org, "installation_id", install.GetID())
+
+		// Create a per-installation API client so that the token is scoped to
+		// this org. GitHub App installation tokens only bypass IP allowlists for
+		// their own org; using a cross-org token causes 403s.
+		client, err := connector.APIClientForInstallation(install.GetID())
```
camgunz, line 1446: This (in most cases) relies on us having run `mapReposToInstallations` to populate the connector, and while that is currently true nothing enforces it, so it's kind of a trap for future refactor-ers.

pkg/sources/github/repo.go
```diff
@@ -338,13 +345,26 @@ func (s *Source) cacheGistInfo(g *github.Gist) {
...
+		}).String() + ".git"
+	}
+	wikiURL := strings.TrimSuffix(wikiBaseURL, ".git") + "/wiki"
```
camgunz: Up in `connector_app.go:194` we do `wikiURL := strings.TrimSuffix(repoURL, ".git") + ".wiki.git"`, which is slightly different; does it matter? We also kind of do the opposite in `github.go` (`repoURLForInstallationLookup`): `strings.TrimSuffix(repoURL, ".wiki.git") + ".git"`, if that's useful information.

pkg/sources/github/repo.go
```diff
@@ -359,12 +379,22 @@ func (s *Source) wikiIsReachable(ctx context.Context, repoURL string) bool {
 func (s *Source) normalizeRepo(repo string) (string, error) {
 
 	// If it's a full URL (has protocol), normalize it
-	if regexp.MustCompile(`^[a-z]+://`).MatchString(repo) {
-
+	if hasURLScheme.MatchString(repo) {
 		return giturl.NormalizeGithubRepo(repo)
 	}
+	if strings.Contains(repo, "@") && strings.Contains(repo, ":") {
```
camgunz: Do we do this because we have a hard time with user/pass info? Mostly I'm just concerned that looking for `":"` could either be the scheme or the password separator, which makes this check a little ambiguous.

camgunz (approved):
I don't think there's any real issues, so I think we can merge. I wrote a whole bunch, mostly in the interest of just getting any opportunity for improvement out there, but also just noting things as I dug through the PR and the GitHub source.

I think if I had to pick one thing that would simplify this new functionality it would be to do the installation enumerating and repo -> installation mapping during initialization, guaranteed outside of a concurrent context. This way we wouldn't need to worry so much about locking, or helper method call order, or whether we're calling remote APIs more than we need to, etc. A bonus 2nd thing would be to modify the `Connector` interface to accept a repository URL, which would cut down on a lot of the type assertion branching we need to do otherwise.

dustin-decker:
Thanks for the detailed pass here. I pushed a follow-up that takes the broader cleanup path:

- `Connector` now has repo-aware API/GraphQL client methods, so the `*appConnector` branching is mostly pushed behind the connector boundary.
- `scan_all_installations` now lists app installations once during `Init`, caches that list, and maps explicit repos during initialization before metadata fetching.
- Enumeration/member/mapping paths reuse the cached installation list, so we avoid repeated `ListInstallations` calls.
- Partial installation mapping errors now use `errors.Join` while keeping the best-effort behavior: continue if all requested repos are mapped, fail if requested repos cannot be mapped.
- Added tests for init-time mapping/failures, cached installation reuse, non-app connector behavior, and GitHub host detection.

Validation passed locally: `go vet ./pkg/sources/github`, `go test ./pkg/sources/github`, and `make lint`.

pkg/sources/github/github.go
dustin-decker: Agreed. I updated `Connector` to expose `APIClientForRepo` and `GraphQLClientForRepo`, with non-app connectors delegating to their default clients. That removes the source-level app-specific helper branching and lets `appConnector` own repo-scoped client selection.

pkg/sources/github/github.go
dustin-decker: Yep, this was the main cleanup in the follow-up. Installations are now loaded once during `Init`, explicit repos are mapped during `Init`, and later paths reuse the cached installation list.

I kept `enumerateAllInstallationRepos` and `mapReposToInstallations` separate because they still serve different outputs: discovery enumerates and reports all repos, while mapping resolves configured/targeted/resume repos. They now share the installation list, so the repeated `ListInstallations` concern should be addressed.

pkg/sources/github/github.go
dustin-decker: Yep, this was the main cleanup in the follow-up. Installations are now loaded once during `Init`, explicit repos are mapped during `Init`, and later paths reuse the cached installation list.

I kept `enumerateAllInstallationRepos` and `mapReposToInstallations` separate because they still serve different outputs: discovery enumerates and reports all repos, while mapping resolves configured/targeted/resume repos. They now share the installation list, so the repeated `ListInstallations` concern should be addressed.

pkg/sources/github/github.go
dustin-decker: Good call. This now uses `errors.Join` for partial installation errors. The behavior is still best-effort: if all requested repos are mapped, we log joined partial installation errors and continue; if requested repos remain unmapped, we return the joined error.

pkg/sources/github/github.go
dustin-decker: Good call. This now uses `errors.Join` for partial installation errors. The behavior is still best-effort: if all requested repos are mapped, we log joined partial installation errors and continue; if requested repos remain unmapped, we return the joined error.

pkg/sources/github/github.go
dustin-decker: Fixed by caching installations during `Init`. Enumeration, explicit mapping, and member enumeration now reuse that cached list instead of each calling `ListInstallations`.

pkg/sources/github/github.go
dustin-decker: Fixed this specific case and added coverage. The GitHub cloud host check now avoids treating hosts like `legithub.com` as `github.com`.

pkg/sources/github/github.go
dustin-decker: Updated the target-link repo URL construction to use `url.URL.JoinPath`, and added/kept tests around the targeted scan URL behavior.

pkg/sources/github/repo.go
dustin-decker: I made the clone-side wiki mapping explicit with a helper and documented the inverse lookup path. The intended behavior is: repository clone URL maps to `repo.wiki.git` for clone/token selection, and wiki clone URLs map back to the owning repo URL for installation lookup.

pkg/sources/github/repo.go
dustin-decker: That check is specifically for SCP-style SSH repo strings like `git@github.com:org/repo.git`. I kept that path because `GitURLParse` handles the user/host/path pieces correctly there, but target metadata now has a comment explaining when we treat a repository value as a bare repo name versus something that needs normalization.

pkg/sources/github/github.go
dustin-decker: Fixed in the follow-up.

pkg/sources/github/github.go
dustin-decker: Cleaned this up in the follow-up: the redundant scheme check is gone, and there is now a comment explaining that bare repository names use the owner/host from the link as the authoritative identity.

pkg/sources/github/github.go
dustin-decker: I kept the behavior intentionally best-effort where possible. Missing/unmatched configured repos still fail because otherwise we'd fall back to the default installation token and potentially scan with the wrong auth context. Partial installation failures are joined/logged when all requested repos were mapped, so unrelated installation issues don't fail an otherwise valid explicit-repo scan.

pkg/sources/github/github.go
dustin-decker: This is less fragile after the follow-up: explicit repos are mapped during `Init`, discovery maps repos during enumeration, and dynamic `ChunkUnit`/targeted paths still lazily map before metadata/API access. The connector API now accepts repo URLs too, so callers don't need to know whether the underlying auth mode has per-repo installation routing.

pkg/sources/github/connector_app.go
dustin-decker: Agree that a per-installation bundle would be cleaner long-term. I scoped this follow-up to the repo-aware routing and installation init/cache changes without restructuring the connector internals further. Separate maps keep the diff smaller, and the default `apiClient`/`graphqlClient` fields preserve the existing configured-installation fast path. Happy to do the unified bundle/single-lock refactor separately if you'd like.

pkg/sources/github/connector_app.go
dustin-decker: Fair point on the mental model: REST and GraphQL clients for an installation should stay paired. The GraphQL path uses the double-check pattern so it never holds `graphqlClientsMu` while taking the API-client lock. Duplicate GraphQL construction on a race would be harmless, and this path is not concurrent in normal enumeration today. I'd rather keep the broader unified client-bundle refactor separate from this PR.

pkg/sources/github/connector_app.go
dustin-decker: Agreed. `NewAppConnector` initializes the maps, so these checks are defensive rather than required. I left them alone to keep this pass focused, but I'm fine dropping them in a small cleanup if you prefer constructor-only initialization.

pkg/sources/github/connector_app.go
dustin-decker: Intentional for the current `Connector` shape: the repo-aware client methods still do not return errors, so unknown repos fall back to the configured installation. The fail-closed behavior now happens earlier for configured/targeted repos during installation mapping, which avoids silently scanning an explicitly requested repo with the wrong token. Making this lookup erroring would ripple through all connector implementations.

pkg/sources/github/connector_app.go
dustin-decker: Yes, there is overlap. I kept it inline to avoid changing constructor behavior while addressing the review feedback, but extracting shared transport/client setup into a helper would be a reasonable follow-up cleanup.

pkg/sources/github/connector_app.go
dustin-decker: Yes, `wikiCloneURLForRepo` only derives `*.wiki.git` when the repo URL is already a clone-style `.git` URL, which is what we get from app repo enumeration. Non-`.git` inputs are not expected on that path, and the inverse installation lookup handles wiki clone URLs explicitly.

pkg/sources/github/repo.go
dustin-decker: The `.git` in that expression matches clone URL shape before stripping it for the web `/wiki` reachability path, so behavior is the same either way. I agree `JoinPath(owner, name, "wiki")` would read more directly; I’d treat that as cosmetic cleanup rather than part of the auth/mapping fix.

pkg/sources/github/github.go
dustin-decker: Noted. I kept `endsWithGithub` and `hasURLScheme` close to the existing GitHub source patterns for this PR, aside from tightening the `github.com` host match. A broader URL-normalization cleanup would be worthwhile separately.

pkg/sources/github/github.go
```diff
@@ -663,6 +693,412 @@ func (s *Source) enumerateWithApp(ctx context.Context, connector *appConnector,
...
+			installCtx.Logger().Error(err, "error enumerating repos for installation")
+			if err := reporter.UnitErr(installCtx, fmt.Errorf("error enumerating repos for installation %d: %w", installID, err)); err != nil {
+				return err
+			}
+			continue
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
dustin-decker: This overlap is intentional for now: the two methods produce different outputs (`Enumerate` needs discovery units, while mapping resolves configured/targeted/resume repos to installations). The follow-up already removed the expensive duplication by caching installations during `Init`; both paths now reuse that list. I’d keep deeper unification as a separate refactor rather than expand this PR further.

pkg/sources/github/repo.go
dustin-decker: Followed up in `ff341f7eb`: wiki web URLs now use shared helpers and `JoinPath(owner, name, "wiki")` instead of constructing a clone-shaped URL and trimming `.git`. The reachability probe also no longer uses an authenticated API client for the web HEAD request.

pkg/sources/github/connector_app.go
dustin-decker: Implemented in `ff341f7eb`. REST and GraphQL clients now live in the same per-installation bundle, so the GraphQL path is paired with the API client for the same installation. Client creation happens outside the exclusive lock and stores with a double-check.

pkg/sources/github/connector_app.go
dustin-decker: Implemented in `ff341f7eb`. `appConnector` now has a single `clientsByInstallationID map[int64]*appInstallationClients`, where each entry owns both REST and GraphQL clients for that installation. The separate top-level `apiClient` / `graphqlClient` fields are gone, and connector state is guarded by one `RWMutex`.

pkg/sources/github/connector_app.go
dustin-decker: Cleaned up in `ff341f7eb`. The defensive nil-map initialization is gone from connector methods; `NewAppConnector` and the test fixtures now own initialization of the connector maps.

pkg/sources/github/connector_app.go
dustin-decker: Followed up in `ff341f7eb` by making the fallback explicit without changing the public connector contract. `installationIDForRepo` now returns `(id, mapped)`, and tests document mapped vs. fallback behavior. Configured/targeted repos still fail closed during mapping if they cannot be found.

pkg/sources/github/connector_app.go
dustin-decker: Followed up in `ff341f7eb`. The wiki clone helper is now centralized and refuses already-wiki clone URLs, so we don't derive `repo.wiki.wiki.git`. Non-`.git` repo URLs still do not derive wiki clone URLs because the installation mapping path uses clone URLs from GitHub.

pkg/sources/github/connector_app.go
dustin-decker: Addressed in `ff341f7eb`. `NewAppConnector` and per-installation API client creation now share small helpers for app transport and GitHub client construction, so the transport setup is no longer duplicated inline.

pkg/sources/github/github.go
dustin-decker: Followed up in `ff341f7eb`. The old endpoint/scheme regexes are replaced with URL helper logic for cloud vs. GHE endpoints, URL scheme detection, and SCP-style repo detection. Added coverage for `legithub.com`, schemeless GHE endpoints, credentialed HTTPS, and SCP-style repo strings.

pkg/sources/github/github.go
dustin-decker: Yes, this changed in `ff341f7eb`. I kept `enumerateAllInstallationRepos` and `mapReposToInstallations` as separate flows because they still have different outputs and error semantics, but extracted the shared `Apps.ListRepos` page listing/rate-limit loop. That removes the duplicated API mechanics without merging discovery and explicit mapping behavior.

pkg/sources/github/repo.go
dustin-decker: Followed up in `ff341f7eb` by centralizing the wiki URL conversions. Repo clone URLs derive `repo.wiki.git`, wiki clone URLs normalize back to the owning repo for installation lookup, and already-wiki URLs are guarded so we do not create `repo.wiki.wiki.git`.

pkg/sources/github/repo.go
dustin-decker: Tightened this in `ff341f7eb` with an explicit SCP-style helper: we only treat `user@host:owner/repo.git` as SCP-style when there is no URL scheme. Credentialed HTTPS and `ssh://` URLs stay on the normal URL path.

pkg/sources/github/connector_app.go
```diff
@@ -108,15 +203,67 @@ func (c *appConnector) InstallationClient() *github.Client {
...
+func (c *appConnector) createAPIClientForInstallation(installationID int64) (*github.Client, error) {
+	appsTransport, err := newAppsTransport(c.apiEndpoint, c.appID, c.appPrivateKey)
+	if err != nil {
+		return nil, fmt.Errorf("could not create app transport for installation %d: %w", installationID, err)
+	}
+
+	transport := ghinstallation.NewFromAppsTransport(appsTransport, installationID)
+	transport.BaseURL = c.apiEndpoint
+
+	client, err := newGitHubClientWithTransport(c.apiEndpoint, transport)
```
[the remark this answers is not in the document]
dustin-decker: No code change here. This is intentional: the shared helper in `ff341f7eb` deduplicates construction code, but it does not share a single `AppsTransport` instance across cached installation clients. Earlier in this PR we moved to per-client app transports to avoid sharing token-refresh state across installation clients; this follow-up keeps that behavior while cleaning up the setup code.

camgunz (approved):
Hell yeah! Thanks so much, love it ✅

pkg/sources/github/urls.go
```diff
@@ -0,0 +1,200 @@
...
+	"net/url"
+	"strings"
+)
+
+func wikiCloneURLForRepo(repoURL string) (string, bool) {
+	if !strings.HasSuffix(repoURL, ".git") {
+		return "", false
+	}
+	return strings.TrimSuffix(repoURL, ".git") + ".wiki.git", true
+}
```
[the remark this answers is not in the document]
dustin-decker: Fixed in the latest push by keeping wikiCloneURLForRepo conservative for existing .wiki.git URLs and only deriving .wiki.wiki.git aliases when repository metadata confirms the repo itself is named with a .wiki suffix.

pkg/sources/github/github.go
```diff
@@ -364,17 +380,22 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, tar
...
+		},
 	}
 	err := s.Enumerate(ctx, noopReporter)
 	if err != nil {
-		return fmt.Errorf("error enumerating: %w", err)
+		return errors.Join(fmt.Errorf("error enumerating: %w", err), errors.Join(enumerationErrs...))
 	}
 
-	return s.scan(ctx, chunksReporter)
+	return errors.Join(append(enumerationErrs, s.scan(ctx, chunksReporter))...)
```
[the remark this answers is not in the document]
dustin-decker: This behavior is intentional. For scan_all_installations, if one installation cannot be enumerated, the scan result is incomplete even if the successfully enumerated repos scan cleanly, so Chunks should return an error instead of silently reporting success.

pkg/sources/github/urls.go
```diff
@@ -0,0 +1,204 @@
...
+	return wikiCloneURLForRepoName(repoURL, info.name)
+}
+
+func wikiCloneURLForRepoName(repoURL, name string) (string, bool) {
+	_, parts, err := getRepoURLParts(repoURL)
+	if err != nil || len(parts) != 3 || !strings.EqualFold(parts[2], name) || !strings.HasSuffix(repoURL, ".git") {
+		return "", false
+	}
+	return strings.TrimSuffix(repoURL, ".git") + ".wiki.git", true
+}
```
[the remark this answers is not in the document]
dustin-decker: This is intentional for repositories actually named with a .wiki suffix. GitHub wiki clone URLs are formed by appending .wiki.git to the repository clone base, so the wiki for a repo named foo.wiki is foo.wiki.wiki.git. The conservative helper still refuses already-wiki inputs; this metadata-aware helper only derives the double-wiki URL after confirming the repository name itself is foo.wiki, and TestMapExplicitReposToInstallationsPrefersRealWikiSuffixedRepo / TestWikiCloneURLForRepoInfo cover that case.

PR #4780, opened 2026-03-03
Thread original chunk data through engine pipeline

dustin-decker:
## Summary

Adds fields to preserve original pre-decode chunk data through the scan pipeline so downstream consumers (thog dispatcher) can access it for secret storage encryption.

**3 files, 1 field per stage:**

1. `Chunk.OriginalData` (`pkg/sources/sources.go`) -- new `[]byte` field on the `Chunk` struct
2. `engine.go` -- sets `chunk.OriginalData = chunk.Data` before `iterativeDecode`, so it retains the original content even after `Data` is replaced with decoded forms
3. `ResultWithMetadata.ChunkData` (`pkg/detectors/detectors.go`) -- populated by `CopyMetadata` from `OriginalData` (falls back to `Data` when nil)

`OriginalData` propagates through `iterativeDecode` naturally because `chunkCopy := *chunk` copies all fields and only `chunkCopy.Data` is replaced.

Corresponding thog-side PR: trufflesecurity/thog#5738

## Test plan

- [x] `CopyMetadata` populates `ChunkData` from `OriginalData` when present
- [x] `CopyMetadata` falls back to `Data` when `OriginalData` is nil
- [x] Additive change -- no existing behavior modified, all existing tests pass

---

> [!NOTE]
> **Medium Risk**
> Touches the core scan pipeline and increases per-chunk memory footprint by threading additional data through `Chunk` and `ResultWithMetadata`, which could affect performance/memory or downstream consumers if assumptions about chunk contents change.
> 
> **Overview**
> Preserves *pre-decode* source bytes through the scan pipeline by adding `sources.Chunk.OriginalData` and setting it in `engine.scannerWorker` before `iterativeDecode` mutates `Chunk.Data`.
> 
> Threads this original content into detector output by adding `ResultWithMetadata.ChunkData` and updating `CopyMetadata` to prefer `OriginalData` (fallback to `Data`), and ensures decoders (e.g. `EscapedUnicode`) carry `OriginalData` forward when emitting decoded chunks.
> 
> Adds unit tests for the `CopyMetadata` behavior and updates the `Chunk` struct size guard to account for the new field.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ed42c05ca813a984804d036f6dbdf98bd34f8d02. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/decoders/escaped_unicode.go
+++ b/pkg/decoders/escaped_unicode.go
@@ -105,6 +105,7 @@ func (d *EscapedUnicode) FromChunk(chunk *sources.Chunk) *DecodableChunk {
 			DecoderType: d.Type(),
 			Chunk: &sources.Chunk{
 				Data:           chunkData,
+				OriginalData:   chunk.OriginalData,
 				SourceName:     chunk.SourceName,
 				SourceID:       chunk.SourceID,
 				JobID:          chunk.JobID,
--- a/pkg/detectors/copy_metadata_test.go
+++ b/pkg/detectors/copy_metadata_test.go
@@ -3,6 +3,8 @@ package detectors
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
 
@@ -19,9 +21,7 @@ func TestCopyMetadata_ChunkDataFromOriginalData(t *testing.T) {
 
 	rwm := CopyMetadata(chunk, result)
 
-	if string(rwm.ChunkData) != "original-source-data" {
-		t.Errorf("ChunkData = %q, want %q", rwm.ChunkData, "original-source-data")
-	}
+	assert.Equal(t, "original-source-data", string(rwm.ChunkData))
 }
 
 func TestCopyMetadata_ChunkDataFallsBackToData(t *testing.T) {
@@ -36,7 +36,5 @@ func TestCopyMetadata_ChunkDataFallsBackToData(t *testing.T) {
 
 	rwm := CopyMetadata(chunk, result)
 
-	if string(rwm.ChunkData) != "only-data" {
-		t.Errorf("ChunkData = %q, want %q", rwm.ChunkData, "only-data")
-	}
+	assert.Equal(t, "only-data", string(rwm.ChunkData))
 }
--- a/pkg/detectors/copy_metadata_test.go
+++ b/pkg/detectors/copy_metadata_test.go
@@ -1,0 +1,42 @@
+package detectors
+
+import (
+	"testing"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+func TestCopyMetadata_ChunkDataFromOriginalData(t *testing.T) {
+	chunk := &sources.Chunk{
+		Data:         []byte("decoded-data"),
+		OriginalData: []byte("original-source-data"),
+		SourceName:   "test-source",
+	}
+	result := Result{
+		DetectorType: 1,
+		Raw:          []byte("secret"),
+	}
+
+	rwm := CopyMetadata(chunk, result)
+
+	if string(rwm.ChunkData) != "original-source-data" {
+		t.Errorf("ChunkData = %q, want %q", rwm.ChunkData, "original-source-data")
+	}
+}
+
+func TestCopyMetadata_ChunkDataFallsBackToData(t *testing.T) {
+	chunk := &sources.Chunk{
+		Data:       []byte("only-data"),
+		SourceName: "test-source",
+	}
+	result := Result{
+		DetectorType: 1,
+		Raw:          []byte("secret"),
+	}
+
+	rwm := CopyMetadata(chunk, result)
+
+	if string(rwm.ChunkData) != "only-data" {
+		t.Errorf("ChunkData = %q, want %q", rwm.ChunkData, "only-data")
+	}
+}
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -224,6 +224,8 @@ type ResultWithMetadata struct {
 
 // CopyMetadata returns a detector result with included metadata from the source chunk.
 func CopyMetadata(chunk *sources.Chunk, result Result) ResultWithMetadata {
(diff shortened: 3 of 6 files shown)


dustin-decker:
* **#4780** <a href="https://app.graphite.com/github/pr/trufflesecurity/trufflehog/4780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/trufflesecurity/trufflehog/4780?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

pkg/decoders/escaped_unicode.go
```diff
@@ -105,6 +105,7 @@ func (d *EscapedUnicode) FromChunk(chunk *sources.Chunk) *DecodableChunk {
 			DecoderType: d.Type(),
 			Chunk: &sources.Chunk{
 				Data:           chunkData,
+				OriginalData:   chunk.OriginalData,
```
rosecodym, line 108: All of the other decoders need to be updated as well.

pkg/detectors/copy_metadata_test.go
```diff
@@ -0,0 +1,42 @@
...
+	result := Result{
+		DetectorType: 1,
+		Raw:          []byte("secret"),
+	}
+
+	rwm := CopyMetadata(chunk, result)
+
+	if string(rwm.ChunkData) != "original-source-data" {
+		t.Errorf("ChunkData = %q, want %q", rwm.ChunkData, "original-source-data")
+	}
```
rosecodym: This should use `testify/assert`

pkg/detectors/copy_metadata_test.go
```diff
@@ -0,0 +1,42 @@
...
+	result := Result{
+		DetectorType: 1,
+		Raw:          []byte("secret"),
+	}
+
+	rwm := CopyMetadata(chunk, result)
+
+	if string(rwm.ChunkData) != "only-data" {
+		t.Errorf("ChunkData = %q, want %q", rwm.ChunkData, "only-data")
+	}
```
rosecodym: This should use`testify/assert`

pkg/detectors/detectors.go
```diff
@@ -217,10 +217,17 @@ type ResultWithMetadata struct {
...
+	// for secret storage encryption in the dispatcher.
+	ChunkData []byte
 }
 
 // CopyMetadata returns a detector result with included metadata from the source chunk.
 func CopyMetadata(chunk *sources.Chunk, result Result) ResultWithMetadata {
+	chunkData := chunk.OriginalData
+	if chunkData == nil {
+		chunkData = chunk.Data
+	}
```
rosecodym, line 232: When does this happen?

pkg/sources/sources.go
rosecodym: This comment is no longer true and should be removed.

rosecodym (changes requested):
requesting changes for the decoder changes

pkg/decoders/escaped_unicode.go
dustin-decker: EscapedUnicode is the only decoder that constructs a new `sources.Chunk` (to avoid data races), which is why it needs the explicit `OriginalData` copy. The other three decoders (Base64, UTF8, UTF16) reuse the passed-in chunk pointer and only reassign `chunk.Data`, so `OriginalData` and all other fields are inherited automatically.

pkg/detectors/copy_metadata_test.go
dustin-decker: Done — switched to `testify/assert`.

pkg/detectors/copy_metadata_test.go
dustin-decker: Done — switched to `testify/assert`.

pkg/detectors/detectors.go
dustin-decker: Added an inline comment explaining this. OriginalData is always set in the engine's scannerWorker before decode, so this nil case only occurs if CopyMetadata is called outside the normal engine pipeline (e.g., in tests or external consumers that construct chunks directly).

pkg/sources/sources.go
dustin-decker: Removed.

rosecodym (approved):

pkg/sources/sources.go
```diff
@@ -21,14 +21,13 @@ type (
 )
 
 // Chunk contains data to be decoded and scanned along with context on where it came from.
-//
-// **Important:** The order of the fields in this struct is specifically designed to optimize
-// struct alignment and minimize memory usage. Do not change the field order without carefully considering
-// the potential impact on memory consumption.
-// Ex: https://go.dev/play/p/Azf4a7O-DhC
```
shahzadhaider1, line 28: this doesn't apply anymore?

shahzadhaider1 (approved):

pkg/sources/sources.go
```diff
@@ -21,14 +21,13 @@ type (
...
 type Chunk struct {
 	// Data is the data to decode and scan.
 	Data []byte
+	// OriginalData holds the pre-decode source data, preserved for secret
+	// storage. Set before iterative decoding so it retains the original
+	// content even after Data is replaced with decoded forms.
+	OriginalData []byte
```
trufflesteeeve, line 30: As described in the deleted comment above, this does significantly increase the memory size of chunk. Going from 80 bytes to 104 bytes. Are we okay with that?

trufflesteeeve (approved):
I think we should probably leave in the deleted comment about chunk memory size. But other than that, this looks fine to me.

pkg/sources/sources.go
shahzadhaider1: oh, just saw [this](https://github.com/trufflesecurity/trufflehog/pull/4780#discussion_r2878483847)

PR #4781, opened 2026-03-03
[INS-345] Add New Relic Insights Query Key detector

mustansir14:
### Description:
This PR adds the New Relic Insights Query Key Detector.

**Regex**: 
Key: `\b(NRIQ-[a-zA-Z0-9-_]{25})`
Account ID: `detectors.PrefixRegex([]string{"relic", "account", "id"}) + '\b(\d{4,10})\b'`
The key is the actual credential but account ID is required for verification because the verification endpoint requires an account ID in the path, and there is no other deterministic way to verify the credential without specifying the valid account ID (invalid or malformed account IDs return the same response as an invalid key)

**Verification**:
For verification, we use the Insights Query API with a simple select query: `https://insights-api.newrelic.com/v1/accounts/[account_id]/query?nrql=SELECT%%201`. 
We send a GET request. A response code of `200` means the key is valid. `401` means it is an invalid/rotated key.
Note: For EU region keys, the host should be `insights-api.eu.newrelic.com`

**Corpora Test**:
image
image

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive, feature-flagged detector with read-only verification HTTP calls; no changes to core scan or auth paths beyond registering the scanner.
> 
> **Overview**
> Adds **New Relic Insights Query Key** (`NRIQ-…`) detection to TruffleHog, alongside the existing New Relic detectors.
> 
> The scanner matches keys with `NRIQ-[a-zA-Z0-9-_]{25}` and pairs them with a nearby **account ID** (prefix-aware regex). Results expose `key`, `account_id`, and optional `region` in `SecretParts`. When verification is on, it calls the legacy Insights Query API with `X-Query-Key` and a minimal `SELECT 1` NRQL query, trying **US** (`insights-api.newrelic.com`) and **EU** (`insights-api.eu.newrelic.com`) until one returns 200.
> 
> Wiring follows other gated detectors: new `DetectorType_NewRelicInsightsQueryKey` (1067), registration in `defaults`, `NewRelicInsightsQueryKeyDetectorEnabled` enabled from `main`, and unit/integration/benchmark tests for the new package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80ad5e0986564979b9a8e185ab21ecd489ec6317. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
```diff
@@ -0,0 +1,124 @@
...
+		req.Header.Set("X-Query-Key", key)
+
+		res, err := client.Do(req)
+		if err != nil {
+			return false, nil, fmt.Errorf("error making request: %w", err)
+		}
+		defer func() {
+			_, _ = io.Copy(io.Discard, res.Body)
+			_ = res.Body.Close()
+		}()
```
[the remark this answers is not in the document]
mustansir14: It is fine because there are always at max two iterations of the loop.

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
```diff
@@ -0,0 +1,126 @@
...
+	dataStr := string(data)
+
+	keyMatches := keyPat.FindAllStringSubmatch(dataStr, -1)
+	accountIDMatches := accountIDPat.FindAllStringSubmatch(dataStr, -1)
+	uniqueAccountIDMatches := make(map[string]struct{})
+	for _, match := range accountIDMatches {
+		uniqueAccountIDMatches[match[1]] = struct{}{}
+	}
+
+	for _, keyMatch := range keyMatches {
```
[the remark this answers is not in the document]
mustansir14: This is intended. The idea is to report multiple results if the key appears in multiple places, so that the user can remove it from all of those places.

Account ID matches are deduplicated because they are not the primary credential.

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// verify checks if the provided key is valid by making a request to the New Relic Insights Query API.
+// It checks both the US and EU endpoints before returning an error.
+// Account ID is required to verify as the API endpoint is account-specific.
+func (s Scanner) verify(ctx context.Context, key string, accountID string) (bool, map[string]string, error) {
+	regionUrls := map[string]string{
+		"us": fmt.Sprintf("https://insights-api.newrelic.com/v1/accounts/%s/query?nrql=SELECT%%201", accountID),
+		"eu": fmt.Sprintf("https://insights-api.eu.newrelic.com/v1/accounts/%s/query?nrql=SELECT%%201", accountID),
+	}
```
shahzadhaider1, line 103: (optional) suggestion: use `net/url` insted of `fmt.Sprintf()`

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
```diff
@@ -0,0 +1,126 @@
...
+			ctx, http.MethodGet, regionUrl, http.NoBody)
+		if err != nil {
+			return false, nil, fmt.Errorf("error constructing request: %w", err)
+		}
+		req.Header.Set("X-Query-Key", key)
+
+		res, err := client.Do(req)
+		if err != nil {
+			return false, nil, fmt.Errorf("error making request: %w", err)
+		}
```
shahzadhaider1: If the US request times out or fails at the network level, you bail out entirely without trying the EU region endpoint. You probably want to try the next region instead?

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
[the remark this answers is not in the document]
shahzadhaider1: Yeah, but don't you think if res1 body isn't drained before iteration 2, the connection can't be reused, forcing a new TCP connection unnecessarily. (defer inside a loop is a well-known Go anti-pattern BTW)

> It is fine because there are always at max two iterations of the loop.

I am wondering how will this behave in case multiple chunks contain secret of the same type?

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
mustansir14: I think that would unnecessarily complicate the code. We prefer `net/url` when we are concatenating user inputted URLs. Here we are simply inserting the account ID (which also comes via a strict regex so it does not have any chances of being malformed) into a URL.

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
mustansir14: Hmm good catch. I agree. I'll just extract the inside of the loop into a separate method, which should also fix the deferred response body cleanup issue by bugbot.

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
```diff
@@ -0,0 +1,137 @@
...
+				isVerified, extraData, verificationErr := s.verify(ctx, keyResMatch, accountIDResMatch)
+				s1.Verified = isVerified
+				s1.ExtraData = extraData
+				s1.SetVerificationError(verificationErr)
+			}
+
+			results = append(results, s1)
+		}
+
+	}
```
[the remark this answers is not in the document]
mustansir14: Account ID is required for verification. The atlassian detector you mentioned captures the `organization_id` only to pass it into `AnalysisInfo`. It's not needed for verification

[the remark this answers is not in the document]
shahzadhaider1: So, the key is useless without an account ID? It can't be used to access any sort of data on its own?

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
```diff
@@ -0,0 +1,137 @@
...
+			accountIDResMatch := strings.TrimSpace(accountID)
+
+			s1 := detectors.Result{
+				DetectorType: s.Type(),
+				Raw:          []byte(keyResMatch),
+				RawV2:        []byte(keyResMatch + accountIDResMatch),
+				Redacted:     keyResMatch[:8] + "...",
+			}
+
+			if verify && accountIDResMatch != "" {
```
shahzadhaider1: in what scenarios, `accountIDResMatch` can be empty at this point?

shahzadhaider1 (approved):

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
mustansir14: It cannot. Good catch. I'll remove the check

pkg/detectors/newrelicinsightsquerykey/newrelicinsightsquerykey.go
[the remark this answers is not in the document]
mustansir14: Yes, all endpoints that it can access require an account ID

proto/detector_type.proto
```diff
@@ -1065,4 +1065,5 @@ enum DetectorType {
   HashiCorpVaultBatchToken = 1061;
   HashiCorpVaultToken = 1062;
   Duo = 1063;
+  NewRelicInsightsQueryKey = 1064;
```
unsmith: Same thing, conflict with proto enum number.

unsmith (changes requested):
I think I agree with all the logic decisions here, but I admit I only brushed through it with robot help.  The proto enum conflict needs to be resolved here as well though.

proto/detector_type.proto
MuneebUllahKhan222: Will resolve them once the PR has been approved and is ready to be merged.

unsmith (approved):

PR #4790, opened 2026-03-05
Add Figma Personal Access Token v3 Detector

shahzadhaider1:
### Summary

Adds a v3 detector for Figma Personal Access Tokens covering the `figp_` prefixed token format, and consolidates verification for all three versions into a single shared function. Note this is **not a pure refactor** — it also fixes a verification bug affecting v1 and v2, described below.

### Changes

**New v3 detector (`figp_` prefix)**
- Regex: `(figp_[a-zA-Z0-9_=-]{40,54})(?:[^a-zA-Z0-9_=-]|\z)`
- Keyword pre-filter uses `figp_`.

**Shared verification**
- `VerifyMatch` now lives in the parent package `pkg/detectors/figmapersonalaccesstoken`; v1, v2 and v3 all import and call it.

**Verification semantics (behavior change for v1 and v2)**
- `GET https://api.figma.com/v1/me` with the `X-Figma-Token` header.
- Figma returns 403 for *every* authentication outcome, so the response body is the only discriminator. Previously any 403 was reported as definitively unverified, which meant a live token lacking the `current_user:read` scope that `/v1/me` requires was silently reported as dead.
- `200` -> verified.
- `403 {"err":"Invalid token"}` -> unverified, no error.
- `403 {"err":"Invalid scope(s): ..."}` -> **verified** (the token is live, it just lacks `current_user:read`).
- `403` with any other message -> verification error, so an unrecognised message surfaces rather than becoming a silent false negative.
- `401` -> unverified. Undocumented by Figma and never observed in testing; kept defensively.
- Any other status (including 429 throttling) -> verification error.

### Evidence

Tested against live tokens generated with distinct scopes. A revoked token returns `"Invalid token"`, and mutating any single character of a valid token also returns `"Invalid token"`, so that message reliably indicates a token that does not resolve.

### Open question on `figp_`

The `figp_` token I tested returns `403 {"err":"Request denied"}` on `/v1/me`, `/v1/files/...` and `/v1/teams/...` — never `"Invalid token"`, and mutating one character flips it to `"Invalid token"`, so Figma does recognise it. Figma's docs state plan access tokens cannot use `/v1/me`, which matches this behavior. If `figp_` is a plan access token rather than a personal one, `/v1/me` can never return 200 for it and this detector may be misnamed. Treated as indeterminate for now rather than verified, since expiry behavior is untested. Confirmation of the token type would be welcome.

### Testing
- Unit tests for the v3 pattern (valid matches and invalid rejections).
- Unit tests for `VerifyMatch` covering 200, the three real 403 bodies, an unparseable 403 body, 401, 429 and 500.
- v1 and v2 integration tests updated for new `detectors.Result` fields in the `cmpopts.IgnoreFields` list.
- v3 integration tests added but **skipped** pending confirmation of the `figp_` token type and provisioning of v3 secrets in GCP.

### Checklist:

image

image

image

image

image

---

> [!NOTE]
> **Medium Risk**
> Changes how v1/v2 Figma tokens are classified (verified vs unverified vs error), which can alter scan output; otherwise limited to detector/verification code with tests.
> 
> **Overview**
> Adds a **v3** Figma Personal Access Token detector for the `figp_` format (custom end anchoring for `=`/`-` in tokens) and wires it into default detectors behind **`FigmaV3DetectorEnabled`** (on by default in OSS).
> 
> **Shared verification** moves Figma API checks into `pkg/detectors/figmapersonalaccesstoken.VerifyMatch`; v1, v2, and v3 scanners call it instead of inlined HTTP logic. **Verification semantics change for v1/v2:** `GET /v1/me` with `X-Figma-Token` now treats `403` with `"Invalid scope(s):..."` as **verified** (live token missing `current_user:read`), `"Invalid token"` as unverified, and other `403` bodies as verification errors; non-403 unexpected statuses (e.g. 429) also surface as errors instead of being lumped with silent unverified 403s.
> 
> Unit tests cover `VerifyMatch` and v3 regex; v1/v2 integration `cmpopts` ignore lists are updated; v3 integration tests are present but skipped until GCP secrets exist.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 309a434fb043f019c8a6ba227b0e80c14591ec92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amanfcp (commented):
LGTM other than the comment below

pkg/detectors/figmapersonalaccesstoken/v2/figmapersonalaccesstoken_v2.go
```diff
@@ -57,28 +60,9 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-				if res.StatusCode >= 200 && res.StatusCode < 300 {
-					s1.Verified = true
-				} else if res.StatusCode != 403 {
-					err = fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
-					s1.SetVerificationError(err, resMatch)
-				}
-			} else {
-				s1.SetVerificationError(err, resMatch)
-			}
+			isVerified, verificationErr := v1.VerifyMatch(ctx, s.getClient(), resMatch)
```
amanfcp: Wouldn't it be better if we keep the `VerifyMatch` function outside v1? Something like [this](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/aws/common.go) and maybe name it `figmapersonalaccesstoken.go`

shahzadhaider1: done

amanfcp (approved):
💯

pkg/detectors/figmapersonalaccesstoken/figmapersonalaccesstoken.go
```diff
@@ -0,0 +1,37 @@
...
+		return false, err
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		return true, nil
```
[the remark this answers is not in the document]
shahzadhaider1: ok cursor

pkg/detectors/figmapersonalaccesstoken/v3/figmapersonalaccesstoken_integration_test.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			for i := range got {
+				if len(got[i].Raw) == 0 {
+					t.Fatalf("no raw secret present: \n %+v", got[i])
+				}
+				if (got[i].VerificationError() != nil) != tt.wantVerificationErr {
+					t.Fatalf("wantVerificationError = %v, verification error = %v", tt.wantVerificationErr, got[i].VerificationError())
+				}
+			}
+			ignoreOpts := cmpopts.IgnoreFields(detectors.Result{}, "Raw", "verificationError", "primarySecret")
```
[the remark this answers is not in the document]
shahzadhaider1: Tests in v3 are skipped at the moment -- please refer to the message at the start of the test for details.

pkg/detectors/figmapersonalaccesstoken/v3/figmapersonalaccesstoken.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.Versioner = (*Scanner)(nil)
+
+func (Scanner) Version() int { return 3 }
+
+var (
+	defaultClient = common.SaneHttpClient()
+	keyPat        = regexp.MustCompile(`(figp_[a-zA-Z0-9_=-]{40,54})`)
```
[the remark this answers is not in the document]
shahzadhaider1: A too-long `figp_` value does get cut to 54 chars, but that truncated string is just the wrong token, it fails verification (403) and never shows up as a verified result.

A negative lookahead isn't an option since we run on go-re2, which has no lookarounds. And `\b` doesn't work here because the token charset includes `=` and `-`, which aren't word characters. The only remaining route is a manual boundary check in code, but no other detector does that, and it'd be extra logic just to reject malformed input that already can't verify. Keeping it simple for consistency.

pkg/detectors/figmapersonalaccesstoken/figmapersonalaccesstoken.go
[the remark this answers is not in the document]
shahzadhaider1: This is intentional. It's highly unlikely to receive `201` and `204` in our case.

shahzadhaider1:
bugbot run

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalacesstoken_integration_test.go
```diff
@@ -55,6 +55,7 @@ func TestFigmaPersonalAccessToken_FromChunk(t *testing.T) {
 					ExtraData: map[string]string{
 						"version": "1",
 					},
+					SecretParts: map[string]string{"token": secret},
```
MuneebUllahKhan222: Nit: No need for it, since we are ignoring `SecretParts` in `cmp.Diff`

pkg/detectors/figmapersonalaccesstoken/v3/figmapersonalaccesstoken.go
MuneebUllahKhan222: We could update this regex to use explicit token boundaries instead of relying solely on a leading word boundary. For example:

```regex
\b(figp_[A-Za-z0-9_=-]{40,54})(?:[^A-Za-z0-9_=-]|\z)
```

The trailing `(?:[^A-Za-z0-9_=-]|\z)` ensures the match ends only when the next character is not part of the token alphabet (or we've reached the end of the input), which helps avoid partial matches. This is something we are already doing this in `jiradatacenterpat`, `dockerhub`  and other detectors.

pkg/detectors/figmapersonalaccesstoken/figmapersonalaccesstoken.go
```diff
@@ -0,0 +1,37 @@
...
+	case http.StatusForbidden:
```
MuneebUllahKhan222: Figma [docs](https://developers.figma.com/docs/rest-api/discovery-endpoints/) insinuate that they return 401 and 403 both in-case of an invalid token. So I think to be safe we should return `false, nil` for 401.

pkg/engine/defaults/defaults.go
```diff
@@ -1794,6 +1796,8 @@ func buildDetectorList() []detectors.Detector {
 			return !feature.CloudinaryDetectorEnabled.Load()
 		case *gitlaboauth2.Scanner:
 			return !feature.GitLabOAuthDetectorEnabled.Load()
+		case *figmapersonalaccesstokenv3.Scanner:
```
MuneebUllahKhan222, line 1837: Need to add `detector_typepb.DetectorType_FigmaPersonalAccessToken` in `excludedFromDefaultList` of defaults_test.go under the comment
```
	// Feature flag gated detectors
	// These should be removed from this list when we remove the feature flag
```

MuneebUllahKhan222 (changes requested):
Mostly the PR is good. Just need to do small tweaks.

pkg/engine/defaults/defaults.go
shahzadhaider1: Good catch but I don't think we can add it here. Unlike the other feature-gated detectors, the Figma type is shared across v1/v2/v3, and v1 + v2 are registered unconditionally in buildDetectorList() (only v3 is behind the flag). So `FigmaPersonalAccessToken` is always present in the default list. Adding it to `excludedFromDefaultList` would trip the reverse check in the test ("in excludedFromDefaultList but also present in buildDetectorList()") and fail.

Confirmed the test passes as-is.

pkg/detectors/figmapersonalaccesstoken/figmapersonalaccesstoken.go
shahzadhaider1: I didn't get this status code while testing the API, but yeah it's good we add this for safety. Done.

MuneebUllahKhan222 (approved):
Thanks for addressing the comments🚀

pkg/detectors/figmapersonalaccesstoken/figmapersonalaccesstoken.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// VerifyMatch checks if the provided Figma token is valid by making a request to the Figma API.
+func VerifyMatch(ctx context.Context, client *http.Client, token string) (bool, error) {
```
martinlocklear, line 20: I think this function might need better test coverage, or unit coverage for it specifically.

shahzadhaider1: Added.

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -10,32 +10,39 @@ import (
...
 func (Scanner) Version() int { return 1 }
 
 var (
 	defaultClient = common.SaneHttpClient()
-
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"figma"}) + `\b([0-9]{6}-[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12})\b`)
+	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"figma"}) + `\b([0-9]{6}-[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12})\b`)
 )
 
-// Keywords are used for efficiently pre-filtering chunks.
```
martinlocklear, line 32: I don't understand why this comment was removed.  Was the comment misleading?  If not, I think this is super useful information for folks not very familiar with this area of code.

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -10,32 +10,39 @@ import (
...
-// Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
 	return []string{"figma"}
 }
 
-// FromData will find and optionally verify FigmaPersonalAccessToken secrets in a given set of bytes.
```
martinlocklear, line 38: Same comment as here: https://github.com/trufflesecurity/trufflehog/pull/4790/changes#r3638432858

pkg/detectors/figmapersonalaccesstoken/v3/figmapersonalaccesstoken_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+				if (got[i].VerificationError() != nil) != tt.wantVerificationErr {
+					t.Fatalf("wantVerificationError = %v, verification error = %v", tt.wantVerificationErr, got[i].VerificationError())
+				}
+			}
+			ignoreOpts := cmpopts.IgnoreFields(detectors.Result{}, "Raw", "verificationError", "primarySecret", "SecretParts", "chunkOffset", "chunkOffsetSet")
+			if diff := cmp.Diff(got, tt.want, ignoreOpts); diff != "" {
+				t.Errorf("FigmaPersonalAccessToken.FromData() %s diff: (-got +want)\n%s", tt.name, diff)
+			}
+		})
+	}
```
martinlocklear, line 154: These kinds of tabular tests are, to my eye, an unreadable, unmaintainable mess.

Thanks for including tests, but please consider refactoring this to use [testify testsuites](https://pkg.go.dev/github.com/stretchr/testify/suite) (or just breaking them out into separate tests).  Here's [an example](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/log/log_cores_test.go#L32-L52) of their use in this repo (that example could use some refactoring to better leverage setup suite and setup test hooks, but 🤷 )

I understand that go-folks seem to do this kind of thing a lot, but there has to be a limit - I think this crosses that line.

martinlocklear: I had a robot simulate that refactor [here](https://gist.github.com/martinlocklear/38a2e3dde588c949c808aa4a1bbc019e).  That same strategy can likely be applied to the other test files we're adding here.

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
```diff
@@ -10,32 +10,39 @@ import (
...
 var _ detectors.Detector = (*Scanner)(nil)
 var _ detectors.Versioner = (*Scanner)(nil)
 
 func (Scanner) Version() int { return 1 }
 
 var (
 	defaultClient = common.SaneHttpClient()
-
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"figma"}) + `\b([0-9]{6}-[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12})\b`)
+	keyPat        = regexp.MustCompile(detectors.PrefixRegex([]string{"figma"}) + `\b([0-9]{6}-[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12})\b`)
```
martinlocklear, line 29: I would make a note here about why using `\b` isn't safe.  It's in the PR, but future folks might not know to look there for it.

martinlocklear (commented):
Left some non-blocking comments - but please take a look a the test refactor.

Edit - I feel like I should have noticed these things earlier - apologies for the re-re-review cycle here

pkg/detectors/figmapersonalaccesstoken/v3/figmapersonalaccesstoken_integration_test.go
shahzadhaider1: Thanks. I agree the big table here is dense. My hesitation with a full testify-suite rewrite is consistency: table-driven FromData tests are the established pattern across effectively every detector in the repo, so converting only the Figma detectors would make them the outlier and set up a mismatch for anyone copying an existing detector as a template. That's a repo-wide convention call I don't think should be decided in this PR.

If the team genuinely wants to standardize on testify suites for detector tests, I'd suggest raising it separately so it can be applied uniformly rather than piecemeal here. What do you think?

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
shahzadhaider1: Restored

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
shahzadhaider1: Restored

pkg/detectors/figmapersonalaccesstoken/v1/figmapersonalaccesstoken.go
shahzadhaider1: Done

pkg/detectors/figmapersonalaccesstoken/figmapersonalaccesstoken.go
```diff
@@ -0,0 +1,37 @@
...
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		return true, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return false, nil
```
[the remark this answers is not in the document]
shahzadhaider1: In case of 403, we now check the API response body and then decide whether the secret is live or not. More details can be found in the function comment.

martinlocklear (approved):
🤠 + 🤖

PR #4794, opened 2026-03-06
[INS-339] Add New Relic User Key detector

mustansir14:
### Description:
This PR adds the New Relic User Key Detector.

**Regex**: `\b(NRAK-[A-Z0-9]{27})\b`

**Verification**:
For verification, we use the New relic NerdGraph API: `https://api.newrelic.com/graphql`. 
We send a POST request with body `{ requestContext { userId } }`. A response code of `200` means the key is valid. `401` means it is an invalid/rotated key and `403` means the region is incorrect.
Note: For EU region keys, the endpoint should be `https://api.eu.newrelic.com/graphql"`

**Corpora Test**:
image
image

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive secret detector following existing New Relic patterns; verification only calls New Relic APIs when enabled and does not change core scan or auth paths.
> 
> **Overview**
> Adds a **New Relic User API Key** detector for tokens matching `NRAK-[A-Z0-9]{27}`, wired as proto type `NewRelicUserKey` (1066) and registered in default detectors behind `NewRelicUserKeyDetectorEnabled` (enabled at startup in `main.go`).
> 
> When verification is on, the scanner POSTs a minimal NerdGraph `{ requestContext { userId } }` query to US and EU GraphQL endpoints, treating **200** as valid and attaching **region** and **user_id** in extra data; **401** / **403** are treated as invalid or wrong-region without failing the scan.
> 
> Unit and integration tests cover regex matching, live US/EU verification, and inactive keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec760f467a0c4801f87e55c7f7a0924f01b73c68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/newrelicuserkey/newrelicuserkey.go
```diff
@@ -0,0 +1,130 @@
...
+			ctx, http.MethodPost, regionUrl, strings.NewReader(body))
+		if err != nil {
+			return false, nil, fmt.Errorf("error constructing request: %w", err)
+		}
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("API-Key", key)
+
+		res, err := client.Do(req)
+		if err != nil {
+			return false, nil, fmt.Errorf("error making request: %w", err)
```
shahzadhaider1: same suggestion as [here](https://github.com/trufflesecurity/trufflehog/pull/4781)

pkg/detectors/newrelicuserkey/newrelicuserkey.go
```diff
@@ -0,0 +1,141 @@
...
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		var resp graphqlResponse
+		if err := json.NewDecoder(res.Body).Decode(&resp); err != nil {
+			return false, nil, fmt.Errorf("error decoding response for region %s: %w", region, err)
+		}
+		return true, map[string]string{"region": region, "user_id": resp.Data.RequestContext.UserID}, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		// 401 means the key is invalid, 403 means the region is incorrect
```
shahzadhaider1, line 142: 403 means the region is incorrect, but does this mean the credential is valid, or we don't know if its valid/invalid?

mustansir14: As per my testing, 403 does actually also mean that the credential is valid, in which case we may not even need to try the other region.

But there is no official documentation to support this, and I think it's better to be safe here and verify it with the right region, just in case something changes in the future and the API starts to return 403s even for revoked keys that belong to a different region.

pkg/detectors/newrelicuserkey/newrelicuserkey.go
```diff
@@ -0,0 +1,141 @@
...
+		results = append(results, s1)
+	}
+
+	return results, nil
+}
+
+type graphqlResponse struct {
+	Data struct {
+		RequestContext struct {
+			UserID string `json:"userId"`
```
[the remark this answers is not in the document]
mustansir14: The API returns a string, not integer.

Image

shahzadhaider1 (approved):

proto/detector_type.proto
```diff
@@ -1065,4 +1065,5 @@ enum DetectorType {
   HashiCorpVaultBatchToken = 1061;
   HashiCorpVaultToken = 1062;
   Duo = 1063;
+  NewRelicUserKey = 1064;
```
unsmith: Same thing here, conflict with previous NewRelic PRs on the proto enum.

unsmith (changes requested):
Also needs conflict resolution with proto enum numbers.

proto/detector_type.proto
MuneebUllahKhan222: Will resolve them once the PR has been approved and is ready to be merged.

unsmith (approved):

PR #4803, opened 2026-03-10
[INS-335] Added AWS Appsync Detector

MuneebUllahKhan222:
### Description

This PR adds the **AWS AppSync API Key Detector** for TruffleHog.  
It scans for AWS AppSync GraphQL API keys (prefix `da2-` followed by 26 lowercase alphanumeric characters) and associated AppSync GraphQL endpoints, and optionally verifies them via the AppSync GraphQL API.

**API Key Regex**: `\b(da2-[a-z0-9]{26})\b`  
**Endpoint Regex**: `https:\/\/[a-z0-9]{26}\.appsync-api\.[a-z0-9-]+\.amazonaws\.com\/graphql`

### Verification
For verification, we send a POST request to the detected AppSync GraphQL endpoint using the API key in the `x-api-key: <key>` header. The request executes a minimal read-only GraphQL query:

`query { __typename }`

A response code of **200 OK** indicates the API key is valid for the endpoint.  
A response code of **401 Unauthorized** or **403 Forbidden** indicates the key is invalid, expired, or does not have permission to access the API.

This verification request is safe because it only executes a read-only GraphQL query and does not perform any mutations or state-changing operations.

### Corpora Test
image

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Adds a new verified detector that makes outbound HTTP requests to AppSync GraphQL endpoints and expands the detector type enum, which could affect scan results and verification behavior if the patterns or status-code handling are off.
> 
> **Overview**
> Adds a new `AWSAppSync` detector that finds AppSync API keys (`da2-…`) alongside AppSync endpoints, emits combined `RawV2` secrets (`endpoint:key`), and optionally verifies keys by POSTing a minimal GraphQL query to the detected `/graphql` endpoint (treating 200 and a specific 502 body as valid).
> 
> Registers the detector in the default detector list and introduces a new `DetectorType_AWSAppSync` enum value, with unit and integration tests (including verification timeout/unexpected status cases) and a benchmark.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66a7033fe69790e0830a7d4cbdccd6bccb27282d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/aws/appsync/appsync.go
```diff
@@ -0,0 +1,163 @@
...
+		for endpoint := range endpoints {
+
+			result := detectors.Result{
+				DetectorType: detector_typepb.DetectorType_AWSAppSync,
+				Raw:          []byte(key),
+				RawV2:        []byte(fmt.Sprintf("%s:%s", endpoint, key)),
+				SecretParts: map[string]string{
+					"key":      key,
+					"endpoint": endpoint,
+				},
```
mustansir14, line 78: Can we also include the endpoint in `ExtraData`?

pkg/detectors/aws/appsync/appsync.go
```diff
@@ -0,0 +1,163 @@
...
+var (
+	defaultClient = common.SaneHttpClient()
+
+	apiKeyPat = regexp.MustCompile(
+		`\b(da2-[a-z0-9]{26})\b`,
+	)
+
+	endpointPat = regexp.MustCompile(
+		`(https:\/\/[a-z0-9]{26}\.appsync-api\.[a-z0-9-]+\.amazonaws\.com(?:\/graphql)?)`,
+	)
```
mustansir14, line 34: Instead of conditionally matching the `/graphql` path here and then normalizing the endpoints later on, why not just match the endpoint using the regex and append the `/graphql` ourselves?

pkg/detectors/aws/appsync/appsync.go
```diff
@@ -0,0 +1,163 @@
...
+	}()
+
+	switch res.StatusCode {
+
+	case http.StatusOK:
+		return true, nil
+
+	case http.StatusUnauthorized,
+		http.StatusForbidden:
+		return false, nil
```
mustansir14: Are we sure that the API can return both these statuses and both of them mean that the credential is determinately unverified? If yes, can you please add that as a comment? Probably at the top of the verification method, also linking to the documentation. Thanks

mustansir14 (changes requested):
Please check the inline comments for some suggestions/questions. Thanks

pkg/detectors/aws/appsync/appsync.go
MuneebUllahKhan222: I investigated this further and found a couple of interesting behaviors in AppSync:
1. AppSync returns a `502 Bad Gateway` with a specific error message when the endpoint is valid but the schema is not configured.
2. AppSync returns `403 Forbidden` for a short period after an API key is deleted or revoked. After some time, the endpoint becomes unreachable or stops responding consistently

So I have additionally added the `502` handling in the new commit and also the documentation link.

pkg/detectors/aws/appsync/appsync.go
MuneebUllahKhan222: Good point.

I removed the optional /graphql from the regex so that we can also detect base AppSync endpoints that don’t include it in the source text. After extraction, I normalize the endpoint by appending /graphql to ensure it’s always in a consistent format for verification.

mustansir14 (approved):

pkg/detectors/aws/appsync/appsync.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	bodyBytes, err := io.ReadAll(res.Body)
```
shahzadhaider1: `bodyBytes` is only used in case of 502 so reading the body could be avoided in other cases

camgunz (approved):

PR #4809, opened 2026-03-10
Improve logging and context management in the Filesystem source

camgunz:
### Description:
When trying to tidy up logging and context management in the Filesystem source, we realized that we'd often be logging messages with duplicate keys (this is an issue when using Uber's zap logger w/ Stackdriver) and that in some cases we'd log the paths of things we scan, which we'd rather not do. So, beyond unifying the patterns, using common names for everything, making logging levels for different situations consistent (V(2) for skipping, V(5) for tracing, etc), and adding trace logging, this avoids logging paths of things we scan and introduces a new pattern for avoiding duplicate keys: `lctx` (local context). In lieu of pulling the logger out of a context (which we avoid because modifying them modifies them for everyone, which quickly becomes confusing) this creates local contexts that are never passed to other functions/methods. I just made this up, feel free to thumbs down. But this is blocking some stuff, so if we don't like it I'm just gonna pull this PR entirely.

### DIFF
--- a/pkg/sources/filesystem/filesystem.go
+++ b/pkg/sources/filesystem/filesystem.go
@@ -111,7 +111,7 @@ func (s *Source) canFollowSymlinks() bool {
 // Chunks emits chunks of bytes over a channel.
 func (s *Source) Chunks(ctx trContext.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
 	for i, rootPath := range s.paths {
-		log := trContext.Logger().WithValues(
+		logger := ctx.Logger().WithValues(
 			"root_path", rootPath,
 		)
 
@@ -123,38 +123,38 @@ func (s *Source) Chunks(ctx trContext.Context, chunksChan chan *sources.Chunk, _
 
 		cleanRootPath := filepath.Clean(rootPath)
 
-		log = log.WithValues(
+		logger = logger.WithValues(
 			"clean_root_path", cleanRootPath,
 		)
 
 		fileInfo, err := os.Lstat(cleanRootPath)
 		if err != nil {
-			log.Error(err, "unable to get file info")
+			logger.Error(err, "unable to get file info")
 			continue
 		}
 
 		if fileInfo.Mode()&os.ModeSymlink != 0 {
 			// if the root path is a symlink we scan the symlink
-			log.V(5).Info("Root path is a symlink")
+			logger.V(5).Info("Root path is a symlink")
 			initialDepth := 0
 			err = s.scanSymlink(ctx, chunksChan, rootPath, initialDepth, cleanRootPath)
 			s.ClearEncodedResumeInfoFor(rootPath)
 		} else if fileInfo.IsDir() {
-			log.V(5).Info("Root path is a dir")
+			logger.V(5).Info("Root path is a dir")
 			initialDepth := 0
 			err = s.scanDir(ctx, chunksChan, rootPath, initialDepth, cleanRootPath)
 			s.ClearEncodedResumeInfoFor(rootPath)
 		} else {
 			if !fileInfo.Mode().IsRegular() {
-				log.V(2).Info("Root path is a non-regular file; skipping")
+				logger.V(2).Info("Root path is a non-regular file; skipping")
 				continue
 			}
-			log.V(5).Info("Root path is a file")
+			logger.V(5).Info("Root path is a file")
 			err = s.scanFile(ctx, chunksChan, cleanRootPath)
 		}
 
 		if err != nil && !errors.Is(err, io.EOF) {
-			log.Error(err, "error scanning filesystem")
+			logger.Error(err, "error scanning filesystem")
 		}
 	}
 
@@ -168,22 +168,22 @@ func (s *Source) scanSymlink(
 	depth int,
 	path string,
 ) error {
-	log := trContext.Logger().WithValues(
+	logger := ctx.Logger().WithValues(
 		"root_path", rootPath,
 		"start_depth", depth,
 	)
 
-	log.V(5).Info("scanSymlink")
+	logger.V(5).Info("scanSymlink")
 
 	if !s.canFollowSymlinks() {
 		// If the file or directory is a symlink but the followSymlinks is disable ignore the path
-		log.V(2).Info("Path is a symlink, but following symlinks is not allowed; skipping")
+		logger.V(2).Info("Path is a symlink, but following symlinks is not allowed; skipping")
 		return nil
 	}
 
 	depth++
 
-	log = log.WithValues(
+	logger = logger.WithValues(
 		"depth", depth,
 	)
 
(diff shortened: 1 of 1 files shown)


pkg/sources/filesystem/filesystem.go
```diff
@@ -159,58 +166,71 @@ func (s *Source) scanSymlink(
 	depth int,
 	path string,
 ) error {
+	ctx = trContext.WithValues(ctx,
+		"root_path", rootPath,
```
kashifkhan0771, line 172: `root_path` is already added at line#114

pkg/sources/filesystem/filesystem.go
```diff
@@ -111,41 +111,48 @@ func (s *Source) canFollowSymlinks() bool {
...
 				continue
 			}
-			ctx.Logger().V(5).Info("Root path is a file", "path", cleanPath)
+			ctx.Logger().V(5).Info("Root path is a file")
 			err = s.scanFile(ctx, chunksChan, cleanPath)
 		}
 
 		if err != nil && !errors.Is(err, io.EOF) {
-			logger.Error(err, "error scanning filesystem")
+			ctx.Logger().Error(err, "error scanning filesystem")
```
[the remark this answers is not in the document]
kashifkhan0771: Don't copy me cursor :)

pkg/sources/filesystem/filesystem.go
```diff
@@ -111,41 +111,48 @@ func (s *Source) canFollowSymlinks() bool {
 // Chunks emits chunks of bytes over a channel.
 func (s *Source) Chunks(ctx trContext.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
 	for i, rootPath := range s.paths {
-		logger := ctx.Logger().WithValues("path", rootPath)
+		ctx = trContext.WithValues(ctx,
+			"root_path", rootPath,
+		)
```
mcastorina: This appends to the current `ctx` in a loop, which will output multiple `root_path` KV pairs. We need to shadow the variable to avoid that:

```go
ctx := trContext.WithValues(ctx, "root_path", rootPath)
//  ^^
```

camgunz: Oop, good catch! (this is why small changes are the best changes!)

pkg/sources/filesystem/filesystem.go
camgunz: (This freaked me out a little so I wrote some test code to confirm what I'm about to write)

I was worried that this would be similar to forgetting to shadow the ctx variable in a loop, but it doesn't accumulate in the parent context--the only way that can happen is if you have access to the parent scope, which you don't inside a function/method. So, all that said I like the pattern of adding the arguments up top for trace logging, even though they're sometimes redundant. It's less mental work to think about all the ways we could end up here and thus what might be in the context.

pkg/sources/filesystem/filesystem.go
```diff
@@ -159,58 +166,71 @@ func (s *Source) scanSymlink(
...
-		ctx.Logger().V(2).Info("skipping, following symlinks is not allowed", "path", path)
+		ctx.Logger().V(2).Info("Path is a symlink, but following symlinks is not allowed; skipping")
 		return nil
 	}
 
 	depth++
 
+	ctx = trContext.WithValues(ctx,
+		"depth", depth,
+	)
```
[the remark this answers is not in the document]
camgunz: Woof is this true? Let me look around.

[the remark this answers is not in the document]
camgunz: This is not true

shahzadhaider1 (approved):

pkg/sources/filesystem/filesystem.go
rosecodym: This is going to emit the key/value pair twice, though. It's genuinely pretty annoying to see in the logs.

camgunz: OK I was wrong about this; mea culpa. In my defense I only went 2 (3?) layers deep into logr's `zapr` code, not 3 (4?) layers deep to zap's code. Oh and look, [there's an issue](https://github.com/uber-go/zap/issues/622) about precisely this, because they're relying on the JSON decoder to dedupe, and Stackdriver's doesn't 🤦🏻.

But, even if Stackdriver did, it's *bananas* to accumulate these things. They're still there accumulating, even if a decoder overwrites them as it's decoding.

Anyway, alright I'll rejigger. "Wrap zap's `JSONEncoder` with something that dedupes, even slowly" goes on my "would love to, but will never, fix list".

pkg/sources/filesystem/filesystem.go
```diff
@@ -248,6 +280,10 @@ func (s *Source) scanDir(
 	resumptionKey := rootPath
 	resumeAfter := s.GetEncodedResumeInfoFor(resumptionKey)
 
+	ctx = trContext.WithValues(ctx,
+		"resume_after", resumeAfter,
```
rosecodym: Isn't this local path information? I think we consider that sensitive.

mcastorina:
> In lieu of pulling the logger out of a context (which we avoid because modifying them modifies them for everyone, which quickly becomes confusing)

I find this surprising, but maybe I'm confused! What do you mean by this?

camgunz:
> > In lieu of pulling the logger out of a context (which we avoid because modifying them modifies them for everyone, which quickly becomes confusing)

> I find this surprising, but maybe I'm confused! What do you mean by this?

@mcastorina Oh I misremembered the exact opposite thing; I was trying to remember [this Scanning Team Hygiene](https://docs.google.com/document/d/1r_obwmGldLWTYtFPpM6NsOJnqIwgrLmrBz0ZOWW8sbw/edit?tab=t.0#heading=h.rd75r9jy7ogw) thing. IDC if you think it's more straightforward to use loggers (I... probably agree) then I can do that.

mcastorina:
> > > In lieu of pulling the logger out of a context (which we avoid because modifying them modifies them for everyone, which quickly becomes confusing)
> 
> > I find this surprising, but maybe I'm confused! What do you mean by this?
> 
> @mcastorina Oh I misremembered the exact opposite thing; I was trying to remember [this Scanning Team Hygiene](https://docs.google.com/document/d/1r_obwmGldLWTYtFPpM6NsOJnqIwgrLmrBz0ZOWW8sbw/edit?tab=t.0#heading=h.rd75r9jy7ogw) thing. IDC if you think it's more straightforward to use loggers (I... probably agree) then I can do that.

Okay good! I don't have a preference between contexts / loggers, though the checklist item makes sense to me. I just wanted to make sure our context / log packages aren't giving people footguns.

pkg/sources/filesystem/filesystem.go
```diff
@@ -111,41 +111,50 @@ func (s *Source) canFollowSymlinks() bool {
 // Chunks emits chunks of bytes over a channel.
 func (s *Source) Chunks(ctx trContext.Context, chunksChan chan *sources.Chunk, _ ...sources.ChunkingTarget) error {
 	for i, rootPath := range s.paths {
-		logger := ctx.Logger().WithValues("path", rootPath)
+		lctx := trContext.WithValues(ctx,
```
rosecodym: Any reason to not shadow `ctx`? We do it elsewhere and it would reduce diff noise.

pkg/sources/filesystem/filesystem.go
```diff
@@ -381,66 +406,78 @@ func (s *Source) scanFile(ctx trContext.Context, chunksChan chan *sources.Chunk,
 		SourceVerify: s.verify,
 	}
 
-	return handlers.HandleFile(fileCtx, inputFile, chunkSkel, sources.ChanReporter{Ch: chunksChan})
+	return handlers.HandleFile(ctx, inputFile, chunkSkel, sources.ChanReporter{Ch: chunksChan})
```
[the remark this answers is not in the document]
camgunz: This was on purpose, to not log sensitive info

pkg/sources/filesystem/filesystem.go
```diff
@@ -236,9 +252,17 @@ func (s *Source) scanDir(
 	depth int,
 	path string,
 ) error {
+	lctx := trContext.WithValues(ctx,
+		"root_path", rootPath,
+		"depth", depth,
+	)
```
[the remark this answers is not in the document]
camgunz: This was on purpose, to not log sensitive info

pkg/sources/filesystem/filesystem.go
camgunz: I agree w/ this; it's the main reason I'm gonna pull out a logger instead.

camgunz:
> is the cursor stuff real?

Probably my #1 annoyance right now is bugbot comments showing up like 10 minutes after a commit. No it was on purpose.

PR #4819, opened 2026-03-17
[INS-355] Added Hashicorp vault token detector

MuneebUllahKhan222:
###Description
This PR adds the `HashiCorp Vault Token Detector`.
It scans for various types of HashiCorp Vault authentication tokens (including standard `service tokens`, `periodic tokens`, and `legacy tokens`) and associated Vault server endpoints. The detector supports live verification against the custom endpoints.

Token Regex: `\b(hvs\.[A-Za-z0-9_-]{90,120}|s\.[A-Za-z0-9_-]{18,40})\b`

Endpoint Regex: `(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`

Verification
Verification is performed by sending a GET request to the Vault server's `auth/token/lookup-self` endpoint using the detected token in the `X-Vault-Token` header.

A response code of `200 OK` indicates the token is valid. In this case, the detector extracts and returns metadata about the token to assist with remediation, including:

- Policies: The permissions associated with the token.

- Token Type: Whether it is a service or batch token.

- Entity ID: Useful for identifying the identity/owner and revoking the token.

- Attributes: orphan and renewable status.

A response code of `401 Unauthorized or 403 Forbidden` indicates the token is invalid or has been revoked.

This verification is safe as lookup-self is a read-only metadata operation that does not consume secrets or trigger state changes within the Vault cluster.

### Corpora Test
image

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Introduces live verification against customer Vault endpoints using leaked tokens (secrets-adjacent), though lookup-self is read-only; shared refactor touches existing batch-token verification behavior.
> 
> **Overview**
> Adds a **HashiCorp Vault Token** detector for `hvs.*` and legacy `s.*` tokens paired with `*.hashicorp.cloud` URLs, with optional live verification via `GET /v1/auth/token/lookup-self` and remediation metadata (policies, type, entity ID, orphan/renewable).
> 
> Vault token verification is **factored into** `pkg/detectors/hashicorpvault` (`VerifyVaultToken`); the batch-token detector now calls that helper instead of duplicating HTTP logic. The new detector is registered behind `HashiCorpVaultTokenDetectorEnabled` (enabled in OSS `main.go`), with proto type `HashiCorpVaultToken` (1062), defaults wiring, and unit/integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b835138d048a83509c8e4d0993957c9ac05a37af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,182 @@
...
+	// legacy tokens are around 18-40 chars and start with s.
+	vaultTokenPat = regexp.MustCompile(
+		`\b(hvs\.[A-Za-z0-9_-]{90,120}|s\.[A-Za-z0-9_-]{18,40})\b`,
+	)
+
+	vaultUrlPat = regexp.MustCompile(`(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`)
+)
+
+func (s Scanner) Keywords() []string {
+	return []string{"hvs.", "s."}
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Need to think about this as it is definitely important for detecting legacy tokens.

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,181 @@
...
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	// Vault tokens:
+	// newer vault tokens are around 90-120 chars and start with hvs. (HashiCorp Vault Service token)
+	// legacy tokens are around 18-40 chars and start with s.
+	vaultTokenPat = regexp.MustCompile(
+		`\b(hvs\.[A-Za-z0-9_-]{90,120}|s\.[A-Za-z0-9_-]{18,40})\b`,
+	)
+
+	vaultUrlPat = regexp.MustCompile(`(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: This is not an issue.

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,203 @@
...
+	vaultTokenPat = regexp.MustCompile(
+		`\b(hvs\.[A-Za-z0-9_-]{90,120}|s\.[A-Za-z0-9_-]{18,40})(?:$|[^A-Za-z0-9_-])`,
+	)
+
+	vaultUrlPat = regexp.MustCompile(`(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`)
+)
+
+func (s Scanner) Keywords() []string {
+	// We cant use s. as a keyword because it is too broad so that's why we are using "vault".
+	return []string{"hvs.", "vault"}
```
shahzadhaider1, line 40: Did you mean `hvs.` in the comment?

pkg/detectors/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,203 @@
...
+		endpoints = append(endpoints, endpoint)
+	}
+
+	for token := range uniqueTokens {
+		emitted := false
+		for _, endpoint := range s.Endpoints(endpoints...) {
+			result := detectors.Result{
+				DetectorType: detector_typepb.DetectorType_HashiCorpVaultToken,
+				Raw:          []byte(token),
+				RawV2:        []byte(token + endpoint),
```
shahzadhaider1: A separator between token and endpoint could be helpful here

pkg/detectors/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,203 @@
...
+
+			results = append(results, result)
+			emitted = true
+		}
+		if !emitted {
+			// No reachable URL in context — emit an unverified token-only
+			// result and annotate why we couldn't verify.
+			results = append(results, detectors.Result{
+				DetectorType: detector_typepb.DetectorType_HashiCorpVaultToken,
+				Raw:          []byte(token),
```
shahzadhaider1: RawV2 is missing in this case; the same token scanned twice (once with an endpoint nearby, once without) becomes two distinct findings with incompatible identity strategies. What do you think?

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
MuneebUllahKhan222: No.
The legacy tokens starts with `s.` but it is too short to be a keyword so that's why we are using vault.

pkg/detectors/hashicorpvaulttoken/hashicorpvaulttoken.go
MuneebUllahKhan222: I may have taken inspiration from another detector that didn’t use a separator when creating this one. Nevertheless, I’ve now added `:` as a separator.

pkg/detectors/hashicorpvaulttoken/hashicorpvaulttoken.go
MuneebUllahKhan222: I have added RawV2 to the `results`.

Regarding the duplicate detection, I think it’s acceptable to report the same token twice when it appears in different contexts (with and without an endpoint). This ensures we don’t risk missing a valid token regardless of how it is discovered.

MuneebUllahKhan222: I took a look at other detectors that require a combination of `url` and `token` like **jira** and found that those detectors only report a result a combination of both `url` and `token` is present so I have updated this detector to mimic the same behaviour.

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,186 @@
...
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	// Vault tokens:
+	// newer vault tokens are around 90-120 chars and start with hvs. (HashiCorp Vault Service token)
+	// legacy tokens are around 18-40 chars and start with s.
+	vaultTokenPat = regexp.MustCompile(
+		`\b(hvs\.[A-Za-z0-9_-]{90,120}|s\.[A-Za-z0-9_-]{18,40})(?:$|[^A-Za-z0-9_-])`,
+	)
+
+	vaultUrlPat = regexp.MustCompile(`(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`)
```
shahzadhaider1, line 35: will this always end in `hashicorp.cloud` or could there be some enterprise vault urls with custom domains?

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
```diff
@@ -0,0 +1,186 @@
...
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+var (
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	// Vault tokens:
+	// newer vault tokens are around 90-120 chars and start with hvs. (HashiCorp Vault Service token)
+	// legacy tokens are around 18-40 chars and start with s.
+	vaultTokenPat = regexp.MustCompile(
+		`\b(hvs\.[A-Za-z0-9_-]{90,120}|s\.[A-Za-z0-9_-]{18,40})(?:$|[^A-Za-z0-9_-])`,
```
shahzadhaider1, line 32: comment says newer vault tokens are around 90-120 chars, but the regex will capture tokens with 124 chars as well.

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
MuneebUllahKhan222: The cloud urls will always end with `hashicorp.cloud`. The enterprise user can host there `hashicorp vault` instance on their custom domain and for those type of users I have implement the `EndpointCustomizer` interface so they can configure their instance url.

pkg/detectors/hashicorpvault/hashicorpvaulttoken/hashicorpvaulttoken.go
MuneebUllahKhan222: Well that's excluding the prefix `hvs.`. Let me rephrase this comment for better clarity.

shahzadhaider1 (approved):

camgunz (approved):

amanfcp (approved):

pkg/detectors/hashicorpvault/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
```diff
@@ -90,9 +88,10 @@ func (s Scanner) FromData(
 			}
 
 			if verify {
-				verified, verificationResp, verificationErr := verifyVaultToken(
+				verified, verificationResp, verificationErr := hashicorpvault.VerifyVaultToken(
 					ctx,
 					s.getClient(),
+					detector_typepb.DetectorType_HashiCorpVaultBatchToken,
```
shahzadhaider1: just a suggestion: `s.Type()` is also same but short

shahzadhaider1 (approved):

PR #4824, opened 2026-03-18
[INS-410] Added batch token detector

MuneebUllahKhan222:
###Description
This PR adds the `HashiCorp Vault Token Detector`.
It scans for various types of HashiCorp Vault authentication tokens (including standard `service tokens`, `periodic tokens`, and `legacy tokens`) and associated Vault server endpoints. The detector supports live verification against the custom endpoints.

Token Regex: `\b(hvb\.[A-Za-z0-9_.-]{50,300})\b`

Endpoint Regex: `(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`

Verification
Verification is performed by sending a GET request to the Vault server's `auth/token/lookup-self` endpoint using the detected token in the `X-Vault-Token` header.

A response code of `200 OK` indicates the token is valid. In this case, the detector extracts and returns metadata about the token to assist with remediation, including:

- Policies: The permissions associated with the token.

- Entity ID: Useful for identifying the identity/owner and revoking the token.

- Attributes: orphan and renewable status.

A response code of `401 Unauthorized or 403 Forbidden` indicates the token is invalid or has been revoked.

This verification is safe as lookup-self is a read-only metadata operation that does not consume secrets or trigger state changes within the Vault cluster.

### Corpora Test
image

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Introduces live verification that sends detected tokens to Vault endpoints; rollout is feature-flagged but enabled by default in OSS main.
> 
> **Overview**
> Adds a **HashiCorp Vault batch token** detector for `hvb.*` tokens paired with `*.hashicorp.cloud` Vault URLs in the same chunk.
> 
> When verification is on, it calls **`GET /v1/auth/token/lookup-self`** with `X-Vault-Token` and attaches token metadata (policies, entity ID, orphan/renewable/type) on success. Tokens without a co-located URL are not reported.
> 
> Wiring includes proto type **`HashiCorpVaultBatchToken` (1056)**, registration in default detectors, **`HashiCorpVaultBatchTokenDetectorEnabled`** (on in OSS `main.go`), unit/integration tests, and the usual feature-gated exclude-list entry in defaults tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 603525bb0e92515eabde27edd1bb4deb8915d4d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
```diff
@@ -0,0 +1,181 @@
...
+
+var (
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	// Batch tokens: hvb.<50-300 chars>
+	batchTokenPat = regexp.MustCompile(
+		`\b(hvb\.[A-Za-z0-9_.-]{50,300})\b`,
+	)
+
+	vaultUrlPat = regexp.MustCompile(`(https?:\/\/[^\s\/]*\.hashicorp\.cloud(?::\d+)?)(?:\/[^\s]*)?`)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: The duplication is not a issues as bot detectors have different regex for keys but same format of url.

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
```diff
@@ -0,0 +1,181 @@
...
+		ctx,
+		http.MethodGet,
+		url,
+		http.NoBody,
+	)
+	if err != nil {
+		return false, nil, err
+	}
+
+	req.Header.Set("X-Vault-Token", token)
```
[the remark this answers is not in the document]
MuneebUllahKhan222: According to the [docs](https://developer.hashicorp.com/vault/api-docs/auth/token#lookup-a-token-self) the `lookup-self` endpoint that we are using doesn't require `X-Vault-Namespace` header and I have also verified this behaviour using postman.

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
```diff
@@ -0,0 +1,185 @@
...
+
+var _ detectors.Detector = (*Scanner)(nil)
+var _ detectors.EndpointCustomizer = (*Scanner)(nil)
+
+var (
+	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
+
+	// Batch tokens: hvb.<50-300 chars>
+	batchTokenPat = regexp.MustCompile(
+		`\b(hvb\.[A-Za-z0-9_.-]{50,300})\b`,
```
kashifkhan0771: We're using a word boundary at the end, but our character set also includes `_`, `.`, and `-`. If the token can legitimately end with any of those characters, the word boundary won't treat them as part of the word, which can lead to incorrect matching.

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
```diff
@@ -0,0 +1,185 @@
...
+					s.getClient(),
+					endpoint,
+					token,
+				)
+				result.SetVerificationError(verificationErr, token)
+				result.Verified = verified
+
+				if verificationResp != nil {
+					result.ExtraData = map[string]string{
+						"policies":  strings.Join(verificationResp.Data.Policies, ", "),
```
kashifkhan0771: If the policy represents an ACL, shouldn't it be part of `analyze` instead? Including what a token is allowed to do in `ExtraData` isn't a pattern we typically follow.

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
```diff
@@ -0,0 +1,185 @@
...
+		url,
+		http.NoBody,
+	)
+	if err != nil {
+		return false, nil, err
+	}
+
+	req.Header.Set("X-Vault-Token", token)
+
+	res, err := client.Do(req)
```
kashifkhan0771: Non-blocking comment: As now we have a `DoWithDedup` we can use it in new detectors.
Example:
```go
resp, err := detectors.DoWithDedup(client, detector_typepb.DetectorType_Meraki, match, req)
if err != nil {
  return nil, false, err
}
```

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/engine/ahocorasick"
+)
+
+func TestBatchToken_PatternWithURL(t *testing.T) {
```
kashifkhan0771, line 14: Pattern test cases are one of the most critical parts of any detector, so I'd suggest taking the time to write them thoroughly.

We've already spent a lot of time cleaning up and improving the test cases across existing detectors, and I'd like to avoid creating another backlog of fixes.

For reference, have a look at:
- https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/browserstack/browserstack_test.go
- Example PR where we improved the test cases: https://github.com/trufflesecurity/trufflehog/pull/3564

kashifkhan0771 (changes requested):

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
MuneebUllahKhan222: Good catch, weird that bugbot didn't point this out.

pkg/detectors/hashicorpvaultbatchtoken/hashicorpvaultbatchtoken.go
MuneebUllahKhan222: I have moved it inside `SecretParts`.

kashifkhan0771 (approved):
LGTM! You need to resolve conflicts.

camgunz (approved):

amanfcp (approved):

PR #4831, opened 2026-03-24
fix: improve detector unit test data for kraken, netsuite, viewneo

### LINKED ISSUE #3817: Fix broken & inaccurate detector unit tests
The detector unit tests were created by "reverse engineering" the patterns, thus they do not provide any real value or confirm the accuracy/currency of the detectors.

Additionally, the structure of the tests are (in my opinion) not maintainable as they are difficult to understand and do not reflect any realistic scenarios. 

https://github.com/trufflesecurity/trufflehog/blob/def734a783b2d4542f2b353d326130d618f3c421/pkg/detectors/netsuite/netsuite_test.go#L14-L44

## Examples

There are dozens, if not hundreds, of problematic test files. These two are illustrative.

### BombBomb

The "valid" tests for BombBomb do not match the detector's pattern.

https://github.com/trufflesecurity/trufflehog/blob/def734a783b2d4542f2b353d326130d618f3c421/pkg/detectors/bombbomb/bombbomb.go#L24

https://github.com/trufflesecurity/trufflehog/blob/def734a783b2d4542f2b353d326130d618f3c421/pkg/detectors/bombbomb/bombbomb_test.go#L15

### Kraken

The "valid" pattern is nonsensical and [not correct base64 encoding](https://base64.guru/learn/base64-characters). The detector should not match this, that is a defect.

https://github.com/trufflesecurity/trufflehog/blob/def734a783b2d4542f2b353d326130d618f3c421/pkg/detectors/kraken/kraken_test.go#L16

image
https://support.kraken.com/hc/en-us/articles/360000919966-How-to-create-an-API-key

### viewneo

A few hundred detectors contain tests tightly coupled to the current implementation of [`PrefixRegex`](https://github.com/trufflesecurity/trufflehog/blob/bcd89e63f88be1126959c20210f134a81b3d252a/pkg/detectors/detectors.go#L229-L233
). Any changes to the prefix pattern will break the detector tests, which seems inadvisable.

https://github.com/trufflesecurity/trufflehog/blob/def734a783b2d4542f2b353d326130d618f3c421/pkg/detectors/viewneo/viewneo_test.go#L38-L42

raajheshkannaa:
Closes #3817 (partial)

## Summary
Addresses three detectors explicitly cited in the issue where test data was reverse-engineered from regex patterns rather than matching real credential formats:

- **kraken**: Replaced invalid base64 test strings (with `=` padding mid-string) with properly encoded keys matching real Kraken API format
- **netsuite**: Used realistic sandbox account ID format, added documentation comments explaining TBA credential structure
- **viewneo**: Replaced opaque blob with structured JWT-like token (header.payload.signature), fixed invalid pattern placement

## Test plan
- [x] `go test ./pkg/detectors/kraken/...` passes
- [x] `go test ./pkg/detectors/netsuite/...` passes
- [x] `go test ./pkg/detectors/viewneo/...` passes
- [x] Test data now resembles real credential formats rather than regex artifacts

---

> [!NOTE]
> **Low Risk**
> Test-only updates to fixture data and assertions; no changes to detector regex or verification logic.
> 
> **Overview**
> **Refreshes pattern-matching unit tests** for the Kraken, NetSuite, and Viewneo detectors so fixtures look like real secrets instead of regex-shaped blobs. **Detector behavior is unchanged**—only `*_test.go` files are touched.
> 
> **Kraken** swaps old strings (including invalid mid-string `=` padding) for plausible base64 API key/secret lengths, documents why invalid cases use `!`, and drives the happy path through `api_key` / `api_secret` config text instead of quoted keyword pairs. Expected matches are spelled out explicitly (no `strings.TrimSpace`).
> 
> **NetSuite** adds comments for OAuth 1.0 TBA shape, uses a sandbox-style account ID (`TSTDRV_1234`), keeps invalid credentials as `?` outside the allowed charset, and moves expected pair concatenations into the test function.
> 
> **Viewneo** replaces one huge opaque token with a three-part JWT-like string aligned to segment length rules, puts the valid case in a `VIEWNEO_API_TOKEN` env-style line, and fixes the negative test so `?` breaks the first segment while still using a `token =` layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f355571981f1a70b3d03309eb80221034bb17bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/kraken/kraken_test.go
+++ b/pkg/detectors/kraken/kraken_test.go
@@ -3,7 +3,6 @@ package kraken
 import (
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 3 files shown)


raajheshkannaa:
Squashed the fixup commits into one. The bot's finding is resolved: `keyword` is used in all four `fmt.Sprintf` calls in `viewneo_test.go`, matching how `kraken_test.go` and `netsuite_test.go` do it.

```
ok  github.com/trufflesecurity/trufflehog/v3/pkg/detectors/viewneo
ok  github.com/trufflesecurity/trufflehog/v3/pkg/detectors/kraken
ok  github.com/trufflesecurity/trufflehog/v3/pkg/detectors/netsuite
```

`go vet` is clean. Ready for review.

PR #4836, opened 2026-03-25
[INS-407] Fixed AWS detector producing non deterministic output

MuneebUllahKhan222:
### Description:
When multiple AWS credentials are detected in a single file and verification is disabled, the output of the AWS detector becomes non-deterministic due to unordered processing.

Can be reproduced by creating a file with multiple AWS credentials and running this command
``` trufflehog --no-update --no-verification filesystem example-creds-file --json | jq -r .RawV2 ```

To address this, this PR introduces logic to generate and return all possible combinations of Access Key IDs and Secret Access Keys. This ensures that no potentially valid credential pair is missed during reporting.

When verification is enabled, the existing behaviour remains unchanged.

- verification OFF → all combinations
- verification ON → unchanged behaviour

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes secret-scanning output volume and the shared `CustomResultsCleaner` API; unverified multi-credential files may report more (including incorrect cross-pairs), while verified behavior stays the same.
> 
> **Overview**
> Fixes **non-deterministic AWS detector output** when a chunk has multiple access keys and secrets and scans run with **`--no-verification`**. AWS cleaners used to collapse results to one entry per access key ID even without verification, so which pair survived depended on map iteration order.
> 
> **`CleanResults` now takes `verificationEnabled`**, and the engine passes **`e.verify`** from `filterResults`. **`aws.CleanResults` is a no-op when verification is disabled**, so every ID×secret (and session-key) combination from detection is kept. With verification on, behavior is unchanged: at most one result per ID, preferring verified matches.
> 
> Access-key and session-key detectors delegate to the shared AWS helper; **`CustomResultsCleaner`** and default **`detectors.CleanResults`** signatures are updated for compatibility. Tests cover multi-credential inputs and engine filtering when a custom cleaner runs irrespective of config but verification is off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 801c1b8cd283543b9cb7834c5d68c1e64dd8ef7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


camgunz (commented):
I'd prefer it if disabling verification **only** disabled verification and didn't opt the user into lots of other non-verification-related code paths. So like, ideally we'd just sort the result list before returning it, even though it's more computationally intensive. So, mostly I just wonder if we can do that instead.

MuneebUllahKhan222:
> I'd prefer it if disabling verification **only** disabled verification and didn't opt the user into lots of other non-verification-related code paths. So like, ideally we'd just sort the result list before returning it, even though it's more computationally intensive. So, mostly I just wonder if we can do that instead.

We can definitely do that but at the expense of not reporting a legit combination of secret that should/could have been reported. If you are ok with that trade-off I can change the logic.

camgunz (approved):

mustansir14 (approved):

PR #4838, opened 2026-03-25
fix: replace release-guard workflow with revert-latest job

sysread:
## Summary

The `release-guard.yml` workflow was broken from day one: the `unset-latest` job had no `actions/checkout` and no `GH_REPO`, so `gh` couldn't resolve the repo ([original failure](https://github.com/trufflesecurity/trufflehog/actions/runs/22491672284)).

Rather than patch the guard (#4776), this removes it entirely and replaces the inline "Mark release as latest" step with a separate `mark-latest` job gated on `needs: Release`. The release is only promoted to `latest` after the entire Release job succeeds, including post-steps.

Changes:
- delete `release-guard.yml`
- replace the inline `Mark release as latest` step and `revert-latest` job with a `mark-latest` job that runs after `Release` succeeds
- `mark-latest` sets `GH_REPO` explicitly (no checkout needed)
- add comments to `release.yml` documenting the state progression through GoReleaser's non-atomic pipeline and what to check on failure at each phase
- add comment to `.goreleaser.yml` explaining why `make_latest: false` is set

Supersedes #4776.

If `mark-latest` fails, the release and all artifacts exist but `/releases/latest` still points to the previous release. The workflow comments include the manual promotion command.

## Test plan

- verify workflow YAML is syntactically valid
- confirm `mark-latest` job has `GH_REPO` set (the original bug)
- review state-progression comments for accuracy against GoReleaser pipeline

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ jobs:
     env:
       DOCKER_CLI_EXPERIMENTAL: "enabled"
     steps:
+      # Setup steps - no external side effects.
       - name: Checkout
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -54,6 +54,32 @@ jobs:
...
+    if: failure()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Restore previous release as latest
+        run: |
+          LATEST_TAG=$(gh release list --json tagName,isLatest -q '.[] | select(.isLatest) | .tagName')
+          if [ "$LATEST_TAG" != "$RELEASE_TAG" ]; then
+            echo "Tag $RELEASE_TAG is not marked as latest (latest is $LATEST_TAG), skipping."
+            exit 0
+          fi
```
j2fw: We edit the release to be latest as the last step in the `Release` job.

I think this means that early exit here makes this job only continue if that last step succeeded in setting the release with the pushed tag as latest, but one of the gh actions post steps failed at the end of the job. Even if those happened, the resulting artifacts exist and are valid, right?

If that's true, I wonder if this `revert-latest` job is worth maintaining.

j2fw: Maybe it would be cleaner to mark the release as latest in a separate gh action job in this same workflow, with a `needs` condition.

Then, it would only be eligible to run once `Release` has completed all steps and post-steps successfully. I expect that this would also handle cancellation of the job, which so far hasn't been handled with the `if: failure()` condition.

eg:

```yaml
  mark-latest:
    needs: Release
    runs-on: ubuntu-latest
    steps:
      - run: gh release edit "$TAG" --latest
        env:
          GH_TOKEN: ${{ github.token }}
          GH_REPO: ${{ github.repository }}
          TAG: ${{ github.ref_name }}
```

sysread: You're right. TBH, this whole release workflow makes me a little itchy. GoReleaser has already promoted the package, image, and brew tap. I wonder _why_ we have `make_release: false` in `.goreleaser.yml`... ah, it's because goreleaser tags the release as `latest` _before_ any artifacts get uploaded. Lovely.

I guess we can do a post-step after cleanup, but I definitely want to add more helpful error messaging around this. We have a lot of ways to end up in an intermediate release state here.

PR #4840, opened 2026-03-25
Add HTML decoder for secret detection in HTML-formatted sources

alafiand:
## Summary

- Adds a new `HTML` decoder to the decoder pipeline that extracts clean text from HTML content, enabling secret detection in sources like MS Teams and Confluence that emit HTML rather than plain text.
- Parses HTML to extract text nodes, high-signal attribute values (`href`, `src`, `value`, `xlink:href`, etc.), script/style/comment content, and code/pre blocks with proper token boundary preservation.
- Handles syntax-highlight boundary detection (hljs-* classes), zero-width/invisible character stripping, URL decoding in attributes, and double-encoded HTML entity cleanup.
- Adds `HTML = 5` to the `DecoderType` protobuf enum and registers the decoder in `DefaultDecoders()`.
- feature is gated behind `htmlDecoder` feature flag in ConfigCat.

## Test plan

- [x] Unit tests covering: split secrets across tags, attribute extraction, URL decoding, script/style/comment content, code blocks, syntax-highlight boundaries, zero-width character stripping, double-encoded entities, feature flag gating
- [X] Integration tested via thog dev deploy with live MS Teams and Confluence scanning ([companion thog PR](https://github.com/trufflesecurity/thog/pull/5761))

---

> [!NOTE]
> **Medium Risk**
> Adds a new decoder to the default decoding pipeline and extends the `DecoderType` protobuf enum, which can affect scanning output/telemetry and downstream consumers despite being gated behind a runtime feature flag.
> 
> **Overview**
> Adds a new, feature-flagged `HTML` decoder to the default decoder chain to normalize HTML-formatted source content into scan-friendly text.
> 
> The decoder parses HTML and emits visible text plus high-signal attribute values and script/style/comment content, with heuristics for token boundaries (block elements, `hljs-*` spans), URL-unescaping, double-encoded entity cleanup, and stripping invisible Unicode characters. This introduces `DecoderType_HTML` in the protobuf enum and a new `feature.HTMLDecoderEnabled` toggle, with extensive unit tests covering real-world Teams/Confluence-style inputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0d437a36102ec61ce4f684dc4fdd03f7aa74cc4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/decoders/html.go
+++ b/pkg/decoders/html.go
@@ -1,0 +1,251 @@
+package decoders
+
+import (
+	"bytes"
+	"net/url"
+	"regexp"
+	"strings"
+
+	"golang.org/x/net/html"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detectorspb"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+// HTML is a decoder that extracts textual content from HTML documents.
+// It produces a normalized view containing visible text, attribute values,
+// script/style content, and HTML comments with entities and URL-encoding decoded.
+type HTML struct {
+	// Enabled controls whether the decoder is active. When nil, the decoder
+	// is always active. Inject a function that checks a feature flag to
+	// allow dynamic toggling without restarting the scanner.
+	Enabled func() bool
+}
+
+func (d *HTML) Type() detectorspb.DecoderType {
+	return detectorspb.DecoderType_HTML
+}
+
+var htmlTagPattern = regexp.MustCompile(`<[a-zA-Z][a-zA-Z0-9]*[\s>/]`)
+
+// highSignalAttrs are attribute names whose values are extracted into the
+// decoded output because they commonly contain URLs, tokens, or other secrets.
+var highSignalAttrs = map[string]bool{
+	"href":       true,
+	"src":        true,
+	"action":     true,
+	"value":      true,
+	"content":    true,
+	"alt":        true,
+	"title":      true,
+	"xlink:href": true,
+}
+
+// syntaxHighlightPrefixes lists CSS class prefixes used by syntax highlighting
+// libraries. Elements with these classes mark logical line boundaries in code
+// blocks where the platform (e.g. Teams) strips actual newlines.
+var syntaxHighlightPrefixes = []string{"hljs-"}
+
+// residualEntityReplacer decodes common HTML entities that survive double-encoding.
+// When content is entity-encoded twice (e.g. &amp;amp;), the parser's first pass
+// leaves residual entity sequences that this replacer cleans up.
+var residualEntityReplacer = strings.NewReplacer(
+	"&amp;", "&",
+	"&lt;", "<",
+	"&gt;", ">",
+	"&quot;", `"`,
+	"&#39;", "'",
+	"&apos;", "'",
+)
+
+// invisibleReplacer strips zero-width and invisible Unicode codepoints that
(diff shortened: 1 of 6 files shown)


pkg/decoders/html.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+// HTML is a decoder that extracts textual content from HTML documents.
+// It produces a normalized view containing visible text, attribute values,
+// script/style content, and HTML comments with entities and URL-encoding decoded.
+type HTML struct {
+	// Enabled controls whether the decoder is active. When nil, the decoder
+	// is always active. Inject a function that checks a feature flag to
+	// allow dynamic toggling without restarting the scanner.
+	Enabled func() bool
```
casey-tran: Is `Enabled` really needed if nothing by default uses this HTML package?

pkg/decoders/html.go
```diff
@@ -0,0 +1,250 @@
...
+	"value":      true,
+	"content":    true,
+	"alt":   true,
+	"title": true,
+}
+
+// syntaxHighlightPrefixes lists CSS class prefixes used by syntax highlighting
+// libraries. Elements with these classes mark logical line boundaries in code
+// blocks where the platform (e.g. Teams) strips actual newlines.
+var syntaxHighlightPrefixes = []string{"hljs-"}
```
casey-tran, line 43: Do you think it would be helpful to note that `hljs-` is a MS Teams specific use case? In case people need to append more to the slice over time as sources change.

casey-tran (commented):
Looks pretty good to me! Left a few comments.

pkg/decoders/html.go
alafiand: I think the reason to include `Enabled` is so we have a kill switch for EE (my understanding is that rollout will be gradual), but OSS will be able to use it out of the box.

pkg/decoders/html.go
alafiand: I agree, I edited the comment to call out the Teams use case and guide future non-Teams additions.

pkg/decoders/html.go
casey-tran: Yes but for EE blocking, you should just need to deal with the feature flag in [thog](https://github.com/trufflesecurity/thog/blob/e870d5b9318aad21cfeaf4fefa1e1f11dd7ac191/scanner/pkg/config/feature/feature.go#L22). If the intention for OSS is to not have a flag at all.

alafiand: Ah okay I hear you. I think the tradeoff is that checking the flag in `pipeline.go` only evaluates at startup, so after toggling in configcat, we would require a scanner restart for it to take effect (which may be expected behavior from customers). That would be fine for the initial release, but the `Enabled` callback would let us toggle per-customer via ConfigCat at runtime without restarts. I imagine something similar to this has been debated before, so it is very possible my position is known to be no good.

alafiand: Confirming our offline chat, I agree with your suggestion to move this to thog.

casey-tran: The ConfigCat flags should be checked every 10 seconds [here](https://github.com/trufflesecurity/thog/blob/e870d5b9318aad21cfeaf4fefa1e1f11dd7ac191/scanner/pkg/config/feature/feature.go#L125).

casey-tran (approved):

PR #4842, opened 2026-03-26
handle AADSTS50173 as explicit revocation signal for azure refresh tokens

jordanTunstill:
AADSTS50173 error messages were falling through to the default error branch, surfacing
revoked tokens as verification errors (unknown) rather than clean
unverified results. 
Added ErrTokenRevoked sentinel and handled it
deterministically in processMatches.
The error message is: "unexpected error ‘invalid_grant': AADSTS50173: The provided grant has expired due to it being revoked, a fresh auth token is needed"
Additionally, ErrTokenExpired previously used continue TokenLoop, silently skipping expired tokens and emitting no result. Expired tokens now follow the same path as revoked tokens, breaking out of the client loop and emitting a clean verified: false result.

### Description:

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Changes only affect Azure Entra refresh token verification error classification, reducing false erroring but potentially altering how some tokens are reported as unverified vs errored.
> 
> **Overview**
> Improves Azure Entra refresh token verification by recognizing `AADSTS50173` as an explicit revocation signal (`ErrTokenRevoked`) instead of falling through to an "unexpected" verification error.
> 
> Adjusts `processMatches` error handling so expired/revoked tokens stop verification attempts and return a clean unverified result (with confidence-adjusted client/tenant fields) rather than short-circuiting the whole token scan or surfacing an unknown verification error.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 91f58a9de3fbade06d2d3885e2f545db1acd853b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/detectors/azure_entra/refreshtoken/refreshtoken.go
```diff
@@ -134,6 +134,10 @@ TokenLoop:
 							continue
 						} else if errors.Is(verificationErr, ErrTokenExpired) {
 							continue TokenLoop
+						} else if errors.Is(verificationErr, ErrTokenRevoked) {
+							// Token was explicitly revoked; emit a clean unverified result.
+							r = createResult(token, clientId, tenantId, false, nil, nil)
+							break ClientLoop
```
shahzadhaider1: This change here is correct, but I see asymmetry with `ErrTokenExpired`:
ErrTokenExpired: continue `TokenLoop` -> no result emitted
ErrTokenRevoked: break `ClientLoop` -> unverified result emitted
Is this intentional?

jordanTunstill: @shahzadhaider1 I was operating under the idea that the rest of this detector was working as intended, so I wanted to implement my change without affecting the rest of the functionality. LMK your thoughts, happy to revisit it if the other aspects of the detector should also be changed.

shahzadhaider1: Yeah, totally understand wanting to keep the scope tight. I was just flagging it since we're already updating this detector and the asymmetry stood out during review.

The concern is about multi-scan behavior: if a token was previously reported as verified and then expires before the next scan, `ErrTokenExpired` -> continue TokenLoop means no result is emitted. Downstream systems receive silence, and the token stays open with stale verified state indefinitely.

This PR actually handles revocation correctly from that perspective; re-emitting as unverified lets consumers update their state, which makes the expired case look worse by comparison. Would it be worth giving `ErrTokenExpired` the same treatment?

Happy to discuss further or address it in a separate PR if you'd prefer to keep this one focused.

jordanTunstill: @shahzadhaider1 sounds good, while I wait for a product-eng review, I'll look into adding this.

jordanTunstill: @shahzadhaider1 just added that functionality :)

pkg/detectors/azure_entra/refreshtoken/refreshtoken.go
```diff
@@ -133,7 +133,13 @@ TokenLoop:
 							invalidClients[clientId] = struct{}{}
 							continue
 						} else if errors.Is(verificationErr, ErrTokenExpired) {
-							continue TokenLoop
+							// Token has expired; emit a clean unverified result
+							r = createResult(token, clientId, tenantId, false, nil, nil)
+							break ClientLoop
```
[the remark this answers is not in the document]
jordanTunstill: taken into account in newest commit

pkg/detectors/azure_entra/refreshtoken/refreshtoken.go
```diff
@@ -128,17 +127,22 @@ TokenLoop:
...
+						// Tenant is valid but the ClientID doesn't exist.
+						invalidClients[clientId] = struct{}{}
+						continue
+					} else if errors.Is(verificationErr, ErrTokenExpired) {
+						// Token has expired; break to the r==nil block which applies
+						// confidence-adjusted clientId/tenantId before creating the result.
+						break ClientLoop
+					} else if errors.Is(verificationErr, ErrTokenRevoked) {
+						// Token was explicitly revoked; same as above.
+						break ClientLoop
```
shahzadhaider1, line 140: This is optional: the two error cases could share a branch, but keeping them separate is also fine for documentation clarity.

shahzadhaider1 (approved):
Good work. 
The PR description focuses on AADSTS50173/revoked tokens. Let's update the description with the second change we made.

pkg/detectors/azure_entra/refreshtoken/refreshtoken.go
shahzadhaider1: Great, thanks!

nabeelalam: @shahzadhaider1 the issue you pointed out is reflected in the [Azure Entra SP v2 detector as well](https://github.com/trufflesecurity/trufflehog/blob/6542ab1c5bb68d915742e4ae44c2f7c3307852d3/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go#L118-L119), and this may potentially be causing expired secrets to be dropped without emitting a result.

shahzadhaider1: Yeah, we should fix that as well (in another PR).

pkg/detectors/azure_entra/refreshtoken/refreshtoken.go
jordanTunstill: Yeah, I didn't want to collapse them into one branch, in case we ever did want to treat them differently.

jordanTunstill: happy to change it to share a branch if that would be cleaner though, LMK

martinlocklear (approved):

PR #4845, opened 2026-03-27
[CSM-1857] Fix expired Azure secrets being silently dropped

dipto-truffle:
## Summary

- **Bug:** The Azure Entra service principal v2 detector dropped findings entirely when Azure returned AADSTS7000222 (secret expired). The `continue SecretLoop` in `ProcessData` skipped result creation, causing expired secrets to vanish from output and stop receiving `last_seen` updates.
- **Fix:** Changed the `ErrSecretExpired` handler to emit an unverified result with the expiry error preserved (`r = createResult(...); break ClientLoop`), consistent with how `ErrSecretInvalid` and `ErrConditionalAccessPolicy` are handled.
- **Tests:** Added unit tests for `ProcessData` verification error handling covering expired, invalid, conditional access, and verified scenarios using a mock Azure HTTP transport.

Corresponds to trufflesecurity/thog#5935.

## Test plan

- [x] `go test ./pkg/detectors/azure_entra/serviceprincipal/v2/ -run TestProcessData -v` passes all 5 subtests
- [x] `TestProcessData_VerificationErrors/expired_secret_produces_unverified_result` confirms that an expired secret now produces 1 unverified result (was 0 before the fix)
- [x] `TestProcessData_ExpiredSecretShouldEmitResult` confirms the result carries `ErrSecretExpired` as the verification error and has `RawV2` set
- [ ] Verify that other verification error subtests (invalid, conditional access, verified) still behave correctly
- [ ] `go test ./pkg/detectors/azure_entra/...` all pass

---

> [!NOTE]
> **Medium Risk**
> Changes Azure Entra SP v2 verification control-flow to return findings for expired secrets, which can alter scan output/`last_seen` behavior but is narrowly scoped and covered by new unit tests.
> 
> **Overview**
> Fixes a bug in the Azure Entra service principal v2 detector where `ErrSecretExpired` (AADSTS7000222) caused the secret to be skipped entirely; expired secrets now emit an **unverified** result instead of being silently dropped.
> 
> Adds focused unit tests for `ProcessData` using a mocked Azure HTTP client to cover expired/invalid/conditional-access/verified token responses and assert `RawV2` population when tenant/client IDs are known.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c101c7e2d87787ec42440f31732f778aea9e14c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
+++ b/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
@@ -74,7 +74,6 @@ func ProcessData(ctx context.Context, clientSecrets, clientIds, tenantIds map[st
 	logCtx := logContext.AddLogger(ctx)
 	invalidClientsForTenant := make(map[string]map[string]struct{})
 
-SecretLoop:
 	for clientSecret := range clientSecrets {
 		var (
 			r        *detectors.Result
@@ -116,7 +115,8 @@ SecretLoop:
 						case errors.Is(verificationErr, serviceprincipal.ErrSecretInvalid):
 							continue ClientLoop
 						case errors.Is(verificationErr, serviceprincipal.ErrSecretExpired):
-							continue SecretLoop
+							r = createResult(tenantId, clientId, clientSecret, false, nil, verificationErr)
+							break ClientLoop
 						case errors.Is(verificationErr, serviceprincipal.ErrTenantNotFound):
 							// Tenant doesn't exist. This shouldn't happen with the check above.
 							delete(tenantIds, tenantId)
--- a/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
+++ b/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
@@ -115,7 +115,7 @@ func ProcessData(ctx context.Context, clientSecrets, clientIds, tenantIds map[st
 						case errors.Is(verificationErr, serviceprincipal.ErrSecretInvalid):
 							continue ClientLoop
 						case errors.Is(verificationErr, serviceprincipal.ErrSecretExpired):
-							r = createResult(tenantId, clientId, clientSecret, false, nil, verificationErr)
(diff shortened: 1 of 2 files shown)


pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
```diff
@@ -116,7 +115,8 @@ SecretLoop:
 						case errors.Is(verificationErr, serviceprincipal.ErrSecretInvalid):
 							continue ClientLoop
 						case errors.Is(verificationErr, serviceprincipal.ErrSecretExpired):
```
nabeelalam, line 117: Verification errors indicate that trufflehog couldn't determinately validate/invalidate the credentials. According to [the documentation for this error code](https://login.microsoftonline.com/error?code=7000222) it implies the secret, so we should set the error to `nil`.

nabeelalam:
@dipto-truffle thanks for putting this up! This is at least one case where this detector will drop a potentially valid but rotated secret and not produce any result for it.

I'm currently looking into whether a deleted tenant can also potentially cause a previously detected live secret to not be marked rotated.

pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
```diff
@@ -116,7 +115,8 @@ SecretLoop:
...
-							continue SecretLoop
+							r = createResult(tenantId, clientId, clientSecret, false, nil, verificationErr)
+							break ClientLoop
```
shahzadhaider1, line 119: Yeah, I agree with Nabeel here; we should not pass the error as an argument as the secret is definitively invalid (expired), not in an uncertain state.

shahzadhaider1 (approved):
Thank you for taking care of this

casey-tran (approved):

PR #4846, opened 2026-03-27
[CSM-1857] Stabilize Azure detector hash_v2 with deterministic iteration

dipto-truffle:
## Summary

- **Root cause:** `ProcessData` in the Azure Entra v2 detector iterated Go maps (`clientSecrets`, `clientIds`, `tenantIds`) non-deterministically, causing the same credential to produce different `(clientId, tenantId)` pairings across scanner runs. This yielded different `RawV2` values, different `hash_v2` hashes, and duplicate secret rows in the database (the secondary issue in CSM-1857).
- **Fix:** Iterate map keys in sorted order via `slices.Sorted(maps.Keys(...))` and clone caller maps with `maps.Clone` before verification-driven mutations. Same inputs now always produce the same `RawV2`/`hash_v2`.
- **Tests:** Added unit tests for RawV2 stability (ID count sensitivity, determinism over 50 runs, caller map immutability, hash divergence chain).

Corresponds to trufflesecurity/thog#5936.

## Test plan

- [x] `go test ./pkg/detectors/azure_entra/serviceprincipal/v2/ -v` — all 8 tests pass
- [x] `TestProcessData_DeterministicRawV2` confirms identical RawV2 across 50 repeated calls with the same inputs
- [x] `TestProcessData_DoesNotMutateCallerMaps` confirms caller maps are not modified
- [x] `TestProcessData_RawV2DependsOnIDCount` confirms RawV2 is populated only with unambiguous IDs
- [x] `TestProcessData_SameSecretDifferentRawV2` documents the RawV2 divergence when chunk context differs
- [x] `go test ./pkg/detectors/azure_entra/...` — all pass

---

> [!NOTE]
> **Medium Risk**
> Changes the credential pairing/selection logic used to compute `RawV2`, which can affect deduplication and reported results for Azure secrets (and is reused by v1). Scope is limited and covered by new determinism and immutability tests.
> 
> **Overview**
> **Stabilizes Azure Entra service principal `hash_v2` generation.** `ProcessData` now iterates `clientSecrets`, `clientIds`, and `tenantIds` in sorted-key order and clones the caller-provided ID maps before performing verification-driven `delete` operations, eliminating cross-run nondeterminism and input mutation.
> 
> Adds a new `spv2_determinism_test.go` suite to assert deterministic `RawV2` across repeated runs, document when `RawV2` is omitted due to ambiguous IDs, and ensure `ProcessData` does not mutate the caller’s maps during verification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a1c6a7b7c92d6cee7b9358a20f99a20a87eaf44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
+++ b/pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
@@ -100,6 +100,7 @@ func ProcessData(ctx context.Context, clientSecrets, clientIds, tenantIds map[st
 				}
 				tenantId = tId
 
+				// Skip known-invalid client/tenant combinations.
 				invalidClients := invalidClientsForTenant[tenantId]
 				if invalidClients == nil {
 					invalidClients = map[string]struct{}{}
@@ -111,9 +112,11 @@ func ProcessData(ctx context.Context, clientSecrets, clientIds, tenantIds map[st
 
 				if verify {
 					if !azure_entra.TenantExists(logCtx, client, tenantId) {
+						// Tenant doesn't exist.
 						delete(activeTenants, tenantId)
 						continue
 					}
+					// Tenant exists, ensure this isn't attempted as a clientId.
 					delete(activeClients, tenantId)
 
 					isVerified, extraData, verificationErr := serviceprincipal.VerifyCredentials(ctx, client, tenantId, clientId, clientSecret)
@@ -127,14 +130,17 @@ func ProcessData(ctx context.Context, clientSecrets, clientIds, tenantIds map[st
 							r = createResult(tenantId, clientId, clientSecret, false, nil, nil)
 							break ClientLoop
 						case errors.Is(verificationErr, serviceprincipal.ErrTenantNotFound):
+							// Tenant doesn't exist; shouldn't happen given the check above.
 							delete(activeTenants, tenantId)
 							continue
 						case errors.Is(verificationErr, serviceprincipal.ErrClientNotFoundInTenant):
+							// Tenant is valid but the client ID doesn't exist in it.
 							invalidClients[clientId] = struct{}{}
 							continue
 						}
 					}
 
+						// The result is verified or there's only one associated client and tenant.
 					if isVerified || (len(activeClients) == 1 && len(activeTenants) == 1) {
 						r = createResult(tenantId, clientId, clientSecret, isVerified, extraData, verificationErr)
 						break ClientLoop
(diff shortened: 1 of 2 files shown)


pkg/detectors/azure_entra/serviceprincipal/v2/spv2.go
```diff
@@ -138,11 +144,10 @@ SecretLoop:
 		}
 
 		if r == nil {
-			// Only include the clientId and tenantId if we're confident which one it is.
-			if len(clientIds) != 1 {
+			if len(activeClients) != 1 {
```
shahzadhaider1, line 154: I might be mistaken, but removing the comments feels a bit aggressive; the PR strips out almost all comments in ProcessData. What do you think?

dipto-truffle: [Agent-generated] Good call — restored the comments that explain non-obvious intent:

- "Skip known-invalid client/tenant combinations." (corrected from the misleading original "Skip known invalid tenants")
- "Tenant doesn't exist." / "Tenant exists, ensure this isn't attempted as a clientId." (cross-map deletion rationale)
- "Tenant doesn't exist; shouldn't happen given the check above." (defensive-check note)
- "Tenant is valid but the client ID doesn't exist in it."
- "The result is verified or there's only one associated client and tenant." (break condition)
- "Only include the clientId and tenantId if we're confident which one it is." (ambiguous-ID clearing)

Dropped only `// Handle errors.` since it was immediately above `if verificationErr != nil` and added no value.

shahzadhaider1 (approved):

camgunz (commented):
It seems plausible this is the right thing, but I have Qs:
* Why does the non-determinism matter here? I see that it generates different permutations, and that the order of those permutations will be different because Go map iteration order is inconsistent, but why does that cause problems? Mostly I'm wondering if it's interacting with overlap detection or some other downstream thing in a weird way.
* `slices.Sorted` necessarily makes a copy of the map's keys. We're already at O(N^3) here, so it's possible this would create a lot of new memory pressure if there are even moderate numbers of secrets/clients/tenants. That's why I'm wondering if we _really_ need to do it. If we do (totally possible), it might be worth reworking the algorithm here to avoid having 2 data structures (the map and the slice) with nearly exactly the same data.

PR #4864, opened 2026-04-03
Fix nil pointer panics in GitHub analyzer gist/repo binding functions

shahzadhaider1:
## Problem
A panic was caught in the GitHub analyzer worker:

```
runtime error: invalid memory address or nil pointer dereference
github.com/trufflesecurity/trufflehog/v3/pkg/analyzer/analyzers/github.secretInfoToGistBindings
    pkg/analyzer/analyzers/github/github.go:129
```

The GitHub API returns `description` and `owner` as optional nullable fields on gist objects. When a gist has no description, the API returns `null`, leaving `Description` as a nil `*string`. The code dereferenced it directly (`*gist.Description`) without any nil guard.

The same class of bug existed across several functions in the file: `*gist.Owner.Login`, `*gist.ID`, `*repo.Name`, `*repo.FullName`, `*user.Login`, `*user.Type`; all raw dereferences of fields that the go-github library defines as optional pointers.

## Why safe getters alone weren't enough for `Owner`
For `Description`, replacing the dereference with `GetDescription()` (returns `""` on nil) is sufficient, a gist without a description is still attributable.

For `Owner`, falling back to an empty string silently produces corrupt output:
- `FullyQualifiedName: "gist.github.com//abc123"`: malformed
- `Parent` resource with empty `Name` and `Type`

A gist with no owner cannot be meaningfully attributed, so it is **skipped** rather than emitting bad data. This is consistent with how `PrintGists` in `common/github.go` already handles nil gist pointers.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes GitHub permission analysis output in edge cases (skipping ownerless gists and emitting repos without a parent), but the changes are localized and add coverage for the nil-field scenarios that previously panicked.
> 
> **Overview**
> Prevents nil-pointer panics in the GitHub analyzer by switching to go-github safe getters (`GetLogin`, `GetType`, `GetName`, `GetFullName`, `GetDescription`, `GetID`) when building user/repo/gist resources.
> 
> Adjusts binding behavior for missing owners: repositories are still emitted with `Parent` unset if `Owner` is nil, while gists with a nil `Owner` are skipped to avoid generating malformed `FullyQualifiedName`s. Adds unit tests covering these nil-field edge cases for `secretInfoToGistBindings` and `secretInfoToRepoBindings`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43833316a95db6bd9ca0538905257060bb0af02c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/analyzer/analyzers/github/github.go
```diff
@@ -114,9 +114,14 @@ func secretInfoToRepoBindings(info *common.SecretInfo) []analyzers.Binding {
 	}
 	var bindings []analyzers.Binding
 	for _, repo := range repos {
+		// A repo without an owner cannot be attributed; skip it rather than
+		// producing a corrupt parent resource with empty name/FQN.
+		if repo.GetOwner() == nil {
```
MuneebUllahKhan222: Skipping a repo resource just because we don't know about it's owner doesn't seem right to me. Maybe we should set the parent resource as nil for such cases.

MuneebUllahKhan222 (approved):
LGTM! Just one small comment that we should think over

camgunz (approved):

pkg/analyzer/analyzers/github/github.go
camgunz: I probably agree; dunno what the analyzer implications are for a resource w/o a parent though (I am _far_ from an analyzer expert haha)

shahzadhaider1: Updated: keeping the repo binding with nil Parent when Owner is absent

PR #4872, opened 2026-04-07
[INS-402] Add Jira Data Center PAT Detector

mustansir14:
### Summary
Adds a new detector for [Jira Data Center Personal Access Tokens (PATs)](https://confluence.atlassian.com/enterprise/using-personal-access-tokens-1026032365.html).

### Regex
PATs are base64-encoded strings of the form `<12-digit-id>:<20-random-bytes>` (33 bytes, 44 chars, no padding). Since the first byte is always an ASCII digit, the first base64 character is always M, N, or O. The trailing boundary `(?:[^A-Za-z0-9+/=]|\z)` is used instead of `\b` to correctly handle tokens ending in `+` or `/`, while still rejecting matches that are a prefix of a longer or padded base64 string.

```
(?i:jira|atlassian)(?:.|[\n\r]){0,40}?\b([MNO][A-Za-z0-9+/]{43})(?:[^A-Za-z0-9+/=]|\z)
```

Server URLs are captured using the same keyword prefix:

```
(?i:jira|atlassian)(?:.|[\n\r]){0,40}?(https?://[A-Za-z0-9][A-Za-z0-9.\-]*(?::\d{1,5})?)
```

Both patterns require a `jira` or `atlassian` keyword within 40 characters to reduce false positives. Extracted URLs are tried alongside any user-configured endpoints.

### Verification
Verifies against `GET /rest/api/2/myself` using `Authorization: Bearer <token>`. Returns `display_name` and `email_address` as extra data on 200. Treats 401 as invalid and anything else as a verification error. Docs: https://developer.atlassian.com/server/jira/platform/rest/v10002/api-group-myself/#api-api-2-myself-get

### Tests
Pattern tests cover valid tokens, URL detection near `jira`/`atlassian` keywords, and negative cases. Verification tests use `gock` to mock the `/rest/api/2/myself` endpoint, covering verified, unverified (401), unexpected status, timeout, and no-verify cases.

Integration tests against a live Jira Data Center instance were not possible because Jira Data Center requires a paid license — there is no free tier or open-source image that runs fully without one. Unlike detectors such as Redis or MongoDB where a fully functional Docker container can be spun up freely, the `atlassian/jira-software` Docker image requires a valid license key to operate. Atlassian's evaluation licenses are time-limited and account-bound, making them unsuitable for automated CI.

### Corpora Test
The detector does not appear in the Corpora Test Results.
image
image

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Introduces new network-based verification logic and expands the default detector set, which could impact scan behavior and outbound request patterns despite being largely isolated and well-tested.
> 
> **Overview**
> Adds a new `JiraDataCenterPAT` detector that finds Jira Data Center Personal Access Tokens via a keyword-scoped base64 pattern plus a structural decode check, and can optionally verify tokens by calling `GET /rest/api/2/myself` with `Authorization: Bearer`.
> 
> Integrates the detector into the default engine detector set and updates the detector type protobuf/Go enums, with comprehensive unit tests covering matching/negative cases, URL+endpoint handling, verification success/failure/timeout, and the no-URL verification path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2449ee559038328ae9c633ae7d11ba24c5b68dcd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -0,0 +1,151 @@
...
+			s1 := detectors.Result{
+				DetectorType: detector_typepb.DetectorType_JiraDataCenterPAT,
+				Raw:          []byte(token),
+				RawV2:        []byte(token + endpoint),
+				Redacted:     token[:3] + "..." + token[len(token)-3:],
+			}
+
+			if verify {
+				isVerified, extraData, verificationErr := verifyPAT(ctx, s.getClient(), endpoint, token)
+				s1.Verified = isVerified
```
amanfcp, line 96: Had a look at Jira Analyzer and it seems it does support custom domains. Can you verify if the analyzer works for this detector too? If yes, we should add AnalysisInfo

amanfcp (approved):

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Ensure the Scanner satisfies the interfaces at compile time.
+var (
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
+)
+
+var (
+	defaultClient = common.SaneHttpClient()
```
shahzadhaider1: Worth switching from `common.SaneHttpClient()` to `detectors.DetectorHttpClientWithNoLocalAddresses` so RFC1918 addresses are blocked by default.
The other Jira detectors use it as well. What do you think?

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -0,0 +1,151 @@
...
+
+type Scanner struct {
+	client *http.Client
+	detectors.EndpointSetter
+}
+
+// Ensure the Scanner satisfies the interfaces at compile time.
+var (
+	_ detectors.Detector           = (*Scanner)(nil)
+	_ detectors.EndpointCustomizer = (*Scanner)(nil)
```
shahzadhaider1: Do you think this detector might be a candidate for `detectors.DefaultMultiPartCredentialProvider`?

mustansir14: Yes, good catch. Will add

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
mustansir14: Hmm, with this my only concern is that an on-prem instance could be hosted at an internal private server. Using `detectors.DetectorHttpClientWithNoLocalAddresses` would cause secrets for those to not be verified.

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
mustansir14: I checked the analyzer, and while it supports custom domains, it is only restricted to Jira Cloud. The authentication method and API it uses is not compatible with Jira On-Prem.

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// FromData will find and optionally verify Jira Data Center PAT secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	var tokens []string
+	for _, match := range patPat.FindAllStringSubmatch(dataStr, -1) {
+		tokens = append(tokens, match[1])
+	}
```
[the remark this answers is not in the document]
mustansir14: It's fine. The idea is to report multiple results for duplicate credentials.

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
rosecodym: @mustansir14 if a detector did find a live credential for one of those addresses, there's a very good chance that whoever is doing the remediation wouldn't be able to interact with the token anyway, because they might not have an easy way to get onto the network. I think you've identified a potentially valid gap that is nevertheless extremely narrow. Consistency with the other detectors also has value; I'm comfortable revisiting the exclusion of private addresses only if and when we actually find one in the wild.

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -0,0 +1,187 @@
...
+				DetectorType: detector_typepb.DetectorType_JiraDataCenterPAT,
+				Raw:          []byte(token),
+				Redacted:     token[:3] + "..." + token[len(token)-3:],
+				ExtraData:    map[string]string{"message": "No Jira Data Center URL was found or configured. To verify this token, set the Jira instance base URL as a custom endpoint."},
+			})
+			continue
+		}
+
+		for _, endpoint := range endpoints {
+			s1 := detectors.Result{
```
rosecodym, line 90: Should the result include the endpoint? It looks like we do that for Confluence Data Center.

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -0,0 +1,187 @@
...
+				DetectorType: detector_typepb.DetectorType_JiraDataCenterPAT,
+				Raw:          []byte(token),
+				RawV2:        []byte(token + endpoint),
```
rosecodym: I know there's no consistency across detectors, but my understanding is that adding a separator has generally been found to be useful.

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
mustansir14: That makes sense. I'll make the change

pkg/detectors/jiradatacenterpat/jiradatacenterpat.go
mustansir14: It does include the endpoint in `RawV2`

mustansir14: Added it in `ExtraData` if that's what you meant

rosecodym: Yeah, but users can't see that. They can see things in extra data.

rosecodym (approved):

PR #4884, opened 2026-04-14
[INS-444] Fix verification logic in Mesibo detector

mustansir14:
### Problem

The Mesibo detector was incorrectly marking unverified secrets as verified. The original implementation sent a `GET` request to `https://api.mesibo.com/api.php?op=useradd&token=<token>` and checked whether the response body contained the string `"AUTHFAIL"`. This endpoint no longer works correctly — it now returns 502 on all requests, meaning the `"AUTHFAIL"` string is never present and every candidate token is marked as verified regardless of its validity. Even setting aside the broken endpoint, checking for a specific error string in the body is an unreliable verification strategy compared to inspecting a structured response field.

### Fix

Switched to the current Mesibo backend API (`POST https://api.mesibo.com/backend`) as documented at https://docs.mesibo.com/api/backend-api/.

The backend API always returns HTTP 200. The actual outcome is encoded in the JSON response body via a `code` field, which contains an RFC 9110 compliant HTTP status code. We probe using the `useradd` operation:

- A **valid token** passes authentication but fails the operation due to missing required user parameters, yielding `code: 400`.
- An **invalid or unauthorized token** yields a different code (e.g. `401`).

The verification logic was refactored into a dedicated `verify()` method that:

1. POSTs a JSON payload with `op=useradd` and the candidate token.
2. Checks that the HTTP status is 200, returning a verification error otherwise.
3. Parses the JSON response into a typed `apiResponse` struct.
4. Returns `true` only if `code == 400`.

All errors are wrapped with context using `fmt.Errorf(..., %w)` and surfaced via `SetVerificationError`.

### Tests

Added a test case `"found, verification error on 502"` to the integration test suite using `common.ConstantResponseHttpClient(502, "")` to simulate an unexpected HTTP status from the API, asserting that the result is unverified and a verification error is set.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes live secret verification behavior and external API interaction; incorrect interpretation of Mesibo response codes could flip verified/unverified results, but the scope is limited to this detector and is covered by tests.
> 
> **Overview**
> Mesibo secret verification is rewritten to use the documented backend API (`POST https://api.mesibo.com/backend`) and to decide validity via the JSON `code` field (treating `400` as *auth succeeded* and `401` as unverified) instead of string-matching `AUTHFAIL` from the deprecated `api.php` endpoint.
> 
> The detector now encapsulates verification in a dedicated `verify()` method, supports injecting an HTTP client on `Scanner`, and surfaces unexpected HTTP/response conditions via `SetVerificationError`. Integration tests are updated to use `go-cmp` and add coverage for a verification-error scenario (simulated `502`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52cedea4e7d88650a1a2947cc9348d7d7f0683bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/mesibo/mesibo.go
+++ b/pkg/detectors/mesibo/mesibo.go
@@ -28,6 +28,8 @@ type apiResponse struct {
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
+	defaultClient = common.SaneHttpClient()
+
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"mesibo"}) + `\b([0-9A-Za-z]{64})\b`)
 )
@@ -36,7 +38,7 @@ func (s Scanner) getClient() *http.Client {
 	if s.client != nil {
 		return s.client
 	}
-	return common.SaneHttpClient()
+	return defaultClient
 }
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/mesibo/mesibo.go
+++ b/pkg/detectors/mesibo/mesibo.go
@@ -112,9 +112,15 @@ func (s Scanner) verify(ctx context.Context, token string) (bool, error) {
 	}
 	// The `code` field contains an RFC 9110 compliant HTTP status
 	// code indicating the outcome of the operation.
-	// code 400 means valid token (bad request due to missing params, but auth passed)
-	// any other code means invalid/unauthorized
-	return result.Code == http.StatusBadRequest, nil
+	switch result.Code {
+	case http.StatusBadRequest:
+		// code 400 means valid token (bad request due to missing params, but auth passed)
+		return true, nil
+	case http.StatusUnauthorized:
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected code field: %d", result.Code)
+	}
 }
 
 func (s Scanner) Type() detector_typepb.DetectorType {
--- a/pkg/detectors/mesibo/mesibo.go
+++ b/pkg/detectors/mesibo/mesibo.go
@@ -1,28 +1,43 @@
 package mesibo
 
 import (
+	"bytes"
 	"context"
-	regexp "github.com/wasilibs/go-re2"
+	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
+
+	regexp "github.com/wasilibs/go-re2"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detector_typepb"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
+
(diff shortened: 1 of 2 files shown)


pkg/detectors/mesibo/mesibo.go
```diff
@@ -70,6 +74,49 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	res, err := s.getClient().Do(req)
+	if err != nil {
+		return false, fmt.Errorf("failed to execute request: %w", err)
+	}
+	defer res.Body.Close()
+
+	// The backend API always returns HTTP 200, with the actual result encoded in the
+	// JSON response body.
+	if res.StatusCode != http.StatusOK {
+		return false, fmt.Errorf("unexpected status code: %d", res.StatusCode)
```
MuneebUllahKhan222, line 103: We should return `false, nil` for **401 or 403** because it falls in the category of **unverified result**(subset of expected outcome from detector). 
We only return `false, err` only in-case of unknown/unexpected results from the detector.

MuneebUllahKhan222 (changes requested):
Just need to address this one small change. Otherwise the PR is as good as it gets.

pkg/detectors/mesibo/mesibo.go
mustansir14: As mentioned in the PR description, the endpoint always returns a `200` status code. This is true for unauthorized/unauthenticated requests as well. So in case the status code is something other than `200`, we can consider it as a verification error.

shahzadhaider1 (approved):

pkg/detectors/mesibo/mesibo.go
MuneebUllahKhan222: Ok, but for cases other than `401 and 400`. Shouldn't we return an unexpected error? Probably before this [line](https://github.com/trufflesecurity/trufflehog/pull/4884/changes#diff-f6667485c38996a4edbcdd0d7e779ed61929d6ae070a3a024a90c17ccd72ff8bR117)

mustansir14: Oh yeah, that seems like a valid point. I'll make the change.

MuneebUllahKhan222 (approved):

pkg/detectors/mesibo/mesibo.go
```diff
@@ -70,6 +74,55 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	// code indicating the outcome of the operation.
+	switch result.Code {
+	case http.StatusBadRequest:
+		// code 400 means valid token (bad request due to missing params, but auth passed)
+		return true, nil
+	case http.StatusUnauthorized:
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected code field: %d", result.Code)
+	}
```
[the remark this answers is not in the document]
mustansir14: That is a mistake in the comment. Will fix that, but the API returns 401 for invalid/revoked tokens.

mariduv (approved):

alafiand (approved):

trufflesteeeve (approved):

PR #4886, opened 2026-04-14
Add Confluence Data Center PAT detector

amanfcp:
## Summary
Adds a detector for **Confluence Data Center Personal Access Tokens**.

- **No branded prefix.** DC PATs are plain 44-char base64. To avoid drowning in false positives, the detector base64-decodes each candidate and checks for the `<numeric_id>:<random_bytes>` structural shape at the byte level
- **URL regex captures `http?://host(:port)?`**, not just `https://`. On-prem Confluence commonly runs plain HTTP inside corporate networks and on non-standard ports (`:8090`, `:8443`).
- **Token-only emission.** When no URL is found in the chunk, we still emit an unverified result.
 
## Testing
- **Unit tests**: pattern coverage, verification coverage via `gock`.
- **No CI integration test.** Confluence Data Center requires a paid Atlassian license.
- **Manual end-to-end**: stood up Confluence DC locally via the official Docker image using Atlassian's [time-bomb licenses for testing server apps](https://developer.atlassian.com/platform/marketplace/timebomb-licenses-for-testing-server-apps/), created and verified valid PATs.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Adds a new network-verifying detector and a new protobuf enum value; risk is mostly around false positives/verification behavior and compatibility for consumers depending on stable detector type IDs.
> 
> **Overview**
> Adds a new `ConfluenceDataCenter` detector to identify and optionally verify Confluence Data Center Personal Access Tokens (PATs), including support for pairing tokens with on-prem instance base URLs found in the same chunk.
> 
> Detection uses keyword-scoped regexes plus a base64 decode *structural check* (expects `<numeric_id>:<random_bytes>`) to reduce false positives, emits token-only findings when no reachable URL is present, and verifies tokens via `GET /rest/api/user/current` with Bearer auth while caching unreachable hosts to avoid repeated DNS failures. The detector is wired into the default engine detector set, the “no cloud endpoint” test allowlist, and introduces a new `DetectorType_ConfluenceDataCenter` enum value with unit tests covering matching and verification behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 085c6814e76d18b2287911a66d66967508d8ad09. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/confluencedatacenter/confluencedatacenter.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (Scanner) CloudEndpoint() string { return "" }
+
+var (
+	defaultClient = detectors.DetectorHttpClientWithLocalAddresses
+
+	keywords = []string{"confluence", "atlassian", "wiki"}
+
+	// 44-char base64 PAT; decoded form must match the structural check below.
+	tokenPat = regexp.MustCompile(detectors.PrefixRegex(keywords) + `\b([A-Za-z0-9+/]{44})\b`)
```
mustansir14: Is there any way we can tighten this regex? Although I see that you have an additional check that checks the structure of the pattern, which is good.

Jira also has a similar pattern (same `<numeric id>:<random bytes>` structure). I found that having this structure also means that the credential will always start with an M, N or O. By any chance is that the same case with Confluence?

pkg/detectors/confluencedatacenter/confluencedatacenter.go
```diff
@@ -0,0 +1,182 @@
...
+	if colon <= 0 || colon == len(raw)-1 {
+		return false
+	}
+	for _, b := range raw[:colon] {
+		if b < '0' || b > '9' {
+			return false
+		}
+	}
+	return true
+}
```
mustansir14, line 86: I really like this. It's a great check for noise reduction. Will implement this in my Jira PR as well.

pkg/detectors/confluencedatacenter/confluencedatacenter.go
```diff
@@ -0,0 +1,182 @@
...
+
+	for token := range uniqueTokens {
+		if len(uniqueURLs) == 0 {
+			// Token-only: report unverified since we can't reach a host.
+			results = append(results, detectors.Result{
+				DetectorType: detector_typepb.DetectorType_ConfluenceDataCenter,
+				Raw:          []byte(token),
+				RawV2:        []byte(token),
+			})
+			continue
```
mustansir14: I like the idea of reporting tokens as unverified if no URL was configured/found. Even though it can be misleading in case the token was actually live, it is still better than not reporting it at all.

A suggestion: we can indicate this (as a message in `ExtraData` maybe?) that we were unable to verify the token because of absence of URL/host. That can serve as a good distinction for the user.

pkg/detectors/confluencedatacenter/confluencedatacenter.go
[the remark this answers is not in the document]
mustansir14: This is valid.

pkg/detectors/confluencedatacenter/confluencedatacenter.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return true, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
```
mustansir14: Are we sure that the API can return both 401 and 403 for an expired/invalid token? Usually it's one of these, and 403 can indicate lack of permissions instead of the credential being invalid.

mustansir14 (approved):
Overall looks good to me. I have some questions/suggestions which you can look into.

Also it seems the credential pattern for this is the same as [Jira Data Center](https://github.com/trufflesecurity/trufflehog/pull/4872), so there may be some overlap in results. I guess that's okay?

amanfcp:
image
image

The detector does not appear in the corpora test results

amanfcp:
Waiting on #4872  to get merged, will resolve conflicts then.

pkg/detectors/confluencedatacenter/confluencedatacenter.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	uniqueTokens := make(map[string]struct{})
+	for _, m := range tokenPat.FindAllStringSubmatch(dataStr, -1) {
+		if isStructuralPAT(m[1]) {
+			uniqueTokens[m[1]] = struct{}{}
+		}
```
rosecodym, line 98: I'm not sure how expensive or hot this code is, but it looks like you could avoid some work by checking for set presence before checking the match's structure, right? (If it's in the set, you don't need to check to see whether it's a PAT.)

rosecodym (approved):
Please tell me if I've got this right: If we find only a single candidate URL in a chunk, and that URL doesn't resolve, we return all findings as determinately unverified. This means that something like a transient DNS error could cause determinate unverification even though for our other detectors it doesn't do that.

I don't say this because I see an obvious way around it - doing so seems like it would need some sort of heuristic analysis, which would be new ground for us. I just want to ensure I understand the current implementation.

(I also left a non-blocking note about a possible optimization, which I'll leave up to your discretion.

amanfcp:
> Please tell me if I've got this right: If we find only a single candidate URL in a chunk, and that URL doesn't resolve, we return all findings as determinately unverified. This means that something like a transient DNS error could cause determinate unverification even though for our other detectors it doesn't do that.
> 
> I don't say this because I see an obvious way around it - doing so seems like it would need some sort of heuristic analysis, which would be new ground for us. I just want to ensure I understand the current implementation.
> 
> (I also left a non-blocking note about a possible optimization, which I'll leave up to your discretion.

You've got the trace right, thanks for catching this.

To name what we actually lose: tokens after the first one land in the **determinate-unverified** bucket (no verification error attached), while the first token correctly lands in **indeterminate** with the DNS error. Downstream tooling that filters "unverified" as noise will drop tokens 2..N, even though they hit the same transient failure. I agree this inconsistency is a real problem, it's not a theoretical edge case, and I don't want to wave it away.

**Why the fallback exists in the first place**: this is a deliberate departure from how `jiratoken` / `artifactoryreferencetoken` and other similar detectors handle the no-URL case. Those detectors target hosted products where either 
    (a) a token without an obvious URL is meaningless, or 
    (b) the URL is a branded domain that's trivially inferable. 
    
Confluence Data Center has neither property: the tokens are unbranded 44-char base64, and the URL is a self-hosted instance that could be anywhere. The two happy paths for this detector are 
    (1) the operator configured `--confluence-data-center-endpoint`, or 
    (2) the URL is co-located with the token in the chunk. 
    
The sad path is "token exists, no URL context, no configured endpoint", and for on-prem secrets, that sad path is exactly where silent data loss hurts most. An operator who forgets the flag and scans a repo with real leaked PATs could see zero Confluence findings and conclude they're clean.

So the token-only emission is less a fallback than an **intimation**: "we found something that structurally matches a Confluence DC PAT, we couldn't verify it because we had nowhere to reach, here's a note explaining why." The `verification_note` is deliberately written so operators who missed the endpoint flag get a clear prompt to configure it and rescan.

Keeping the current behavior for this PR trades classification fidelity for N-1 tokens in a rare path against guaranteed surfacing of all structurally-valid tokens, which for on-prem feels like the right way to err.

On the 2..N mismatch: I don't have a confident answer on the right fix yet. I'd suggest we think it through properly and ship it separately than bundle a half-considered change into this PR. Happy to pick it up as a follow-up. What do you think?

pkg/detectors/confluencedatacenter/confluencedatacenter.go
```diff
@@ -0,0 +1,204 @@
...
+	if colon <= 0 || colon == len(raw)-1 {
+		return false
+	}
+	for _, b := range raw[:colon] {
+		if b < '0' || b > '9' {
+			return false
+		}
+	}
+	return true
+}
```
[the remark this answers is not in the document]
amanfcp: We can defer this. I've created a separate ticket for this.

PR #4892, opened 2026-04-16
adding customizable successRanges and rotatedRanges to customDetector

jordanTunstill:
### Description:
This would allow users to add successRanges and rotatedRanges to their customDetectors, for verification purposes, allowing them control/configuration over what is considered live/rotated for these custom detections.
If both successRanges and rotatedRanges are implemented, any response that is not within said ranges is treated as inconclusive. Here is how the change handles edge cases:
Only successRanges set: non-matching responses mean rotated
Only rotatedRanges set: non-matching responses mean live
Neither set: old behavior, only 200 means verified; other statuses are just unverified without a verification error.
### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes verification behavior for custom detectors by interpreting verifier HTTP status codes via configurable ranges and introducing new "inconclusive" verification errors, which could affect existing workflows if misconfigured. Backward compatibility is preserved when no ranges are provided, but verification outcomes and error reporting now depend on additional config.
> 
> **Overview**
> Adds configurable HTTP status handling for custom detector webhook verification by introducing `rotatedRanges` alongside existing `successRanges`, allowing verifiers to explicitly mark secrets as *live*, *rotated*, or *inconclusive* based on response codes.
> 
> Updates the verifier loop to evaluate these ranges (including single-range fallbacks and legacy no-range behavior), records truncated response bodies via a shared helper, and emits a verification error when ranges are configured but no verifier returns a definitive match.
> 
> Extends the protobuf schema (`VerifierConfig`) and YAML parsing/validation to support `rotatedRanges`, adds `StatusCodeMatchesRanges`, and introduces focused unit tests covering parsing, range matching, mixed legacy+ranged verifiers, and edge-case status handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d12d4d9166f8f18b82e93a957264afde52d36c0e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/custom_detectors/custom_detectors.go
```diff
@@ -302,27 +313,54 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
+				storeResponseBody(resp, result.ExtraData)
+			}
+			break
+		}
+
+		// Both configured but neither matched -- try the next verifier.
+	}
+
+	if rangesInEffect && !definitive {
+		result.SetVerificationError(fmt.Errorf("verification response status code did not match any configured successRanges or rotatedRanges"))
```
kashifkhan0771: No format args in `fmt.Errorf` we can use errors.New.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -302,27 +313,54 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
-				continue
+		if len(successRanges) == 0 && len(rotatedRanges) == 0 {
+			// Backward compat: no ranges configured, use legacy behavior.
+			if resp.StatusCode == http.StatusOK {
+				result.Verified = true
+				definitive = true
+				storeResponseBody(resp, result.ExtraData)
+				break
 			}
+			continue
```
kashifkhan0771, line 331: I think we need to set a `definitive` as true here before continue for legacy non 200 path.

Scenario that can trigger without this:
```
Verifier 1: ranged, both configured, returns 500 or any code not in ranges
  - rangesInEffect = true                                                                                                                                                                   
  - bothConfigured = true, neither range matches
  - loop continues (no break, no definitive)                                                                                                                                                
                                                                                                                                                                                              
Verifier 2: legacy (no ranges), returns 401
  - hits continue at line 327                                                                                                                                                             
  - definitive never set
          
After loop:
  - rangesInEffect = true, definitive = false
  - SetVerificationError fires

kashifkhan0771 (approved):

pkg/custom_detectors/custom_detectors.go
jordanTunstill: Thank you, just made this change

pkg/custom_detectors/custom_detectors.go
jordanTunstill: done, thank you!

rosecodym (approved):

PR #4907, opened 2026-04-22
[INS-455] Unify common logic in Atlassian Data Center detectors

mustansir14:
## Background

Three detectors target Atlassian Data Center (self-hosted) products: `JiraDataCenterPAT`, `ConfluenceDataCenter`, and `BitbucketDataCenter`. Reviewing them side-by-side revealed significant duplication: identical structural-validation logic, the same bearer-auth HTTP plumbing, and the same URL-extraction pipeline repeated verbatim in each file.

### New shared package: `pkg/detectors/atlassiandatacenter/`

Following the pattern established by `pkg/detectors/aws/`, the three detectors are now co-located under a common parent that also houses shared utilities in `common.go`.

**`GetDCTokenPat(prefixes []string) *regexp.Regexp`**
Returns the compiled PAT regex for Jira/Confluence DC tokens (44-char base64 strings whose decoded form is `<numeric-id>:<random-bytes>`). Previously each detector hand-rolled the same pattern with inline literal strings.

**`IsStructuralPAT(candidate string) bool`**
Validates that a base64 candidate decodes to the `<digits>:<bytes>` structure. This function was copy-pasted verbatim between Jira and Confluence.

**`GetURLPat(prefixes []string) *regexp.Regexp`**
Returns the compiled URL regex for self-hosted Atlassian instance URLs. Each detector calls this once at package init time and stores the result in a package-level var, preserving the same compile-once behaviour as before. `go-re2` compiles via CGo/FFI so doing this per-chunk would be a meaningful regression.

**`FindEndpoints(data string, urlPat *regexp.Regexp, resolve func(...string) []string) []string`**
Extracts keyword-scoped URLs from a chunk using the pre-compiled `urlPat`, passes them through `s.Endpoints` (which merges configured endpoints), deduplicates, and strips trailing slashes. All three detectors had their own multi-step version of this pipeline.

**`MakeVerifyRequest(ctx, client, fullURL, token string) (bool, map[string]any, error)`**
Sends a Bearer-authenticated GET and interprets the response: `200` → `(true, decoded-JSON-body, nil)`, `401` → `(false, nil, nil)`, other → `(false, nil, error)`. Previously each detector built the request inline and contained its own copy of the `200 / 401 / default` status-code switch. Jira reads `body["displayName"]` and `body["emailAddress"]` from the returned map; Confluence and Bitbucket discard it.

### Important Note

It appears that Bitbucket DC detector was introduced but never registered in `defaults.go`. This PR also registers it. With that in mind, and of course to test the other changes as well, I also ran the usual corpora tests we do for new detectors.

### Detector changes

Each detector now imports the shared package and delegates to it:

| Before | After |
|--------|-------|
| Private `isStructuralPAT` (Jira + Confluence) | `atlassiandatacenter.IsStructuralPAT` |
| Inline PAT regex literal (Jira + Confluence) | `atlassiandatacenter.GetDCTokenPat(keywords)` |
| 10–15 line URL-find + dedup + resolve block (all three) | `atlassiandatacenter.FindEndpoints(...)` |
| Inline request build + status-code switch (all three) | `atlassiandatacenter.MakeVerifyRequest(...)` |

The URL regex is also standardised across all three detectors. Confluence and Bitbucket previously used an unbounded `\d+` port pattern and allowed hostnames starting with `.` or `-`; all three now use `[a-zA-Z0-9][a-zA-Z0-9.\-]*` with `\d{1,5}` for the port (matching the stricter Jira original).

A `keywords` package-level variable is introduced in Jira (Confluence and Bitbucket already had one) so the keyword list is defined once and reused by the regex, `FindEndpoints`, and `Keywords()`.

### `TestIsStructuralPAT` migrated

The structural-PAT test previously lived in `jiradatacenterpat_test.go` and called the private function. It is now in `common_test.go` and covers tokens from all three products. `common_test.go` also adds tests for `GetDCTokenPat`, `FindEndpoints`, and `MakeVerifyRequest`.

## What did not change

- Each detector retains its own package, `Scanner` type, `Keywords()`, `Type()`, `Description()`, and `FromData()` logic.
- Jira's JSON response parsing (`displayName`, `emailAddress`) is unchanged.
- Confluence's `invalidHosts` cache and `errNoHost` sentinel are unchanged.
- Bitbucket's `?limit=1` query parameter is unchanged.
- All existing tests pass without modification.

## Corpora Tests
image
image

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to shared refactor of token/URL matching and verification request handling across multiple detectors, plus enabling `BitbucketDataCenter` in the default detector set which can change scan coverage and findings volume.
> 
> **Overview**
> Introduces a new shared `atlassiandatacenter` package that centralizes Data Center PAT regexes, structural token validation, keyword-scoped URL extraction/resolution, and the common Bearer-auth verification GET flow.
> 
> Updates the Jira, Confluence, and Bitbucket DC detectors to delegate to these shared helpers (including a standardized URL pattern) and removes duplicated in-detector implementations; Jira still extracts `displayName`/`emailAddress` from the shared verification response.
> 
> Registers `BitbucketDataCenter` in `defaults.DefaultDetectors()` (and updates related tests), and adds dedicated unit tests for the new shared helpers in `common_test.go`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05e1280513eac369fa9b6153c7555eacdcae1abb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MuneebUllahKhan222:
This is completely irrelevant to the goal of this PR. But can you please add similar logic in `bitbucketdatacenter` detector
```
if !emitted {
	// No reachable URL in context — emit an unverified token-only
	// result and annotate why we couldn't verify.
	results = append(results, detectors.Result{
		DetectorType: detector_typepb.DetectorType_ConfluenceDataCenter,
		Raw:          []byte(token),
		RawV2:        []byte(token),
		ExtraData: map[string]string{
			"verification_note": "no reachable Confluence Data Center URL found in context; token reported unverified",
		},
	})
}
```

pkg/detectors/atlassiandatacenter/confluencedatacenter/confluencedatacenter.go
```diff
@@ -170,35 +136,13 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 func verifyPAT(ctx context.Context, client *http.Client, baseURL, token string) (bool, error) {
 	endpoint := strings.TrimRight(baseURL, "/") + "/rest/api/user/current"
```
MuneebUllahKhan222, line 143: No need to do `TrimRight` here because we are already doing that in `FindEndpoints` function.

pkg/detectors/atlassiandatacenter/bitbucketdatacenter/bitbucketdatacenter.go
```diff
@@ -105,30 +91,8 @@ func verifyMatch(ctx context.Context, client *http.Client, secretPat, baseURL st
 	q.Set("limit", "1")
 	u.RawQuery = q.Encode()
```
MuneebUllahKhan222, line 96: We can get rid of all this and do
```
baseURL + "/rest/api/1.0/projects?limit=1"
```

pkg/detectors/atlassiandatacenter/jiradatacenterpat/jiradatacenterpat.go
```diff
@@ -122,65 +107,19 @@ func verifyPAT(ctx context.Context, client *http.Client, baseURL, token string)
 	}
 	u.Path = "/rest/api/2/myself"
```
MuneebUllahKhan222, line 113: We can get rid of all this and do
```
baseURL + "/rest/api/2/myself"
```

MuneebUllahKhan222 (changes requested):
Couple of changes that need to be addressed, Other than that the PR is good to go.

pkg/engine/defaults/defaults.go
```diff
@@ -966,6 +967,7 @@ func buildDetectorList() []detectors.Detector {
 		&billomat.Scanner{},
 		&bingsubscriptionkey.Scanner{},
 		&bitbar.Scanner{},
+		&bitbucketdatacenter.Scanner{},
```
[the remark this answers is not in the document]
mustansir14: Yes, this was missed in the original PR. We want to enable it. Added a note about this in the PR description as well

mustansir14:
> This is completely irrelevant to the goal of this PR. But can you please add similar logic in `bitbucketdatacenter` detector
> 
> ```
> if !emitted {
> 	// No reachable URL in context — emit an unverified token-only
> 	// result and annotate why we couldn't verify.
> 	results = append(results, detectors.Result{
> 		DetectorType: detector_typepb.DetectorType_ConfluenceDataCenter,
> 		Raw:          []byte(token),
> 		RawV2:        []byte(token),
> 		ExtraData: map[string]string{
> 			"verification_note": "no reachable Confluence Data Center URL found in context; token reported unverified",
> 		},
> 	})
> }
> ```

I don't really think that's a good idea, as this will add more complexity to the PR and will require additional tests. I would say we stick to the goal of this PR and do this in a separate optimizations PR.

pkg/detectors/atlassiandatacenter/bitbucketdatacenter/bitbucketdatacenter.go
mustansir14: I was trying to make as minimal changes as possible to the detectors' specific logic. Also we generally prefer to use `net/url` to work with URLs and paths. This is valid for your other comment as well.

pkg/detectors/atlassiandatacenter/confluencedatacenter/confluencedatacenter.go
mustansir14: You might be right here, but doing so would add to the changes in this PR and my goal is to keep detector specific changes as minimal as possible. Also it doesn't really harm to have additional check

pkg/detectors/atlassiandatacenter/jiradatacenterpat/jiradatacenterpat.go
mustansir14: Same comment as above

MuneebUllahKhan222 (approved):
The changes I requested earlier seems to be out of scope for this PR and maybe we can cater them in some other PR. So I am approving this.

rosecodym (approved):

PR #4910, opened 2026-04-22
Fix line numbers for duplicate secrets within a chunk

### LINKED ISSUE #2502: rew

### TruffleHog Version
v3.4.3 to present

### Trace Output
N/A

### Expected Behavior
Should calculate the line of code correctly

### Actual Behavior
The line of code is not correctly calculated

### Steps to Reproduce
1. Create a file with multiple instances of the exact same secret, one after another
2. Scan the file with TruffleHog v3.4.3 or later

## Environment
 * OS: All
 * Version: v3.4.3 or later

## Additional Context
Root cause analysis documented in detail here: https://github.com/0x736E/trufflehog_v3_loc_bug

When there are multiple secrets of the same type, in the same data "chunk", TruffleHog incorrectly reports all secrets with the same raw value as having the same line of code.

### References

* #1876
* #1031

amanfcp:
## Summary

When the same secret appears multiple times in a chunk, every finding was reported with the line number of the first occurrence. This fixes it so each finding reports the line where it actually sits.

Closes #2502

## Root cause

`FragmentLineOffset` in `pkg/engine/engine.go` uses `bytes.Cut(chunk.Data, secret)`, which always splits at the first match. Because each `detectors.Result` only carries the secret bytes (not its byte position in the chunk), every duplicate result went through this first-match lookup and inherited the same line number.

Duplicates survive to this function under normal scans. `CleanResults` deduplication only runs under `--only-verified`, or when a detector's `ShouldCleanResultsIrrespectiveOfConfiguration()` returns true. By default, any detector that emits one result per regex match hits the bug.

## Fix

Localized to the engine. No detector changes required.

1. **`pkg/detectors/detectors.go`**: add a private `chunkOffset` / `chunkOffsetSet` pair to `Result` with `SetChunkOffset` / `ChunkOffset` / `HasChunkOffset` accessors.
2. **`pkg/engine/engine.go`**: add `AssignDuplicateLineOffsets(chunk, results)`, which groups results by secret value, walks the chunk with `bytes.Index` to find each successive occurrence, and stamps the offset onto the corresponding result. Called once per result batch inside `detectChunk`, before results are dispatched. Unique secrets are skipped (zero overhead).
3. **`FragmentLineOffset`**: gains a fast path. When `result.HasChunkOffset()` is true, compute the line directly from the pre-assigned offset and run the ignore-tag check against *that* occurrence's line. The original `bytes.Cut` logic is preserved as a fallback for any caller that didn't go through `AssignDuplicateLineOffsets`, so the change is backward compatible.
4. **Test fixtures**: `pkg/detectors/datadogapikey/datadogapikey_test.go` and `pkg/custom_detectors/custom_detectors_test.go` compared `detectors.Result` values via `cmp.Diff` with `IgnoreFields` enumerating the then-existing unexported fields. Switched them to `cmpopts.IgnoreUnexported(detectors.Result{})` so they tolerate the new private fields and are future-proof against similar additions.

A side effect worth calling out: the `trufflehog:ignore` check now runs per-occurrence instead of always inspecting the first occurrence's line. That's a real behavioral improvement and is covered explicitly by a test.

## Why not fix it in detectors?

Threading byte offsets through every detector's `FromData` would touch 700+ implementations and require switching most of them from `regex.FindAllStringSubmatch` to `FindAllStringIndex`. The engine is the single place in the pipeline that has both the full chunk data *and* the full result batch, so a localized engine-level fix is the minimal correct change.

## Performance

- `Result` struct: 160 → 176 bytes (+16 bytes for the int64 + bool + alignment padding). Results are short-lived per-chunk, so there is no meaningful memory pressure.
- `AssignDuplicateLineOffsets`: O(N) with one map allocation per call. The inner `bytes.Index` loop only runs for groups with duplicates.

## End-to-end verification

Reproduced the reporter's scenario against a fresh filesystem scan using a `CustomRegex` detector and a file containing `FAKE_SECRET_ABC123XYZ` on lines 2, 5, and 8:

| Binary | Reported lines |
|---|---|
| baseline | `line=2, line=2, line=2` |
| with this fix | `line=2, line=5, line=8` |

## Test plan

- [x] `TestFragmentLineOffset_DuplicateSecrets`: regression test for the bug (fails without the fix, passes with it)
- [x] `TestAssignDuplicateLineOffsets`: unit test covering unique secrets, duplicates, and result ordering
- [x] `TestFragmentLineOffset_DuplicateSecretsWithIgnoreTag`: verifies per-occurrence `trufflehog:ignore` handling
- [x] Existing `TestFragmentLineOffset` and `TestFragmentLineOffsetWithPrimarySecret*` still pass (fallback path unchanged)
- [x] Full `go test ./pkg/engine/...` and `go test ./pkg/detectors/...` pass
- [x] Manual CLI verification against a `CustomRegex` detector matches the expected line numbers

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches the engine’s line-number/ignore-tag calculation, which affects how findings are reported; logic is localized and covered by new regression tests, but could impact reporting accuracy across sources.
> 
> **Overview**
> Fixes incorrect line-number reporting when the *same secret appears multiple times in a single chunk* by adding per-result chunk byte offsets and using them when computing line offsets.
> 
> The engine now pre-assigns offsets for duplicate secrets via `AssignDuplicateLineOffsets`, and `FragmentLineOffset` uses these offsets so each occurrence (and its `trufflehog:ignore` check) maps to the correct line. Tests are added for duplicate-secret cases (including ignore-tag behavior), and a couple of detector tests switch to `cmpopts.IgnoreUnexported(detectors.Result{})` to tolerate the new private `Result` fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e61c20da3b0fdfb98b9e811c288fc6bb0d1e162e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/engine/engine.go
```diff
@@ -1335,7 +1337,27 @@ func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64,
...
+
+		afterSecret := chunk.Data[offset+len(secretBytes):]
+		endLine := bytes.Index(afterSecret, []byte("\n"))
+		if endLine == -1 {
+			endLine = len(afterSecret)
+		}
+		if bytes.Contains(afterSecret[:endLine], []byte(ignoreTag)) {
+			return lineNumber, true
+		}
+		return lineNumber, false
```
mustansir14: It seems we have duplicated this logic from below. Can we unify it somehow?

mustansir14 (approved):
Thanks for fixing this long overdue bug!

pkg/engine/engine.go
```diff
@@ -1352,6 +1367,49 @@ func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64,
...
+		indices []int
+	}
+	seen := make(map[string]int) // secret -> index into groups slice
+	var groups []group
+
+	for i := range results {
+		secret := results[i].GetPrimarySecretValue()
+		if secret == "" {
+			secret = string(results[i].Raw)
+		}
```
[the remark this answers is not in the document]
amanfcp: addressed

mustansir14 (approved):

pkg/detectors/detectors.go
```diff
@@ -173,6 +178,22 @@ func (r *Result) GetPrimarySecretValue() string {
...
+
+// ChunkOffset returns the byte position of this result's secret within the chunk data.
+func (r *Result) ChunkOffset() int64 {
+	return r.chunkOffset
+}
+
+// HasChunkOffset reports whether a chunk offset has been explicitly set on this result.
+func (r *Result) HasChunkOffset() bool {
+	return r.chunkOffsetSet
+}
```
kashifkhan0771, line 198: Can't we just use one field `chunkOffset` and by default set it to a negative value `-1`. If the value is `-1` that means it is not set else return whatever the value is.

amanfcp: Considered this, but kept the two-field design deliberately. The zero value of int64 is 0, which is a valid offset (the first byte of the chunk). For a -1 sentinel to mean "not set," every Result{...} literal in the codebase would need to explicitly initialize chunkOffset: -1. That's ~800+ detector callsites. With the current chunkOffset + chunkOffsetSet pair, a zero-value Result{} naturally means "not set" and falls back to the existing bytes.Index path. No detector changes required, which was the whole point of keeping this fix engine-only.

The two-field invariant is also safe by construction: both fields are private, SetChunkOffset always sets both, and HasChunkOffset / ChunkOffset are the only read paths, so external code can't produce a half-set state.

kashifkhan0771 (approved):

gaborbernat (approved):
Validated the fix end-to-end against a real internal repo (`DCA/PreppyService preppy/modules/config.py`) that had 4 identical `[FAIL]` values at lines 87, 95, 103, 111.

**Before (upstream/main):** all 4 reported as line 87
**After (this PR):** correctly reported as 87, 95, 103, 111

Tested with `trufflehog filesystem --json --config=... --include-detectors=CustomRegex --no-verification`. Both the synthetic case (3 identical secrets) and the real-world case pass.

mariduv (approved):

0x736E:
very happy to see this fixed, great job 👍

PR #4913, opened 2026-04-22
Add a static check for detectors that don't set SecretParts

mcastorina:
## Summary

Adds a small static-analysis tool under `hack/checksecretparts/` that scans `pkg/detectors/` and warns about detector packages that construct `detectors.Result` values without ever populating the new `SecretParts` field.

It's wired into the `Lint` workflow as **warning-only** (`continue-on-error: true`): findings show up in CI but don't block merges. The intent is to land the check *before* every detector is migrated to aid in migrating the packages.

Stacked on #4911 (the rename that introduces `SecretParts`); that needs to land first.

## Why

`SecretParts` replaces the old `AnalysisInfo` field and is expected to be populated by every detector going forward. Today only ~11 detectors set it. Rather than wait for a big-bang migration, this PR adds the enforcement mechanism first so each incremental migration PR gets immediate feedback, and so the flip from warning to hard-fail becomes a one-line workflow change once the migration is complete.

## What it does

- For each directory under `pkg/detectors/`, finds composite literals of the form `detectors.Result{...}` or `&detectors.Result{...}` in non-test `.go` files.
- If the package never mentions `SecretParts` in any non-test source (neither in the literal nor in a later `x.SecretParts = ...` assignment), emits one warning per construction site.
- Ignores `_test.go` files on both sides — some tests zero the field for cmp comparisons (e.g. `pkg/detectors/gitlab/v1/gitlab_integration_test.go`) and those references would otherwise mask real findings.

No new module dependencies — just `go/ast` and `go/parser`.

## How to run

```sh
# Warning mode (default), exit 0 even with findings:
go run ./hack/checksecretparts

# Fail mode, exit 1 on findings — for use after the full migration:
go run ./hack/checksecretparts -fail

# Scoped to specific packages:
go run ./hack/checksecretparts ./pkg/detectors/aws ./pkg/detectors/github
```

Full instructions for flipping warning → hard-fail are in `hack/checksecretparts/README.md`.

## Verification

- `go test ./hack/checksecretparts/` — 8 cases covering missing-field, literal-populated, later-assigned, pointer literal, multiple sites, test-file-only references (must not suppress), and no-op packages.
- Against the current tree: 825 findings. Spot-checked that every detector already known to populate `SecretParts` (postmark, databricks, figma, ngrok, monday, airbrake, digitalocean, datadog, tableau) is correctly exempted.
- `go vet ./...` and `go build ./...` are clean.

## Out of scope

- Populating `SecretParts` on any real detector — that's a separate migration.
- Flipping the check to hard-fail — to be done in a one-liner follow-up once migration completes.
- Deeper dataflow (per-code-path verification). The package-level heuristic is deliberately coarse; it's enough to surface unmigrated detectors and doesn't risk false positives on the migration PRs themselves.

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a new warning-only CI/static-analysis tool and tests without changing runtime detector behavior; main impact is additional lint workflow time/output.
> 
> **Overview**
> Adds `hack/checksecretparts`, a small Go AST-based scanner that reports (and optionally fails on) detector packages that build `detectors.Result{}` / `&detectors.Result{}` without any non-test reference to `SecretParts`, plus unit tests and documentation.
> 
> Wires the scanner into `.github/workflows/lint.yml` as a **warning-only** `checksecretparts` job (`continue-on-error: true`) to surface unmigrated detectors in CI without blocking merges.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc9bdaa9fbad796a02678ba622da08f0658c62f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -50,3 +50,18 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: semgrep --config=hack/semgrep-rules/detectors.yaml pkg/detectors/
+  checksecretparts:
+    # Warning-only: reports detector packages that construct detectors.Result
+    # without populating SecretParts. Flip to hard-fail (drop continue-on-error
+    # and run with -fail) after every detector has been migrated. See
+    # hack/checksecretparts/README.md.
+    name: checksecretparts (warning)
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.25"
+      - name: Run checksecretparts
+        run: go run ./hack/checksecretparts ./pkg/detectors
--- a/hack/checksecretparts/README.md
+++ b/hack/checksecretparts/README.md
@@ -1,0 +1,59 @@
+# checksecretparts
+
+Static analysis check that finds detector packages which construct
+`detectors.Result` values without populating the `SecretParts` field.
+
+## What it checks
+
+For each directory under `pkg/detectors/` (recursing into subpackages):
+
+1. Find every composite literal of the form `detectors.Result{...}` or
+   `&detectors.Result{...}` in non-test `.go` files.
+2. If the package does not mention `SecretParts` anywhere (neither in the
+   literal nor in a later `x.SecretParts = ...` assignment), emit a warning
+   for each construction site.
+
+Test files (`_test.go`) are ignored on both sides — construction sites in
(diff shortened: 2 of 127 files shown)


rosecodym (approved):
I only barely skimmed the new check itself, but I like the approach and the test suite.

kashifkhan0771 (approved):

trufflesteeeve (approved):

PR #4915, opened 2026-04-23
[INS-461] Add test to ensure new detectors are registered in defaults.go

mustansir14:
## Motivation

TruffleHog has over 1,000 defined `DetectorType` proto enum entries, each corresponding to a secret detector. The `buildDetectorList()` function in `pkg/engine/defaults/defaults.go` is the single place where all active detectors are registered and made available to the engine. However, there was no CI check to verify that a newly written detector is actually registered there. A developer could write a complete, working detector — with its own package, `Scanner` struct, keywords, and `Type()` — and simply forget to add it to `buildDetectorList()`, causing it to silently never run.

## What changed

`TestAllDetectorTypesAreInDefaultList` is added to `pkg/engine/defaults/defaults_test.go`. It iterates over every `DetectorType` defined in the protobuf enum and asserts that each one is either:

- present in the list returned by `DefaultDetectors()`, or
- explicitly listed in `excludedFromDefaultList` with a documented reason.

A reverse check is also included: if a type is in `excludedFromDefaultList` but is also present in the active list, the test fails. This catches stale exclude-list entries (e.g. a detector that was once disabled but has since been re-enabled).

The `excludedFromDefaultList` map groups excluded types into four categories:

- **Mistakenly missed** — implementations exist but were never registered (see below)
- **Reserved / special** — e.g. `CustomRegex`, which is added dynamically via engine config
- **Deprecated** — proto fields marked `deprecated=true`; the underlying service no longer exists or has been superseded
- **Intentionally disabled** — implementations exist but are commented out in `buildDetectorList()` due to API issues, false positives, or expired service domains
- **Not yet implemented** — proto enum entries reserved for future detectors that have not been written yet

## Detectors found to be unregistered

Running the test immediately surfaced detectors with full implementations that were never added to `buildDetectorList()`:

`BitbucketDataCenter` (this is added in #4907 ), `DatadogApikey`, `Guru`, `IPInfo`, `Lob`, `Rev`, `TLy`, `Tru`, `User`, `Wit`

These detectors are **not added to `buildDetectorList()` in this PR**. Enabling a detector that has never been active is a customer-facing change: it can produce noise, and this is something we have been careful about while adding new detectors. Each detector should be enabled in a dedicated PR after:

1. Verifying the detector still works correctly against its target service
2. Running corpora tests to assess the volume and quality of new findings
3. Confirming the customer impact is acceptable

In the meantime, these types are tracked in `excludedFromDefaultList` under a `TODO` comment so the gap remains visible and the test continues to pass.

---

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Low risk: adds test-only coverage to prevent new `DetectorType` enums from being silently unregistered; no runtime detector selection logic changes.
> 
> **Overview**
> Adds `TestAllDetectorTypesAreInDefaultList` to assert every proto `DetectorType` is either present in `DefaultDetectors()` (via `buildDetectorList()` in `defaults.go`) or explicitly listed in a new `excludedFromDefaultList` allowlist.
> 
> The test also performs a reverse check to fail if an excluded type becomes active, and introduces a categorized exclusion map (including a TODO group for implemented-but-not-yet-enabled detectors) to make intentional omissions explicit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b04da359c071d229598c4d257121b22001ec769b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MuneebUllahKhan222 (approved):
This test should have been there all along but I guess better late than never. Great Work!

amanfcp (approved):

shahzadhaider1 (approved):

rosecodym (approved):

PR #4927, opened 2026-04-29
Automate corpora testing in CI

mustansir14:
## Motivation

When adding or modifying a detector, the key question is: how much noise will this regex produce against real-world code? Too many false positives means alert fatigue; a regex that's too tight misses real secrets.

Previously this was a fully manual process — download a large corpus locally, run the pipeline, inspect the DuckDB output. There was no enforcement in CI, so it was easy to skip or forget, especially under time pressure.

This PR automates it. Every PR that touches detector regex or keyword changes now gets a comment showing exactly how many unique matches the changed detector produces, compared to the main baseline. The workflow runs once on PR open; reviewers or authors can re-run it manually via `workflow_dispatch` if needed.

## What it tells you

The bench scans a corpus of real-world public code (~2.8 GB compressed, ~29 GB uncompressed) using **only the detectors changed in the PR**, with verification disabled. It reports unique match counts for the PR build vs. the main baseline:

- **New matches** — the regex picks up strings main didn't; could be intentional (broader coverage) or noise (too loose)
- **Removed matches** — the regex narrowed; could be intentional (tighter pattern) or a regression (missing real secrets)

PRs that add a new detector will see a 🆕 row with an absolute match count and no baseline comparison.

## Example output

image

## What changed

- **`.github/workflows/detector-corpora-test.yml`** — new workflow; triggers on PRs touching `pkg/detectors/**`; detects which detectors changed, builds PR and main binaries in parallel, runs both scans against the corpus concurrently, posts a sticky comment with the diff
- **`scripts/test/detect_changed_detectors.sh`** — resolves changed detector directories to their proto enum names for `--include-detectors` scoping; skips detectors whose diff doesn't touch regex patterns or `Keywords()` so PRs that only change verification or redaction logic don't trigger a bench run
- **`scripts/test/detector_corpora_test.sh`** — streams corpus files (S3 or local), runs trufflehog with `--no-verification`, outputs JSONL
- **`scripts/test/diff_corpora_results.py`** — diffs two JSONL result sets and renders the Markdown report posted to the PR

## Performance

The corpus is ~2.8 GB compressed (~29 GB uncompressed). A typical run takes ~21 minutes.

Three things keep it manageable:

**Scoped scanning** — Only the detectors changed in the PR are passed via `--include-detectors`. Scanning the full detector set against 29 GB would be prohibitive; scoping to 1–3 detectors cuts runtime proportionally.

**Single S3 stream** — Both PR and main binaries consume the same S3 download via a named FIFO. S3 is downloaded and decompressed once; both scans run in parallel against the same stream.

**Main scan caching** — The main binary always scans the same merge-base commit for a given PR. Results are cached in GitHub Actions keyed by `merge-base SHA + detector set`. On a manual re-run without a rebase, the entire main side (worktree checkout, `go build`, S3 download, scan) is skipped.

## Fork PRs

Excluded for fork PRs — S3 credentials are not available to fork-originated workflows. Maintainers need to run this manually for forked PRs.

## Running locally

```bash
./scripts/test/detector_corpora_test.sh /path/to/contents.jsonl.zstd
```

Results land in `/tmp/corpora_results.jsonl` with a DuckDB summary table printed to stdout.

---

> [!NOTE]
> **Medium Risk**
> Adds a new GitHub Actions workflow that uses AWS credentials to stream large corpora from S3 and post sticky PR comments, so misconfiguration could leak secrets, waste CI resources, or spam PRs.
> 
> **Overview**
> Adds an automated *corpora regression bench* in CI for detector regex/keyword changes: a new `detector-corpora-test.yml` workflow scopes scans to only changed detectors, builds both PR and merge-base binaries, streams the same S3 corpus to both, diffs JSONL results, and posts/updates a sticky PR comment (or a skip comment when no relevant detector changes). 
> 
> Introduces helper tooling in `scripts/test/`: `detect_changed_detectors.sh` maps changed detector dirs to `--include-detectors` IDs, filters out non-regex/`Keywords()`-only edits, and distinguishes new detectors from `defaults.go` import diffs; `detector_corpora_test.sh` runs scoped `trufflehog` scans (optionally dual-binary via `tee`) and emits JSONL; `diff_corpora_results.py` generates the Markdown report with regression/warn thresholds and new-detector handling. Also updates `.gitignore` for Python cache/artifacts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8d2732cabb79c3fa6a5d8ef76f06de7dd6a234a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/test/detector_corpora_test.sh
```diff
@@ -0,0 +1,134 @@
...
+if [[ "$RUN_DUCKDB_SUMMARY" == "1" ]]; then
+    duckdb -c "
+CREATE TABLE t AS FROM read_json_auto('$OUTPUT_JSONL', ignore_errors=true);
+
+SELECT
+    t.DetectorName detector,
+    COUNT(*) total,
+    SUM(CASE WHEN Verified AND VerificationError IS NULL THEN 1 ELSE 0 END) verified,
+    SUM(CASE WHEN NOT Verified AND VerificationError IS NULL THEN 1 ELSE 0 END) unverified,
+    SUM(CASE WHEN VerificationError IS NOT NULL THEN 1 ELSE 0 END) \"unknown\"
```
bradlarsen: Running with `--no-verification` above makes these values not meaningful, right?

scripts/test/detector_corpora_test.sh
```diff
@@ -0,0 +1,134 @@
...
+                stdin >> "$OUTPUT_JSONL" 2>> "$STDERR_FILE"
+        rc=$?
+    fi
+    set -e
+    return $rc
+}
+
+for CORPORA_FILE in "$@"; do
+    if [[ "$CORPORA_FILE" == s3://* ]]; then
+        aws s3 cp "$CORPORA_FILE" - | scan /dev/stdin
```
bradlarsen, line 109: Note: though `stdin` is likely fine for this kind of testing, the newer `json-enumerator` input source in TruffleHog might make this a bit more straightforward. That input source expects NDJSON where each value looks like this:
```
{"data": "utf8 string to scan", "metadata": <arbitrary JSON value>}
```
or this:
```
{"data_b64": "base64-encoded bytestring to scan", "metadata": <arbitrary JSON value>}
```

scripts/test/detector_corpora_test.sh
```diff
@@ -0,0 +1,134 @@
...
+                --no-verification \
+                --allow-verification-overlap \
+                --log-level=3 \
+                --concurrency=8 \
+                --json \
+                --print-avg-detector-time \
+                "${INCLUDE_FLAG[@]}" \
+                stdin >> "$OUTPUT_JSONL" 2>> "$STDERR_FILE"
+        rc=$?
+    fi
```
bradlarsen, line 102: When I last used `trufflehog stdin` like this, I found that it would timeout after 1 minute, leaving you with an incomplete scan when dealing with large inputs. The workaround was, confusingly, to specify `--archive-timeout=6h` (or some similarly large value).

You might want to check that the scan is not being terminated early! This is another reason why you might prefer the `json-enumerator` input source, which doesn't have this timeout gotcha.

.github/workflows/detector-corpora-test.yml
```diff
@@ -0,0 +1,266 @@
...
+      - 'scripts/test/detector_corpora_test.sh'
+      - 'scripts/test/diff_corpora_results.py'
+      - 'scripts/test/detect_changed_detectors.sh'
+
+env:
+  DATASETS: |
+    s3://trufflehog-corpora-datasets/contents.2025-11-04.jsonl.zstd
+    s3://trufflehog-corpora-datasets/contents.jsonl.zstd
+
+jobs:
```
bradlarsen, line 19: Note, some of the actions used here are old versions. Also, you might consider pinning the action versions used here to reduce risk of possible supply-chain attacks.

zizmor is helpful: https://docs.zizmor.sh/

.github/workflows/detector-corpora-test.yml
```diff
@@ -0,0 +1,266 @@
...
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'pkg/detectors/**'
+      - 'pkg/engine/defaults/defaults.go'
+      - '.github/workflows/detector-corpora-test.yml'
+      - 'scripts/test/detector_corpora_test.sh'
+      - 'scripts/test/diff_corpora_results.py'
+      - 'scripts/test/detect_changed_detectors.sh'
```
bradlarsen, line 13: Since the CPU work required by a single run of this workflow is pretty expensive (30+ minutes), is this something we want running automatically on pull requests (as it looks like it does here), or only as an opt-in workflow?

bradlarsen (commented):
This looks useful!

scripts/test/detector_corpora_test.sh
mustansir14: Agreed! Thanks for catching this. Will remove

scripts/test/detector_corpora_test.sh
mustansir14: Hmm, `json-enumerator` would give better chunk isolation and carry path metadata through. But in this context we're running with `--no-verification` and the metadata isn't used downstream, so the practical difference is minimal. Keeping `stdin` for now to avoid any unexpected behavioral changes, but happy to revisit if there's a specific benefit you had in mind.

scripts/test/detector_corpora_test.sh
mustansir14: I don't think this is an issue in here. The tests consistently go 30+ minutes and successfully complete without truncation. I think it's because we decompress the corpus ourselves with `unzstd` before piping to `stdin`, so TruffleHog never sees a compressed archive and the archive timeout never applies.

.github/workflows/detector-corpora-test.yml
mustansir14: Thanks for this. Really helpful! I'll do the needful

.github/workflows/detector-corpora-test.yml
mustansir14: Good question. The workflow is scoped to only trigger when a PR modifies regex patterns or `Keywords()` in a detector. Purely structural changes (verification logic, redaction, comments, etc.) are filtered out and skip the bench entirely. In practice this means it only runs on PRs that actually affect match behavior.

We also don't currently merge a detector without running this test manually, so automating it in CI seems like the right call. it ensures the check never gets skipped and gives reviewers the data they need without having to ask for it.

scripts/test/detector_corpora_test.sh
bradlarsen: Makes sense! Just a suggestion.

scripts/test/detector_corpora_test.sh
bradlarsen: > The tests consistently go 30+ minutes and successfully complete without truncation.

How long do the actual scans run? If you look closely, do you see that the actual scans (not the builds etc) take just ~1 minute each?

I just was able to reproduce the behavior I was seeing earlier on my local machine, using the latest published trufflehog (3.95.2). When I try to scan a large stream of textual data on stdin, the scan stops after 60s (the default value of `--archive-timeout`). Adjust the value of `--archive-timeout` to something shorter and you will see:

```
time trufflehog --no-update --log-level=4 --no-verification -j --archive-timeout 5s stdin < SEVERAL_GIGABYTES.txt
...
{"level":"error","ts":"2026-05-07T10:09:40-04:00","logger":"trufflehog","msg":"error writing to data channel","source_manager_worker_id":"dwpxz","unit_kind":"unit","unit":"<stdin>","mime":"text/plain; charset=utf-8","timeout":5,"error":"context deadline exceeded"}
...
{"level":"info-0","ts":"2026-05-07T10:09:40-04:00","logger":"trufflehog","msg":"finished scanning","chunks":106929,"bytes":1423438848,"verified_secrets":0,"unverified_secrets":77,"scan_duration":"5.0252335s","trufflehog_version":"3.95.2","verification_caching":{"Hits":0,"Misses":0,"HitsWasted":0,"AttemptsSaved":0,"VerificationTimeSpentMS":0}}
trufflehog --no-update --log-level=4 --no-verification -j --archive-timeout 5  49.61s user 1.97s system 829% cpu 6.219 total
```

.github/workflows/detector-corpora-test.yml
mustansir14: I've improved on this by having it run only once when a PR opens. Subsequent runs can be done manually.

bradlarsen (approved):
This looks like it would add some useful test feedback for detector regex-related changes.

My only remaining question is whether the 60s timeouts I've observed with the `trufflehog stdin` source are affecting the testing done here.

mustansir14:
> My only remaining question is whether the 60s timeouts I've observed with the `trufflehog stdin` source are affecting the testing done here.

It turns out you were right. The scans were indeed timing out after 1 minute. The CI wasn't showing this because scanner logs were being piped into a temp file. Once I made it log on CI, I was able to see the error. I added the `--archive-timeout=2h` flag and now I see in the logs that scans are successfully completing.

Also this impacted the running time heavily. It was running for more than an hour now. This won't work in CI, so I removed the larger dataset (32GB one) and now it's only running on the smaller (2.8GB compressed, 29GB uncompressed) dataset. The running time for this is ~21 minutes, which is fine I think.

Even the smaller dataset returned 865 results for a loosened JDBC regex, so I think it would still really benefit us to have it on CI.

@bradlarsen Thanks much for this! Great catch!

scripts/test/detector_corpora_test.sh
mustansir14: You are right. This was indeed an issue. See my comment on the PR. Thanks for catching!

scripts/test/detect_changed_detectors.sh
```diff
@@ -0,0 +1,206 @@
...
+    local file
+    while IFS= read -r file; do
+        [[ "$file" == *_test.go ]] && continue
+        local head_body base_body
+        head_body=$(git show "$HEAD_REF:$file" 2>/dev/null \
+            | awk '/func[[:space:]].*Keywords\(\)[[:space:]]*\[\]string/,/^[[:space:]]*\}/' \
+            | tail -n +2)
+        base_body=$(git show "$MERGE_BASE:$file" 2>/dev/null \
+            | awk '/func[[:space:]].*Keywords\(\)[[:space:]]*\[\]string/,/^[[:space:]]*\}/' \
+            | tail -n +2)
```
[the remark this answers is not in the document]
mustansir14: This is not an issue because the `Keywords` method never contains conditional logic or any sort of block enclosed by braces.

scripts/test/detect_changed_detectors.sh
```diff
@@ -0,0 +1,206 @@
...
+
+# Step 2b — skip detectors whose diff doesn't touch regex patterns or Keywords.
+# Corpora results only change when the matching logic changes; verification,
+# redaction, or structural changes don't affect match counts.
+has_pattern_change() {
+    local dir="$1"
+
+    # Fast path: regex or Keywords() signature on a changed line.
+    git diff "$MERGE_BASE...$HEAD_REF" -- "$dir"/*.go 2>/dev/null \
+        | grep -qE '^[+-][^+-].*(regexp\.|MustCompile|Keywords)' && return 0
```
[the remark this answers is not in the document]
mustansir14: Test files never contain `regexp.MustCompile` or `Keywords`. Not an issue

camgunz (approved):

PR #4931, opened 2026-04-29
fix nested archive handler error wraps to preserve errors.Is identity

johnelliott:
## Summary

Two same-shape `%v` error wraps in nested archive handlers drop the inner
error's identity from the `errors.Is` chain, causing the
`isFatal` classifier in `pkg/handlers/handlers.go` to misclassify
cancellation and deadline-exceeded conditions as non-fatal warnings.
This is the identical bug fixed for the canonical chunk-reader wrap in
PR #4929.

Affected sites:

- `pkg/handlers/ar.go:109`
- `pkg/handlers/rpm.go:120`

Both wrap the return value of `handleNonArchiveContent`, which can be
`ctx.Err()` (`context.Canceled` / `context.DeadlineExceeded`) when its
internal `CancellableWrite` fails. The wrapped error feeds
`isFatal` at `pkg/handlers/handlers.go:425`. With `%v`, neither
`errors.Is(err, context.Canceled)` nor
`errors.Is(err, context.DeadlineExceeded)` matches, so the classifier
falls through to the `ErrProcessingWarning` case and processing
continues despite a fatal cause.

## Fix

Change the inner `%v` to `%w` in both wraps, matching the shape used by
PR #4929. Multi-`%w` in a single `fmt.Errorf` is supported on Go
1.20+; this repo is on Go 1.24.

## Behavior change

Cancelled or deadline-exceeded processing inside an AR or RPM nested
archive now correctly terminates instead of being logged as a warning
and continuing.

## Tests

Added a regression test for each handler:

- `TestARHandler_NonArchiveErrPreservesIdentity` in `ar_test.go`
- `TestRPMHandler_NonArchiveErrPreservesIdentity` in `rpm_test.go`

Each test injects a chunk reader that cancels the parent context before
delivering its chunk, forcing `handleNonArchiveContent`'s
`CancellableWrite` to return `context.Canceled`, which the surrounding
handler wraps. The tests assert the outer `ErrProcessingWarning` wrap is
preserved, the inner cancellation cause is inspectable via
`errors.Is`, and `isFatal` classifies the wrapped error as fatal.

Verified the tests catch the regression by reverting just the wrap
fix: both tests fail on `errors.Is(err, context.Canceled)` and
`isFatal(err)`, then pass again once the fix is reapplied.

## Test plan

- [x] `go test ./pkg/handlers/ -count=1` passes
- [x] `go vet ./pkg/handlers/` clean
- [x] Catches-the-bug verification: revert `%w` -> `%v` -> tests fail; reapply -> tests pass

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to error wrapping plus new tests; behavior only changes for cancellation/deadline paths in AR/RPM processing and is unlikely to impact normal success flows.
> 
> **Overview**
> Fixes AR and RPM nested archive handlers to wrap `handleNonArchiveContent` failures with `%w` (instead of `%v`) while still tagging them with `ErrProcessingWarning`, preserving `errors.Is` identity for underlying causes like `context.Canceled`/`context.DeadlineExceeded`.
> 
> Adds regression tests (`TestARHandler_NonArchiveErrPreservesIdentity`, `TestRPMHandler_NonArchiveErrPreservesIdentity`) that inject a cancelling chunk reader and assert both the outer warning wrap and the inner cancellation cause remain detectable and are classified as fatal by `isFatal`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b66333863d2bf71eef63b2b124188ec6cac91de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/handlers/ar_test.go
+++ b/pkg/handlers/ar_test.go
@@ -1,8 +1,7 @@
 package handlers
 
 import (
-	stdctx "context"
-	"errors"
+	"context"
 	"io"
 	"os"
 	"testing"
@@ -11,7 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	trContext "github.com/trufflesecurity/trufflehog/v3/pkg/context"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
 
@@ -20,7 +19,7 @@ func TestHandleARFile(t *testing.T) {
 	assert.Nil(t, err)
 	defer file.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := trContext.WithTimeout(trContext.Background(), 3*time.Second)
 	defer cancel()
 
 	rdr, err := newFileReader(ctx, file)
@@ -28,7 +27,7 @@ func TestHandleARFile(t *testing.T) {
 	defer rdr.Close()
 
 	handler := newARHandler()
-	dataOrErrChan := handler.HandleFile(context.AddLogger(ctx), rdr)
+	dataOrErrChan := handler.HandleFile(trContext.AddLogger(ctx), rdr)
 	assert.NoError(t, err)
 
 	wantChunkCount := 102
@@ -54,7 +53,7 @@ func TestARHandler_NonArchiveErrPreservesIdentity(t *testing.T) {
 	require.NoError(t, err)
 	defer file.Close()
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := trContext.WithCancel(trContext.Background())
 	defer cancel()
 
 	rdr, err := newFileReader(ctx, file)
@@ -66,7 +65,7 @@ func TestARHandler_NonArchiveErrPreservesIdentity(t *testing.T) {
 	// chunk's data sees the cancelled context and returns context.Canceled,
 	// which handleNonArchiveContent returns and processARFiles wraps. This
 	// is the only path that exercises the ar.go:107-110 wrap.
-	cancellingChunkReader := sources.ChunkReader(func(_ context.Context, _ io.Reader) <-chan sources.ChunkResult {
+	cancellingChunkReader := sources.ChunkReader(func(_ trContext.Context, _ io.Reader) <-chan sources.ChunkResult {
 		ch := make(chan sources.ChunkResult, 1)
 		cancel()
 		ch <- sources.NewChunkResult([]byte("data"), 4)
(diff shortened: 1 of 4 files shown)


johnelliott:
Question on the test assertions: any reason these use `assert.True(t, errors.Is(warnErr, target), "...")` rather than `assert.ErrorIs(t, warnErr, target, "...")`? 

The testify helper produces a better diagnostic on failure (it prints the actual error chain), and we already import `github.com/stretchr/testify/assert`. Same applies to the matching test in `rpm_test.go`. Happy to leave as-is if there's a reason; just wanted to ask.

pkg/handlers/ar_test.go
```diff
@@ -1,13 +1,18 @@
 package handlers
 
 import (
+	stdctx "context"
```
rosecodym: Our idiom is to leave the stdlib `context` package unaliased and to alias `github.com/trufflesecurity/trufflehog/v3/pkg/context` as `trContext`.

pkg/handlers/rpm_test.go
```diff
@@ -34,3 +39,55 @@ func TestHandleRPMFile(t *testing.T) {
...
+	var warnErr error
+	for _, d := range got {
+		if d.Err != nil && errors.Is(d.Err, ErrProcessingWarning) {
+			warnErr = d.Err
+			break
+		}
+	}
+	require.Error(t, warnErr, "expected wrapped warning from rpm.go non-archive error path")
+
+	assert.True(t, errors.Is(warnErr, ErrProcessingWarning),
```
rosecodym: Did you avoid `assert.ErrorIs` intentionally?

rosecodym (commented):
left some style notes

pkg/handlers/ar_test.go
johnelliott: Done in 4a0369e — dropped the stdctx alias on the stdlib import and aliased trufflehog's pkg/context as trContext throughout both test files. Thanks for the pointer to the package idiom.

pkg/handlers/rpm_test.go
johnelliott: Not intentional — switched to assert.ErrorIs in 4a0369e for both ar_test.go and rpm_test.go. Better diagnostic on failure, agreed.

pkg/handlers/ar_test.go
```diff
@@ -34,3 +38,63 @@ func TestHandleARFile(t *testing.T) {
 
 	assert.Equal(t, wantChunkCount, count)
 }
+
+// TestARHandler_NonArchiveErrPreservesIdentity is a regression test for the
+// %v wrap at ar.go:109. Before the fix, wrapping handleNonArchiveContent's
```
alafiand: suggest specifying portion of the file, e.g. `dataOrErrChan` case, so the line number doesn't accidentally become stale. An LLM might be smart enough to catch in the future.

johnelliott: Good call — replaced the line refs in both `ar_test.go` and `rpm_test.go` docstrings with the symbolic shape (`dataOrErrChan` ErrProcessingWarning send in `processARFiles`/`processRPMFiles`' `handleNonArchiveContent` error path). Pushed in 7b66333.

PR #4942, opened 2026-05-04
fix: report TimeCamp verifier errors

alephnull-sh:
## Summary

Small scoped follow-up to trufflesecurity/trufflehog#4051 for the TimeCamp detector.

Before this change, TimeCamp verification treated request failures and unexpected API responses the same as invalid credentials: the result came back unverified, with no `VerificationError`. This keeps `401` and `403` as normal invalid-token responses, but records a `VerificationError` when the verifier cannot make a clean decision.

The patch also adds a test-only client path so the verifier cases can be covered without hitting the live TimeCamp API.

## Test plan

- `go test ./pkg/detectors/timecamp`
- `git diff --check`

### DIFF
--- a/pkg/detectors/timecamp/timecamp.go
+++ b/pkg/detectors/timecamp/timecamp.go
@@ -2,22 +2,27 @@ package timecamp
 
 import (
 	"context"
-	regexp "github.com/wasilibs/go-re2"
+	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


shahzadhaider1 (approved):

MuneebUllahKhan222 (approved):
LGTM

PR #4944, opened 2026-05-04
Fix GrafanaServiceAccount detector: emit unverified results, fix status codes, deduplicate

asivaprasad09:
### Problem
The existing `GrafanaServiceAccount` detector had several issues:

1. **Silent drop** — if no `*.grafana.net` domain was found in the same chunk as the `glsa_` token, the detector emitted **no result at all**. Valid tokens stored separately from their domain (e.g. token in `.env`, domain in `config.yaml`) were completely invisible.
2. **`403` treated as error** — a `403 Forbidden` response was falling into the indeterminate error bucket instead of being treated as a determinately invalid token.
3. **Connection leaks** — `defer res.Body.Close()` without draining the body prevents HTTP connection reuse.
4. **Wrong HTTP client** — used `DetectorHttpClientWithNoLocalAddresses` instead of `common.SaneHttpClient()` which is the standard across all other detectors.
5. **No deduplication** — the same token could produce duplicate results if the regex matched multiple times in a chunk.

### Changes

- **Always emit the token** — if no domain is found in the chunk, an unverified result is emitted so the token is never silently lost. If a domain is found, verification is attempted as before.
- **Fix 403 handling** — `401` and `403` both correctly return `false, nil` (determinately invalid).
- **Proper body drain** — `io.Copy(io.Discard)` before `Close()` to avoid connection leaks.
- **Switch to `common.SaneHttpClient()`** — consistent with the rest of the codebase.
- **Deduplicate** — use `uniqueKeys` and `uniqueDomains` maps before processing.
- **Extract `verifyGrafanaServiceAccount()`** — verification logic moved to a named helper, consistent with detector conventions.

## Before vs After

| Scenario | Before | After |
|---|---|---|
| Token + domain in same chunk | ✅ Verified | ✅ Verified |
| Token alone, no domain | ❌ Silently dropped | ⚠️ Flagged unverified |
| Invalid token + domain | ✅ Unverified | ✅ Unverified |
| `403` response | ❌ Indeterminate error | ✅ Determinately invalid |
| Duplicate matches | ❌ Duplicate results | ✅ Deduplicated |

## Test plan
- [x] Token + domain in same file → `✅ Found verified result` with `--only-verified`
- [x] Token alone, no domain → unverified result emitted (not silently dropped)
- [x] Verified against real `glsa_` tokens: `200` = valid, `401` = invalid

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes secret detection/verification behavior (including emitting additional unverified findings) and adds a new detector plus a new `DetectorType` enum value, which can affect downstream integrations and result volume.
> 
> **Overview**
> **GrafanaServiceAccount detector fixes:** tokens are now deduplicated and **always emitted** even when no `*.grafana.net` domain is present in the same chunk (unverified instead of being dropped). Verification logic is extracted to `verifyGrafanaServiceAccount`, switches to `common.SaneHttpClient()`, correctly treats `401/403` as determinately invalid, and drains response bodies to improve connection reuse.
> 
> **New Together AI detector:** adds a `togetherai` detector for `tgp_v1_...` API keys with optional online verification against `GET /v1/models`, registers it in defaults, and introduces `DetectorType_TogetherAI` in `proto/detector_type.proto` and generated enum code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c1e6a014ff6e640ed2743041af50f155dcccac9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
+++ b/pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
@@ -3,11 +3,13 @@ package grafanaserviceaccount
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
 
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detector_typepb"
 )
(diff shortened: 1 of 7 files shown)


pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
```diff
@@ -21,65 +23,69 @@ type Scanner struct {
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	defaultClient = detectors.DetectorHttpClientWithNoLocalAddresses
-	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat    = regexp.MustCompile(`\b(glsa_[0-9a-zA-Z_]{41})\b`)
+	defaultClient = common.SaneHttpClient()
```
mustansir14, line 26: Any particular reason why you've made this switch? We prefer to use `detectors.DetectorHttpClientWithNoLocalAddresses` with detectors that verify the credential against a custom domain.

pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
```diff
@@ -21,65 +23,69 @@ type Scanner struct {
...
+
+	// Grafana service account tokens have a 41-character body after glsa_.
+	// Older tokens can contain underscores in that body, so keep accepting them
+	// instead of requiring only the newer checksum-style separator.
+	keyPat    = regexp.MustCompile(`\b(glsa_[0-9A-Za-z_]{41})\b`)
 	domainPat = regexp.MustCompile(`\b([a-zA-Z0-9-]+\.grafana\.net)\b`)
 )
 
-// Keywords are used for efficiently pre-filtering chunks.
-// Use identifiers in the secret preferably, or the provider name.
```
mustansir14, line 31: nit: can we get this comment back please?

pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
```diff
@@ -21,65 +23,69 @@ type Scanner struct {
...
+		if len(uniqueDomains) == 0 {
+			// No domain found in this chunk — emit unverified so the token is not silently dropped.
+			s1 := detectors.Result{
+				DetectorType: detector_typepb.DetectorType_GrafanaServiceAccount,
+				Raw:          []byte(key),
+				SecretParts:  map[string]string{"key": key},
+			}
+			results = append(results, s1)
+			continue
+		}
```
mustansir14, line 72: While I agree with the concept of emitting credentials that couldn't be verified because of absence of a domain, I believe we need to make the distinction clear between this and a token that was marked determinately unverified using a domain.

`Result.ExtraData` is a good place to do this. You can add a `message` field inside that indicates this.  The [ConfluenceDataCenter detector](https://github.com/trufflesecurity/trufflehog/blob/0c381f12b3f9a934f33fc61bf003599f5323ff55/pkg/detectors/atlassiandatacenter/confluencedatacenter/confluencedatacenter.go#L133) also does this.

pkg/detectors/grafanaserviceaccount/grafanaserviceaccount.go
```diff
@@ -89,6 +95,33 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return true, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
```
mustansir14, line 118: Can you provide references to official documentation or examples where it says that a 403 indicates that the key is determinately invalid?

pkg/detectors/togetherai/togetherai.go
mustansir14, line 1: Please add this new detector in a separate PR

mustansir14 (changes requested):
Hey, thanks for this!

I've requested changes mainly for the new detector you added. We would like you to do that in a separate PR, to keep the scope of this PR intact.

I have some questions/suggestions for the `grafanaserviceaccount` detector changes. Please check them out.

pkg/detectors/togetherai/togetherai.go
mustansir14: I just saw #4943. It looks like you branched off from this feature branch instead of main for this PR. Can you please update the base branch?

PR #4949, opened 2026-05-04
fix: stream CircleCI action output

alephnull-sh:
### Description:
CircleCI action output is build-log data, but this path was reading the whole response body before passing it to the chunker. Large step output can make the source hold much more data in memory than it needs to.

This keeps the response body as a stream and passes it straight into the existing chunk reader. The CircleCI config string still goes through the same chunk path with a `strings.Reader`.

The output fetch now also returns an error for 4xx/5xx responses, draining and closing the response body before returning.

This follows the same idea as the Jenkins build-log streaming fix in #4225, but keeps the change scoped to the CircleCI source.

### Tests:
* [x] `GOCACHE=/Users/nsatyakrishna/Sandbox/.cache/go-build GOMODCACHE=/Users/nsatyakrishna/Sandbox/.cache/go-mod go test ./pkg/sources/circleci -run 'Test(GetOutputUrlResponseStreamsBody|ChunkReadsFromReader)'`
* [x] `make lint`

I ran the source-specific tests here because `make test-community` skips `pkg/sources`, so it does not cover this change.

### Checklist:
* [x] Targeted tests passing
* [x] Lint passing

### DIFF
--- a/pkg/sources/circleci/circleci.go
+++ b/pkg/sources/circleci/circleci.go
@@ -152,7 +152,7 @@ func (s *Source) Chunks(ctx context.Context, chunksChan chan *sources.Chunk, _ .
 
 				for _, step := range projBuildJobs.Steps {
 					for _, action := range step.Actions {
-						data, err := s.getOutputUrlResponse(action.OutputUrl)
+						data, err := s.getOutputUrlResponse(ctx, action.OutputUrl)
 						if err != nil {
 							ctx.Logger().Error(err, "error getting action output url response")
 							errors = append(errors, err)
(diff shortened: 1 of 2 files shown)


pkg/sources/circleci/circleci.go
```diff
@@ -328,34 +331,31 @@ func (s *Source) listProjBuildJobs(ctx context.Context, proj *project, buildNum
...
+
+	if resp.StatusCode >= http.StatusBadRequest {
 		_, _ = io.Copy(io.Discard, resp.Body)
 		_ = resp.Body.Close()
-	}()
-
-	bodyBytes, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return "", fmt.Errorf("failed to read response body: %w", err)
+		return nil, fmt.Errorf("unexpected status code: %d while fetching action output", resp.StatusCode)
```
shahzadhaider1, line 348: Maybe we should add this new behavior in the PR description.

shahzadhaider1 (approved):

PR #4951, opened 2026-05-07
[INS-377] Add V2 for weights and biases detector

mustansir14:
## Summary

Weights & Biases recently introduced a new API key format prefixed with `wandb_v1_`, distinct from the legacy 40-character hex format. This PR adds a v2 detector to cover the new format while keeping the v1 detector active for existing keys.

## What changed

- **`pkg/detectors/weightsandbiases/base.go`** — defines `BaseScanner`, a base struct embedded by both versioned scanners; provides `FromData`, `verifyMatch`, `Type`, and `Description`; the HTTP client is held on the struct so no need to pass it through call chains; version is passed as a parameter to `FromData` and included in `ExtraData` of every result alongside the account fields returned by the API
- **`pkg/detectors/weightsandbiases/v1/`** — existing detector moved here from the package root; implements `Versioner` returning `1`; delegates to `BaseScanner` for all shared logic
- **`pkg/detectors/weightsandbiases/v2/`** — new detector targeting the `wandb_v1_<27 chars>_<49 chars>` format; implements `Versioner` returning `2`; delegates to `BaseScanner` for all shared logic
- **`pkg/engine/defaults/defaults.go`** — updated import paths; both `v1` and `v2` scanners registered

## Regex

| Version | Pattern |
|---------|---------|
| v1 | `[0-9a-f]{40}` (with `wandb` prefix) |
| v2 | `wandb_v1_[A-Za-z0-9]{27}_[A-Za-z0-9]{49}` |

The v2 pattern is self-anchored on the `wandb_v1_` literal prefix, so no `PrefixRegex` wrapper is needed. Segment lengths are exact based on observed key samples.

## Verification

Both versions verify against the W&B GraphQL API (`POST https://api.wandb.ai/graphql`) using the `viewer` query with Basic auth (`api:<token>`). A 200 response with a non-empty `username` field is considered verified; 401 is definitively invalid; anything else is a verification error.

## Corpora Testing

The detector does not appear in the lists.
image
image

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped detector refactor with shared verification; v2 rollout is feature-flagged and v1 behavior is preserved aside from version metadata in results.
> 
> **Overview**
> Adds support for Weights & Biases’ new `wandb_v1_…` API key format while keeping detection for legacy 40-character hex keys.
> 
> The monolithic detector is split into a shared **`BaseScanner`** (regex-driven `FromData`, GraphQL verification, and **`ExtraData["version"]`**) plus **`v1`** and **`v2`** scanners that implement **`Versioner`**. **v1** keeps the existing `wandb` + hex pattern; **v2** matches `wandb_v1_[A-Za-z0-9]{27}_[A-Za-z0-9]{49}` with keyword `wandb_v1_`. Both scanners are registered in default detectors; **v2** is gated by **`WeightsAndBiasesV2DetectorEnabled`** (on by default in **`main.go`**). Integration and pattern tests were updated for v1 and added for v2.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0e9ba89d51a7bb5a9f1368f732edd404616ecaf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/weightsandbiases/common.go
shahzadhaider1, line 120: I don't see the detector version being set in extraData anywhere in this PR -- can we add it?

shahzadhaider1 (commented):
I like how you’ve unified the common code into a single implementation--it looks clean. That said, it’s a bit inconsistent with how we handle newer versions in other detectors across the codebase. Just something I noticed.

amanfcp (approved):

pkg/detectors/weightsandbiases/common.go
amanfcp: Agreed. We should add version info in ExtraData to follow convention

mustansir14: Added

mustansir14:
> I like how you’ve unified the common code into a single implementation--it looks clean. That said, it’s a bit inconsistent with how we handle newer versions in other detectors across the codebase. Just something I noticed.

Yeah, this is done because the implementations for both versions are exactly the same. The only difference is the regex pattern, so It didn't make sense for me to duplicate all of the code.

Other versioned detectors usually have some differences, mostly in the verification logic.

amanfcp (approved):

MuneebUllahKhan222 (approved):
LGTM, need to resolve merge conflicts

trufflesteeeve (approved):

PR #4917, opened 2026-04-23
Add Pinecone API key detector

dylanTruffle:
Adds a new detector for Pinecone vector database API keys (pcsk_* format).

Detection:
- Regex matches the pcsk_{label}_{secret} structure with tight bounds (5–6 char label, exactly 63 char secret, word boundaries), verified against real keys. This minimizes false positives and prevents over-reading when a key is adjacent to other alphanumeric text.
- The label is captured as a regex group and surfaced as ExtraData["key_id"] unconditionally, which maps to the key entry in the Pinecone console and aids revocation.
- SecretParts["key"] is populated on every result, satisfying the framework's checksecretparts static check.

Verification:
- Uses GET https://api.pinecone.io/indexes (non-state-changing, read-only) with Api-Key header auth and X-Pinecone-Api-Version: 2025-10.
- Validates response body structure (requires the top-level "indexes" JSON key), not just status codes, to be resilient against generic 200 responses from intermediate proxies.
- 200: Verified. Extracts project_id from the first index host, total_indexes, and metadata for up to 5 indexes (name, host, cloud, region).
- 401: Invalid key. Returned unverified with no VerificationError (documented behavior).
- 403: Valid key with restricted permissions (e.g. DataPlane-only roles). Marked as verified with permission=restricted metadata. Not documented in Pinecone's official API reference but observed in practice, since API keys can be created with custom permission scopes.
- 500 / other: Indeterminate. Returned unverified with a VerificationError so callers can retry.

Uses common.SaneHttpClient() with standard timeouts and io.LimitReader (1 MiB cap) on the response body. The verification endpoint is hardcoded (api.pinecone.io/indexes), so there is no user-controlled destination and the hardened no-local-addresses client is not needed. No SDK dependencies.

Tests:
- pinecone_test.go: regex coverage (valid/invalid shapes including too-short label, trailing alphanumerics, multi-key strings) plus table-driven tests for extractProjectID and mock-transport tests for verifyMatch (success, malformed indexes key, invalid indexes payload, metadata preservation on verification error).
- pinecone_integration_test.go (//go:build detectors): live-credential coverage for verified, unverified, not-found, timeout, and unexpected-API-surface cases.

Registered as DetectorType Pinecone = 1048.

### Checklist:
* [x] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

image

---------------------------------------------------------------------------------------------------

image

---

[...]

PR #4953, opened 2026-05-09
feat(detectors): add Mistral AI API key detector

ipunithgowda:
## Summary
Adds a new detector for Mistral AI API keys and verifies them against `api.mistral.ai/v1/models`. Mirrors the structure of the existing Groq and Anthropic detectors.

## Detection
Mistral does not document a key prefix. GitGuardian classifies the detector as non-prefixed, and Mistral console keys empirically present as 32-character mixed-case alphanumeric strings.

Because there is no distinctive prefix, the regex uses `detectors.PrefixRegex` to require the keyword `mistral` within 40 characters of the candidate token, consistent with how 750+ other trufflehog detectors handle low-specificity formats (`hunter`, `elevenlabs/v1`, etc.).

Effective pattern: `(?i:mistral)(?:.|[\n\r]){0,40}?\b([A-Za-z0-9]{32})\b`

## Verification
- `GET https://api.mistral.ai/v1/models` with `Authorization: Bearer <key>` (read-only, idempotent).
- 200 → verified
- 401 → determinately unverified
- Other → verification error (matches existing detector convention)

## Registration
- New `DetectorType` enum value: `Mistral = 1050` (next free after `Pinecone = 1049` from #4917).
- Updated `proto/detector_type.proto`.
- Updated `pkg/pb/detector_typepb/detector_type.pb.go`: const block, name map, value map, raw descriptor bytes.
- Registered `&mistral.Scanner{}` in `pkg/engine/defaults/defaults.go` between `miro` and `mite` to maintain alphabetical order.

### Note on the proto descriptor edit
The `pkg/pb/detector_typepb/detector_type.pb.go` raw-descriptor bytes were edited by hand because I could not run `make protos` locally (Docker daemon not available in my environment). The inserted block follows the exact same encoding pattern used by the Pinecone PR (#4917) — 14 added bytes: `0x12 0x0c 0x0a 0x07 "Mistral" 0x10 0x9a 0x08`, which decodes as inner-length 12, name "Mistral", varint 0x9a 0x08 = 1050. The outer file-descriptor length field was incremented from 17452 to 17466 to match the addition. Runtime reflection round-trips name↔number both directions cleanly and `go test ./pkg/pb/detector_typepb/...` passes.

If maintainers prefer, running `make protos` on merge should produce a byte-for-byte identical (or trivially equivalent) result with the toolchain pinned in `trufflesecurity/protos:1.22`.

## Test plan
- [x] `go build ./...` — clean
- [x] `go test ./pkg/detectors/mistral/...` — 5/5 PASS (typical_pattern, env_var_style, no_keyword_nearby, wrong_length_31, wrong_length_33)
- [x] `go test ./pkg/engine/defaults/...` — PASS
- [x] `go test ./pkg/pb/detector_typepb/...` — PASS (no test files; descriptor `init()` did not panic, which means the raw bytes are valid)
- [x] `go vet ./pkg/detectors/mistral/... ./pkg/engine/defaults/... ./pkg/pb/detector_typepb/...` — clean
- [ ] Maintainers may want to add `MISTRAL` and `MISTRAL_INACTIVE` GCP secrets so the integration test (build tag `detectors`) can run in CI. Same convention as Groq / Anthropic.

## Notes
- Pattern confidence is medium by design: Mistral does not publish key format. The verifier (200 vs 401 from `api.mistral.ai`) is the actual ground truth, which is the philosophy of every other trufflehog AI-provider detector. If you have a tighter regex from internal samples, the regex line is a one-liner change.
- Unit tests use synthetic 32-char alphanumeric strings only; no real keys present in the diff.

---

> [!NOTE]
> **Medium Risk**
> Adds a new secret detector with live HTTP verification against an external API and updates protobuf-generated detector type enums, which can impact runtime classification if the enum/descriptor changes are incorrect.
> 
> **Overview**
> Adds a new `mistral` detector that matches 32-char alphanumeric tokens near the keyword "mistral" and can verify candidates by calling `GET https://api.mistral.ai/v1/models` with a Bearer token, reporting 200 as verified and 401 as unverified.
> 
> Registers the detector in the default detector list and introduces a new `DetectorType` enum value (`MistralAI = 1050`) by updating `proto/detector_type.proto` and the generated `detector_type.pb.go`. Unit, integration (tagged `detectors`), and benchmark tests are added for the new detector.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afc9cd443665e9d9c7a39259f81d60ff0af6b69a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/mistral/mistral.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if verify {
+			client := s.client
+			if client == nil {
+				client = defaultClient
+			}
+
+			isVerified, extraData, verificationErr := verifyMatch(ctx, client, match)
+			s1.Verified = isVerified
+			s1.ExtraData = extraData
```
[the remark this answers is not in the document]
ipunithgowda: Fixed in a5eaaa1e466057be59a4f5dd6fbd6308cb7fa2f3. Switched to a merge loop so ExtraData["rotation_guide"] survives verification, and any future keys returned by verifyMatch are appended rather than replacing the constructor-set values. Tests still green.

proto/detector_type.proto
```diff
@@ -1051,4 +1051,5 @@ enum DetectorType {
   ConfluenceDataCenter = 1047;
   Cloudinary = 1048;
   Pinecone = 1049;
+  Mistral = 1050;
```
shahzadhaider1: What do you think about naming this detector `MistralAI`?

pkg/detectors/mistral/mistral.go
```diff
@@ -0,0 +1,118 @@
...
+
+			isVerified, extraData, verificationErr := verifyMatch(ctx, client, match)
+			s1.Verified = isVerified
+			// Merge verifier-supplied extra data into the existing map so the
+			// rotation_guide URL set above is preserved (verifyMatch currently
+			// returns nil for the map, but the merge is correct for future
+			// verifier additions and for the nil case alike).
+			for k, v := range extraData {
+				s1.ExtraData[k] = v
+			}
```
shahzadhaider1: https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_mapsloop

pkg/detectors/mistral/mistral.go
```diff
@@ -0,0 +1,118 @@
...
+	switch res.StatusCode {
+	case http.StatusOK:
+		return true, nil, nil
+	case http.StatusUnauthorized:
+		// The secret is determinately not verified (nothing to do)
+		return false, nil, nil
+	default:
+		return false, nil, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
+}
```
shahzadhaider1: This function never returns `extraData`, we should remove it from the return statements.

pkg/detectors/mistral/mistral.go
ipunithgowda: Good catch. Since verifyMatch was always returning a nil map (your other comment), I went one step further and dropped the merge entirely rather than swap the for-loop for maps.Copy. The ExtraData["rotation_guide"] set at construction is now preserved naturally because nothing in the verify branch touches the map. Fixed in 202a40d4a.

pkg/detectors/mistral/mistral.go
ipunithgowda: Agreed, removed. verifyMatch is now (bool, error). Fixed in 202a40d4a.

proto/detector_type.proto
ipunithgowda: Good call. Renamed to MistralAI in afc9cd443665e9d9c7a39259f81d60ff0af6b69a. Matches the OpenAI precedent where the AI suffix is part of the actual brand (vs Anthropic/Groq/Cohere which drop the suffix). Re-hand-edited the proto descriptor bytes for the 7->9 char name expansion: inner enum value entry 14->16 bytes, outer enum-type length varint 17466->17468. Runtime roundtrip verified both directions.

gugacyber:
Nice work, the verifyMatch simplification and the ExtraData preservation fix are clean.
One thing worth flagging: MistralAI = 1050 conflicts with DetectorType_GitLabOauth2 = 1050 which landed in main recently (PR #4961 range). Worth rebasing against current main and picking the next free value.
Also, the integration test cmpopts.IgnoreFields list only ignores "Raw" and "verificationError" but the detector sets ExtraData (with rotation_guide) and SecretParts on results. The want structs leave both as nil, so the integration test will mismatch when run against real credentials. The Anthropic detector includes "SecretParts" and "ExtraData" in its ignore list for exactly this reason.

PR #4954, opened 2026-05-11
fix(twilio): deduplicate matches to prevent O(N×M) result explosion

kashifkhan0771:
### Description:
Both Twilio detectors (twilio, twilioapikey) were emitting O(N×M) duplicate results when scanning. A document with the same SID appearing N times and the same API key M times produced N×M findings with identical secret values.
This fix deduplicates regex matches before the nested loop so only unique (sid, key) pairs are processed. The detectors also adopt `NewClientWithDedup/DoWithDedup` to combine concurrent verification requests for the same credential pair into a single network call. Two engine-level issues (dedup condition logic and LRU cache size) are noted in engine.go as known issues for a follow-up pass.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes Twilio/TwilioApiKey detectors’ matching/verification flow to reduce duplicate findings and network calls, which could alter result counts and verification behavior in edge cases. Engine changes are comments only but highlight an existing dedup/cache limitation that remains unresolved.
> 
> **Overview**
> Fixes Twilio secret scanning to avoid an *O(N×M)* explosion of duplicate findings by deduplicating regex matches before generating `(sid,key)` / `(apiKey,secret)` pairs.
> 
> Twilio verifications now use `detectors.NewClientWithDedup` plus `detectors.DoWithDedup` to coalesce concurrent verification requests for the same credential pair, and simplifies merging verification `ExtraData` via `maps.Copy`.
> 
> Adds *known-issue* notes in `engine.go` calling out the small LRU dedupe cache size and a dedupe condition that still allows duplicates for the same decoder type (no functional engine change in this PR).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 186678b1f2d16d940c969b18f34ae3997f70c503. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/twilioapikey/twilioapikey.go
+++ b/pkg/detectors/twilioapikey/twilioapikey.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 
 	regexp "github.com/wasilibs/go-re2"
@@ -23,7 +24,7 @@ type Scanner struct {
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	defaultClient = common.SaneHttpClient()
+	defaultClient = detectors.NewClientWithDedup(common.SaneHttpClient())
 	apiKeyPat     = regexp.MustCompile(`\bSK[a-zA-Z0-9]{32}\b`)
 	secretPat     = regexp.MustCompile(`\b[0-9a-zA-Z]{32}\b`)
 )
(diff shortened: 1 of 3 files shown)


pkg/detectors/twilioapikey/twilioapikey.go
```diff
@@ -103,7 +108,7 @@ func verifyTwilioAPIKey(ctx context.Context, client *http.Client, apiKey, secret
 	req.Header.Add("Accept", "*/*")
 	req.SetBasicAuth(apiKey, secret)
 
-	resp, err := client.Do(req)
+	resp, err := detectors.DoWithDedup(client, detector_typepb.DetectorType_TwilioApiKey, apiKey+secret, req)
```
shahzadhaider1, line 111: This detector falsely sets the detector type of the results as `DetectorType_Twilio` but you are passing `DetectorType_TwilioApiKey` as an argument here. Will that create any issues?

kashifkhan0771: Nope, and that’s actually a good thing. The secrets detected by both detectors can overlap, and we accidentally assigned the same detector type to both. Preserving the actual detector type for the Dedup client won’t change the result - it just ensures that identical secrets coming from both detectors are treated separately instead of overlapping.

shahzadhaider1 (approved):

pkg/detectors/twilioapikey/twilioapikey.go
```diff
@@ -54,11 +55,17 @@ func (s Scanner) Keywords() []string {
...
+	for _, s := range secretPat.FindAllString(dataStr, -1) {
+		uniqueSecrets[s] = struct{}{}
+	}
 
-	for _, apiKey := range apiKeyMatches {
-		for _, secret := range secretMatches {
+	for apiKey := range uniqueAPIKeys {
+		for secret := range uniqueSecrets {
 			s1 := detectors.Result{
 				DetectorType: detector_typepb.DetectorType_Twilio,
```
MuneebUllahKhan222, line 70: Since we're already making changes here, we can probably update the `DetectorType` as well.

MuneebUllahKhan222 (approved):
LGTM, Just one non blocking comment.

pkg/detectors/twilioapikey/twilioapikey.go
kashifkhan0771: Actually, that’s the tricky part. If we update `DetectorType` now, all previously discovered secrets associated with this detector will become orphaned.

camgunz (approved):

PR #4959, opened 2026-05-12
fix(detectors/azureappconfigconnectionstring): allow hyphen and colon in Id

### LINKED ISSUE #4955: Azure App Configuration not detected
### TruffleHog Version

3.95.3

### Trace Output

N/A — issue is reproducible from the detector regex alone; no scan trace needed.

### Expected Behavior

The `AzureAppConfigConnectionString` detector should match Azure App Configuration access-key connection strings in the canonical format produced by Azure:

```
Endpoint=https://<store>.azconfig.io;Id=<prefix>:<base64>;Secret=<base64>
```

The `Id` portion is documented as `<prefix>:<base64>`, where `<prefix>` typically contains letters, digits, and hyphens. The detector should flag this as a finding.

### Actual Behavior

The detector silently misses real Azure App Configuration access keys because the `Id` character class in the regex only allows the base64 alphabet (`[a-zA-Z0-9+\/=]`). Any real-world `Id` value containing `-` or `:` (which is every Azure-issued key, since the colon separates prefix from base64 body) fails to match, and no finding is emitted.

Current pattern in `pkg/detectors/azureappconfigconnectionstring/azureappconfigconnectionstring.go`:

```go
connectionStringPat = regexp.MustCompile(`Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/=]+);Secret=([a-zA-Z0-9+\/=]+)`)
```

Proposed fix — extend the `Id` character class to include `:` and `-`:

```go
connectionStringPat = regexp.MustCompile(`Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/:=-]+);Secret=([a-zA-Z0-9+\/=]+)`)
```

### Steps to Reproduce

1. Save the following representative (fake) connection string to a file, e.g. `/tmp/azconfig.txt`:

   ```
   Endpoint=https://myappconfig001.azconfig.io;Id=aB1c-d2-e3:+XyZ12345AbCdEfGhIjKl;Secret=MnOpQrSt67UvWxYz89AbCdEf0123GhIjKlMnOpQrStUvWxYz1234=
   ```

   The `Id` (`aB1c-d2-e3:+XyZ12345AbCdEfGhIjKl`) matches the canonical Azure-issued format: hyphenated prefix, `:`, base64 body.

2. Run trufflehog against the file with verification disabled to isolate the regex behavior:

   ```
   trufflehog filesystem /tmp/azconfig.txt --no-update --json --no-verification
   ```

3. Observe `unverified_secrets: 0` — the detector emits no finding even though the string is a syntactically valid Azure App Configuration connection string.

4. Apply the regex change above and rebuild; the same input now produces an `AzureAppConfigConnectionString` finding as expected.

## Environment

* Version: trufflehog 3.95.3

## Additional Context

The fix is a minimal character-class extension on a single regex; the verification logic downstream already handles arbitrary `Id`/`Secret` byte content correctly, so no other changes are required. Happy to open a PR with the fix and a detector unit test covering the hyphen/colon `Id` shape if useful.

### References

* `pkg/detectors/azureappconfigconnectionstring/azureappconfigconnectionstring.go` — current detector

ChrisJr404:
## Summary

The `AzureAppConfigConnectionString` detector misses real Azure-issued connection strings because the `Id` capture group only allows the base64 alphabet. Azure's documented connection-string format is

```
Endpoint=https://<store>.azconfig.io;Id=<prefix>:<base64>;Secret=<base64>
```

where `<prefix>` is a short identifier that can contain letters, digits, and hyphens, joined to the base64 portion with a colon. See Microsoft's docs on the connection-string shape and on rotating access keys, both of which show the `prefix:base64` form for `Id`:

- https://learn.microsoft.com/en-us/azure/azure-app-configuration/howto-create-snapshots#connection-strings
- https://learn.microsoft.com/en-us/azure/azure-app-configuration/howto-best-practices

The current regex is

```go
Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/=]+);Secret=([a-zA-Z0-9+\/=]+)
```

The colon separator alone is enough to break the match, and any hyphen inside the prefix breaks it too. Both happen on every store created in the portal or via `az appconfig credential list`.

## Fix

Add `:` and `-` to the `Id` character class. The `Endpoint` host class already permits hyphens, the `Secret` is unchanged, and no other detector calls were affected.

```diff
-Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/=]+);Secret=([a-zA-Z0-9+\/=]+)
+Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/=:-]+);Secret=([a-zA-Z0-9+\/=]+)
```

## Test plan

- Added a new subtest `valid pattern - id with prefix and hyphen` to `azureappconfigconnectionstring_test.go` that feeds a real-shape `Id=abcd-l0:u+DeEXamPleIdValue==` connection string.
- `go test ./pkg/detectors/azureappconfigconnectionstring/...` passes all four subtests.
- Manually verified that the new subtest fails before the regex change and passes after it, so the test exercises the bug.

Closes #4955

---

> [!NOTE]
> **Low Risk**
> Low risk: a small regex broadening plus a new unit test; behavior change is limited to matching more valid Azure App Config connection strings.
> 
> **Overview**
> Improves `AzureAppConfigConnectionString` detection by widening the `Id` capture group in `connectionStringPat` to accept `-` and `:` (matching Azure’s `prefix:base64` `Id` format).
> 
> Adds a new pattern test case covering an `Id` with a hyphenated prefix and colon separator to prevent regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7793ac09d17241e6bade18b878a10a91bdd7f7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/azureappconfigconnectionstring/azureappconfigconnectionstring.go
+++ b/pkg/detectors/azureappconfigconnectionstring/azureappconfigconnectionstring.go
@@ -27,7 +27,7 @@ var _ detectors.Detector = (*Scanner)(nil)
 
 var (
 	defaultClient       = common.SaneHttpClient()
-	connectionStringPat = regexp.MustCompile(`Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/=]+);Secret=([a-zA-Z0-9+\/=]+)`)
+	connectionStringPat = regexp.MustCompile(`Endpoint=(https:\/\/[a-zA-Z0-9-]+\.azconfig\.io);Id=([a-zA-Z0-9+\/=:-]+);Secret=([a-zA-Z0-9+\/=]+)`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
--- a/pkg/detectors/azureappconfigconnectionstring/azureappconfigconnectionstring_test.go
+++ b/pkg/detectors/azureappconfigconnectionstring/azureappconfigconnectionstring_test.go
@@ -40,6 +40,11 @@ func TestAzureAppConfigConnectionString_Pattern(t *testing.T) {
 			want: []string{"Endpoint=https://iTHzRfnepCddRiYoBbPj-drVzUjwTNduwb3EUOTsuSAgg1e83Q7bw.azconfig.io;Id=eO04L+/m9rYn;Secret=G4jQ3GmcsYqlLkkG8uoIVbx08PZIJSdfB/7"},
 		},
 		{
+			name:  "valid pattern - id with prefix and hyphen",
+			input: `Endpoint=https://trufflesecurity.azconfig.io;Id=abcd-l0:u+DeEXamPleIdValue==;Secret=80DtxZkndXpM2mV2J1JjX2vL1x4gm1hHn8Y3JeFJ4N0PPLSO5D70JQQJ99BBAC1i4FpQkb5wAAACAAZC26dr`,
+			want:  []string{"Endpoint=https://trufflesecurity.azconfig.io;Id=abcd-l0:u+DeEXamPleIdValue==;Secret=80DtxZkndXpM2mV2J1JjX2vL1x4gm1hHn8Y3JeFJ4N0PPLSO5D70JQQJ99BBAC1i4FpQkb5wAAACAAZC26dr"},
+		},
+		{
 			name:  "invalid pattern",
 			input: `Endpoint=https://trufflesecurity.azconfig.io;Secret=80DtxZkndXpMTmV2J3JjX2vL1x4gm1hHn8Y3KeFV4N0PPLSO5D70JQQJ79BBAC1i4FpRkb5wAAACAAZC26dr`,
 			want:  nil,


gugacyber:
Fix looks correct, the Azure docs linked in the description confirm the prefix:base64 format for Id. Good call adding : and - to the character class rather than making the regex too permissive.
One thing worth considering: the Id group now accepts :- anywhere in the string, not just as a separator between prefix and base64. A tighter pattern like [a-zA-Z0-9-]+:[a-zA-Z0-9+\/=]+ would reflect the actual documented format more precisely and avoid matching malformed strings. Not a blocker since false positives here are low-risk, but worth noting.

ChrisJr404:
Good eye. I went with the character-class widening on purpose rather than a `prefix:base64` shape — not every Id in the wild carries the colon separator (the existing "valid pattern" test case, `Id=eO04L+/m9rYn`, has none), so requiring `[...]+:[...]+` would turn those into false negatives. Since the Secret group is still strict, the blast radius of a slightly loose Id is low. Happy to make the prefix+colon optional instead if you'd rather encode both shapes explicitly.

Issue #4962, opened 2026-05-13
`# trufflehog:ignore` not honored for Postgres URLs without an explicit port


mrstanwell:
### TruffleHog Version
- 3.95.3

### Trace Output
https://gist.github.com/mrstanwell/dbd00c7ae2996dedcf8a83aca6d06b16

### Expected Behavior
`# trufflehog:ignore` on a line containing a Postgres URL suppresses the finding regardless of whether the URL includes an explicit port number.

### Actual Behavior
When a Postgres URL omits the port (e.g. `postgresql://user:pass@host/db`), `# trufflehog:ignore` does not suppress the finding. Adding an explicit `:5432` to the same URL makes the annotation work correctly.

### Steps to Reproduce
```bash
# Case 1: no explicit port — trufflehog:ignore is NOT honored (exit 183)
printf 'DB=postgresql://user:secret@host/db  # trufflehog:ignore\n' > /tmp/config.env
trufflehog filesystem /tmp/config.env --fail
echo "exit: $?"

# Case 2: explicit port — trufflehog:ignore IS honored (exit 0)
printf 'DB=postgresql://user:secret@host:5432/db  # trufflehog:ignore\n' > /tmp/config.env
trufflehog filesystem /tmp/config.env --fail
echo "exit: $?"
```

### Workaround
Add a port number to the postgres URL and the `# trufflehog:ignore` directive will be honored.

## Environment
- OS: macOS 15.7.3

## Additional Context
I notice that the output shows `Raw result: postgresql://user:secret@host:5432`, even when there was no actual port number in the URL in the file.  That is not, in fact, a "raw" result; and I suspect it is causing a [`!found` result in `FragmentLineOffset`, thus pre-empting the check for the ignore tag](https://github.com/trufflesecurity/trufflehog/blob/2edd4d326abd10ea6320e03f42c3b6a7cf259b3d/pkg/engine/engine.go#L1338).  (I considered submitting a fix, but I don't know why the URL is being normalized with the port number included, or what the implications would be of not doing that...)

MuneebUllahKhan222:
Fixed in [PR#4968](https://github.com/trufflesecurity/trufflehog/pull/4968)

PR #4968, opened 2026-05-15
fix(postgres): honor ignore tags for default port URLs

Edneam:
## Summary
- Preserve the original Postgres URI match as the primary secret so line-offset and ignore-tag handling use the source text.
- Add regression coverage for `trufflehog:ignore` on Postgres URLs that omit the explicit port and are normalized to `:5432`.

## Tests
- `go test ./pkg/detectors/postgres ./pkg/engine`
- `make test-community`
- `make lint`
- `go run ./hack/checksecretparts -fail ./pkg/detectors`
- `make man && git diff --exit-code docs/man/trufflehog.1`

Addresses #4962.

---

> [!NOTE]
> **Low Risk**
> Localized detector and test changes; behavior improves ignore/line matching without altering verification or connection logic.
> 
> **Overview**
> The Postgres detector now keeps the **exact URI substring** from the scanned file as the primary secret, while `Raw` / `RawV2` stay in the normalized connection form (including default port `:5432`). That aligns Postgres with how other detectors use primary secrets so the engine can find the right line and honor **`trufflehog:ignore`** on the text that actually appears in the source.
> 
> `findUriMatches` returns a small `uriMatch` wrapper (parsed params plus `rawURI`) instead of only parameter maps. Regression tests cover primary vs normalized values and an engine case where ignoring a URL **without an explicit port** still works after normalization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddd94fc702e831dddb457396e6f221857c88336a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gugacyber:
Fix is correct and well-scoped. The root cause is clear: findUriMatches was normalizing the URI (adding :5432) before returning it, so the raw text used for ignore-tag line matching no longer corresponded to what was in the source file.
The uriMatch struct is the right abstraction — keeps params and rawURI together without leaking the raw string into unrelated code paths.
One minor observation: SetPrimarySecretValue(candidateURI.rawURI) sets the primary secret to the raw URI, but Raw and RawV2 are still set to the normalized raw value a few lines above. For ignore-tag matching this is fine since SetPrimarySecretValue is what drives line offset lookup, but worth confirming that having Raw != primary secret value doesn't break any downstream consumers that assume they're the same.

Edneam:
Thank you for the review will do and commit again

Edneam:
i have commit again with your comment in mind .. please review and respond @gugacyber

gugacyber:
The new TestPostgres_RawVsPrimarySecret test addresses exactly the concern I raised — it explicitly asserts that Raw/RawV2 hold the normalized form while GetPrimarySecretValue() returns the original source text, and the comment in the code explains the design decision clearly.
The approach is correct and well-documented. LGTM. Congrats on the fix!

Edneam:
Thank you ..

MuneebUllahKhan222 (approved):
Hi @Dawn-Fighter, Thanks for the contribution, this seems like a well put together PR. I will ask product team for a review on it so we can get it merged.

Edneam:
Thank you

camgunz (approved):

amanfcp (approved):

PR #4971, opened 2026-05-18
[INS-468] Add improved lob detector to defaults.go

mustansir14:
## Summary

The Lob detector existed in the codebase but was never registered in the default detector list in `defaults.go`. This PR adds it to the defaults and, after discovering via corpora testing that the original regex was too loose and produced significant noise, refactors the detector to be more precise and follow current practices.

**Regex tightened to reduce noise (the core fix):**

The original regex relied on a loose proximity-based prefix match against the word `"lob"` and matched any 40-character alphanumeric string:
```
PrefixRegex([]string{"lob"}) + `\b([a-zA-Z0-9_]{40})\b`
```
Corpora testing showed this was extremely noisy. Lob API keys have a well-defined format — they always begin with `live_` or `test_` — so the new regex anchors on that structure:
```
`\b((live|test)_[a-zA-Z0-9_]{35})\b`
```

**Keywords updated to match key prefix:**
- Before: `["lob"]`
- After: `["live_", "test_"]`

This makes pre-filtering align with the actual key format rather than relying on a nearby context word.

**Additional improvements (following current detector practices):**
- `Scanner` struct now accepts an injectable `*http.Client` (via `getClient()` helper) to support test mocking without a global variable.
- Package-level `client` renamed to `defaultClient` to avoid shadowing.
- Verification logic extracted into a dedicated `verify()` method.
- Verification endpoint changed from `GET /v1/addresses` to `POST /v1/us_verifications`. The old endpoint returns `401 Unauthorized` both for invalid keys and for active keys with no billing method on file, making it impossible to distinguish between the two cases. The new endpoint returns `403 Forbidden` for active keys with no billing method, allowing a correct verification signal. Status code handling:
  - `403 Forbidden` → verified (active key, no billing method on file)
  - `422 Unprocessable Entity` → verified (active key, request body is invalid — expected for an empty POST)
  - `401 Unauthorized` → not verified
  - anything else → verification error
- Duplicate matches are now deduplicated before result construction.
- `ExtraData` field added to expose the key environment (`live` or `test`).

**Gating behind feature flag**

Since this is considered a new detector addition, it is gated behind a feature flag. This is why the PR is based off of #4969 which contains some require plumbing for this.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Enables a previously unregistered detector (even if flag-gated) and changes live verification behavior against Lob’s API; regex tightening is intentional but alters what gets reported on scans.
> 
> **Overview**
> Registers the **Lob** detector in the default engine list and gates rollout behind **`LobDetectorEnabled`**, with OSS **`main`** turning that flag on like other new detectors.
> 
> The Lob scanner is tightened to match real API keys (`live_` / `test_` plus 35 alphanumeric/underscore characters) instead of a loose `lob`-proximity 40-character pattern, with Aho-Corasick keywords updated to **`live_`** and **`test_`**. Verification moves to **`POST /v1/us_verifications`**, treating **403** and **422** as active keys and **401** as invalid; results are deduplicated and **`ExtraData.environment`** is set. The scanner also supports an injectable HTTP client via **`verify()`**. Unit and integration tests reflect the new patterns and metadata.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 471f4eae2bc15a837d84108ed6eab4a8f7a34055. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/lob/lob.go
```diff
@@ -12,22 +13,31 @@ import (
...
+		return s.client
+	}
+	return defaultClient
+}
+
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
-	return []string{"lob"}
+	return []string{"live_", "test_"}
```
shahzadhaider1, line 40: Hmmm, several other secret types contain this, so expect high volume in this detector processing

pkg/detectors/lob/lob.go
```diff
@@ -66,6 +73,30 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	switch res.StatusCode {
+	case http.StatusForbidden, http.StatusUnprocessableEntity:
+		// 403 indicates key is active but no billing method on file
+		// 422 indicates key is active but request body is invalid
+		return true, nil
+	case http.StatusUnauthorized:
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected status code: %d", res.StatusCode)
+	}
```
shahzadhaider1, line 97: no 200?

pkg/detectors/lob/lob.go
mustansir14: Yes, that is true. But I believe it is still better than having a three letter keyword that produced much larger noise.

pkg/detectors/lob/lob.go
mustansir14: No. The verification call is a POST request with an empty body, so we do not expect a 200 response.

shahzadhaider1 (approved):

kashifkhan0771 (approved):

camgunz (approved):

pkg/detectors/lob/lob.go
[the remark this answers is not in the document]
mustansir14: Already addressed [here](https://github.com/trufflesecurity/trufflehog/pull/4971#discussion_r3266108059)

PR #4972, opened 2026-05-18
Added source config flags to sharepoint proto

MuneebUllahKhan222:
### Description:
This PR introduces a `include_version`, `include_comments`, `include_attachments` source configs flags for sharepoint source in soures.protos file.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive protobuf fields default to false and are backward compatible; no runtime scan logic changes in this PR.
> 
> **Overview**
> Extends the **Sharepoint** source message in `sources.proto` with three optional boolean config flags: `include_versions`, `include_comments`, and `include_attachments` (fields 3–5), alongside existing OAuth and `site_url`.
> 
> Regenerated **`sources.pb.go`** adds struct fields and getters for those flags, and **`sources.pb.validate.go`** treats them as unconstrained booleans. This is the wire/API contract for toggling version history, comments, and attachments during SharePoint scans; behavior in the connector is not part of this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2a2c12294f8307f501dd9ecbff592d07c126369. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gugacyber (commented):
The poto changes look correct — field numbers 3, 4, 5 are fresh and don't conflict with existing fields, and the raw descriptor bytes in sources.pb.go are consistent with the new fields.
One thing worth noting: the three new flags default to false (protobuf bool default), meaning existing configs silently opt out of scanning versions, comments, and attachments after upgrading. If the previous behavior was to scan these by default, this is a silent breaking change for existing users. Worth documenting in the PR whether the old behavior scanned these or not, and if so, defaulting to true or adding a migration note.
Also, the PR description mentions scan_version (singular) but the field added is scan_versions (plural) — minor inconsistency in the description.

proto/sources.proto
```diff
@@ -467,6 +467,9 @@ message Sharepoint {
     credentials.Oauth2 oauth = 1;
   }
   string site_url = 2;
+  bool scan_versions = 3;
+  bool scan_comments = 4;
+  bool scan_attachments = 5;
```
mustansir14: Can we rename these to `include_xxx` instead of `scan_xxx`? So that it's consistent with similar flags on other sources, such as [confluence](https://github.com/trufflesecurity/trufflehog/blob/0ec3634f6cf66a61912a923fee9d20cc45633a67/proto/sources.proto#L155)

MuneebUllahKhan222: Ahh It will be a hassle to do it in all PRs but I guess I will have to do that.🤦‍♂️

camgunz (approved):

proto/sources.proto
camgunz: I approved but, agree w/ Mustansir here ➕

camgunz:
> One thing worth noting: the three new flags default to false (protobuf bool default), meaning existing configs silently opt out of scanning versions, comments, and attachments after upgrading.

We can fix this w/ feature flags like "override_sharepoint_include_attachments", which we'll ✂️ once SharePoint source users update their configs.

proto/sources.proto
MuneebUllahKhan222: I have renamed the fields, merging now.

PR #4981, opened 2026-05-25
[SCAN-795] HTML decoder: ASPX and entity-encoded HTML support

mustansir14:
## Summary

The HTML decoder failed to process ASPX files in two distinct scenarios, leaving secrets undetected. This PR fixes both gaps and improves attribute extraction for ASP.NET server controls.

## Problem

**1. Fragment ASPX pages (no `<html>` wrapper)**

Master-page-based ASP.NET Web Forms pages omit the outer `<html>` element. Their content is wrapped in namespace-prefixed server controls:

```aspx
<%@ Page MasterPageFile="~/Site.Master" %>
<asp:Content ContentPlaceHolderID="MainContent" runat="server">
  <asp:SqlDataSource ConnectionString="Server=prod;Password=secret" runat="server" />
</asp:Content>
```

The old `looksLikeHTML` regex (`<[a-zA-Z][a-zA-Z0-9]*[\s>/]`) did not match namespace-prefixed tags — the `:` in `<asp:Content` stopped the match before the required terminator. The decoder returned nil and the file was skipped entirely.

**2. Mid-file chunks entirely within entity-encoded blobs**

SharePoint ASPX pages store large HTML content blocks entity-encoded inside XML metadata fields:

```xml
<mso:CanvasContent1 msdt:dt="string">
  &lt;div&gt;&lt;p&gt;-----BEGIN OPENSSH PRIVATE KEY-----&lt;br&gt;...&lt;/p&gt;&lt;/div&gt;
</mso:CanvasContent1>
```

The chunker splits files into 10KB chunks with 3KB overlap. Chunks that land entirely within this blob contain no literal `<tag>` patterns — only `&lt;div&gt;` sequences. The decoder returned nil for these chunks, leaving any secrets they contained undetected.

**3. ASP.NET server control attributes silently dropped**

Even when detection passed, secrets in server control attributes were not extracted. `emitAttributes` filtered by `highSignalAttrs` (`href`, `src`, `value`, etc.) and `data-*` prefixes — neither covers PascalCase attributes like `ConnectionString`, `Text`, or `SelectCommand`.

### `pkg/decoders/html.go`

**Extended `htmlTagPattern`** with an optional namespace group to match `<asp:Content`, `<mso:CanvasContent1>`, etc.:

```go
// Before
var htmlTagPattern = regexp.MustCompile(`<[a-zA-Z][a-zA-Z0-9]*[\s>/]`)

// After
var htmlTagPattern = regexp.MustCompile(`<[a-zA-Z][a-zA-Z0-9]*(?::[a-zA-Z][a-zA-Z0-9]*)?[\s>/]`)
```

**Added `htmlEntityPattern`** to detect chunks that are entirely entity-encoded HTML. The terminator `(?:[\s>/]|&gt;)` covers both attribute-bearing tags (`&lt;div `) and bare tags (`&lt;p&gt;`), while correctly excluding comparison operators (`x &lt; threshold`) and template placeholders (`&lt;YOUR_KEY&gt;` — `_` breaks the alphanumeric run before `&gt;` can match):

```go
var htmlEntityPattern = regexp.MustCompile(
    `&lt;[a-zA-Z][a-zA-Z0-9]*(?::[a-zA-Z][a-zA-Z0-9]*)?(?:[\s>/]|&gt;)`)

func looksLikeHTML(data []byte) bool {
    return htmlTagPattern.Match(data) || htmlEntityPattern.Match(data)
}
```

When an entity-encoded chunk passes the gate, `html.Parse` decodes the entities to literal HTML markup. The engine's `iterativeDecode` then re-applies the HTML decoder at the next depth level to fully extract text and attributes from the inner markup — no changes needed in the decoder itself for this case.

**Emit all attributes for namespace-prefixed elements** in `emitAttributes`. Elements where `n.Data` contains `:` (e.g. `asp:textbox`, `mso:canvascontent1`) are ASP.NET server controls or XML metadata nodes — all their attributes are data payloads:

```go
isNamespaced := strings.Contains(n.Data, ":")
if !isNamespaced && !highSignalAttrs[attr.Key] && !isDataAttr {
    continue
}
```

### `pkg/decoders/html_test.go`

**`TestLooksLikeHTML` additions:**
- Positive: namespace-prefixed literal tags (`<asp:Content`, `<mso:WikiField>`, `<WebPartPages:WebPartZone>`), entity-encoded tags (`&lt;div `, `&lt;p&gt;`, `&lt;br/&gt;`, `&lt;asp:Content&gt;`)
- Negative: comparison operators (`x &lt; threshold`), template placeholders (`&lt;YOUR_KEY&gt;`), double-encoded entities (`&amp;lt;div`), plain text

**`TestHTML_FromChunk` additions:**
- Fragment ASPX with no `<html>` wrapper
- `ConnectionString` and `Text` attribute extraction from ASP.NET server controls
- Entity-encoded chunk with no literal tags (mid-chunk scenario)
- SharePoint `mso:CanvasContent1` pattern showing entity decoding at depth 1

### `main.go`

Enabled `HTMLDecoderEnabled` by default for OSS builds.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Default-on HTML decoding and broader tag/attribute extraction change what content reaches detectors (more coverage, slightly higher false-positive surface on entity heuristics).
> 
> **Overview**
> **Enables the HTML decoder by default** for OSS builds via `HTMLDecoderEnabled` in `main.go`, so markup-heavy sources get normalized text without extra flags.
> 
> The **HTML decoder** now recognizes **namespace-prefixed tags** (e.g. `asp:`, `mso:`) and **entity-encoded tag shapes** (`&lt;div&gt;`, etc.) in `looksLikeHTML`, so fragment ASPX and SharePoint-style chunks are no longer skipped. For elements whose tag name contains `:`, **all attributes** are emitted (not only `href`/`data-*`), surfacing secrets in ASP.NET controls like `ConnectionString` and `Text`. Entity-encoded chunks still decode through `html.Parse` and can be re-processed on later decode depths.
> 
> Tests cover ASPX fragments, server-control attributes, SharePoint `mso:CanvasContent1`, and positive/negative `looksLikeHTML` cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed5e2cbbadd77fa4cb2eb9cb4be1630d185cb21b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


main.go
```diff
@@ -525,6 +525,9 @@ func run(state overseer.State, logSync func() error) {
 	// OSS Default using github graphql api for issues, pr's and comments
 	feature.UseGithubGraphQLAPI.Store(false)
 
+	// OSS Default Use HTML Decoder on
+	feature.HTMLDecoderEnabled.Store(true)
```
rosecodym, line 529: I'm surprised this wasn't already the case. Do you know why it wasn't?

pkg/decoders/html.go
```diff
@@ -196,9 +208,17 @@ func hasSyntaxHighlightClass(n *html.Node) bool {
...
+	// data payloads that may carry secrets (ConnectionString, Text, SelectCommand,
+	// etc.), so we emit every attribute rather than filtering by highSignalAttrs.
+	// After html.Parse the colon is preserved in n.Data even though the name is
+	// lowercased, making strings.Contains a reliable namespace check.
+	isNamespaced := strings.Contains(n.Data, ":")
+
 	for _, attr := range n.Attr {
 		isDataAttr := strings.HasPrefix(attr.Key, "data-")
-		if !highSignalAttrs[attr.Key] && !isDataAttr {
+		if !isNamespaced && !highSignalAttrs[attr.Key] && !isDataAttr {
```
rosecodym: It looks like you can check `isNamespaced` before the prefix comparison (and potentially skip it). There might be a performance benefit in doing that (it certainly wouldn't hurt anything) and it _might_ clarify the code's intention a bit. Is that something you explored?

rosecodym (commented):
Thanks for taking a look at this!

main.go
mustansir14: I'm not really sure. The original PR that added it (#4840) doesn't address this, so my guess is that it was just missed?

pkg/decoders/html.go
mustansir14: You're absolutely right. I'll make the change.

main.go
alafiand: I can't recall why I made that decision in #4840, but agree it should be turned on.

pkg/decoders/html.go
```diff
@@ -116,7 +128,7 @@ func (d *HTML) FromChunk(chunk *sources.Chunk) *DecodableChunk {
 }
 
 func looksLikeHTML(data []byte) bool {
-	return htmlTagPattern.Match(data)
+	return htmlTagPattern.Match(data) || htmlEntityPattern.Match(data)
 }
```
alafiand, line 132: Both patterns share the same core structure and only differ in prefix/terminator, so we _could_ combine them into a single compiled regex. For the `not HTML` path, it would save us an additional scan.

The tradeoff for this minor performance improvement is readability, so splitting it out seems fine for now. We may likely decide to modify these again in the future, so preventing it from being a long abomination is more sustainable.

alafiand (approved):
Looks good, helpful improvements.

shahzadhaider1 (approved):

PR #4993, opened 2026-05-29
[INS-497] Add Pganalyze Read Key Detector

MuneebUllahKhan222:
## Description
This PR adds the pganalyze Read API Key Detector for TruffleHog.

It scans for pganalyze Read API keys and optionally verifies them via the official pganalyze GraphQL API.

Regex:

```regex
\b(pgar_[A-Za-z0-9]{27})\b
```

The detector uses the contextual keyword `pgar_` to reduce false positives.

## Verification
For verification, we use the pganalyze GraphQL API endpoint:

```text
https://app.pganalyze.com/graphql
```

We send a `POST` request with the token in the `Authorization` header using the format:

```text
Authorization: Token <token>
```

A response code of:
- `200 OK` means the API key is valid
- `401 Unauthorized` means the API key is invalid.

This API endpoint is part of the official pganalyze API surface and can be used safely for verification. The verification request performs a read-only GraphQL query and does not perform any destructive actions.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive detector following existing feature-flag and verification patterns; no changes to core scan or auth paths beyond optional outbound HTTP during verify.
> 
> **Overview**
> Adds **PgAnalyze Read Key** detection for tokens matching `pgar_` plus 27 alphanumeric characters, with the `pgar_` keyword for Aho-Corasick pre-filtering.
> 
> When verification is enabled, keys are checked via a `POST` to `https://app.pganalyze.com/graphql` using `Authorization: Token <key>`; **200** marks verified and **401** invalid.
> 
> Wires the scanner into default detectors behind `PgAnalyzeReadKeyDetectorEnabled` (enabled in OSS `main.go`), registers `DetectorType_PgAnalyzeReadKey = 1054` in proto/generated code, and adds unit, pattern, and integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bd646f7620e47ef0213f5d49acd751d3c26f88a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/pganalyzereadkey/pganalyzereadkey.go
```diff
@@ -0,0 +1,144 @@
...
+		uniqueTokens[match[1]] = struct{}{}
+	}
+
+	for token := range uniqueTokens {
+		result := detectors.Result{
+			DetectorType: detector_typepb.DetectorType_PgAnalyzeReadKey,
+			Raw:          []byte(token),
+			SecretParts: map[string]string{
+				"key":         token,
+				"access_type": "read",
```
shahzadhaider1, line 72: I think it's worth adding this to the `ExtraData` as well.

shahzadhaider1: Since the detector name is `PgAnalyzeReadKey`, which already indicates that this is a read-only key, what additional value does `access_type: read` provide in the secret metadata?

Also, can this secret type have other permission levels (e.g. write or admin), or is it always read-only?

MuneebUllahKhan222: The token only has read permissions. Also, I think we should keep `access_type` because if we decide to build an analyzer for this detector in the future, we'll likely want a single analyzer that can handle different types of pganalyze tokens. Since` detector_type` isn't available in the analyzer input, the analyzer wouldn't know which detector triggered it. Having `access_type` would provide the necessary context to determine how the token should be analyzed.

MuneebUllahKhan222: Also `ExtraData` doesn't make sense here because it is clear from the detector that the found token is a read key.

camgunz (approved):

amanfcp (approved):
The wiring LGTM

PR #5007, opened 2026-06-02
[INS-472] [INS-515] Add user detector to defaults.go, gate it behind feat flag, update verification logic and add custom ep configuration

MuneebUllahKhan222:
### Description:
This PR adds the `User` detector to `defaults.go` and gates it behind the appropriate feature flag.

It also updates the detector verification logic to align with the latest API behavior and validation requirements, including support for configuring a custom User.com endpoint since API keys must be verified against the workspace-specific domain rather than a single global endpoint.

Additionally, the PR tightens the detector regex by removing support for the `+`, `-`, `_`, `.`, and `=` characters.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Turning on a new default-gated detector changes scan behavior and outbound verification traffic; the stricter regex and endpoint pairing may alter findings versus the old fixed-host detector.
> 
> **Overview**
> Registers the **User** (User.com) detector in the default engine list behind a new `UserDetectorEnabled` feature flag (enabled in OSS `main.go`), matching other gated detectors.
> 
> The **User** scanner is reworked so secrets are **token + workspace host** pairs: it finds 64-character alphanumeric API keys and `*.user.com` hosts in chunk data (or via configured endpoints), emits `RawV2` / `SecretParts` with both parts, and verifies with `GET /api/public/users/` on that host using deduplicated HTTP and a cache for bad hostnames. The key regex no longer allows `+`, `-`, `_`, `.`, or `=`; keywords shift to `user.com`.
> 
> Tests cover endpoint-in-data vs custom endpoint-only flows and updated pattern expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05055ae50f6af75d69ff4f16bf9839f5289d4d59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/main.go
+++ b/main.go
@@ -536,12 +536,9 @@ func run(state overseer.State, logSync func() error) {
 	feature.DatadogApiKeyDetectorEnabled.Store(true)
 	feature.TlyDetectorEnabled.Store(true)
 	feature.WitDetectorEnabled.Store(true)
-<<<<<<< user-detector
-	feature.UserDetectorEnabled.Store(true)
-=======
 	feature.RevDetectorEnabled.Store(true)
+	feature.UserDetectorEnabled.Store(true)
 
->>>>>>> main
 	conf := &config.Config{}
 	if *configFilename != "" {
 		var err error
--- a/main.go
+++ b/main.go
@@ -536,7 +536,7 @@ func run(state overseer.State, logSync func() error) {
 	feature.DatadogApiKeyDetectorEnabled.Store(true)
 	feature.TlyDetectorEnabled.Store(true)
 	feature.WitDetectorEnabled.Store(true)
-
+	feature.UserDetectorEnabled.Store(true)
 	conf := &config.Config{}
 	if *configFilename != "" {
 		var err error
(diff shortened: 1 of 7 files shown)


unsmith:
I feel like this is similar to https://github.com/trufflesecurity/trufflehog/pull/5006 (TLy) but doesn't include some of the same improvements, like fixing the resource leak in the defer() calls or the refactored verify function.  Those might be good to include here as well.

MuneebUllahKhan222:
> I feel like this is similar to #5006 (TLy) but doesn't include some of the same improvements, like fixing the resource leak in the defer() calls or the refactored verify function. Those might be good to include here as well.

Yeah, this was the first detector I worked on for adding to the defaults.go list. Initially, the intent was only to add it there, but later I thought that since I was already touching the detector, it would make sense to refactor it as well to align with our modern detector patterns.

I forgot to include that refactor in this PR, but I have updated it.

mustansir14 (approved):

unsmith (approved):

PR #5011, opened 2026-06-02
Enigma detector enhance verification to prevent false verified findings

shahzadhaider1:
## Summary
We have seen an Enigma finding that was marked verified but is a false positive. The finding was verified at the first scan then flipped to unverified and marked rotated about 12 hours later. The old verification logic marked a secret verified on any 2xx response from the businesses match endpoint, so a transient upstream 2xx was enough to produce a false verified result.

This PR hardens the verification path so the verified flag reflects a real auth check.

## Changes

- Verify against the GraphQL endpoint at `https://api.enigma.com/graphql` using a minimal query that selects only `__typename`. This does not depend on Enigma schema changes or any specific entity id.
- Mark verified only on HTTP 200 with a JSON content type. A 200 with a non JSON body is treated as a verification error so an intercepting proxy or block page is not mistaken for a live key.
- Treat 401 as unverified.
- Treat any other status as a verification error instead of silently passing or failing.
- Surface transport and verification errors through SetVerificationError.

## What we know
- Confirmed by Postman that a valid key returns 200 with a JSON body and an invalid or missing key returns 401.
- The reported finding self-corrected on rescan. The DB record shows `verified false` and `verification_ever_determinate true`.
- We could not reproduce a 2xx for an invalid key against the old endpoint. The most likely cause is transient upstream behavior.

## Out of scope
- The detection regex is left as is. Valid Enigma keys are 40 character alphanumeric strings and tightening the regex risks missing real keys.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes live verification behavior for Enigma (fewer false positives, more unknowns on odd responses); default OSS enables the gated detector, which may increase Enigma hits when enabled.
> 
> **Overview**
> Hardens **Enigma API key verification** so the `verified` flag is not set on weak or misleading HTTP responses (addressing false verified findings from the old businesses-match check that treated any 2xx as success).
> 
> Verification now **POSTs to** `https://api.enigma.com/graphql` with a minimal `{ __typename }` query, marks keys **verified only on HTTP 200** when `Content-Type` includes `application/json`, treats **401 as unverified**, and reports other statuses or non-JSON 200s via **`SetVerificationError`**. Detection regex is unchanged.
> 
> Wires **`EnigmaDetectorEnabled`** (default **on** in OSS `main.go`) so the Enigma scanner is included in the default detector list only when the flag is set, consistent with Pinecone/Cloudinary/GitLab OAuth. Integration tests use the **`detectors6`** secret bundle and expect **`SecretParts`** on results.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e638c998ccefdb1505399bb6bae586139662ce2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/main.go
+++ b/main.go
@@ -532,6 +532,7 @@ func run(state overseer.State, logSync func() error) {
 	feature.PineconeDetectorEnabled.Store(true)
 	feature.CloudinaryDetectorEnabled.Store(true)
 	feature.GitLabOAuthDetectorEnabled.Store(true)
+	feature.EnigmaDetectorEnabled.Store(true)
 
 	conf := &config.Config{}
 	if *configFilename != "" {
--- a/pkg/detectors/enigma/enigma.go
+++ b/pkg/detectors/enigma/enigma.go
@@ -2,6 +2,8 @@ package enigma
 
 import (
 	"context"
+	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
@@ -33,7 +35,6 @@ func (s Scanner) Keywords() []string {
 // FromData will find and optionally verify Enigma secrets in a given set of bytes.
 func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
 	dataStr := string(data)
-
 	matches := keyPat.FindAllStringSubmatch(dataStr, -1)
 
 	for _, match := range matches {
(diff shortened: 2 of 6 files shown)


camgunz (approved):
Nice, also good idea to migrate off the legacy REST API to the GraphQL API 👍🏻

amanfcp (approved):
Skimmed through. Thanks for the refactor too.

PR #5013, opened 2026-06-03
fix(facebookoauth): verify app creds against /{app-id} instead of /me

lukem-ts:
### Description:
The verifier built a valid app access token ({app-id}|{app-secret}) but queried the /me node, which only accepts user/page tokens. Facebook rejects an app token there with HTTP 400 OAuthException code 2500 ("An active access token must be used to query information about the current user."), so the check returned 400 for every App ID/Secret pair — valid or not — and could never mark a result verified.

Query the documented validation endpoint instead: GET /{app-id}?fields=roles with the app token. Valid credentials return 200; invalid pairs return 400 OAuthException 190. Status handling now distinguishes verified (200), unverified (400/401), and indeterminate (other statuses -> verification error).

image

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes outbound verification to Facebook’s Graph API and how secrets are classified (verified vs unverified vs error); behavior is more correct but still involves live credential checks against a third-party API.
> 
> **Overview**
> **Fixes Facebook App ID/secret verification** so valid pairs can be marked verified instead of always failing.
> 
> Verification no longer calls **`/me`** with an app access token (`{app-id}|{secret}`), which Facebook rejects for every pair (HTTP 400). It now uses **`GET /{app-id}?fields=roles`** with the same app token: **200** → verified, **400/401** → unverified, other statuses → verification error via **`SetVerificationError`**.
> 
> On a successful **200**, optional **`ExtraData`** is populated from the roles payload (`app_roles_count`, comma-separated `user:role` in `app_roles`). Response bodies are drained on close.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d9eff7f28eaa0c4fd1bf59042a92d3c643d1ab7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/facebookoauth/facebookoauth.go
+++ b/pkg/detectors/facebookoauth/facebookoauth.go
@@ -65,9 +65,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 				isVerified, extraData, verificationErr := verifyFacebookOAuth(ctx, apiIdRes, apiSecretRes)
 				s1.Verified = isVerified
 				s1.ExtraData = extraData
-				if verificationErr != nil {
-					s1.SetVerificationError(verificationErr, apiSecretRes)
-				}
+				s1.SetVerificationError(verificationErr, apiSecretRes)
 			}
 
 			results = append(results, s1)
--- a/pkg/detectors/facebookoauth/facebookoauth.go
+++ b/pkg/detectors/facebookoauth/facebookoauth.go
@@ -2,8 +2,11 @@ package facebookoauth
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
+	"strconv"
 	"strings"
 
 	regexp "github.com/wasilibs/go-re2"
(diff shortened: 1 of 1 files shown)


pkg/detectors/facebookoauth/facebookoauth.go
```diff
@@ -59,18 +62,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-				if err == nil {
-					defer func() { _ = res.Body.Close() }()
-					if res.StatusCode >= 200 && res.StatusCode < 300 {
-						s1.Verified = true
-					}
+				isVerified, extraData, verificationErr := verifyFacebookOAuth(ctx, apiIdRes, apiSecretRes)
+				s1.Verified = isVerified
+				s1.ExtraData = extraData
+				if verificationErr != nil {
+					s1.SetVerificationError(verificationErr, apiSecretRes)
```
shahzadhaider1: no need for this `if` here. `SetVerificationError` only adds error if it exists

lukem-ts: updated, thanks!

shahzadhaider1 (approved):
lgtm, thanks!

MuneebUllahKhan222 (approved):

bryanbeverly (approved):
Approving on behalf of @trufflesecurity/ftes

PR #5016, opened 2026-06-04
Improve YouTube API key verifier errors

srkyn:
## Description
Improves the YouTube API key detector verifier by returning explicit verification errors for transport failures and unexpected API responses.

This is a narrow slice related to the detector verifier error-handling work discussed in #4051. The PR avoids closing keywords so the umbrella issue can stay open for the remaining detectors.

## Changes
- Adds injectable HTTP client support to the YouTube API key scanner
- Extracts verification into a helper that returns verified state plus verification error
- Treats 400 and 401 responses as unverified without a verification error
- Treats 403 responses as verified because YouTube uses them for recognized keys blocked by permissions or quota
- Adds local verifier tests using constant HTTP responses

## Tests
- `go test ./pkg/detectors/youtubeapikey`
- `go vet ./pkg/detectors/youtubeapikey`

---

> [!NOTE]
> **Low Risk**
> Scoped to one detector’s verify path and test doubles; behavior change is treating 403 as verified, which reduces false negatives for restricted keys.
> 
> **Overview**
> Refactors the YouTube API key detector so verification reports **verified vs unverified** separately from **verification errors**, matching the broader detector verifier work in #4051.
> 
> Verification moves into `verifyYoutubeAPIKey`, which returns a bool and an error. The scanner can use an injectable `http.Client` (via `getClient`) for tests. **200** and **403** count as verified (403 covers valid keys with quota/permission limits); **400** and **401** are unverified with no error; other status codes and transport failures surface through `SetVerificationError`.
> 
> Adds `TestYoutubeApiKey_Verification` with a stub HTTP client for OK, 403, 400, and unexpected 500 responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b33f557d6b18b323b284242ecf0f5c38723482f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/youtubeapikey/youtubeapikey.go
+++ b/pkg/detectors/youtubeapikey/youtubeapikey.go
@@ -2,6 +2,7 @@ package youtubeapikey
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
@@ -14,6 +15,7 @@ import (
 
 type Scanner struct {
 	detectors.DefaultMultiPartCredentialProvider
+	client *http.Client
 }
 
 // Ensure the Scanner satisfies the interface at compile time.
@@ -31,6 +33,13 @@ var (
 // Use identifiers in the secret preferably, or the provider name.
 func (s Scanner) Keywords() []string {
 	return []string{"youtube"}
+}
+
+func (s Scanner) getClient() *http.Client {
+	if s.client != nil {
+		return s.client
+	}
+	return client
 }
 
 // FromData will find and optionally verify YoutubeApiKey secrets in a given set of bytes.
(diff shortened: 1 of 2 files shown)


pkg/detectors/youtubeapikey/youtubeapikey.go
```diff
@@ -74,6 +75,27 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+	if err != nil {
+		return false, err
+	}
+	defer func() { _ = res.Body.Close() }()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		return true, nil
+	case http.StatusBadRequest, http.StatusUnauthorized, http.StatusForbidden:
+		return false, nil
```
shahzadhaider1, line 94: AFAIK, Google does return 403 for valid credentials that lack proper authorization. Could you double check this?
```
Error type	Error detail	Description
forbidden (403)	forbidden	Access forbidden. The request may not be properly authorized.
```

srkyn: Good catch. Google's YouTube API docs list 403 for authorization and quota failures, so it can come from a recognized key whose request is blocked. I now treat 403 as verified, added a regression test, and rebased on current main. `go test ./pkg/detectors/youtubeapikey` and `go vet ./pkg/detectors/youtubeapikey` both pass.

PR #5018, opened 2026-06-04
fix(detectors): validate Kraken base64 matches

oxura:
## Summary
- Fix the Kraken detector so regex boundary characters are not included in captured API/private key values.
- Reject malformed Kraken key candidates that do not decode as standard base64 before emitting results.
- Update Kraken pattern tests with valid base64 fixtures plus a quoted-env-var regression case.

## Problem
Kraken credentials are base64-encoded values. The existing detector test fixtures included `=` padding in the middle of the candidate strings, which is not valid base64. The detector also included the trailing boundary character in the regex capture group, so common quoted forms like `KRAKEN_API_KEY=...` could include the closing quote in the captured secret.

## Fix
The Kraken detector now captures only the credential value and treats the following space/quote/newline as a boundary outside the capture group. It also base64-decodes both the API key and private key candidates before emitting a result, reducing false positives from malformed padding.

## Tests
- `go test ./pkg/detectors/kraken -run '^TestKraken_Pattern$' -tags=detectors -count=1` ✅

## Notes
- Related to #3817. This is intentionally scoped to Kraken and includes production detector behavior, not only fixture cleanup.
- Full `go test ./pkg/detectors/kraken -tags=detectors` is not runnable in this local environment because `TestKraken_FromChunk` requires GCP Secret Manager application default credentials for test secrets.
- I noticed #4831 also touches Kraken test fixtures; this PR is narrower and also fixes the detector capture/base64 validation behavior.

---

> [!NOTE]
> **Low Risk**
> Scoped Kraken secret-detection changes that reduce false positives; no auth or broad infrastructure impact.
> 
> **Overview**
> The Kraken detector now captures only the credential token in regex (trailing space/quote/newline are boundaries, not part of the match) and **drops candidates that do not decode as base64** for both API and private keys, including unpadded private keys via `RawStdEncoding`. Verification reuses the validated decode for signing instead of ignoring decode errors.
> 
> Pattern tests use real base64 fixtures, add quoted env-var and unpadded-private-key cases, and expect no hits when padding is malformed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 360ba92f8ff2743dcb6e562f90b4e40a9c75cf3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/kraken/kraken.go
+++ b/pkg/detectors/kraken/kraken.go
@@ -32,8 +32,8 @@ var (
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	// Bounds have been removed because there are some cases that tokens have trailing frontslash(/) or plus sign (+)
-	keyPat     = regexp.MustCompile(detectors.PrefixRegex([]string{"kraken"}) + `\b([0-9A-Za-z\/\+=]{56}[ "'\r\n]{1})`)
-	privKeyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"kraken"}) + `\b([0-9A-Za-z\/\+=]{86,88}[ "'\r\n]{1})`)
+	keyPat     = regexp.MustCompile(detectors.PrefixRegex([]string{"kraken"}) + `\b([0-9A-Za-z\/\+=]{56})[ "'\r\n]`)
+	privKeyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"kraken"}) + `\b([0-9A-Za-z\/\+=]{86,88})[ "'\r\n]`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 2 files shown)


oxura:
Thanks for the review. I pushed a follow-up that addresses both points:

- `decodeBase64` now accepts both padded standard base64 and unpadded standard base64, so 86/87-character Kraken private keys are preserved and verification uses the decoded bytes from the same helper.
- The invalid API-key fixture now has malformed padding in the middle of the candidate, and the tests cover malformed API key and malformed private key independently.
- Added a regression case for an unpadded private key.

Verification:

- `go test ./pkg/detectors/kraken -run '^TestKraken_Pattern$' -tags=detectors -count=1`
- `git diff --check`

oxura:
﻿Thanks, this was a valid catch. I pushed a follow-up that puts a second `kraken` keyword before the private-key candidate in both invalid fixture cases, matching the structure used by the valid paired-key cases. That keeps each candidate inside `PrefixRegex` range, so the tests now exercise the API-key and private-key `decodeBase64` rejection paths instead of passing before the private-key regex runs.

I also mutation-checked the coverage locally:

- temporarily bypassing API-key base64 validation makes `invalid pattern - malformed api key base64 padding` fail with one unexpected result
- temporarily bypassing private-key base64 validation makes `invalid pattern - malformed private key base64 padding` fail with one unexpected result

Verification after restoring the implementation:

- `go test ./pkg/detectors/kraken -run '^TestKraken_Pattern$' -tags=detectors -count=1`
- `git diff --check`

PR #5021, opened 2026-06-05
fix(detectors): honor ignores for normalized Postgres URIs

### LINKED ISSUE #4962: `# trufflehog:ignore` not honored for Postgres URLs without an explicit port

oxura:
## Description

Fixes #4962.

Postgres URI detection normalizes candidates before emitting `Raw`, including adding the default `:5432` port when the original URI omitted it. The engine uses the emitted secret value to locate the source line for `trufflehog:ignore`, so an ignored line like:

```text
postgresql://user:password@host/db # trufflehog:ignore
```

could still report because the normalized value (`...@host:5432`) does not appear in the scanned chunk.

This change keeps the normalized `Raw`/`RawV2` behavior for the detector result, but also stores the exact matched URI as the primary secret value so the engine can resolve line numbers and same-line ignore tags against the original source text.

## Tests

- `go test ./pkg/detectors/postgres -count=1`
- `go test ./pkg/engine -run TestEngineignoreLine -count=1`
- `go test ./pkg/engine -count=1`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Localized Postgres detector and ignore-line behavior; normalized secret output is unchanged aside from metadata used for line matching.
> 
> **Overview**
> Fixes **`trufflehog:ignore`** on Postgres connection strings when the detector **normalizes** URIs (e.g. injects default **`:5432`**) so the emitted `Raw` no longer matches the text in the file.
> 
> The Postgres scanner now threads the **exact matched URI** through `findUriMatches` via a `uriMatch` struct and calls **`SetPrimarySecretValue`** with that string while keeping normalized `Raw`/`RawV2` unchanged. The engine already uses the primary secret (via `effectiveSecret`) for line lookup and same-line ignore tags, so ignores work when the source omits an explicit port.
> 
> Tests assert the primary secret stays the unnormalized URI and add an engine case for an ignored Postgres URL without a port; one existing ignore-line fixture gets a whitespace-only alignment fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b97594fba2e08e5493b694b50caadd4a930e402. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/postgres/postgres.go
+++ b/pkg/detectors/postgres/postgres.go
@@ -52,6 +52,11 @@ var (
 	connStrPartPattern                    = regexp.MustCompile(`([[:alpha:]]+)='(.+?)' ?`)
 )
 
+type uriMatch struct {
+	params map[string]string
+	uri    string
+}
+
 type Scanner struct {
 	detectors.DefaultMultiPartCredentialProvider
 	detectLoopback bool // Automated tests run against localhost, but we want to ignore those results in the wild
@@ -95,10 +100,11 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
 	var results []detectors.Result
 	candidateParamSets := findUriMatches(data, s.ignorePatterns)
 
-	for _, params := range candidateParamSets {
+	for _, candidate := range candidateParamSets {
 		if common.IsDone(ctx) {
 			break
 		}
+		params := candidate.params
 		user, ok := params[pgUser]
 		if !ok {
 			continue
@@ -141,6 +147,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) ([]dete
 			RawV2:        raw,
 			SecretParts:  map[string]string{"connection_string": string(raw)},
 		}
+		result.SetPrimarySecretValue(candidate.uri)
 
 		// We don't need to normalize the (deprecated) requiressl option into the (up-to-date) sslmode option - pq can
 		// do it for us - but we will do it anyway here so that when we later capture sslmode into ExtraData we will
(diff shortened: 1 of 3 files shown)


oxura:
Closing this as superseded by #4968, which was merged into `main` and addresses #4962 with the same primary-secret approach and regression coverage.

I rebased locally after seeing this PR become conflicting, and the conflict is because `main` now already keeps the original Postgres URI match as the primary secret while preserving the normalized `Raw`/`RawV2` value. Keeping this PR open would only leave a stale duplicate conflict.

Thanks for taking care of the fix.

PR #5022, opened 2026-06-05
fix: scan files with lines exceeding bufio's default 64 KB token limit

MuneebUllahKhan222:
### Description:
- Raise `bufio.Scanner` max token size from 64 KB to 10 MB in `gitChunk` so files with long lines (minified JS, base64 blobs, etc.) are fully scanned
- Keep the initial buffer at 4 KB (bufio default) to avoid unnecessary allocations on the common case of small diffs
- Add `Scanner.Err()` check after the scan loop to surface scanner failures as logged errors instead of silent no-ops

### Root Cause
`gitChunk` used `bufio.NewScanner` with the default 64 KB max token size. Any line exceeding that limit caused `Scan()` to return `false` with `bufio.Scanner: token too long` — but since `Err()` was never checked, the function exited silently as if it had reached EOF. No secrets were scanned, no error was reported.

### Test Changes
- Added `TestGitChunk_LongLine` — creates a local git repo with a single 100 KB line (well above the old 64 KB cap), scans it, and asserts at least one chunk is produced. This test would have hung/failed before the fix.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core git chunking for large diffs and allows up to 10 MB per scanned line, which can increase memory use on pathological files but fixes missed secret detection.
> 
> **Overview**
> Fixes a **silent scan gap** in large git diffs: `gitChunk` no longer stops at the first line over bufio’s default **64 KB** token limit.
> 
> **`gitChunk`** now configures `bufio.Scanner` with a **10 MB** max token and a **4 KB** starting buffer so long single-line files (minified JS, base64, etc.) are read fully and still flow through the existing oversize-line chunking. After the scan loop, **`Scanner.Err()`** is checked and failures are logged instead of treating them like EOF.
> 
> Adds **`TestGitChunk_LongLine`**, which commits a **100 KB** one-line file and asserts the git source emits the expected chunks (behavior that failed before the fix).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a715feff498d057b0a54fcd3cbf2bc12291a3f6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -750,7 +750,6 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 
 		email := commit.Author
 		when := commit.Date.UTC().Format("2006-01-02 15:04:05 -0700")
-
 		if fullHash != lastCommitHash {
 			depth++
 			lastCommitHash = fullHash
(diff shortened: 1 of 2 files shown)


camgunz (approved):
Great catch!

PR #5028, opened 2026-06-10
Fix: Resolve known dedup issues in notifierWorker

MuneebUllahKhan222:
## Problem

`notifierWorker` could report the same credential multiple times due to a broken deduplication condition.

### Old deduplication condition

The condition guarding the deduplication cache only suppressed duplicates when the decoder type differed between two results for the same key:

```go
if val, ok := e.dedupeCache.Get(key); ok && (val != result.DecoderType) {
    continue
}
```

When the decoder type was the same, the common case `val != result.DecoderType` evaluated to `false`, so every occurrence passed through even if the key was already present in the cache.

This commonly occurred for results produced from the peek-overlap between consecutive chunks, where the same credential at a chunk boundary appeared in both:

* Chunk N (as peek data)
* Chunk N+1 (as main content)

As a result, the same finding could be reported multiple times.

### Cache eviction

Even if the deduplication condition had been working correctly, the cache size of 512 entries was small enough that entries could be evicted during larger scans, allowing previously seen findings to be re-admitted and reported again.

## Fix

Simplify the deduplication logic to suppress any result whose key is already present in the cache:

```go
if _, ok := e.dedupeCache.Get(key); ok {
    continue
}
```

This is safe because the deduplication key already contains sufficient source-specific metadata, including:

* File path
* Line numbers
* Commit hash
* Other location-specific metadata

As a result, genuinely distinct occurrences of the same credential generate different keys and will not be incorrectly suppressed.

### Increase cache size

Increase the deduplication cache size from **512** to **5,000** entries to reduce eviction-driven re-admission during larger scans.

### Store MD5 hashes instead of full keys

The cache now stores MD5 hashes of deduplication keys rather than the full key strings.

Since an MD5 hash is 16 bytes, a cache containing 5,000 entries requires approximately:

```text
5,000 × 16 bytes = 80,000 bytes
```

or roughly **80 KB** of storage for the cached hashes themselves, making the larger cache size a reasonable memory trade-off.

### Remove unused cache value

Change the cache value type from:

```go
lru.Cache[string, detectorspb.DecoderType]
```

to:

```go
lru.Cache[string, struct{}]
```
### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches the path that decides which findings are emitted to users; incorrect dedup could hide real hits, though keys still include location metadata and detector identity.
> 
> **Overview**
> **Fixes duplicate secret reporting** in `notifierWorker` by treating any cache hit as a duplicate instead of only when decoder types differed (which let same-decoder repeats through, including peek-overlap at chunk boundaries).
> 
> The dedupe key now includes **`DetectorName`** (for custom detectors sharing a type), is stored as an **MD5 hash** of the full key string, and the LRU grows from **512 to 5,000** entries with **`struct{}`** values instead of decoder types. A false-positive overlap test expecting two emissions after the behavior change was removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f55d96f89f6c05a8345b6294fb51dbb751b9d6e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -215,7 +215,7 @@ type Engine struct {
 
 	// dedupeCache is used to deduplicate results by comparing the
 	// detector type, raw result, and source metadata
-	dedupeCache *lru.Cache[string, detectorspb.DecoderType]
+	dedupeCache *lru.Cache[string, struct{}]
 
 	// verify determines whether the scanner will attempt to verify candidate secrets.
 	verify bool
@@ -503,14 +503,12 @@ func filterDetectors(filterFunc func(detectors.Detector) bool, input []detectors
 // deduplication efforts, allowing the engine to quickly check if a chunk has
 // been processed before, thereby saving computational overhead.
 func (e *Engine) initialize(ctx context.Context) error {
-	// TODO (ahrav): Determine the optimal cache size.
-	// KNOWN ISSUE: 512 entries is far too small for large scans. Under concurrent notifier
-	// workers a single burst of unique findings easily evicts previously seen keys, allowing
-	// the same secret to be re-emitted on every subsequent pass. Raise to at least 10000
-	// (or make configurable via Config).
-	const cacheSize = 512 // number of entries in the LRU cache
+	// The cache size is set to 10000 entries, which is a balance between memory usage and the need for effective deduplication.
+	// On average a cache entry would be between 500-1000 bytes, so this would use around 5-10 MB of memory.
+	// This should be sufficient for most scans while keeping memory usage reasonable.
+	const cacheSize = 10000
 
-	cache, err := lru.New[string, detectorspb.DecoderType](cacheSize)
+	cache, err := lru.New[string, struct{}](cacheSize)
 	if err != nil {
 		return fmt.Errorf("failed to initialize LRU cache: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -503,14 +503,12 @@ func filterDetectors(filterFunc func(detectors.Detector) bool, input []detectors
...
-	// TODO (ahrav): Determine the optimal cache size.
-	// KNOWN ISSUE: 512 entries is far too small for large scans. Under concurrent notifier
-	// workers a single burst of unique findings easily evicts previously seen keys, allowing
-	// the same secret to be re-emitted on every subsequent pass. Raise to at least 10000
-	// (or make configurable via Config).
-	const cacheSize = 512 // number of entries in the LRU cache
-
-	cache, err := lru.New[string, detectorspb.DecoderType](cacheSize)
+	// The cache size is set to 10000 entries, which is a balance between memory usage and the need for effective deduplication.
+	// On average a cache entry would be between 500-1000 bytes, so this would use around 5-10 MB of memory.
```
camgunz: One thing you could consider is hashing; MD5 is fast and we wouldn't be worried about collisions here. A 1K cache key is biiiiiiiig lol

camgunz (approved):
LGTM; but see what @trufflesecurity/scanning says

pkg/engine/engine_test.go
```diff
@@ -925,7 +925,7 @@ func TestEngine_FalsePositivesRetainedCorrectly(t *testing.T) {
 				passthroughDetector{detectorType: detector_typepb.DetectorType(-2), keywords: []string{"ample"}},
 			},
 			retainFalsePositives:      true,
-			wantUnverifiedSecretCount: 2,
+			wantUnverifiedSecretCount: 1,
```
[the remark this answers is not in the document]
MuneebUllahKhan222: Yeah, this test doesn't make much sense anymore. Both detectors now produce exactly the same key because they have the same `DetectorType (AliBaba)`. As a result, the test is no longer validating any meaningful behavior, so I'm going to remove it.

pkg/engine/engine.go
MuneebUllahKhan222: Thanks for the MD5 suggestion, I hadn't considered it, but it was a great idea, so I went ahead and implemented it.

Although a cache size of 10,000 wouldn't really be an issue anymore since we're now hashing the keys (10,000 entries would only require about 160 KB for the hashes themselves), I've reduced the cache size to a more conservative value of 5,000 entries for now.

If we find that we need a larger cache in the future, we can always increase it.

rosecodym (approved):
I think this is ok. I'm a little confused by why it was originally written the way it was, though - it's not that there's no obvious purpose, its that it was apparently known to be wrong at that time but written incorrectly anyway. Maybe we didn't have rich enough metadata then to make a proper dedup key?

PR #5031, opened 2026-06-11
fix(detectors): report Omnisend verifier errors

whitewooood:
### Description:

Related to #4051. This updates the Omnisend detector verifier to report indeterminate verification failures instead of silently ignoring them.

Before this change, Omnisend verification only marked 2xx responses as verified. Network/client errors and unexpected API statuses were dropped, which made verifier failures indistinguishable from determinately invalid credentials.

This PR keeps behavior narrow:
- 2xx responses verify the secret.
- 401 and 403 remain determinately unverified without a verification error.
- client/request failures and unexpected statuses now populate `VerificationError`.
- response bodies are drained and closed on verifier responses.

### Local validation:

- [x] `gofmt -w pkg/detectors/omnisend/omnisend.go pkg/detectors/omnisend/omnisend_test.go`
- [x] `git diff --check`
- [x] `go test -count=1 ./pkg/detectors/omnisend`
- [x] `go test -count=1 ./pkg/detectors`
- [x] `go vet ./pkg/detectors/omnisend`
- [x] `go run ./hack/checksecretparts -fail ./pkg/detectors`
- [x] `go test -timeout=5m -count=1 ./pkg/detectors/...`

### Checklist:

* [x] Tests passing (`go test` commands listed above)
* [x] Lint-adjacent checks passing (`go vet`, `git diff --check`, `checksecretparts` listed above)

---

> [!NOTE]
> **Low Risk**
> Scoped to Omnisend detector verification behavior and tests; no auth or core engine changes, with a slight tightening from any 2xx to 200-only for verified.
> 
> **Overview**
> Refactors the **Omnisend** detector verifier so indeterminate failures are visible instead of being dropped silently.
> 
> Verification moves into `verifyMatch` with an injectable `http.Client` on `Scanner` (defaulting to the shared sane client). **200 OK** marks the key verified; **401/403** stay unverified with no verification error; request failures and other HTTP statuses set **`VerificationError`**. Response bodies are drained and closed after the contacts API call.
> 
> Adds unit tests that mock the transport to assert the verification URL/header and cover success, auth failures, network errors, and unexpected status codes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cb2345fc1019a0af0cbc74f2d4dc5afdd0b2cdf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/omnisend/omnisend.go
+++ b/pkg/detectors/omnisend/omnisend.go
@@ -2,6 +2,8 @@ package omnisend
 
 import (
 	"context"
+	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
@@ -12,13 +14,15 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detector_typepb"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"omnisend"}) + `\b([a-z0-9A-Z-]{75})\b`)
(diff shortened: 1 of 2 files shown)


pkg/detectors/omnisend/omnisend.go
```diff
@@ -66,6 +66,35 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
+
+	switch res.StatusCode {
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return false, nil
+	default:
+		if res.StatusCode >= 200 && res.StatusCode < 300 {
+			return true, nil
+		}
+		return false, fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
+	}
```
shahzadhaider1, line 94: It'd be better to test the actual api with real credentials and observe the API behavior and then adjust these cases accordingly, instead of setting generic ranges/codes.

whitewooood: I narrowed this to endpoint-specific status handling instead of accepting any 2xx response. The verifier now only treats `200 OK` as verified, keeps `401`/`403` as unverified, and reports other statuses as verification errors.

I do not have valid Omnisend credentials to confirm a successful account response. I did verify the live endpoint with an invalid key and without the header; both returned `403 {"error":"Forbidden"}`. The Omnisend v3 List contacts docs list `200`, `401`, `403`, `408`, `422`, and `429` for this endpoint, so I added a regression test to ensure `201 Created` is no longer treated as verified.

PR #5041, opened 2026-06-16
fix(sources/filesystem): order resume comparison by path component

genisis0x:
### Description:

Resuming a filesystem scan could silently skip an entire sibling directory when one directory name is a prefix of another (e.g. `blue-team` and `blue-team-deprecated`).

`scanDir` decided whether a directory was already scanned by comparing its path against the stored resume point with a raw string comparison (`path < resumeAfter`). Raw string order does not match the depth-first order `os.ReadDir` produces, because the separator `/` (0x2F) sorts *after* characters that are valid inside a path component such as `-` (0x2D) and `.` (0x2E).

So when resuming inside `blue-team` with resume point `/root/blue-team/project-notes.txt`, the check evaluated:

```
"/root/blue-team-deprecated" < "/root/blue-team/project-notes.txt"  // true: '-' < '/'
```

and skipped `blue-team-deprecated` entirely — every file beneath it (e.g. `AWSCredentials.txt`) was never enumerated or scanned. Renaming `blue-team` to `blue-team2` made the problem disappear, matching the report.

The fix compares the two paths component by component, which matches `os.ReadDir`'s ordering (`blue-team` sorts before `blue-team-deprecated`). The in-subtree `HasPrefix` checks and the per-entry loop were already correct and are unchanged.

Closes #5039

Tests:
- `TestResumptionWithPrefixSiblingDirectories` — integration test reproducing the report (resume inside `blue-team`, assert the sibling `blue-team-deprecated/AWSCredentials.txt` is still scanned).
- `TestComparePathsForResume` — unit test for the comparison helper, including the prefix-sibling case and the existing before/after/ancestor cases.
- Existing `TestResumption*` tests still pass.

### Checklist:
* [x] Tests passing (`make test-community`)?
* [x] Lint passing (`make lint`)?

---

> [!NOTE]
> **Medium Risk**
> Incorrect resume ordering could miss files during interrupted scans; the change is narrow and covered by new regression tests.
> 
> **Overview**
> Fixes **filesystem scan resumption** skipping whole sibling directories when one directory name is a prefix of another (e.g. `blue-team` vs `blue-team-deprecated`).
> 
> `scanDir` used to treat a directory as already scanned when `path < resumeAfter` as **raw strings**. That order disagrees with `os.ReadDir` because `/` sorts after characters like `-` inside a name, so `blue-team-deprecated` could look “before” a resume point under `blue-team/` and never be visited.
> 
[...]

PR #5043, opened 2026-06-16
fix(azuresastoken): match SAS tokens regardless of parameter order

### LINKED ISSUE #4732: Azure sas storage tokens not detected
### TruffleHog Version: trufflehog 3.93.0

Noticed that some Azure Storage SAS tokens where not detected by TruffleHog

Example SAS url generated via Microsoft Storage Explorer that is not detected
```
https://accountname.blob.core.windows.net/sorted?sv=2023-01-03&st=2025-06-18T08%3A45%3A11Z&se=2025-06-19T08%3A45%3A11Z&sr=c&sp=r&sig=ow2a1XbXmD4%2BEv9LBUkek8R%2FrAjvrQFpenUbzztILn8%3D
```

It seems like this regex is a bit restrictive: https://github.com/trufflesecurity/trufflehog/blob/7635b24fd512a2e817dd3e9dd661caaf035a079d/pkg/detectors/azuresastoken/azuresastoken.go#L36

Tricky thou to build a regex that detects http query params that can be in any order. 

For example this regex works for my url, but would fail on one with sip set, other ordering of parameters etc.
```
'(sv=[0-9-]{10}&st=[0-9-%TAZ]{24}&se=[0-9-%TAZ]{24}&sr=[a-z]{1,10}&sp=[a-z]{1,10}&sig=(?:[A-Za-z0-9+\/=]|%3D|%2B|%2F){44})'
```

genisis0x:
### Description:

The `azuresastoken` detector only matched SAS tokens whose query parameters appeared in one fixed order (`sp`, `st`, `se`, `[sip]`, `[spr]`, `sv`, `sr`, `sig`) and whose timestamps used a literal `:`.

Real SAS tokens vary. Azure Storage Explorer, for example, emits the parameters in a different order (`sv` first, `sp` last) and URL-encodes the `:` in `st`/`se` as `%3A`:

```
https://accountname.blob.core.windows.net/sorted?sv=2023-01-03&st=2025-06-18T08%3A45%3A11Z&se=2025-06-19T08%3A45%3A11Z&sr=c&sp=r&sig=ow2a1XbXmD4%2BEv9LBUkek8R%2FrAjvrQFpenUbzztILn8%3D
```

Tokens like this were silently missed.

Instead of encoding one parameter order in a single regex, the detector now matches a contiguous run of query parameters and validates the SAS-specific parameters (`sp`, `sv`, `sr`, `sig`, and at least one of `st`/`se`, plus an optional `sip`) in code. This makes detection order-independent and tolerant of URL-encoded values, while preserving the validations the previous regex enforced (permission set, resource type, timestamp shape, and IP format) so the existing false-positive cases still produce no result.

Closes #4732

Tests:
- Added a case for the Storage-Explorer token above (alternate order + URL-encoded timestamps).
- Added a non-SAS query string case to confirm it is ignored.
- All existing positive and negative `TestAzureSASToken_Pattern` cases still pass.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped to secret-detection heuristics in one detector with added validation and tests; verification behavior is unchanged.
> 
> **Overview**
> The **Azure SAS token** detector no longer requires query parameters in a fixed order or literal `:` in timestamps. It now finds contiguous `key=value` runs with a looser regex, then confirms SAS shape in **`keyMatchIsSASToken`** (`sp`, `sv`, `sr`, `sig`, at least one of `st`/`se`, optional `sip`), including **URL-encoded** times (`%3A` / `%3a`) and signatures.
> 
> This fixes misses from tools like **Storage Explorer** (e.g. `sv` before `sp`, encoded colons). The regex avoids swallowing a preceding short `sas=` assignment so the real `sp` parameter is validated. Tests cover alternate ordering, encoded timestamps, the `sas=` case, and ignoring non-SAS query strings like `comp=list&restype=container`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac198e19aa5bfbb67ce1583c6e140ac16a63995b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/azuresastoken/azuresastoken.go
+++ b/pkg/detectors/azuresastoken/azuresastoken.go
@@ -37,7 +37,12 @@ var (
 	// signature as %2B, %2F, %3D). Rather than enumerate every permutation in one
 	// regex, match a contiguous run of query parameters and validate the SAS-specific
 	// parameters in keyMatchIsSASToken below. This keeps detection order-independent.
-	sasQueryPat = regexp.MustCompile(`[a-z]{2,4}=[^&\s"'<>]+(?:&[a-z]{2,4}=[^&\s"'<>]+)+`)
+	// The value class excludes `=` so a SAS token preceded by a short lowercase
+	// `key=` (e.g. `sas=sp=r&...`) is not absorbed from that preceding key, which
+	// would bury the real `sp` parameter inside the first value and fail
+	// validation. Azure URL-encodes any `=` inside a value (the signature padding
+	// becomes %3D), so excluding the literal does not drop a legitimate value.
+	sasQueryPat = regexp.MustCompile(`[a-z]{2,4}=[^&=\s"'<>]+(?:&[a-z]{2,4}=[^&=\s"'<>]+)+`)
 
 	// Validators for the individual SAS parameters. The `sp` permission set and `sr`
 	// resource set match the formats the previous, order-locked regex accepted.
--- a/pkg/detectors/azuresastoken/azuresastoken.go
+++ b/pkg/detectors/azuresastoken/azuresastoken.go
@@ -31,9 +31,22 @@ var (
 	// microsoft storage resource naming rules: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage:~:text=format%3A%0AVaultName_KeyName_KeyVersion.-,Microsoft.Storage,-Expand%20table
 	urlPat = regexp.MustCompile(`https://([a-zA-Z0-9][a-z0-9_-]{1,22}[a-zA-Z0-9])\.blob\.core\.windows\.net/[a-z0-9]([a-z0-9-]{1,61}[a-z0-9])?(?:/[a-zA-Z0-9._-]+)*`)
 
-	keyPat = regexp.MustCompile(
-		detectors.PrefixRegex([]string{"azure", "sas", "token", "blob", ".blob.core.windows.net"}) +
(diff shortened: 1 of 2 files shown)


genisis0x:
Addressed the Bugbot note in b718eca: `timeValuePat` now accepts both `%3A` and `%3a` (`%3[Aa]`) for the URL-encoded `:` in the start/expiry timestamps, since percent-encoding hex digits are case-insensitive per RFC 3986. Added a lowercase-hex test case alongside the existing uppercase one.

genisis0x:
Hi @mariduv, thanks again for merging #5041. Whenever you have a spare moment, would appreciate a look at this one, no rush. It closes #4732: Storage-Explorer SAS tokens were missed because the matcher required a fixed parameter order and literal-colon timestamps. The fix validates the SAS params order-independently and tolerates URL-encoded timestamps. CLA is signed and all checks are green. Happy to adjust anything.

genisis0x:
Addressed the Cursor Bugbot finding (sasQueryPat absorbing a preceding key) in ac198e19a.

The value class now excludes `=` (`[a-z]{2,4}=[^&=\s"'<>]+...`), so a SAS token preceded by a short lowercase `key=` (e.g. `sas=sp=r&...`, or `oken=` inside `token=`) no longer matches from that key and buries the real `sp` parameter in the first value. The parameter run now correctly starts at `sp`. Azure URL-encodes any `=` inside a value (the signature padding is `%3D`), so excluding the literal drops no legitimate value. Added a regression test for a token preceded by a lowercase assignment, which returns zero results on the old class and one with the fix. gofmt/vet/golangci-lint(0)/package tests all green.

mariduv (approved):

PR #5057, opened 2026-06-19
github: scan a user's public repos when --org names a non-owner

### LINKED ISSUE #4517: user (instead of org) with token is not working
### TruffleHog Version
trufflehog 3.90.11

### Expected Behavior

Considering this checks an user's repositories:
```sh
trufflehog github --org $user
```

And this checks an organization's repositories:
```sh
trufflehog github --org $org --token $GITHUB_TOKEN
```

This also should check an user's repositories, but now using a token:
```sh
trufflehog github --org $user --token $GITHUB_TOKEN
```

### Actual Behavior

When a user is passed to `--org` and `--token` is specified, the repositories of the token's owner are being analyzed (instead of the specified user).
 
### Steps to Reproduce

1. Run the following command using a user instead of an organization:
```sh
trufflehog github --org $user --token $GITHUB_TOKEN
```

2. Notice that the specified user' repositories will not be checked.

### Potential Solution
I have a suggestion that worked here:

1. Go to the function `getReposByOrgOrUser` inside [pkg/sources/github/repo.go](../tree/main/pkg/sources/github/repo.go);
2. Remove the parameter `authenticated`, and pass `false` to the function `getReposByUser()`;
3. Inside [pkg/sources/github/github.go](../tree/main/pkg/sources/github/github.go), update all the references that call `getReposByOrgOrUser`, removing the boolean parameter. 

I did not create a PR because I feel this solution may have collateral effects that I cannot understand now, having a very superficial context of the code.

ashvinctrl:
### Description:
Fixes #4517.

When a username is passed via `--org` together with a token, `enumerateWithToken` listed repositories through the authenticated `/user/repos` endpoint. That endpoint ignores the requested account and always returns the *token owner's* repositories, so `trufflehog github --org someuser --token ...` silently scanned the token owner instead of `someuser`.

The fix only uses the authenticated listing when the requested account matches the token owner; for any other user it falls back to the public `/users/{user}/repos` endpoint. This keeps the token owner's private repos in scope (the behaviour added in #4426) while scanning the right account for everyone else.

Scoped to the token enumeration path. Added two regression tests covering both branches: a different user resolves to the public endpoint, and the token owner still uses the authenticated one.

### Checklist:
* [x] Tests passing (`make test-community`)? — ran the affected package (`go test ./pkg/sources/github/...`), plus `gofmt` and `go vet`, all clean on Windows.
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))? — golangci-lint not installed locally; `gofmt`/`go vet` clean, relying on CI for the full lint.

---

> [!NOTE]
> **Medium Risk**
> Changes which repositories are enumerated for a common CLI pattern; wrong behavior previously caused silent mis-scans, but the fix may reduce private-repo coverage when scanning another user's account by name.
> 
> **Overview**
> Fixes incorrect scan targets when **`--org`** names a GitHub user that is not the token owner during token-based enumeration.
> 
> **`enumerateWithToken`** no longer always passes **`authenticated: true`** into **`getReposByOrgOrUser`**. It now sets **`authenticated`** only when the org/account name matches the token owner (case-insensitive), so **`/user/repos`** is used for the owner (including private repos, per #4426) and **`/users/{user}/repos`** is used for other users instead of silently enumerating the token owner's repos (#4517).
> 
> Adds two HTTP-mocked regression tests for the non-owner public path and the owner authenticated path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6b58e51b039a0b6905785c7baff385e18ca7dd2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/github/github.go
+++ b/pkg/sources/github/github.go
@@ -609,7 +609,13 @@ func (s *Source) enumerateWithToken(ctx context.Context, isGithubEnterprise bool
 	if len(s.orgsCache.Keys()) > 0 {
 		for _, org := range s.orgsCache.Keys() {
 			orgCtx := context.WithValue(ctx, "account", org)
-			userType, err := s.getReposByOrgOrUser(ctx, org, true, reporter)
+			// The authenticated /user/repos endpoint ignores the requested account
+			// and returns the token owner's repos, so only use it when org is the
+			// token owner. For any other user passed via --org, fall back to the
+			// public listing instead of silently scanning the token owner's repos.
+			// https://github.com/trufflesecurity/trufflehog/issues/4517
+			authenticated := strings.EqualFold(org, ghUser.GetLogin())
+			userType, err := s.getReposByOrgOrUser(ctx, org, authenticated, reporter)
 			if err != nil {
 				orgCtx.Logger().Error(err, "Unable to fetch repos for org or user")
 				continue
--- a/pkg/sources/github/github_test.go
+++ b/pkg/sources/github/github_test.go
@@ -624,6 +624,90 @@ func TestEnumerateWithToken(t *testing.T) {
 	assert.True(t, gock.IsDone())
 }
 
+// Passing a username via --org that isn't the token owner should list that
+// user's public repos via /users/{user}/repos, not the token owner's repos
+// via /user/repos. Regression test for issue #4517.
+func TestEnumerateWithToken_DifferentUserListsPublicRepos(t *testing.T) {
+	defer gock.Off()
+
+	// The token owner is "token-owner", but we scan "super-secret-user".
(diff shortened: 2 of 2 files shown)


ashvinctrl:
Bumping this one. All checks are passing and the CLA is signed, so I don't think it's blocked on anything from my end — looks like it just got missed.

If the needs-reconciliation label on #4517 means the behaviour is still open for discussion, I'd rather hear that than leave the PR sitting. Happy to rebase or narrow the scope if that helps.

PR #5063, opened 2026-06-22
feat: add Black Duck API token detector

### LINKED ISSUE #3012: Detection of Blackduck API tokens as secret
## Description
Can you add the secret detection criteria for Blackduck tool
https://www.synopsys.com/software-integrity/software-composition-analysis-tools/black-duck-sca.html
https://github.com/blackducksoftware

Blackduck Patterns
bash <(curl -s
https://detect.synopsys.com/detect9.sh)
--blackduck.url="
https://blackduck-server.com/"
--blackduck.api.token="XXXXXXXXXXXXXXXXXXXX"
Examples
text: |
    bash <(curl -s
https://detect.synopsys.com/detect9.sh)
--blackduck.url="
https://blackduck-server.com/"
--blackduck.api.token="MWI5YTAwNDItMmI1NC00NjgxLTlmM2YtNzM4ZDdjYjY5NjVhOjgzZWIxYWQ4LWM4OGUtNDJjMy1hYWFiLWE1MDc0MzU1NjBhMw=="

### Preferred Solution
Blackduck tokens are detected as secrets

New1Direction:
Adds a detector for Black Duck API tokens (the SCA tool). Closes #3012.

Black Duck tokens are base64 of `uuid:uuid` (100 chars ending in `==`). Matching just the base64 shape is noisy, so the detector decodes the candidate and only keeps it when it's a colon-separated pair of UUIDs.

Black Duck is self-hosted, so verifying needs the server URL. I grab it from the surrounding context and POST to `{server}/api/tokens/authenticate` with `Authorization: token <token>`. 200 = valid, 401/403 = invalid, anything else is left indeterminate.

Tests: `go test -tags=detectors ./pkg/detectors/blackduck/`

A couple of notes:
- I couldn't spin up a real Black Duck instance, so the verification path follows the documented token-auth API but hasn't been run against a live server.
- A token with no server URL nearby isn't reported, since it can't be verified without the endpoint (same approach as the portainertoken detector). Happy to also flag token-only matches if you'd prefer.

---

> [!NOTE]
> **Medium Risk**
> Verification performs outbound HTTP to user-supplied Black Duck hosts with discovered tokens; behavior follows existing self-hosted detector patterns but live verification was not validated against a real server.
> 
> **Overview**
> Adds a **Black Duck** secret scanner that finds base64 API tokens (decoded as `uuid:uuid`) near `blackduck` / `black_duck` keywords, pairs them with a co-located server URL, and registers `DetectorType_BlackDuck` (1053) in the default engine list.
> 
> Detection filters out raw base64 noise by requiring a valid UUID pair after decode. **Findings are only emitted when a server URL is present in the same chunk** (same pattern as Portainer token), with optional verification via `POST {url}/api/tokens/authenticate` and `Authorization: token <token>` (200 = verified, 401/403 = unverified).
> 
> Includes unit tests for token+URL matching, invalid token shape, token-only (no result), and `BLACK_DUCK_*` env-style prefixes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21c659fc18693ea1005d303ceb9de98d94ff216f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


New1Direction:
Thanks for the review — both findings were valid and are fixed in 21c659f:

- **`Keywords()` missing `black_duck`**: the regex accepts a `black_duck` prefix, but `Keywords()` only had `blackduck`, and `black_duck` isn't a substring of it — so the Aho-Corasick pre-filter would skip env-var style `BLACK_DUCK_TOKEN=...`. Added `black_duck` to the keyword set and a regression test that detects the underscore-only form against a host with no `blackduck` substring.
- **Swallowed request error in `verifyToken`**: now returns the `http.NewRequestWithContext` error instead of `nil`, so an indeterminate failure surfaces via `SetVerificationError` like the other detectors.

`go test -tags=detectors ./pkg/detectors/blackduck/` passes (4 cases).

PR #5072, opened 2026-06-24
[INS-465] Skip unverified JWT Detector results when feature flag is enabled

MuneebUllahKhan222:
# Summary

- Adds a `--drop-unverified-jwt-results` CLI flag that, when enabled, filters out JWT results that are definitively not live.
- The filter applies only when:
  - Verification was enabled (`verify=true`)
  - The JWT result is unverified (`Verified=false`)
  - No verification error is present (`VerificationError() == nil`)
- This ensures only conclusively invalid JWTs are dropped (expired tokens, rotated signing keys, invalid signatures), while JWTs that could not be verified due to environmental conditions (network failures, non-OIDC issuers, etc.) continue to be reported.
- The feature flag is backed by an `atomic.Bool` in `pkg/feature`, allowing staged per-customer rollout.

## Why This Filter Lives in the JWT Detector

An alternative implementation would have been to apply this filtering in the engine (`processResult`), where all detector results pass through.

However, that approach would require evaluating:

```go
!result.Verified && result.VerificationError() == nil
```

for every result produced by every detector, even though only JWT detectors can produce this specific condition.

By performing the filtering inside `jwt.FromData()`:

- The check only runs for JWT results.
- The engine hot path remains unchanged.
- Unrelated detectors incur no additional overhead.
- JWT-specific behavior remains encapsulated within the JWT detector.

## Meaning of "Unverified With No Error"

The JWT detector returns:

```go
Verified = false
VerificationError() = nil
```

when the token has been conclusively determined to be invalid.

Examples:

- Token is expired.
- Token's signing key has been rotated away (`kid` not found in JWKS).
- Signature verification fails against the issuer's public key.

The detector returns:

```go
Verified = false
VerificationError() != nil
```

when verification could not be completed and the token may still be valid.

Examples:

- Network failure while fetching metadata.
- OIDC discovery endpoint unavailable.
- Issuer does not support OIDC discovery.
- Other environmental verification failures.

These results are **not dropped**, since the token may still be live and the verification failure was inconclusive.

## Test Plan

- [x] `TestJwt_SkipUnverified` covers both feature-flag states:
  - With `drop-unverified-jwt-results=true`, a token that fails signature verification produces no results.
  - With `drop-unverified-jwt-results=false`, the same token is reported as before.

### Rollout Notes

This change implements **Stage 1** of the agreed JWT noise-reduction strategy:

- Unverified JWTs are dropped at scan time.
- Verified JWTs continue through the pipeline unchanged.

A future Stage 2 change will address filtering of verified-but-expired JWTs during API ingestion.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Opt-in flag changes which JWT findings are emitted when verification is enabled; incorrect filtering could hide or over-drop findings, but scope is limited to the JWT detector and only the unverified-without-error case.
> 
> **Overview**
> Adds an opt-in **`--drop-unverified-jwt-results`** CLI flag (documented in the man page) wired through **`feature.DropUnverifiedJWTResults`** in startup, matching other feature toggles.
> 
> When verification is on and the flag is set, the JWT detector **omits** findings that are **unverified with no verification error**—i.e. tokens judged definitively invalid (e.g. expired claims or failed signature checks) while **keeping** unverified results that still have a verification error (inconclusive failures such as network or OIDC issues). Filtering runs inside **`jwt.FromData`** so only JWT hits pay the cost.
> 
> **`TestJwt_SkipUnverified`** exercises flag on vs off with an expired EdDSA token under **`verify=true`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97d370db97961275eb83e2853ba854031199fa4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


unsmith (approved):
Seems solid to me.  I imagine this is something customers would intentionally turn on, so they are aware unverified JWT results are getting dropped.

main.go
```diff
@@ -512,6 +513,10 @@ func run(state overseer.State, logSync func() error) {
 		feature.UserAgentSuffix.Store(*userAgentSuffix)
 	}
 
+	if *dropUnverifiedJWTResults {
+		feature.DropUnverifiedJWTResults.Store(*dropUnverifiedJWTResults)
```
shahzadhaider1: Inside the `if`, value is always true, we could make it consistent with surrounding flag style

shahzadhaider1 (approved):

PR #5067, opened 2026-06-23
fix(detectors/datadogtoken): report verification endpoint failures

genisis0x:
### Description:

Closes #4541.

When the DatadogToken verifier could not reach the verification endpoint, the failure was silently dropped: the detector inlined the API call and only acted on the response inside an `if err == nil` branch, so a transport-level error (unreachable endpoint, DNS failure, timeout) left the result unverified with **no** verification error. Under `--results=verified,unknown` the secret was reported as `unknown` with no indication that verification had actually failed, unlike AWS (and unlike the sibling DatadogApiKey detector).

This change captures the request error and, when no endpoint verifies the token, surfaces it via `SetVerificationError`, so endpoint failures are reported as verification issues. The `DatadogApiKey` detector in this same package already follows this pattern, so this brings `DatadogToken` in line with it.

To keep the verifier testable the same way `DatadogApiKey` is, the scanner gains an optional injectable `client` (via `getClient()`); behavior is unchanged when it is nil.

### Checklist:
* [x] Tests passing (`make test-community`)? Added `TestDataDogToken_FromData_VerificationEndpointFailure`, which drives `FromData` with an error-returning transport and asserts a verification error is reported. It fails on the prior code (error swallowed) and passes with this fix.
* [x] Lint passing (`make lint`)? `golangci-lint run ./pkg/detectors/datadogtoken/...` reports 0 issues; `gofmt`/`go vet` clean.

---

> [!NOTE]
> **Low Risk**
> Localized detector verification and error-reporting behavior; no auth or data-path changes beyond clearer failure signaling.
> 
> **Overview**
> **DatadogToken** verification no longer drops transport failures (DNS, timeout, unreachable host). When `client.Do` fails across tried endpoints, the last error is recorded and, if nothing verifies, exposed via **`SetVerificationError`**—aligning with **DatadogApiKey** and other detectors so `--results=verified,unknown` can distinguish “couldn’t verify” from a plain unverified hit.
> 
> The scanner adds an optional injectable **`http.Client`** through **`getClient()`** (default behavior unchanged when nil), enabling the new **`TestDataDogToken_FromData_VerificationEndpointFailure`** test with a failing transport.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8062329ab13bf650950dd941c4ef1a708b4e4bda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5076, opened 2026-06-25
fix(detectors/hunter): surface indeterminate verification as a VerificationError

genisis0x:
## Problem
The Hunter verifier only set `Verified=true` on a 2xx response and silently ignored every other outcome. A transport failure (network blip), a `429` rate limit, or a `5xx` all left the result `Verified=false` with **no** `VerificationError`, making them indistinguishable from a definitively-rejected key. Under the default `--results=verified` filter a live key could be dropped with no signal that it was never actually checked.

This mirrors the verification-error convention already used across the detector suite (e.g. `SetVerificationError` on transport/indeterminate failures).

## Fix
Extract `verifyMatch` into a tri-state result:
- `2xx` -> verified;
- `401` / `403` -> definitively invalid (no error);
- transport failure or any other status (429, 5xx, ...) -> returned as an error and recorded via `SetVerificationError`.

Also adds an injectable client for testing and drains the response body before close.

## Tests
Added `TestHunter_VerificationError` covering the valid (2xx), invalid (401), rate-limited (429), and transport-error cases. The existing `TestHunter_Pattern` is unchanged.

`go test ./pkg/detectors/hunter/...`, `go vet`, `gofmt`, and `golangci-lint` are clean.

---

> [!NOTE]
> **Low Risk**
> Scoped to Hunter detector verification behavior and tests; aligns with existing detector conventions without changing auth or data paths.
> 
> **Overview**
> **Hunter API key verification** no longer treats every non-2xx or failed request as a quietly unverified finding. Verification is moved into `verifyMatch`, which returns a clear outcome: **2xx** marks the key verified; **401/403** mark it definitively invalid with no error; **transport failures, 429, 5xx, and other statuses** return an error and are recorded via **`SetVerificationError`**, matching other detectors.
> 
> The scanner gains an optional injectable **`http.Client`** (with a shared default) so tests can stub HTTP. Response bodies are drained before close.
> 
> **`TestHunter_VerificationError`** covers 2xx, 401, 429, and transport errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05ba949225ad03d751315f6b6d7e94d9362aaa59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/hunter/hunter.go
+++ b/pkg/detectors/hunter/hunter.go
@@ -2,6 +2,8 @@ package hunter
 
 import (
 	"context"
+	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
@@ -12,16 +14,25 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detector_typepb"
 )
 
-type Scanner struct{}
+type Scanner struct {
+	client *http.Client
+}
 
 // Ensure the Scanner satisfies the interface at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
 var (
-	client = common.SaneHttpClient()
+	defaultClient = common.SaneHttpClient()
 
 	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"hunter"}) + `\b([a-z0-9_-]{40})\b`)
 )
+
+func (s Scanner) getClient() *http.Client {
+	if s.client != nil {
+		return s.client
+	}
+	return defaultClient
+}
 
 // Keywords are used for efficiently pre-filtering chunks.
 // Use identifiers in the secret preferably, or the provider name.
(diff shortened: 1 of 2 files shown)


genisis0x:
@shahzadhaider1 gentle nudge on this one whenever you have a window (and thanks for the look on #5067).

It's small and self-contained: the Hunter verifier only sets a result on a `2xx` and silently swallows everything else, so a transport failure, a `429`, or a `5xx` all come back `Verified=false` with no `VerificationError` — indistinguishable from a definitively rejected key. Under the default `--results=verified` that means a live key can be dropped with no signal it was never actually checked. The change routes those non-2xx/transport outcomes into a `VerificationError`, matching the convention the other detectors already follow, with a test covering each case. Green and CLA signed.

Happy to reshape it if you'd prefer a different approach.

PR #5084, opened 2026-06-28
fix(detectors/jdbc): do not dial local hosts during verification

### LINKED ISSUE #3679: JDBC detector should not connect to local hosts
The URI detector does not connect to local IP addresses ("dialing local IP addresses is not allowed"), and the same should be applied to the JDBC and other database detectors.
```
Found unverified result 🐷🔑❓
Verification issue: dial tcp [::1]:3306: connect: connection refused
Detector Type: JDBC
Decoder Type: PLAIN
Raw result: jdbc:mysql://localhost/authority
Commit: 68b80fa14f8c2ece71df20ac0a3e3e710bbcd368
File: webgoat-5.4/doc/attic/Eclipse-Workspace.zip
Line: 144
Link: https://github.com/attiasas/WebGoat/blob/68b80fa14f8c2ece71df20ac0a3e3e710bbcd368/webgoat-5.4/doc/attic/Eclipse-Workspace.zip#L144
Repository: https://github.com/attiasas/WebGoat.git
Timestamp: 2012-09-09 21:53:44 +0000
```

Ideally this wouldn't be logged as an error every time.

arpitjain099:
Fixes #3679. The JDBC detector opens real database connections to local hosts during verification, so scanning something like `jdbc:postgresql://127.0.0.1:5432/db` or `jdbc:mysql://user:pass@localhost:3306/db` makes TruffleHog dial loopback/private addresses. The URI detector already refuses this through DetectorHttpClientWithNoLocalAddresses; this brings JDBC to the same behavior. It's the noisy verification error from the issue and also a small SSRF-adjacent surface, since scanned content drives the outbound connection.

The change adds a local-host guard in the jdbc package that mirrors the isLocalIP check in pkg/detectors/http.go (loopback, link-local, private, unspecified), and bails out before any dial for the postgres, mysql, and sqlserver sub-protocols. It strips the MySQL `tcp(...)` wrapper and optional port and resolves names with net.LookupIP so `localhost` is caught too. Detection is unchanged; the finding just comes back unverified instead of dialing.

Tests cover host classification and a per-protocol refusal for each of the three sub-protocols. `go test ./pkg/detectors/jdbc/` and `go vet` pass.

Two notes for review:
- I kept this to JDBC to keep it small. The issue mentions other database detectors, and I'm happy to extend the same guard in a follow-up.
- I duplicated a small isLocalIP helper in the jdbc package rather than reach for the unexported one in pkg/detectors. If you'd rather I export and reuse it, just say so.

I've signed the CLA. Thanks for taking a look.

---

> [!NOTE]
> **Medium Risk**
> Changes outbound connection behavior during secret verification (SSRF-adjacent hardening); scoped to JDBC ping paths with tests, but DNS resolution for hostnames is new behavior on the verification path.
> 
> **Overview**
> Stops JDBC **verification** from opening real DB connections to loopback, private, link-local, or unspecified hosts—aligning with the URI detector’s local-address refusal (fixes #3679).
> 
> Adds `isLocalIP` / `hostIsLocal` in the JDBC package (with MySQL `tcp(...)`, port, and SQL Server `host\INSTANCE` normalization plus DNS lookup for names like `localhost`). **Postgres, MySQL, and SQL Server** `ping` paths now return a determinate `detectors.ErrNoLocalIP` before `sql.Open`/`Ping`; detection and redaction are unchanged, so local URLs still surface as **unverified** findings.
> 
> New tests cover host classification and per-protocol refusal for named-instance SQL Server cases that previously bypassed the check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6aab4bacf12583def1a2dd55e1faf7a9c2607736. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/jdbc/jdbc.go
+++ b/pkg/detectors/jdbc/jdbc.go
@@ -5,7 +5,9 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"net"
 	"regexp"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


arpitjain099:
Both taken, and the named-instance one was a real hole rather than a theoretical one. Measured against the current branch before changing anything:

```
hostIsLocal("localhost")                = true
hostIsLocal("127.0.0.1")                = true
hostIsLocal("localhost\SQLEXPRESS")     = false
hostIsLocal("127.0.0.1\SQLEXPRESS")     = false
hostIsLocal("localhost\SQLEXPRESS:1433") = false
```

So the guard was fully bypassed for any SQL Server named instance, including an IP-literal one, and verification would have dialled the local address. `hostIsLocal` now strips the `\INSTANCE` suffix after the port split, and `TestHostIsLocal` covers `localhost`, `127.0.0.1`, `10.0.0.5` and a `:1433` variant, plus `db.example.com\SQLEXPRESS` staying false so the strip does not over-block. `TestJdbcPingRefusesLocalHosts` got two named-instance pingers as well. Removing the strip fails five cases.

On the second point, the `detectors.ErrNoLocalIP` half is right and is done: mysql, postgres and sqlserver all return the exported sentinel now, so an `errors.Is` against it matches the JDBC path too.

I left `isLocalIP` duplicated though, and I do not think it can be removed: the one in `pkg/detectors/http.go` is lowercase, so it is not reachable from this package without exporting it or moving it somewhere shared. That felt like a bigger change than belongs in this PR, but happy to do it if you would prefer that.

PR #5102, opened 2026-07-01
fix(sentryorgtoken): match variable-length org auth tokens

Synvoya:
## Summary

The `sentryorgtoken` detector hardcodes exactly 197 characters after `sntrys_eyJ`, so it silently misses most real Sentry organization auth tokens — a false negative on valid secrets.

## Root cause

A Sentry org auth token is `sntrys_` + base64(JSON payload) + `_` + secret, where the JSON encodes `iat`, `url`, `region_url`, and `org`. The org slug and the url/region_url vary per organization and region, so the base64 payload — and therefore the total token length — is not fixed. The pattern `\b(sntrys_eyJ[a-zA-Z0-9=_+/]{197})\b` only matches tokens that happen to total 197 characters. The single existing fixture (org slug `truffle-security-jd`) coincidentally hits 197, so the gap was never caught, and the detector has no fallback pattern — so any other length is dropped.

Using the real token structure: an org slug of `acme` yields 177 characters after `sntrys_eyJ`, and a longer slug yields 225 — both valid tokens the old pattern misses.

## Fix

Widen the quantifier to `{130,300}` to bracket real payload/secret length variance. The highly specific `sntrys_eyJ` prefix keeps false positives negligible.

## Tests

Added short- and long-length regression cases to `TestSentryToken_Pattern` (built at runtime so the fixtures don't trip push-protection secret scanning). Both fail against the old `{197}` pattern and pass with the fix. `go test ./pkg/detectors/sentryorgtoken/` passes; `gofmt` and `go vet` clean.

---

> [!NOTE]
> **Low Risk**
> Scoped regex and test changes in one detector; no auth or verification logic changes, with low false-positive risk due to the specific `sntrys_eyJ` prefix.
> 
> **Overview**
> Fixes a **false-negative** in the Sentry org auth token detector: the regex required exactly **197** characters after `sntrys_eyJ`, so most real tokens (length varies with org slug and region URLs) were never reported.
> 
> The pattern now allows **`{130,300}`** characters after the prefix, with a short comment explaining why length is not fixed. **`TestSentryToken_Pattern`** adds runtime-built short (177) and long (225) fixtures that the old rule missed, without embedding secrets that would trigger push protection.
> 
> Verification (`verifySentryOrgToken`) is unchanged; only extraction breadth improves.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 899baed10d68aa71f1655f8f8a3af1977ab07652. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/sentryorgtoken/sentryorgtoken.go
+++ b/pkg/detectors/sentryorgtoken/sentryorgtoken.go
@@ -22,7 +22,11 @@ var _ detectors.Detector = (*Scanner)(nil)
 
 var (
 	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	orgAuthTokenPat = regexp.MustCompile(`\b(sntrys_eyJ[a-zA-Z0-9=_+/]{197})\b`)
+	// The base64 payload encodes the org slug plus url/region_url, which vary
+	// per organization/region, so the token length is not fixed. Match a range
+	// rather than the single hardcoded length that only matched tokens which
+	// happened to total 197 characters.
+	orgAuthTokenPat = regexp.MustCompile(`\b(sntrys_eyJ[a-zA-Z0-9=_+/]{130,300})\b`)
 )
 
 // Keywords are used for efficiently pre-filtering chunks.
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
LGTM. Expanding the regex to support a bounded variable-length payload fixes the false negatives caused by the previous fixed-length pattern while keeping matching constrained. The added tests for both shorter and longer organization tokens provide good coverage for the regression.

PR #4995, opened 2026-05-29
[INS-197] Add redhatpyxis api key detector

MuneebUllahKhan222:
## Description
This PR adds the Red Hat Pyxis API Key Detector for TruffleHog.

It scans for Red Hat Pyxis API keys and optionally verifies them via the official Red Hat Ecosystem Catalog API.

Regex:

```regex
\b([a-z0-9]{32})\b
```

The detector uses contextual keyword `redhat` to reduce false positives.

## Verification
For verification, we use the Red Hat Pyxis Projects API:

```text
https://catalog.redhat.com/api/containers/v1/projects/certification/requests/images?page_size=1&page=0
```

We send a `GET` request with the token in the `X-API-KEY` header.

A response code of:
- `200 OK` means the API key is valid
- `401 Unauthorized` means the API key is invalid.

This API endpoint is part of the official Red Hat Ecosystem Catalog API and can be used safely for verification. It is a read-only endpoint and does not perform any destructive actions.

### Checklist:
* [x] Tests passing (`make test-community`)?
* [x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

---

> [!NOTE]
> **Low Risk**
> Additive detector-only change following existing patterns; verification uses a read-only external API and is gated like other new detectors.
> 
> **Overview**
> Adds a **Red Hat Pyxis** secret detector so scans can find 32-character lowercase hex API keys when `redhat` or `pyxis` appears nearby (`PrefixRegex` + `\b([a-z0-9]{32})\b`).
> 
> Optional verification issues a read-only `GET` to the Red Hat Ecosystem Catalog certification images endpoint with `X-API-KEY`, treating **200** as valid and **401** as invalid (other statuses surface as verification errors). Wiring matches other gated detectors: proto `RedHatPyxis = 1055`, `redhatpyxis.Scanner` in default detector list, `RedHatPyxisDetectorEnabled` feature flag (enabled in `main.go`), and unit/integration/pattern tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 869474554d2a90c100e1e7c62ee61d3128cd12be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5103, opened 2026-07-01
feat(detectors): add Resend detector

mwoss:
### Description:
Hi Trufflehog team! I work at Resend (the email API this detector targets), submitting this on behalf of the team. Happy to be the point of contact for the key format and verification behavior, and I can provide sanctioned test credentials (`RESEND_API_KEY`, `RESEND_API_KEY_SENDING_ACCESS`, `RESEND_API_KEY_INACTIVE`) for trufflehog-testing/detectors5 via any secure channel you prefer.

Key format: `re_[1-9A-HJ-NP-Za-km-z]{8}_[1-9A-HJ-NP-Za-km-z]{24}` (base58 alphabet, excluding the ambiguous characters 0, O, I, and l).

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Verification sends discovered keys to Resend’s API (standard for TruffleHog); misclassified tokens or API behavior changes could affect verification accuracy, but scope is limited to a new feature-flagged detector.
> 
> **Overview**
> Adds a **Resend** detector so TruffleHog can find and verify Resend email API keys in scanned content.
> 
> Detection uses a strict `re_` + base58 pattern (excluding ambiguous `0/O/I/l`). When verification is on, it issues a read-only **GET** to `https://api.resend.com/api-keys`: **200** marks **full_access** keys verified; **401** with error name `restricted_api_key` still verifies **sending_access** keys. Results include a rotation guide URL and permission in extra data.
> 
> The scanner is registered in default detectors behind **`ResendDetectorEnabled`** (on by default in OSS `main.go`), with proto type **`Resend = 1071`**. Unit, pattern, and integration tests cover matching, deduplication, and live verification paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4a1d33c8c75be3ab2e116d0c0b2fb59a3341450. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mwoss:
Closes https://github.com/trufflesecurity/trufflehog/issues/5107

rajpratham1 (approved):
LGTM. The detector follows the existing TruffleHog detector pattern, uses a well-constrained key format, implements sensible verification logic for both full_access and sending_access API keys, and includes comprehensive unit and integration tests. I don't see any issues with this implementation.

mwoss:
@kashifkhan0771 @MuneebUllahKhan222 Could you take a look in your free time? :D

pkg/detectors/resend/resend.go
```diff
@@ -0,0 +1,149 @@
...
+	case http.StatusUnauthorized:
+		// Resend returns 401 for two distinct cases. We must look at the
+		// error name to tell them apart:
+		//   - restricted_api_key: active `sending_access` key trying to
+		//     reach a non-send route. The key is real and can send mail.
+		//   - missing_api_key:    no Authorization header (should not
+		//     happen here since we set one above).
+		var body resendError
+		if err := json.NewDecoder(res.Body).Decode(&body); err != nil {
+			return false, nil, nil
```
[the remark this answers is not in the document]
mwoss: Good point. If we can't read the 401 body we can't really tell if the key is valid, so returning an error there makes more sense. Gonna fix it.

pkg/detectors/resend/resend.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Resend API keys: prefix `re_`, 8 chars of base58,
+	// underscore, then 24 chars from the same alphabet.
+	keyPat = regexp.MustCompile(`\b(re_[1-9A-HJ-NP-Za-km-z]{8}_[1-9A-HJ-NP-Za-km-z]{24})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"re_"}
```
shahzadhaider1, line 44: How do you feel about replacing `re_` with `resend`? I’m asking because `re_` seems to be quite prone to false positives.

shahzadhaider1 (commented):
This looks good, thank you for the contribution. one small feedback.

Could you also introduce a feature flag for this detector? Examples: [redhatpyxis](https://github.com/trufflesecurity/trufflehog/pull/4995), [SolarWinds](https://github.com/trufflesecurity/trufflehog/pull/5199)

pkg/detectors/resend/resend.go
mwoss: That's true, but our API keys are currently constructed with `re` + `short token` + `long_token` chunks. We would have to change the schema of API keys then to support `resend`. Or am I misunderstanding the purpose of the ' Keywords' method? 
If we change it won't match lines like `apiKey = "re_12345678_aaaaaaaaaaaaaaaaaaaaaaaa"` yes?

shahzadhaider1: No, I’m not suggesting changing the API key schema. The `Keywords` function looks for the specified keywords anywhere in the input. So with `re_`, the detector could be invoked for things like `pre_commit`, `azure_config`, or anything else containing `re_`. Since `re_` is fairly common, this could cause the detector to run unnecessarily often.

What I was suggesting is adding another keyword that is commonly found near Resend API keys but is more specific/less common, so we can reduce unnecessary detector invocations.

And yes, if we replace `re_` with `resend`, it wouldn’t match:

`apiKey = "re_12345678_aaaaaaaaaaaaaaaaaaaaaaaa"`

but it would match something like:

`resend_apiKey = "re_12345678_aaaaaaaaaaaaaaaaaaaaaaaa"`

assuming `resend` is used as the keyword.

mwoss: Oh, I got it now. Thanks! It makes sense, but I'm scared that if we use `resend` we gonna miss cases like the above, but when we gonna use `resend` + `re_` to match them all, we would still hit performance issues, as keywords are or-matched (if I'm not wrong).

(Sorry for the late reply, I was on short PTO :3)

shahzadhaider1: Yeah, `re_` is a bit scary. We’ve had similar cases before where overly broad keywords ended up causing a lot of unnecessary processing.

mwoss: But when using `resend` as a keyword, wouldn't we miss a lot of potential keys? Like the one mentioned by you above?

mwoss: What do you think @shahzadhaider1? :D I just think that when we move to `resend` as a keyword, checks won't be as effective. Maybe it's worth trade off, but I would deffer to you, as you have more experienec in this space :D

shahzadhaider1 (approved):

mwoss:
@shahzadhaider1 is it mergable? :D

shahzadhaider1:
> @shahzadhaider1 is it mergable? :D

I believe so, yes. We need one more maintainer's approval.

mariduv (approved):

PR #5105, opened 2026-07-02
feat(detectors): add Arweave wallet private key detector

### LINKED ISSUE #4643: Arweave wallet private keys
TruffleHog currently lacks a detector for Arweave wallet private keys in JWK (JSON Web Key) format. Arweave is a decentralized storage network with its own cryptocurrency (AR). Leaked private keys can result in theft of AR tokens and unauthorized access to stored data.

  Arweave wallets use 4096-bit RSA keys stored in JWK format, which differs from the PEM format that existing RSA detectors target.

  Preferred Solution

  Add a detector that identifies Arweave JWK private keys by looking for JSON objects containing the RSA-JWK structure with these required fields:

  {
      "kty": "RSA",
      "e": "AQAB",
      "n": "<base64url-encoded-modulus>",
      "d": "<base64url-encoded-private-exponent>",
      "p": "<base64url-encoded-prime1>",
      "q": "<base64url-encoded-prime2>",
      "dp": "<base64url-encoded-exponent1>",
      "dq": "<base64url-encoded-exponent2>",
      "qi": "<base64url-encoded-coefficient>"
  }

  Detection heuristics:
  - kty equals "RSA"
  - Presence of private key components: d, p, q, dp, dq, qi
  - The n (modulus) field is ~683 characters when base64url-encoded (indicating 4096-bit key)
  - e is typically "AQAB" (base64url for 65537)

  Verification: The detector could verify the key by:
  1. Deriving the wallet address from the public components (n and e)
  2. Querying https://arweave.net/wallet/<address>/balance to check if the wallet exists

  Additional Context

  Example structure (NOT a real key):
  {
      "kty": "RSA",
      "e": "AQAB",
      "n": "0vx7agoebGcQSuu...<~683 chars total>...HKoZIzj0DAQcDQgAE",
      "d": "X4cTteJY_gn4Fgn...<~683 chars total>...ZDrQRIhTnCFhWP",
      "p": "83i-7IvMGXoMXC...<~341 chars total>...Uwthj0Iw",
      "q": "3dfOR9cuYq-0S-m...<~341 chars total>...8rBYnzTIOQw",
      "dp": "G4sPXkc6Ya9y8o...<~341 chars total>...dpCexxQvdcw",
      "dq": "s9lAH9fggBsoFR...<~341 chars total>...DOGhtASj1kcw",
      "qi": "rCz5MhkRZxOo...<~341 chars total>...2k6xqo7c"
  }

  How to generate a test key:

  # Using arweave npm package
  npm install arweave

  # Generate wallet
  node -e "
  const Arweave = require('arweave');
  const arweave = Arweave.init({});
  arweave.wallets.generate().then(key => console.log(JSON.stringify(key, null, 2)));
  "

  Or use the https://arweave.app or https://github.com/ArweaveTeam/arweave-js.

  Common locations where these keys appear:
  - arweave-keyfile-<address>.json files
  - Environment variables (e.g., ARWEAVE_PRIVATE_KEY, AR_WALLET)
  - CI/CD configs for Arweave deployment tools (Bundlr, Irys, ArDrive)

  References

  - Arweave Docs - Wallets: https://docs.arweave.org/developers/wallets
  - arweave-js SDK: https://github.com/ArweaveTeam/arweave-js
  - JWK RFC 7517: https://datatracker.ietf.org/doc/html/rfc7517
  - Related: Bundlr/Irys uses the same key format for Arweave funding

momomuchu:
## Description

Closes #4643.

Adds a detector for **Arweave wallet private keys**. An Arweave wallet file is an RSA JSON Web Key (JWK): a JSON object with `"kty":"RSA"` plus a `"d"` field holding the RSA private exponent in base64url. Leaking this key gives full control of the wallet and all AR tokens it holds.

### Detection rationale
- Requires **both** the `"kty":"RSA"` marker and a long (≥100 char) base64url `"d"` value. The `d` (private exponent) is absent from public keys, so this combination is high-signal.
- Rejects RSA public keys (no `d`), EC/oct JWKs (`kty` != RSA), and short `d` false positives.
- No public Arweave endpoint validates a raw JWK, so results are `unverified`, consistent with other key-material detectors (e.g. `privatekey`).

### Test coverage
8 pattern cases: 3 positive (compact JWK, spaced JWK, arweave-keyword JWK) and 5 negative (EC key, RSA public key, short `d`, no keyword, wrong `kty`). All synthetic/fake key material, no real secrets.

### Note on generated protobuf
`proto/detector_type.proto` is updated with the new `ArweaveKey` enum, and the generated `detector_type.pb.go` carries the matching constant + name/value map entries so the package builds and the enum resolves. The embedded descriptor blob was **not** regenerated (the `make protos` toolchain runs in a pinned Docker image unavailable in my environment). Running `make protos` before merge will regenerate it cleanly from the updated proto source.

### Checklist
- [x] `go test ./pkg/detectors/arweavekey/...` passing
- [x] `go vet` / `gofmt` clean
- [x] Registered in `pkg/engine/defaults`

---

> [!NOTE]
> **Low Risk**
> Additive secret scanner with no auth or runtime behavior changes; main review focus is false-positive/negative tradeoffs on generic RSA JWKs and confirming protobuf regeneration before merge.
> 
> **Overview**
> Adds an **Arweave wallet private key** detector that flags RSA JWKs with a long base64url `"d"` private exponent, wired into default engine scans as `DetectorType_ArweaveKey`.
> 
> Matching requires `"kty":"RSA"` plus a `"d"` value of at least 100 characters. To avoid labeling generic JWT/Auth0-style 2048-bit RSA JWKs as Arweave, findings also need either an **arweave** mention in the chunk or a `"d"` field at least **600** characters (4096-bit wallet sizing). Keywords `kty` and `arweave` drive the Aho–Corasick prefilter; results are **unverified** (no validation API). **`MaxSecretSize` 4096** extends the engine window so full keyfiles are not truncated before `"d"`.
> 
> Includes pattern and engine span-extraction tests, `proto/detector_type.proto` / generated enum **1062**, and registration in `pkg/engine/defaults`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1d5e9669c2ba53a1440da12067daa639e8e7cfc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rajpratham1 (approved):
Thanks for the contribution! The detector follows the existing detector pattern well, uses a reasonably strict signature (kty=RSA plus a long d field), includes good positive/negative test coverage, and documents the lack of online verification. I don't see any issues with the implementation. LGTM.

rajpratham1 (changes requested):
I agree with the Cursor review. This PR appears functionally incomplete because the detector likely won't work in the production scanning pipeline.

Main issue

The detector only implements:

type Scanner struct{}

and

but does not implement detectors.MaxSecretSizeProvider.

Production scanning does not call FromData() with the entire file. Instead, the engine extracts a window around the matched keyword ("kty" or "arweave"). If no custom maximum secret size is provided, that window defaults to roughly 512 bytes.

An actual Arweave RSA JWK is much larger:

full JWK: ~1.4–3 KB+
"d" field often hundreds of base64url characters
"d" commonly appears well after "kty"

So the engine may pass only:

{
  "kty":"RSA",
  "e":"AQAB",
  "n":"...."

without the complete "d":"..." value.

Since detection depends on

dFieldPat.FindAllStringSubmatch(...)

the detector simply returns no results.

Why the tests don't catch it

The tests call:

d.FromData(...)

directly with the complete JSON.

That bypasses the production span/window logic entirely, so they validate only the regex—not integration with the engine.

What should be added

This detector should implement MaxSecretSizeProvider, returning a window large enough to encompass a full Arweave JWK, similar to other large-secret detectors already in the repository (e.g. private key detectors).

Ideally, add an integration-style test exercising the detector through the engine's span extraction rather than only calling FromData() directly.

What's good
Regex logic is straightforward.
Deduplication is handled.
Detector registration is complete.
Documentation/comments are clear.

momomuchu:
Addressed both findings. Regenerated the detector proto descriptor so DetectorType_ArweaveKey.String() returns the right name, and implemented MaxSecretSizeProvider (4096) with a test that exercises the engine span-extraction path (not just FromData on the full JSON) to prove a full Arweave JWK isn't truncated before the d field. Ready for another look.

momomuchu:
Fixed. The detector now only reports an Arweave match when the d field is sized like a 4096-bit Arweave key (about 683 base64url chars) or the text explicitly mentions arweave, so a generic 2048-bit RSA JWK (JWT/Auth0/Azure-style) no longer gets mislabeled. Added a regression test for the generic-JWK case.

PR #5106, opened 2026-07-02
feat(detectors): add RSA SecurID software token detector

### LINKED ISSUE #4395: Add support for RSA SecurID MFA software tokens
## Description
Add support for RSA SecurID virtual MFA tokens
https://www.rsa.com/products/securid/

### Preferred Solution
Expose serial number & seed of the virtual token. The seed allows to duplicate the token, then bypass MFA, thus proves the leak. The serial number allows an admin to invalidate the token.

### Additional Context
Custom detection logic:
```
detectors:
- name: RSASecurIDDetector
  keywords:
  - TKNHeader
  regex:
    SerialNumber: |-
      <SN>\d+?</SN>
    Seed: |-
      <Seed>.+?</Seed>
  primary_regex_name: SerialNumber
```

#### References
Example tokens available at https://github.com/stoken-dev/stoken

momomuchu:
## Description

Closes #4395.

Adds a detector for **RSA SecurID software tokens** distributed as `.sdtid` XML files. Each token export contains a `<SN>` serial (12 digits) and a `<Seed>` TOTP seed (128-bit hex, or an AES-encrypted seed in base64). The seed lets an attacker generate valid one-time passcodes and defeat the MFA factor, so both parts are reported as a multi-part credential (serial + seed).

### Detection rationale
- `<SN>(\d{12})</SN>` **and** `<Seed>([A-Za-z0-9+/]{16,}={0,2})</Seed>` must both be present.
- A structural guard requires an RSA SecurID marker (`<TKNBatch>`, `<TKNBasic>`, or `.sdtid`) before emitting, so generic XML with unrelated `<SN>`/`<Seed>` elements is not flagged. This is the primary false-positive defense.
- No public endpoint can validate a software token, so results are `unverified`.

### Test coverage
8 pattern cases: 2 positive (hex seed, base64/encrypted seed) and 6 negative (SN only, Seed only, wrong element names, short seed, wrong-length serial, missing structural marker). All synthetic/fake values, no real token material.

### Note on generated protobuf
`proto/detector_type.proto` is updated with the new `RSASecurID` enum, and the generated `detector_type.pb.go` carries the matching constant + name/value map entries so the package builds and the enum resolves. The embedded descriptor blob was **not** regenerated (the `make protos` toolchain runs in a pinned Docker image unavailable in my environment). Running `make protos` before merge will regenerate it cleanly from the updated proto source.

### Checklist
- [x] `go test ./pkg/detectors/rsasecurid/...` passing
- [x] `go vet` / `gofmt` clean
- [x] Registered in `pkg/engine/defaults`

---

> [!NOTE]
> **Low Risk**
> Additive secret scanner with structural guards and no verification or auth changes; main risk is false positives on SecurID-shaped XML in repos.
> 
> **Overview**
> Adds a new **RSA SecurID** detector for `.sdtid` XML exports that pairs a 12-digit `<SN>` with a `<Seed>` (hex or base64) and reports them as a multi-part credential (`serial` + `seed`, `RawV2` as `serial:seed`). Detection requires SecurID structure markers (`<TKNBatch>`, `<TKNBasic>`, or `.sdtid`) so unrelated XML is ignored; there is no remote verification, so findings stay **unverified**.
> 
> `Keywords()` includes `TKNBatch`, `TKNBasic`, and `sdtid` so the Aho-Corasick prefilter does not skip valid blobs. Tests cover keyword/context alignment, a `<TKNBasic>`-only end-to-end path, and positive/negative pattern cases.
> 
> The scanner is wired into default detectors and `DetectorType_RSASecurID` (1058) is added in `proto/detector_type.proto` with matching generated Go enum/maps (descriptor blob updated in the diff).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a6d7d9b04753cfde8d21977ed1820b9ede59c67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rajpratham1 (changes requested):
Good addition overall, but I think the Cursor finding is valid. contextPat recognizes both <TKNBatch> and <TKNBasic>, whereas Keywords() only returns "TKNBatch" and "sdtid". Since the Aho-Corasick prefilter gates execution of FromData, a valid token containing only <TKNBasic> may never be scanned. Could you add "TKNBasic" to Keywords() (or otherwise align the prefilter with contextPat) so all supported formats are reachable?

momomuchu:
Addressed: `Keywords()` now returns `["TKNBatch", "TKNBasic", "sdtid"]` so a `<TKNBasic>`-only token survives the Aho-Corasick prefilter, plus a test that scans a TKNBasic-only blob through the real prefilter path. Ready for another look.

momomuchu:
Good catch. Fixed: `RSASecurID` now uses `1058`, the next free value on `main` after `OpenRouter = 1057`. Regenerated `detector_type.pb.go` via `make protos` so the constant, the name/value maps, and the binary descriptor all match (that also clears the earlier stale-descriptor `String()` issue). Note the sibling PR #5105 also targets 1058; whichever merges second I'll rebase-bump by one.

PR #5125, opened 2026-07-12
Add bearer token auth support for OpenSearch clusters in Elasticsearch source

### LINKED ISSUE #5124: Support scanning OpenSearch clusters via the Elasticsearch source
## Description

The elasticsearch source can't currently scan OpenSearch clusters, even though OpenSearch
shares Elasticsearch's REST API surface and is a common target.
Pointing the Trufflehog with the latest cersion at an OpenSearch cluster fails for several reasons at once:

1. OpenSearch's JWT auth plugin expects a raw Authorization: Bearer <jwt>
   header. None of the existing auth options (basic, API key, service token) produce that.
2. go-elasticsearch client rejects any response
   missing the `X-Elastic-Product` header. Real Elasticsearch always sends it while OpenSearch
   doesn't reliably send it
3. OpenSearch exposes PIT creation at
   `POST /{index}/_search/point_in_time`, not Elasticsearch's `POST /{index}/_pit`, so the
   vendored esapi.OpenPointInTimeRequest (hardcoded to the ES8 path) 404s.
4. Elasticsearch's PIT tiebreaker sort field is `_shard_doc`;
   OpenSearch doesn't recognize it and rejects the search with a `query_shard_exception`,
   but accepts the equivalent `_doc`.
5. When no `--query-json`/`--since-timestamp` is set, the search
   body's `query` field is `{}`. Elasticsearch tolerates this as an implicit match-all;
   OpenSearch's stricter parser rejects it with `illegal_argument_exception: query
   malformed, empty clause found`, silently zeroing out every scan.

### Preferred Solution

A new `--bearer-token` flag / `ELASTICSEARCH_BEARER_TOKEN` env var on the existing
`elasticsearch` command that:
- Sets the `Authorization: Bearer` header directly for JWT-based auth.
- Wraps the HTTP transport to inject `X-Elastic-Product` on successful responses when
  bearer-token auth is used, so the client's product check passes against OpenSearch.
- Uses an OpenSearch-specific PIT-open call and `_doc` tiebreaker when bearer-token auth
  indicates an OpenSearch target.
- Sends an explicit `match_all` instead of an empty query clause (valid and
  behavior-preserving on real Elasticsearch too, so this fix isn't even OpenSearch-gated).

All of the above only activate when `--bearer-token` is set, so existing Elasticsearch
behavior (basic/API-key/service-token auth) is unchanged.

#### References

(none yet. Will link the PR here once opened)

anon767:
Fixes #5124

### Summary
OpenSearch clusters share TruffleHog's Elasticsearch API surface but aren't fully compatible with the existing `elasticsearch` source, so scanning an OpenSearch cluster currently fails. This adds a `--bearer-token` flag (`ELASTICSEARCH_BEARER_TOKEN` env var) that works around five incompatibilities (see linked issue), all gated behind bearer-token auth being set so existing Elasticsearch behavior (basic/API-key/service-token) is unchanged.

### Test plan
- [x] `go build ./...` passes
- [x] Existing Elasticsearch integration test updated for the new `isOpenSearch` parameter (passed `false`, no behavior change)
- [x] Manually verified against a real OpenSearch 2.17.0 cluster with the JWT auth plugin enabled: minted a JWT signed with the cluster's configured HMAC key, confirmed `--bearer-token` authenticates, PIT opens via the OpenSearch-specific endpoint, the `_doc` tiebreaker sort works, and documents are correctly fetched and chunked for scanning. Seeded a document containing a JWT-shaped test secret and confirmed TruffleHog's own JWT detector correctly fired on it end-to-end through this path (source: Elasticsearch, index test-logs) -- this testing is also what surfaced the empty-query-clause bug above.

### Notes for reviewers
Nothing about existing Elasticsearch source behavior changes -- the auth/PIT/tiebreaker workarounds only activate when `bearer_token` is non-empty; the query-clause fix is a strict improvement on both backends.

---

> [!NOTE]
> **Low Risk**
> Small, additive credential wiring with no change to default Elasticsearch auth when bearer token is omitted; secrets still follow the same flag/env patterns as other cluster credentials.
> 
> **Overview**
> Adds **`--bearer-token`** (and **`ELASTICSEARCH_BEARER_TOKEN`**) to the `elasticsearch` scan command so OpenSearch clusters that use JWT/Bearer auth can authenticate the same way as other Elasticsearch credential flags.
> 
> The value is passed through **`sources.ElasticsearchConfig`** into the scan engine’s Elasticsearch protobuf connection, and the man page documents the new option. Existing username/password, API key, and service-token flows are unchanged when bearer token is not set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07fabdeea02b707d83c2912ec5d176bfc3f0ec3a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MuneebUllahKhan222:
Can you add some tests that demonstrate the current Elasticsearch source doesn't work with OpenSearch, but that they pass with this PR? I think that would clearly show the value of the change and help ensure we don't regress in the future.

anon767:
Done

PR #5127, opened 2026-07-14
fix(detectors/privatekey): report encrypted keys with uncrackable passphrases

### LINKED ISSUE #5115: Privatekey detector silently drops encrypted keys when the passphrase isn't in the built-in wordlist
**Summary**
When a private key is passphrase-protected and Crack() can't guess the passphrase from the built-in wordlist (pkg/detectors/privatekey/list.txt, ~250 common passwords), the detector discards the finding entirely — it's not reported as verified or unverified. An encrypted key in a repo is still a real exposure (offline-crackable, weak legacy KDFs, passphrase often reused/committed nearby), so it should surface as an unverified finding.

**Cause**
In pkg/detectors/privatekey/privatekey.go (FromData), the failed-crack branch continues past the results = append(...):

parsedKey, passphrase, err = Crack([]byte(token))
if err != nil {
s1.SetVerificationError(err, token)
continue // <-- drops the finding (and the encrypted:"true" metadata just set)
}

**Reproduction**

ssh-keygen -t ed25519 -N "$(head -c24 /dev/urandom | base64)" -f ./enc_key
trufflehog filesystem ./enc_key --no-verification

Expected: one unverified PrivateKey finding.
Actual: zero findings.

**Control:**
same command with -N password (a wordlist entry) or -N "" (no passphrase) → finding is reported. The only variable is whether the passphrase is crackable.

**Suggested fix**
Append the result as unverified before continuing:
s1.SetVerificationError(err, token)
results = append(results, s1)
continue

genisis0x:
Closes #5115

## Problem

When a private key is passphrase-protected and `Crack()` cannot recover the passphrase from the built-in wordlist (`pkg/detectors/privatekey/list.txt`, ~250 common passwords), the detector discards the finding entirely — it is reported as neither verified nor unverified.

In `FromData`, the failed-crack branch set a verification error but then `continue`d past the `results = append(...)`:

```go
parsedKey, passphrase, err = Crack([]byte(token))
if err != nil {
    s1.SetVerificationError(err, token)
    continue // drops the finding (and the encrypted:"true" metadata just set)
}
```

An encrypted key committed to a repository is still a real exposure: it is offline-crackable, may use a weak or legacy KDF, and the passphrase is frequently reused or committed nearby. It should surface as an unverified finding.

## Fix

Append the result before continuing, so the encrypted key is reported as unverified with its verification error and `encrypted:"true"` metadata — matching how unencrypted keys are already reported under `--no-verification`.

## Reproduction

```
ssh-keygen -t ed25519 -N "$(head -c24 /dev/urandom | base64)" -f ./enc_key
trufflehog filesystem ./enc_key --no-verification
```

- Before: zero findings.
- After: one unverified `PrivateKey` finding.
- Control (`-N password` — a wordlist entry, or `-N ""` — no passphrase): finding is reported either way; the only variable is whether the passphrase is crackable.

## Test

Added `TestPrivatekey_EncryptedKeyReported`: an ed25519 key protected by a passphrase absent from the wordlist is scanned and asserted to yield exactly one unverified result carrying a verification error and `encrypted:"true"`. Proven to fail on the current code (0 results) and pass with the fix. `go test`/`go vet`/`golangci-lint` on the package are green.

---

> [!NOTE]
> **Low Risk**
> Narrow detector behavior change with a focused regression test; no auth or data-path impact beyond surfacing more findings.
> 
> **Overview**
> Fixes a gap where **passphrase-protected private keys** whose passphrase is not in the built-in wordlist were **silently dropped** after `Crack()` failed, even though `encrypted: true` was set.
> 
> When cracking fails, the detector now **appends an unverified finding** (with the verification error) before continuing, so encrypted keys still show up in scans such as `--no-verification`.
> 
> Adds **`TestPrivatekey_EncryptedKeyReported`** with an OpenSSH ed25519 key and a non-wordlist passphrase, asserting one unverified result with `encrypted: true` and a verification error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbc159a9b5204254f76b2262cb4b8c86ed7025ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MuneebUllahKhan222 (approved):
This looks like a good change to me.
I will also ask a `FTE` for their review.

shahzadhaider1 (approved):

unsmith (approved):
Looks good to me as well - I agree that reporting this situation as unverified is better than just dropping it.

PR #5130, opened 2026-07-15
Log analyze errors for HuggingFace analyzer

kashifkhan0771:
### Description:
Thread ctx through HuggingFace analyzer's request chain, log every failure at origin.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Observability and HTTP hygiene changes in a single analyzer; the stricter non-200 handling on model listing may surface errors that were previously swallowed.
> 
> **Overview**
> Threads **`context.Context`** through the HuggingFace analyzer (`Analyze` → `AnalyzePermissions` → `getTokenInfo` / `getModelsByAuthor`) so outbound calls use **`http.NewRequestWithContext`** and failures can be logged at the source with **`ctx.Logger()`**.
> 
> Adds **structured error logging** for missing credentials, request/transport/decode failures, and **V(2) info** for invalid tokens. **`getModelsByAuthor`** now **returns an error on non-200** responses instead of attempting to decode the body. Response bodies are **drained** with `io.Copy(io.Discard, …)` before close. The CLI helper **`AnalyzeAndPrintPermissions`** passes **`context.Background()`** into the updated API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 102a7478b24d322d44f01c424fdeedd10ff1d617. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/analyzer/analyzers/huggingface/huggingface.go
+++ b/pkg/analyzer/analyzers/huggingface/huggingface.go
@@ -5,6 +5,7 @@ package huggingface
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strings"
@@ -30,13 +31,15 @@ type Analyzer struct {
 
 func (Analyzer) Type() analyzers.AnalyzerType { return analyzers.AnalyzerTypeHuggingFace }
 
-func (a Analyzer) Analyze(_ context.Context, credInfo map[string]string) (*analyzers.AnalyzerResult, error) {
+func (a Analyzer) Analyze(ctx context.Context, credInfo map[string]string) (*analyzers.AnalyzerResult, error) {
 	key, ok := credInfo["key"]
 	if !ok || key == "" {
-		return nil, analyzers.NewAnalysisError(a.Type().String(), analyzers.OperationValidateCredentials, analyzers.ServiceConfig, "", fmt.Errorf("key not found in credentialInfo"))
+		err := fmt.Errorf("key not found in credentialInfo")
+		ctx.Logger().Error(err, "huggingface credentials missing key")
+		return nil, analyzers.NewAnalysisError(a.Type().String(), analyzers.OperationValidateCredentials, analyzers.ServiceConfig, "", err)
 	}
 
-	info, err := AnalyzePermissions(a.Cfg, key)
+	info, err := AnalyzePermissions(ctx, a.Cfg, key)
 	if err != nil {
 		return nil, analyzers.NewAnalysisError(a.Type().String(), analyzers.OperationAnalyzePermissions, analyzers.ServiceAPI, "", err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/analyzer/analyzers/huggingface/huggingface.go
```diff
@@ -330,28 +333,37 @@ func getModelsByAuthor(cfg *config.Config, key string, author string) ([]Model,
 	// send the request
 	resp, err := client.Do(req)
 	if err != nil {
+		ctx.Logger().Error(err, "failed to send huggingface models request", "author", author)
 		return modelsJSON, err
 	}
 
 	// defer the response body closing
 	defer func() { _ = resp.Body.Close() }()
```
MuneebUllahKhan222: Need to drain the response body
```go
defer func() {
    _, _ = io.Copy(io.Discard, resp.Body)
    _ = resp.Body.Close()
}()
```

pkg/analyzer/analyzers/huggingface/huggingface.go
MuneebUllahKhan222: Need to drain the response body
```go
defer func() {
    _, _ = io.Copy(io.Discard, resp.Body)
    _ = resp.Body.Close()
}()
```

MuneebUllahKhan222 (approved):
Two non-blocking comments.

pkg/analyzer/analyzers/huggingface/huggingface.go
```diff
@@ -361,19 +377,25 @@ func getTokenInfo(cfg *config.Config, key string) (HFTokenJSON, bool, error) {
...
 	}
 
+	// defer the response body closing
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
 	// check if the response is 200
 	if resp.StatusCode != 200 {
```
unsmith: legacy nit: should this use http.StatusOK to avoid magic numbers?

pkg/analyzer/analyzers/huggingface/huggingface.go
```diff
@@ -361,19 +377,25 @@ func getTokenInfo(cfg *config.Config, key string) (HFTokenJSON, bool, error) {
...
+		ctx.Logger().Error(fmt.Errorf("invalid/revoked huggingface access token"), "huggingface token invalid while fetching whoami", "status_code", resp.StatusCode)
```
unsmith: Noting this here, but it kinda applies to all the loglines - do we want these to be at error level, or are they just informational for debugging?  There's two places where an error is logged but no error is actually returned, suggesting it's not really an error.  I'm wondering if this will be overly noisy in production at an Error level.

unsmith (approved):
Logic looks sound, left a couple comments to think about.  Otherwise LGTM!

pkg/analyzer/analyzers/huggingface/huggingface.go
kashifkhan0771: Good suggestion - I reverted few to Info level which are not really an error.

PR #5132, opened 2026-07-15
Retry git clone on transient network errors

shahzadhaider1:
### Issue
Currently a repo clone gets exactly one attempt. If the connection drops mid-transfer (`curl 56 Recv failure: Connection reset by peer`, `HTTP/2 stream reset by server`, `early EOF`, etc.), the whole scan of that repo fails, even though an immediate retry would almost certainly succeed. We've seen this repeatedly with large repos, where long transfer times make a transient network blip much more likely to hit.

### Solution
This PR makes `CloneRepo` retry when the failure is classified as a network error (using the existing `ClassifyCloneError`, which was previously only used for metrics). Up to 3 total attempts with a short backoff (5s, then 10s), each starting from a fresh clone directory. 
All other failure classes - auth, permissions, rate limits, not found, clone timeouts, still fail immediately, same as before.

Along the way, the directory setup logic was extracted into `createClonePath` so the initial attempt and retries share one code path. Temp directories keep their `os.MkdirTemp` 0700 permissions on retry, and `--clone-path` directories keep 0755.

### Notes:
* This applies to all git-based sources, not just GitHub, since they all clone through `CloneRepo`.
* Clone timeouts are deliberately not retried. Note that the timeout now applies per attempt, so a repo that repeatedly fails with network errors near the cap could take up to 3x the configured timeout in the worst case. Happy to make it a shared budget across attempts if that's preferred.
* Clone failure metrics are recorded per attempt, so failure counts may tick up slightly even as overall clone reliability improves.

### Tests: 
- Added unit tests for the `isRetryableCloneError` and for `createClonePath`. 
- Also verified the retry flow end-to-end with a stub git binary that fails the first attempt and succeeds on the second.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core clone behavior for all git-based sources; worst case can extend wall time (per-attempt timeout × retries) and slightly inflate clone-failure metrics per attempt.
> 
> **Overview**
> **`CloneRepo`** now retries up to **3** times when failures are classified as transient **network** errors (via existing **`ClassifyCloneError`** / **`cloneFailureNetwork`**), with linear backoff (**5s**, then **10s**) and a **fresh clone directory** on each attempt. **Auth**, **403/429**, **not found**, and **clone timeouts** still fail immediately without retry.
> 
> Clone directory setup is moved into **`createClonePath`** so the first attempt and retries share one path; partial clones are removed before retrying, with cleanup on final failure unchanged.
> 
> Unit tests cover **`isRetryableCloneError`** and **`createClonePath`** (temp vs **`--clone-path`**, permissions, uniqueness).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b845e62de03131be2d7fc3318b4cb4cb057c8c45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/git/git.go
+++ b/pkg/sources/git/git.go
@@ -38,7 +38,15 @@ import (
 	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
 )
 
-const SourceType = sourcespb.SourceType_SOURCE_TYPE_GIT
+const (
+	SourceType = sourcespb.SourceType_SOURCE_TYPE_GIT
+	// maxCloneAttempts is the total number of times a clone is attempted when
+	// each failure is classified as a transient network error.
+	maxCloneAttempts = 3
+	// cloneRetryBackoff is the base wait between clone attempts; it is
+	// multiplied by the number of failed attempts so far.
+	cloneRetryBackoff = 5 * time.Second
+)
 
 type Source struct {
 	name     string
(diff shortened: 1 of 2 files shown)


pkg/sources/git/git_test.go
```diff
@@ -61,6 +62,131 @@ func TestClone_Timeout(t *testing.T) {
...
+		"could not clone repo: https://github.com/org/repo.git, error executing git clone: exit status 128, remote: The requested URL returned error: 403",
+		"could not clone repo: https://github.com/org/repo.git, error executing git clone: exit status 128, The requested URL returned error: 429",
+		"git clone timed out (after 1h0m0s)",
+		"could not clone repo: https://github.com/org/repo.git, error executing git clone: exit status 128, fatal: repository 'https://github.com/org/repo.git/' not found",
+	}
+	for _, msg := range notRetryable {
+		assert.False(t, isRetryableCloneError(errors.New(msg)), "expected not retryable: %q", msg)
+	}
+	assert.False(t, isRetryableCloneError(nil))
+}
```
MuneebUllahKhan222, line 94: Better to refactor this in a table test.

MuneebUllahKhan222 (approved):
LGTM. Just two small refactoring changes that can be done.

pkg/sources/git/git.go
```diff
@@ -474,38 +482,91 @@ type cloneParams struct {
...
+		case <-ctx.Done():
+			CleanOnError(&err, path)
+			return "", nil, err
+		case <-time.After(cloneRetryBackoff * time.Duration(attempt)):
 		}
 	}
 
-	timeout := time.Duration(feature.GitCloneTimeoutDuration.Load())
+	return path, repo, nil
+}
```
MuneebUllahKhan222, line 536: I think we can re-write this function better like this: (To make it easier to read)
```go
func CloneRepo(ctx context.Context, userInfo *url.Userinfo, gitURL string, clonePath string, authInUrl bool, args ...string) (string, *git.Repository, error) {
      timeout := time.Duration(feature.GitCloneTimeoutDuration.Load())

      var path string
      var repo *git.Repository
      var err error
      for attempt := 1; attempt <= maxCloneAttempts; attempt++ {
              path, err = createClonePath(gitURL, clonePath)
      var err error
      for attempt := 1; attempt <= maxCloneAttempts; attempt++ {
              path, err = createClonePath(gitURL, clonePath)
              if err != nil {
                      return "", nil, err
              }

              repo, err = executeClone(ctx, cloneParams{userInfo, gitURL, args, path, authInUrl, timeout})
              if err == nil {
                      return path, repo, nil
              }

              if attempt >= maxCloneAttempts || !isRetryableCloneError(err) || common.IsDone(ctx) {
                      CleanOnError(&err, path)
                      return "", nil, err
              }

              if rmErr := os.RemoveAll(path); rmErr != nil {
                      return "", nil, fmt.Errorf("failed to clean clone path for retry: %w (original clone error: %w)", rmErr, err)
              }

              ctx.Logger().Info("git clone interrupted by network error; retrying",
                      "attempt", attempt, "max_attempts", maxCloneAttempts, "error", err.Error())

              select {
              case <-ctx.Done():
                      CleanOnError(&err, path)
                      return "", nil, err
              case <-time.After(cloneRetryBackoff * time.Duration(attempt)):
              }
      }
      return "", nil, err
}
```

unsmith: Might be a typo at the top of this block (a couple of duplicated lines?) but otherwise I'm on board with this.  I don't feel strongly about it because the logic does look sound as-is, but if you both agree it would be a readability improvement then I'd say go ahead.  I'll approve anyway.

unsmith (approved):
> Clone timeouts are deliberately not retried. Note that the timeout now applies per attempt, so a repo that repeatedly fails with network errors near the cap could take up to 3x the configured timeout in the worst case. Happy to make it a shared budget across attempts if that's preferred.

I think this is fine as-is, but I'd wonder what sort of time loss we'd be absorbing if we did retry timeouts.  Maybe as a future improvement, if we have a sense of how long a clone can take.

pkg/sources/git/git.go
shahzadhaider1: done

shahzadhaider1:
> > Clone timeouts are deliberately not retried. Note that the timeout now applies per attempt, so a repo that repeatedly fails with network errors near the cap could take up to 3x the configured timeout in the worst case. Happy to make it a shared budget across attempts if that's preferred.
> 
> I think this is fine as-is, but I'd wonder what sort of time loss we'd be absorbing if we did retry timeouts. Maybe as a future improvement, if we have a sense of how long a clone can take.

Yeah, good call for a follow-up. Main reason I skipped timeouts: network errors fail fast so a retry just costs the backoff, but a timeout retry burns a full budget each attempt with no sign it'll succeed, worst case 3x the cap just waiting.

PR #5134, opened 2026-07-18
fix(privatekey): report encrypted keys as unverified when passphrase is uncrackable

sergioperezcheco:
When a private key is passphrase-protected and the built-in wordlist in Crack() can't recover the passphrase, the detector was discarding the finding entirely (continue without appending to results). An encrypted private key committed to a repo is still a real exposure — it's offline-crackable, often uses weak legacy KDFs, and the passphrase is frequently reused or committed nearby — so it should surface as an unverified finding rather than vanish.

The fix appends the result with encrypted="true" metadata and the crack error as the verification error before continuing, matching how other unverified findings are reported. Reported in #5115.

I added a regression test (TestPrivatekey_EncryptedUncrackable) that generates an ed25519 key encrypted with a random passphrase outside the wordlist and asserts exactly one unverified result with the encrypted flag and a verification error. Existing TestPrivatekey_Pattern still passes.

---

> [!NOTE]
> **Low Risk**
> Narrow detector behavior change that increases visibility of findings; no auth or data-path changes, covered by a focused regression test.
> 
> **Overview**
> **Fixes silent drops of passphrase-protected private keys** when `Crack()` cannot recover the passphrase from the built-in wordlist (#5115).
> 
> Previously, encrypted keys that failed cracking were skipped entirely. The detector now **returns an unverified finding** with `encrypted=true`, attaches the crack failure via `SetVerificationError`, and appends it to results—same pattern as other unverified secrets.
> 
> Adds **`TestPrivatekey_EncryptedUncrackable`**: generates an ed25519 key encrypted with a non-wordlist passphrase and asserts one unverified result with the encrypted flag and a verification error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ce6f216957fe8da1bc8f5a3d0b68ad3809c56aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/privatekey/privatekey.go
+++ b/pkg/detectors/privatekey/privatekey.go
@@ -81,7 +81,13 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 			s1.ExtraData["encrypted"] = "true"
 			parsedKey, passphrase, err = Crack([]byte(token))
 			if err != nil {
+				// The key is encrypted and its passphrase is not in the built-in
+				// wordlist. An encrypted private key in a repo is still a real
+				// exposure (offline-crackable, weak legacy KDFs, passphrase often
+				// reused/committed nearby), so surface it as an unverified finding
+				// instead of silently dropping it.
 				s1.SetVerificationError(err, token)
+				results = append(results, s1)
 				continue
 			}
 			if passphrase != "" {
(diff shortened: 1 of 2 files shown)


sergioperezcheco:
Closing as duplicate — the same fix was merged in #5127 (by @genisis0x). Our diff was functionally identical, differing only in comment phrasing. Happy to see this addressed upstream!

PR #5142, opened 2026-07-20
Fix `scan_all_installations` Rejecting Org Member Personal Repos

shahzadhaider1:
### Problem
With `scan_all_installations` enabled alongside `scan_users`, scans fail on org members' personal repos:
Configured repos were not found in any GitHub App installation: Member enumeration adds members' public personal repos as scan units, but the installation mapping in `ChunkUnit` rejects any repo that no app installation owns. GitHub App installations only ever contain repos owned by the installing account, so member personal repos can never match, even though they're public and clone fine with the default installation token (single-installation mode scans them without issue). A single unmappable unit also fails the whole run.

### Changes
- Enumeration: repos discovered outside installation listings (member personal repos) are now mapped to the default installation up front, so the mapping in ChunkUnit short-circuits with no extra API calls.
- Chunk time: for the case where `ChunkUnit` runs without enumeration state, a new fallback (mapRemainingAccessibleRepos) runs after installation listings and metadata matching both miss; if the default installation token can read the repo, it's mapped to the default installation and scanned; if not, the mapping still fails closed.
- `ensureRepoInstallation` does its check-and-set under a single lock so a concurrent mapping to a real installation is never overwritten.

### Testing
- New regression tests: member personal repo maps to the default installation (both at enumeration and chunk time); an unreadable repo still fails the mapping, so the check doesn't become fail-open.
- Flipped `TestMapExplicitReposToInstallationsRejectsRepoMissingFromInstallationList` -> `FallsBackToDefaultForReadableRepo`: its mock declares the repo readable by the default token, which is exactly the case this change legalizes. Host-mismatch, wiki, and inaccessible-repo rejection tests pass unchanged.
- Verified end-to-end against a real GitHub App on a test org: previously the scan died on the first member repo; now all org and member repos scan cleanly with identical results to a single-installation run.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes GitHub App installation-to-repo mapping and clone token selection for scan_all_installations; behavior is scoped but affects which installation token is used for non-owned repos.
> 
> **Overview**
> Fixes **`scan_all_installations` + `scan_users`** runs that failed when org members’ **personal repos** were enumerated but **no app installation owns** them, even when they’re public and cloneable with the default installation token.
> 
> **Enumeration:** While listing user repos (e.g. member personal repos), repos are **pre-mapped to the configured default installation** via `ensureRepoInstallation`, so later `ChunkUnit` mapping can short-circuit without extra API calls. Mapping is **check-then-set under one lock** so an existing installation mapping is never overwritten.
> 
> **Chunk / explicit mapping:** After installation listings and metadata matching still leave repos unmatched, **`mapRemainingAccessibleRepos`** probes each repo with the **default installation** client; if GitHub returns the repo (and host matches), it’s mapped to the default installation and scanned. Repos the default token **cannot** read still fail with the existing “not found in any installation” error (**fail-closed**).
> 
> Tests cover readable-repo fallback, member personal repo at map and enumerate time, and inaccessible repos still rejected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b34895ad1ee4df3c9f123c465ef19592013260e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/sources/github/github.go
```diff
@@ -927,6 +927,16 @@ func (s *Source) mapReposToInstallations(ctx context.Context, connector *appConn
 		}
 	}
 
+	if len(wantedRepos) > 0 {
+		errs = errors.Join(errs, s.mapRemainingAccessibleRepos(ctx, connector, wantedRepos))
+		if len(wantedRepos) == 0 {
+			if errs != nil {
+				ctx.Logger().Error(errs, "some installations could not be fully enumerated while mapping repos")
```
MuneebUllahKhan222: Should we return `errs` here?

MuneebUllahKhan222 (approved):
LGTM.

pkg/sources/github/github.go
camgunz: Ah I've probably never linked you to our [logging guidance](https://www.notion.so/Logging-Best-Practices-dcb5323c1fcc48d9881f9bff16cab2a2#dcb5323c1fcc48d9881f9bff16cab2a2). We generally choose either to log **or** return an error; we almost never do both.

pkg/sources/github/github.go
```diff
@@ -1031,6 +1041,55 @@ func (s *Source) mapRemainingReposByMetadata(
 	return errs
 }
 
+// mapRemainingAccessibleRepos handles repos that no installation owns, such
+// as org members' personal repos enumerated via scan_users. Installations
+// only ever contain repos owned by the installing account, but the default
+// installation token can still read repos that are visible to it (public
+// repos in particular), so a repo that the API confirms accessible is mapped
+// to the default installation instead of failing the scan.
+func (s *Source) mapRemainingAccessibleRepos(ctx context.Context, connector *appConnector, wantedRepos map[string]repoMappingRequest) error {
```
camgunz: **suggestion:** Some lines here are so long! It's not blocking but can we keep 'em under 120?

camgunz (approved):

pkg/sources/github/github.go
shahzadhaider1: Haha, how could I forget? 😄
Done!

PR #5143, opened 2026-07-20
Propagate job attempts through source progress

juvoni:
### Description:

Overlapping runs of the same source job can currently share UnitHook correlation state because progress is identified only by source, job, and unit. One attempt can overwrite or finish another attempt's in-flight metrics when those identifiers are reused.

This adds an optional, caller-defined `JobAttemptID` to source progress. Callers attach it through `ContextWithJobAttemptID`, and every SourceManager entry point preserves it on the returned `JobProgressRef`, including preflight errors. UnitHook includes the attempt in its internal correlation key for both unit and non-unit sources, so distinct attempts remain independent.

The zero value remains unspecified and is omitted from JSON, preserving behavior for existing callers. `JobProgressRef` also remains comparable. This draft makes the smallest OSS-facing contract concrete for maintainer feedback before downstream adoption.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Backward-compatible optional API in `pkg/sources` with no auth or data-path changes; behavior unchanged when attempt ID is not set.
> 
> **Overview**
> Adds an optional, caller-defined **`JobAttemptID`** so concurrent retries or overlapping runs of the same source/job no longer collide in **`UnitHook`** metrics (previously keyed only by source, job, and unit).
> 
> Callers set the attempt via **`ContextWithJobAttemptID`**; **`SourceManager`** entry points (`EnumerateAndScan`, `Enumerate`, `Scan`) read it from context, attach it to **`JobProgress`**, and expose it on **`JobProgressRef`** (including on preflight failures). **`UnitHook`** includes the attempt in its internal correlation key for unit and non-unit paths.
> 
> **`job_attempt_id`** is omitted from JSON when zero; **`JobProgressRef`** stays comparable. Tests cover propagation, preflight, JSON, and attempt-separated hook metrics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd7fbed1f8ec62babcd04ad3bcb6a698dc99692e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/job_progress_hook.go
+++ b/pkg/sources/job_progress_hook.go
@@ -67,7 +67,7 @@ func (u *UnitHook) id(ref JobProgressRef, unit SourceUnit) string {
 		id, kind := unit.SourceUnitID()
 		unitID = fmt.Sprintf("%s:%s", kind, id)
 	}
-	return fmt.Sprintf("%d/%d/%s", ref.SourceID, ref.JobID, unitID)
+	return fmt.Sprintf("%d/%d/%d/%s", ref.SourceID, ref.JobID, ref.JobAttemptID, unitID)
 }
 
 func (u *UnitHook) ejectFinishedMetrics(metrics UnitMetrics) {
--- a/pkg/sources/job_progress_hook.go
+++ b/pkg/sources/job_progress_hook.go
@@ -67,7 +67,7 @@ func (u *UnitHook) id(ref JobProgressRef, unit SourceUnit) string {
 		id, kind := unit.SourceUnitID()
 		unitID = fmt.Sprintf("%s:%s", kind, id)
 	}
-	return fmt.Sprintf("%d/%d/%d/%s", ref.SourceID, ref.JobID, ref.JobAttemptID, unitID)
+	return fmt.Sprintf("%d/%d/%s/%d", ref.SourceID, ref.JobID, unitID, ref.JobAttemptID)
 }
 
 func (u *UnitHook) ejectFinishedMetrics(metrics UnitMetrics) {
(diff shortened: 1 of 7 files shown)


kashifkhan0771 (approved):

pkg/sources/job_progress_test.go
```diff
@@ -36,10 +37,12 @@ func TestJobProgressFatalErrors(t *testing.T) {
 }
 
 func TestJobProgressRef(t *testing.T) {
-	jp := NewJobProgress(123, 456, "source name")
+	jp := NewJobProgress(123, 456, "source name", withJobAttemptID(7))
 	ref := jp.Ref()
 	assert.Equal(t, JobID(123), ref.JobID)
 	assert.Equal(t, SourceID(456), ref.SourceID)
+	assert.Equal(t, JobAttemptID(7), ref.JobAttemptID)
+	assert.Contains(t, map[JobProgressRef]bool{ref: true}, ref)
```
mcastorina: Is this line testing anything? It looks like it just asserts a map of `ref: true` contains `ref`?

pkg/sources/job_progress_hook.go
```diff
@@ -67,7 +67,7 @@ func (u *UnitHook) id(ref JobProgressRef, unit SourceUnit) string {
 		id, kind := unit.SourceUnitID()
 		unitID = fmt.Sprintf("%s:%s", kind, id)
 	}
-	return fmt.Sprintf("%d/%d/%s", ref.SourceID, ref.JobID, unitID)
+	return fmt.Sprintf("%d/%d/%d/%s", ref.SourceID, ref.JobID, ref.JobAttemptID, unitID)
```
mcastorina: **Nit:** This is an internal `id`, but it might be useful to put the attempt ID at the end, so we could do prefix matching.

mcastorina (commented):
Is there somewhere I can read about job attempts? Some questions I have:

* Why are there overlapping runs of the same source job?
* What is a job attempt describing? Is it just an opaque value from the caller to uniquely identify a job? Why is job ID not enough?
* Does the attempt ID increment for the same `JobID`, or would the `JobID` be different every time?

PR #5154, opened 2026-07-24
fix(sources): stop UnitHook Close racing in-flight Finish sends

bill-rich:
## What

Two-part fix for `panic: send on closed channel` in `UnitHook.ejectFinishedMetrics`:

1. Reorders the deferred calls at all three SourceManager job-spawn sites so `wg.Done()` is declared first (runs last). Previously `defer progress.Finish()` was declared before `defer s.wg.Done()`, so with LIFO ordering the WaitGroup released before Finish ran its hooks, and `Wait()` could close the hooks while a Finish hook was still sending on `finishedMetrics`.
2. Hardens `UnitHook.Close`: it is now idempotent, waits for in-flight sends before closing the channel, and sends that start after Close are dropped instead of panicking, protecting any caller that closes the hook while jobs are still running.

## Why

This panic is crashing scanner processes built against this package: 14 occurrences in the last 30 days across a hosted fleet, the most frequent panic signature there. Stack: `UnitHook.Finish` -> `ejectFinishedMetrics` (job_progress_hook.go:77) via `JobProgress.Finish` from `SourceManager.EnumerateAndScan.func1`.

The defer ordering also explains a latent flake in `TestSourceManagerUnitHookNoUnits`: metrics could be lost (or panic) when Close won the race against a finishing job.

## Verification

New `TestUnitHookCloseConcurrentWithFinish` reproduces the panic without the fix (5 runs, 6 panics) and passes with it. Full `./pkg/sources` suite passed 24 consecutive `-race` runs with the fix.

---

> [!NOTE]
> **Medium Risk**
> Touches concurrent job lifecycle and channel shutdown in the core source manager path; behavior change is intentional (metrics may be dropped after close) but reduces production panics.
> 
> **Overview**
> Fixes **`panic: send on closed channel`** when scanner jobs finish while `SourceManager.Wait()` is tearing down progress hooks.
> 
> **SourceManager** job goroutines now declare `defer s.wg.Done()` *before* `defer progress.Finish()` so LIFO teardown runs **Finish (and hook sends) before** the wait group drops to zero. That keeps `Wait()` from closing `UnitHook` while `Finish` is still writing to `finishedMetrics`.
> 
> **UnitHook** shutdown is coordinated: `Close` is idempotent, waits on in-flight `ejectFinishedMetrics` sends, then closes the channel; sends that start after close are dropped via a `done` signal instead of blocking or panicking. The blocking retry path after back-pressure also respects `done`.
> 
> Adds **`TestUnitHookCloseConcurrentWithFinish`** to stress concurrent `Finish` vs `Close`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e07febb48b6f5e567c6926d1b505fab1c929bf95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/job_progress_hook.go
+++ b/pkg/sources/job_progress_hook.go
@@ -17,6 +17,13 @@ type UnitHook struct {
 	mu              sync.Mutex
 	finishedMetrics chan UnitMetrics
 	logBackPressure func()
+	// closeMu, closed, done, and inflight coordinate Close with in-flight
+	// ejectFinishedMetrics sends from still-running job goroutines, so
+	// finishedMetrics is never closed while a send is possible.
+	closeMu  sync.Mutex
+	closed   bool
+	done     chan struct{}
+	inflight sync.WaitGroup
 	NoopHook
 }
 
@@ -36,6 +43,7 @@ func NewUnitHook(ctx context.Context, opts ...UnitHookOpt) (*UnitHook, <-chan Un
 	hook := UnitHook{
 		metrics:         make(map[string]*UnitMetrics, runtime.NumCPU()),
 		finishedMetrics: make(chan UnitMetrics, 1024),
+		done:            make(chan struct{}),
 		logBackPressure: func() {
 			once.Do(func() {
 				ctx.Logger().Info("back pressure detected in unit hook")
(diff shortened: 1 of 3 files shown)


pkg/sources/job_progress_hook.go
```diff
@@ -189,6 +213,19 @@ func (u *UnitHook) InProgressSnapshot() []UnitMetrics {
 }
 
 func (u *UnitHook) Close() error {
```
unsmith, line 215: Logic seems sound to me - stop new senders, then wait for existing senders to finish, then close. 👍

unsmith (approved):
Looks solid!

Issue #3262, opened 2024-09-02
Trufflehog not detecting valid secrets as verified result


rnillavazagan:
### TruffleHog Version
3.81.9

trufflehog filesystem scan not detecting a valid Github Server-Server token (GHS) when `--only-verified` is applied.

### Expected Behavior

Scan results should report the GHS secret as verified result.

### Actual Behavior

No secrets are reported.

### Steps to Reproduce

1. Place a valid Github Server-Server token (GHS) in any of the sample file.
2. Execute trufflehog filesystem scan against this file.

## Additional Context
Same issues persists for GHS tokens present in docker images as well.

kashifkhan0771:
I conducted testing and discovered that the GitHub Server-to-Server token (ghs_*) must have the appropriate permissions granted in order to perform specific operations. If the token lacks the required permissions, it results in a 403 Forbidden error. Conversely, if the token is expired or invalid, it returns a 401 Unauthorized error.

In our GitHub verification call using the /user API, it’s important to note that not all ghs_* tokens will necessarily have the permission to access user profile information. This means that requests to the /user endpoint can fail if the token lacks the necessary permissions. During my testing, even after granting user information access to the token, the API still returned a 403 error, which was perplexing.

On the other hand, when I made a request to the /issues API after providing the token with the necessary issues read access, it worked as expected.

Therefore, I guess we need a more refined approach to handle permissions for Server-to-Server(`ghs_*`) tokens.

@zricethezav @abasit-folio3

jglick:
https://github.com/trufflesecurity/trufflehog/blob/e1f9e5d1d55d12e4a976f06deef4ccf20e099644/pkg/detectors/github/v2/github.go#L36 will not (I think) match as of https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/

PR #5156, opened 2026-07-27
fix(github): detect new installation token format

Eljees:
### Description:

GitHub is rolling out a new, variable-length format for App installation tokens: `ghs_APPID_JWT`. The existing detector treats every GitHub token as a bounded alphanumeric string, so it truncates these tokens at the first JWT separator and cannot verify them.

Match the documented three-segment JWT form before the legacy token form while preserving all existing token patterns. The regression test uses a synthetic ~520-character token and also verifies that a trailing base64url `-` is retained.

Addresses the token-format failure noted in #3262. This does not change the separate API-permission behavior discussed earlier in that issue.

### Testing:

* [x] `go test ./pkg/detectors/github/v2 -count=1`
* [x] Regression test fails on `main` because the token is truncated before the first `.`
* [x] `gofmt` reports no changes
* [ ] Full `make test-community` suite (targeted detector tests were run)
* [ ] Full `make lint` (formatting was checked locally)

---

> [!NOTE]
> **Low Risk**
> Changes are limited to GitHub v2 secret matching and buffer sizing, with regression tests for truncation and over-matching; no auth or data-path changes in the product runtime.
> 
> **Overview**
> Updates the **GitHub v2 detector** so it can find and pass through full **GitHub App installation tokens** in the new `ghs_<appId>_JWT` shape, which are longer than classic tokens and were getting cut off at JWT `.` separators under the old pattern and default scan window.
> 
> The scanner now implements **`MaxSecretSizeProvider`** with a **4096-byte** window (instead of the engine default **512**), and **`keyPat`** tries the three-segment JWT form first, then the existing `ghp`/`gho`/`ghu`/`ghs`/`ghr`/`github_pat` patterns.
> 
> Tests add a **~520-character** synthetic installation token through the Aho–Corasick engine path and assert **legacy `ghs_` tokens** are not over-matched when followed by something like `.config.yaml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 608af4a41e3e31073650de24aefa43156dd28f94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Eljees:
Addressed both review findings in 507a71a1: the GitHub v2 scanner now uses a 4096-byte MaxSecretSize window and the test exercises DetectorMatch.Matches() through the engine; the new installation-token branch also requires JWT-like eyJ prefixes so legacy ghs_ tokens followed by dot-separated text are not over-matched. Focused detector tests pass.

Eljees:
Ping - this has been open since 27 July without a review. Happy to rebase it, split it, or close it if the direction is wrong.

Eljees:
Rebased onto current `main` in 608af4a4 - the branch had gone `CONFLICTING`, so it could not be merged even if someone wanted to.

The conflict was in `pkg/detectors/github/v2/github_test.go` only, and it was a both-added one: `main` gained `TestGithubTokenType` and the `keyPat` drift guard while this branch had added the engine-window and over-match regression tests. Kept both sides, plus the `newInstallationToken` fixture the latter needs. Net diff against `main` is unchanged at two files, +48/-2.

Worth noting because it is the one place the two could really have collided: the drift guard still holds against the widened `keyPat`. `prefixAlternationPat` finds the same `(?:ghp|gho|ghu|ghs|ghr|github_pat)_` group, and the new `ghs_<appid>_<jwt>` alternative introduces no prefix that is missing from `tokenTypesByPrefix`, so `TestGithubTokenType_MappingMatchesKeyPatExactly` passes unchanged.

`go vet` is clean and all six tests in the package pass, the four from `main` included.

Both review findings from July are still addressed: the scanner declares `MaxSecretSize() 4096` so the full installation token reaches `FromData` through the engine window, and the JWT branch requires `eyJ` on both of the first two segments so a legacy `ghs_` token followed by dot-separated text is not over-matched.

PR #5159, opened 2026-07-27
fix(analyzers/groq): surface free and paid plan resources

mattbrady-1:
## Summary
- Surface whatever the Groq key can access: models on every plan, plus batches and files on paid Developer plans.
- Free keys no longer look like they have zero permissions; paid keys still pick up batch/file resources when available.
- Fix metadata and binding construction so analyze results are populated correctly.

## Test plan
- [ ] Free-tier key: `go run . analyze groq` shows model resources with `full_access`
- [ ] Paid Developer key with batches/files: same command also shows those resources with `full_access`
- [ ] `go build ./pkg/analyzer/analyzers/groq/`
- [ ] `go test ./pkg/analyzer/analyzers/groq/ -count=1` (needs GCP test secrets)

### DIFF
--- a/pkg/analyzer/analyzers/groq/groq.go
+++ b/pkg/analyzer/analyzers/groq/groq.go
@@ -19,7 +19,7 @@ type Analyzer struct {
 	Cfg *config.Config
 }
 
-// SecretInfo hold the information about the groq key
+// SecretInfo holds information gathered about a Groq API key.
 type SecretInfo struct {
 	Valid         bool
 	Reference     string
@@ -28,7 +28,7 @@ type SecretInfo struct {
 	Misc          map[string]string
 }
 
-// GroqResource is a single groq resource which can be accessed with groq api key
+// GroqResource is a single resource accessible with a Groq API key.
 type GroqResource struct {
 	ID         string
 	Name       string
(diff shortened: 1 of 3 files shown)


pkg/analyzer/analyzers/groq/groq.go
```diff
@@ -37,17 +37,17 @@ type GroqResource struct {
...
-// appendGroqResource append the single groq resource to secretinfo groqresources list
+// appendGroqResource appends a resource onto the secret info list.
 func (s *SecretInfo) appendGroqResource(resource GroqResource) {
 	s.GroqResources = append(s.GroqResources, resource)
 }
 
-// updateMetadata safely update the metadata of the groq resource
-func (g GroqResource) updateMetadata(key, value string) {
+// updateMetadata sets a metadata key on the resource. Pointer receiver so
+// writes persist when callers mutate before appending.
```
unsmith, line 46: The fix is good here, but I don't think it's necessary to call out a pointer receiver here, that's just standard Go for having a mutable struct.

pkg/analyzer/analyzers/groq/requests.go
```diff
@@ -90,96 +100,126 @@ func makeGroqRequest(client *http.Client, url, key string) ([]byte, int, error)
...
+			resource.updateMetadata("created at", time.Unix(m.Created, 0).UTC().Format("2006-01-02 15:04:05 UTC"))
+			secretInfo.appendGroqResource(*resource)
+		}
+		return nil
+	case http.StatusUnauthorized:
+		secretInfo.Valid = false
+		return fmt.Errorf("invalid Groq API key")
+	default:
+		return fmt.Errorf("unexpected status code: %d", statusCode)
+	}
```
unsmith, line 138: This doesn't have any resiliency for transient failures - if Groq's endpoint(s) give a temporary 500 or 503 or similar, this analyzer will treat that as a permanent(?) failure.  Not something specifically _wrong_ here as the other endpoints here operate the same way, just something to note.  If we're fixing up Groq's analyzer to be better, it might be worth it to also introduce some retry resiliency.

unsmith (approved):
Approving, with a couple of notes about possible improvements.  Nothing blocking.

PR #5165, opened 2026-07-30
feat(output): add SARIF output format for GitHub code scanning

kashifkhan0771:
### Description:
Adds --sarif flag emitting a SARIF 2.1.0 log, buffered across the scan and flushed once finished since SARIF isn't a streamable format like the existing printers. Verified results map to "error", unverified to "warning", with a stable per-finding fingerprint for cross-scan new/fixed tracking when uploaded via github/codeql-action/upload-sarif.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive output path and CLI flag with no changes to detection or verification; main operational note is higher memory use on very large result sets when buffering SARIF.
> 
> **Overview**
> Adds a **`--sarif`** CLI flag that emits **SARIF 2.1.0** after a scan finishes, aimed at **GitHub code scanning** via `github/codeql-action/upload-sarif`.
> 
> A new **`SarifPrinter`** buffers findings during the run (SARIF is one JSON document, not line-streamed), then **`Flush`** writes the log from `main` when the scan completes. Verified findings use SARIF level **error**, unverified **warning**, with **partial fingerprints** for stable new/fixed tracking across runs. File/line metadata is shared through a new **`extractFileAndLine`** helper (also used by the GitHub Actions printer).
> 
> README, man page, and a GitHub Actions workflow snippet document `--sarif` and the upload flow; tests cover SARIF shape, empty runs, fingerprints, and verification-stable IDs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d2be0955183a438397e953946637769b9d9ee18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/output/sarif.go
+++ b/pkg/output/sarif.go
@@ -1,0 +1,268 @@
+package output
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/context"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/version"
+)
+
+// SARIF (Static Analysis Results Interchange Format) 2.1.0 identifiers.
+// See https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html
+const (
+	sarifSchemaURI = "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json"
+	sarifVersion   = "2.1.0"
+
+	// sarifLevelError/sarifLevelWarning map to SARIF's result.level. Verified secrets are
+	// confirmed live credentials and are reported as errors; unverified secrets matched a
+	// detector pattern but could not be confirmed, and are reported as warnings so that tools
+	// consuming SARIF (e.g. GitHub code scanning) can triage them separately.
+	sarifLevelError   = "error"
+	sarifLevelWarning = "warning"
+)
+
+// SarifPrinter is a printer that accumulates results and, once the scan completes, emits them
+// as a single SARIF 2.1.0 log. Unlike the other Printer implementations, SARIF results cannot be
+// streamed one-per-line: the spec requires one JSON document containing every run and result, so
+// Print only buffers results and Flush performs the actual marshal/write.
+//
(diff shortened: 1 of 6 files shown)


pkg/output/sarif.go
```diff
@@ -0,0 +1,263 @@
...
+
+	rules := make([]*sarifRule, 0, len(p.rules))
+	for _, rule := range p.rules {
+		rules = append(rules, rule)
+	}
+	// Sort for deterministic output; map iteration order is randomized in Go.
+	sort.Slice(rules, func(i, j int) bool { return rules[i].ID < rules[j].ID })
+
+	results := p.results
+	if results == nil {
```
MuneebUllahKhan222, line 129: I think the condition should be
```go
if results == nil || len(results) == 0 {
...
}
```

pkg/output/sarif.go
```diff
@@ -0,0 +1,263 @@
...
+// as a single SARIF 2.1.0 log. Unlike the other Printer implementations, SARIF results cannot be
+// streamed one-per-line: the spec requires one JSON document containing every run and result, so
+// Print only buffers results and Flush performs the actual marshal/write.
+type SarifPrinter struct {
+	mu      sync.Mutex
+	results []sarifResult
+	rules   map[string]*sarifRule // keyed by detector type name, de-duplicated across results
+}
+
+// Print buffers a single result for inclusion in the SARIF document written by Flush.
```
MuneebUllahKhan222, line 47: This is actually concerning since accumulating all the results in a `results` slice can lead to OOM issue.

MuneebUllahKhan222 (changes requested):
Requesting few changes.

pkg/output/sarif.go
kashifkhan0771: Not needed - results is in a struct, never initialized to an empty literal, so it stays nil unless we append to it. Here `== nil` and `len() == 0` are equivalent.

pkg/output/sarif.go
kashifkhan0771: For now, I've documented this limitation in the README and added a TODO in the SARIF file. Addressing it will require a separate PR and a dedicated strategy to resolve the underlying limitation.

I think this is sufficient for the initial release. It should work well for small GitHub organizations and is likely to cover many medium-sized GitHub organizations as well.

MuneebUllahKhan222 (approved):

unsmith (approved):
It looks like all concerns have been addressed and I can't find anything additional, so I think this one's good to go 🚢

PR #5175, opened 2026-08-03
Give each clone its own directory under `--clone-path`

shahzadhaider1:
With `--clone-path` set and concurrency > 1, repos whose URLs end in the same segment get cloned into the same directory, so workers clone into and delete each other's directories. `gitlab.com/group-a/api.git` and `gitlab.com/group-b/api.git` both land in `<clone-path>/trufflehog-api`.

Depending on the code path, this either aborts the whole scan or silently skips the repo and reports fewer findings 

## Cause
`createClonePath` named the directory after the URL's last segment and created it with `os.MkdirAll`, which quietly succeeds if the directory already exists.

## Fix
Use `os.MkdirTemp`, which appends a random suffix and claims the name atomically:

```go
path, err := os.MkdirTemp(clonePath, "trufflehog-"+slug+"-")
```

Giving `<clone-path>/trufflehog-api-1729579074`. The `trufflehog-` prefix stays because `cleantemp.CleanTempDirsForLegacyJSON` matches on it, and `0755` is kept via an explicit `os.Chmod` since `MkdirTemp` creates `0700`. The default (no `--clone-path`) branch is untouched; it already used `cleantemp.MkdirTemp()`.

## Notes
- `--no-cleanup` now leaves one directory per clone rather than reusing a stable name. Flag help updated on `git`, `github`, and `gitlab`.
- Three new tests cover the collision and the concurrent case; five existing subtests asserted the old fixed name and were updated. 
- Reproduced against `main` on real GitLab; two `test-project` repos in different groups at `--concurrency=8` fail, and pass on this branch.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core git clone path logic used by git/github/gitlab scans; behavior is safer under concurrency but `--no-cleanup` now leaves many directories instead of one stable name per repo.
> 
> **Overview**
> Fixes concurrent scans with `--clone-path` where repos sharing the same URL basename (e.g. different GitLab groups both named `api`) were cloned into one fixed path via `os.MkdirAll`, causing workers to overwrite or delete each other's trees.
> 
> **`createClonePath`** now uses `os.MkdirTemp` under the base path with names like `trufflehog-<slug>-<random>`, keeps the `trufflehog-` prefix for legacy cleanup, and `chmod`s to `0755`. Each clone and retry gets its own directory.
> 
> **CLI/docs:** `--no-cleanup` help on `git`, `github`, and `gitlab` (and man page) notes that retained clones are one directory per clone and disk usage accumulates. GitLab `--clone-path` help gets a trailing period for consistency.
> 
> **Tests:** Unit tests updated for the new naming; added coverage for basename collisions, concurrent `createClonePath`, and an end-to-end concurrent `CloneRepo` case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a7b6419256e6271c1f0473bd877021384e5bb3b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/git/git_test.go
+++ b/pkg/sources/git/git_test.go
@@ -123,11 +123,14 @@ func TestCreateClonePath(t *testing.T) {
 		assert.NotEqual(t, first, second)
 	})
 
-	t.Run("clonePath set builds trufflehog-<repo> subdirectory", func(t *testing.T) {
+	t.Run("clonePath set builds trufflehog-<repo>-<random> subdirectory", func(t *testing.T) {
 		base := t.TempDir()
 		path, err := createClonePath("https://github.com/org/repo.git", base)
 		assert.NoError(t, err)
-		assert.Equal(t, filepath.Join(base, "trufflehog-repo"), path)
+		assert.Equal(t, base, filepath.Dir(path))
+		// The trufflehog- prefix is what CleanTempDirsForLegacyJSON sweeps.
+		assert.True(t, strings.HasPrefix(filepath.Base(path), "trufflehog-repo-"),
+			"unexpected directory name %q", filepath.Base(path))
 
 		info, err := os.Stat(path)
 		assert.NoError(t, err)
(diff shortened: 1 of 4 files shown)


pkg/sources/git/git_test.go
```diff
@@ -141,50 +144,161 @@ func TestCreateClonePath(t *testing.T) {
...
+		assert.NoError(t, os.WriteFile(base, []byte("x"), 0644))
 
 		path, err := createClonePath("https://github.com/org/repo.git", base)
 		assert.Error(t, err)
 		assert.Empty(t, path)
 		assert.Contains(t, err.Error(), "failed to create clone path")
 	})
+
+	t.Run("distinct repos sharing a basename get distinct paths", func(t *testing.T) {
+		// Only the last URL segment is used as the directory slug, so repos
```
MuneebUllahKhan222, line 195: I think the comment should explain what the code does not what it fixes.

MuneebUllahKhan222 (approved):
Good test coverage. LGTM overall.

unsmith (approved):
This looks fine, but I do have a minor concern about the disk filling up with N workers all using their own clone path.  I also don't have any decent ideas on how to mitigate this; i think it's just going to be a thing.  Hopefully not many people are using `--no-cleanup`, or if they do, they understand the risks.

Maybe a warning in the docs, asking the user to verify they have sufficient disk space to clone their repos?

robersonroland17-ux:
Apple tv mls start Adobe dead

PR #5178, opened 2026-08-05
make cwd-changing normalization tests non-parallel to fix flaky clone failures

jordanTunstill:
### Description:
stop TestPrepareRepoWithNormalizationBare from running in parallel, to fix test that was failing PRs.
```
OSS prs are failing the test pull-request test on this error:
failed to clone file Git repo (file:///tmp/TestGitConfigSanitization508720935/001/test-repo): could not clone repo: file:///tmp/TestGitConfigSanitization508720935/001/test-repo, error executing git clone: exit status 128, fatal: Unable to read current working directory: No such file or directory
``` 

The test chdir'd into its own t.TempDir while parallel, so concurrently
forked `git clone` subprocesses inherited a working directory that was
deleted on cleanup, failing with "Unable to read current working
directory". Switched both normalization tests to t.Chdir so the
directory is restored before cleanup removes it.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Test-only changes to working-directory handling; no production git source behavior is modified.
> 
> **Overview**
> Fixes CI flakes where parallel git normalization tests changed the process working directory into a `t.TempDir()` that could be removed before child `git clone` processes finished, causing **"Unable to read current working directory"**.
> 
> **`TestPrepareRepoWithNormalization`** and **`TestPrepareRepoWithNormalizationBare`** now use **`t.Chdir`** instead of manual `os.Chdir` plus defer restore, so the test harness restores cwd before temp cleanup. **`TestPrepareRepoWithNormalizationBare`** no longer calls **`t.Parallel()`**, and its setup is reordered so the test chdirs into its temp directory before running bare-repo clone scenarios (including relative `file://` URIs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51364b74acf972503c39bccf57b7cb13dccb66d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/git/git_test.go
+++ b/pkg/sources/git/git_test.go
@@ -1286,19 +1286,14 @@ func TestNormalizeFileURI(t *testing.T) {
 	}
 }
 
+// Relative file URIs resolve against the process working directory, so this
+// test has to chdir and cannot run in parallel: every other test in this
+// package shells out to git, and a git subprocess that inherits a working
+// directory this test later removes fails with "Unable to read current
+// working directory".
 func TestPrepareRepoWithNormalization(t *testing.T) {
 	repoPath := setupTestRepo(t, "test-repo")
 	addTestFileAndCommit(t, repoPath, "test.txt", "test content")
-
-	originalDir, err := os.Getwd()
-	if err != nil {
-		t.Fatalf("failed to get current directory: %v", err)
-	}
-	defer func() {
-		if err := os.Chdir(originalDir); err != nil {
-			t.Fatalf("failed to restore original directory: %v", err)
-		}
-	}()
 
 	// Test absolute paths first (without changing directory)
 	absoluteTests := []struct {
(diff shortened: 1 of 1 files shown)


mariduv (approved):
The only changes I would make here are that the block comment is unnecessary with the change to use the proper t.Chdir() helper, which will panic if someone comes back to try adding t.Parallel(), so things don't need a redundant explanation beyond the method's docs and the tests breaking with a message that you can't do both if you try.  And also, that while here I would try to move the t.Chdir() as near to the top of the function as possible, so it's immediately clear this test will change the process's directory.

jordanTunstill:
> The only changes I would make here are that the block comment is unnecessary with the change to use the proper t.Chdir() helper, which will panic if someone comes back to try adding t.Parallel(), so things don't need a redundant explanation beyond the method's docs and the tests breaking with a message that you can't do both if you try. And also, that while here I would try to move the t.Chdir() as near to the top of the function as possible, so it's immediately clear this test will change the process's directory.

thanks Meredith! I made both changes.

PR #5180, opened 2026-08-05
Anthropic detector: surface API error detail on non-2xx verification

Ahsan-Sarbaz:
### Description:
The Anthropic detector treated any non-200/401/404 response as a plain "unexpected HTTP response status" error. A 400 from the API is `invalid_request_error` and never a bad-key signal (bad keys return 401), so it must stay indeterminate rather than being read as not-live, but the error gave no way to tell why a request was rejected.

This enriches that error with the API's own `error.type`/`error.message` when the body carries them, so a 400 (or any other unexpected status) is diagnosable from the log alone. Also drains the response body before closing it, drops a dead branch in the key-type dispatch, and adds test coverage for the verification status codes (200/401/404/400/429/500) plus a malformed-body fallback.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped to Anthropic detector verification and logging; behavior for 401/404 unchanged, with added tests for edge statuses.
> 
> **Overview**
> Improves **Anthropic API key verification** when the HTTP response is not 200, 401, or 404. Those unexpected statuses still leave the key **unverified** (not treated as definitively dead), but verification errors now include Anthropic’s **`error.type` and `error.message`** when the response body is valid JSON.
> 
> Also **propagates errors** from building the outbound request (instead of dropping them), **drains the response body** before close, and **simplifies** admin vs API key dispatch by removing an unreachable branch.
> 
> Adds **`TestAnthropic_VerifyMatch`** covering 200/401/404/400/429/500 and non-JSON error bodies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 447b1ea80be02ade2cd271f82a4763e4d749ab87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/anthropic/anthropic.go
+++ b/pkg/detectors/anthropic/anthropic.go
@@ -83,15 +83,8 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	return results, nil
 }
 
-/*
-verifyAnthropicKey verify the anthropic key passed against the endpoint
-
-Endpoints:
-
-  - For api keys: https://docs.anthropic.com/en/api/models-list
-
-  - For admin keys:  https://docs.anthropic.com/en/api/admin-api/apikeys/list-api-keys
-*/
+// verifyAnthropicKey verifies key against endpoint (models-list for API keys,
+// admin api_keys-list for admin keys).
 func verifyAnthropicKey(ctx context.Context, client *http.Client, endpoint, key string) (bool, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, http.NoBody)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


kashifkhan0771 (approved):
LGTM!

pkg/detectors/anthropic/anthropic.go
shahzadhaider1: Since we are at it, we should probably fix this as well: request construction due to any failure should report the error, instead of falsely claiming the secret as not-live, which the caller records as a determinate not-live.

shahzadhaider1 (commented):
Your changes are sound. I just have one concern

unsmith (approved):

PR #5186, opened 2026-08-07
fix twilio detectortype bug

lukem-ts:
### Description:
The `TwilioApiKey` detector was incorrectly referencing `DetectorType: ... Twilio`.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Single-field metadata fix in one detector with no change to matching or verification logic.
> 
> **Overview**
> Fixes a labeling bug in the **Twilio API key** scanner: findings from `FromData` now set `DetectorType` to `DetectorType_TwilioApiKey` instead of `DetectorType_Twilio`.
> 
> That matches `Scanner.Type()` and verification/dedup paths that already use the API-key detector enum, so reported secrets are classified as Twilio API keys rather than the separate Twilio detector.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2f2e7ceed64df8bf59e4e32e895719fef9d89c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/twilioapikey/twilioapikey.go
+++ b/pkg/detectors/twilioapikey/twilioapikey.go
@@ -67,7 +67,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	for apiKey := range uniqueAPIKeys {
 		for secret := range uniqueSecrets {
 			s1 := detectors.Result{
-				DetectorType: detector_typepb.DetectorType_Twilio,
+				DetectorType: detector_typepb.DetectorType_TwilioApiKey,
 				Raw:          []byte(apiKey),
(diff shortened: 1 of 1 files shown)


shahzadhaider1 (commented):
Yup, this needs fixing.

One thing worth flagging before it merges though: on the enterprise side we key a secret's identity partly on its detector type, so changing it means the same credential won't match what's already stored. We'd end up with a second record for a secret we already have. The old record just sits there and goes stale since nothing updates it anymore.

I’d suggest checking with the Scanning team first, since I believe they handle the storage and lifecycle of secrets.

PR #5188, opened 2026-08-07
Update regex pattern and 403 handling in the Lob detector to fix false positives

### LINKED ISSUE #5184: Lob detector reports Verified for non-credentials: 422/403 treated as verification, and the pattern has no entropy check
## Summary

The Lob detector reports `Verified: true` for strings that are not credentials. Two independent defects combine: the pattern matches ordinary source identifiers, and the verifier treats error responses that any request would produce as proof of a live key.

The practical effect is that **any Python `unittest` method whose name is exactly 40 characters is reported as a verified live Lob credential.** `test_` is the mandatory prefix for unittest methods, so this is a systematic collision rather than a coincidence — it can occur in any Python test suite.

Detector: `pkg/detectors/lob/lob.go`, added in #4971.

## Defect 1 — the verifier accepts non-credentials

`verify()` sends `POST https://api.lob.com/v1/us_verifications` with HTTP Basic auth (candidate as username, empty password) and an empty request body, then:

| status | result |
|---|---|
| 403 | `Verified = true` |
| 422 | `Verified = true` |
| 401 | `Verified = false` |
| other | error |

`422 Unprocessable Entity` is the natural response to a POST with an empty body — it reports that the *request* is malformed, and is not evidence about the *credential*. #4971's own description notes 422 is expected for an empty POST. Treating it as proof of an active key means the verification step cannot fail for any candidate that reaches it and does not return 401.

`403 → Verified = true` looks similarly inverted: a forbidden response is being read as confirmation.

The net effect is that `verify()` confirms the *shape* of a candidate rather than its *validity*, which removes the distinction the verified flag exists to provide.

## Defect 2 — the pattern has no entropy check

```go
\b((live|test)_[a-zA-Z0-9_]{35})\b
```

Any 40 word-characters beginning `test_` or `live_` match. There is no entropy or character-class-mix requirement, so ordinary identifiers qualify — underscores and lowercase letters are sufficient.

Python's `unittest` requires test methods to begin with `test_`, which makes descriptive 40-character method names a recurring natural match. A synthetic example that matches:

```python
class ExampleTests(unittest.TestCase):
    def test_returns_none_when_payload_is_bad_ok(self):   # exactly 40 chars
        ...
```

Combined with Defect 1, that method name is reported as a **verified live credential**.

## Reproduction

1. Create a Python file containing a method named `test_returns_none_when_payload_is_bad_ok`.
2. Run trufflehog in filesystem mode over it.
3. The Lob detector reports a finding with `Verified: true`.

## Observed

Three separate identifiers of this kind — all plain unittest method names in ordinary test files, none of them credentials, in a codebase with no Lob account or Lob integration of any kind — were each reported as verified. Three for three. Results were fully deterministic across repeated runs.

That determinism is worth stating because it rules out the more comfortable explanation: this is not a flaky network verification. The verifier is behaving exactly as written, and what is written accepts non-credentials.

## Suggested fixes

1. **Do not treat 422 or 403 as verification.** Restrict `Verified = true` to a response that can only be produced by a valid key. If no such response is available from this endpoint without a well-formed body, sending a minimal valid body would let 200-vs-401 carry the signal.
2. **Add an entropy or character-mix requirement** to the pattern. Real Lob keys are high-entropy; identifiers made of lowercase words and underscores are not, and the two are separable.

Either fix alone removes most of the false positives; both together would close the class.

## Secondary observation, less thoroughly verified

In the same runs, each match appeared **twice** with different line numbers — once at its true line and once a few lines earlier — apparently from a second decoder pass (HTML entity decoding). The files involved contained no HTML entities, so the decode should have been a no-op on identical bytes. This turned 3 matches into 6 reported findings.

Flagging this separately and with lower confidence than the two defects above: it was observed but not traced to specific code, and it may warrant its own issue if it reproduces for you.

## Impact

For consumers who gate CI on trufflehog, a verified finding is normally treated as the highest-severity signal and can block merges. A detector that verifies plain identifiers turns routine test-method naming into a build failure, and — more corrosively — it teaches people that "verified" findings are often wrong, which erodes trust in the flag for every other detector that computes it correctly.

### LINKED ISSUE #5187: lob: keyPat false-positives on snake_case identifiers same length as key
The Lob detector regex `\b((live|test)_[a-zA-Z0-9_]{35})\b` allows underscores in the 35-char suffix. Real keys appear to be alphanumeric only (see [test fixtures](https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/lob/lob_test.go#L15-L16)), so any snake_case identifier of the right length collides — e.g. a function name like `test_calculate_total_price_with_tax_rate` (35 chars after `test_`) matches and gets flagged as a raw finding.

Fix: drop `_` from the suffix char class (`[a-zA-Z0-9]{35}`), regression test included. Happy to open a PR.

MachiavelliII:
Fixes #5187

### Description

Two independent defects combined to make this detector report non-credentials as verified.

**Pattern.** `\b((live|test)_[a-zA-Z0-9_]{35})\b` matched any 40-character identifier beginning `live_`/`test_`, so ordinary snake_case names collided with it - Python's unittest mandates a `test_` prefix, which made the collision systematic rather than coincidental. Real Lob secret keys are 35 lowercase hex characters after the prefix. Tightened to `[a-f0-9]{35}`, which also rejects the `camelCase` collisions an alphanumeric class would still match. Publishable keys (`(live|test)_pub_` + 31 hex) no longer match; they are meant to be embedded in client-side code and are not secrets.

**Verification.** 403 was treated as proof of an active key, on the assumption it means "key is valid, no billing method on file". The status code alone cannot separate a valid key from an invalid one. 403 now branches on `error.code` - `billing_address_required` and `feature_limit_reached` mean the key authenticated and the account was rejected, `invalid_api_key` means it didn't, an unrecognised code is a verification error, and a body that isn't Lob's error JSON is reported as an invalid API response rather than as a verdict either way. 422 and 401 are unchanged.

### Checklist:
* [x] Tests passing (`go test ./pkg/detectors/lob/...`) — targeted package tests pass, including the new regression test. The full `make test-community` run didn't finish locally (large dependency graph, slow on this machine) but CI will run it.
* [X] Lint passing (`make lint`)? 

---

> [!NOTE]
> **Low Risk**
> Scoped to the Lob detector’s matching and verification logic; changes reduce false positives rather than altering core product auth or data paths.
> 
> **Overview**
> Fixes false-positive Lob secret detection and incorrect “verified” results for non-credentials.
> 
> **Pattern matching** now requires 35 lowercase hex digits after `live_`/`test_` (was any alphanumeric/underscore), so identifiers like Python `test_*` function names no longer match. Publishable `*_pub_*` keys are intentionally excluded from detection.
> 
> **API verification** no longer treats every HTTP 403 as proof of a valid key. On 403, the response body is parsed for Lob’s `error.code`: `billing_address_required` and `feature_limit_reached` still count as verified (key authenticated), `invalid_api_key` as unverified, unknown codes and non-JSON bodies surface errors instead of a wrong verdict. 422/401 behavior is unchanged.
> 
> Tests cover the new pattern cases and a mocked `verify` matrix for status codes and error payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1a7b0a11b76ab640a4f86eef1a5335d96475e6f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


duy0699cat (approved):

pkg/detectors/lob/lob.go
```diff
@@ -86,10 +89,27 @@ func (s Scanner) verify(ctx context.Context, key string) (bool, error) {
...
+		var body struct {
+			Error struct {
+				Code string `json:"code"`
+			} `json:"error"`
+		}
+		if err := json.NewDecoder(res.Body).Decode(&body); err != nil {
+			return false, errors.New("invalid API response")
+		}
+		switch body.Error.Code {
+		case "billing_address_required", "payment_method_unverified", "feature_limit_reached":
```
unsmith: One tweak needed here - `payment_method_unverified` is a 401, not 403: https://help.lob.com/developer-docs/error-reference#error-codes-advanced-6

unsmith (changes requested):
Just need to tweak the handling of `payment_method_unverified`, otherwise looks good.

pkg/detectors/lob/lob.go
shahzadhaider1: Looks like this is unreachable, let me remove `payment_method_unverified` from 403.

MuneebUllahKhan222 (approved):
LGTM!

unsmith (approved):

PR #5195, opened 2026-08-09
docs: add generic config-secret custom detector example

### LINKED ISSUE #4957: Feature Request: TruffleHog git scan does not detect generic hardcoded configuration secrets in .properties files
## Description

While integrating TruffleHog into our CircleCI pipeline, I noticed that certain hardcoded configuration secrets are not being detected during repository scans.

Below is an example:

```properties
knox.auth.client.secret=Ntde8=qa.wi6
jwt.secretkey=abcdsecret
test.auth.password=hello@01
```

The following command does not report any findings for the above values:

```bash
trufflehog git file://. --fail
```

I also tested similar hardcoded secrets in `.properties` files and observed that they are not detected during scans.

I would like clarification on whether this is expected behavior, whether additional configuration is required, or whether support for detecting generic hardcoded configuration secrets is planned.

---

### Preferred Solution

It would be helpful if TruffleHog could optionally detect common hardcoded configuration secrets such as:

```properties
*.password=
*.secret=
*.client.secret=
*.api.key=
```

especially within:

* `.properties`
* `.yaml`
* `.env`
* application configuration files

Ideally this could be:

* configurable
* optional
* rule-based
* with allowlist support to reduce false positives

This would improve detection coverage for CI/CD pipelines where teams want to prevent accidental commits of hardcoded configuration credentials.

---

### Additional Context

Environment:

* CircleCI pipeline
* Docker image: `trufflesecurity/trufflehog:latest`

Current scan command:

```bash
trufflehog git file://. --fail
```

The same behavior is observed both locally and in CircleCI.

The request is mainly to understand:

* whether this behavior is intentional
* whether additional detectors/rules are recommended
* or whether this would be considered a potential enhancement request

#### References

None currently.

pgmpofu:
## Summary

- Custom Detectors already support omitting the `verify` webhook block — matches are then simply reported as unverified. This makes them usable for flagging generic hardcoded secrets (e.g. `*.password=`, `*.secret=`) in config files like `.properties`, `.env`, or `.yaml`, which TruffleHog's built-in, verified detectors intentionally don't cover.
- Adds `examples/generic_config_secrets.yml`, tuned for that use case: looser matching than `examples/generic_with_filters.yml` (which requires the match to contain both a digit and a special character), with an entropy filter and exclude rules for common placeholders (`changeme`, `${VAR}`, `{{ .template }}`, env-var references, etc.) to keep false positives down.
- Links to the new example from `README.md` and `pkg/custom_detectors/CUSTOM_DETECTORS.md`, right next to the existing docs explaining that `verify` is optional.

Addresses #4957, where a user reported that a default `trufflehog git ... --fail` scan doesn't flag hardcoded secrets in a `.properties` file. That's expected (no built-in detector targets arbitrary config keys), but the ask can already be met with Custom Detectors — this PR just makes that path discoverable with a ready-to-use example.

## Test plan

- [x] Built from source and reproduced the issue: default scan of a `.properties` file containing `knox.auth.client.secret=...`, `jwt.secretkey=...`, `test.auth.password=...` reports 0 findings.
- [x] Same file scanned with `--config=examples/generic_config_secrets.yml` reports all 3 as unverified findings.
- [x] Verified the exclude rules correctly skip common false-positive patterns in the same file: `${DB_PASSWORD}`, `changeme`, `{{ .Values.token }}`, and a plain `true` value.
- [x] Docs-only change elsewhere; no code paths modified.

---

> [!NOTE]
> **Low Risk**
> Documentation and example YAML only; no runtime or detector code changes.
> 
> **Overview**
> Documents that Custom Detectectors work **without** a `verify` webhook (findings stay unverified) and adds a ready-to-use example for catching hardcoded secrets in config files (`.properties`, `.env`, `.yaml`) that built-in verified detectors miss.
> 
> **`examples/generic_config_secrets.yml`** defines `generic-config-secret`: keyword-triggered regex for `key=value` / `key: value` patterns, entropy `2.5`, and capture-level excludes for placeholders (`changeme`, `${VAR}`, `{{ templates }}`, env refs, vault URLs, etc.)—looser than `generic_with_filters.yml`, which enforces digit/special-char validations and no `verify` block.
> 
> **README** and **`pkg/custom_detectors/CUSTOM_DETECTORS.md`** now call out the optional-`verify` workflow and link to the new example beside the existing filter docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75c4ad6489f12eda0355010928664873befab3b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -798,6 +798,8 @@ Custom Detectors support a few different filtering mechanisms: entropy, regex ta
 and excluded word lists checked against the secret (captured group if present, entire match if capture group is not present). Note that if
 your custom detector has multiple `regex` set (in this example `hogID`, and `hogToken`), then the filters get applied to each regex. [Here](examples/generic_with_filters.yml) is an example of a custom detector using these filters.
 
+The `verify` section is optional — if you omit it, matches are still reported as unverified, no webhook required. This makes Custom Detectors useful for flagging generic hardcoded secrets (e.g. `*.password=`, `*.secret=`) in config files like `.properties`, `.env`, or `.yaml` that TruffleHog's built-in, verified detectors won't otherwise catch. [Here](examples/generic_config_secrets.yml) is an example tuned for that use case.
+
 **NB:** This feature is alpha and subject to change.
 
 ### Regex Detector Example
--- a/examples/generic_config_secrets.yml
+++ b/examples/generic_config_secrets.yml
@@ -1,0 +1,44 @@
+detectors:
+- name: generic-config-secret
+  keywords:
+  - secret
+  - password
+  - passwd
+  - pwd
+  - apikey
+  - api_key
+  - api-key
+  - token
+  - credential
+  - cred
+  - auth
+  regex:
+    secret: |-
+      (?i)[\w.-]{0,50}?(?:secret|passw(?:or)?d|pwd|api[_-]?key|token|credentials?|creds?|auth)[\w.-]{0,20}\s*[=:]\s*["'`]?([^\s"'`,;]{6,150})["'`]?\s*(?:$|[\r\n#;])
+  entropy: 2.5
+  exclude_regexes_match:
+  - '(?i)\$\{[^}]*\}'
+  - '(?i)\{\{[^}]*\}\}'
+  - '(?i)%[A-Z_]+%'
+  - '(?i)\$[A-Z_][A-Z0-9_]*'
+  - '(?i)(?:env|process\.env|os\.environ|getenv|System\.getenv)\b'
+  - '(?i)(?:vault|kms|secretsmanager|parameterstore)://'
+  exclude_words:
+  - "changeme"
+  - "change_me"
+  - "change-me"
+  - "changethis"
+  - "xxxxxx"
+  - "example"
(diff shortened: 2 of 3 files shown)


pgmpofu:
Good catch — fixed in 75c4ad6: moved the placeholder/env-reference excludes from `exclude_regexes_match` to `exclude_regexes_capture` so they only inspect the captured value, not the whole `key=value` line. Verified `env.password=<realsecret>` and `app.env.secret=<realsecret>` are now correctly detected, while `${DB_PASSWORD}`, `process.env.X`, etc. are still excluded.

pgmpofu:
Friendly nudge — this has been sitting for about a week with all checks green (CLA signed, Bugbot's finding addressed in 75c4ad6). @trufflesecurity/integrations, would appreciate a look when you get a chance. Thanks!

kashifkhan0771 (approved):
LGTM!

unsmith (approved):

PR #5203, opened 2026-08-11
fix(blogger): anchor regex to AIzaSy prefix to eliminate false positives

unsmith:
### Description:
Fix false positive in the Blogger detector triggered by Java class names (e.g. `EJBLogger.logfinderReturnedMultipleValuesLoggable`). The regex `[0-9A-Za-z-]{39}` matches any 39-character alphanumeric string within 40 characters of "blogger" (case-insensitive). This false positive above just so happened to be exactly 39 characters between word boundaries, so it was detected.

All Google API keys start with `AIzaSy`, the fix anchors the capture group to that prefix and also adds `_` to the character class since Google keys can contain underscores.

A note on the Corpora "regression" - the test files were adjusted to start looking for the right key prefixes, so while they were technically removed, this is expected I believe, since the corpus of public code hopefully does not contain any exposed Blogger API keys.

### Checklist:

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit e385039ffc4d243bbead75908ac203832af25fba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/blogger/blogger.go
+++ b/pkg/detectors/blogger/blogger.go
@@ -22,8 +22,10 @@ var _ detectors.Detector = (*Scanner)(nil)
 var (
 	client = common.SaneHttpClient()
 
-	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
-	keyPat = regexp.MustCompile(detectors.PrefixRegex([]string{"blogger"}) + `\b([0-9A-Za-z-]{39})\b`)
+	// Google API keys always start with "AIzaSy" (6-char prefix + 33 remaining = 39 total).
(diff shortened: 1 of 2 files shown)


MuneebUllahKhan222 (approved):
LGTM!

martinlocklear (approved):

martinlocklear:
@unsmith - Can you please investigate [this comment](https://github.com/trufflesecurity/trufflehog/pull/5203#issuecomment-5257040357) and see if it's worth addressing before merging?

PR #5204, opened 2026-08-11
fix(huggingface): move unscoped fine-grained models to UnboundedResources

unsmith:
### Description:

The HuggingFace analyzer produces bindings with empty `Permission.Value` for fine-grained tokens when a model's owner isn't in the token's scoped permissions. This causes `CreateAnalysisResults` to reject the request because the proto validation requires `Permission.Name` to be at least 1 rune.

The root cause is that `bakefineGrainedBindings()` creates a `Binding` for every model returned by the `/api/models` API, but the HuggingFace API returns all visible models (including public ones) regardless of the token's fine-grained scope. Models outside the scope end up with a zero-value `Permission{}`.

The fix splits models into two groups: those that resolve a non-empty permission become `Binding` entries, and those without a matching scope go into`UnboundedResources`.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Localized HuggingFace analyzer output shaping with a focused regression test; no auth or shared infrastructure changes.
> 
> **Overview**
> Fixes **fine-grained HuggingFace** analysis so models that are visible via the API but **not covered by the token’s scoped permissions** no longer produce `Binding` entries with empty `Permission.Value` (which broke downstream **proto validation** on `CreateAnalysisResults`).
> 
> `bakefineGrainedBindings` now returns both bindings and extra resources: models that resolve a non-empty permission stay in **Bindings**; everything else is appended to **`UnboundedResources`** from `secretInfoToAnalyzerResult`. A unit test locks in scoped vs unscoped model placement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63c3f3af5cb58114c3700d0274a83bb99703ef87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/analyzer/analyzers/huggingface/huggingface_test.go
+++ b/pkg/analyzer/analyzers/huggingface/huggingface_test.go
@@ -118,3 +118,132 @@ func TestAnalyzer_Analyze(t *testing.T) {
 		})
 	}
 }
+
+// Test_secretInfoToAnalyzerResult_FineGrainedUnboundedResources verifies that
+// models outside the token's fine-grained permission scopes land in
+// UnboundedResources rather than producing bindings with empty Permission.Value.
+func Test_secretInfoToAnalyzerResult_FineGrainedUnboundedResources(t *testing.T) {
+	info := &SecretInfo{
+		Token: HFTokenJSON{
+			Username: "testuser",
+			Name:     "Test User",
+			Auth: struct {
+				AccessToken struct {
+					Name        string `json:"displayName"`
+					Type        string `json:"role"`
+					CreatedAt   string `json:"createdAt"`
+					FineGrained struct {
+						Global []string `json:"global"`
+						Scoped []struct {
+							Entity struct {
+								Type string `json:"type"`
+								Name string `json:"name"`
+								ID   string `json:"_id"`
+							} `json:"entity"`
+							Permissions []string `json:"permissions"`
+						} `json:"scoped"`
+					} `json:"fineGrained"`
+				}
+			}{
+				AccessToken: struct {
+					Name        string `json:"displayName"`
+					Type        string `json:"role"`
+					CreatedAt   string `json:"createdAt"`
+					FineGrained struct {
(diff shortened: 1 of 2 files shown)


pkg/analyzer/analyzers/huggingface/huggingface_test.go
```diff
@@ -118,3 +118,132 @@ func TestAnalyzer_Analyze(t *testing.T) {
...
+		}
+	}
+	if !foundScoped {
+		t.Error("expected testuser/scoped-model in Bindings")
+	}
+
+	// Unscoped models should appear in UnboundedResources, not Bindings.
+	unscopedNames := map[string]bool{
+		"other-org/unscoped-model": false,
+		"another-org/also-unscoped": false,
```
shahzadhaider1: Is this gofmt-clean?

shahzadhaider1 (approved):

pkg/analyzer/analyzers/huggingface/huggingface_test.go
unsmith: Interesting, that seems to have escaped my linter checks.  Fixed.

pkg/analyzer/analyzers/huggingface/huggingface_test.go
```diff
@@ -118,3 +118,132 @@ func TestAnalyzer_Analyze(t *testing.T) {
 		})
 	}
 }
+
+// Test_secretInfoToAnalyzerResult_FineGrainedUnboundedResources verifies that
+// models outside the token's fine-grained permission scopes land in
+// UnboundedResources rather than producing bindings with empty Permission.Value.
+func Test_secretInfoToAnalyzerResult_FineGrainedUnboundedResources(t *testing.T) {
+	info := &SecretInfo{
```
martinlocklear, line 126: Tabular tests, gross

martinlocklear (approved):

PR #5206, opened 2026-08-12
feat(detectors): enforce endpoint order, support multi-cloud endpoints, fix stop-on-verify bugs

kashifkhan0771:
### Description:
Endpoints() now always tries configured, then cloud, then found endpoints in that fixed order, with an optional per-call suffix filter on found endpoints to guard against SSRF from scanned text. CloudProvider now returns []string, letting detectors declare multiple cloud endpoints; migrated 4 NewRelic detectors and sumologickey off ad-hoc endpoint logic onto this shared path, and fixed a latent bug where datadogapikey/datadogtoken could silently overwrite a verified result with a later failed endpoint check.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes verification target selection and ordering for many detectors, including SSRF-related filtering of scanned URLs; incorrect suffix or ordering could mis-verify secrets or skip valid hosts.
> 
> **Overview**
> Refactors detector endpoint handling so verification always tries **configured → cloud → found** endpoints, with deduplication across sources and found URLs only accepted when they pass an optional per-call host suffix filter (SSRF mitigation).
> 
> **`CloudProvider` / `EndpointCustomizer`** now expose plural APIs (`CloudEndpoints()`, `SetCloudEndpoints`, `UseCloudEndpoints`) and `Endpoints(found []string, allowedSuffixes ...)` instead of variadic found URLs and a single cloud URL. **`EndpointSetter`** implements the new ordering, multi-cloud lists, and suffix checks via `hostOf`.
> 
> Call sites across Artifactory, GitHub/GitLab, Datadog, Jira, Vault, Tableau, Atlassian DC helpers, engine defaults, and custom-verifier wiring are updated to the new signatures.
> 
> **Behavior fixes:** Datadog API key/token verification stops after the first successful endpoint so a later failure cannot overwrite a verified result. **Sumo Logic** routes hosts through `Endpoints()` and adds `confidentEndpoint` plus scheme normalization for `--verifier-endpoint` URLs. **Four New Relic detectors** adopt `EndpointSetter` with US/EU cloud endpoints instead of hard-coded regional loops.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23ce397016099327a4702929b34345aabc03671e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/endpoint_customizer.go
+++ b/pkg/detectors/endpoint_customizer.go
@@ -2,18 +2,22 @@ package detectors
 
 import (
 	"fmt"
+	"net/url"
+	"strings"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
 )
 
 // EndpointSetter implements a sensible default for the SetEndpoints function
 // of the EndpointCustomizer interface. A detector can embed this struct to
-// gain the functionality.
+// gain the functionality. See Endpoints for the order sources are tried in.
 type EndpointSetter struct {
 	configuredEndpoints []string
-	cloudEndpoint       string
-	useCloudEndpoint    bool
-	useFoundEndpoints   bool
+
+	cloudEndpoints   []string
+	useCloudEndpoints bool
+
+	useFoundEndpoints bool
 }
 
 func (e *EndpointSetter) SetConfiguredEndpoints(userConfiguredEndpoints ...string) error {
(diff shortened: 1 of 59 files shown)


pkg/detectors/endpoint_customizer.go
```diff
@@ -40,13 +54,83 @@ func (e *EndpointSetter) UseFoundEndpoints(enabled bool) {
...
+			if foundEndpointAllowed(found, allowedFoundSuffixes) {
+				add(found)
+			}
+		}
 	}
+
 	return endpoints
 }
+
+func foundEndpointAllowed(rawURL string, allowedSuffixes []string) bool {
```
MuneebUllahKhan222, line 102: A suffix allow list for `foundEndpoints` doesn't make much sense because the suffix check can be done through regex use to find these endpoints.

pkg/detectors/endpoint_customizer.go
```diff
@@ -40,13 +54,83 @@ func (e *EndpointSetter) UseFoundEndpoints(enabled bool) {
...
+// verification at an arbitrary host (SSRF). Configured and cloud endpoints
+// are never filtered - they're operator/developer trusted, not scanned text.
+func (e *EndpointSetter) Endpoints(foundEndpoints []string, allowedFoundSuffixes ...string) []string {
+	endpoints := make([]string, 0, len(e.configuredEndpoints)+len(e.cloudEndpoints)+len(foundEndpoints))
+	seen := make(map[string]struct{}, cap(endpoints))
+	add := func(endpoint string) {
+		if _, ok := seen[endpoint]; ok {
+			return
+		}
+		seen[endpoint] = struct{}{}
```
MuneebUllahKhan222, line 73: I always wonder if we should do `url.Parse` to make sure an endpoint is a valid `url` before appending it.

MuneebUllahKhan222 (changes requested):
This is an initial review of the changes in this PR. I haven’t reviewed the individual detector changes yet; so far, I’ve only reviewed the main `endpoint_customizer` changes.

pkg/detectors/endpoint_customizer.go
kashifkhan0771: Yes - that's why it's optional. Some detectors found-endpoint regex matches broad domain patterns and don't restrict suffix. Editing regex per-detector is riskier (easy to break matching). Suffix filter at Endpoints() call site is safer and reusable across detectors than baking it into each regex.

pkg/detectors/endpoint_customizer.go
kashifkhan0771: We could, but url.Parse is very lenient - it rarely errors even on garbage input, so it wouldn't really catch invalid endpoints. Also endpoints here can be either bare hostnames or full URLs by design, so a strict URL check doesn't fit well

pkg/detectors/endpoint_customizer.go
```diff
@@ -2,18 +2,22 @@ package detectors
...
-// gain the functionality.
+// gain the functionality. See Endpoints for the order sources are tried in.
 type EndpointSetter struct {
 	configuredEndpoints []string
-	cloudEndpoint       string
-	useCloudEndpoint    bool
-	useFoundEndpoints   bool
+
+	cloudEndpoints   []string
+	useCloudEndpoint bool
```
camgunz: **suggestion:** _Could_ make this `useCloudEndpoints`

PR #5207, opened 2026-08-12
fix(action): warn when an unhandled event falls back to a full-history scan

### LINKED ISSUE #5190: GitHub Action: unhandled event types silently fall through to a full-history scan (BASE/HEAD left empty)
## Summary

The action's event dispatch has branches for `push`, `workflow_dispatch`/`schedule`, and `pull_request`. Any other `github.event_name` falls out of the ladder with `BASE` and `HEAD` still unset, and the scan then runs with empty `--since-commit` and `--branch`. Combined with the `fetch-depth: 0` checkout the action's own README recommends, the effect is a silent full-history scan across all fetched refs — where the user expected a diff scan, and with no indication in the log that the scope changed.

`release` is the case that bit us, but the same applies to `merge_group`, `pull_request_target`, `push_tag`, and anything else not in the list.

## The code

`action.yml`:

```bash
else
  if [ "${{ github.event_name }}" == "push" ]; then
    ...
  elif [ "${{ github.event_name }}" == "workflow_dispatch" ] || [ "${{ github.event_name }}" == "schedule" ]; then
    BASE=""
    HEAD=""
  elif [ "${{ github.event_name }}" == "pull_request" ]; then
    BASE=${{github.event.pull_request.base.sha}}
    HEAD=${{github.event.pull_request.head.sha}}
  fi
fi
...
docker run --rm -v .:/tmp -w /tmp \
"${IMAGE}:${VERSION}" \
git file:///tmp/ \
--since-commit \
${BASE:-''} \
--branch \
${HEAD:-''} \
```

There is no `else` on the inner ladder. For an unhandled event both variables stay empty and `${BASE:-''}` / `${HEAD:-''}` expand to empty strings.

Worth contrasting with the guard immediately above it, which handles the *opposite* failure explicitly:

```bash
## If BASE == HEAD, exit with error
echo "::error::BASE and HEAD commits are the same. TruffleHog won't scan anything. Please see documentation (...)"
```

"Scans nothing" gets a hard error. "Scans everything, on every ref, unasked" gets nothing.

## Observed

A `release: [published]` run, on a workflow whose other jobs are shared with its PR checks:

```
if [ "release" == "push" ]; then ...
elif [ "release" == "workflow_dispatch" ] || [ "release" == "schedule" ]; then ...
elif [ "release" == "pull_request" ]; then ...
fi
env:
  BASE:
  HEAD:
```

```
scanning repo {"unit": "/tmp/trufflehog-19-2005203197", "repo": "file:///tmp"}
finished scanning {"chunks": 5730, "bytes": 3110451, "verified_secrets": 20, ...}
##[error]Process completed with exit code 183
```

5730 chunks is the entire history, not the release diff.

## Why this is worse than a scope surprise

The findings a history scan surfaces are, by construction, things you cannot fix by changing code — they're immutable commits. So the failure is not actionable in the way a diff finding is, and it arrives at release time rather than at review time.

In our case all 20 were false positives from the Lob detector (#5184), the oldest from 2024. The PR that introduced each of those filenames had passed, correctly, because PR runs scan a diff. The same commits then blocked a release two years later. Two defects were needed to produce that outcome, and this is the one that decided *when* it landed and *how much* it surfaced.

It also makes the behaviour non-obvious to debug: identical code, identical action ref, one job passes and one fails, and nothing in the log says the scan scope differed.

## Suggested fix

Any of these would be an improvement, roughly in order of preference:

1. **Add an `else` that fails loudly**, mirroring the `BASE == HEAD` guard — `::error::` naming the unhandled `event_name` and pointing at the `base`/`head` inputs. Unhandled input should not silently become the most expensive possible behaviour.
2. **Handle `release` explicitly.** `github.event.release.tag_name` and the previous tag are both reachable, so a tag-to-tag diff is well defined.
3. **Log the effective scan scope unconditionally** — one line stating diff-vs-full and the resolved BASE/HEAD. Cheap, and it would have made this self-diagnosing.

Happy to open a PR for (1) plus (3) if you'd take it.

## Environment

- Action `trufflesecurity/trufflehog@v3.96.0`, scanner image `3.96.0`
- `actions/checkout@v7` with `fetch-depth: 0`
- `extra_args: --only-verified`
- Runner `ubuntu-24.04`

Haseeb-1698:
### Description:

Fixes #5190.

The event-dispatch ladder in `action.yml` has branches for `push`, `workflow_dispatch`/`schedule` and `pull_request`. Any other `github.event_name` falls out of the ladder with `BASE` and `HEAD` still unset, so the scan runs with empty `--since-commit` and `--branch`:

```bash
elif [ "${{ github.event_name }}" == "pull_request" ]; then
  BASE=${{github.event.pull_request.base.sha}}
  HEAD=${{github.event.pull_request.head.sha}}
fi                       # <- no else
...
--since-commit ${BASE:-''} --branch ${HEAD:-''}
```

Combined with the `fetch-depth: 0` checkout the README recommends, that silently turns an expected diff scan into a full-history scan across every fetched ref, with nothing in the log to indicate the scope changed. As the issue notes, `release` is one case; `merge_group`, `pull_request_target`, `create` and `workflow_run` behave the same way.

Worth contrasting with the guard immediately above, which handles the opposite failure explicitly — when `BASE == HEAD` it prints `::error::` and exits, because scanning *nothing* is worth shouting about. Scanning *everything*, unasked, was silent.

This adds the missing `else`. It sets `BASE`/`HEAD` explicitly so the fallback reads as intentional rather than incidental, and emits a `::warning::` naming the event that caused it and pointing at the `base`/`head` inputs.

**Deliberately a warning, not an error.** Erroring out would break any workflow currently relying on the fallback, and "scan everything" is sometimes the right outcome for these events — it just should not be invisible. The scan still runs; only the log gains a line.

### Verification

The change is confined to `action.yml`; no Go code is touched.

I modelled the dispatch ladder in a shell harness and ran every event type against the current and patched logic:

| Event | Before | After |
|---|---|---|
| `push` | `since-commit=<before-sha> branch=<after-sha>` | identical |
| `workflow_dispatch` | `since-commit=<> branch=<>` | identical |
| `schedule` | `since-commit=<> branch=<>` | identical |
| `pull_request` | `since-commit=<base-sha> branch=<head-sha>` | identical |
| `release` | `since-commit=<> branch=<>`, no warning | same scan, **warning emitted** |
| `merge_group` | as above | **warning emitted** |
| `pull_request_target` | as above | **warning emitted** |
| `create` | as above | **warning emitted** |
| `workflow_run` | as above | **warning emitted** |
| `issue_comment` | as above | **warning emitted** |

All four handled events are byte-identical, so there is no regression; every unhandled event now warns.

Also confirmed `action.yml` still parses as YAML, and the extracted `run:` body passes `bash -n`.

### Checklist:
* [ ] Tests passing (`make test-community`)? — not run: this change touches no Go code, only `action.yml`. Happy to run it if you would like the confirmation anyway.
* [ ] Lint passing (`make lint`)? — same; `golangci-lint` does not cover this file.
* [x] Man page — not affected; no CLI flags or subcommands added, removed or renamed.

---

> [!NOTE]
> **Low Risk**
> Composite action shell-only change; handled events are unchanged and unhandled events still scan but are now visible in logs.
> 
> **Overview**
> **Unhandled GitHub events** (`release`, `merge_group`, `pull_request_target`, etc.) no longer fall through the `action.yml` dispatch ladder with empty `BASE`/`HEAD` and no signal. An explicit `else` keeps the same full-history scan behavior but emits a **`::warning::`** naming the event and pointing users at the `base`/`head` inputs.
> 
> **Every run** now logs a one-line **scan scope** summary (diff vs full branch vs full fetched refs) with event name and resolved `since-commit` / `branch` values before TruffleHog starts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d315c36ee67b134c200b7f06f5d8c72c8d5020f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -93,6 +93,14 @@ runs:
           elif [ "${{ github.event_name }}" == "pull_request" ]; then
             BASE=${{github.event.pull_request.base.sha}}
             HEAD=${{github.event.pull_request.head.sha}}
+          else
+            # Unhandled event type (release, merge_group, pull_request_target, ...).
+            # BASE and HEAD stay empty, so the scan below covers the full history of
+            # every fetched ref rather than a diff. That may be what you want, but it
+            # should not be silent: say so, and name the event that caused it.
+            BASE=""
+            HEAD=""
+            echo "::warning::TruffleHog has no diff range for event '${{ github.event_name }}', so it is scanning the full history of all fetched refs. Set the 'base' and 'head' inputs explicitly to scan a specific range. Please see documentation (https://github.com/trufflesecurity/trufflehog#octocat-trufflehog-github-action)."
           fi
         fi
         ##########################################
--- a/action.yml
+++ b/action.yml
@@ -104,6 +104,19 @@ runs:
           fi
         fi
         ##########################################
+        ## Log the resolved scan scope          ##
+        ##########################################
+        if [ -n "$BASE" ] && [ -n "$HEAD" ]; then
+          SCOPE="diff"
+        elif [ -n "$HEAD" ]; then
+          SCOPE="full history of branch '$HEAD'"
+        elif [ -n "$BASE" ]; then
+          SCOPE="every commit after '$BASE'"
+        else
+          SCOPE="full history of all fetched refs"
+        fi
+        echo "TruffleHog scan scope: ${SCOPE} [event='${{ github.event_name }}', since-commit='${BASE}', branch='${HEAD}']"
+        ##########################################
         ##          Run TruffleHog              ##
         ##########################################
         docker run --rm -v .:/tmp -w /tmp \


shahzadhaider1 (commented):
Looks fine but I believe the ask no. 3 in the attached issue in still not addressed.

Haseeb-1698:
Right, I only did (1). (3) is pushed now.

```
TruffleHog scan scope: diff [event='pull_request', since-commit='<base>', branch='<head>']
TruffleHog scan scope: full history of all fetched refs [event='release', since-commit='', branch='']
```

It sits after the whole ladder, so it covers the explicit `base`/`head` inputs too, not just the event branches. Plain `echo` rather than `::notice::`, since this runs every time and an annotation on every run would be noise in the summary; the `::warning::` stays as the line that actually asks for action. Four states rather than two, because the ladder can resolve a `head` with no `base` (first push) and the inputs path can do the reverse.

Verified by modelling the ladder in a shell harness and running all 16 paths against the current and patched `action.yml`: the resolved `--since-commit` / `--branch` are byte-identical in every case, only the new log line appears.

I left (2), explicit `release` handling, out on purpose: a tag-to-tag diff changes scan behaviour rather than just reporting it, so it seemed better as its own PR. Happy to add it here if you would prefer it in one.

Haseeb-1698:
Hi @shahzadhaider1 — just checking in on this one.

Ask (3) from #5190 went in on 13 Aug: the action now logs the effective scan scope unconditionally, one line naming diff-vs-full plus the resolved BASE/HEAD, so both the handled and fall-through paths are self-diagnosing. That's alongside ask (1), the loud `::error::` for unhandled `event_name`.

All 11 checks are green and the branch is up to date with main. Happy to rebase, split it, or adjust the wording of the log line if you'd prefer it shaped differently — just let me know what would help.

Thanks for the review.

shahzadhaider1 (approved):

PR #5215, opened 2026-08-16
Carry GitHub App installation ID on repo units

bill-rich:
## Problem

In `scan_all_installations` mode, enumeration already resolves which GitHub App installation owns each repo (`enumerateAllInstallationRepos` records it as it lists each installation's repos). But `RepoUnit` only carried `{name, url}`, so that fact was lost at the enumerate to scan boundary.

Each SCAN job runs with a cold connector, so `ChunkUnit` re-derived the mapping via `mapReposToInstallations`, which lists every installation's repos to locate the single repo being scanned. At large multi org installations this runs once per scanned unit, which stalls scans on GitHub rate limits (INT-790).

## Change

* Add `InstallationID` to `RepoUnit` and record the resolved installation during enumeration (`processRepos`).
* In `ChunkUnit`, use the installation carried on the unit and only fall back to `mapReposToInstallations` when a unit has no installation. This preserves the existing listing based ownership semantics rather than guessing from the repo owner.

Units enumerated before this field existed have no installation and fall back to the current derivation path, so rollout is safe.

---

> [!NOTE]
> **Medium Risk**
> Changes GitHub App scan-all-installations token selection and source-unit unmarshalling. Legacy units still fall back to the old mapping path, but a bad ID or unmarshal bug could clone with the wrong installation token.
> 
> **Overview**
> Avoids re-listing every GitHub App installation’s repos on each SCAN job in `scan_all_installations` mode (INT-790). Enumeration already knew which installation owned a repo; that mapping was dropped because `RepoUnit` only had name and URL.
> 
> `RepoUnit` now carries `InstallationID`. Enumeration (`processRepos` and pre-configured repos) records it, and `ChunkUnit` applies it on the cold connector instead of calling `mapReposToInstallations`. Units without an ID (legacy payloads) still use the old listing-based fallback.
> 
> GitHub source-unit unmarshalling is custom so `unit_data` (and thus the installation ID) survives the enumerate→scan boundary, replacing `CommonSourceUnitUnmarshaller`. Tests cover round-trip, legacy units, and emission during enumeration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3f42157a0eb6fd3c504683f9c60f7ab5acf1edf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/sources/github/github.go
```diff
@@ -110,6 +118,74 @@ type GistUnit struct {
...
+				var unit RepoUnit
+				if err := json.Unmarshal(decoded, &unit); err == nil && unit.URL != "" {
+					return unit, nil
+				}
+			}
+		}
+	}
+
+	// Full enumeration payload: the unit JSON itself, with no separate id field.
+	if envelope.URL != "" {
```
MuneebUllahKhan222: Nit: I might be wrong here. but from my findings/understanding `&apipb.SourceUnit` that is received by the scan job never has URL so `envelope.URL != ""` branch is  dead code and can be removed.

MuneebUllahKhan222 (approved):
LGTM!

kashifkhan0771 (approved):

pkg/sources/github/github.go
```diff
@@ -2448,15 +2448,12 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
...
-	if ok && s.conn.ScanAllInstallations && kind == "repo" {
+	if ac, ok := s.connector.(*appConnector); ok && s.conn.ScanAllInstallations && kind == "repo" {
 		if ru, ok := unit.(RepoUnit); ok && ru.InstallationID > 0 {
 			ac.setRepoInstallationForRepoName(repoURL, ru.Name, ru.InstallationID)
 		} else if err := s.mapReposToInstallations(ctx, ac, []string{repoURL}); err != nil {
 			return err
 		}
-	} else if err := s.mapReposToInstallations(ctx, ac, []string{repoURL}); err != nil {
-		return err
 	}
```
MuneebUllahKhan222, line 2517: We can re-write the condition like this
```go
  if ac, ok := s.connector.(*appConnector); ok && s.conn.ScanAllInstallations && kind == "repo" {
    ru := unit.(RepoUnit)
    if ru.InstallationID > 0 {
      ac.setRepoInstallationForRepoName(repoURL, ru.Name, ru.InstallationID)
    } else if err := s.mapReposToInstallations(ctx, ac, []string{repoURL}); err != nil {
      return err
    }
```
Since `kind == "repo"`is always true only for `repoUnit` as defined by
```go
func (r RepoUnit) SourceUnitID() (string, sources.SourceUnitKind) { return r.URL, "repo" }
```
we don't need to do `if ru, ok := unit.(RepoUnit); ok` assertion

MuneebUllahKhan222 (approved):
The PR LGTM, only one small nit if we wanna do it.

mariduv (approved):

pkg/sources/github/github.go
shahzadhaider1: Ummm, wouldn't this panic if the unit comes back as a `CommonSourceUnit`? `kind` is `"repo"` either way, so I think the `, ok` is doing real work here.

pkg/sources/github/github.go
```diff
@@ -2428,8 +2444,14 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
...
 	ctx = context.WithValue(ctx, "repo", repoURL)
 
-	if connector, ok := s.connector.(*appConnector); ok && s.conn.ScanAllInstallations && kind == "repo" {
-		if err := s.mapReposToInstallations(ctx, connector, []string{repoURL}); err != nil {
+	// If the unit has an installation ID, use it. Fetching it is slow,
+	// specifically it stalls large multi-org scans on rate limits (INT-790).
+
+	// [CG] This is pretty ugly; if you can clean it up please do (I failed).
+	if ac, ok := s.connector.(*appConnector); ok && s.conn.ScanAllInstallations && kind == "repo" {
+		if ru, ok := unit.(RepoUnit); ok && ru.InstallationID > 0 {
```
shahzadhaider1, line 2512: Hey, might be misreading this so feel free to disregard, but when the `sources.CommonSourceUnitUnmarshaller` embed came back in "Simplify the implementation somewhat", doesn't that make `UnmarshalSourceUnit` only ever return a `CommonSourceUnit`? If so, `unit.(RepoUnit)` in `ChunkUnit` would never hit for a rebuilt unit and we'd always take the `mapReposToInstallations` fallback, so the INT-790 cost stays. I tried a few serialized shapes locally and couldn't get a `RepoUnit` back from any of them, but I'm not sure I'm matching what the scan job really receives. Could you double check?

pkg/sources/github/github.go
MuneebUllahKhan222: Yeah you are right I didn't consider the `CommonSourceUnit`.

pkg/sources/github/github.go
camgunz: You were right! I fixed it 🛠️

shahzadhaider1 (approved):

PR #5219, opened 2026-08-18
🐛 fix(engine): map decoded results to source lines

### LINKED ISSUE #5218: HTML decoder reports transformed line numbers
Please review the [Community Note](https://github.com/trufflesecurity/trufflehog/blob/main/.github/community_note.md)
before submitting.

### TruffleHog Version

3.97.0

### Expected Behavior

Both decoder results should point to line 6, where the synthetic value appears in the source file. TruffleHog could also
suppress the transformed duplicate when the plain decoder finds the same value.

### Actual Behavior

The plain decoder reports line 6. The HTML decoder reports line 3 for the same value.

```json
{"decoder":"PLAIN","line":6,"raw":"synthetic-secret-value-123456"}
{"decoder":"HTML","line":3,"raw":"synthetic-secret-value-123456"}
```

### Steps to Reproduce

Create `input.txt`:

```text
# Date format: <yyyymmdd>




token = "synthetic-secret-value-123456"
```

Create `config.yaml`:

```yaml
detectors:
  - name: Line Mapping Repro
    keywords:
      - token
    regex:
      token: |-
        (?m)^token[ \t]*=[ \t]*"([^"]+)"$
```

Run:

```shell
trufflehog filesystem --no-update --json --concurrency=1 \
  --config=config.yaml --include-detectors=CustomRegex \
  --no-verification input.txt
```

### Root Cause

The HTML heuristic accepts `<yyyymmdd>` as a tag. HTML extraction then collapses blank lines before detection.
`FragmentLineOffset` counts newlines in the transformed chunk and adds that count to the original chunk's starting line,
so the transformed result cannot identify its source line.

The bug affects any decoder that changes line structure without retaining a mapping to the source bytes.

### References

- #1434 tracks the same source-mapping problem for Base64-decoded content.
- #4840 introduced the HTML decoder.

gaborbernat:
Fixes #5218, where HTML decoding removes source lines before detector results reach `FragmentLineOffset`. Findings that survive decoding can then point to a transformed line instead of their source line.

Line lookup now uses the matching occurrence in `OriginalData` when the detected value survives decoding, in three steps of increasing cost:

- Value absent from the source, as with a base64-decoded secret: keep decoded-data offsets.
- Same number of occurrences in both buffers: decoders emit the occurrences they keep in source order, so the nth decoded match is the nth source match. One linear scan.
- Counts disagree, meaning the decoder dropped or merged occurrences: score each candidate source occurrence by how much of the decoded neighbourhood still reads out of the source around it, allowing gaps on the source side for what was removed, and take the best.

That last case has no exact answer. A value the decoder dropped and a value it kept are only distinguishable by their surroundings, so it is a ranking, not a lookup. Measured against generated ground truth it lands on the right source line 99.9% of the time, against 95.0% for a full edit script and 31% for giving up and reporting the decoded line.

Nothing on this path allocates.

---

> [!NOTE]
> **Medium Risk**
> Changes how every finding’s reported line and trufflehog:ignore are computed in the engine, but scope is limited to offset mapping with extensive tests and no new allocations on the hot path.
> 
> **Overview**
> Fixes incorrect **line numbers** (and **ignore-tag** checks) when decoders change newlines or drop markup before secrets are found in `chunk.Data`.
> 
> `FragmentLineOffset` now remaps each match in decoded data onto **`OriginalData`** when the secret bytes still exist in the source: same occurrence count uses ordinal matching; when decoding drops or merges duplicates it picks the best source candidate via **surrounding-text alignment** scoring. If the value never existed in the source (e.g. fully decoded secrets), behavior stays on decoded offsets.
> 
> Adds **property tests** (`rapid`), targeted regression cases (HTML blank lines, duplicate occurrences, UTF-8/invalid bytes, ignore tags in source-only markup), and **benchmarks**. Ignores local `testdata/rapid/` failure recordings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d83fc1bc99d06af961f39328cf27b56a1694345. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -14,10 +14,12 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unicode/utf8"
 
 	"github.com/adrg/strutil"
 	"github.com/adrg/strutil/metrics"
 	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/sergi/go-diff/diffmatchpatch"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
@@ -1410,23 +1412,16 @@ func effectiveSecret(r *detectors.Result) string {
 func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64, bool) {
 	secretBytes := []byte(effectiveSecret(result))
 
-	data, offset := chunk.Data, -1
-	if chunk.OriginalData != nil {
-		occurrence := 0
-		if result.HasChunkOffset() {
-			occurrence = bytes.Count(chunk.Data[:result.ChunkOffset()], secretBytes)
-		}
-		offset = nthIndex(chunk.OriginalData, secretBytes, occurrence)
-		if offset >= 0 {
-			data = chunk.OriginalData
-		}
+	offset := bytes.Index(chunk.Data, secretBytes)
+	if result.HasChunkOffset() {
+		offset = int(result.ChunkOffset())
+	} else if offset == -1 {
+		return 0, false
 	}
-	if offset < 0 {
-		if result.HasChunkOffset() {
-			offset = int(result.ChunkOffset())
-		} else if offset = bytes.Index(chunk.Data, secretBytes); offset == -1 {
-			return 0, false
-		}
+
+	data := chunk.Data
+	if originalOffset := sourceOffset(chunk.OriginalData, chunk.Data, offset, len(secretBytes)); originalOffset >= 0 {
+		data, offset = chunk.OriginalData, originalOffset
 	}
 
 	lineNumber := int64(bytes.Count(data[:offset], []byte("\n")))
@@ -1444,18 +1439,31 @@ func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64, 
 	return lineNumber, false
 }
 
-func nthIndex(data, value []byte, occurrence int) int {
-	offset := 0
-	for matchNumber := 0; matchNumber <= occurrence; matchNumber++ {
-		index := bytes.Index(data[offset:], value)
-		if index == -1 {
-			return -1
+func sourceOffset(originalData, data []byte, offset, length int) int {
+	if originalData == nil || offset < 0 || offset+length > len(data) {
+		return -1
+	}
+	if bytes.Equal(originalData, data) {
+		return offset
+	}
+	if !utf8.Valid(originalData) || !utf8.Valid(data) {
+		return -1
+	}
+
+	var originalOffset, dataOffset int
+	for _, diff := range diffmatchpatch.New().DiffMain(string(originalData), string(data), true) {
+		switch diff.Type {
+		case diffmatchpatch.DiffDelete:
+			originalOffset += len(diff.Text)
+		case diffmatchpatch.DiffInsert:
+			dataOffset += len(diff.Text)
+		case diffmatchpatch.DiffEqual:
+			if offset >= dataOffset && offset+length <= dataOffset+len(diff.Text) {
+				return originalOffset + offset - dataOffset
+			}
+			originalOffset += len(diff.Text)
+			dataOffset += len(diff.Text)
 		}
-		offset += index
-		if matchNumber == occurrence {
-			return offset
-		}
-		offset += len(value)
 	}
 	return -1
 }
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -1410,24 +1410,30 @@ func effectiveSecret(r *detectors.Result) string {
 func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64, bool) {
 	secretBytes := []byte(effectiveSecret(result))
 
-	// Locate the byte offset of the secret in chunk.Data. If a chunk offset was
-	// pre-assigned (for duplicate secrets), use it directly to find the correct
-	// occurrence instead of always matching the first one.
-	var offset int
-	if result.HasChunkOffset() {
-		offset = int(result.ChunkOffset())
-	} else {
-		offset = bytes.Index(chunk.Data, secretBytes)
-		if offset == -1 {
+	data, offset := chunk.Data, -1
+	if chunk.OriginalData != nil {
+		occurrence := 0
+		if result.HasChunkOffset() {
+			occurrence = bytes.Count(chunk.Data[:result.ChunkOffset()], secretBytes)
+		}
+		offset = nthIndex(chunk.OriginalData, secretBytes, occurrence)
+		if offset >= 0 {
+			data = chunk.OriginalData
+		}
+	}
+	if offset < 0 {
+		if result.HasChunkOffset() {
+			offset = int(result.ChunkOffset())
+		} else if offset = bytes.Index(chunk.Data, secretBytes); offset == -1 {
 			return 0, false
 		}
 	}
 
-	lineNumber := int64(bytes.Count(chunk.Data[:offset], []byte("\n")))
+	lineNumber := int64(bytes.Count(data[:offset], []byte("\n")))
 	result.SetPrimarySecretLine(lineNumber)
 
 	// If the line containing the secret has the ignore tag, we should ignore the result.
-	after := chunk.Data[offset+len(secretBytes):]
+	after := data[offset+len(secretBytes):]
 	endLine := bytes.Index(after, []byte("\n"))
 	if endLine == -1 {
 		endLine = len(after)
@@ -1436,6 +1442,22 @@ func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64, 
 		return lineNumber, true
 	}
 	return lineNumber, false
+}
+
+func nthIndex(data, value []byte, occurrence int) int {
+	offset := 0
+	for matchNumber := 0; matchNumber <= occurrence; matchNumber++ {
+		index := bytes.Index(data[offset:], value)
+		if index == -1 {
+			return -1
+		}
+		offset += index
+		if matchNumber == occurrence {
+			return offset
+		}
+		offset += len(value)
+	}
+	return -1
 }
 
 // AssignDuplicateLineOffsets pre-computes byte offsets for results that share the same
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -1448,7 +1448,7 @@ func sourceOffset(originalData, data []byte, offset, length int) int {
 	}
 
 	var originalOffset, dataOffset int
-	for _, diff := range diffmatchpatch.New().DiffMain(bytesAsRunes(originalData), bytesAsRunes(data), true) {
+	for _, diff := range diffmatchpatch.New().DiffMainRunes(bytesAsRunes(originalData), bytesAsRunes(data), true) {
 		diffLength := utf8.RuneCountInString(diff.Text)
 		switch diff.Type {
 		case diffmatchpatch.DiffDelete:
@@ -1466,13 +1466,13 @@ func sourceOffset(originalData, data []byte, offset, length int) int {
 	return -1
 }
 
-// DiffMain replaces invalid UTF-8, so encode each source byte as one valid rune.
-func bytesAsRunes(data []byte) string {
+// Map each source byte to one rune so invalid UTF-8 survives diffing.
+func bytesAsRunes(data []byte) []rune {
 	runes := make([]rune, len(data))
 	for i, value := range data {
 		runes[i] = rune(value)
 	}
-	return string(runes)
+	return runes
 }
 
 // AssignDuplicateLineOffsets pre-computes byte offsets for results that share the same
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -1446,26 +1446,33 @@ func sourceOffset(originalData, data []byte, offset, length int) int {
 	if bytes.Equal(originalData, data) {
 		return offset
 	}
-	if !utf8.Valid(originalData) || !utf8.Valid(data) {
-		return -1
-	}
 
 	var originalOffset, dataOffset int
-	for _, diff := range diffmatchpatch.New().DiffMain(string(originalData), string(data), true) {
+	for _, diff := range diffmatchpatch.New().DiffMain(bytesAsRunes(originalData), bytesAsRunes(data), true) {
+		diffLength := utf8.RuneCountInString(diff.Text)
 		switch diff.Type {
 		case diffmatchpatch.DiffDelete:
-			originalOffset += len(diff.Text)
+			originalOffset += diffLength
 		case diffmatchpatch.DiffInsert:
-			dataOffset += len(diff.Text)
+			dataOffset += diffLength
 		case diffmatchpatch.DiffEqual:
-			if offset >= dataOffset && offset+length <= dataOffset+len(diff.Text) {
+			if offset >= dataOffset && offset+length <= dataOffset+diffLength {
 				return originalOffset + offset - dataOffset
 			}
-			originalOffset += len(diff.Text)
-			dataOffset += len(diff.Text)
+			originalOffset += diffLength
+			dataOffset += diffLength
 		}
 	}
 	return -1
+}
+
+// DiffMain replaces invalid UTF-8, so encode each source byte as one valid rune.
+func bytesAsRunes(data []byte) string {
+	runes := make([]rune, len(data))
+	for i, value := range data {
+		runes[i] = rune(value)
+	}
+	return string(runes)
 }
 
 // AssignDuplicateLineOffsets pre-computes byte offsets for results that share the same
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -284,6 +284,20 @@ func TestFragmentLineOffsetSkipsRemovedOriginalDataOccurrence(t *testing.T) {
 	assert.Equal(t, int64(1), lineOffset)
 }
 
+func TestFragmentLineOffsetUsesInvalidOriginalData(t *testing.T) {
+	secret := []byte("synthetic-secret-value-123456")
+	originalData := append([]byte("heading\n\xff\n"), secret...)
+	chunk := &sources.Chunk{
+		Data:         originalData,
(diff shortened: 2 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -1438,6 +1439,42 @@ func FragmentLineOffset(chunk *sources.Chunk, result *detectors.Result) (int64,
...
+func sourceOffset(originalData, data []byte, offset, length int) int {
+	if originalData == nil || offset < 0 || offset+length > len(data) {
+		return -1
+	}
+	if bytes.Equal(originalData, data) {
+		return offset
+	}
+
+	var originalOffset, dataOffset int
+	for _, diff := range diffmatchpatch.New().DiffMain(bytesAsRunes(originalData), bytesAsRunes(data), true) {
```
rosecodym: I'm not too familiar with this library, but it _looks_ like if you use `DiffMainRunes` instead of `DiffMain` you can elide the stringification at the end of `bytesAsRunes. (I'm not sure what the actual performance impact would be, but all else equal, fewer operations = better, right?)

rosecodym (commented):
Thank you for this! (And for the detailed bug report it addresses!) Your new tests inspire great confidence in the change's correctness, so my primary concern is about whether it might impose any unfortunate performance costs given the additional text comparisons it's added to the scanning engine. (Given that you're fixing a bug, these might be unavoidable, but we'd like to understand them _before_ releasing 🙂.) I'm rounding up some additional review for that, but we won't let this linger too long waiting for it.

As far as non-blocking requests from me in the interim: I've left one inline comment regarding a library call that might not matter but I don't think would hurt. Also, your great new tests exercise cases that aren't immediately obvious, so any time you can spare to add a brief doc comment to each explaining what it's testing would be appreciated.

kashifkhan0771 (approved):
LGTM!

pkg/engine/engine.go
```diff
@@ -1410,24 +1412,23 @@ func effectiveSecret(r *detectors.Result) string {
...
 	if result.HasChunkOffset() {
 		offset = int(result.ChunkOffset())
-	} else {
-		offset = bytes.Index(chunk.Data, secretBytes)
-		if offset == -1 {
-			return 0, false
-		}
+	} else if offset == -1 {
+		return 0, false
+	}
```
mcastorina, line 1420: I don't understand why this change was included. It looks like we *always* calculate the `offset` now?

mcastorina (commented):
Thanks for the fix (and the tests!). We will need to look into the performance aspect of this.

mcastorina (changes requested):
Requesting changes to prevent merging.

@MuneebUllahKhan222 ran some benchmarks and unfortunately this is a large enough regression that we need to weigh correctness vs performance.

```
                                                 │   old.txt   │                   new.txt                    │
                                                 │   sec/op    │     sec/op       vs base                     │
FragmentLineOffset_Identical_Small-10              29.62n ± 0%       31.92n ± 1%        +7.78% (p=0.000 n=10)
FragmentLineOffset_Identical_Large-10              1.408µ ± 0%       1.439µ ± 0%        +2.17% (p=0.000 n=10)
FragmentLineOffset_Diffed_Small-10                 29.54n ± 1%    42727.50n ± 0%   +144542.86% (p=0.000 n=10)
FragmentLineOffset_Diffed_Large-10                 1.417µ ± 0%   23044.929µ ± 0%  +1626218.21% (p=0.000 n=10)
FragmentLineOffset_Diffed_MaxRealisticChunk-10     828.3n ± 0%   9021587.0n ± 0%  +1089069.02% (p=0.000 n=10)
FragmentLineOffset_Diffed_DefaultChunkSize-10      180.0n ± 0%    889702.5n ± 1%   +494179.17% (p=0.000 n=10)
FragmentLineOffset_Identical_DefaultChunkSize-10   180.2n ± 0%       183.8n ± 0%        +2.00% (p=0.000 n=10)
FragmentLineOffsetStart-10                         27.62n ± 1%       26.98n ± 5%             ~ (p=0.089 n=10)
FragmentLineOffsetMiddle-10                        564.4n ± 0%       578.9n ± 0%        +2.57% (p=0.000 n=10)
FragmentLineOffsetEnd-10                           1.074µ ± 0%       1.098µ ± 0%        +2.19% (p=0.000 n=10)
geomean                                            245.2n            8.050µ          +3183.35%

                                                 │    old.txt     │             new.txt              │
                                                 │      B/op      │     B/op      vs base            │
FragmentLineOffset_Identical_Small-10                0.000 ± 0%       0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffset_Identical_Large-10                0.000 ± 0%       0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffset_Diffed_Small-10                   0.0Ki ± 0%     154.8Ki ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Diffed_Large-10                  0.00Mi ± 0%     38.47Mi ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Diffed_MaxRealisticChunk-10      0.00Mi ± 0%     20.80Mi ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Diffed_DefaultChunkSize-10      0.000Mi ± 0%     3.278Mi ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Identical_DefaultChunkSize-10     0.000 ± 0%       0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffsetStart-10                           0.000 ± 0%       0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffsetMiddle-10                          0.000 ± 0%       0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffsetEnd-10                             0.000 ± 0%       0.000 ± 0%  ~ (p=1.000 n=10) ¹
geomean                                                         ²                 ?                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                                                 │    old.txt    │             new.txt             │
                                                 │   allocs/op   │  allocs/op   vs base            │
FragmentLineOffset_Identical_Small-10               0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffset_Identical_Large-10               0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffset_Diffed_Small-10                    0.0 ± 0%      251.0 ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Diffed_Large-10                  0.00k ± 0%     33.16k ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Diffed_MaxRealisticChunk-10      0.00k ± 0%     18.47k ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Diffed_DefaultChunkSize-10      0.000k ± 0%     3.206k ± 0%  ? (p=0.000 n=10)
FragmentLineOffset_Identical_DefaultChunkSize-10    0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffsetStart-10                          0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffsetMiddle-10                         0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=10) ¹
FragmentLineOffsetEnd-10                            0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=10) ¹
geomean                                                        ²                ?                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

gaborbernat:
> ran some benchmarks and unfortunately this is a large enough regression that we need to weigh correctness vs performance.

Dunno, correctness IMHO is more important in general. Giving the wrong answer but fast just causes user confusion. If performance is a worry we should look to find way to mitigate it, not per se accept incorrect behaviour.

gaborbernat:
Fair point on the benchmarks, and the diff turned out to be the wrong tool. Two problems with it beyond being quadratic:

- It ran on every decoded chunk, including the common case where the detected value has no source counterpart at all (a base64-decoded secret, say). That work was always discarded.
- At 64KB it hit go-diff's one-second default `DiffTimeout`, so a single result could stall a scanner worker for a full second.

Replaced it with an occurrence-count mapping. Decoders emit the occurrences they keep in source order, so when both buffers hold the same number of occurrences of the value, the nth decoded match is the nth source match. That is a linear, allocation-free scan. When the counts disagree the decoder dropped or merged occurrences and nothing cheap can say which one survived, so those results keep decoded-data line numbers, the behavior they have on `main`. `go-diff` is no longer used in this path at all.

I pushed the benchmarks as well (ba07db9) so this stays measurable: identical and decoded chunks at small, default, max-realistic and 64KB sizes, plus the count-mismatch fallback. M4 Pro, `main` vs the commit you benchmarked vs now:

```
                                                 │    main     │                      reviewed                      │                  now                   │
                                                 │   sec/op    │       sec/op        vs base                        │    sec/op      vs base                 │
FragmentLineOffsetStart-14                         27.45n ± 0%          28.02n ± 1%          +2.11% (p=0.001 n=8+6)     28.30n ± 1%     +3.12% (p=0.000 n=8)
FragmentLineOffsetMiddle-14                        606.6n ± 1%          603.5n ± 1%          -0.50% (p=0.010 n=8+6)     638.0n ± 6%     +5.18% (p=0.000 n=8)
FragmentLineOffsetEnd-14                           1.148µ ± 2%          1.149µ ± 0%               ~ (p=0.726 n=8+6)     1.172µ ± 0%     +2.05% (p=0.000 n=8)
FragmentLineOffset_Identical_Small-14              39.88n ± 1%          43.56n ± 1%          +9.23% (p=0.001 n=8+6)     43.45n ± 2%     +8.95% (p=0.000 n=8)
FragmentLineOffset_Identical_DefaultChunkSize-14   2.864µ ± 1%          2.868µ ± 1%               ~ (p=0.922 n=8+6)     2.859µ ± 2%          ~ (p=0.977 n=8)
FragmentLineOffset_Identical_Large-14              17.91µ ± 1%          17.91µ ± 1%               ~ (p=0.950 n=8+6)     17.86µ ± 1%          ~ (p=0.593 n=8)
FragmentLineOffset_Diffed_Small-14                 18.90n ± 1%        9350.50n ± 1%      +49373.54% (p=0.001 n=8+6)     58.11n ± 3%   +207.49% (p=0.000 n=8)
FragmentLineOffset_Diffed_DefaultChunkSize-14      79.16n ± 1%    46949563.00n ± 2%   +59309605.66% (p=0.001 n=8+6)   2916.00n ± 1%  +3583.68% (p=0.000 n=8)
FragmentLineOffset_Diffed_MaxRealisticChunk-14     103.2n ± 1%     78740930.5n ± 0%   +76299251.26% (p=0.001 n=8+6)    3777.0n ± 1%  +3559.88% (p=0.000 n=8)
FragmentLineOffset_Diffed_Large-14                 557.1n ± 0%   1001429771.0n ± 0%  +179741400.67% (p=0.001 n=8+6)   18375.5n ± 1%  +3198.12% (p=0.000 n=8)
FragmentLineOffset_Unmapped_DefaultChunkSize-14                                                                         2.790µ ± 1%
geomean                                            283.6n               32.94µ           +11516.36%                     1.040µ        +232.28%
```

```
                                                 │  allocs/op   │  allocs/op   vs base             │ allocs/op   vs base                │
FragmentLineOffset_Identical_Small-14              0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=8+6) ¹   0.000 ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Identical_DefaultChunkSize-14   0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=8+6) ¹   0.000 ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Identical_Large-14              0.000 ± 0%      0.000 ± 0%  ~ (p=1.000 n=8+6) ¹   0.000 ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Diffed_Small-14                   0.0 ± 0%      168.0 ± 0%  ? (p=0.001 n=8+6)       0.0 ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Diffed_DefaultChunkSize-14      0.00k ± 0%     28.39k ± 0%  ? (p=0.001 n=8+6)     0.00k ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Diffed_MaxRealisticChunk-14     0.00k ± 0%     36.98k ± 0%  ? (p=0.001 n=8+6)     0.00k ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Diffed_Large-14                   0.0 ± 0%      175.0 ± 0%  ? (p=0.001 n=8+6)       0.0 ± 0%       ~ (p=1.000 n=8) ¹
FragmentLineOffset_Unmapped_DefaultChunkSize-14                                                      0.000 ± 0%
geomean                                                       ²                ?                 ²               +0.00%               ²
```

What is left over `main` is one extra pass across `OriginalData` to locate the value and count the newlines before it. That is what mapping a result onto its source line costs; `main` is cheaper only because it does not do the mapping. Worst case is 18µs on a 64KB chunk, with zero allocations everywhere.

@mcastorina the always-computed `bytes.Index` is back to the original branch, that was an accident rather than intentional. @rosecodym the `DiffMainRunes` suggestion is moot now the library call is gone, and the test doc comments landed in de127e1.

pkg/engine/engine.go
gaborbernat: Accident on my part, not intentional. Reverted to the original branch so the pre-assigned offset path no longer scans for an offset it already has.

pkg/engine/engine.go
gaborbernat: Switched to `DiffMainRunes` at the time, but the whole diff is gone now that mapping is done by occurrence counting, so this no longer applies.

gaborbernat:
Correcting my last comment: I had the ambiguous case fall back to decoded-data line numbers, and said that was no worse than `main`. I then wrote a property test that generates a decoded buffer from a source buffer by dropping spans, so ground truth is known by construction, and pointed it at each strategy. Falling back gets the source line right **31%** of the time, so it was very much worse. That is fixed in a42d783.

The edit script does not look good on that test either. Over ~11k generated ambiguous cases:

| strategy | correct source line | cost, 10KB chunk |
| --- | --- | --- |
| full edit script (what you reviewed) | 95.0% | 47ms |
| local alignment (now) | 99.9% | 6µs |
| fall back to decoded line | 31% | 3µs |

The replacement scores each candidate source occurrence by how much of the decoded neighbourhood still reads out of the source around it, allowing gaps on the source side for whatever the decoder removed, and takes the best. That is the same rule a minimal edit script applies, evaluated locally per candidate over a 1KB window rather than globally, which is both cheaper and, on this data, more accurate. It runs only when the occurrence counts disagree; the common cases are settled by two linear scans before that.

I want to be straight about what it is: a ranking, not a lookup. When a decoder drops one copy of a value and keeps another, the copies are distinguishable only by their surroundings, so exactness is not available at any price. `TestFragmentLineOffsetReportsLineHoldingSecret` pins the invariant that does hold, that the reported line contains the value in the source, and `TestFragmentLineOffsetMapsDroppedSourceSpans` pins exactness for the decidable case.

Updated numbers, `main` vs the commit you benchmarked vs now:

```
                                                 │    main     │                      reviewed                      │                  now                   │
                                                 │   sec/op    │       sec/op        vs base                        │    sec/op      vs base                 │
FragmentLineOffsetStart-14                         27.45n ± 0%          28.02n ± 1%          +2.11% (p=0.001 n=8+6)     28.28n ± 2%     +3.06% (p=0.000 n=8)
FragmentLineOffsetMiddle-14                        606.6n ± 1%          603.5n ± 1%          -0.50% (p=0.010 n=8+6)     607.4n ± 2%          ~ (p=0.820 n=8)
FragmentLineOffsetEnd-14                           1.148µ ± 2%          1.149µ ± 0%               ~ (p=0.726 n=8+6)     1.143µ ± 2%          ~ (p=0.490 n=8)
FragmentLineOffset_Identical_Small-14              39.88n ± 1%          43.56n ± 1%          +9.23% (p=0.001 n=8+6)     43.65n ± 1%     +9.45% (p=0.000 n=8)
FragmentLineOffset_Identical_DefaultChunkSize-14   2.864µ ± 1%          2.868µ ± 1%               ~ (p=0.922 n=8+6)     2.808µ ± 2%     -1.96% (p=0.000 n=8)
FragmentLineOffset_Identical_Large-14              17.91µ ± 1%          17.91µ ± 1%               ~ (p=0.950 n=8+6)     17.74µ ± 1%     -0.94% (p=0.021 n=8)
FragmentLineOffset_Diffed_Small-14                 18.90n ± 1%        9350.50n ± 1%      +49373.54% (p=0.001 n=8+6)     57.26n ± 1%   +202.96% (p=0.000 n=8)
FragmentLineOffset_Diffed_DefaultChunkSize-14      79.16n ± 1%    46949563.00n ± 2%   +59309605.66% (p=0.001 n=8+6)   2884.00n ± 2%  +3543.25% (p=0.000 n=8)
FragmentLineOffset_Diffed_MaxRealisticChunk-14     103.2n ± 1%     78740930.5n ± 0%   +76299251.26% (p=0.001 n=8+6)    3752.0n ± 2%  +3535.66% (p=0.000 n=8)
FragmentLineOffset_Diffed_Large-14                 557.1n ± 0%   1001429771.0n ± 0%  +179741400.67% (p=0.001 n=8+6)   18278.5n ± 2%  +3180.71% (p=0.000 n=8)
FragmentLineOffset_Ambiguous_DefaultChunkSize-14                                                                         5.965µ ± 1%
geomean                                            283.6n               32.94µ           +11516.36%                     1.101µ        +227.92%
```

Zero allocations on every case, against 37K allocs/op at 13KB before. The remaining gap over `main` is one extra pass across `OriginalData` to locate the value and count the newlines before it, which is what mapping onto a source line costs.

Happy to share the ground-truth harness if you want to run it against other strategies, and happy to revisit the 1KB window if you would rather trade accuracy down for the last microseconds.

rosecodym (approved):
Thank you for digging into this! And thank you particularly for the changes you made after our performance analysis.

camgunz (approved):

mcastorina (approved):
Thanks for taking the performance into consideration!

PR #5223, opened 2026-08-20
detectors/github/v2: differentiate token type in ExtraData

hartsock:
## Problem
`keyPat` in `pkg/detectors/github/v2/github.go` matches all six GitHub token prefixes with one regex: `ghp_`/`github_pat_` (PATs), `gho_` (OAuth), `ghu_`/`ghs_`/`ghr_` (GitHub App tokens). Every match gets the same `DetectorType_Github` and the same hardcoded PAT description. The prefix is captured in `token := match[1]` but discarded before it reaches `Result`.

Nothing downstream can tell a leaked OAuth grant or GitHub App token from a leaked PAT. That matters: each token type revokes through a different GitHub settings page. A user following PAT-revocation steps on a `ghu_`/`ghs_` finding revokes nothing.

## Fix
Add `token_type` and `remediation` to `ExtraData`, keyed off the matched prefix. Follows the existing `ExtraData["token_type"]` convention (`launchdarkly`, `slack`, `larksuite`).

## Why this matters
Hit this at NVIDIA: three findings kept reopening across repeated full PAT revocations because the leaked credential was OAuth/GitHub-App, not a PAT, and the finding never said so. Cost incident-response time and triggered a false "revocation is broken" alarm before we traced it here.

## Testing
`go build`, `go vet`, `gofmt -l`, `TestGithub_Pattern` all pass. Additive `ExtraData` fields only; no detection/verification behavior change.

---

> [!NOTE]
> **Low Risk**
> Additive ExtraData only; no change to matching, verification, or secret handling.
> 
> **Overview**
> GitHub v2 findings now include `ExtraData["token_type"]` derived from the matched prefix (`ghp_`, `github_pat_`, `gho_`, `ghu_`, `ghs_`, `ghr_`), so responders can tell a PAT from an OAuth grant or GitHub App token.
> 
> Detection and verification are unchanged. Tests cover each prefix, plus drift guards so `keyPat` and `tokenTypesByPrefix` cannot diverge without failing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4072f231ee06aa34258cf1ff0ab7ab0bf64a858c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/github/v2/github.go
+++ b/pkg/detectors/github/v2/github.go
@@ -39,31 +39,31 @@ var (
 	// TODO: Oauth2 client_id and client_secret
 	// https://developer.github.com/v3/#oauth2-keysecret
 
-	// tokenTypesByPrefix maps each GitHub token prefix to a human-readable type
-	// and its distinct remediation path. Each prefix identifies a materially
-	// different credential (PAT, OAuth grant, or GitHub App token) that is
-	// revoked/rotated through a different GitHub settings page.
+	// tokenTypesByPrefix maps each GitHub token prefix to a human-readable type.
+	// Each prefix identifies a materially different credential (PAT, OAuth
+	// grant, or GitHub App token) that is revoked/rotated through a different
+	// GitHub settings page.
 	// https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
-	tokenTypesByPrefix = map[string]struct{ TokenType, Remediation string }{
-		"ghp_":        {"Personal Access Token (classic)", "Revoke at https://github.com/settings/tokens"},
-		"github_pat_": {"Personal Access Token (fine-grained)", "Revoke at https://github.com/settings/personal-access-tokens"},
-		"gho_":        {"OAuth Access Token", "Revoke at https://github.com/settings/applications (Authorized OAuth Apps)"},
-		"ghu_":        {"GitHub App User-to-Server Token", "Revoke at https://github.com/settings/apps/authorizations (Authorized GitHub Apps) or uninstall at https://github.com/settings/installations"},
-		"ghs_":        {"GitHub App Server-to-Server (installation) Token", "Uninstall the GitHub App at https://github.com/settings/installations"},
-		"ghr_":        {"GitHub App Refresh Token", "Uninstall the GitHub App at https://github.com/settings/installations"},
+	tokenTypesByPrefix = map[string]string{
+		"ghp_":        "Personal Access Token (classic)",
+		"github_pat_": "Personal Access Token (fine-grained)",
+		"gho_":        "OAuth Access Token",
+		"ghu_":        "GitHub App User-to-Server Token",
+		"ghs_":        "GitHub App Server-to-Server (installation) Token",
+		"ghr_":        "GitHub App Refresh Token",
 	}
 )
 
-// githubTokenType returns the human-readable token type and remediation path
-// for a matched token, keyed off its prefix. Falls back to a generic label if
-// no known prefix matches (should not happen given keyPat, but keeps this safe).
-func githubTokenType(token string) (tokenType, remediation string) {
-	for prefix, info := range tokenTypesByPrefix {
+// githubTokenType returns the human-readable token type for a matched token,
+// keyed off its prefix. Falls back to a generic label if no known prefix
+// matches (should not happen given keyPat, but keeps this safe).
+func githubTokenType(token string) string {
+	for prefix, tokenType := range tokenTypesByPrefix {
 		if strings.HasPrefix(token, prefix) {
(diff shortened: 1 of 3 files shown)


hartsock:
@dylanTruffle, @Miccah,
this fixes a real remediation gap: keyPat matches all six GitHub token prefixes but tags every one as DetectorType_Github with a hardcoded  PAT description. Downstream consumers can't tell a leaked OAuth/GitHub App token from a PAT, so a user following PAT-revocation steps on a ghu_/ghs_ finding revokes nothing. I spent nine months tracing this.

pkg/detectors/github/v2/github.go
```diff
@@ -37,8 +38,34 @@ var (
...
+	// different credential (PAT, OAuth grant, or GitHub App token) that is
+	// revoked/rotated through a different GitHub settings page.
+	// https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
+	tokenTypesByPrefix = map[string]struct{ TokenType, Remediation string }{
+		"ghp_":        {"Personal Access Token (classic)", "Revoke at https://github.com/settings/tokens"},
+		"github_pat_": {"Personal Access Token (fine-grained)", "Revoke at https://github.com/settings/personal-access-tokens"},
+		"gho_":        {"OAuth Access Token", "Revoke at https://github.com/settings/applications (Authorized OAuth Apps)"},
+		"ghu_":        {"GitHub App User-to-Server Token", "Revoke at https://github.com/settings/apps/authorizations (Authorized GitHub Apps) or uninstall at https://github.com/settings/installations"},
+		"ghs_":        {"GitHub App Server-to-Server (installation) Token", "Uninstall the GitHub App at https://github.com/settings/installations"},
+		"ghr_":        {"GitHub App Refresh Token", "Uninstall the GitHub App at https://github.com/settings/installations"},
```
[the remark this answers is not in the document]
unsmith: This seems like a legit finding, perhaps the remediation string was copy/pasted and missed on update?  My take from the dogs is that `ghr_` should be revoked at `settings/apps/authorizations`.

pkg/detectors/github/v2/github.go
```diff
@@ -37,8 +38,34 @@ var (
 
 	// TODO: Oauth2 client_id and client_secret
 	// https://developer.github.com/v3/#oauth2-keysecret
+
+	// tokenTypesByPrefix maps each GitHub token prefix to a human-readable type
+	// and its distinct remediation path. Each prefix identifies a materially
+	// different credential (PAT, OAuth grant, or GitHub App token) that is
+	// revoked/rotated through a different GitHub settings page.
+	// https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
+	tokenTypesByPrefix = map[string]struct{ TokenType, Remediation string }{
```
unsmith: I like the granularity here, but I think we might want this folded into the `howtorotate.com` docs instead (or at least, in addition).  Adding another field to the `ExtraData` is additive but we would need to consider downstream consumers of this and what they would do with the new field.

EDIT: We're going to look into expanding the documentation on `howtorotate.com` to handle the differing token types, so I'll ask that the `remediation` field be removed from `ExtraData` here.

unsmith (commented):
This looks ok, differentiating between GitHub token types by way of `ExtraData`.  Can you add some tests for the new behavior(s) as well, something to test that token strings map to the expected `tokenType` and to catch drift if the pattern is updated but the mapping is not?

unsmith (changes requested):
As mentioned, please remove the `remediation` field from `ExtraData` and we'll handle documentation on `howtorotate.com`.

pkg/detectors/github/v2/github.go
hartsock: I think the main _pain point_ for me was just that this kept getting flagged as a PAT and displaying the PAT remediation. If we linked to something about the different types of tokens & categorized them with _possible_ remediations. That would have been less painful and would have resolved the 9 month incident much faster.

It simply never occurred to me that the TH report was about an OAuth token b/c the TH report kept insisting it was a PAT.

hartsock:
Thanks for the review @unsmith ... now that you've mentioned it, I _completely_ understand the desire to keep the metadata surfaces clean as possible. I've updated the PR.

hartsock:
@unsmith I've added some additional integration tests that will check that the data-structure remains well formed if a future developer _does_ decide to modify the metadata format. The additional `reflect` based tests should help guard from a malformed metadata in the future.

unsmith (approved):
Thanks for following up - everything looks good to me now.  This'll be a good addition to the Github detector to help differentiate the various token types.

PR #5229, opened 2026-08-24
Implement Enumerate for the Jenkins source

Ahsan-Sarbaz:
### Description:

Implements `Enumerate` for the Jenkins source, reporting one source unit per job with kind `job`. The unit ID is the job's path relative to the Jenkins instance — Jenkins reports absolute job URLs whose host comes from its own root URL setting, and `Chunks` already discards that host and keeps only the path, so the path is what the source relies on and is stable across runs.

The recursive folder walk is extracted into `walkJobs`, which streams jobs to a callback and routes errors to a second one. `Enumerate` passes `reporter.UnitErr`, so an unreadable folder is reported and the walk continues. `RecursivelyGetJenkinsObjectsForPath` passes a fail-fast handler, keeping the behavior `Chunks` has always had.

Also adds `UnmarshalSourceUnit`. `ChunkUnit` is intentionally not implemented yet, so `SourceManager` still takes the non-unit path and scan behavior is unchanged.

Tests cover enumeration over a nested folder tree, rerun stability, continuing past an unreadable folder, and parity with the jobs `Chunks` walks. Folder recursion previously had no coverage outside the `localInfra` integration test.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Refactors shared job traversal used by `Chunks` while adding enumeration; behavior is heavily tested for parity, but any walk regression could change which Jenkins jobs get scanned.
> 
> **Overview**
> Adds **source-unit enumeration** for Jenkins so each scannable job can be discovered as a unit with kind `job` and a stable ID based on the job URL **path** (not host).
> 
> The recursive folder walk is refactored into shared **`walkJobs`**, which streams jobs through a **`UnitReporter`**. **`Enumerate`** reports one unit per `FreeStyleProject` / `WorkflowJob` and **keeps walking** after reporting unreadable folders. **`GetJenkinsJobs`** / **`Chunks`** reuse the same walk via a **`jobCollector`** that **still fails fast** on folder listing errors.
> 
> Also adds **`JenkinsJob`** (ID, display name) and **`UnmarshalSourceUnit`** for persisted envelopes or bare JSON. **`ChunkUnit` is not implemented**, so **`SourceManager` continues to scan via `Chunks`** until unit chunking lands.
> 
> New tests cover nested folders, enumerate vs chunk parity, error handling differences, and unmarshalling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10e55d3add43d407e3e5d9d4a94435a27aa6f8fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/jenkins/unit.go
+++ b/pkg/sources/jenkins/unit.go
@@ -1,0 +1,63 @@
+package jenkins
+
+import (
+	"strings"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/sources"
+)
+
+const SourceUnitKindJob sources.SourceUnitKind = "job"
+
+// jobPathSegment is the path segment Jenkins puts before every job or folder
+// name in a job URL, e.g. /job/folder1/job/build.
+const jobPathSegment = "job"
+
+// JobUnit is a single Jenkins job, identified by its path relative to the
+// Jenkins instance. Jenkins reports absolute job URLs whose host comes from the
+// instance's own root URL setting, which can differ from the endpoint being
+// scanned, so only the path is retained. Chunks does the same when it scans a
+// job, so the path is the identity the source already relies on.
+type JobUnit struct {
+	Path string `json:"path"`
+}
+
+var _ sources.SourceUnit = JobUnit{}
+
+func (j JobUnit) SourceUnitID() (string, sources.SourceUnitKind) {
+	return j.Path, SourceUnitKindJob
+}
+
+// Display returns the job's folder path without the "job" segments Jenkins
+// interleaves, e.g. /jenkins/job/folder1/job/build becomes folder1/build. It is
+// derived from Path rather than stored so that a unit rebuilt from its ID alone
+// still renders a name.
+func (j JobUnit) Display() string {
+	segments := strings.Split(strings.Trim(j.Path, "/"), "/")
+
+	// Walk in pairs so that a job named "job" and any instance base path
+	// (/jenkins/) are both handled: a name is only ever the segment directly
+	// after a "job" segment.
+	names := make([]string, 0, len(segments)/2)
+	for i := 0; i < len(segments)-1; i++ {
+		if segments[i] == jobPathSegment {
+			names = append(names, segments[i+1])
+			i++
+		}
+	}
+
+	if len(names) == 0 {
+		return j.Path
+	}
+	return strings.Join(names, "/")
+}
+
+// unitEnvelope is the JSON shape a host application persists for an
+// enumerated unit between an enumerate pass and a later scan pass: the
+// SourceUnit proto marshalled with encoding/json, where unit_data carries the
+// original unit as base64-encoded bytes.
+type unitEnvelope struct {
+	ID       string `json:"id"`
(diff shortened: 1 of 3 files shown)


pkg/sources/jenkins/unit.go
```diff
@@ -0,0 +1,63 @@
...
+// still renders a name.
+func (j JobUnit) Display() string {
+	segments := strings.Split(strings.Trim(j.Path, "/"), "/")
+
+	// Walk in pairs so that a job named "job" and any instance base path
+	// (/jenkins/) are both handled: a name is only ever the segment directly
+	// after a "job" segment.
+	names := make([]string, 0, len(segments)/2)
+	for i := 0; i < len(segments)-1; i++ {
+		if segments[i] == jobPathSegment {
```
MuneebUllahKhan222, line 48: In-case a single folder can contain multiple jobs this will return same string.
But I am not sure if a scenario like this can happen.

MuneebUllahKhan222 (changes requested):
This PR is mostly fine just one change requested.

pkg/sources/jenkins/unit.go
Ahsan-Sarbaz: A folder holding multiple jobs is the normal case, but the display strings still differ. The loop appends the name after every job segment, including the leaf, so /job/folder1/job/a becomes folder1/a and /job/folder1/job/b becomes folder1/b. Two jobs only collide if their paths are identical, which means they are the same job.

Display is also just the human readable label. Identity is SourceUnitID, which returns the raw path.

I added a sibling case to the display table plus TestJobUnitDisplaySiblingsDiffer, which fails on any collision across same folder siblings, the same leaf name in different folders, and the same leaf at different depths.

MuneebUllahKhan222: Is something like `job/folder1/job/a` and  `job/folder1/job/b` not a possibility.

Ahsan-Sarbaz: It is possible and is the norm

MuneebUllahKhan222 (approved):

camgunz (commented):
I haven't worked all the way through this, so some details of what I'm about to say might be wrong, so bear with me :). Also this is really good, especially for only having been here a very short time 🏆. It's an elegant way to straddle both the Scan Details and non-Scan Details flows, and I love the test coverage.

Okay, we're implementing Scan Details in the source, and conceptually that means we should start dealing with `SourceUnit`s. I think the natural choice here is `JenkinsJob`, with `JenkinsBuild` as its subunit, that is:
- `JenkinsJob` should implement `SourceUnit`
- `walkJobs` should take a `UnitReporter`
- `GetJenkinsJobs` should build/use its own `UnitReporter` to collect the jobs
- `UnmarshalSourceUnit` should deal in `JenkinsJob`

I'm not requesting changes because I haven't really thought through if it's all a good idea, so (please) feel free to tell me it's not. But the more we can standardize on existing stuff the better, I think.

camgunz (approved):
Looks great!

PR #5236, opened 2026-08-25
fix(s3): cut a child context per object so key/size log values don't pile up

mariduv:
### Description:
Line 545 was reusing the same context from the outer scope in each iteration, reassigning it back with key/size added each time, making the log lines grow over the course of a page.  This makes a context inside each loop - I could've made this a 1 character change by adding `:` to shadow the outer ctx but it seemed better to me to give it a distinct name.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Logging/context scoping fix only; behavior is unchanged aside from correct per-object log fields and isolation from accidental context mutation across iterations.
> 
> **Overview**
> In **`pageChunker`**, each object in a list page now gets its own context **`octx`** derived from the page-level **`ctx`** (bucket + page number), instead of reassigning **`ctx`** on every loop iteration.
> 
> Previously, **`context.WithValues`** was applied to the same variable each time, which chained structured log fields so later objects inherited every prior object's **`key`** and **`size`** on log lines. Per-object work—cancellation checks, logging, checkpoint updates, **`GetObject`** timeouts, **`HandleFile`**, and the job pool closure—now consistently uses **`octx`**, so logs and context values reflect only the current object.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eab9f6949b20c01f045828af23367b48d4aed07d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/sources/s3/s3.go
+++ b/pkg/sources/s3/s3.go
@@ -542,59 +542,59 @@ func (s *Source) pageChunker(
 	checkpointer.Reset() // Reset the checkpointer for each PAGE
 	ctx = context.WithValues(ctx, "bucket", metadata.bucket, "page_number", metadata.pageNumber)
 	for objIdx, obj := range metadata.page.Contents {
-		ctx = context.WithValues(ctx, "key", *obj.Key, "size", *obj.Size)
-		if common.IsDone(ctx) {
+		octx := context.WithValues(ctx, "key", *obj.Key, "size", *obj.Size)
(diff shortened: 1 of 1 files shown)


kashifkhan0771 (approved):

unsmith (approved):
Heh, interesting bug.  Good fix!

PR #5253, opened 2026-08-31
Set all verification errors in detectors

shahzadhaider1:
Some detectors threw away a finding when verification failed for a reason that wasn't "bad credential"; a body read error, a failed request build, a transport error. Nothing got reported, so a secret that had since been revoked never had its status updated and stayed looking live.

This fixes 45 of them. Each moves its verification into a `verifyX(...) (bool, error)` and appends the result unconditionally, recording the failure via `SetVerificationError`. Status-code logic is copied over verbatim; no verified/unverified outcome changes.

One commit per detector, each with its own feature flag.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Wide mechanical refactor across many detectors that perform live credential checks; behavior change is mainly surfacing more findings as unknown on infra errors, with per-detector flags to mitigate rollout risk.
> 
> **Overview**
> Refactors **~45 secret detectors** so verification no longer silently drops matches when HTTP verification fails for operational reasons (request build, transport, body read, parse errors). Each detector now uses a **`verifyMatch` (or equivalent) `(bool, error)` helper**, always appends the finding, sets **`Verified`**, and records failures with **`SetVerificationError`** instead of `continue`-ing past the result.
> 
> Verification **success/failure rules are unchanged** (same endpoints, status codes, and body heuristics); only error propagation and result retention differ. This aligns findings with **`unknown`** / verification-error output when APIs are unreachable or responses are malformed.
> 
> **Rollout:** adds per-detector **`atomic.Bool` feature flags** (INT-595), wires them in **`main.go`**, **`pkg/feature`**, and **`pkg/engine/defaults`** so each detector can be gated during the verification-error reporting rollout; defaults tests list the gated detector types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cba3a0def2048de22e1ae26aef88c16f4902fd88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MuneebUllahKhan222 (approved):
This LGTM. My only concern with this PR is the addition of a feature flag gate for each of these 45 detectors. Since they were already enabled, it suggests they aren't generating a significant number of false positives, so I think it should be fine to leave them ungated.

shahzadhaider1:
> This LGTM. My only concern with this PR is the addition of a feature flag gate for each of these 45 detectors. Since they were already enabled, it suggests they aren't generating a significant number of false positives, so I think it should be fine to leave them ungated.


I gated them because this changes verification behaviour in 45 detectors at once. If one of them turns out to be noisy in prod; say it starts emitting indeterminate results for a whole customer's scan, we can switch off that single detector instead of reverting the whole PR. That said, I get your point: if one more person agrees with you, I'll drop them.

mariduv (approved):
Thrilled just to see this turns things into better patterns for testing in the future too :joy:

PR #5255, opened 2026-08-31
[INT-943] Add HumioIngestToken detector

unsmith:
### Developer Commentary

This implements a detector and verifier for Humio Ingest Tokens.  These are _write-only_ tokens that allow a user to send log data into the LogScale system.  As they are write-only, it presents some challenges for verification.  We attempt a low-impact query here, a zero-length HEC event to see if the token is at least accepted.  This likely creates a log event in the customer's system, but there's no other way.

As these are just UUIDs, we lean on keyword matching as well to look for "things that reasonably look like Humio ingest tokens."

A small wrinkle - we don't currently (and might not ever) have access to a real, live Humio instance to try creating our own keys.  This detector/verifier is best-effort based on research and testing with expired/invalidated tokens.

### Description:

Add a new detector for Humio/CrowdStrike Falcon LogScale repository ingest tokens ([INT-943](https://linear.app/truffle-security/issue/INT-943/implement-humioingesttoken-detector), child of [INT-829](https://linear.app/truffle-security/issue/INT-829/humio-api-keys-detector-csm-1715)).

Ingest tokens are UUID-format credentials scoped to a single LogScale repository that authorize sending log data. A leaked ingest token allows an attacker to inject arbitrary log data into the customer's repository.

**Detection:** UUID regex anchored via `PrefixRegex(["humio", "logscale"])` to avoid false positives on generic UUIDs. Shannon entropy filtering rejects placeholder UUIDs. Aho-Corasick pre-filter keywords are `humio` and `logscale`.

**Verification:** Sends a minimal empty-event POST to the HEC endpoint (`/api/v1/ingest/hec`). Ingest tokens are write-only — LogScale provides no read or validation endpoint that accepts them — so this is the only verification path available. Each verification creates one log entry in the customer's repo.

**Response code handling:**
- 200 → verified
- 403 → verified (token recognized but blocked by IP filter or permission change)
- 422 → verified (token recognized but repo deleted or unresolvable)
- 401 → not verified (unknown token)
- Other → indeterminate verification error

**Endpoint support:** Implements `EndpointCustomizer` + `CloudProvider` so the detector works against the cloud default (`cloud.us.humio.com`) and self-hosted instances whose URLs appear in scanned data.

**New files:**
- `pkg/detectors/humioingesttoken/humioingesttoken.go`
- `pkg/detectors/humioingesttoken/humioingesttoken_test.go`
- `pkg/detectors/humioingesttoken/humioingesttoken_integration_test.go`

**Wiring:**
- Proto enum `HumioIngestToken = 1070` (regenerated via `make protos`)
- Feature flag `HumioIngestTokenDetectorEnabled` in `feature.go`, gated in `defaults.go`, enabled in `main.go`

### Checklist:
- [x] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?


pkg/detectors/humioingesttoken/humioingesttoken.go
```diff
@@ -0,0 +1,159 @@
...
+// data"). This means every verification creates a log entry in the customer's
+// repository. The event body is an empty string to minimize noise, but scans
+// with many matches will produce that many entries.
+//
+// Response semantics: 200 means fully valid. 403 or 422 still signal a
+// recognized token (blocked by IP filter / permissions, or pointing at a
+// deleted repo) so both count as verified. Only 401 means the token is
+// completely unknown.
+func verifyIngestToken(ctx context.Context, client *http.Client, baseURL, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, baseURL+"/api/v1/ingest/hec", bytes.NewBufferString(`{"event":""}`))
```
MuneebUllahKhan222: According to my robot, this would cause a mutation to the state of the CrowdStrike instance, which we never want to do. For these kinds of cases, we usually send malformed JSON instead.

pkg/detectors/humioingesttoken/humioingesttoken_test.go
```diff
@@ -0,0 +1,206 @@
...
+func TestHumioIngestToken_Pattern(t *testing.T) {
+	d := Scanner{}
+	ahoCorasickCore := ahocorasick.NewAhoCorasickCore([]detectors.Detector{d})
+	tests := []struct {
+		name  string
+		input string
+		want  []string
+	}{
+		{
+			name:  "valid pattern - humio keyword",
```
MuneebUllahKhan222: Can you take some inspiration from `solarwindsobservability_test.go` and reshape the inputs for these test cases to follow the same input pattern used in that file?

MuneebUllahKhan222 (changes requested):
I have requested a couple of small changes. Other than that the PR seems fine.

pkg/detectors/humioingesttoken/humioingesttoken.go
camgunz: When building the URL, let's use [`url.URL`](https://pkg.go.dev/net/url#URL) and [`url.JoinPath`](https://pkg.go.dev/net/url#JoinPath); we've had a bunch of bugs shake out of using string concatenation (etc) to build URLs.

camgunz (changes requested):
Just had 1 thing about URL construction

pkg/detectors/humioingesttoken/humioingesttoken.go
unsmith: Fair point.  Ingestion tokens are write-only, and the only endpoint to verify against that I've been able to find is that HEC endpoint.  According to their docs, the HEC endpoint can operate on structured or unstructured data and my take is that they just accept a blob and attempt to parse it.  It's entirely possible that malformed JSON would be interpreted as a raw, unstructured event and stored that way anyway, defeating the attempt to have a bad payload not create a log entry on their side.  I guess we need to decide as a team how to proceed here.

pkg/detectors/humioingesttoken/humioingesttoken_test.go
unsmith: Can do, yes.

pkg/detectors/humioingesttoken/humioingesttoken.go
unsmith: Will do.

PR #5257, opened 2026-08-31
Low-memory Git Scanning

mariduv:
### Description:
This PR adds an alternate mode to git scanning which significantly reduces the peak memory usage for a given repo, for a small trade in speed.  Currently a single `git log` command is run on a repo, listing all history and patches, and this git process's peak RSS always increases with repo size, so very large repos can reach high peak memory use that is hard to plan for in containers and quotas, especially with concurrency and with other repos of different sizes.

Instead, with this mode, we'd still lean on git to deal with the repo, but we split into a `git log` that orders the commit graph and returns only IDs, and then we use `git show` to get those in patch form, in batches.  Peak memory usage with this change will still depend a bit on the specific repo, but it is more tied to the size of the largest commit in the repo rather than that _and_ the total length of the commit history.

I ran a series of tests on repos of different sizes to 1) ensure there's no loss of coverage and 2) gather time and memory numbers to judge a good batch size.

I can share full data but this is the short version.  The repos listed are ~2k, ~12k, ~250k, and ~1M commits in size. 

Max RSS (mb)
|           | base   | branch | change  | %    |
|-----------|-------:|-------:|--------:|-----:|
| watchexec | 54.4   | 25.8   | -28.6   | -53% |
| thog      | 1292.0 | 416.6  | -875.3  | -68% |
| rust      | 2346.7 | 688.0  | -1658.7 | -71% |
| firefox   | 7438.2 | 1431.9 | -6006.2 | -80% |

Total Run Time (s)
|           | base   | branch | change | %
|-----------|-------:|-------:|-------:|-----:|
| watchexec | 1.1    | 1.2    | 0.1    | 9%   | 
| thog      | 30.1   | 32.9   | 2.8    | 9%   |
| rust      | 352.4  | 327.1  | -25.3  | -7%  |
| firefox   | 2855.2 | 3029.2 | 174.0  | 6%   |

So, peak memory use is always smaller, but there's generally a slowdown, though it's not guaranteed.  The rust repo consistently ran faster with this branch, and at all reasonable batch sizes.

There may also may be some time that can be won back by changing the channel pattern here but I did test a few other options already.  I can also confirm there doesn't seem to be much value in parallelizing the `git show` commands (they are intentionally slightly _pipelined_ but not remarkably parallel here) - as you'd lose some of the memory benefit and the reader consuming `*Diff`s becomes a bottleneck.  If it can be improved enough later, maybe it doesn't need to be an alternate mode.

This mode is currently controlled by a hidden feature flag so we can give it further exercise or optimization.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes how git history is fetched and parsed on an opt-in path; bugs could affect scan completeness or ordering, though default behavior is unchanged.
> 
> **Overview**
> Adds an optional **low-memory git scan** path controlled by the hidden `--git-low-memory-scan` flag (wired to `feature.UseGitLowMemoryScan`).
> 
> When enabled, repo history is no longer streamed from a single `git log --patch`. The parser first runs `git log` for commit hashes only, batches them (75 commits per group), then fetches patches with grouped `git show` calls and merges the results back into one diff channel—aiming to cap peak memory closer to large commits than total history length, at some runtime cost.
> 
> Git argument construction is centralized in `prepGitArgs`, and `RepoPath`/`executeCommand` are refactored accordingly (including using the parser’s `waitDelay` and returning errors on pipe setup failures). The git source enables `gitparse.UseLowMemoryScan()` when the feature flag is on. Tests cover `prepGitArgs` and an integration-style low-memory scan.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bee83532499af21f3dd6d47151f12666aa6ea390. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/gitparse/gitparse.go
+++ b/pkg/gitparse/gitparse.go
@@ -3,11 +3,13 @@ package gitparse
 import (
 	"bufio"
 	"bytes"
+	"cmp"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -233,8 +235,17 @@ func NewParser(options ...Option) *Parser {
 	return parser
 }
 
+type gitArgs struct {
+	env    []string
+	global []string
+	log    []string
+	show   []string
+	paths  []string
+}
+
 // RepoPath parses the output of the `git log` command for the `source` path.
-// The Diff chan will return diffs in the order they are parsed from the log.
+// The Diff chan will return diffs in the order they are parsed from the log,
+// though the diffs are generated using `git show` in groups.
 func (c *Parser) RepoPath(
 	ctx context.Context,
 	source string,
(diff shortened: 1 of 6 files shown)


pkg/gitparse/gitparse.go
```diff
@@ -233,60 +251,213 @@ func NewParser(options ...Option) *Parser {
...
+		defer close(diffGroups)
+
+		for group := range commitGroups {
+			if common.IsDone(ctx) {
+				return
+			}
+
+			showCmd := exec.CommandContext(ctx,
+				"git",
+				slices.Concat(args.global, []string{"show"}, args.show, group, args.paths)...,
```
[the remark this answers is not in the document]
mariduv: Addressing this in a separate PR - the old path doesn't get merge commit output at all, new mis-parses but the parser resets and continues.

pkg/gitparse/gitparse.go
```diff
@@ -233,60 +251,213 @@ func NewParser(options ...Option) *Parser {
...
+			)
+			showCmd.Env = args.env
+
+			diffGroup, err := c.executeCommand(ctx, showCmd, false)
+			if err != nil {
+				ctx.Logger().Error(err, "Error executing git show for commit group.")
+				return
+			}
+			diffGroups <- diffGroup
```
shahzadhaider1, line 320: One thing I wasn't sure about: git.go breaks out of `for diff := range diffChan` on --max-depth and on reaching the base commit, and repoCtx never gets cancelled during a scan. These sends are unbuffered with no ctx.Done() case, so once nobody drains downstream they park permanently and a blocked channel send won't unwind on cancellation either. Might be misjudging how these scans terminate in practice.

pkg/gitparse/gitparse.go
```diff
@@ -233,60 +251,213 @@ func NewParser(options ...Option) *Parser {
...
+		}()
+
+		s := bufio.NewScanner(stdOut)
+		commitGroup := make([]string, 0, showGroupSize)
+
+		for s.Scan() && !common.IsDone(ctx) {
+			commitGroup = append(commitGroup, s.Text())
+
+			if len(commitGroup) == showGroupSize {
+				commitGroups <- commitGroup
```
shahzadhaider1, line 387: Same question here (and for the final partial-batch, if the consumer goroutine returns early on a git show error or on IsDone, this send has nobody receiving, so the deferred cmd.Wait() never runs and the git log process is left behind. A ctx.Done() case on the send would cover it?

shahzadhaider1 (commented):
Great work.

pkg/gitparse/gitparse.go
mariduv: Good point, I missed that!  I was focused on trying the channel patterns a few ways and forgot about context.  Fixing...

unsmith (approved):
My robot assistant noted this:

git show on a merge emits diff --cc / @@@ hunks, while FromReader expects diff --git / @@ from git log --patch. Every merge commit in the repo history will hit a parse failure and log an error. The parser resets and continues (no coverage loss), but high-merge repos could flood Sentry with error logs.

This seems fine since the flag is hidden and iteration is still happening, just wanted to note.

mariduv:
> git show on a merge emits diff --cc / @@@ hunks, while FromReader expects diff --git / @@ from git log --patch. Every merge commit in the repo history will hit a parse failure and log an error. The parser resets and continues (no coverage loss), but high-merge repos could flood Sentry with error logs.
> 
> This seems fine since the flag is hidden and iteration is still happening, just wanted to note.

Yeah, I have a followup for this (the other scan path simply doesn't see merge commits at all) and I think bots only picked it up bc i have a branch out about the problem 😂

PR #5258, opened 2026-08-31
[INT-942] Add HumioAPIToken detector

unsmith:
### Developer Commentary

This implements a detector and verifier for Humio API Tokens. This is less impactful than ingest tokens; we can hit the `/api/v1/health-json` endpoint to check things.  The token format is distinct enough that we don't need a prefix on the regex detector.

A small wrinkle - we don't currently (and might not ever) have access to a real, live Humio instance to try creating our own keys. This detector/verifier is best-effort based on research and testing with expired/invalidated tokens.

### Description:

Adds a new TruffleHog detector for Humio/CrowdStrike Falcon LogScale API tokens (INT-942). These are personal and repository-scoped tokens in a distinctive tilde-delimited base62 format (`{24 or 32}~{44}`) that grant read/write access to LogScale repositories via the REST and GraphQL APIs.

**Key design decisions:**

- **Regex**: Standalone (no PrefixRegex) since the tilde-delimited format is structurally distinctive. Matches exactly 24- or 32-character prefixes via `{24}(?:{8})?` to avoid accepting intermediate lengths.
- **Keywords**: `["humio", "logscale"]` for the Aho-Corasick pre-filter. The tilde character was considered but rejected as too noisy.
- **Verification**: Read-only `GET /api/v1/health-json` endpoint — no side effects in customer accounts. 200 = valid, 403 = recognized but IP-filtered (still verified), 401 = unknown/expired.
- **Multi-region**: `EndpointCustomizer`/`CloudProvider` support for US (`cloud.us.humio.com`), EU (`cloud.humio.com`), CrowdStrike LogScale regional instances, and self-hosted deployments discovered via URL patterns in scanned data.
- **Token type classification**: `ExtraData["token_type"]` reports "Personal API Token" (24-char prefix) or "Repository API Token" (32-char prefix).
- **Stale error handling**: Verification errors tracked in a local variable and set once after the endpoint loop to prevent `SetVerificationError(nil)` no-op from leaving stale errors.
- **URL regex**: Restricted to `cloud.*` subdomains on `humio.com` to avoid false verification against docs/marketing sites. CrowdStrike branch captures full hostnames including customer subdomains and regional labels.

**Note**: Enum value 1070 is shared with INT-943 (HumioIngestToken) on a parallel branch; conflict will be resolved at merge time.

Feature-flag gated behind `HumioAPITokenDetectorEnabled`.

### Checklist:
- [x] Lint passing (`make lint`)?


pkg/detectors/humioapitoken/humioapitoken_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	tests := []struct {
+		name    string
+		s       Scanner
+		args    args
+		want    []detectors.Result
+		wantErr bool
+	}{
+		{
+			name: "found, verified",
```
MuneebUllahKhan222, line 54: Use the input pattern similar to the tests in `solarwindsobservability_test.go`.

MuneebUllahKhan222 (approved):
LGTM!

pkg/detectors/humioapitoken/humioapitoken_integration_test.go
unsmith: Thanks for flagging that - as I compare, I'm seeing that the solar winds tests look more like real-world chunks of data which will be more realistic tests.  I'll update accordingly.

mariduv (approved):

PR #3440, opened 2024-10-16
Ignore invalid MongoDB connection strings

rgmz:
### Description:
This ignores matches for MongoDB that are not valid URLs and can never succeed verification.

```
Found unverified result 🐷🔑❓
Verification issue: net/url: invalid userinfo
Detector Type: MongoDB
Decoder Type: PLAIN
Raw result: mongodb+srv://<user>:<password>@myMongoCluster.mongocluster.cosmos.azure.com
Rotation_guide: https://howtorotate.com/docs/tutorials/mongo/
Commit: 23ddc72b02c957662d1e7b51914dd90bad6adeb0
Email: Peng Jiahui <46921893+Alancere@users.noreply.github.com>
File: sdk/resourcemanager/cosmos/armcosmos/mongoclusters_client_example_test.go
Line: 453
Link: https://github.com/Azure/azure-sdk-for-go/blob/23ddc72b02c957662d1e7b51914dd90bad6adeb0/sdk/resourcemanager/cosmos/armcosmos/mongoclusters_client_example_test.go#L453
Repository: https://github.com/Azure/azure-sdk-for-go.git
Timestamp: 2023-11-27 08:12:14 +0000
```

It's also a fix/follow-up to #2518. The connection string's query parameters and path should not be discarded.

### Checklist:
* [x] Tests passing (`make test-community`)?
* [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

PR #5259, opened 2026-08-31
Int 595 auth errors

shahzadhaider1:
These detectors dropped a finding entirely when verification failed for a reason other than "bad credential", so the result never reached the pipeline, `last_seen` was never updated, and a revoked secret kept looking live.

Stacked on #5253; review that one first.

One commit per detector:
- **mongodb**: an auth error `continue`d past the append, so a credential the server actively rejected produced no result at all. This restores the pre-#3440 handling of that branch; #3440's actual subject, skipping unparseable connection strings, is unchanged.
- **jdbc**: an unparseable connection string was dropped when `verify=true` and kept when `verify=false`. Now reported with the parse error.
- **postgres**: an expired deadline `break`ed the loop, abandoning the current candidate and every remaining one, valid credentials included. Now reported with `context.DeadlineExceeded`.
- **uri**: the host-not-found cache skipped the whole finding rather than just the DNS lookup, so later candidates for a dead host vanished, including in later chunks. The cache stays; it now suppresses the lookup only.
- **gitlab**: no drop bug here. `SetVerificationError` was missing the secret argument, so redaction wouldn't apply if an error path ever carried the token. Hardening only.

Verification logic is unchanged everywhere else, same status codes, same determinate/indeterminate classification.

Checked against real servers (mongo, postgres, mysql, nginx basic-auth containers). The last commit removes the per-detector feature flags added by the earlier commits, so the net diff adds none. Kept as a separate commit so it can be reverted if reviewers want them back.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes which secrets appear in scan output and how they are classified (verified / unverified / unknown), which can shift customer-facing counts and CI behavior when verification is enabled.
> 
> **Overview**
> This PR tightens **tri-state verification** across several detectors so findings stay consistent whether or not `--no-verification` is used, and so **auth rejections** are not treated like inconclusive failures.
> 
> **MongoDB** no longer drops matches when verification returns a determinate auth error; those stay **unverified with no verification error**. Only indeterminate failures (timeouts, unreachable hosts, etc.) call `SetVerificationError`.
> 
> **JDBC**, **URI**, and **Postgres** stop silently omitting candidates during verification: unparsed JDBC strings are emitted with a parse error; URI hits for hosts already known not to resolve are still reported with a verification error instead of `continue`; Postgres marks expired context deadlines as indeterminate instead of breaking out and skipping the rest of the queue.
> 
> **GitLab** v1–v3 pass the matched token into `SetVerificationError` so errors are redacted correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01c067bb01496b3a821bb9ab1e30c81959e1d9c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/mongodb/mongodb.go
+++ b/pkg/detectors/mongodb/mongodb.go
@@ -46,7 +46,7 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 	dataStr := string(data)
 
 	type mongoMatch struct {
-		password string
+		password  string
 		parsedURL *url.URL
 	}
 	uniqueMatches := make(map[string]mongoMatch)
@@ -102,8 +102,8 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		r := detectors.Result{
 			DetectorType: detector_typepb.DetectorType_MongoDB,
 			Raw:          []byte(connStr),
-			ExtraData:   extraData,
-			SecretParts: map[string]string{"key": connStr},
+			ExtraData:    extraData,
+			SecretParts:  map[string]string{"key": connStr},
 		}
 
 		if verify {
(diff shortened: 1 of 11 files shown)


pkg/detectors/mongodb/mongodb.go
```diff
@@ -102,8 +102,8 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 		r := detectors.Result{
 			DetectorType: detector_typepb.DetectorType_MongoDB,
 			Raw:          []byte(connStr),
-			ExtraData:   extraData,
-			SecretParts: map[string]string{"key": connStr},
+			ExtraData:    extraData,
+			SecretParts:  map[string]string{"key": connStr},
```
camgunz: My `gofmt` bumped a `map` field in a struct one space to the left yesterday... what is going on here I wonder....

camgunz (approved):

pkg/detectors/mongodb/mongodb.go
shahzadhaider1: We are but humble servants of gofmt. 😄

PR #5261, opened 2026-09-01
Int 595 verify asymmetry

shahzadhaider1:
## Problem
Setting `verify: true` changed *which* findings were reported, not just their verification status. Seven detectors dropped a candidate entirely when its host didn't resolve, so `last_seen` never advanced and a revoked secret stayed looking live.

## Change
Verification now only decorates a finding, never decides whether it exists. The host cache still suppresses the DNS lookup; the result is reported with the error attached.

- **Dropped findings** (artifactory, algoliaadminkey, azure_cosmosdb, azurecontainerregistry, azureapimanagement/repositorykey): `continue` skipped the `append`.
- **Misclassified transport errors** (azureapimanagementsubscriptionkey, azuredirectmanagementkey): `client.Do` errors returned `(false, nil)`, a determinate rejection that sets `date_rotated` and marks a live secret dead. Now `(false, err)`.
- In artifactory, algoliaadminkey and azure_cosmosdb the cache guard sat outside the `verify` block, so a globally cached host also dropped findings in later `verify: false` runs. Moved inside.
- DNS detection switched to `errors.As(&net.DNSError{}) && IsNotFound`. repositorykey's sentinel retexted to `"no such host"` so it classifies as DNS.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes secret-detection and verification semantics across multiple detectors; fixes incorrect “dead secret” signals but may increase reported findings and alter downstream rotation tracking when verify is on.
> 
> **Overview**
> Fixes **verify asymmetry** across several Azure-, Artifactory-, and Algolia-related detectors: enabling verification no longer changes *which* secrets are emitted when a host does not resolve or a request never reaches the server.
> 
> For **Algolia, Artifactory, Cosmos DB, ACR, and Azure API Management** detectors, unresolved hosts used to **`continue`** (and sometimes ran the invalid-host guard even when `verify` was false), so candidates were dropped entirely. Those paths now **always append** the result; cached bad hosts only skip the HTTP/git check and attach a verification error via **`SetVerificationError`**. DNS failures are detected with **`errors.As` → `*net.DNSError` (`IsNotFound`)** instead of string-matching `"no such host"` or custom wrappers in `verifyMatch`.
> 
> **Azure APIM subscription** and **direct management** keys now return **`(false, err)`** from `client.Do` failures instead of **`(false, nil)`**, so network/DNS problems are indeterminate verification errors rather than looking like a rejected credential (which could incorrectly drive rotation/`last_seen` behavior).
> 
> **Cosmos DB** sets **`ExtraData["DB Type"]`** from the account URL for every match, not only when verification runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41ac72aec875c866d5bae27b1925e92d5608ae3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/azure_cosmosdb/azure_cosmosdb.go
+++ b/pkg/detectors/azure_cosmosdb/azure_cosmosdb.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -74,11 +75,6 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
 
 	for key := range uniqueKeyMatches {
 		for accountUrl := range uniqueAccountMatches {
-			if invalidHosts.Exists(accountUrl) {
-				delete(uniqueAccountMatches, accountUrl)
-				continue
-			}
-
 			s1 := detectors.Result{
 				DetectorType: detector_typepb.DetectorType_AzureCosmosDBKeyIdentifiable,
 				Raw:          []byte(key),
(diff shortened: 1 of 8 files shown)


camgunz (approved):

pkg/detectors/azure_cosmosdb/azure_cosmosdb.go
```diff
@@ -90,30 +86,43 @@ func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (result
...
-				if strings.Contains(accountUrl, ".documents.azure.com") {
-					verified, verificationErr = verifyCosmosDocumentDB(client, accountUrl, key)
-					s1.ExtraData["DB Type"] = "Document"
+			if verify {
+				if invalidHosts.Exists(accountUrl) {
+					// An earlier candidate already proved this host does not resolve, so the lookup is
+					// skipped. The finding is still reported: dropping it here would make the reported
+					// secrets depend on whether verification was requested, and on the order in which
+					// candidates happened to be processed.
+					s1.SetVerificationError(errNoHost)
```
MuneebUllahKhan222: Should it be `s1.SetVerificationError(errNoHost, key)`

MuneebUllahKhan222 (approved):
LGTM!

The only thing we might want to change is to call `s1.SetVerificationError(err, key)` with the main secret, such as the key or token. We can also pass multiple secrets to it, so we could call it like this:

```go
s1.SetVerificationError(err, key, url)
```

PR #5272, opened 2026-09-03
s3: fail the scan when a configured bucket cannot be listed

kvnphvm:
### Description:

**Description authored by robot**

`ListObjectsV2` returning `AccessDenied` on a bucket named with `--bucket` finished as a
successful scan of an empty bucket, so a permissions problem was indistinguishable from a
clean result with no findings.

A bucket the user named is a target they expect to reach, so a listing or region-resolution
failure on one now fails the scan. Buckets that came from enumeration are unchanged: denials
there are routine and are still skipped.

| Buckets named | A bucket cannot be listed |
| --- | --- |
| `--bucket` given | scan fails, error names the bucket |
| none (enumerated) | bucket skipped, logged at V(3) |

The split is now just "did the user name buckets", where it previously also required an
assumed role. Scanning an account without a role is still enumeration, so it gets the same
treatment as the role case.

Three supporting changes fall out of returning the error instead of logging it:

- **Every bucket is still attempted.** `scanBuckets` collects failures and continues rather
  than returning on the first one, so an unreachable bucket cannot hide findings in the
  rest; the errors are joined. `Chunks` does the same across roles, since `visitRoles` stops
  at its first visitor error and a bucket one role cannot reach is often reachable under
  another.
- **Interruption is a clean stop, not a failure.** Cancellation and deadline-exceeded are
  filtered at the `Chunks` and `ChunkUnit` boundaries, so stopping a scan does not report the
  source as broken.
- **Resume state survives a failure.** Unit scans clear resume info only after a complete
  scan rather than unconditionally, so a retry continues from the last checkpoint. For
  multi-bucket passes an interruption keeps the checkpoint but rewinds it to the earliest
  failed bucket (new `Checkpointer.ResumeFrom`), so a retry cannot start past a bucket that
  never succeeded.

Per-object `GetObject` failures and the bucket-skip paths now log at Info/V(3) instead of
Error. They are expected during enumeration, and scan-level failure is carried by the
returned error rather than by log severity.

### Testing:

httptest-backed coverage for the matrix above: named vs enumerated denials, partial access
across several buckets, error aggregation, resume retention on mid-bucket failure, and resume
rewinding on interruption. Each resume/cancellation test was validated by reverting the fix
under test and confirming the test fails.

Verified end-to-end against a local S3-compatible mock with 3 readable and 2 denied buckets:

- enumerated: exit 0 even under `--fail-on-scan-errors`, all 5 buckets attempted, readable
  ones fully chunked
- named including denied: exit 1, terminal error names a denied bucket, readable ones still
  scanned
- named readable only: exit 0

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core S3 scan success criteria, error propagation, and resume/checkpoint semantics, which can alter exit codes and retry behavior for production scans.
> 
> **Overview**
> **Configured buckets that cannot be listed or resolved now fail the scan** instead of finishing as an empty bucket. **Enumerated buckets** (no explicit `--bucket` list) still treat list/region denials as skippable; the rule no longer depends on assuming a role.
> 
> **`scanBuckets` and `Chunks` propagate errors**: every configured bucket is still attempted, failures are joined, and per-role failures are aggregated because `visitRoles` stops on the first visitor error. **Cancellation and deadlines** are treated as a clean stop at the `Chunks`/`ChunkUnit` boundaries (not a source failure). **Resume behavior** keeps unit checkpoints until a bucket completes; multi-bucket passes can **`ResumeFrom`** the earliest failed bucket on interrupt so retries do not skip buckets that never succeeded. **`pageChunker`** stops scheduling on cancel but waits for in-flight workers before return so checkpoints cannot race with rewind.
> 
> Per-object **`GetObject`** issues log at Info rather than Error when scan-level failure is expressed via returned errors. Tests add httptest coverage for the named vs enumerated matrix, error aggregation, resume lifecycle, and interrupt rewind.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fab35314666ac7a8af7baded8cc2c5860c1e30f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/sources/s3/s3.go
```diff
@@ -397,7 +398,10 @@ func (s *Source) scanBuckets(
 			)
 		}
 
-		objectCount := s.scanBucket(ctx, client, role, bucket, sources.ChanReporter{Ch: chunksChan}, startAfter, checkpointer)
+		objectCount, err := s.scanBucket(ctx, client, role, bucket, sources.ChanReporter{Ch: chunksChan}, startAfter, checkpointer)
```
bill-rich: This used to try all the buckets, but now stops and returns on the first error. I believe it would be possible for there to be a mix of buckets with different list permissions.

pkg/sources/s3/s3.go
```diff
@@ -342,15 +342,16 @@ func determineResumePosition(ctx context.Context, tracker *Checkpointer, buckets
...
+// bucket the identity cannot list) are returned so the job is fatal.
 func (s *Source) scanBuckets(
 	ctx context.Context,
 	client *s3.Client,
 	role string,
 	bucketsToScan []string,
 	chunksChan chan *sources.Chunk,
 	totalObjectCount *uint64,
-) {
+) error {
```
bill-rich: The error now get propagated up and causes the scan to show up as failed if any buckets are unable to be scanned (like an access denied error). I think it would be worth limiting to only fatal errors getting sent up and somewhat expected errors get logged. Maybe also include a check based on number of buckets scanned successfully.

pkg/sources/s3/s3.go
```diff
@@ -417,22 +423,29 @@ func (s *Source) scanBucket(
...
-		return 0
+		// Same expected-vs-fatal split as listing: role enumeration walks every
+		// bucket in the account and will fail region lookup on many of them.
+		if s.listErrorsAreExpected(role) {
+			ctx.Logger().V(3).Info("could not get regional client for bucket", "err", err)
+			return 0, nil
+		}
+		// The returned error is the job-level report (unhealthy scan). This is
+		// customer access failure, so it is not logged at Error.
+		return 0, fmt.Errorf("could not get regional client for bucket %q: %w", bucket, err)
```
bill-rich: Similar to the other comments, but getting repeated transient errors like `SlowDown` would cause an error to be sent up. The errors that get returned can be scoped to certain ones to avoid that. Better error handling overall for things like retrying would be good at some point, but probably out of scope for the PR.

pkg/sources/s3/s3.go
```diff
@@ -417,22 +423,29 @@ func (s *Source) scanBucket(
...
-) uint64 {
+) (uint64, error) {
 	s.metricsCollector.RecordBucketForRole(role)
 
 	ctx = context.WithValue(ctx, "bucket", bucket)
 
 	if common.IsDone(ctx) {
 		ctx.Logger().Error(ctx.Err(), "context done, while scanning bucket")
-		return 0
+		return 0, ctx.Err()
```
bill-rich, line 488: Don't want a context error to be a fatal error for the scan.

pkg/sources/s3/s3.go
bill-rich: Any errors returned will now cause the scan to be classified as `FAILED` since this gets bubbled up to `Chunk`. Some errors are allowable for a scan to still be successful.

pkg/sources/s3/s3.go
```diff
@@ -397,7 +398,10 @@ func (s *Source) scanBuckets(
 			)
 		}
 
-		objectCount := s.scanBucket(ctx, client, role, bucket, sources.ChanReporter{Ch: chunksChan}, startAfter, checkpointer)
+		objectCount, err := s.scanBucket(ctx, client, role, bucket, sources.ChanReporter{Ch: chunksChan}, startAfter, checkpointer)
+		if err != nil {
```
bill-rich, line 438: `scanBuckets` will now stop and return on the first bucket error instead of trying all the ones in the list. Should probably try all and keep track of the errors to consolidate them later.

pkg/sources/s3/s3.go
```diff
@@ -847,8 +861,8 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
 		startAfterPtr = &startAfter
 	}
 	defer s.ClearEncodedResumeInfoFor(unitID)
```
bill-rich: It used to be that the scan would always run to the end, so this was ok. Now if a bucket fails part way through, it will have to start at the beginning again because the resume info is cleared. If the scan is going to stop for any type of error, the logic on this should change so it only clears when the scan is completed.

pkg/sources/s3/s3.go
bill-rich: I think it would be better change the error between the expected vs unexpected just to make it easier to tell which was triggered. The keys/bucket info is slightly different, but it is easy to overlook.

pkg/sources/s3/s3.go
```diff
@@ -448,20 +461,21 @@ func (s *Source) scanBucket(
...
+				return objectCount, nil
 			}
-			s.metricsCollector.RecordBucketListError(bucket, role)
-			break
+			// This can also be a failure to assume the role itself: role credentials
+			// are resolved lazily, so the first request that needs them surfaces the
+			// STS error here rather than at client construction.
+			// Returning the error marks the job fatal so a configured bucket the
+			// identity cannot list is not reported as a successful empty scan.
+			return objectCount, fmt.Errorf("could not list objects in bucket %q: %w", bucket, err)
```
bill-rich: This changes the behavior so that the first error returns instead of continuing to try as long as `HasMorePages`. That might be the right call, but I'm not familiar enough with how the SDK handled S3 to know for sure.

bill-rich (changes requested):
I put some comments, but they are mostly related to one thing that was left out in the original ticket. The list of buckets can come from just using creds and listing buckets, or they can come from an user-provided list of buckets. If the list if provided by the user, we want to return an error on any failure. If the listing just comes from the credentials, I believe we're fine to still say the scan is healthy (not bubbling up errors) and just log instead. 

I'd also like to see the `could not get S3 object` error logs downgraded to an info log.

pkg/sources/s3/s3.go
```diff
@@ -439,22 +445,29 @@ func (s *Source) scanBucket(
...
-) uint64 {
+) (uint64, error) {
 	s.metricsCollector.RecordBucketForRole(role)
 
 	ctx = context.WithValue(ctx, "bucket", bucket)
 
 	if common.IsDone(ctx) {
 		ctx.Logger().Error(ctx.Err(), "context done, while scanning bucket")
-		return 0
+		return 0, ctx.Err()
```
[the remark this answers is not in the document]
KevinPham-TruffleSec: Fixes from Bill's comment addresses.

pkg/sources/s3/s3.go
```diff
@@ -888,8 +902,8 @@ func (s *Source) ChunkUnit(ctx context.Context, unit sources.SourceUnit, reporte
 		startAfterPtr = &startAfter
 	}
 	defer s.ClearEncodedResumeInfoFor(unitID)
-	s.scanBucket(ctx, defaultClient, s3unit.Role, s3unit.Bucket, reporter, startAfterPtr, checkpointer)
-	return nil
+	_, err = s.scanBucket(ctx, defaultClient, s3unit.Role, s3unit.Bucket, reporter, startAfterPtr, checkpointer)
+	return err
```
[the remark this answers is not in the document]
KevinPham-TruffleSec: Fixes from Bill's comment addresses.

pkg/sources/s3/s3.go
KevinPham-TruffleSec: Resolved logs instead of error. Test created.

pkg/sources/s3/s3.go
KevinPham-TruffleSec: Resolved, return the errors together now, test created ensuring every bucket is still attempted and failures appear in the error list

pkg/sources/s3/s3.go
KevinPham-TruffleSec: ClearEncodedResumeInfoFor should now only run after a unit scans to completion rather than defer. Test created.

pkg/sources/s3/s3.go
KevinPham-TruffleSec: Agreed. Done.

pkg/sources/s3/s3.go
KevinPham-TruffleSec: main did the same thingg, it just used break instead of returning, so the only real change is that the error now makes it out instead of the bucket looking like it finished. we can't keep paginating anyway, since a failed NextPage leaves the token untouched and just replays the same request forever.

PR #5273, opened 2026-09-03
Clean up use of wasilibs/go-re2

bradlarsen:
### Overview

TruffleHog makes extensive use of regular expressions in its detectors and its decoders. A while back, _most_ detectors were switched from using stdlib `regexp` to use `wasilibs/go-re2`, a build of the C++ `re2` library compiled to WASM. That implementation typically has better performance when dealing with complicated patterns or large inputs. However, today there are a couple dozen remaining uses of `regexp` (perhaps ones that were added after the initial switchover).

This PR switches over all but 2 extant uses of stdlib `regexp`, and updates the `wasilibs/go-re2` dependency to the latest version. These changes appear to improve scanning performance across the board, giving 1.08-1.15x wall clock speedup when tested with the `json-enumerator` source.

### Benchmarks

I ran a simple benchmark to quantify the performance impacts of this change.

**Test machine:** a 12-core Nov 2024 M4 Pro MBP
**TruffleHog versions:** `main` baseline; `test` this branch
**Input:** 10 files of zstd-compressed NDJSON, 8.5GiB on disk, given to the `json-enumerator` source
**Verification:** enabled (default) or disabled (`--no-verification`)

Verification disabled emphasizes CPU performance, as it eliminates the network component of the workload.
The scans take 4-6min of real time and 30-45 CPU minutes.

**When verification is disabled:** 1.15x wall clock speedup vs baseline; 1.20x user+sys speedup

**When verification is enabled:** 1.08x wall clock speedup vs baseline; 1.17x user+sys speedup

Much of the speedup comes from the "escaped unicode" decoder running much faster with `wasilibs/go-re2`. (There is still potential for significant further improvement in that part of the code...)

### Details:

- Update `wasilibs/go-re2` from v1.9.0 to v1.12.0

- Switch nearly all remaining uses of stdlib `regexp` to `wasilibs/go-re2`

- Hoisted a constant regexp compilation operation into a package-level variable for the Snowflake detector

- 2 detectors still use stdlib `regexp`, as it handles invalid UTF-8 differently than `wasilibs/go-re2`. With a bit of investigation, it appeared that this _might_ be significant in those cases (`jdbc` and custom detectors). I've taken the conservative approach here to avoid possible regressions from that.

---

> [!NOTE]
> **Medium Risk**
> Broad changes across secret detection and decoding paths could alter match behavior versus stdlib regex, though exceptions were kept for known UTF-8-sensitive cases.
> 
> **Overview**
> This PR **migrates almost all regular-expression usage from stdlib `regexp` to `github.com/wasilibs/go-re2`**, bumps that library (and related wazero deps) to **v1.12.0**, and adds **golangci `depguard`** so new stdlib `regexp` imports are blocked unless explicitly exempted.
> 
> Touched areas include **decoders** (notably escaped Unicode/HTML), **path filters** (`FilterRuleSet` now stores `*regexp.Regexp` instead of copied values), **analyzers**, **handlers**, **git sources**, and several **detectors**. **Snowflake** hoists username/password regex helpers to package-level vars to avoid recompiling on every scan.
> 
> **Stdlib `regexp` is intentionally kept** (with `//nolint:depguard`) in **JDBC** and **custom detector** code paths where UTF-8 handling differs from go-re2; tests were adjusted where compiled-regex identity checks no longer apply.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e96f42772e87b51a2e941593474256befd2e6545. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashifkhan0771 (approved):

mariduv (approved):
If you don't mind, we can add a `.golangci.yml`:

```yaml
version: "2"
linters:
  enable:
    - depguard
  settings:
    depguard:
      rules:
        no-stdlib-regexp:
          deny:
            - pkg: "regexp$"
              desc: "use wasilibs/go-re2 unless necessary"
```

and change the couple exception comments to

` //nolint:depguard // used instead of github.com/wasilibs/go-re2 due to differences in utf-8 handling`

to keep regexp banished.  I can also do this in a quick followup PR too.

bradlarsen:
Good suggestion @mariduv!  I went ahead and added that. Thanks for the review.

PR #5277, opened 2026-09-05
Implement wakatime detector

kvnphvm:
### Description:
Implement wakatime previously unimplemented as part of a POC for learning.

### Checklist:
* [ X] Tests passing (`make test-community`)?
* [ X] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?

---

> [!NOTE]
> **Low Risk**
> Isolated new detector following existing TruffleHog patterns; verification only performs outbound HTTP when scanning with verify enabled.
> 
> **Overview**
> Adds a **WakaTime** secret scanner for the existing `DetectorType_Wakatime` enum entry, which was previously marked as not implemented.
> 
> The detector finds `waka_`-prefixed UUID v4 keys via regex (keyword prefilter `waka_`), deduplicates matches, and optionally verifies them with HTTP basic auth against `GET /api/v1/users/current` (200 = verified, 401 = invalid). It is registered in the default detector list, and WakaTime is removed from the engine’s “not yet implemented” test set. Proto comments are updated to reflect that the type is implemented.
> 
> Unit tests cover regex/Aho–Corasick matching; integration tests (build tag `detectors`) cover live verification, inactive keys, timeouts, and unexpected API responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb0c00c1f2946eb3652bc9af3e6976a5679ef1e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/wakatime/wakatime.go
+++ b/pkg/detectors/wakatime/wakatime.go
@@ -1,0 +1,104 @@
+package wakatime
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+
+	regexp "github.com/wasilibs/go-re2"
+
+	"github.com/trufflesecurity/trufflehog/v3/pkg/common"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/detectors"
+	"github.com/trufflesecurity/trufflehog/v3/pkg/pb/detector_typepb"
+)
+
+type Scanner struct {
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
+	keyPat = regexp.MustCompile(`\b(waka_[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\b`)
+)
+
+// Keywords are used for efficiently pre-filtering chunks.
+// Use identifiers in the secret preferably, or the provider name.
+func (s Scanner) Keywords() []string {
+	return []string{"waka_"}
+}
+
+// FromData will find and optionally verify Wakatime secrets in a given set of bytes.
+func (s Scanner) FromData(ctx context.Context, verify bool, data []byte) (results []detectors.Result, err error) {
+	dataStr := string(data)
+
+	uniqueMatches := make(map[string]struct{})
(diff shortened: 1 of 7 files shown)


pkg/detectors/wakatime/wakatime.go
```diff
@@ -0,0 +1,104 @@
...
+type Scanner struct {
+	client *http.Client
+}
+
+// Ensure the Scanner satisfies the interface at compile time.
+var _ detectors.Detector = (*Scanner)(nil)
+
+var (
+	defaultClient = common.SaneHttpClient()
+	// Make sure that your group is surrounded in boundary characters such as below to reduce false positives.
```
unsmith, line 25: I'd remove this comment.

pkg/detectors/wakatime/wakatime.go
```diff
@@ -0,0 +1,104 @@
...
+	keyPat = regexp.MustCompile(`\b(waka_[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\b`)
```
unsmith, line 26: I think this needs to be case-insensitive, based on https://github.com/wakatime/vscode-wakatime/blob/9378973d/src/utils.ts#L26.

pkg/detectors/wakatime/wakatime.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, res.Body)
+		_ = res.Body.Close()
+	}()
+
+	switch res.StatusCode {
+	case http.StatusOK:
+		// If the endpoint returns useful information, we can return it as a map.
+		return true, nil, nil
```
unsmith, line 89: The comment appears to be at odds with the code here.  If there's something in the response to capture, we should do that instead of returning nil.

pkg/detectors/wakatime/wakatime.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		results = append(results, s1)
+	}
+
+	return
+}
+
+func verifyMatch(ctx context.Context, client *http.Client, token string) (bool, map[string]string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://wakatime.com/api/v1/users/current", nil)
```
unsmith, line 70: I think the convention here is to use a named constant for the verify URL instead of inlining it here.

unsmith (changes requested):
The detector gap is a blocker here, and I'd like to see some other stuff cleaned up, boilerplate comments removed etc.  Seems solid otherwise!

PR #5286, opened 2026-09-07
Object filtering in S3

shahzadhaider1:
Adds four flags to the `s3` command for filtering objects within a bucket, so a scan can skip archives, media, and build artifacts instead of downloading everything.

- `--include-prefix` / `--exclude-prefix` match a literal key prefix
- `--include-extension` / `--exclude-extension` match the file extension, written without a leading dot

Matching happens on the `ListObjectsV2` metadata, before `GetObject`, so a filtered object costs no download. Both scan paths (`Chunks` and `ChunkUnit`) share `pageChunker`, so the filter applies to both. Behavior is unchanged when nothing is configured.

Corresponding `include_prefixes`, `exclude_prefixes`, `include_extensions`, and `exclude_extensions` fields on the S3 connection proto.

**Include and exclude interact differently per axis.** Prefixes accept both lists at once, since scanning a subtree minus one directory is a common shape, and exclusion wins on a key matching both. Extensions accept only one, because naming the extensions to scan already excludes every other one; both together fails at startup, matching how buckets already behave.

**Skipped objects are still checkpointed.** Skipping without calling `UpdateObjectCompletion` would stall the resume low-water mark at the first filtered object and make a resumed scan redo everything after it. There's a test for this specifically.

Adds `pkg/sources/s3/README.md` covering the source as a whole, including the custom endpoint support from #5202, which was never documented.

### Known limits

Filtering runs after listing, so it cuts `GetObject` cost but not `ListObjectsV2` cost, a bucket is still paginated in full. Pushing prefixes into `ListObjectsV2Input.Prefix` would fix that and is worth a follow-up; it needs `ResumeInfo` to carry a checkpoint per prefix.

Extension matching uses the last extension only, so `tar.gz` matches nothing while `gz` works. A dotfile counts as all extension, so `.env` has extension `env; an include list omitting it will skip every `.env` in the bucket. Both are documented in the README.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Misconfigured include lists can silently shrink scan coverage and miss secrets; resume/checkpoint behavior on filtered keys is security-relevant but covered by tests.
> 
> **Overview**
> Adds **object-level filtering** to `trufflehog s3` so scans can skip unwanted keys before `GetObject`, cutting download cost on large buckets.
> 
> Four repeatable CLI flags (and matching S3 proto / `S3Config` fields) wire through to a new **`objectFilter`**: `--include-prefix` / `--exclude-prefix` (literal `HasPrefix` matching; exclude wins when both apply) and `--include-extension` / `--exclude-extension` (case-insensitive, last extension only; the two extension modes cannot be combined at startup).
> 
> Filtering runs in **`pageChunker`** on list metadata only—listing is still full-bucket pagination. Filtered objects are **checkpointed** like other skips so resume does not redo work after the first excluded key. When a filter removes every object, the scan logs **`Scanned no objects in bucket`** with an excluded count.
> 
> Docs cover behavior and edge cases (e.g. `.env` as extension `env`, archives not inspected by extension rules) in the root README, man page, and new **`pkg/sources/s3/README.md`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad842f482f6383209f46ad94fce6beaa8de409ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/sources/s3/s3.go
```diff
@@ -471,13 +493,25 @@ func (s *Source) scanBucket(
...
 		}
 		s.pageChunker(ctx, pageMetadata, processingState, reporter, checkpointer)
 
 		pageNumber++
 	}
+
+	// A filter that excludes everything otherwise looks exactly like a clean scan
+	// of an empty bucket, so say so rather than finishing silently.
+	if objectCount == 0 && filteredCount > 0 {
+		if len(s.conn.GetBuckets()) > 0 {
```
MuneebUllahKhan222: I think we should log this at `V(0)` regardless of whether the user explicitly configured any buckets.

pkg/sources/s3/s3.go
```diff
@@ -112,6 +114,25 @@ func (s *Source) Init(
...
+	if err != nil {
+		return err
+	}
+	s.objectFilter = filter
+
+	if filter.isConfigured() {
+		ctx.Logger().V(1).Info("Object filter configured",
+			"include_prefixes", filter.includePrefixes,
+			"exclude_prefixes", filter.excludePrefixes,
+			"include_extensions", sortedExtensions(filter.includeExtensions),
```
MuneebUllahKhan222: There is no apparent need to sort extension. Maybe we can remove this.

MuneebUllahKhan222 (approved):
LGTM. I just requested two small, non-blocking changes that can be ignored.

Also, maybe we should add an integration test showcasing the filtering logic that was added.

I also think the design documented for the S3 source should be added in a separate PR.

pkg/sources/s3/s3.go
shahzadhaider1: The reason for the split was the role case: with `--role-arn` and no `--bucket` we walk every bucket in the account, so a narrow prefix means most of them legitimately filter everything out and each logs a line. I was mirroring `listErrorsAreExpected` just below, which makes the same explicit-vs-enumerated distinction for the same reason.

dipto-truffle (approved):
Approving for scanning owned files

PR #5291, opened 2026-09-09
perf(engine): lowercase prefilter chunks as ASCII in a pooled buffer

kashifkhan0771:
### Description:
FindDetectorMatches searched a bytes.ToLower copy but cut spans from the original chunk, so a rune whose lowercase differs in byte length shifted every later position and could hide a credential. Folding only A-Z keeps the lengths equal and a pooled buffer drops the copy's allocation. Adds a package README.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes the hot path for every scanned chunk and fixes span logic that could mis-route detector input or miss secrets; behavior is heavily tested but custom detectors with non-ASCII cased keywords may match differently than before.
> 
> **Overview**
> **Fixes a correctness bug and cuts per-chunk allocations** in the Aho–Corasick keyword prefilter used to decide which detectors run on each scan chunk.
> 
> `FindDetectorMatches` no longer uses `bytes.ToLower` / `strings.ToLower`. It builds a search copy with **`appendASCIILower`** (only `A-Z` → `a-z`) and matches against that while still slicing **spans from the original bytes**. Full Unicode lowercasing can change byte length (e.g. `İ`), which desynchronized match positions and could hand detectors the wrong window or miss hits; length-preserving ASCII folding keeps offsets aligned. Keywords are registered with the same folding so chunk and trie stay consistent.
> 
> A **`sync.Pool`** of reusable lowercasing buffers (`lowerBuf`, 1 MiB max retained size) replaces allocating a fresh copy every chunk—benchmarks show **0 allocs/op** vs one allocation per chunk when unpooled.
> 
> Adds a **package README** documenting prefilter behavior, the ASCII-only folding tradeoff (custom non-ASCII cased keywords), and the pool. **Tests** cover case insensitivity, multibyte padding, property/fuzz checks on spans, pool reuse, concurrency, default detector keywords all ASCII, and `BenchmarkLowerBuf`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8aa1f5ff40c3eeb5b29caea844fa08fd6abf6606. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/engine/ahocorasick/ahocorasickcore.go
+++ b/pkg/engine/ahocorasick/ahocorasickcore.go
@@ -1,8 +1,7 @@
 package ahocorasick
 
 import (
-	"bytes"
-	"strings"
+	"sync"
 
 	ahocorasick "github.com/BobuSumisu/aho-corasick"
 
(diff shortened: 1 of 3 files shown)


bradlarsen:
This sounds good in theory! Do you have any experimental results showing impact (like time or memory use improvements)?

pkg/engine/ahocorasick/ahocorasickcore.go
```diff
@@ -11,6 +12,92 @@ import (
...
+//
+// A keyword that is plain ASCII, or that uses a script without letter case such
+// as Chinese or Arabic, folds to the same string both ways and is recorded once.
+//
+// Keywords mixing cases of several non-ASCII letters, such as a capitalised
+// Cyrillic word, are not fully covered. Handling those would mean recording one
+// spelling per combination of those letters.
+func keywordSpellings(kw string) []string {
+	lower := string(appendASCIILower(nil, []byte(strings.ToLower(kw))))
+	upper := string(appendASCIILower(nil, []byte(strings.ToUpper(kw))))
```
bradlarsen: Is there a problem here by using `strings.ToLower` on the keyword, which will also lowercase non-ascii unicode characters, whereas the new explicit lower-casing via `appendASCIILower` only touches ASCII A-Z?

kashifkhan0771: Using Unicode lowercasing on the chunk can change the byte length - İ is 2 bytes, its lowercase i is 1. We search the lowercased copy but cut the span out of the original chunk, so any length change shifts every position after it and the detector gets handed the wrong bytes. ASCII-only folding always keeps the same length, which is what makes the positions safe.
Once the chunk is folded that way, keywords have to be folded the same way or the two sides stop agreeing. Since A-Z folding leaves É and é as different bytes, each keyword is recorded under both its lowercase and uppercase spelling (café → ["café", "cafÉ"]) so either casing in the text still matches.

kashifkhan0771:
> This sounds good in theory! Do you have any experimental results showing impact (like time or memory use improvements)?

Yes, I added the benchmark test case for that and also updated the README.md with the results.

pkg/engine/ahocorasick/ahocorasickcore_test.go
```diff
@@ -434,3 +439,225 @@ func TestFindDetectorMatches(t *testing.T) {
...
+	chunk := []byte(strings.Repeat("İ", 700) + " padding " + keyword + " trailing")
+
+	matches := core.FindDetectorMatches(chunk)
+	assert.Len(t, matches, 1)
+	for _, m := range matches[0].matches {
+		assert.Contains(t, string(m), keyword)
+	}
+}
+
+// No input may produce a span that would panic when used to slice the chunk.
```
rosecodym: This is a _great_ candidate for property-based testing. We use `rapid` in some other tests in the codebase - you can use that as prior art.

rosecodym (commented):
If I understand the problem correctly, you could also solve it by cutting the detector-specific spans based on the _lowercased_ chunk data, not the original chunk data. Do I understand correctly? And if so, why did you choose this approach over that one?

kashifkhan0771:
> If I understand the problem correctly, you could also solve it by cutting the detector-specific spans based on the _lowercased_ chunk data, not the original chunk data. Do I understand correctly? And if so, why did you choose this approach over that one?

You've got the mechanism right, but those spans are the exact bytes we hand to the detector, and whatever it finds in them becomes the reported secret. Secrets are case-sensitive, so cutting from the lowercased copy would hand back a mangled credential that can fail verification - the copy can only tell us where a keyword is, never supply the bytes.

rosecodym (approved):

PR #5296, opened 2026-09-09
Postgres: drop non-connection URI params before verifying

jordanTunstill:
### Description:
For Postgres, drop non-connection URI params before verifying
Prisma and other ORMs append query-string arguments such as `schema`,
`connection_limit`, and `pgbouncer` to Postgres connection URIs. These
are not libpq connection keywords, so lib/pq and pgx forward them to the
server as startup runtime parameters, which the server rejects with
42704 ("unrecognized configuration parameter") before authentication can
resolve. This trapped valid credentials in indeterminate verification.
This change adds a denylist of known non-connection params out of the connection
string on both the lib/pq and pgx (Neon) paths. Each entry was diffed
against the libpq keyword list; params that ORMs share with libpq
(connect_timeout, sslmode, sslcert) are intentionally kept. 
### Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped to Postgres verification string assembly with denylist filtering and unit tests; no auth logic changes beyond avoiding bogus server GUC errors.
> 
> **Overview**
> Fixes **indeterminate Postgres verification** when connection URIs include ORM-specific query arguments (e.g. Prisma’s `schema`, `connection_limit`, `pgbouncer`). Those keys are not libpq keywords; **lib/pq** and **pgx** would forward them as server startup parameters and the server returns **42704** before auth completes, so valid secrets could not be classified.
> 
> The detector now maintains a **denylist** of known non-connection params and **drops them** when building strings for both the **pgx (Neon)** path (`pgxConnString`) and the **lib/pq** path (`verifyPostgresPq`). Real shared keywords like `connect_timeout` and `sslmode` stay.
> 
> **`connStrPartPattern`** is updated to allow **underscores** in parameter names so keys like `connection_limit` parse fully instead of being truncated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e830e8b62d3af08db5c168b71426677a9a2d4e14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/postgres/postgres.go
+++ b/pkg/detectors/postgres/postgres.go
@@ -38,6 +38,24 @@ const (
 	pgDbType         = "db_type"
 )
 
+// nonConnectionParams are query-string arguments that ORMs append to
+// Postgres connection URIs but that are not libpq connection keywords. lib/pq and pgx
+// forward any key they don't recognize to the server as a startup runtime parameter,
+// which the server rejects with 42704. Exluding them prevents this
+var nonConnectionParams = map[string]struct{}{
+	"schema":           {}, // Prisma: search_path selector
+	"connection_limit": {}, // Prisma: client-side pool size
+	"pool_timeout":     {}, // Prisma: pool-acquisition wait
+	"socket_timeout":   {}, // Prisma: per-query timeout
+	"pgbouncer":        {}, // Prisma: PgBouncer compatibility mode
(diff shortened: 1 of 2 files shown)


kashifkhan0771 (approved):

mariduv (approved):
This might be better as an allow-list instead but I'll make a ticket for someone to sort that out later. 👍

PR #5298, opened 2026-09-10
Introduce a new optional detector interface that will allow us to verify credentials not in the cache

kvnphvm:
### Description:
It sounds like from brief git history, this was a design choice because allowing for verification of those in a cache miss would require updating implementation for all of our detectors. So, maybe this should be a 1-off basis depending on the detector.

Today, a span can have 2 outcomes, if every candidate in it is cached, the facade returns cached answers and makes zero network calls. If at least 1 candidate is missing, it discards the lookups it already did and re-verifies the entire span. So for a span with N candidates, the cost is either 0 requests or N requests.

At 100 candidates per span: 23,480 verifications for 4,096 distinct pairs. The engine reported 239 spans containing at least one cache miss, and each span holds ~98 candidates — 239 × 98 ≈ 23,400. Most of those requests re-checked credentials the cache already knew, purely because of one cache miss.

image

ACTUAL TLDR:
- OSS today remembers which credentials it has already checked, but it decides whether to use the cache is all-or-nothing.
- If at least 1 cache miss exists, it re-checks every one of them.

The fix is to verify only those that the cache miss, instead of all for Github OAuth2

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes how cached verification is applied for opt-in detectors and refactors GitHub OAuth2 remote checks; behavior is heavily tested but incorrect cache/miss handling would affect verification accuracy and API load.
> 
> **Overview**
> Fixes **all-or-nothing verification caching**: when any candidate in a chunk missed the cache, the engine used to re-run `FromData(verify=true)` and re-check every pair. Chunks with many GitHub OAuth2 ID/secret combinations made that especially costly.
> 
> This PR adds an optional **`detectors.ResultVerifier`** hook so the verification cache can call **`verifyCacheMisses`**: one extraction pass with `verify=false`, then apply cached status per result and remotely verify **only cache misses** (including duplicates within a chunk after the first verify populates the cache). **`forceCacheUpdate`** and **`verify=false`** still follow the existing full-chunk paths.
> 
> **GitHub OAuth2** is the first implementer: verification moves into shared **`VerifyResult`**, used from both `FromData` and the cache. Metrics text now counts verify time from per-result calls too. Tests cover partial/full/no cache hits, duplicates, and bypass cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5aaf068fce193c8c77f118e3a7fd87f6e324fd34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/detectors.go
+++ b/pkg/detectors/detectors.go
@@ -375,3 +375,10 @@ func withDedupKey(ctx context.Context, detType detector_typepb.DetectorType, cre
 func DoWithDedup(client *http.Client, detType detector_typepb.DetectorType, credential string, req *http.Request) (*http.Response, error) {
 	return client.Do(req.WithContext(withDedupKey(req.Context(), detType, credential)))
 }
+
+// ResultVerifier is an optional interface that a detector can implement to verify a single
+// previously-extracted result independently of the chunk it came from, which lets the
+// verification cache verify only cache misses instead of re-verifying an entire chunk.
+type ResultVerifier interface {
+	VerifyResult(ctx context.Context, result *Result)
+}
--- a/pkg/detectors/github_oauth2/github_oauth2.go
+++ b/pkg/detectors/github_oauth2/github_oauth2.go
@@ -19,10 +19,20 @@ type Scanner struct {
 // Ensure the Scanner satisfies the interfaces at compile time.
 var _ detectors.Detector = (*Scanner)(nil)
 
+// This detector emits the cartesian product of every client ID and client secret in a
+// chunk, so chunks routinely carry dozens of candidate pairs. Implementing ResultVerifier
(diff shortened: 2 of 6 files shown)


KevinPham-TruffleSec:
Should clearing of secret parts happen in this PR? @rosecodym

casey-tran:
> Should clearing of secret parts happen in this PR? @rosecodym

@KevinPham-TruffleSec That will be a separate PR that I will handle [here](https://github.com/trufflesecurity/trufflehog/pull/5318).

camgunz (approved):
I don't believe this will make our lives more difficult down the road; lgtm 👍🏻

PR #5301, opened 2026-09-10
return error body for fetchBearerToken

jordanTunstill:
### Description:
For twitterconsumerkey detector, surface token-endpoint error body in verification errors
### Checklist:

---

> [!NOTE]
> **Low Risk**
> Only enriches error strings on failed token requests; success path and credential handling are unchanged.
> 
> **Overview**
> When **Twitter consumer key** verification fails at the OAuth2 token endpoint, errors now include a trimmed snippet of the HTTP response body (up to 1024 bytes) instead of only the status code.
> 
> This makes `fetchBearerToken` failures easier to debug (e.g. invalid credentials vs rate limits) when results are marked with `SetVerificationError` during verification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit adca0ebd6373730868572843cab1041867664748. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/detectors/twitterconsumerkey/twitterconsumerkey.go
+++ b/pkg/detectors/twitterconsumerkey/twitterconsumerkey.go
@@ -5,6 +5,7 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
@@ -142,6 +143,10 @@ func fetchBearerToken(ctx context.Context, client *http.Client, key, secret stri
 		}
 		return token.AccessToken, nil
 	default:
+		body, _ := io.ReadAll(io.LimitReader(res.Body, 1024))
+		if len(body) > 0 {
+			return "", fmt.Errorf("unexpected HTTP response status %d: %s", res.StatusCode, strings.TrimSpace(string(body)))
+		}
 		return "", fmt.Errorf("unexpected HTTP response status %d", res.StatusCode)
 	}
 }


MuneebUllahKhan222 (approved):
LGTM. IMO, it would be better not to limit the response body to `1024 bytes` and instead log the full response body. This would make it easier to determine why the verification is failing.

pkg/detectors/twitterconsumerkey/twitterconsumerkey.go
```diff
@@ -142,6 +143,10 @@ func fetchBearerToken(ctx context.Context, client *http.Client, key, secret stri
 		}
 		return token.AccessToken, nil
 	default:
+		body, _ := io.ReadAll(io.LimitReader(res.Body, 1024))
+		if len(body) > 0 {
+			return "", fmt.Errorf("unexpected HTTP response status %d: %s", res.StatusCode, strings.TrimSpace(string(body)))
+		}
```
shahzadhaider1, line 149: The error is swallowed here and what we return doesn't make sense to me, a truncated fragment ships as the error while the read failure stays hidden.

jordanTunstill: I'll address this today!

jordanTunstill:
> IMO, it would be better not to limit the response body to `1024 bytes` and instead log the full response body. This would make it easier to determine why the verification is failing.

Just removed this, thanks!

pkg/detectors/twitterconsumerkey/twitterconsumerkey.go
jordanTunstill: It uses the whole response body now, ty

mariduv (approved):

PR #5320, opened 2026-09-16
[PERF] Remove \b anchor from escaped-unicode codePointPat

Ahsan-Sarbaz:
### Description:

`codePointPat` in the escaped-unicode decoder dominated the cost of that decoder. Profiling a 200 MiB scan put `EscapedUnicode.FromChunk` at 4.2% of CPU, and 4.56s of its 4.71s was one line: `codePointPat.Match`.

The cause is the leading `\b`. Both regexp engines extract a required literal prefix and use a fast memory search to jump straight to candidate offsets. An empty-width assertion in front blocks that:

```go
regexp.MustCompile(`\bU\+([a-fA-F0-9]{4}).?`).LiteralPrefix()  // ""
regexp.MustCompile(`U\+([a-fA-F0-9]{4}).?`).LiteralPrefix()    // "U+"
regexp.MustCompile(`\\U([a-fA-F0-9]{8})`).LiteralPrefix()      // "\U"
```

The seven sibling patterns all have a prefix and are cheap. This one had none, so it was tried at every offset in every chunk. Worst case is a chunk with no match, since `Match` short-circuits when one exists. Over 99% of chunks contain no `U+` at all.

Measured on an 11.2 KB chunk, `Match` alone, after the move to `wasilibs/go-re2` in #5273:

| pattern | go-re2 | stdlib regexp |
|---|---|---|
| `\bU\+...` | 39.0 µs | 153.7 µs |
| `U\+...` | 2.8 µs | 0.22 µs |

**Fix:** drop the `\b` and apply the boundary in Go. `U` is a word character, so `\b` holds exactly when the preceding byte is not one. `codePointMatches` filters on that; match indices and group numbering are unchanged.

**Second change:** each pattern needs one fixed literal to be present (`\`, `&#x`, `%u`, `U+`). Searching for it is cheaper than running the pattern, so each pattern now runs only when its anchor is there. This skips all eight patterns for ~99.6% of chunks. It is a shortcut, not the fix: `CPU+GPU` is ordinary text containing `U+`, so it defeats the anchor but not the `\b` removal.

`FromChunk` also needs the match list to decide whether this pattern applies, which previously meant scanning once to decide and once to decode. It now scans once and hands the indices to the decode step.

### Equivalence

The regex engine advances one byte when `\b` fails; the filter skips the whole rejected match. That cannot hide a good match: `+` is not a hex digit, so a second `U+` cannot start inside the four digits, and the only other overlap is
the trailing optional byte covering a `U` that follows a hex digit, which `\b` would reject anyway.

Checked rather than argued. `escaped_unicode_equivalence_test.go` keeps the old implementation and compares match indices, decoded output, and the decode/skip decision against it:

- 600,000 generated inputs over a `U`/`+`/hex/underscore/whitespace alphabet,
  plus 32 hand-picked cases (`AU+0041`, `U+0041U+0042`, `U+0041\nU+0042`,
  `9U+0041`, `U+ZZZZ`, `UU+0041`, invalid UTF-8)
- `go test -fuzz`, 3,179,065 execs

Zero divergence. The baseline in that test is built on `go-re2`, the same engine the decoder uses, so it holds the engine fixed and varies only this change.

### Numbers

M3 Pro, `FromChunk` on an 11.2 KB chunk, medians of 8 runs:

| workload | before | after |
|---|---|---|
| ordinary text, no `U+` | 89.4 µs | 1.22 µs |
| contains `U+`, no real match | 92.7 µs | 6.24 µs |
| dense real `U+XXXX` | 489.0 µs | 422.1 µs |

Row 1 is the common case and the literal gates carry it. Row 2 is the case they cannot help with, since the gate passes and the pattern still runs; only the `\b` removal helps there. Row 3 shows the Go-side boundary check costs nothing
even when every match is genuine.

Full scan of a 200 MiB corpus, three builds interleaved, five runs each, medians: main 3.53s, literal gates only 3.30s, gates plus the `\b` removal 3.26s. About 8% overall, and most of it is the gates, since over 99% of chunks carry none of the literals.

Detection counts agree run for run within the corpus's own variance: repeat runs of any single build land between 103,861 and 103,867. The equivalence and fuzz tests above are what establish that decoding itself is unchanged.

### Checklist:


mcastorina (approved):
This looks good to me, thanks for sharing all the numbers and testing methodology. I see you added a `hasBackslash` check before a lot of the regex checks, could that have had an influence on the performance too?

Ahsan-Sarbaz:
Yes, and for ordinary chunks it is the bigger of the two changes. On a chunk with no escape sequences in it, the gate takes the decoder from about 50 µs to 0.49 µs, because five backslash regexes no longer run on a chunk that has no
backslash. Over 99% of chunks are like that, so most of the scan win comes from there rather than from the `\b` removal.

The `\b` removal covers the case the gate cannot: a chunk containing `U+` that has no real match, like the text `CPU+GPU`. The gate passes and the pattern runs anyway, and that is where removing `\b` takes it from 39.4 µs to 5.64 µs.

Whole scan, 200 MiB, three builds interleaved, five runs each, medians:

| build | scan |
|---|---|
| main | 3.53 s |
| literal gates only | 3.30 s |
| gates + `\b` removal | 3.26 s |

Two corrections to the description while I am here. It claims 11% on a full scan; that came from a noisier session and the interleaved runs above put it at about 8%. It also claims detections are identical at 103,864, which is stated too strongly: the count moves between 103,861 and 103,867 across repeat runs of main alone on this corpus. The equivalence and fuzz tests are the real evidence that decoding is unchanged. I will fix both lines.

PR #5322, opened 2026-09-16
feat(detectors): add Composio API key detector

### LINKED ISSUE #5321: Add Composio API key detector
### Summary
Request to add a detector for Composio API keys. I work at Composio (composio.dev) and am the point of contact for the key formats; a PR with the detector and pattern tests follows.

### Key formats
All three types are a fixed prefix followed by URL-safe nanoid characters `[A-Za-z0-9_-]`, no checksum:

- project keys: `ak_` + 20 characters
- organization keys: `oak_` + 20 characters
- user keys: `uak_` + 43 characters

`oak_` and `uak_` contain `ak_`, so the project pattern needs a non-alphabet character before the prefix. Keys can end in `-`, so `\b` alone misses about 1 in 64 of them; the patterns use an explicit non-alphabet boundary on both sides.

### Verification
Not in the first PR. Composio is finalizing a dedicated verification endpoint and will follow up with a second PR that adds it, plus sanctioned test credentials for the integration test through whichever secure channel you prefer.

groovyBugify:
### Description:
I work at Composio (composio.dev) and am submitting this detector on behalf of the company; I am the point of contact for the key formats. Closes #5321.

Composio issues three API key types, all a fixed prefix plus URL-safe nanoid characters `[A-Za-z0-9_-]`:

- project keys `ak_` + 20 characters
- organization keys `oak_` + 20 characters
- user keys `uak_` + 43 characters

`oak_` and `uak_` contain `ak_`, so each pattern requires a non-alphabet character before the prefix and after the key; `\b` alone misses keys that end in `-`. Results carry `key_type` in ExtraData. Keys have no checksum.

### Verification (01a1b2d)

Each key type authenticates through its own header and its own routes. A key sent under another type's header is answered with the same 401 as no header at all, so the endpoint cannot be shared: every type is checked on the cheapest read-only route it can authenticate.

| Key | Request |
|---|---|
| `ak_` | `GET /api/v3/toolkits/categories`, header `x-api-key` |
| `oak_` | `GET /api/v3/org/owner/project/list`, header `x-org-api-key` |
| `uak_` | `GET /api/v3/org/list?limit=1`, header `x-user-api-key` |

Composio answers a revoked key, an IP-allowlisted key, a banned caller address and a stripped header all with 401, so the verdict comes from `error.slug` in the JSON body, never from the status alone:

- 200: verified.
- 401 with a key-specific slug (`APIKey_InvalidAPIKey`, `APIKey_APIKeyExpired`, `UserApiKey_Unauthorized`, or the org key's `HTTP_Unauthorized` "Invalid x-org-api-key"): the hash matched no live row. Unverified, no error.
- 401 `Auth_Unauthorized` naming the key's IP allowlist, or any 403 carrying the envelope: the key resolved and a later check refused the request. Verified, with `restriction` in ExtraData.
- Anything else (a banned scanner address answers "Access denied", a WAF page has no envelope, 429, 5xx, network): undecided, surfaced as a verification error. The key never appears in the error text.

Only the cloud endpoint (`https://backend.composio.dev`) and endpoints passed with `--verifier composio=<url>` are called; URLs found in scanned data are never used. The scanner embeds `detectors.EndpointSetter` and implements `CloudProvider`, so self-hosted deployments work with `--verifier`.

One property of the API worth knowing: Composio bans a caller address after IP-allowlist denials on five distinct keys. Dead keys do not count, so scanning stale keys is safe. A scanner that hits five live IP-restricted keys from one address gets "Access denied" from then on, which this detector reports as undecided rather than as a verdict.

### Testing

- Unit: `composio_verify_test.go` runs against `httptest` and covers the request shape per key type (path, header, no other key header sent), every response class above, `verify=false` sending nothing, and endpoint fall-through.
- Integration (`-tags detectors`): `TestComposio_FromChunk` expects `COMPOSIO_PROJECT_KEY` and `COMPOSIO_PROJECT_KEY_INACTIVE` in `detectors5`. Test credentials from a dedicated Composio test org can be provided privately.
- Checked against production before pushing: a live project key verifies, a revoked project key and never-minted org and user keys come back unverified with no error. Scanning this repository's full history with the detector enabled finds only the four synthetic fixtures in `composio_test.go`.

The rotation guide now points at the public revocation endpoint's reference page; the previous URL was a 404.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Verification sends discovered keys to Composio over the network and uses nuanced slug-based classification, so mis-handled API responses could affect verified/unverified reporting for this detector only.
> 
> **Overview**
> Adds a **Composio** secret detector for three key shapes (`ak_`, `oak_`, `uak_`) with prefix/boundary-aware matching so org/user keys are not double-counted as project keys. Findings include `key_type`, a rotation doc link, and deduplicated tokens.
> 
> When verification is enabled, each key type is checked against `https://backend.composio.dev` (or configured endpoints only—never URLs from scanned data) via a dedicated header and minimal read-only route. Verdicts rely on Composio’s JSON error **slugs**, not status alone: explicit invalid/expired slugs mark dead keys; IP allowlist and permission-style 401/403 responses can still count as **live** with a `restriction` in ExtraData; ambiguous responses surface as verification errors without leaking the secret.
> 
> The scanner is registered in default detectors behind `ComposioDetectorEnabled` (enabled at startup like peer detectors) and `DetectorType_Composio` (1072) in the proto. Unit tests cover matching edge cases and httptest-driven verification behavior; a `detectors`-tagged integration test hits live Composio with GCP test credentials.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01a1b2db2671bc640fb548bb9f8e156e5c7a202d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/composio/composio.go
```diff
@@ -0,0 +1,73 @@
...
+// nanoid characters. oak_ and uak_ contain "ak_", so every pattern requires a
+// non-alphabet character (or start of input) before the prefix, and a
+// non-alphabet character (or end of input) after the key.
+type keyPattern struct {
+	keyType string
+	re      *regexp.Regexp
+}
+
+var keyPatterns = []keyPattern{
+	{keyType: "project", re: regexp.MustCompile(`(?:^|[^A-Za-z0-9_-])(ak_[A-Za-z0-9_-]{20})(?:[^A-Za-z0-9_-]|$)`)},
```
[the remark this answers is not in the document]
groovyBugify: Good catch, thanks. Fixed in 9324266: the trailing boundary is now checked in code on the match index (end of input or a non-alphabet byte) instead of a consuming group, so the delimiter stays available for the next match. Added a regression test for two keys separated by a single comma and one for a key glued to a following alphabet character.

pkg/detectors/composio/composio.go
```diff
@@ -0,0 +1,90 @@
...
+				DetectorType: detector_typepb.DetectorType_Composio,
+				Raw:          []byte(token),
+				SecretParts:  map[string]string{"key": token},
+				ExtraData: map[string]string{
+					"key_type":       pattern.keyType,
+					"rotation_guide": "https://docs.composio.dev/docs/api-keys",
+				},
+			})
+		}
+	}
```
shahzadhaider1, line 153: Detectors in trufflehog do two things: detection and verification. I don't see any secret verification in this detector.

shahzadhaider1 (commented):
This is great. Thanks for the contribution. I've added a small comment.

pkg/detectors/composio/composio.go
groovyBugify: Added in 01a1b2d. Each key type is verified on the cheapest read-only route it can authenticate, under its own header; the API answers a key sent under the wrong header the same way as no header at all, so the endpoint cannot be shared across types. The verdict comes from the error slug in the body rather than the status code, because Composio uses 401 for a revoked key, an IP-allowlisted key and a banned caller address alike: key-specific slugs come back unverified with no error, an allowlist denial or a 403 carrying the envelope comes back verified with a `restriction`, and anything else is left as a verification error rather than a guess. Unit tests run against httptest. The integration test behind the `detectors` tag needs `COMPOSIO_PROJECT_KEY` and `COMPOSIO_PROJECT_KEY_INACTIVE` in `detectors5`; test credentials can be provided through whichever channel you use. The full mapping is in the PR description.

PR #5323, opened 2026-09-16
Add Okta OAuth support in Okta detector

shreyas-sriram:
### Description:

Implements the `// TODO: Oauth client secrets` in `pkg/detectors/okta/okta.go` — the Okta detector previously only found SSWS API tokens (`domain` + `00...` token). This adds detection and verification for Okta OAuth 2.0 client credentials (`client_id` + `client_secret`) as well.

- New patterns detect Okta app/client IDs (`0oa` + 17 base62 chars, matching Okta's own object-ID format) and client secrets (40–64 chars, anchored on a nearby `client_secret`/`clientsecret`/`client-secret` keyword to avoid false positives).
- Verification exchanges the `client_id`/`client_secret` pair for a token via the `client_credentials` grant against `https://{domain}/oauth2/default/v1/token`. A `400 unauthorized_client` response is also treated as verified: Okta authenticates the client before checking whether the grant type is authorized for the app, so that response still confirms the secret itself is valid — it just means the app isn't configured as a service app for this grant.
- Added a `CustomResultsCleaner` implementation so that when result cleaning runs (e.g. `--filter-unverified`), the pre-existing SSWS-token findings and the new OAuth-credential findings don't collapse into a single arbitrary result — each distinct secret identity is preserved.
- Regression tests cover the new pattern (including a boundary edge case where a secret starting/ending in `-` previously failed to match or was silently truncated) and the OAuth verification response handling (200/400 `unauthorized_client`/400 `invalid_client`/401/unexpected status).

### Checklist:
* [x] Tests passing (`go test ./pkg/detectors/okta/...`)
* [x] Lint passing (`golangci-lint run --enable bodyclose,copyloopvar,misspell --timeout 10m ./pkg/detectors/okta/...` — 0 issues)

---

> [!NOTE]
> **Medium Risk**
> Adds live outbound OAuth token requests using discovered credentials and changes which findings survive result cleaning, though scope is limited to the Okta detector with regression tests.
> 
> **Overview**
> Extends the **Okta** detector beyond SSWS API tokens (`domain` + `00…` token) to also find **OAuth 2.0 client credentials** (`client_id`, `client_secret`, and `domain`).
> 
> New regexes match Okta app IDs (`0oa` + 17 chars) and keyword-anchored client secrets (40–64 chars), with delimiter-based boundaries so secrets that start or end with `-` are not missed. When verification is enabled, OAuth pairs are checked via `client_credentials` against `https://{domain}/oauth2/default/v1/token`; a **400** body containing `unauthorized_client` counts as verified because Okta validates the secret before grant-type authorization.
> 
> SSWS results now get a **Redacted** identity (`domain:first4…last4`). The scanner implements **`CustomResultsCleaner`** so SSWS and OAuth hits in the same chunk are deduped per `Redacted` identity instead of collapsing to a single generic cleaned result. Tests cover OAuth patterns (including hyphen edge cases), verification response handling, and custom cleaning behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2159aeede90d015497f57d34322893e7bb7cf73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/detectors/okta/okta.go
```diff
@@ -111,6 +160,77 @@ func verifyOktaToken(ctx context.Context, client *http.Client, domain, token str
...
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		body, _ := io.ReadAll(resp.Body)
+		return strings.Contains(string(body), "\"access_token\""), nil
+	case http.StatusBadRequest:
+		body, _ := io.ReadAll(resp.Body)
+		// Okta authenticates the client before checking grant-type authorization, so
+		// "unauthorized_client" still confirms client_secret is valid.
+		return strings.Contains(string(body), "unauthorized_client"), nil
```
martinlocklear: `return strings.Contains(string(body), "unauthorized_client"), nil`

I'm worried about false negatives here. This treats any 400 that isn't unauthorized_client as proof the secret is bad, but some 400s say nothing about the secret at all - an org without API Access Management has no /oauth2/default, so it returns a feature-not-enabled error and a live secret comes back as a negative. Same if Okta adds an error code later. I think we might want to follow [a pattern like the box detector](https://github.com/trufflesecurity/trufflehog/blob/b2159aeede90d015497f57d34322893e7bb7cf73/pkg/detectors/boxoauth/boxoauth.go#L154-L160) and return an error when we don't find a key word we recognize. That at least would make this not fail silently.

pkg/detectors/okta/okta.go
```diff
@@ -111,6 +160,77 @@ func verifyOktaToken(ctx context.Context, client *http.Client, domain, token str
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		body, _ := io.ReadAll(resp.Body)
```
martinlocklear: I think we might want to propagate this error up, so that a truncated body surfaces as a verification error instead of a unverified result. A 200 here means Okta issued a token, so a failed read currently turns a near-certain hit into a negative. [Same at line 193.](https://github.com/trufflesecurity/trufflehog/pull/5323/changes#diff-449b7e95081c2079083b9b08d4eae00e24f02b7b0bb522c7868437b2674ca26eR193)  What do you think?

martinlocklear (changes requested):
Hi @shreyas-sriram !  Thanks for putting this PR together - apologies for the delay in review.  I'm jumping into some new territory here and I needed to really sink more into the subject. 

Added notes on a couple of small things that I think are worth addressing before merging this in - let me know if you disagree and we can re-evaluate.

pkg/detectors/okta/okta.go
shreyas-sriram: Good call — adopted the boxoauth pattern in 9428940: `unauthorized_client` → verified, `invalid_client` → unverified, and any other 400 body now returns a verification error instead of failing silently.

pkg/detectors/okta/okta.go
shreyas-sriram: Agreed — fixed in 9428940. Read errors are now propagated on both the 200 and 400 paths so they surface as verification errors.

martinlocklear (approved):
Looks good to me!

PR #5325, opened 2026-09-17
give git the real base..head range for diff scans

kvnphvm:
### Description:
A diff scan (--since-commit) could skip commits and still exit sucessfuly. On a branch that had merged its base branch in on any PR that stays open long enough to pull in main, commits before that merge were never scaned, so secrets introduced in them went unreported with no warnings or errors.

The fix is to get the range directly and delete the read-until-base logic. Only scans that specify a base are affected, full history scans are unchanged.

### Checklist:

---

> [!NOTE]
> **High Risk**
> Changes which commits are scanned for every git diff scan (`--since-commit`), directly affecting secret detection; incorrect range logic previously caused silent misses.
> 
> **Overview**
> Fixes **silent gaps in `--since-commit` / diff scans** (INT-1054) where commits on a feature branch could be skipped—especially after merging main in—while the scan still succeeded.
> 
> **Git log construction** now passes an explicit **base** into `gitparse` and builds **`head` + `^base`** so Git defines **`base..head`**, instead of walking history and stopping when the base commit appears (which broke when commit dates or merge topology didn’t match log order). Full-history scans are unchanged; diff scans no longer apply `--diff-filter=AM` so deletions in the range are included. **`--since-commit` with no `--branch`** stays **`--all ^base`** (all refs since the base), not an implicit `HEAD`.
> 
> The git source **drops the in-loop “reached base commit” break**, relies on the parsed range, and logs when a diff range has **no commits**. **Pre-commit** sets **`BaseRef` and `HeadRef` to `HEAD`** so the commit walk is empty and only staged changes are scanned.
> 
> Large regression coverage: merged-base topology, base-only vs explicit head, pre-commit, bare mirrors, shallow clones, invalid bases, and swapped range ends.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2e84b2dfab9079fbdc069e3ea520a648ea72401. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/gitparse/gitparse.go
```diff
@@ -449,7 +459,12 @@ func (c *Parser) prepGitArgs(source string, head string, abbreviatedLog bool, ex
 	// https://git-scm.com/docs/git-log#Documentation/git-log.txt---all
 	args.log = append(args.log, cmp.Or(head, "--all"))
 
-	// And then potentially add -- to args here
+	// `^base head` is base..head, and unlike the two-dot form it composes with --all.
+	// https://git-scm.com/docs/gitrevisions#_specifying_ranges
+	if base != "" {
+		args.log = append(args.log, "^"+base)
+	}
```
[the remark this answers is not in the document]
kvnphvm: i agree

pkg/gitparse/gitparse.go
```diff
@@ -445,11 +455,29 @@ func (c *Parser) prepGitArgs(source string, head string, abbreviatedLog bool, ex
...
+
+	// `head ^base` is base..head. Keeping the two revisions as separate args
+	// means head is always the positive end of the range and the base is
+	// never spliced into a string git has to parse.
+	// https://git-scm.com/docs/gitrevisions#_specifying_ranges
+	if base != "" {
+		args.log = append(args.log, "^"+base)
+	}
 
-	// And then potentially add -- to args here
```
casey-tran, line 452: Did you find this comment misleading? I'm just wondering why it is being taken out.

pkg/gitparse/gitparse.go
```diff
@@ -445,11 +455,29 @@ func (c *Parser) prepGitArgs(source string, head string, abbreviatedLog bool, ex
 		args.show = append(args.show, "--diff-filter=AM")
 	}
 
-	// Keep head or all last, before the --, not required but sensible
+	// The positive end of the walk, kept last before the -- (not required but
+	// sensible). A base with no head is a diff scan up to the checked-out
+	// commit: the pre-commit hook and `--since-commit X` without `--branch`.
```
casey-tran, line 460: Do you think it would be easier to read if the comment:
```
A base with no head is a diff scan up to the checked-out commit: the pre-commit hook and `--since-commit X` without `--branch`.`
```
Went to the specific `base != "":` case?

casey-tran (approved):
Looks good to me, but I'd wait for someone from Integrations to do a review too. Left a few comments and questions.

pkg/gitparse/gitparse.go
kvnphvm: I particularly did not find value in the comment, but I can add it back. It's a fair argument.

kvnphvm: How about this?

Pathspecs live in their own slice so `--` always trails every revision.

pkg/gitparse/gitparse.go
kvnphvm: agreed, can move for better readability

pkg/gitparse/gitparse.go
```diff
@@ -445,11 +455,29 @@ func (c *Parser) prepGitArgs(source string, head string, abbreviatedLog bool, ex
...
+	// Defaulting to --all there would walk every ref not reachable from base,
+	// which is not a diff of the change being made.
 	// https://git-scm.com/docs/git-log#Documentation/git-log.txt---all
-	args.log = append(args.log, cmp.Or(head, "--all"))
+	switch {
```
MuneebUllahKhan222, line 523: Concern: `--since-commit X` without `--branch` loses coverage of other refs

Consider this history:

```
      o1 - o2         other
     /
a - b - c             main   <- HEAD is main
```

Previously --since-commit a` (no `--branch`) ran `git log --all` and stopped when it reached `a`. In this history that scanned `b, c, o1, o2`.

After the changes introduced in this PR it runs `git log HEAD ^a`, which scans only `b, c`. Commits that exist only on other branches (`o1`, `o2`) are no longer scanned. Users who rely on "everything since `a` across all refs" (mirror/bare clones, remote URL scans) would lose coverage silently, with exit code 0.

What we can do is make the hook explicit instead of changing the default.

In the pre-commit override in `main.go`:

```go
gitCfg.BaseRef = "HEAD"
gitCfg.HeadRef = "HEAD" // hook range is explicitly HEAD..HEAD, not "all refs"
```

In `prepGitArgs`, drop the `case base != "": HEAD` branch and keep `--all` as the default when head is empty:

```go
switch {
case head != "":
    args.log = append(args.log, head)
default:
    args.log = append(args.log, "--all")
}
if base != "" {
    args.log = append(args.log, "^"+base)
}
```

| Invocation | Command | Result |
|---|---|---|
| Pre-commit hook | `git log HEAD ^HEAD` | empty (staged changes are covered by `ScanStaged`, as before) |
| `--since-commit a` (no `--branch`) | `git log --all ^a` | `c b o2 o1` (old coverage kept) |
| `--since-commit a --branch main` | `git log main ^a` | `c b` |

This keeps the previous behavior and the hook working. Git still computes the range, so the merge/date-ordering fix from this PR still applies to every case.

pkg/sources/git/git.go
```diff
@@ -1037,6 +1034,10 @@ func (s *Git) ScanCommits(ctx context.Context, repo *git.Repository, path string
 			return err
 		}
 	}
+
+	if scanOptions.BaseHash != "" && depth == 0 {
+		logger.V(1).Info("no commits in range", logValues...)
```
MuneebUllahKhan222: Could we log "no commits in range" at Info instead of V(1)? V(1) isn't printed at the default log level, so an empty base..head range would be invisible.

MuneebUllahKhan222 (changes requested):
It took me longer than expected to review this PR carefully.

After going through it in detail, I’ve come to the conclusion that the comments below are blocking and should be addressed before merging this PR.

I could be wrong about some of my comments, though, so if anything feels off or unclear, please let me know and we can discuss it.

MuneebUllahKhan222 (approved):
LGTM!

PR #5327, opened 2026-09-17
Add OAuth2 token acquisition and verification for custom detectors

unsmith:
### Description:

Add support for OAuth2-authenticated verification endpoints in custom detectors. When a custom detector config includes an `auth` block with OAuth2 credentials, TruffleHog acquires a token via the configured grant type (currently ROPC per RFC 6749 §4.3) and uses it to authenticate verification requests.

**Key changes:**

- **Proto schema** (`custom_detectors.proto`): New `VerifierAuth`, `OAuth2Config`, and `ROPCConfig` messages. The optional `auth` field on `VerifierConfig` is backward-compatible — existing configs without it work unchanged.
- **Token acquisition** (`custom_detectors.go`): `BuildTokenSource` constructs an `oauth2.ReuseTokenSource` with automatic expiry-aware caching and refresh. Token endpoint URLs are validated against the same `unsafe` flag as verification endpoints.
- **Shared verification loop** (`verify.go`): Single `VerifyCredential` function used by both the inline custom detector path and the engine-level OAuthVerifier path. Supports customer-defined request body templates with `$token`/`$secret`/`$detector_type`/`$detector_name` substitution, custom headers, and HTTP status code interpretation via `successRanges`/`rotatedRanges`.
- **Engine routing** (`engine.go`, `oauth_verifier.go`): When a detector satisfies `OAuthVerifier` and has a token source, verification bypasses the built-in path. Detection runs with `verify=false`, then results are verified via the shared `VerifyCredential`.
- **Correlation logging**: `TracedTokenSource` carries a single correlation ID (`oauth2_trace`) that flows through all token acquisition and verification log messages via context enrichment.
- **Interface additions** (`detectors.go`, `endpoint_customizer.go`): `OAuthVerifier` interface, `OAuth2TokenSource` type alias, `TracedTokenSource` wrapper, and `EndpointSetter` extensions.

All non-auth code paths are completely unchanged. The `auth` field is optional at the proto level; detectors without it follow the existing verification logic with no behavioral difference.

### Checklist:
- [ ] Lint passing (`make lint` this requires [golangci-lint](https://golangci-lint.run/welcome/install/#local-installation))?


pkg/custom_detectors/custom_detectors.go
```diff
@@ -283,91 +514,39 @@ func (c *CustomRegexWebhook) createResults(ctx context.Context, match map[string
...
-		result.SetVerificationError(errors.New("verification response status code did not match any configured successRanges or rotatedRanges"))
+	outcome := VerifyCredential(ctx, httpClient, endpoints, defaultBody, vars)
+	result.Verified = outcome.Verified
+	if outcome.RespBody != "" {
+		result.ExtraData["response"] = outcome.RespBody
+	}
+	if outcome.Attempted && !outcome.Definitive {
+		// Verification was configured but no endpoint gave a conclusive
+		// answer — either all transports failed or no range matched.
+		result.SetVerificationError(errors.New("verification attempted but no endpoint gave a definitive answer"))
```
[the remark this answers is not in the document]
unsmith: This is intentional.  There was a bug here, on main when all verification endpoints are unreachable the secret was reported as unverified with no error, which is indistinguishable from unverified _because the verification endpoint said so_.  It's now unknown with a verification error - an indeterminate state that should allow checking again when the scanner re-encounters the secret.

pkg/custom_detectors/custom_detectors.go
```diff
@@ -25,11 +26,205 @@ import (
 // for poorly defined regexps.
 const maxTotalMatches = 100
 
+// defaultTokenLifetime is applied when the IdP omits expires_in from
+// the token response. Five minutes is conservative enough to avoid
+// sending stale tokens during long scans while keeping token endpoint
+// traffic reasonable.
+const defaultTokenLifetime = 5 * time.Minute
```
camgunz, line 33: **question:** I thought the max lifetime was 5m, should we make this 4m to be on the safe side?

pkg/custom_detectors/custom_detectors.go
```diff
@@ -25,11 +26,205 @@ import (
...
+	// detection context because oauth2.TokenSource.Token() accepts no
+	// context parameter — there is no way to thread a caller-provided
+	// context through the interface. The SaneHttpClient timeout bounds
+	// the worst case for a stuck token endpoint.
+	httpCtx context.Context
+}
+
+func (s *ropcTokenSource) Token() (*oauth2.Token, error) {
+	logger := logContext.AddLogger(s.httpCtx).Logger()
+	logger.Info("requesting ROPC token from endpoint",
```
camgunz, line 59: **suggestion:** I think we should bump this down to level 2; it sounds like we'll be doing it every few minutes or so

pkg/custom_detectors/custom_detectors.go
```diff
@@ -25,11 +26,205 @@ import (
...
+	// the IdP omits it, oauth2 leaves Expiry as zero — which
+	// ReuseTokenSource treats as "never expired," caching the token
+	// for the lifetime of the process. Force a conservative default
+	// so stale tokens get refreshed during long scans.
+	if tok.Expiry.IsZero() {
+		tok.Expiry = time.Now().Add(defaultTokenLifetime)
+		logger.Info("IdP did not return expires_in; using default token lifetime",
+			"default_lifetime", defaultTokenLifetime,
+		)
+	}
```
camgunz, line 82: **praise:** love it 🏆

pkg/engine/engine.go
```diff
@@ -1170,12 +1183,51 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
...
+		// bypass the built-in verification path entirely: detect only,
+		// then verify each result via OAuth2-authenticated POST. This
+		// intentionally skips verificationCache — the built-in cache
+		// wraps FromData with verify=true, which would run the
+		// detector's unauthenticated verification. The OAuth path
+		// replaces that with its own authenticated verification loop.
+		// Token caching is handled separately by oauth2.ReuseTokenSource.
+		var results []detectors.Result
+		var err error
+		if oauthV, ok := data.detector.Detector.(detectors.OAuthVerifier); ok && oauthV.HasOAuth2() {
```
camgunz, line 1195: **issue:** Regardless of the answer to the above, I think we don't want to bypass the verification cache, which I guess would mean at least moving this logic from here to [there](https://github.com/trufflesecurity/trufflehog/blob/df5b7ce7ec895edc2fc67ae9929f3468216ec85e/pkg/verificationcache/verification_cache.go#L53).

pkg/engine/engine.go
```diff
@@ -1170,12 +1183,51 @@ func (e *Engine) detectChunk(ctx context.Context, data detectableChunk) {
...
-		results, err := e.verificationCache.FromData(
-			ctx,
-			data.detector.Detector,
-			data.verify,
-			data.chunk.SecretID != 0,
-			matchBytes)
+
+		// When the detector has OAuth2 auth on its custom verifier,
+		// bypass the built-in verification path entirely: detect only,
+		// then verify each result via OAuth2-authenticated POST. This
```
camgunz: **question:** I'm curious about the `POST`, like do we also have to modify the request we're making?

And at a higher level, I'm also curious why we're doing this here in detection. I would've guessed that we'd add a method like `SetHTTPClient` to [`EndpointCustomizer`](https://github.com/trufflesecurity/trufflehog/blob/df5b7ce7ec895edc2fc67ae9929f3468216ec85e/pkg/detectors/detectors.go#L88) where we could pass an HTTP client that was pre-authenticated and call it alongside [`SetConfiguredEndpoints`](https://github.com/trufflesecurity/trufflehog/blob/df5b7ce7ec895edc2fc67ae9929f3468216ec85e/pkg/engine/engine.go#L325) if we had OAuth info for that detector's verifier.

pkg/custom_detectors/custom_detectors.go
unsmith: There actually isn't a strict max lifetime for OAuth2 tokens.  It can vary depending on provider, and this default only kicks in when that provider omits `expires_in`.  When `expires_in` is set, we respect that.

pkg/custom_detectors/custom_detectors.go
unsmith: I mostly agree, _but_ I wanted to keep this chatty for the short term while we work out any bugs.  It would only be emitting this while a custom detector is running, which shouldn't be all that often.

pkg/engine/engine.go
unsmith: That's an interesting idea, I'll look more into this.  I agree that bypassing the verification cache isn't ideal and potentially churny.

unsmith: And re: the POST - we do potentially have to modify the request, yes.  If the custom verifier endpoint behind OAuth2 is "standard", as in it works with the same request shape as a normal verifier, then modifying the stock verifier's request body isn't necessary.  BUT if the custom verifier is completely custom so we don't know the request shape at all, that's the case where we need to hand-roll our own POST.

We could do this where if the request shape is the standard one for the detector (no config override) then we inject the authenticated client like you suggested, but we'd still need the custom path anyway, so that doesn't seem worth it.  Having this check the verification cache is definitely doable though, so I'll add that.

unsmith: I got sidetracked, but this now includes a callback for "external" verifiers (so like, shunting off to an OAuth something-or-other) to use the verification cache like a non-custom verifier would.  That should result in cache hits if the same secret is detected in a later chunk.

PR #5331, opened 2026-09-21
fix(gitparse): cheaper low-memory git scan, stop losing trailing commits

kashifkhan0771:
### Description:
Listing commits used `git log`, which diffs every commit just to print hashes; `git rev-list` does it far cheaper. Hashes now go in on stdin, so groups are not capped by command length. Bad revisions return an error instead of looking like an empty repo, and a scan that stops early now kills its git processes. Also fixes a parser bug: a commit with no diffs was lost when last in a stream

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes core secret-scanning git history ingestion; regressions could miss commits or hang scans, though extensive parity tests mitigate this.
> 
> **Overview**
> Fixes **OOM on huge histories** when `--git-low-memory-scan` is enabled by splitting one long `git log --patch` into **`git rev-list`** (commit hashes only) plus **short-lived `git log --no-walk=unsorted --stdin`** runs in groups of **5000** full hashes fed on stdin—not batched `git show` on the command line.
> 
> **Correctness and parity** with the single-process path: **`--diff-filter=AM`** stays on patch args only; **`cleanupParse`** now emits **commits with no file diffs** at the **end of each group stream** (empty commits / message-only secrets); bad refs **fail at `RepoPath`** instead of looking like an empty repo. The git source **`ScanCommits`** wraps a **cancelable context** so max-depth / base-commit early exit **stops background git processes**.
> 
> Adds **`pkg/gitparse/README.md`** and **`lowmemory_test.go`** (byte-for-byte match vs single command, globs, cancel, trailing empty commit).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39f9d1f4d3653a4d4bbb685f16175bb6d0e8e04a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashifkhan0771:
A big thanks to @mariduv for doing 90% of the work to fix the [git OOM issue](https://github.com/trufflesecurity/trufflehog/pull/5257). This PR is the remaining 10%. It also fixes an older parser bug that batching made visible, where a commit with no diffs was dropped when it was last in a stream.

mariduv (approved):

PR #5336, opened 2026-09-21
fix(analyzers/github): prevent panics on unrecognized permission scopes

### LINKED ISSUE #5335: [Bug] GitHub Analyzer panics when encountering unrecognized permission scopes
**Describe the bug**
The GitHub analyzer will panic during token analysis if it encounters an API response that contains unrecognized permission scopes or if certain type assertions fail. This causes the entire 	rufflehog process to crash.

Specifically:
1. In pkg/analyzer/analyzers/github/classic/classictoken.go, Permission.ToString() returns an error for unknown permissions, which is then handled with panic(err). As GitHub frequently adds new OAuth scopes, any encountered new scope will crash 	rufflehog.
2. In pkg/analyzer/analyzers/github/finegrained/finegrained.go, there are panic calls for failed type assertions on RepoAccessMap and UserAccessMap.
3. In pkg/analyzer/analyzers/github/finegrained/finegrained.go, an unexpected 200 OK status when testing for content write permissions triggers a panic("This should never happen...").

**To Reproduce**
1. Provide 	rufflehog with a valid GitHub classic token that has a newly introduced permission scope not yet mapped in PermissionStrings.
2. The GitHub analyzer runs and attempts to format the scopes.
3. 	rufflehog crashes with a panic.

**Expected behavior**
The analyzer should handle unexpected API responses and unknown scopes gracefully. Unknown scopes should be formatted with a fallback string, and unexpected HTTP responses or types should result in an error or safe fallback rather than crashing the entire scanner.

Gracy769:
Fixes #5335

Replaces several panic calls in the GitHub analyzers with safe error handling and fallback values. This prevents the scanner from crashing when:
1. GitHub introduces new OAuth scopes that aren't mapped yet in classic_permissions.go.
2. The GitHub API returns an unexpected 200 OK when testing content write permissions with an invalid payload.
3. Expected type assertions on RepoAccessMap and UserAccessMap fail.

---

> [!NOTE]
> **Low Risk**
> Changes are defensive error handling and status-code mapping in analyzer output paths; they do not alter secret detection or credential handling.
> 
> **Overview**
> Hardens the GitHub token analyzers so unexpected API or mapping data no longer crashes the scanner.
> 
> In the **classic** analyzer, failed `Permission.ToString()` calls now emit a placeholder like `unknown_permission_<id>` instead of panicking when printing scopes and permission lists.
> 
> In the **fine-grained** analyzer, content-write probes treat **200/201** as write access instead of aborting via `log.Fatal` or `panic`, and permission tables are only rendered when `RepoAccessMap` / `UserAccessMap` type-assert to `[]Permission`—otherwise those sections are skipped rather than panicking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f01af1a8b4eb0594a0873a7172687f92e04fe03f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Gracy769:
I've addressed the feedback from Cursor Bugbot! The early returns that were skipping output have been changed to \if ok\ blocks, and the unexpected 200/201 statuses correctly return \ContentsWrite, nil\. I also found and removed a hidden \log.Fatal()\ in that same block which would have also crashed the process.

